PR #2173, opened 2022-12-21
Support/coin framework

sprohaszka-ledger:
### 📝 Description

This PR aims to separate the coin logic into a different package, to ease the separation of the coin families.

With this PR:
* Coin-framework has all the logic needed to implement a new coin family. Duplicate code in live-common has been removed. Live-common depend on this new package.
* Coin-polkadot has all the coin logic of Polkadot. It depends mainly on coin-framework and has no link with live-common.
* Bridge-framework is not usable for the time being. It is not used by live-common. It is a draft of what should be the sole entry point of the live for all coin interaction

**Side effect**: to be able to switch to a strict Typescript compilation, `lru-cache` has been updated. The impact is the replacement of the deprecated properties `maxAge` to its new name `ttl`.

__Coin-Framework__
* Operation.ts : Remove NFT. Keep it in Live-common
* ~Error.ts : Remove unused errors. Keep the rest in Live-common~
* Derivation.ts : Move types from live-common/hw/getAddress/types.ts. Maybe should remove from live-common either.
* account/support::canSend : What to do?
* account/helpers : What to do?
* env.ts : What to do?
* network.ts : What to do?

__Coin-Polkadot__
* WalletAPI (PlatformSDK) dependency : What to do?
* Remove react.js file : DONE
* spec.ts : link to the bot. What to do?
* cli-transaction : What to do?
* bridge.integration.ts : What to do?

__Bridge-Framework__
* Mock : remove from the time being because not used in Polkadot
* MockHelper : Removed. DONE
* getDeviceTransactionConfig : UI functions. To Remove
* Add resources conversion into toAccount and fromAccoutRaw
* Swap types : moved from exchange/swap/types.ts. Is it OK?
* useBridgeTransaction : linked to React ; to remove. DONE.

__Live-Common__

### ❓ Context

- **Impacted projects**: `live-common` 
- **Linked resource(s)**: [LIVE-4362](https://ledgerhq.atlassian.net/browse/LIVE-4362)

### ✅ Checklist

- [x] **Test coverage** 
- [ ] ~~**Atomic delivery**~~ 
- [x] **No breaking changes** 

Tests that have been launched:
* Launching integration test on Algorand, Cosmos and Polkadot
* Bot testing on Algorand, Cosmos and Polkadot

### 🚀 Expectations to reach

Test commands that need to be checked:
 * ```pnpm common jest --runTestsByPath src/families/cosmos/bridge.integration.test.ts```
 * ```pnpm run:cli bot -c algorand```

[LIVE-4362]: https://ledgerhq.atlassian.net/browse/LIVE-4362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-framework/src/account/helpers.ts
```diff
@@ -0,0 +1,414 @@
...
+
+    default:
+      throw new Error("invalid account.type=" + (account as AccountLike).type);
+  }
+};
+
+export const isAccountEmpty = (a: AccountLike): boolean => {
+  // if (a.type === "Account" && a.currency.family === "tron") {
+  //   const tronAcc = a as TronAccount;
+  //   // FIXME: here we compared a BigNumber to a number, would always return false
```
gre, line 97: this comment was actually wrong. `BigNumber("0").eq(0)` is true

libs/coin-framework/src/account/helpers.ts
```diff
@@ -0,0 +1,414 @@
...
+  //   return (
+  //     tronAcc.tronResources && tronAcc.tronResources.bandwidth.freeLimit.eq(0)
```
gre, line 99: by removing this code, I can see we will have a regression here on considering some cases of Tron account non empty again. (TBD what it exactly means)

libs/coin-framework/src/account/helpers.ts
```diff
@@ -0,0 +1,414 @@
...
+  // if (copy.currency.family === "tron") {
+  //   const tronAcc = copy as TronAccount;
+  //   tronAcc.tronResources = {
+  //     ...tronAcc.tronResources,
+  //     cacheTransactionInfoById: {},
+  //   };
+  // }
+  // if (copy.currency.family === "bitcoin") {
+  //   (copy as BitcoinAccount).bitcoinResources = initialBitcoinResourcesValue;
+  // }
```
gre, line 165: we need a solution to allow the coins to define a clear cache. this is farely important, especially for the "clear" action to works on bitcoin family.

libs/coin-framework/src/account/helpers.ts
```diff
@@ -0,0 +1,414 @@
...
+  //     return (mainAccount as TronAccount)?.tronResources?.votes.length || 0;
+  //   case "osmosis":
+  //   case "cosmos":
+  //     return (
+  //       (mainAccount as CosmosAccount)?.cosmosResources?.delegations.length || 0
+  //     );
+  //   default:
+  //     return 0;
+  // }
+  return 0;
```
gre, line 260: this is a regression for analytics

libs/coin-framework/src/account/support.ts
```diff
@@ -0,0 +1,111 @@
...
+// ): boolean {
+//   try {
+//     getAccountBridge(account, parentAccount).createTransaction(
+//       getMainAccount(account, parentAccount)
+//     );
+//     return true;
+//   } catch (e) {
+//     return false;
+//   }
+// }
```
gre, line 61: some code commented, where did this go then? (just a PR comment to remind we will need to either remove the commented code or figure out how to put it back)

libs/coin-framework/src/env.ts
```diff
@@ -0,0 +1,790 @@
+import mapValues from "lodash/mapValues";
```
gre, line 1: ok sorry for missing this before. But I think this file don't belong to `coin-framework` as it contains envs that are not even related to coins.
 I suggest that we make a dedicated lib `env` so we can use `@ledgerhq/env` instead. (quid to guarantee a dedup of this lib, because it's stateful, we have a memory state that changes with setEnv)

libs/coin-framework/types/compressjs/index.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '@ledgerhq/compressjs' {
```
gre, line 1: note to ourself that we may need to extract the Bzip2 out of compressjs to type it properly in future & not maintain this fork. cc @valpinkman @jelbaz-ledger too

libs/ledger-live-common/src/families/polkadot/bridge/js.ts
```diff
@@ -1,57 +1,10 @@
...
-  sync,
-  receive,
-  signOperation,
-  broadcast,
-};
 export default {
-  currencyBridge,
-  accountBridge,
+  currencyBridge: buildCurrencyBridge(withDevice),
+  accountBridge: buildAccountBridge(withDevice),
```
gre, line 9: clean 👍

gre (changes requested):
going through the code again & more extensive review. 

I can see a few regression because of some commented code. 
I don't doubt things were done carefully but I also don't have the ability to do a "diff" on the moved code, so there may be more that I couldn't have seen if we would have removed the code (so kudos on keeping the code commented, but i think it could be valuable for a code reviewer to do a more careful second review)

I can see moving a coin to a lib will require some changes both in term of creating a package as well as converting some code to use the "build with communication" paradigm.
I expect this PR to also document:
- how to migrate code
- the relevant update it would mean in the dev portal

libs/coin-framework/src/account/helpers.ts
sprohaszka-ledger: So you mean, I can safely remove the // FIXME comment?

libs/coin-framework/src/account/helpers.ts
sprohaszka-ledger: I keep a `clearAccount` method instance in *live-common* to manage this specific case.

libs/coin-framework/src/account/helpers.ts
sprohaszka-ledger: I remove this method from *coin-framework*. It seems not to be useful in this package for the time being.

libs/coin-framework/src/account/support.ts
sprohaszka-ledger: I remove this method from *coin-framework*. I leave its declaration inside *live-common*

libs/coin-framework/src/env.ts
sprohaszka-ledger: Unfortunately, this file contains URL for coin family request. There are also variable like `DEBUG_HTTP_RESPONSE` used inside the `network.ts` file.
I fully agree that we should put this file elsewhere. But I afraid the consequences of refactoring it right now will bring too much work for this PR.
WDYT? @gre

gre: indeed. can you prepend in that file a comment?

```
// TODO: this file will be moved to a @ledgerhq/env library
```

I can help working on this after the PR lands.

gre (approved):

PR #2174, opened 2022-12-21
[BUGFIX] make sure currencies param of requestAccount is string array

chabroA:
### 📝 Description

If the currencies array passed to `requestAccount` contains other values than string, the underlying function crashes (cf. demo). 

Make sure currencies param of requestAccount is a string array.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

<details>
<summary>KO 💥 </summary>

https://user-images.githubusercontent.com/9203826/208966317-7a0a5d13-11ee-4e9c-bf53-67a099a5e05a.mov

</details>

<details>
<summary>OK ✅  </summary>

https://user-images.githubusercontent.com/9203826/208966376-6bfbf50d-d7f8-4f2d-b121-207e6c091f85.mov

</details>

### 🚀 Expectations to reach

LL should not crash even if invalid values passed as param for live-app-sdk functions

### DIFF
--- a/.changeset/lazy-sloths-obey.md
+++ b/.changeset/lazy-sloths-obey.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Fix app crash in case requestAccount currencies array param contains non string values
--- a/.changeset/spotty-boxes-battle.md
+++ b/.changeset/spotty-boxes-battle.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Fix app crash in case requestAccount currencies array param contains non string values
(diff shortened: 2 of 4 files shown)


.changeset/five-dragons-jog.md
```diff
@@ -0,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"live-mobile": patch
```
Justkant, line 3: You'll need to split this changeset

Justkant (approved):
IIUC it shouldn't be an issue with the wallet-api because we use zod to validate at runtime right ?

Justkant:
You have a lint error with prettier, just need a quick format on the webview file

chabroA:
> IIUC it shouldn't be an issue with the wallet-api because we use zod to validate at runtime right ?

Indeed it should not. In this case it should throw an error on the wallet-api-client side at runtime (so at worst it's the live app that will crash if they don't handle errors for example) 
Wallets should be safe 👍

PR #2176, opened 2022-12-22
bugfix/FAT-748 Fix  IL fw bug that broke access to My Ledger

juan-cortes:
### 📝 Description

Original While investigating [LIVE-4506](https://ledgerhq.atlassian.net/browse/LIVE-4506) which doesn’t seem to be happening anymore (after this unrelated fix) I noticed that with [FAT-690](https://ledgerhq.atlassian.net/browse/FAT-690) we introduced a loop when accessing the manager.

Basically, when we detect the device model to be Stax, we send an apdu as part of listing apps which fetches the size of the custom lock screen. Since this IL firmware was released before the feature was developed, the firmware doesn’t know how to handle that apdu and throws an error.

This handled that edge case that would prevent Nano X IL devices from accessing in the manager.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/FAT-748` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No need for a demo

### 🚀 Expectations to reach
Since this was originally supposed to be an investigation on [LIVE-4506](https://ledgerhq.atlassian.net/browse/LIVE-4506), I would say that a good test plan is accessing My Ledger with both an il firmware and a real stax device, to ensure that they both work.


alexandremgo (approved):
LGTM

PR #2179, opened 2022-12-22
Feat: add dex deep link LIVE-3788

sarneijim:
### 📝 Description

As a user on LLD, I want to have my information routed to the DEX iframe

### ❓ Context

- **Impacted projects**: 
    - LLD
    - Common
  
- **Linked resource(s)**: 
   - [swap] As a user on LLD, I want to have my information routed to the DEX iframe https://ledgerhq.atlassian.net/browse/LIVE-3788

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

Avoid fill in 2 places the same form. Save user clicks.

[LIVE-3788]: https://ledgerhq.atlassian.net/browse/LIVE-3788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.jsx
```diff
@@ -383,14 +382,24 @@ const SwapForm = () => {
...
       partner: provider,
     });
-    if (navigation) {
-      const { pathname, params } = navigation;
+
+    if (exchangeRate.providerType === "DEX") {
+      const from = swapTransaction.swap.from;
+      const fromAddress = from.parentAccount?.freshAddress || from.account.freshAddress;
+      const providerURL =
+        exchangeRate.providerURL || `/platform/${getProviderName(exchangeRate.provider)}`;
```
cjordan-ledger, line 390: If the provider is 1inch, `/platform/${getProviderName(exchangeRate.provider)}` will be `/platform/1inch`, whereas it looks like it used to be `/platform/1inch-lld` in `DEX_PROVIDERS`. 

Is this okay?

apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
```diff
@@ -20,17 +20,29 @@ type Props = {
...
-
+  const sufix = pathname.substr(
+    pathname.indexOf(match.params.appId) + match.params.appId.length + 1,
+  );
   const localManifest = useLocalLiveAppManifest(appId);
   const remoteManifest = useRemoteLiveAppManifest(appId);
+  const manifest = JSON.parse(JSON.stringify(localManifest || remoteManifest));
 
-  const manifest = localManifest || remoteManifest;
+  const dappUrl = manifest.params.dappUrl.split("?");
```
cjordan-ledger, line 44: I think the snapshot tests are failing on this line https://github.com/LedgerHQ/ledger-live/pull/2179#issuecomment-1363139928

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.jsx
sarneijim: Both are valid, I assume we are redirecting from "1inch" to "1inch-lld" in this case

cjordan-ledger:
> 

@ggilchrist-ledger - unsure if this is something you can help with/or are aware of, but these visual regression tests look like they may need to wait for the content to load before the snapshot is taken.

apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
```diff
@@ -20,17 +20,32 @@ type Props = {
...
+  );
 
   const localManifest = useLocalLiveAppManifest(appId);
   const remoteManifest = useRemoteLiveAppManifest(appId);
-
-  const manifest = localManifest || remoteManifest;
+  const manifest = Object.assign({}, localManifest || remoteManifest);
+  if (sufix !== "") {
+    const dappUrl = manifest?.params?.dappUrl?.split("?");
+    manifest.params.dappUrl =
```
cjordan-ledger, line 46: I think it would be worthwhile to write some unit tests for this as there's now quite a bit of complexity in this component. Maybe we can extract all of this to a hook called something like `useGetManifest(location, appId)` and test that?

apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
```diff
@@ -20,17 +20,32 @@ type Props = {
...
 };
 
-export default function PlatformApp({ match, appId: propsAppId }: Props) {
+export default function PlatformApp({ match, appId: propsAppId, location }: Props) {
   const history = useHistory();
-  const { state: urlParams, search } = useLocation();
+  const { params: urlParams, search, pathname } = location;
   const appId = propsAppId || match.params?.appId;
+  const sufix = pathname.substr(
+    pathname.indexOf(match.params.appId) + match.params.appId.length + 1,
```
cjordan-ledger, line 38: Am I right in thinking we are trying to capture the query params from the `pathname` and then to add them to `manifest.params.dappUrl`? If so, is that data not available in the `search` variable?

ggilchrist-ledger:
Does this affect all discover apps? Also, what was the issue with the visual regression tests? It doesn't look like there have been any changes to the screenshots so was it just flaky? @chrisd-ledger @sarneijim

ggilchrist-ledger:
FYI: issue with the account selected https://ledgerhq.atlassian.net/browse/LIVE-3788?focusedCommentId=268979

sarneijim:
> FYI: issue with the account selected https://ledgerhq.atlassian.net/browse/LIVE-3788?focusedCommentId=268979

Fixed! Good catch!

PR #2189, opened 2022-12-23
Feat: LIVE-4683 Add DEX limit UI

sarneijim:
### 📝 Description

As a user on LLD, i want to know the min/max amount I can swap with DEX providers

### ❓ Context

- **Impacted projects**: 
  - LLD
  - Common
- **Linked resource(s)**: 
  - https://ledgerhq.atlassian.net/browse/LIVE-4683

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

Screenshot 2022-12-23 at 10 25 21

### 🚀 Expectations to reach

Although DEX limits are less restrictives, We do not know exactly the "amount".
We will not show any error message. Only a empty list.

### DIFF
--- a/libs/ledger-live-common/src/exchange/swap/getExchangeRates.test.ts
+++ b/libs/ledger-live-common/src/exchange/swap/getExchangeRates.test.ts
@@ -230,4 +230,86 @@ describe("swap/getExchangeRates", () => {
       headers: expect.anything(),
     });
   });
+  test("should return correct error SwapExchangeRateAmountTooHigh", async () => {
+    const data = [
+      {
+        provider: "changelly",
+        providerType: "CEX",
+        from: "ethereum",
+        to: "ethereum/erc20/usd_tether__erc20_",
+        amountRequested: "1e-15",
+        minAmountFrom: "0.00000001",
+        maxAmountFrom: "0.00000008",
+        tradeMethod: "fixed",
+        status: "error",
+      },
+    ];
+    const resp = {
+      data,
+      status: 200,
+      statusText: "",
+      headers: {},
+      config: {},
+    };
+
+    mockedAxios.mockResolvedValue(Promise.resolve(resp));
+    mockedProviders.mockResolvedValue(Promise.resolve([]));
+    const res = await getExchangeRates(exchange, transaction);
+    expect(res[0]?.error?.name).toEqual("SwapExchangeRateAmountTooHigh");
+  });
+  test("should return correct error SwapExchangeRateAmountTooLow", async () => {
+    const data = [
+      {
+        provider: "changelly",
+        providerType: "CEX",
+        from: "ethereum",
+        to: "ethereum/erc20/usd_tether__erc20_",
+        amountRequested: "1e-15",
+        minAmountFrom: "0.08045622",
+        maxAmountFrom: "105.00000000",
+        tradeMethod: "fixed",
+        status: "error",
+      },
+    ];
+    const resp = {
+      data,
+      status: 200,
+      statusText: "",
+      headers: {},
+      config: {},
+    };
+
+    mockedAxios.mockResolvedValue(Promise.resolve(resp));
+    mockedProviders.mockResolvedValue(Promise.resolve([]));
+    const res = await getExchangeRates(exchange, transaction);
+    expect(res[0]?.error?.name).toEqual("SwapExchangeRateAmountTooLow");
+  });
+  test("should return correct error SwapExchangeRateAmountTooLowOrTooHigh", async () => {
+    const data = [
+      {
+        provider: "paraswap",
+        providerType: "DEX",
+        from: "ethereum",
+        to: "ethereum/erc20/usd_tether__erc20_",
+        tradeMethod: "float",
+        errorCode: 500,
+        errorMessage: "Failed to get rate for paraswap.",
+        status: "error",
+      },
+    ];
+    const resp = {
+      data,
+      status: 200,
+      statusText: "",
+      headers: {},
+      config: {},
+    };
+
+    mockedAxios.mockResolvedValue(Promise.resolve(resp));
+    mockedProviders.mockResolvedValue(Promise.resolve([]));
+    const res = await getExchangeRates(exchange, transaction);
+    expect(res[0]?.error?.name).toEqual(
+      "SwapExchangeRateAmountTooLowOrTooHigh"
+    );
+  });
 });
--- a/libs/ledger-live-common/src/exchange/swap/getExchangeRates.ts
+++ b/libs/ledger-live-common/src/exchange/swap/getExchangeRates.ts
@@ -10,6 +10,7 @@ import { getEnv } from "../../env";
 import {
   SwapExchangeRateAmountTooHigh,
   SwapExchangeRateAmountTooLow,
+  SwapExchangeRateAmountTooLowOrTooHigh,
 } from "../../errors";
 import network from "../../network";
 import type { Transaction } from "../../generated/types";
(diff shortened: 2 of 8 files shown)


cjordan-ledger (approved):

chabroA (changes requested):
The error ordering logic needs to be reworked and new tests cases might need to be added

libs/ledger-live-common/src/exchange/swap/hooks/useProviderRates.ts
```diff
@@ -120,37 +120,44 @@ export const useProviderRates = ({
...
-                  ? rateError
-                  : rate.error;
+              if (rateError?.name !== "SwapExchangeRateAmountTooLowOrTooHigh") {
+                rateError =
+                  (rateError as CustomMinOrMaxError).amount.comparedTo(
+                    (rate.error as CustomMinOrMaxError)?.amount
+                  ) === cmp
+                    ? rateError
+                    : rate.error;
+              }
```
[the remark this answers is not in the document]
sarneijim: Totally agree, wrong approach. It has been recode.

chabroA (commented):
Fixed a condition for the DEX quotes in the `inferError` function (cf. db65b46ffe5a8ec90eef2fc637ea856e9e64aeef).

Fixed the reducing logic in the `useProviderRates` hook (cf. f6fa33c4ebdb676f1c6bb132ffc2dbfa9a286fdc).
It's bad practice to alter (`splice`) the array you are currently reducing on, and the logic is overall less easy to understand.
One could argue it's not optimised performance-wise to keep iterating on the `rates` array once we have found a `SwapExchangeRateAmountTooLowOrTooHigh` error since it takes precedence over everything else. This is debatable and needs to be balanced with easier understanding and maintenance of the code / logic. Also, the performance optimisation would be insignificant since the algorithm complexity is _O(n)_ in both cases.

libs/ledger-live-common/src/exchange/swap/getExchangeRates.ts
```diff
@@ -146,11 +147,29 @@ const inferError = (
...
+  const {
+    amountTo,
+    minAmountFrom,
+    maxAmountFrom,
+    errorCode,
+    errorMessage,
+    status,
+  } = responseData;
+
+  // DEX quotes are out of limit scope error
```
chabroA, line 163: > DEX quotes are out of limit scope error

This does not make sense to me.
What do you mean by _out of limit scope_?
IMO it's either _out of scope_ or _out of limit_ (and in both case, what is the _limit_ or _scope_?)

chabroA (approved):

libs/ledger-live-common/src/exchange/swap/getExchangeRates.ts
sarneijim: DEX quotes are out of limits error. We do not know if it is a low or high limit, neither the amount.

chabroA, line 163:
```suggestion
  // DEX quotes are out of limits error. We do not know if it is a low or high limit, neither the amount.
```

libs/ledger-live-common/src/exchange/swap/getExchangeRates.test.ts
```diff
@@ -230,4 +230,86 @@ describe("swap/getExchangeRates", () => {
...
+      headers: {},
+      config: {},
+    };
+
+    mockedAxios.mockResolvedValue(Promise.resolve(resp));
+    mockedProviders.mockResolvedValue(Promise.resolve([]));
+    const res = await getExchangeRates(exchange, transaction);
+    expect(res[0]?.error?.name).toEqual("SwapExchangeRateAmountTooLow");
+  });
+  test("should return correct error SwapExchangeRateAmountTooLowOrTooHigh", async () => {
```
ggilchrist-ledger, line 287: Is it worth adding a test for when the min and max rates are `""` in the response, like we get with Paraswap? @sarneijim

libs/ledger-live-common/src/exchange/swap/getExchangeRates.test.ts
```diff
@@ -230,4 +230,86 @@ describe("swap/getExchangeRates", () => {
...
+      headers: {},
+      config: {},
+    };
+
+    mockedAxios.mockResolvedValue(Promise.resolve(resp));
+    mockedProviders.mockResolvedValue(Promise.resolve([]));
+    const res = await getExchangeRates(exchange, transaction);
+    expect(res[0]?.error?.name).toEqual("SwapExchangeRateAmountTooHigh");
+  });
+  test("should return correct error SwapExchangeRateAmountTooLow", async () => {
```
ggilchrist-ledger, line 260: Please add a blank line between the tests :)

libs/ledger-live-common/src/exchange/swap/hooks/useProviderRates.test.ts
```diff
@@ -330,6 +333,34 @@ describe("useProviderRates", () => {
...
 
+  it("SwapExchangeRateAmountTooLowOrTooHigh should take precedence over SwapExchangeRateAmountTooLow", async () => {
+    const swapExchangeRateAmountTooLowError = new SwapExchangeRateAmountTooLow(
+      undefined,
+      {
+        amount: new BigNumber(1),
+      }
+    );
+
+    const swapExchangeRateAmountTooLowOrTwoHighError =
```
ggilchrist-ledger, line 344: Small thing: should be `...TooHighError`

libs/ledger-live-common/src/exchange/swap/hooks/useProviderRates.test.ts
```diff
@@ -330,6 +333,34 @@ describe("useProviderRates", () => {
     expect(result.current.rates.value).toBeUndefined();
   });
 
+  it("SwapExchangeRateAmountTooLowOrTooHigh should take precedence over SwapExchangeRateAmountTooLow", async () => {
```
ggilchrist-ledger, line 336: Where does the `AmountTooHigh` error fit into this? Should it be incorporated into this test or maybe have a separate one?

libs/ledger-live-common/src/exchange/swap/getExchangeRates.ts
```diff
@@ -146,11 +147,29 @@ const inferError = (
...
+    errorCode,
+    errorMessage,
+    status,
+  } = responseData;
+
+  // DEX quotes are out of limits error. We do not know if it is a low or high limit, neither the amount.
+  if (
+    (!minAmountFrom || !maxAmountFrom) &&
+    status === "error" &&
+    errorCode !== 300
```
ggilchrist-ledger, line 167: @jnicoulaud-ledger What does the 300 error code mean and why do we use it? I see it's multiple choices but not sure why we use that. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/300#:~:text=The%20HTTP%20300%20Multiple%20Choices,more%20than%20one%20possible%20responses.

libs/ledger-live-common/src/exchange/swap/getExchangeRates.test.ts
sarneijim: we are checking that we are not retrieving a wrong value (cex value)

libs/ledger-live-common/src/exchange/swap/hooks/useProviderRates.test.ts
sarneijim: It can be too small too. (I have tested it with 0.000...01 and works in the same way.

libs/ledger-live-common/src/exchange/swap/getExchangeRates.ts
sarneijim: We are receiving different kind of error from Paraswap and 1Inch, perhaps if we unify them we can make this part of the code cleaner.

jnicoulaud-ledger: @ggilchrist-ledger it's not an HTTP status code, it's a "swap API error code" (and yes it sucks that we use error codes that look like HTTP status codes, we have a plan to change it): https://github.com/LedgerHQ/swap/blob/master/modules/model/src/main/scala/co/ledger/swap/model/ApplicationError.scala#L22

In the meanwhile there is a mapping here to avoid harcoding these codes: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/exchange/swap/index.ts#L198

PR #2205, opened 2022-12-27
feat/LIVE-2613 Improve error handling and test coverage of HSM connections

juan-cortes:
### 📝 Description

Introduce test coverage for our script runner connections, after discussing with the relevant parties, the original scope of this feature which was to map all errors whenever the device returned anything other than `9000` changed. The reason for this conditional map+throw you see in this PR is because we want to remain open to the possibility of a script runner not closing the connection immediately after something other than a status OK response.

In doing this, I tweaked a little bit the variable names too to make them clearer to understand and introduced a test suite which, as ugly as it is, should allow us to detect regressions in the emitted events and thrown errors of this very delicate piece of our code.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-2613` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
Note that some tests are slow due to the need to introduce an artificial delay in order to allow the socket.ts logic to emit the allow manager (renamed to allow secure channel) event to be emitted.
image

### 🚀 Expectations to reach
Accessing the manager, installing apps, onboarding a device, they should all remain unaltered.
Ideally this PR introduces no notable changes from a QA stand point.


libs/ledger-live-common/src/api/socket.ts
```diff
@@ -33,16 +33,11 @@ export const createDeviceSocket = (
...
     let unsubscribed = false; // subscriber wants to stops everything
-
     let correctlyFinished = false; // the socket logic reach a normal termination
-
-    let inBulk = false; // bulk is a mode where we have many apdu to run on device and no longer need the connection
-
-    let timeoutForAllowManager: NodeJS.Timeout | null | number = null; // track if there is an ongoing allow manager step
-
+    let inBulkMode = false; // we have an array of apdus to exchange, without the need of more WS messages.
+    let allowSecureChannelTimeout: NodeJS.Timeout | null | number = null; // track if there is an ongoing allow manager step
```
alexandremgo, line 40: i'm picky 😈 : `track if there is an ongoing secure channel communication step` or equivalent for the comment ;)

libs/ledger-live-common/src/api/socket.ts
```diff
@@ -91,45 +89,61 @@ export const createDeviceSocket = (
...
             }
 
             const r = await transport.exchange(apdu);
 
-            if (timeoutForAllowManager) {
-              // if the exchange is faster than the timeout, we pass through because it's not "awaiting" user action
-              clearTimeout(timeoutForAllowManager as number);
-              timeoutForAllowManager = null;
+            if (allowSecureChannelTimeout) {
+              clearTimeout(allowSecureChannelTimeout as number);
```
alexandremgo, line 112: You can define: `let allowSecureChannelTimeout: NodeJS.Timeout | null = null;` withouth `number` and then only have:

```TS
clearTimeout(allowSecureChannelTimeout);
```

IDK why TS on VSCode wanted `clearTimeout` arg as an input, it was the same on my side. But like above TS should work.

libs/ledger-live-common/src/api/socket.ts
```diff
@@ -91,45 +89,61 @@ export const createDeviceSocket = (
...
+            switch (status) {
+              case StatusCodes.OK:
+                response = "success";
+                break;
 
-            // if allow manager was requested, we either throw if deny or emit accepted
-            if (allowManagerAwaitingUser) {
-              if (status === 0x6985 || status === 0x5501) {
-                o.error(new UserRefusedAllowManager());
+              case StatusCodes.LOCKED_DEVICE:
```
alexandremgo, line 125: I know the comment above was not amazing, because we don't just throw `LockedDeviceError`, but let's keep it:

```
// TransportStatusError will be remapped to a LockedDeviceError with a correct error message
```

libs/ledger-live-common/src/__tests__/api/socket.ts
```diff
@@ -0,0 +1,98 @@
+import {
+  openTransportReplayer,
+  RecordStore,
+} from "@ledgerhq/hw-transport-mocker";
+import { toArray } from "rxjs/operators";
+import scenarios from "./test-helpers/scenarios";
+import WS from "isomorphic-ws";
+
+// import { setEnv } from "@ledgerhq/live-common/env";
```
alexandremgo, line 9: unused import ;)

libs/ledger-live-common/src/__tests__/api/socket.ts
```diff
@@ -0,0 +1,98 @@
+import {
+  openTransportReplayer,
+  RecordStore,
+} from "@ledgerhq/hw-transport-mocker";
+import { toArray } from "rxjs/operators";
+import scenarios from "./test-helpers/scenarios";
```
alexandremgo, line 6: 😈 import after the lib import bellow

libs/ledger-live-common/src/__tests__/api/test-helpers/scenarios.ts
```diff
@@ -0,0 +1,219 @@
+type Event = [string] | [string, string] | [string, string, boolean]; // callback, payload, eager
```
alexandremgo, line 1: FYI you can create "named tuples":

```TS
type Event =
  | [callback: string]
  | [callback: string, payload: string]
  | [callback: string, payload: string, eager: boolean];
```

It should have the same behavior than normal tuples

libs/ledger-live-common/src/__tests__/api/socket.ts
```diff
@@ -0,0 +1,98 @@
...
+          if (payload?.includes("e051000000")) {
+            // Need to delay the next apdu to see the allow secure channel event.
+            transport.setArtificialExchangeDelay(1000);
+          }
+          const maybeCallback = callbacks[callback];
+          if (maybeCallback) {
+            maybeCallback({ data: payload });
+          }
+
+          // Check if there's a next event, and if it's a onclose emit it.
```
alexandremgo, line 54: if it's an `onclose` event or if it's an "eager" event ?

alexandremgo: Not sure i understood the "eager" concept

alexandremgo (commented):
Really nice to have the cached device error now 🔥

Just some comments ;)

libs/ledger-live-common/src/api/socket.ts
juan-cortes: it's for the user confirmation but indeed the comment should've been updated with the rename

libs/ledger-live-common/src/api/socket.ts
juan-cortes: uhmm I didn't write this, but sure, your suggestion works

libs/ledger-live-common/src/api/socket.ts
juan-cortes: Sorry but the case literally tells you that without the comment, anyone reading that code should know or be able to look into what `TransportStatusError` is doing without the need for a comment.

image

libs/ledger-live-common/src/__tests__/api/socket.ts
juan-cortes: As expected

libs/ledger-live-common/src/__tests__/api/socket.ts
juan-cortes: Shouldn't this be caught by ... something? 😪

libs/ledger-live-common/src/__tests__/api/test-helpers/scenarios.ts
juan-cortes: Oh nice!

libs/ledger-live-common/src/__tests__/api/socket.ts
juan-cortes: The comment is basically shit, I needed to automatically emit the next event when needed, that's what the concept of eagerness meant. But the comment implies that this only applies to onclose, when it doesn't. I'll change the comment

libs/ledger-live-common/src/api/socket.ts
alexandremgo: true !

libs/ledger-live-common/src/__tests__/api/socket.ts
alexandremgo: 😢

alexandremgo (approved):

PR #2214, opened 2022-12-28
FAT-717 Use the new device selection screen everywhere

grsoares21:
### 📝 Description
This PR integrates the new device selection screen in all the places in the app where we're prompted to select a device. They are all shown based on the same feature flag.
Unfortunately it wasn't possible to integrate it inside the original SelectDevice component (which would have made the PR much simpler) because the API and the design interactions of the new component are different to the original one.

Test plan
The places where the device selection screens were migrated are:
- AddAccount flow
- Custom image transfer flow (stax)
- Debug Fetch custom Image
- Debug Multi app install
- Debug devices
- Migrate accounts
- Onboarding (connect new device)
- Swap
- Verify acccount (coming from the platform / discover tab)
- Exchange (coming from the platform / discover tab)

### ❓ Context

- **Linked resource(s)**: `FAT-717`

### ✅ Checklist

- [N/A] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes**

### 📸 Demo
https://user-images.githubusercontent.com/6013294/209844698-0fcf9b78-6d85-491d-ad08-774efc059ec3.mp4


apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
```diff
@@ -152,28 +163,14 @@ export default function SelectDevice({ onSelect, stopBleScanning }: Props) {
...
-    });
-  }, [navigation]);
+  const openBlePairingFlow = useCallback(() => {
+    setIsPairingDevices(true);
+  }, []);
+
+  const closeBlePairingFlow = useCallback(() => {
+    setIsPairingDevices(false);
+    setIsAddNewDrawerOpen(false);
+  }, []);
```
juan-cortes, line 173: You **[don't need](https://reactjs.org/docs/hooks-reference.html#usestate)** `useCallback` since you're not changing the behavior based on any internal variable
>React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.

juan-cortes (approved):
lgtm

apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
grsoares21: this is more to avoid rerenders in case we memo the underlying components (if we pass () => setIsPairingDevices(true) directly in the prop, it will always rerender when the parent rerenders, even if the component is memoed)

hedi-edelbloute (approved):

PR #1991, opened 2022-11-28
More Granular CI Checks

valpinkman:
### 📝 Description

Change the way we handle checks on PR to have only relevant ones

### ❓ Context

- **Impacted projects**: `automation` 
- **Linked resource(s)**: https://github.com/LedgerHQ/ledger-live-issues/issues/65 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] **No breaking changes** we will change the way PR checks are handled but all test will be done on this PR

### 📸 Demo

soonTM

### 🚀 Expectations to reach

Have clear checks on PR CI

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #2223, opened 2022-12-29
Support/healthchecks runners

cdiarra-ledger:
### 📝 Description

Add a github action job that will verify that software dependencies are installed and can be executed successfully. Whenever an issue occurs, a slack message will be sent.
That will help isolate CI issues that come from runners configuration.

### ❓ Context

- **Impacted projects**: `` No project will be impacted as it is only a github workflow that will not write any ledger live asset

### ✅ Checklist

- [x] **No breaking changes** - no codebase impact

### 📸 Demo

Preview of the markdown table, built at the end of the workflow :
Screenshot from 2022-12-29 14-39-51

### 🚀 Expectations to reach

### DIFF
--- a/.github/workflows/healthchecks-runners.yaml
+++ b/.github/workflows/healthchecks-runners.yaml
@@ -1,0 +1,575 @@
+name: "[Healthchecks] CI Runners"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 */6 * * *"
+
+jobs:
+  start-runner:
+    name: Start self-hosted EC2 runner
+    runs-on: [ledger-live, linux]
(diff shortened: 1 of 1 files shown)


valpinkman (approved):
i would say we could add python checking for macOS as well, but overall it's pretty solid

valpinkman (changes requested):
Sorry one last comment => it would be great to have a link toward the run in the slack message (cf example here: https://github.com/LedgerHQ/ledger-live/pull/1991/files#diff-83b28405abf3c3a00acab6d012253c9486b2765a29ca6556ab263c3c1adff6b4R174)

After that we are golden

PR #2239, opened 2023-01-03
bugfix - LLD - shows close button on live app screen

cjordan-ledger:
### 📝 Description

Fixes the close button on the live app screen by ensuring the `onClose` prop is forwarded to the WebView component.

### ❓ Context

- **Impacted projects**:  LLD
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-5062

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/119950081/210390999-24c9fa28-d5c6-4bdc-95ad-93ad528d300f.mov

### 🚀 Expectations to reach


sarneijim (approved):
LGTM (tested correct behaviour with buy-sell and card).
I have added Justkant, who I think it is the PlatformApp expert, in the review.

Justkant (changes requested):
Nice catch, I also see that we forgot to add back the `config` props on desktop, can you add that back also ?

sarneijim (commented):
Not sure if we need "onClose" if "config" has "shouldDisplayClose".

cjordan-ledger:
> Not sure if we need "onClose" if "config" has "shouldDisplayClose".

@sarneijim I think we need both. `PlatformApp` only passes in `onClose` and no config.

Justkant:
> Not sure if we need "onClose" if "config" has "shouldDisplayClose".

No need to change anything, `shouldDisplayClose` is checked in the `TopBar`

.changeset/silly-carrots-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"ledger-live-desktop": minor
```
Justkant, line 2: Should just be a patch here

Justkant (changes requested):
Just need to update the changeset to a patch and everything should be good

Justkant (approved):

PR #2248, opened 2023-01-04
fix(wallet-api): filter currencies families

Justkant:
### 📝 Description

We had an issue with the current implementation where we would send unsupported currencies to the wallet-api.
We solve the issue by checking against the `FAMILIES` array exported by the `wallet-api-core`

### ❓ Context

- **Impacted projects**: `ledger-live` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** Tested manually with the debugger (another PR with playwright testing incoming) 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Manifest to test the wallet-api
```json
{
  "id": "debugger",
  "name": "Debugger",
  "url": "https://platform-app-debug-git-wallet-api-conversion-ledgerhq.vercel.app/",
  "homepageUrl": "https://platform-app-debug-git-wallet-api-conversion-ledgerhq.vercel.app/",
  "icon": "",
  "platform": "all",
  "apiVersion": "^2.0.0 || ~0.0.1",
  "manifestVersion": "1",
  "branch": "experimental",
  "categories": [],
  "currencies": ["**"],
  "content": {
    "shortDescription": {
      "en": ""
    },
    "description": {
      "en": ""
    }
  },
  "permissions": [
    "account.list",
    "account.receive",
    "account.request",
    "currency.list",
    "device.close",
    "device.exchange",
    "device.transport",
    "message.sign",
    "transaction.sign",
    "transaction.signAndBroadcast",
    "wallet.capabilities"
  ],
  "domains": ["http://*", "https://*"]
}
```

### 🚀 Expectations to reach


libs/ledger-live-common/src/wallet-api/types.ts
```diff
@@ -14,6 +14,7 @@ export type {
 } from "@ledgerhq/wallet-api-core";
 
 export {
+  FAMILIES as WalletAPI_FAMILIES,
```
chabroA, line 17: `FAMILIES` is not a type. IMO it should not be put in the `types.ts` file to avoid confusion. It could be moved to the `constants.ts` file, since it's more akin to a constant (it's fixed for a given version of the wallet-api).
Otherwise it could also be moved to a new `common.ts` file if it's not considered a constant.

Justkant: Good catch, I'll move that

chabroA (approved):

PR #2252, opened 2023-01-04
feat: create client specifically to get balance

Sheng-Long:
### 📝 Description

Separate Hedera client created for balance specific queries

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/hedera/api/network.ts
+++ b/libs/ledger-live-common/src/families/hedera/api/network.ts
@@ -43,7 +43,7 @@ export async function getAccountBalance(
   try {
     accountBalance = await new hedera.AccountBalanceQuery({
       accountId,
-    }).execute(getClient());
+    }).execute(getBalanceClient());
   } catch {
     throw new HederaAddAccountError();
   }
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/src/families/hedera/api/network.ts
```diff
@@ -55,10 +55,18 @@ export async function getAccountBalance(
...
 function getClient(): hedera.Client {
   _hederaClient ??= hedera.Client.forMainnet().setMaxNodesPerTransaction(1);
 
   //_hederaClient.setNetwork({ mainnet: "https://hedera.coin.ledger.com" });
 
   return _hederaClient;
 }
+
+function getBalanceClient(): hedera.Client {
+  _hederaBalanceClient ??= hedera.Client.forMainnet();
```
henrily-ledger, line 69: First time i see this operation, can you enlighten me about what is it ?
Thank you :)

henrily-ledger (approved):

libs/ledger-live-common/src/families/hedera/api/network.ts
Sheng-Long: If you mean the `??=`, to be honest I had just copy/pasted the existing Client declaration and reused it.

From some googling it seems to be the nullish coalescing assignment: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

I'm not entirely sure the reason it was used here, but it seems that it would prevent reassignment of the client which could be helpful?

haammar-ledger (commented):
Doesn't look harmful, but can you please describe the reason for using separate clients?

Sheng-Long:
> Doesn't look harmful, but can you please describe the reason for using separate clients?

The original issue was this: 
image

Because transactions can only use 1 node, this error can occur when the node selected is down.  Balance queries, however, do not require signing and thus, can use more than one node.  Thus, using `hedera.Client.forMainnet();` for the balance client will allow for more uptime on balance queries when the random node selected is down.

PR #2255, opened 2023-01-05
feat: swap analytics for LLM

cjordan-ledger:
### 📝 Description

LLM Swap Analytics

### ❓ Context

- **Impacted projects**: Mobile
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-4833

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/FabActions/hooks/useAssetActions.tsx
+++ b/apps/ledger-live-mobile/src/components/FabActions/hooks/useAssetActions.tsx
@@ -13,7 +13,7 @@ import {
 } from "../../../reducers/settings";
 import { ActionButtonEvent } from "..";
 import ZeroBalanceDisabledModalContent from "../modals/ZeroBalanceDisabledModalContent";
-import { SWAP_VERSION } from "../../../screens/Swap/utils";
+import { sharedSwapTracking } from "../../../screens/Swap/utils";
 
 type useAssetActionsProps = {
   currency?: CryptoCurrency | TokenCurrency;
@@ -128,9 +128,8 @@ export default function useAssetActions({
                     Icon: iconSwap,
                     event: "button_clicked",
                     eventProperties: {
+                      ...sharedSwapTracking,
                       button: "swap",
-                      flow: "swap",
-                      swapVersion: SWAP_VERSION,
                     },
                     navigationParams: [
                       NavigatorName.Swap,
--- a/apps/ledger-live-mobile/src/components/FabActions/hooks/useAssetActions.tsx
+++ b/apps/ledger-live-mobile/src/components/FabActions/hooks/useAssetActions.tsx
@@ -13,6 +13,7 @@ import {
 } from "../../../reducers/settings";
 import { ActionButtonEvent } from "..";
 import ZeroBalanceDisabledModalContent from "../modals/ZeroBalanceDisabledModalContent";
+import { SWAP_VERSION } from "../../../screens/Swap/utils";
 
 type useAssetActionsProps = {
   currency?: CryptoCurrency | TokenCurrency;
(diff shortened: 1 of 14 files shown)


apps/ledger-live-mobile/src/components/FabActions/index.tsx
```diff
@@ -84,6 +84,7 @@ export const FabButtonBarProvider = ({
   eventProperties?: { [key: string]: unknown };
   children: (value: { quickActions: ActionButtonProps[] }) => ReactNode;
 }) => {
+  const { track } = useAnalytics();
```
sarneijim, line 87: Whiat is different between?
```
``` 
and 
```
const { track } = useAnalytics()
```

apps/ledger-live-mobile/src/components/FabActions/hooks/useAssetActions.tsx
```diff
@@ -123,9 +124,14 @@ export default function useAssetActions({
             ...(availableOnSwap
               ? [
                   {
-                    id: "swap",
```
sarneijim, line 126: Why are we removing this it? Can we use the same structure for any item in the array?

apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
```diff
@@ -73,6 +76,12 @@ export default function SelectFeesStrategy({
 
   const onPressStrategySelect = useCallback(
     (item: FeeStrategy) => {
+      track("button_clicked", {
+        button: item.label,
+        flow: "swap",
+        page: "Swap quotes",
+        swapVersion: SWAP_VERSION,
```
sarneijim, line 83: Can we create a default swap property object an export it across any track? (If tomorrow, we want to change the property the change will be done only in one place)
swapVersion, flow

apps/ledger-live-mobile/src/screens/Swap/Form/Max.tsx
```diff
@@ -2,9 +2,20 @@ import React from "react";
...
 
 export function Max({ swapTx }: { swapTx: SwapTransactionType }) {
   const { t } = useTranslation();
+  const { track } = useAnalytics();
+
+  const onToggle = (event: boolean) => {
+    track("button_clicked", {
+      button: "max",
+      flow: "swap",
+      state: event,
```
sarneijim, line 15: swap version?

apps/ledger-live-mobile/src/components/FabActions/index.tsx
cjordan-ledger: Good question. I used `useAnalytics` as it calls `track` and adds in the current `page` route automatically. However, when I've been testing this, `page` is sometimes wrongly assigned to `portfolio_navigator`.  I also have just noticed that `track`  also adds in a `screen` value which does look correct.

 It's a little bit confusing. I think we should send either `page` or `screen` and ensure that this is set to the right value, but don't really want to fix this issue in this PR as it'll affect all analytics.

apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
cjordan-ledger: Can do, I thought about this but wasn't sure if it was worthwhile as there will only be two fields in that shared object.

apps/ledger-live-mobile/src/components/FabActions/hooks/useAssetActions.tsx
cjordan-ledger: `apps/ledger-live-mobile/src/components/FabActions/index.tsx:127`

We want to add additional event properties like `swapVersion` etc, so we need to replace `id` with `event` in order to do that.

apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
cjordan-ledger: https://github.com/LedgerHQ/ledger-live/pull/2255/commits/6bf6cb751e5882133d7cc8a09edf24853bbf8fc9

sarneijim (approved):

PR #1901, opened 2022-11-21
feat(wallet-api): device.transport LLD & LLM integration [LIVE-4293]

Justkant:
### 📝 Description

Expose a `hw-transport` using 3 RPC calls `device.getTransport`, `device.exchange` and `device.close` on the client & server.
Create a specific DeviceAction using a subject to maintain a single device transport open in the wallet.

### ❓ Context

- **Impacted projects**: `ledger-live-desktop`, `ledger-live-mobile`, `ledger-live-common` 
- **Linked resource(s)**: [`LIVE-4293`](https://ledgerhq.atlassian.net/browse/LIVE-4293) 

### ✅ Checklist

- [ ] **Test coverage** Manual testing using the dummy-wallet-app 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/3009753/208748293-43f6f8c1-54e5-4abd-8ec7-b0ea140bdfea.mov

#### Manifest example
```json
{
  "id": "dummy",
  "name": "Dummy Wallet App",
  "url": "http://localhost:3000",
  "homepageUrl": "http://localhost:3000",
  "icon": "",
  "platform": "all",
  "apiVersion": "^2.0.0 || ~0.0.1",
  "manifestVersion": "1",
  "branch": "experimental",
  "categories": [],
  "currencies": ["*"],
  "content": {
    "shortDescription": {
      "en": ""
    },
    "description": {
      "en": ""
    }
  },
  "permissions": [
    "account.list",
    "account.receive",
    "account.request",
    "currency.list",
    "device.close",
    "device.exchange",
    "device.transport",
    "message.sign",
    "transaction.sign",
    "transaction.signAndBroadcast",
    "wallet.capabilities"
  ],
  "domains": ["http://*", "https://*"]
}
```

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #1900, opened 2022-11-21
feat(wallet-api): LLM & LLD server implementation [LIVE-4394]

Justkant:
### 📝 Description

We keep the old behaviour and codebase for v1 (live-app & live-app-sdk) and only use the Wallet API for v2 compatible manifests.

### ❓ Context

- **Impacted projects**: `ledger-live-desktop`, `ledger-live-mobile`, `ledger-live-common` 
- **Linked resource(s)**: [`LIVE-4394`](https://ledgerhq.atlassian.net/browse/LIVE-4394) 

### ✅ Checklist

- [ ] **Test coverage** Manual testing with the new debugger from https://github.com/LedgerHQ/platform-app-debug/pull/4 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

#### Manifest example for the new wallet-api debugger
```json
{
  "id": "debugger",
  "name": "Debugger",
  "url": "https://platform-app-debug-git-wallet-api-conversion-ledgerhq.vercel.app/",
  "homepageUrl": "https://platform-app-debug-git-wallet-api-conversion-ledgerhq.vercel.app/",
  "icon": "",
  "platform": "all",
  "apiVersion": "^2.0.0 || ~0.0.1",
  "manifestVersion": "1",
  "branch": "experimental",
  "categories": [],
  "currencies": ["*"],
  "content": {
    "shortDescription": {
      "en": ""
    },
    "description": {
      "en": ""
    }
  },
  "permissions": [
    "account.list",
    "account.receive",
    "account.request",
    "currency.list",
    "device.close",
    "device.exchange",
    "device.transport",
    "message.sign",
    "transaction.sign",
    "transaction.signAndBroadcast",
    "wallet.capabilities"
  ],
  "domains": ["http://", "https://"]
}
```

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #2262, opened 2023-01-06
Add walletApiAdapter files to migrate to wallet-api

chabroA:
### 📝 Description

With the migration from live-app-sdk to wallet-api, we also need to migrate transaction adapters logic to the new wallet-api types.

cf:
- https://github.com/LedgerHQ/ledger-live/pull/1900
- https://github.com/LedgerHQ/ledger-live/pull/1901

### ❓ Context

- **Impacted projects**: `LLM`, `LLD`, `LLC`
- **Linked resource(s)**: [LIVE-5085](https://ledgerhq.atlassian.net/browse/LIVE-5085)

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-5085]: https://ledgerhq.atlassian.net/browse/LIVE-5085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/families/ethereum/walletApiAdapter.ts
+++ b/libs/ledger-live-common/src/families/ethereum/walletApiAdapter.ts
@@ -1,15 +1,13 @@
-// TODO: to remove once live-app-sdk is depreciated and removed from LL
-
-import { EthereumTransaction as PlatformTransaction } from "@ledgerhq/live-app-sdk";
+import { EthereumTransaction as WalletAPITransaction } from "@ledgerhq/wallet-api-core";
 import { Transaction } from "./types";
 
 const CAN_EDIT_FEES = true;
 
-const areFeesProvided = (tx: PlatformTransaction): boolean =>
+const areFeesProvided = (tx: WalletAPITransaction): boolean =>
   !!(tx.gasLimit || tx.gasPrice);
 
 const convertToLiveTransaction = (
-  tx: PlatformTransaction
+  tx: WalletAPITransaction
 ): Partial<Transaction> => {
   const hasFeesProvided = areFeesProvided(tx);
 
@@ -26,8 +24,8 @@ const convertToLiveTransaction = (
   return hasFeesProvided ? { ...liveTx, feesStrategy: "custom" } : liveTx;
 };
 
-const getPlatformTransactionSignFlowInfos = (
-  tx: PlatformTransaction
+const getWalletAPITransactionSignFlowInfos = (
+  tx: WalletAPITransaction
 ): {
   canEditFees: boolean;
   hasFeesProvided: boolean;
@@ -40,4 +38,4 @@ const getPlatformTransactionSignFlowInfos = (
   };
 };
 
-export default { getPlatformTransactionSignFlowInfos };
+export default { getWalletAPITransactionSignFlowInfos };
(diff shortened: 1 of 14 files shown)


libs/ledger-live-common/src/families/polkadot/walletApiAdapter.ts
```diff
@@ -0,0 +1,27 @@
...
+const CAN_EDIT_FEES = false;
+
+const convertToLiveTransaction = (
+  tx: WalletAPITransaction
+): Partial<Transaction> => ({
+  ...tx,
+  era: tx.era ? `${tx.era}` : undefined,
+});
+
+const getWalletAPITransactionSignFlowInfos = (
```
IAmMorrow, line 13: Some functions seems to always have the same prototype, would it be a good idea to define a type and import it in all those adapter modules ?

chabroA: Indeed the `getWalletAPITransactionSignFlowInfos` function has the same signature for all coins needing an adapter as of today. The type can indeed be extracted 👍

chabroA: cf. 6d3fea8d1dce8e07eff8b571cc9b3a017a4090b5 ✅

libs/ledger-live-common/src/families/ethereum/walletApiAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+    amount: tx.amount,
+    recipient: tx.recipient,
+    gasPrice: tx.gasPrice,
+    userGasLimit: gasLimit,
+  };
+
+  return hasFeesProvided ? { ...liveTx, feesStrategy: "custom" } : liveTx;
+};
+
+const getWalletAPITransactionSignFlowInfos: GetWalletAPITransactionSignFlowInfos<
```
Justkant, line 33: No tests for this one or just an oversight ?

libs/ledger-live-common/src/wallet-api/types.ts
```diff
@@ -62,3 +62,20 @@ export type AppManifest = {
...
+  LiveTransaction
+> = (tx: WalletAPITransaction) => {
+  canEditFees: boolean;
+  hasFeesProvided: boolean;
+  liveTx: Partial<LiveTransaction>;
+};
+
+export type AreFeesProvided<WalletAPITransaction> = (
+  tx: WalletAPITransaction
+) => boolean;
```
Justkant, line 77: I find it a bit misleading to use the same name for a generic and a real type, `WalletAPITransaction` is exported just above and in this lines `WalletAPITransaction` is a generic type and not the exported one.

Maybe we could simplify it, I have 2 ideas in mind:
```suggestion
export type AreFeesProvided<Transaction> = (
  tx: Transaction
) => boolean;
```
or
```suggestion
export type AreFeesProvided<T> = (
  tx: T
) => boolean;
```

And if you decide to change that we should do the same for `ConvertToLiveTransaction` and `GetWalletAPITransactionSignFlowInfos`

Justkant (changes requested):
Just a small thing with the types and maybe a missing test

libs/ledger-live-common/src/wallet-api/types.ts
chabroA: Fixes done here -> 0ec3e458ab6b359714579a87ef4e471258e52b45

Rename + use of `extends` 👌

libs/ledger-live-common/src/families/ethereum/walletApiAdapter.ts
chabroA: Indeed, fixed here -> f8351c25fdf46ee25bdeabac47f9a3db85d46774 ✅

Justkant (approved):

PR #2263, opened 2023-01-06
feat/FAT-705 BLE Permission rework

juan-cortes:
#### Original description from Juan

The permission system in LLM is pretty unstable right now and relies on an anti-pattern of the component being mounted or not. I reworked both the location and bluetooth permission components' logic into hooks, fixed the navigation issues that were happening (ie, no way of going back from the UI, only swipe and native Back button) and tested on the devices I had at hand, namely iOS and Android 12.

Like I said on the video, I refactored quite a bit of code into hooks and since it was mostly done by @ofreyssinet-ledger and @alexandremgo I'm requesting their take on the matter since I'm off starting today. The only thing I'm not happy with is the timeout for the hooks to consider the request as failed but unless you find a cleaner approach, this works pretty well.

I've added an entry to the debugging screens that simply wraps the requiresBLE component for ease of access but it's already pretty easy to access it via manager or onboarding. In particular, pay attention to the possibility of double back arrow renderings in case the errors are not rendered as a whole screen. In which case, we would have to pass the appropriate prop to the device wrapper component thingie you did @alexandremgo.

#### Update from Alex

Note: to try in a dev build, please make sure you re-built the app entirely (and not just a hot reload with metro)

What has been added:
- new UI components for service disabled and service permission errors

- BLE permission on Android: allow the user to request again to allow permissions
  - taking into account the "never ask again" possibility

- BLE service enable: for both Android and iOS
  - Hook: exposing a method which checks the bluetooth service state and requests (if necessary) the user to enable it.
  - Also hook listening to an update of the bluetooth service state from BleTransport, and updates the state accordingly.
  - This hook can also indicate if bluetooth has enough permissions to be enabled, which is useful for iOS as we don't have an easy way to implement a "useIosBluetoothPermissions" currently.

- Location permissions on Android: allow user to request again
  - taking into account the "never ask again" possibility.

- a new location helper native module for android that:
  - Allows to check the location services state and request (if necessary) the user to enable them.
  - Also listens to the state (enabled/disabled) of the location services, and notify the consumer of any changes
  - Usage in a new hook `useAndroidEnableLocation`

- Location services enable: using the new native module
  - Component that renders its children if the locations services are enabled
  - Otherwise, renders a new location disabled component
  - Possibility for the user to be directly prompted to enable their services again

- Restructured/renamed all the existing components/hooks related to bluetooth and location services
  - `RequiresBLE` should now handle every possible cases (turning on/off services, removing/adding permissions)
 
- Different updates on `GenericError`, `RenderError`, `BleDevicesScanning` etc. to better handle BLE and location services related errors

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/FAT-705` 

### ✅ Checklist

- [ ] **Test coverage**: need to be QAed
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

<details>
  <summary>Android ex</summary>

https://user-images.githubusercontent.com/15096913/213485702-e4c9bf15-1084-44b8-9e7e-da888f7ac2db.mp4

</details>

<details>
  <summary>iOS ex</summary>

https://user-images.githubusercontent.com/15096913/213485861-8cf2676d-6acd-410a-a529-2cea9f494c7a.mp4

</details>
### 🚀 Expectations to reach
It should solve all the problems from the ticket and never leave the app in a state with broken ctas, no UI, or otherwise broken.


apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBLEPermissions.ts
```diff
@@ -0,0 +1,148 @@
...
+    setRetryNonce(new Date().getTime());
+  }, []);
+
+  useEffect(() => {
+    // set hasPermission to false after 400ms if it's still undefined.
+    // without this we risk a black screen when this unmounts or a blink
+    // if we ignore it. Feel free to come up with a better way.
+    timeout.current = setTimeout(() => {
+      setHasPermission(false);
+    }, 200);
```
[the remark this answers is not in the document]
alexandremgo: Should we clear the timeout in the useEffect-cleanup-return-function just in case though ?

apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidLocationPermission.ts
```diff
@@ -0,0 +1,126 @@
...
+  }, []);
+
+  useEffect(() => {
+    // set hasPermission to false after 400ms if it's still undefined.
+    // without this we risk a black screen when this unmounts or a blink
+    // if we ignore it. Feel free to come up with a better way.
+    timeout.current = setTimeout(() => {
+      setHasPermission(false);
+    }, 200);
+  }, []);
```
alexandremgo, line 57: Same: should we clear the timeout in the useEffect-cleanup-return-function just in case though ?

apps/ledger-live-mobile/src/logic/usePromptBluetoothCallback.ts
```diff
@@ -1,70 +1,58 @@
...
  * @returns a function that turns on bluetooth on the phone
  */
 export function usePromptBluetoothCallback() {
-  const { t } = useTranslation();
-  return useCallback(async (): Promise<boolean> => {
+  return useCallback(async (): Promise<number> => {
     try {
-      return await NativeModules.BluetoothHelperModule.prompt();
+      const result = await NativeModules.BluetoothHelperModule.prompt();
+      return parseInt(result, 10);
```
alexandremgo, line 25: Does `usePromptBluetoothCallback` returns a number or a `BluetoothPromptResult` number enum ? The question is: how do we know that `prompt` won't produce a value > 1 ?

apps/ledger-live-mobile/src/logic/usePromptBluetoothCallback.ts
```diff
@@ -1,70 +1,58 @@
...
-          case E_ENABLE_BLE_UNKNOWN_RESPONSE:
-          case E_SECURITY_EXCEPTION:
-          case E_UNKNOWN_ERROR:
-          default:
-            ToastAndroid.show(
-              t("errors.BluetoothRequired.description"),
-              ToastAndroid.LONG,
-            );
+        if (code === E_BLE_CANCELLED) {
+          return BluetoothPromptResult.OK;
```
alexandremgo, line 30: if BLE was not turned on because it was cancelled, it's still ok ?

apps/ledger-live-mobile/src/logic/usePromptBluetoothCallback.ts
```diff
@@ -1,70 +1,58 @@
...
+  }, []);
 }
 
 export function usePromptBluetoothCallbackWithState() {
   const [bluetoothPromptedOnce, setBluetoothPromptedOnce] = useState(false);
-  const [bluetoothPromptSucceeded, setBluetoothPromptSucceeded] =
-    useState(true);
+  const [bluetoothPromptResult, setBluetoothPromptResult] =
+    useState<BluetoothPromptResult>(BluetoothPromptResult.OK);
   const promptBluetoothCallback = usePromptBluetoothCallback();
```
alexandremgo, line 44: Wait, it does not return a function anymore, but an enum/number ?

apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBLEPermissions.ts
```diff
@@ -0,0 +1,148 @@
...
+      generalStatus: allGranted
+        ? RESULTS.GRANTED
+        : someNeverAskAgain
+        ? RESULTS.NEVER_ASK_AGAIN
+        : RESULTS.DENIED,
+    };
+  });
+}
+
+const useAndroidBLEPermissions = () => {
```
alexandremgo, line 56: Should we put in the same location those new hooks together with `usePromptBluetoothCallback` ?

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
```diff
@@ -0,0 +1,219 @@
...
+        override fun onActivityResult(
+            activity: Activity?,
+            requestCode: Int,
+            resultCode: Int,
+            data: Intent?
+        ) {
+            if (requestCode === REQUEST_ENABLE_LOCATION) {
+                when (resultCode) {
+                    Activity.RESULT_OK -> {
+                        locationPromise?.resolve(SUCCESS_LOCATION_ENABLED);
```
ofreyssinet-ledger, line 53: here and below set back locationPromise to null no?

alexandremgo: Yes you're right !

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
```diff
@@ -0,0 +1,224 @@
...
+            }
+            super.onLocationAvailability(p0)
+        }
+    }
+
+    init {
+        // PRIORITY_HIGH_ACCURACY:  request the most precise location possible.
+        // With this setting, the location services are more likely to use GPS to determine the location.
+        // The given permissions will also influence the level of precision.
+        locationRequest.priority = LocationRequest.PRIORITY_HIGH_ACCURACY
```
ofreyssinet-ledger, line 92: I get this deprecation warning for `PRIORITY_HIGH_ACCURACY`:
```
This constant is deprecated. Use Priority.PRIORITY_HIGH_ACCURACY instead.
```

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
```diff
@@ -0,0 +1,224 @@
...
+ */
+class LocationHelperModule(private val reactContext: ReactApplicationContext) :
+    ReactContextBaseJavaModule(reactContext) {
+
+    // Resolves with a success or error value. reject is not used because we want to pass an
+    // information to the JS code.
+    private var locationPromise: Promise? = null
+
+    // LocationRequest used by both LocationSettings and FusedLocationProviderClient
+    private var locationRequest: LocationRequest = LocationRequest.create();
```
ofreyssinet-ledger, line 37: apparently LocationRequest.create() is deprecated

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
```diff
@@ -0,0 +1,224 @@
...
+            }
+        }
+    }
+
+
+    private var listenerCount = 0
+
+    // Signature to respect to add a listener from JS to which we can send events
+    @ReactMethod
+    fun addListener(eventName: String) {
```
ofreyssinet-ledger, line 170: parameter `eventName` never used

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
alexandremgo: Added comment on why i had to use this method

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
alexandremgo: Added comment on why i had to use this constant

libs/ledger-live-common/src/ble/hooks/useBleDevicesScanning.ts
```diff
@@ -45,6 +45,16 @@ const DEFAULT_RESTART_SCANNING_TIMEOUT_MS = 4000;
  *
  * Note: if a communication is started with a device, the scanning should be stopped
  *
+ * The scanningBleError will depend on the bleTransportListen function. Some errors might not be thrown.
```
ofreyssinet-ledger, line 48: what does it mean "some errors might not be thrown"? Is it "in some cases, no error will be thrown" as in "it will silently fail"?

ofreyssinet-ledger: ok I read the rest of the comment and it's a bit clearer, but the wording here could be improved I think

libs/ledger-live-common/src/ble/hooks/useBleDevicesScanning.ts
```diff
@@ -45,6 +45,16 @@ const DEFAULT_RESTART_SCANNING_TIMEOUT_MS = 4000;
...
+ * For ex:
+ * - on Android, using the current Transport from react-native-hw-transport-ble, if the bluetooth is off,
+ * only a BluetoothScanStartFailed error is thrown. But this error could happen for other reason than the BLE being off.
+ * No actual "BluetoothOff" or "BluetoothUnauthorized" error is thrown.
+ * On the other side, if location is off, an error is thrown.
+ * - On iOS, using the current Transport from react-native-hw-transport-ble, if the bluetooth is off, no error is thrown at all.
+ * Hence, handling if the bluetooth is off or not is not handled here. It should be handled by the consumer of this hook.
```
ofreyssinet-ledger, line 55: Maybe highlight this sentence by putting it higher and properly warning the consumer that it's his job to handle the bluetooth activation and fallback logic

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
```diff
@@ -574,14 +574,15 @@ export default class BluetoothTransport extends Transport {
...
 const bleErrorToHwTransportError = new Map([
-  [BleErrorCode.ScanStartFailed, HwTransportErrorType.BleScanStartFailed],
+  [BleErrorCode.ScanStartFailed, HwTransportErrorType.BluetoothScanStartFailed],
   [
     BleErrorCode.LocationServicesDisabled,
-    HwTransportErrorType.BleLocationServicesDisabled,
+    HwTransportErrorType.LocationServicesDisabled,
   ],
   [
+    // BluetoothUnauthorized actually represents a location service unauthorized error
```
ofreyssinet-ledger, line 583: :ohno:
thanks for investigating and clarifying that in the error name!

apps/ledger-live-mobile/src/screens/Settings/Debug/Features/BLEPairingFlow.tsx
```diff
@@ -60,7 +60,7 @@ export default () => {
     pairedDevice: null,
   };
 
-  const promptBluetooth = usePromptBluetoothCallback();
+  const promptBluetooth = usePromptEnableBluetoothCallback();
 
   const goToBlePairingFlow = useCallback(() => {
```
ofreyssinet-ledger, line 65: related to the code below inside the callback, which actually didn't change, but maybe should:

if we first prompt the enabling of Bluetooth _before_ navigating to the bluetooth  device pairing screen, aren't we creating the impression, for a person checking that debugging screen, that the screen itself is handling this enabling of bluetooth?
That screen actually does it, but imagine it breaks in that screen / gets removed, then it won't be detected here, because in any case we try to enable Bluetooth here before actually navigating to the screen (which is supposed to do that too).

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicesScanning.tsx
```diff
@@ -29,6 +30,10 @@ export type BleDevicesScanningProps = {
 /**
  * Runs a BLE scan and list seen devices
  *
+ * This components should be wrapped around a RequiresBLE component.
```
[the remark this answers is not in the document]
alexandremgo: ahah yeah i don't know how to write english properly - and i have Copilot right now with me for this, so i guess i'm a lost cause

apps/ledger-live-mobile/src/components/RequiresBLE/AndroidRequiresBluetoothPermissions.tsx
```diff
@@ -1,118 +1,45 @@
...
 /**
- * Renders an error if location is required & not available,
- * otherwise renders children
+ * Renders an error if bluetooth is required & its associated permissions are not
+ * set. Otherwise renders children.
+ *
+ * Should only be used for Android.
+ *
+ * @param hasBackButtonOnDenied If true, the back button will be displayed on the permission denied screen. Defaults to false.
+ * @param openSettingsOnErrorButton Used for debug purposes. If true, on a permission denied, pressing the button on
```
ofreyssinet-ledger, line 18: if it's for debug purposes only can we change the name of the prop? like prefix it with "debugOnlyDONOTUSE" or some thing like that which will jump to any reviewer's eyes so it's not used in real user situations haha

apps/ledger-live-mobile/src/components/GenericErrorView.tsx
```diff
@@ -41,15 +41,6 @@ const GenericErrorView = ({
   Icon = CloseMedium,
   iconColor = "error.c100",
 }: Props) => {
-  const promptBluetooth = usePromptBluetoothCallback();
```
ofreyssinet-ledger, line 44: nice to finally remove this horror haha thanks

apps/ledger-live-mobile/src/components/RequiresBLE/AndroidRequiresBluetoothPermissions.tsx
alexandremgo: I mean. it could be use in a scenario where a consumer of this component wants to force opening the user settings when pressing the button when there is an error

But i don't really see a case currently where we would want to force this

So i guess yeah we should rename it with `debugOnlyOpenSettingsOnErrorButton` ?

apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBluetoothPermissions.ts
```diff
@@ -0,0 +1,191 @@
...
+        ? RESULTS.GRANTED
+        : someNeverAskAgain
+        ? RESULTS.NEVER_ASK_AGAIN
+        : RESULTS.DENIED,
+    };
+  });
+}
+
+/**
+ * Hook to check and request the BLE permission on Android.
```
ofreyssinet-ledger, line 73: Maybe here document what happens at mount (automatic check and request of the permission)

apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBluetoothPermissions.ts
```diff
@@ -0,0 +1,191 @@
...
+      setHasPermission(false);
+    }, 400);
+
+    return () => {
+      if (timeout.current) clearTimeout(timeout.current);
+    };
+  }, []);
+
+  /**
+   * Goes to state="background" when for ex:
```
ofreyssinet-ledger, line 114: Maybe document this in the body of the listener itself, and rephrase to something like "The value of the state will be "background" when for ex ...".
"Goes" is a bit ambiguous

apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBluetoothPermissions.ts
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @returns an object with the following properties:
+ * - hasPermission: true if the permission is granted, false if it's denied, undefined if it's still loading
+ * - checkAgain: a function to check (and not request) again the permission
+ * - neverAskAgain: true if the user has checked the "never ask again" checkbox (or on Android 11+ if the user has denied the permission several times)
+ * - renderChildren: a boolean to know if the children should be rendered (permissions are granted) or not
+ */
+export const useAndroidBluetoothPermissions = () => {
+  const [hasPermission, setHasPermission] = useState<boolean | undefined>();
+  const [check, setCheck] = useState(false);
```
ofreyssinet-ledger, line 83: I think the naming of "hasPermission" and "check" is not clear (check being a verb, while it's a boolean, maybe checkResult is more appropriate?), and maybe the structure (separating in two states) is weird.
Why are we separating the result of the permission check and the result of the permission request? In the end these two resolve in a state that means "the permissions is / isn't granted", no?
Also i see that `check` and `hasPermission`  are only used in one place, like this: `check || hasPermission` so I feel like we can get rid of one of them.

If you what you want to track is whether the initial check has resolved, then track it in a state with a name that clearly illustrates that.

alexandremgo: Yeah i agree, i'm going to re-work this hook 👌

apps/ledger-live-mobile/src/components/ServiceDisabledView.tsx
```diff
@@ -0,0 +1,109 @@
...
+  ButtonIcon?: IconType;
+  onButtonPress: () => void;
+  hasBackButton?: boolean;
+  buttonEvent?: string;
+};
+
+/**
+ * View component used by specific service disabled error components
+ *
+ * @param title String displayed as a title on the view
```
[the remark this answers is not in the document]
ofreyssinet-ledger: (not suggesting a change, I'm just wondering what's the best way to document props)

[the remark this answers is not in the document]
ofreyssinet-ledger: I think this way will break easily for instance if some properties of the `Props` type change and you forget to update this comment

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
```diff
@@ -0,0 +1,267 @@
...
+     */
+    private val locationServicesUpdateCallback = object : LocationCallback() {
+        override fun onLocationAvailability(p0: LocationAvailability) {
+            val isLocationAvailable = p0.isLocationAvailable
+
+            if (isLocationAvailable) {
+                reactContext
+                    .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter::class.java)
+                    .emit("LocationServiceUpdated", Arguments.createMap().apply {
+                        putString("service", "enabled")
```
ofreyssinet-ledger, line 84: why not just use a boolean here?

apps/ledger-live-mobile/src/components/RequiresBLE/AndroidRequiresBluetoothPermissions.tsx
ofreyssinet-ledger: 🤷‍♂️ yeah or if you think it's not debug only, "forceOpenSettingsOnErrorButton", just something to make a consumer 2nd guess if he really wants to use this prop.
To be honest not super important 😆

apps/ledger-live-mobile/src/components/RequiresBLE/BluetoothPermissionDenied.tsx
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Renders a component that informs the user that they have denied the app permission to use Bluetooth.
+ *
+ * If the user has selected/triggered "Don't ask again", the user will be prompted to open the native settings
+ *
+ * @param onRetry A callback for the user to retry allowing the permission, if neverAskAgain and openSettings are false.
+ *   Otherwise, the user will be prompted to open the native settings.
+ * @param hasBackButton If true, a back button will be displayed in the header. Default to false.
+ * @param neverAskAgain If true, the user has denied the app permission to use Bluetooth and has selected "Don't ask again". Default to true.
+ * @param openSettings Used for debug purposes. If true pressing the button will make the user go to the settings. Defaults to false.
```
ofreyssinet-ledger, line 24: In that case it's not only for debug purposes right? It's also used in case the permission cannot be requested again (on android) ?

apps/ledger-live-mobile/src/components/ServicePermissionDeniedView.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Text, Flex } from "@ledgerhq/native-ui";
+import { ArrowLeftMedium } from "@ledgerhq/native-ui/assets/icons";
+import { TouchableOpacity } from "react-native-gesture-handler";
+import { IconType } from "@ledgerhq/native-ui/components/Icon/type";
+import { useNavigation } from "@react-navigation/native";
+import Button from "./Button";
+
+type Props = {
+  title: string;
+  TitleIcon?: React.ReactNode;
```
ofreyssinet-ledger, line 12: if it's a ReactNode, don't capitalize, only capitalize if it's a React Component (something that you can pass props to)

apps/ledger-live-mobile/src/components/ServicePermissionDeniedView.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <Text
+          mb={10}
+          variant={"body"}
+          fontWeight={"medium"}
+          textAlign="center"
+          color={"neutral.c80"}
+        >
+          {description}
+        </Text>
+        <Text variant={"h3"} mb={8} mt={24} px={20} textAlign="center">
```
ofreyssinet-ledger, line 84: For spacing values, you should always try to use [values from the design system](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ui/packages/native/src/styles/theme.ts#L4) by specifying their index in this array. If a value is missing then use the closest value (and check with the design team if unsure).

apps/ledger-live-mobile/src/components/ServicePermissionDeniedView.tsx
```diff
@@ -0,0 +1,109 @@
...
+          {subTitle}
+        </Text>
+        <Flex alignSelf="stretch" mb={24} px={36}>
```
ofreyssinet-ledger, line 87: same comment as above here for the spacing values

apps/ledger-live-mobile/src/components/RequiresLocation/hooks/useAndroidEnableLocation.ts
```diff
@@ -0,0 +1,98 @@
...
+    try {
+      if (Platform.OS === "android") {
+        const response =
+          await LocationHelperModule.checkAndRequestEnablingLocationServices();
+        return response;
+      }
+
+      return "ERROR_UNKNOWN";
+    } catch (e) {
+      // Should never be reached, a checkAndRequestEnablingLocationServices only resolves
```
ofreyssinet-ledger, line 28: never reached ... unless the native `checkAndRequestEnablingLocationServices` gets removed/renamed in which case you'll get a proper error.
Since there is no straightforward way today to enforce that the contract is respected (that the method is implemented on the native side), I think we should at least log the error (probably even throw it)

apps/ledger-live-mobile/src/components/RequiresBLE/AndroidRequiresBluetoothPermissions.tsx
alexandremgo: i like `forceOpenSettingsOnErrorButton` , i'll use this 👌

apps/ledger-live-mobile/src/screens/Settings/Debug/Features/BLEPairingFlow.tsx
alexandremgo: Especially, now we request if the component `RequiresBLE` is used (as it should be). So if the user refuses once on the prompt coming from this ` const promptBluetooth = usePromptEnableBluetoothCallback();`, then it is again prompt by the `RequiresBLE` just after

apps/ledger-live-mobile/src/components/RequiresLocation/hooks/useAndroidEnableLocation.ts
alexandremgo: Yes you're right - i'm not going to throw here, because then i would try/catch where it is used below, and do basically the same (`setLocationServicesState("disabled")` + log)

So i'm going to log the error, add a new possible returned value `"LOCATION_MODULE_THREW_ERROR"` and return this

apps/ledger-live-mobile/android/app/src/main/java/com/ledger/live/LocationHelperModule.kt
alexandremgo: I think it's more declarative than just saying "true/false", but also this value is serialised (IMO), so i don't think it changes anything

PR #2283, opened 2023-01-10
LLM Camera permissions rework / bugfix NFT claim

ofreyssinet-ledger:
### 📝 Description

- Initial context: 
   In case the permission was not granted yet, the `Camera` component was still already mounted in some cases, and that caused an issue on iOS as the `Camera` component doesn't refresh automatically its internal state depending on the permission, so it has to be mounted in a context where the permission is already granted, or it won't work if the permission gets granted _while_ it is mounted.
   
After encountering other bugs while trying to fix this one, I decided to implement a simple component `RequiresCameraPermissions` and remove the old logic that was used in particular for the _Scan to Import LLD accounts_ flow.

Changes:
- removal of legacy logic for checking camera permissions, also removal of some dead code (`FallbackCameraSend` component)
- new component `RequiresCameraPermissions` to handle that logic simply by wrapping any other component.
- debug screen for camera permissions

### ❓ Context

- **Linked resource(s)**: [FAT-775] 

### ✅ Checklist

- [ ] **Test coverage**  Cannot be covered by a test as camera capabilities aren't available on iOS emulators
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/91890529/211530181-c83ee10f-d74d-4b5f-a21d-b229309d70dd.MP4

### 🚀 Expectations to reach

1. Fresh LLM install
2. Go to post onboarding hub
3. Start scan NFT claim flow
4. Press "Claim ledger market pass" to navigate to the scanning screen
5. When prompted, grant the Camera permission to LLM.
6. **The QR code scanning zone should display whatever the camera sees (and scan a QR code if presented)**

[FAT-775]: https://ledgerhq.atlassian.net/browse/FAT-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/screens/ClaimNft/ClaimNftQrScan.tsx
```diff
@@ -163,9 +165,7 @@ const ClaimNftQrScan = () => {
   return (
     <SafeAreaView style={{ flex: 1 }} edges={["bottom"]}>
       <Flex flex={1}>
-        {!permission?.canAskAgain &&
-        !permission?.granted &&
-        permission?.status === "denied" ? (
```
ofreyssinet-ledger, line 168: This is unrelated to the fix, just a small cleanup:
`permission.granted` is a helper that abstracts into a boolean `permission.status`, in this specific case checking the value of `permission.granted` is enough.

apps/ledger-live-mobile/src/screens/ClaimNft/ClaimNftQrScan.tsx
```diff
@@ -174,13 +174,13 @@ const ClaimNftQrScan = () => {
         ) : (
           <>
             <Flex
-              backgroundColor="neutral.c40"
+              backgroundColor="constant.black"
```
ofreyssinet-ledger, line 177: If the camera component is unmounted, using `constant.black` as a background will make this unmounted state unnoticeable.

thomasrogerlux (approved):

apps/ledger-live-mobile/src/components/RequiresCameraPermissions/useCameraPermissions.ts
```diff
@@ -0,0 +1,59 @@
...
+    requestPermission().then(() => setFirstAutomaticRequestCompleted(true));
+    // only run this once on mount
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const appState = useRef(AppState.currentState);
+  const shouldCheckPermissionOnNextResume = useRef(false);
+  useEffect(() => {
+    const subscription = AppState.addEventListener("change", nextAppState => {
+      if (
+        appState.current.match(/inactive|background/) &&
```
alexandremgo, line 23: Interested for me on Bluetooth and location: is it needed to check that the previous app state was `inactive` or `background` ? The `addEventListener` reacts on `change`: does it trigger `active` several times in a row ?

apps/ledger-live-mobile/src/components/RequiresCameraPermissions/useCameraPermissions.ts
```diff
@@ -0,0 +1,59 @@
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { Camera } from "expo-camera";
+import { AppState, Linking } from "react-native";
+import { useIsMounted } from "../../helpers/useIsMounted";
+
+export default function useCameraPermissions() {
+  const [permission, requestPermission, checkPermission] =
+    Camera.useCameraPermissions();
```
alexandremgo, line 8: That's cheating ahah so much easier - i wish expo had an api for BLE and location permissions

apps/ledger-live-mobile/src/screens/Settings/Debug/Debugging/CameraPermissions.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useState } from "react";
+import { Button, Flex, Switch, Text } from "@ledgerhq/native-ui";
+import { Camera } from "expo-camera";
+import { useIsFocused } from "@react-navigation/native";
+import { CameraType } from "expo-camera/build/Camera.types";
+import RequiresCameraPermissions from "../../../../components/RequiresCameraPermissions";
+import CameraPermissionContext from "../../../../components/RequiresCameraPermissions/CameraPermissionContext";
+
+const CameraPermissions: React.FC<{}> = () => {
```
alexandremgo, line 9: TS is complaning that `{}` means any non-nullish value.

Either remove the `React.FC<{}>` type all together and let TS infers the type. Or use `React.FC<Record<string, never>>` for the type

alexandremgo (commented):
Way cleaner than before ! 👌

Just a small comment on a type, but otherwise all good !

apps/ledger-live-mobile/src/components/RequiresCameraPermissions/useCameraPermissions.ts
ofreyssinet-ledger: I don't know exactly, I think there are ways it can happen on Android. This is from the [react-native documentation template](https://reactnative.dev/docs/appstate#basic-usage).

apps/ledger-live-mobile/src/screens/Settings/Debug/Debugging/CameraPermissions.tsx
ofreyssinet-ledger: nice nice thank you!

PR #2290, opened 2023-01-10
✨ [FEAT] : LIVE-4909 - LLM- MultiSelect-Hide-NFTs

mcayuelas-ledger:
### 📝 Description

Multi Select Hide NFTs

A way to select NFTs to hide them

### ❓ Context

- **Linked resource(s)**: [LIVE-4909] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

(New UI 💄)

https://user-images.githubusercontent.com/112866305/215436562-8b2703a1-c226-4159-82f6-1f4e84887004.mov

### 🚀 Expectations to reach

[LIVE-4909]: https://ledgerhq.atlassian.net/browse/LIVE-4909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/screens/Nft/WalletNftGallery/index.tsx
```diff
@@ -27,14 +29,18 @@ const WalletNftGallery = () => {
...
   return (
     <>
-      <Box mx={6}>
+      <Box>
         {hasNFTs ? (
           <NftList data={nftsOrdered} />
         ) : (
           <CollapsibleHeaderScrollView
-            contentContainerStyle={{ paddingTop: 0 }}
+            contentContainerStyle={{ paddingTop: 0, marginHorizontal: 18 }}
```
nparigi-ledger, line 43: use theme values

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -0,0 +1,108 @@
...
+      }),
+    });
+
+    readOnlyModeAction();
+  }, [readOnlyModeAction, dispatch, nftsToHide, pushToast, t]);
+
+  const multiSelectModeAction = useCallback(() => {
+    setNftsToHide([]);
+    setMultiSelectMode(true);
+    dispatch(updateMainNavigatorVisibility(false));
```
ofreyssinet-ledger, line 64: What happens if you quit the app right after this happens? (without finishing the selection)
Then restart it, won't the tab bar be hidden, putting the app in a blocked state? Because values of the `settings` part of the store are persisted. In my opinion even if you have some logic to "unhide it" automatically on mount (which could solve this issue), it's a bit dangerous to save that state persistently in the settings.

ofreyssinet-ledger: Also for "safety" I'd also call a `dispatch(updateMainNavigatorVisibility(true)` in the cleanup function of a `useEffect` to ensure that we return to a normal state when this hook here is unmounted.

mcayuelas-ledger: https://user-images.githubusercontent.com/112866305/214809677-1bcde5bb-5544-40af-8c52-b0023d2423da.mov

Well yes you're right but seems working well

To solve that maybe create a new part of store that is not written persistently?

ofreyssinet-ledger: You can use `appstate` for that I think

ofreyssinet-ledger: Right now it works because the "crypto/nft tab" is persisted as well so the app always starts on that tab if you left it there, but we should not depend on this logic. The solution is to not persist that value.

On the other point I mentioned, where you should call `dispatch(updateMainNavigatorVisibility(true)` when the hook unmounts (so in the cleanup function), here's a (twisted) example with a video to illustrate:

1. user starts selecting
2. puts the app in background
3. opens a deeplink to Ledger (this could happen for many reasons
4. user is stuck on the page that the deeplink brought them to


https://user-images.githubusercontent.com/91890529/214820570-efa5a5bb-a228-4ff7-aff0-52d28518773b.mp4

ofreyssinet-ledger: Another thing that would be a nice to have regarding this "selection mode" is to take into account "back button presses / gestures" (especially on Android) to cancel the selection mode.

There is an example of how it can be done in `TabBar/Transfer.tsx`. If you do it don't forget to return a boolean value in the listener function that you pass to the `BackHandler` (always return false is the selection mode is off, to let the event bubble up to other listeners and eventually to the native handler). Cf. https://reactnative.dev/docs/backhandler

ofreyssinet-ledger: ^ the reason we need this is that this is the expected behaviour when navigating, and since you're not doing a real navigation here (but a real one in terms of UX), back gesture behaviour is not implemented but would be expected by the user

apps/ledger-live-mobile/src/components/TabBar/CustomTabBar.tsx
```diff
@@ -213,4 +216,29 @@ export default function CustomTabBar({
...
+            state={state}
+            descriptors={descriptors}
+            navigation={navigation}
+            colors={colors}
+            insets={insets}
+          />
+        </Animated.View>
+      )}
+    </Animated.View>
+  );
```
ofreyssinet-ledger, line 243: this entering/exiting animation is neat 👍

apps/ledger-live-mobile/src/components/Nft/SelectNft.tsx
```diff
@@ -0,0 +1,43 @@
...
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+`;
+
+type Props = {
+  isSelected: boolean;
+};
+
+export const SelectNft = ({ isSelected }: Props) => {
```
ofreyssinet-ledger, line 31: maybe rename to `NftSelectionCheckbox` or something more accurate like that?

apps/ledger-live-mobile/src/components/Nft/NftListItem.tsx
```diff
@@ -188,9 +232,37 @@ const NftMediaComponent = ({ status, metadata, nftAmount }: NftMediaProps) => {
...
+  }
+
+  if (selectable) {
+    return (
+      <Box position="relative">
+        <NftMedia
+          style={[
+            styles.image,
+            {
+              opacity: isSelected ? 0.3 : selectable ? 0.7 : 1,
```
ofreyssinet-ledger, line 252: is it expected that the opacity values are different from those used in the previous if block?

IMO this entire component could be simplified/refactored slightly by returning 1 block JSX with the conditions inside. It would ensure consistency for the common props & style

Something like that:
```
const NftMediaComponent = () => {
  return (
    <Box>
      <NftMedia style={yourStyleLogicDependingOnSelectableEtc} otherPropsHere />
      {moreThanOne ? <TheTagThatDisplaysTheCount /> : null}
      {selectable ? <SelectableCheckboxIndicator isSelected={isSelected}> : null}
    </Box>
  )
}
```

apps/ledger-live-mobile/src/components/Nft/NftListItem.tsx
```diff
@@ -158,6 +186,9 @@ const NftListItem = ({ nft, onPress }: Props) => {
       collectionStatus={collectionStatus}
       collectionMetadata={collectionMetadata}
       onPress={handlePress}
+      onLongPress={handleLongPress}
```
ofreyssinet-ledger, line 189: here I think you don't need an intermediary handleLongPress (that does nothing more than calling onLongPress), you can directly pass `onLongPress`

apps/ledger-live-mobile/src/components/Nft/NftList.tsx
```diff
@@ -67,30 +86,123 @@ export function NftList({ data }: Props) {
...
+const StyledButton = styled(Button)`
+  padding: 0;
+  margin: 0;
+`;
+
+const ButtonsContainer = styled(Flex)`
+  position: absolute;
+  bottom: 20px;
+  z-index: 5;
+  padding: 0 18px;
```
ofreyssinet-ledger, line 207: please no hardcoded px value, use a value from the design system (take the closest one if it doesn't exist and give feedback to designers)

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -0,0 +1,108 @@
...
+      icon: "success",
+      title: t("wallet.nftGallery.filters.alertHide", {
+        count: nftsToHide.length,
+      }),
+    });
+
+    readOnlyModeAction();
+  }, [readOnlyModeAction, dispatch, nftsToHide, pushToast, t]);
+
+  const multiSelectModeAction = useCallback(() => {
```
ofreyssinet-ledger, line 61: maybe rename with a verb, like `triggerMultiSelectMode`

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -0,0 +1,108 @@
...
+const TOAST_ID = "SUCCESS_HIDE";
+
+export function useNftList() {
+  const isFocused = useIsFocused();
+  const dispatch = useDispatch();
+  const { t } = useTranslation();
+  const { pushToast } = useToasts();
+  const navigation = useNavigation();
+
+  const [onMultiSelectMode, setMultiSelectMode] = useState<boolean>(false);
```
ofreyssinet-ledger, line 25: rename it to make it clearer that it's a boolean, e.g: `multiSelectModeEnabled`

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const [nftsToHide, setNftsToHide] = useState<ProtoNFT[]>([]);
+
+  // Multi Select ------------------------
+
+  const readOnlyModeAction = useCallback(() => {
```
ofreyssinet-ledger, line 31: Why is this called readOnlyModeAction? Shouldn't it be something closer to what it does, like `exitMultiSelectMode`. Also it should probably be a verb.

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -0,0 +1,108 @@
...
+    readOnlyModeAction();
+  }, [readOnlyModeAction, dispatch, nftsToHide, pushToast, t]);
+
+  const multiSelectModeAction = useCallback(() => {
+    setNftsToHide([]);
+    setMultiSelectMode(true);
+    dispatch(updateMainNavigatorVisibility(false));
+  }, [dispatch]);
+
+  const updateListSelect = useCallback(
```
ofreyssinet-ledger, line 67: Maybe rename this to something like `handleSelectableNftPressed` or `toggleNftSelection`

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -0,0 +1,108 @@
...
+  }, [dispatch]);
+
+  // Detect change of screen
+  useEffect(() => {
+    if (!isFocused) {
+      readOnlyModeAction();
+    }
+  }, [readOnlyModeAction, isFocused]);
+
+  const onClickHide = useCallback(() => {
```
ofreyssinet-ledger, line 44: `hideSelectedNfts` might be better, imagine that the consumer of this hook is not necessarily going to use that for a button

ofreyssinet-ledger: (but i'm being nitpicky here lol it's good like that)

ofreyssinet-ledger (changes requested):
Nice!

- Just a few changes 100% needed IMO:
  - Moving the "selection mode enabled" boolean out of a persistent store, to avoid leaving the app in a blocked state.
  - Disabling this "selection mode enabled" in the cleanup function of a useEffect in the NftList.hook, so that when any component using this hook unmounts, we put back the app state to normal (with a visible tab bar). This will avoid weird issues.
- A nice to have (but more related to UX): a back gesture/press handler to have a normal/expected navigation UX when the multi selection mode is enabled.
- Bonus:
  - some cleanup refactoring of the NftMedia component, to make its rendering logic simpler and less likely to be broken by future modifications.
  - some renaming of functions & variables here and there for better readability.

apps/ledger-live-mobile/src/components/Nft/NftList.hook.ts
```diff
@@ -37,9 +38,26 @@ export function useNftList() {
...
     }
-  }, [readOnlyModeAction, isFocused]);
+
+    return () => {
+      exitMultiSelectMode();
+    };
+  }, [exitMultiSelectMode, isFocused, dispatch]);
+
+  const handleBackPress = useCallback(() => {
+    exitMultiSelectMode();
```
ofreyssinet-ledger, line 50: Here you have to handle the case where you're not in multiSelectMode and then return false (so that other eventual other listeners (somewhere else in the app) can do their job, for instance the default native one which will trigger an app close in that case)

mcayuelas-ledger:
Now should be ok for `TabBar `
https://user-images.githubusercontent.com/112866305/214877335-4c69054b-8599-4808-b1be-b453811ba54e.mov

apps/ledger-live-mobile/src/components/Nft/NftList.tsx
```diff
@@ -1,16 +1,42 @@
...
+import { TrackScreen } from "../../analytics";
+import { useNftList } from "./NftList.hook";
+import BackgroundGradient from "../TabBar/BackgroundGradient";
+
+const darkGradients = [
+  {
+    height: 145,
+    opacity: 1,
+    stops: [
+      <Stop key="0%" offset="0%" stopOpacity={0} stopColor="#131214" />,
```
cgrellard-ledger, line 22: Here you can use the color background.main from the DS, it's the same one. That way you don't need two gradients anymore

apps/ledger-live-mobile/src/components/Nft/NftList.tsx
```diff
@@ -1,16 +1,42 @@
...
+    ],
+  },
+];
+
+const lightGradients = [
+  {
+    height: 145,
+    opacity: 1,
+    stops: [
+      <Stop key="0%" offset="0%" stopOpacity={0} stopColor="#ffffff" />,
```
cgrellard-ledger, line 34: Same as above

apps/ledger-live-mobile/src/components/Nft/NftList.tsx
mcayuelas-ledger: Yes indeed ;)

cgrellard-ledger (approved):

PR #2296, opened 2023-01-10
feat: update test for fee estimates to use scaling usd price

Sheng-Long:
### 📝 Description

Update Hedera fee estimate and get usd price if possible for estimate.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledger-live-common/src/families/hedera/tests/js-estimateMaxSpendable.test.ts
```diff
@@ -50,8 +51,27 @@ const account: Account = {
...
 };
 
-describe("js-estimateMaxSpendable", () => {
-  const estimatedFees = new BigNumber("83300").multipliedBy(2);
+describe("js-estimateMaxSpendable", async () => {
+  // If get hedera price works, use real estimate, otherwise fallback to hard coded
+  let hederaPrice;
+
+  try {
+    const { data } = await axios.get(
```
henrily-ledger, line 59: Because you add some network logic we will need to change it to an integration test.
It's our way to not have network dependency for our CI.
can you rename js-estimateMaxSpendable.test.ts to js-estimateMaxSpendable.integration.test.ts ?

Or you can also change it with a mock test with predicted data

henrily-ledger (changes requested):

libs/ledger-live-common/src/families/hedera/tests/js-estimateMaxSpendable.test.ts
Sheng-Long: This makes sense, will change soon.  Thanks.  I believe that it is desired that we check using the network dependency as this test is currently failing because the value has changed.  To prevent future failures we will need to check a dynamic value to determine the fees.

Sheng-Long: Sorry for the delay, I had accidentally changed this on the wrong branch at first and didn't notice, but it is changed now

haammar-ledger (commented):
The test looks good, but don't you want to also update the actual implementation in `js-estimateMaxSpendable.ts` ?

libs/ledger-live-common/src/families/hedera/utils.ts
```diff
@@ -2,13 +2,35 @@ import estimateMaxSpendable from "./js-estimateMaxSpendable";
...
+
+  let estimatedFees = new BigNumber("212800"); // 0.002128 ℏ (as of 2023-01-09)
+
+  if (hederaPrice) {
+    estimatedFees = new BigNumber("0.0001").dividedBy(
+      new BigNumber(hederaPrice)
+    );
+  }
+
+  return estimatedFees;
```
haammar-ledger, line 31: Just a more concise proposal, if it makes sense.

```suggestion
  try {
    const { data } = await axios.get(
      "https://countervalues.live.ledger.com/latest/direct?pairs=hbar:usd"
    );

    if (data[0]) {
      const hederaPrice = new BigNumber(data[0]);
      return new BigNumber("0.0001").dividedBy(hederaPrice);
    }
  } catch {}
  
  return new BigNumber("212800"); // 0.002128 ℏ (as of 2023-01-09)
```

haammar-ledger (commented):
Hi @Sheng-Long, is this PR still draft?
If not could you please just rebase on develop? Thanks a lot.

libs/ledger-live-common/src/families/hedera/utils.ts
```diff
@@ -2,13 +2,27 @@ import estimateMaxSpendable from "./js-estimateMaxSpendable";
...
 
-// NOTE: Hedera declares stable fees in USD
-//       If we can get the current USD/HBAR price here..
-//       > transfer fee is 0.0001 USD
-export const estimatedFees = new BigNumber("83300"); // 0.000833 ℏ (as of 2021-09-20)
 export const estimatedFeeSafetyRate = 2;
 
+export async function getEstimatedFees(): Promise<BigNumber> {
+  // If get hedera price works, use real estimate, otherwise fallback to hard coded
+  let hederaPrice;
```
haammar-ledger, line 11: Leftover unused variable, you can remove it.

libs/ledger-live-common/src/families/hedera/utils.ts
```diff
@@ -2,13 +2,24 @@ import estimateMaxSpendable from "./js-estimateMaxSpendable";
 import BigNumber from "bignumber.js";
 import type { Account } from "@ledgerhq/types-live";
 import type { Transaction } from "./types";
+import axios from "axios";
```
haammar-ledger, line 5: Not mandatory, but you should use our `network` util if possible, the main benefit is that it uses our logs()

haammar-ledger (changes requested):
A couple errors raised from the tests: https://github.com/LedgerHQ/ledger-live/actions/runs/4316534468/jobs/7532432216#step:9:3139

Sheng-Long:
There seems to be an issue with snapshots, but this doesn't seem diagnosable on my side.  Let me know if you have a way for me to fix this

PR #2320, opened 2023-01-13
FAT-736 FAT-766 FAT-767: New device action for the Firmware update

grsoares21:
### 📝 Description
This creates a new device action using the new proposed device action paradigm for performing firmware updates (which weren't handled by device actions so far).
This PR contains only the first milestone for a complete firmware update, which works only for updates that require a SE upgrade only (defined in [FAT-736]). Updates including flashes to the Bootloaded and MCU will come in a different PR.  

### ❓ Context

- **Linked resource(s)**: [FAT-736](https://ledgerhq.atlassian.net/browse/FAT-736), [FAT-766](https://ledgerhq.atlassian.net/browse/FAT-766), [FAT-767](https://ledgerhq.atlassian.net/browse/FAT-767)

### ✅ Checklist

- [Not yet] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes**

### 📸 Demo
new-fw-update-dev-action

### 🚀 Expectations to reach

[FAT-736]: https://ledgerhq.atlassian.net/browse/FAT-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FAT-736]: https://ledgerhq.atlassian.net/browse/FAT-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FAT-766]: https://ledgerhq.atlassian.net/browse/FAT-766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FAT-767]: https://ledgerhq.atlassian.net/browse/FAT-767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/cli/src/commands-index.ts
+++ b/apps/cli/src/commands-index.ts
@@ -15,6 +15,7 @@ import derivation from "./commands/derivation";
 import devDeviceAppsScenario from "./commands/devDeviceAppsScenario";
 import deviceAppVersion from "./commands/deviceAppVersion";
 import deviceInfo from "./commands/deviceInfo";
+import deviceSDKFirmwareUpdate from "./commands/deviceSDKFirmwareUpdate";
 import deviceSDKGetDeviceInfo from "./commands/deviceSDKGetDeviceInfo";
 import deviceVersion from "./commands/deviceVersion";
 import discoverDevices from "./commands/discoverDevices";
@@ -75,6 +76,7 @@ export default {
   devDeviceAppsScenario,
   deviceAppVersion,
   deviceInfo,
+  deviceSDKFirmwareUpdate,
   deviceSDKGetDeviceInfo,
   deviceVersion,
   discoverDevices,
--- a/libs/ledger-live-common/src/api/Manager.ts
+++ b/libs/ledger-live-common/src/api/Manager.ts
@@ -163,6 +163,7 @@ const listCategories = async (): Promise<Array<Category>> => {
   return r.data;
 };
 
+// TODO: correctly type the endpoint answer
 const getMcus: () => Promise<any> = makeLRUCache(
   async () => {
     const { data } = await network({
@@ -571,6 +572,28 @@ const installMcu = (
   }).pipe(remapSocketError(context));
 };
 
+function retrieveMcuVersion(
+  finalFirmware: FinalFirmware
+): Promise<McuVersion | undefined> {
+  return getMcus()
+    .then((mcus) =>
+      mcus.filter((deviceInfo) => {
+        const provider = getProviderId(deviceInfo);
+        return (mcu) => mcu.providers.includes(provider);
+      })
+    )
+    .then((mcus) =>
+      mcus.filter((mcu) => mcu.from_bootloader_version !== "none")
+    )
+    .then((mcus) =>
+      findBestMCU(
+        finalFirmware.mcu_versions
+          .map((id) => mcus.find((mcu) => mcu.id === id))
+          .filter(Boolean)
+      )
+    );
+}
+
 const API = {
   applicationsByDevice,
   listApps,
@@ -588,5 +611,6 @@ const API = {
   install,
   genuineCheck,
   installMcu,
+  retrieveMcuVersion,
 };
 export default API;
--- a/libs/ledger-live-common/src/deviceSDK/actions/updateFirmware.ts
+++ b/libs/ledger-live-common/src/deviceSDK/actions/updateFirmware.ts
@@ -1,0 +1,125 @@
+import { DeviceId } from "@ledgerhq/types-live";
+import { Observable, of } from "rxjs";
+import { scan, switchMap } from "rxjs/operators";
+import {
+  updateFirmwareTask,
+  UpdateFirmwareTaskEvent,
+} from "../tasks/updateFirmware";
+import {
+  getDeviceInfoTask,
+  GetDeviceInfoTaskErrorEvent,
+} from "../tasks/getDeviceInfo";
+import {
+  FullActionState,
+  initialSharedActionState,
+  sharedReducer,
+} from "./core";
+import {
+  getLatestFirmwareTask,
+  GetLatestFirmwareTaskErrorEvent,
+} from "../tasks/getLatestFirmware";
+
+export type updateFirmwareActionArgs = {
+  deviceId: DeviceId;
+};
+
+export type UpdateFirmwareActionState = FullActionState<{
+  step:
+    | "preparingUpdate"
+    // initial step where we retrieve all the necessary information for the update
+    | "allowSecureChannelRequested"
+    // we need a secure connection to the HSM to install the operating system updater (osu)
+    | "installingOsu"
+    // step that installs the operating system updater, always happen
+    | "installOsuDevicePermissionRequested"
+    // permission requested for istalling the osu
+    | "installOsuDevicePermissionGranted"
+    // the user has confirmed the firmware installation on the device
+    | "installOsuDevicePermissionDenied"
+    // the user has refused the firmware installation on the device
+    | "flashingMcu"
+    // step that flashes the mcu of the device, only happens when necessary
+    | "flashingBootloader"
+    // step that flashes the bootloader of the device, only happen when necessary
+    | "firmwareUpdateCompleted";
+  // final step when the device has reconnected after the firwmare update has been completed
+
+  progress: number;
+  error: { type: "UpdateFirmwareError"; message?: string };
+}>;
+
+export const initialState: UpdateFirmwareActionState = {
+  step: "preparingUpdate",
+  progress: 0,
+  ...initialSharedActionState,
+};
+
+/**
+ * Device Action used to update the firmware installed on a Ledger device
+ * @param deviceId the transport id of the device to be updated
+ * @returns an Observables that emits the evolution of the state machine of the firmware installation
+ */
+export function updateFirmwareAction({
+  deviceId,
+}: updateFirmwareActionArgs): Observable<UpdateFirmwareActionState> {
+  return getDeviceInfoTask({ deviceId })
+    .pipe(
+      switchMap((event) => {
+        if (event.type !== "data") {
+          return of(event);
+        }
+        const { deviceInfo } = event;
+        return getLatestFirmwareTask({ deviceId, deviceInfo });
+      }),
+      switchMap((event) => {
+        if (event.type !== "data") {
+          return of(event);
+        } else {
+          return updateFirmwareTask({
+            deviceId,
+            updateContext: event.firmwareUpdateContext,
+          });
+        }
+      })
+    )
+    .pipe(
+      scan<
+        | UpdateFirmwareTaskEvent
+        | GetLatestFirmwareTaskErrorEvent
+        | GetDeviceInfoTaskErrorEvent,
+        UpdateFirmwareActionState
+      >((currentState, event) => {
+        switch (event.type) {
+          case "taskError":
+            return {
+              ...initialState,
+              error: {
+                type: "UpdateFirmwareError",
+                error: event.error,
+              },
+            };
+          case "installingOsu":
+          case "flashingMcu":
+          case "flashingBootloader":
+            return {
+              ...currentState,
+              step: event.type,
+              progress: event.progress,
+            };
+          case "allowSecureChannelRequested":
+          case "installOsuDevicePermissionRequested":
+          case "installOsuDevicePermissionGranted":
+          case "installOsuDevicePermissionDenied":
+          case "firmwareUpdateCompleted":
+            return { ...currentState, step: event.type };
+          default:
+            return {
+              ...currentState,
+              ...sharedReducer({
+                event,
+              }),
+            };
+        }
+      }, initialState)
+    );
+}
--- a/libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/flashMcuOrBootloader.ts
+++ b/libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/flashMcuOrBootloader.ts
@@ -1,0 +1,63 @@
+import { Observable } from "rxjs";
+import URL from "url";
+import Transport from "@ledgerhq/hw-transport";
+import type { DeviceInfo, SocketEvent } from "@ledgerhq/types-live";
+import { version as livecommonversion } from "../../../../package.json";
+import { getEnv } from "../../../env";
+import { log } from "@ledgerhq/logs";
+import { createDeviceSocket } from "../../../api/socket";
+import { filter, map } from "rxjs/operators";
+
+export type FlashMcuOrBootloaderCommandRequest = {
+  targetId: DeviceInfo["targetId"];
+  version: string;
+};
+
+type ProgressEvent = {
+  type: "bulk-progress";
+  progress: number;
+  index: number;
+  total: number;
+};
+
+const filterProgressEvent = (e: SocketEvent): e is ProgressEvent =>
+  e.type === "bulk-progress";
+
+export type FlashMcuCommandEvent = {
+  type: "progress";
+  progress: number;
+};
+
+/**
+ * Creates a scriptrunner connection with the /mcu API endpoint of the HSM in order to flash the MCU
+ * or the Bootloader of a device (both use the same endpoint with version different parameters)
+ * @param transport The transport object to contact the device
+ * @param param1 The versions details of the MCU or Bootloader to be installed
+ * @returns An observable that emits the events according to the progression of the firmware installation
+ */
+export function flashMcuOrBootloaderCommand(
+  transport: Transport,
+  { targetId, version }: FlashMcuOrBootloaderCommandRequest
+): Observable<FlashMcuCommandEvent> {
+  log("device-command", "flashMcuOrBootloader", {
+    targetId,
+    version,
+  });
+
+  return createDeviceSocket(transport, {
+    url: URL.format({
+      pathname: `${getEnv("BASE_SOCKET_URL")}/mcu`,
+      query: {
+        targetId,
+        livecommonversion,
+        version,
+      },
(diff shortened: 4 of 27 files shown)


libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/installFirmware.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  | {
+      type: "firmwareInstallPermissionRequested";
+    }
+  | {
+      type: "firmwareInstallPermissionGranted";
+    }
+  | UnresponsiveCmdEvent;
+
+export function installFirmwareCommand(
```
alexandremgo, line 53: We should start documenting what each command (and task) are doing - because just by reading the name, i don't see which one does what between `installFirmware` and `flashMcuOrBootloader`

libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/installFirmware.ts
```diff
@@ -0,0 +1,148 @@
...
+  }).pipe(
+    catchError(remapSocketFirmwareError),
+    filter<SocketEvent, FilteredSocketEvent>((e): e is FilteredSocketEvent => {
+      return (
+        e.type === "bulk-progress" || e.type === "device-permission-requested"
+      );
+    }),
+    map<FilteredSocketEvent, InstallFirmwareCommandEvent>((e) => {
+      if (e.type === "bulk-progress") {
+        return e.index === e.total - 1
```
alexandremgo, line 84: it was probably in the former install firmware logic, but this looks like magic ahah

libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/installFirmware.ts
```diff
@@ -0,0 +1,148 @@
...
+      wording: string;
+    };
+
+export type InstallFirmwareCommandEvent =
+  | {
+      type: "progress";
+      progress: number;
+    }
+  | {
+      type: "allowManagerRequested";
```
alexandremgo, line 43: side note: i think we wanted to change the "allow manager" vocabulary to "allow secure channel" - so `allowSecureChannelRequested`

libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/flashMcuOrBootloader.ts
```diff
@@ -0,0 +1,55 @@
...
+import { log } from "@ledgerhq/logs";
+import { createDeviceSocket } from "../../../api/socket";
+import { filter, map } from "rxjs/operators";
+
+export type FlashMcuOrBootloaderCommandRequest = {
+  targetId: DeviceInfo["targetId"];
+  version: string;
+};
+
+const castProgressEvent = (e: SocketEvent) =>
```
alexandremgo, line 16: i think we can use a TS `type predicates` here to narrow the type at runtime

libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/installFirmware.ts
grsoares21: The biggest difference is the endpoint to which we create the secure channel socket connection.

In installFirmware it's the `/install` one which is also the same that's used for installing apps. In flashMcuOrBootloader it's the `/mcu` one.

There's also some extra handling of different types of events in the `/install` since there are more steps or types of apdu that can come from that endpoint.

We should probably document all the endpoints as well that we have. Which ones require secure connections with the HSM etc

libs/ledger-live-common/src/deviceSDK/commands/firmwareUpdate/installFirmware.ts
grsoares21: I'm adding some comments explaining

apps/cli/src/commands/newGetDeviceInfo.ts
```diff
@@ -0,0 +1,32 @@
+import { getDeviceInfoAction } from "@ledgerhq/live-common/deviceSDK/actions/getDeviceInfo"; 
+import { Observable } from "rxjs";
+import { deviceOpt } from "../scan";
+
+export default {
```
alexandremgo, line 5: this CLI should be deleted ;) (i renamed it `deviceSDKGetDeviceInfo` already)

apps/cli/src/commands-index.ts
```diff
@@ -23,6 +23,7 @@ import estimateMaxSpendable from "./commands/estimateMaxSpendable";
 import exportAccounts from "./commands/exportAccounts";
 import firmwareRepair from "./commands/firmwareRepair";
 import firmwareUpdate from "./commands/firmwareUpdate";
+import firmwareUpdateAction from "./commands/firmwareUpdateAction";
```
alexandremgo, line 26: Maybe we should all name them: `deviceSDK<Something>` to gather all the new CLI using the device SDK (so `deviceSDKFirmwareUpdate`) as long as the device SDK is not mainly used). WDYT ?

libs/ledger-live-common/src/api/Manager.ts
```diff
@@ -571,6 +571,28 @@ const installMcu = (
...
 };
 
+function retrieveMcuVersion(
+  finalFirmware: FinalFirmware
+): Promise<McuVersion | undefined> {
+  return getMcus()
+    .then((mcus) =>
+      mcus.filter((deviceInfo) => {
+        const provider = getProviderId(deviceInfo);
+        return (mcu) => mcu.providers.includes(provider);
```
alexandremgo, line 581: mcu has an `any` type - but `getMcus` returns a `Promise<any>` because it's a request .. In the future we should maybe implement a type system for the manager

libs/ledger-live-common/src/api/Manager.ts
```diff
@@ -571,6 +571,28 @@ const installMcu = (
...
+      })
+    )
+    .then((mcus) =>
+      mcus.filter((mcu) => mcu.from_bootloader_version !== "none")
+    )
+    .then((mcus) =>
+      findBestMCU(
+        finalFirmware.mcu_versions
+          .map((id) => mcus.find((mcu) => mcu.id === id))
```
alexandremgo, line 590: what happens if `mcus` is undefined/null (i don't think it's possible, `filter` returns an empty array if no mcu is found) - but what about `getMcus`, if it fails or returns `undefined` ?

Probably not the correct PR to do this - but i don't like this Manager 🙃

libs/ledger-live-common/src/deviceSDK/commands/quitApp.ts
```diff
@@ -0,0 +1,16 @@
+import { Observable } from "rxjs";
+import Transport from "@ledgerhq/hw-transport";
+
+
+export function quitApp(
```
alexandremgo, line 5: Looks like there is some prettier/indent issue ;)

libs/ledger-live-common/src/deviceSDK/commands/quitApp.ts
```diff
@@ -0,0 +1,16 @@
...
+  transport: Transport
+): Observable<void> {
+  return new Observable((observer) => {
```
alexandremgo, line 8: Should it implement the unresponsive timeout ? `transport.setExchangeUnresponsiveTimeout(1000)` etc. like `getVersion` for ex ?

libs/ledger-live-common/src/deviceSDK/tasks/getDeviceInfo.ts
```diff
@@ -40,17 +39,7 @@ function internalGetDeviceInfoTask({
...
-            if (!isDashboardName(name)) {
-              subscriber.next({
-                type: "taskError",
-                error: "DeviceOnDashboardExpected",
-              });
-              return false;
-            }
-            return true;
-          }),
+        quitApp(transport).pipe(
```
alexandremgo, line 42: Smart - i guess we don't need a `getAppAndVersion` after all, if `quitApp` does nothing when the device is already on the dashboard ?

alexandremgo: I'll do some cleaning on `GetDeviceInfoTaskError` when i'll rework the genuine check - because we don't need it anymore, but we want to define a `taskError` maybe ?

libs/ledger-live-common/src/deviceSDK/tasks/getLatestFirmware.ts
```diff
@@ -0,0 +1,70 @@
...
+        switchMap((firmwareUpdateContext) => {
+          if (firmwareUpdateContext) {
+            return of<GetLatestFirmwareTaskEvent>({
+              type: "data",
+              firmwareUpdateContext,
+            });
+          } else {
+            return of<GetLatestFirmwareTaskEvent>({
+              type: "taskError",
+              error: "FirmwareUpToDate",
```
alexandremgo, line 53: Do we want to return a "taskError" for a firmware that is up to date ? Should it be another type of event ?

libs/ledger-live-common/src/deviceSDK/tasks/updateFirmware.ts
```diff
@@ -0,0 +1,283 @@
...
+              type: "error" as const,
+              error: new LockedDeviceError(),
+            });
+          }
+          const { firmwareInfo } = value;
+
+          // TODO: we're repeating the handling of the unresponsive event here... is there a way to make this better
+          // the problem that if we want to change multiple commands that may have unresponsive events
+          // and only continue with the next command if the event is not unresponsive we might have to handle multiple times
+          // the unresponsive event :/
```
alexandremgo, line 66: Yes that the problem of handling this kind of "locked device" (unresponsive from a command) at the command level. But as we discussed, we don't want this unresponsive strategy on every commands (for ex `getAppAndVersion` does not need it) and we want to customized the timeout.

Though maybe we can create a custom `switchMap` and `map` that already handle the `unresponsive` event ? WDYT ? Is it just creating a new abstraction for just 5 lines of code ?

apps/cli/src/commands-index.ts
grsoares21: Done

apps/cli/src/commands/newGetDeviceInfo.ts
grsoares21: Done

libs/ledger-live-common/src/api/Manager.ts
grsoares21: Yeah, this has been directly grabbed from legacy code. So I don't fully know what the types are. It will require a bit of reverse engineering of the endpoint. I'll add a TODO

libs/ledger-live-common/src/api/Manager.ts
grsoares21: getMcus uses some cache mechanism inside libs/coin-framework, I believe it handles the promise failing correctly, because it's nowhere used with a try catch. This whole manager is probably something worth rethinking.

libs/ledger-live-common/src/deviceSDK/commands/quitApp.ts
grsoares21: done

libs/ledger-live-common/src/deviceSDK/tasks/updateFirmware.ts
grsoares21: I'm not completely against it. But I usually prefer to create the abstraction the moment it's actually needed (e.g. when we've created several tasks already that do this).

alexandremgo: yeah i agree - then let's keep it like this (handling manually the event of type `unresponsive`) and see if this pattern is repeated everywhere

libs/ledger-live-common/src/deviceSDK/commands/quitApp.ts
alexandremgo: I documented the command on its behavior with locked devices - what i was proposing is not needed

libs/ledger-live-common/src/api/Manager.ts
alexandremgo: ahah yeah, if to `getMcus` we need something that is located inside `libs/coin-framework` we need to re-work it ;) (but not in this PR)

libs/ledger-live-common/src/deviceSDK/tasks/getDeviceInfo.ts
alexandremgo: done in this PR actually

libs/ledger-live-common/src/deviceSDK/actions/updateFirmware.ts
```diff
@@ -0,0 +1,107 @@
...
+  error: { type: "UpdateFirmwareError"; message?: string };
+}>;
+
+export const initialState: UpdateFirmwareActionState = {
+  step: "preparingUpdate",
+  progress: 0,
+  ...initialSharedActionState,
+};
+
+export function updateFirmwareAction({
```
alexandremgo, line 46: Maybe let's add a "JSDoc" documentation for each action, as they will be provided to our consumers

alexandremgo: Also with a small definition of what each `step` represents, so the consumer knows what to render

libs/ledger-live-common/src/deviceSDK/actions/updateFirmware.ts
```diff
@@ -0,0 +1,107 @@
...
+  deviceId,
+}: updateFirmwareActionArgs): Observable<UpdateFirmwareActionState> {
+  return getDeviceInfoTask({ deviceId })
+    .pipe(
+      filter<GetDeviceInfoTaskEvent, { type: "data"; deviceInfo: DeviceInfo }>(
```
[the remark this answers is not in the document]
alexandremgo: Do we need our `switchMap`, `map` and more that let propagate some events like `{ type: "error", error }` or `{ type: "taskError", error }`

libs/ledger-live-common/src/deviceSDK/actions/updateFirmware.ts
grsoares21: done

libs/ledger-live-common/src/deviceSDK/tasks/getLatestFirmware.ts
grsoares21: I think the error makes sense. In theory the consumer would check independently if there's already an update available for the device and only trigger this if yes.

If we make an analogy to HTTP methods, this would be like a 422 (unprocessable entity): "you're asking me to do something that doesn't make sense, to update the firmware of a device that doesn't have any updates"

That was my reasoning.

libs/ledger-live-common/src/deviceSDK/tasks/core.ts
```diff
@@ -52,3 +74,99 @@ export function sharedLogicTaskWrapper<TaskArgumentsType, TaskEventsType>(
...
+  CommandArgsWithoutTransportType,
+  CommandEventsType
+>({
+  command,
+  refreshTransport,
+  transportRef,
+  allowedErrors,
+}: {
+  command: (
+    args: CommandArgsWithoutTransportType & CommandTransportArgs
```
grsoares21, line 106: can't we put the `withRefreshableTransport` inside here? this way we don't need to pass the refreshTransport and transportRef arguments. This way the task doesn't even have to know about the existence of `withRefreshableTransport`.

Either way these are tightly coupled, using this wrappers **requires** the task to use withRefreshableTransport, so we might as well put it here.

libs/ledger-live-common/src/deviceSDK/tasks/getDeviceInfo.ts
```diff
@@ -1,58 +1,56 @@
...
-            if (!isDashboardName(name)) {
-              subscriber.next({
-                type: "taskError",
-                error: "DeviceOnDashboardExpected",
-              });
-              return false;
-            }
-            return true;
-          }),
+      withRefreshableTransport(deviceId)(({ transportRef, refreshTransport }) =>
```
grsoares21, line 42: concerning my other command about moving withRefreshableTransport to inside the wrapper. 
That would mean that we would move quitApp out of here and rather after the getAppAndVersion in the orikginal withDevice. Which makes sense, it's not quitApp that we're retrying, it's the getVersion command, because we know quitapp can cause a disconnect.

libs/ledger-live-common/src/deviceSDK/tasks/core.ts
alexandremgo: I agree that  `retryOnErrorsCommandWrapper` is coupled to `withRefreshableTransport`

But i don't see how putting the `withRefreshableTransport` inside the wrapper would work ? The idea of having the `withRefreshableTransport` is to do the "logic that was done inside withDevice" only once. And then we refresh the transport if there is an issue

By putting the `withRefreshableTransport` you are going to call it on every wrapped command no ? And you are going to call another `withDevice` for all the other commands no ? And how are you going to keep the transport up-to-date if you both use a `withDevice` and a `withRefreshableTransport` ?

PR #2323, opened 2023-01-13
support/b2cqa 1187 update swap tests

ggilchrist-ledger:
### 📝 Description

Resurrect Swap tests to work with new swap layout and providers with the following test cases (all mocked data and device actions):
- Changelly Floating rate max swap with change of target account

For future:
- Add test to go to DEX and verify the page loads
- Drop a lot of the mocked responses in the frontend code and use Playwrights mocked API responses depending on the provider

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/87659842/212331415-f24eef7c-dd22-4686-9e5c-0b1e7a6d38cb.mov

### 🚀 Expectations to reach


libs/ledger-live-common/src/exchange/swap/types.ts
```diff
@@ -148,7 +148,9 @@ export type GetExchangeRates = (
   providers?: AvailableProviderV3[],
   includeDEX?: boolean
 ) => Promise<ExchangeRate[]>;
-export type GetProviders = () => Promise<AvailableProvider[]>;
+export type GetProviders = () => Promise<
+  AvailableProvider[] | ProvidersResponseV4
```
ggilchrist-ledger, line 152: Will remove this stuff but I have questions about how it functions

apps/ledger-live-desktop/tests/specs/services/swap.spec.ts
```diff
@@ -2,57 +2,68 @@ import test from "../../fixtures/common";
 import { expect } from "@playwright/test";
 import { SwapPage } from "../../models/SwapPage";
 import { DeviceAction } from "../../models/DeviceAction";
-import { Layout } from "../../models/Layout";
 import { Drawer } from "tests/models/Drawer";
+import { Layout } from "../../models/Layout";
 
 test.use({ userdata: "1AccountBTC1AccountETH" });
 
+// process.env.PWDEBUG = "1";
```
cjordan-ledger, line 10: Delete before merge?

cjordan-ledger (commented):
Looks great, left one tiny comment

libs/ledger-live-common/src/exchange/swap/mock.ts
```diff
@@ -128,13 +149,15 @@ export const mockInitSwap = (
     },
   });
 };
+
+// Need to understand how and why this gets used
```
cjordan-ledger, line 153: Is this comment still needed?

cjordan-ledger (approved):
Nicely done 🎉

libs/ledger-live-common/src/exchange/swap/mock.ts
```diff
@@ -153,6 +176,56 @@ export const mockGetProviders: GetProviders = async () => {
     },
   ];
 };
+
+// Providers using V4 schema. For some reason it doesn't work though (still requires V3)
```
sarneijim, line 180: I think this schema is v5, not v4.
We do not support v5 yet.
Why do you have this schema? Have you seen it in any place?

sarneijim, line 180:
```suggestion
// Providers using V5 schema. For some reason it doesn't work though (still requires V4)
```

ggilchrist-ledger: I can't quite remember but I think it's what prod is using (V4). When I tried this schema in the mocks, it didn't work, even after updating the types. You can see in `getProviders.ts` the code shows we are using V4 in prod: 

```js
const getProviders: GetProviders = async () => {
  if (getEnv("MOCK")) return mockGetProviders();

  const res = await network({
    method: "GET",
    url: `${getSwapAPIBaseURL()}/providers`,
    params: { whitelist: getAvailableProviders() },
    paramsSerializer: (params) =>
      qs.stringify(params, { arrayFormat: "comma" }),
  });

  const responseV4 = res.data as ProvidersResponseV4;
  if (!responseV4.providers || !Object.keys(responseV4.providers).length) {
    throw new SwapNoAvailableProviders();
  }
  return Object.entries(responseV4.providers).flatMap(([provider, groups]) => ({
    provider: provider,
    pairs: groups.flatMap((group) =>
      group.methods.flatMap((tradeMethod) =>
        Object.entries(group.pairs).flatMap(([from, toArray]) =>
          (toArray as number[]).map((to) => ({
            from: responseV4.currencies[from],
            to: responseV4.currencies[to.toString()],
            tradeMethod,
          }))
        )
      )
    ),
  }));
};
```

ggilchrist-ledger: Long story short I'm not sure what's going on, but I think we need to understand the different schemas, what's getting used, and then update the mocks. I think another improvement we could make it try to use mocked API responses where appropriate with Playwright HTTP mocking, rather than using `MOCK=1`. To me that's more of an E2E test, and also it reduces the amount of `MOCK` logic in the code.

PR #2342, opened 2023-01-17
Performance: Add transport.exchangeBulk API and extend IPCTransport

gre:
### 📝 Description

- This implements a `exchangeBulk` method on Transport, that is implemented by default for all transports.
- We then can use it in `socket` logic in order to run the bulk of APDU when install apps, firmwares,...
- We overrides the "exchangeBulk" on the the IPC Transport to benefit the gain of performance.

Performance analysis:
- When Experimental Renderer Execution is ON => On a build, this was tested to optimise from **2 minutes** to **25 seconds** the installation of Polkadot app.
- When it is OFF (aka production) => On a build, I was able to see that the current production approach (commands & all is in internal) is relatively the same, but actually a bit slower (**27 seconds** on that try). The reason it could be slower is that Internal Process is today doing MORE than just dealing with devices (it's dealing with coins & other commands,.. NB that i didn't even had an account imported)

### 📸 Demo in video

In order to fully document this, I have made 3 videos:

- Context on why we need to do migrate to renderer: https://youtu.be/WBunb5j2zTM
- DEMO of the performance: https://youtu.be/Ww8NN_GF8Q8
- CODE explanation: https://youtu.be/oT78r1nVLlQ

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5222 

### ✅ Checklist

- [x] **Test coverage** it's manually tested **and it's still behind an experimental flag** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/api/socket.ts
+++ b/libs/ledger-live-common/src/api/socket.ts
@@ -35,6 +35,7 @@ export const createDeviceSocket = (
   new Observable((o) => {
     let deviceError: Error | null = null; // error originating from device (connection/response/rejection...)
     let unsubscribed = false; // subscriber wants to stops everything
+    let bulkSubscription: null | { unsubscribe: () => void } = null; // subscription to the bulk observable
     let correctlyFinished = false; // the socket logic reach a normal termination
     let inBulkMode = false; // we have an array of apdus to exchange, without the need of more WS messages.
     let allowSecureChannelTimeout: NodeJS.Timeout | null = null; // allows to delay/cancel the user confirmation event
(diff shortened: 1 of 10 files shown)


libs/ledger-live-common/src/api/socket.ts
```diff
@@ -184,16 +185,25 @@ export const createDeviceSocket = (
...
+              notify(0);
+              // we also use a subscription to be able to cancel the bulk if the user unsubscribes
+              bulkSubscription = transport.exchangeBulk(
+                data.map((d) => Buffer.from(d, "hex")),
+                {
+                  next: () => {
+                    notify(++i);
+                  },
+                  error: (e) => reject(e),
+                  complete: () => resolve(null),
```
Justkant, line 200: Should we reset `bulkSubscription` on `complete` to avoid an unnecessary call to `unsubscribe` ?

gre: we can, but it should always be a "no op". it's part of the remaining thing to do I think anyway because right now, unsubscribe isn't propagated all the way down the ipc

libs/ledgerjs/packages/hw-transport/src/Transport.ts
```diff
@@ -120,6 +120,40 @@ export default class Transport {
     throw new Error("exchange not implemented");
   }
 
+  /**
+   * Send apdus in batch to the device using a low level API.
+   * The default implementation is to call exchange for each apdu.
+   * @param {Array<Buffer>} apdus - array of apdus to send.
+   * @param Observer<Buffer> observer - an observer that will receive the response of each apdu.
```
alexandremgo, line 127:
```suggestion
   * @param {Observer<Buffer>} observer - an observer that will receive the response of each apdu.
```

If we want it to be correctly formatted ;)

alexandremgo (commented):
Just 2 small things, but otherwise good 👌

alexandremgo (approved):

PR #2328, opened 2023-01-13
[LIVE-3511] Feature: Add tokens support to EVM family

lambertkevin:
### 📝 Description

This PR enables the EVM family to support tokens (ERC20) send/receive.

### ❓ Context

- **Impacted projects**: `live-common`, `ledger-live-desktop`, `ledger-live-mobile`, `hw-app-eth` 
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-3511 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/44363395/215188891-cde374ff-52f7-4c7e-930e-272f937f7afe.mov

https://user-images.githubusercontent.com/44363395/215190826-f1b0ae9a-04b2-4435-9a63-d8a60ed1906e.mov

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

### 🧪 Coverage
Screenshot 2023-01-24 at 23 22 24

PR #2344, opened 2023-01-17
Hydrate all currencies that was loaded in cache

gre:
### 📝 Description

worked with @lambertkevin to solve ongoing bug in LLD.

see https://ledgerhq.atlassian.net/wiki/spaces/~5570585546a11e40f64481b3c5af5e8f0034c0/pages/4025254182/hydrateCurrency+should+be+called+for+each+data+stored+in+localStorage+to+truly+hydrate+everything+even+currency+that+weren+t+imported+as+account+yet

in order to make them available at load of the app (synchronously) BEFORE the accounts are loaded (eg tokens are needed to be loaded in order to not disappear through the serialisation)

### ❓ Context

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/init.jsx
+++ b/apps/ledger-live-desktop/src/renderer/init.jsx
@@ -133,13 +133,14 @@ async function init() {
 
   const isMainWindow = remote.getCurrentWindow().name === "MainWindow";
 
-  // TODO hydrate the store with the bridge/cache
-  listCachedCurrencyIds().then(ids => {
-    ids.forEach(id => {
-      hydrateCurrency(id);
-    });
-  });
-
+  // hydrate the store with the bridge/cache
+  await Promise.allSettled(
+    listCachedCurrencyIds()
+      .map(id => {
+        const currency = findCryptoCurrencyById(id);
+        return currency ? hydrateCurrency(currency) : null;
+      }),
+  );
   let accounts = await getKey("app", "accounts", []);
   if (accounts) {
     accounts = implicitMigration(accounts);
--- a/apps/ledger-live-desktop/src/renderer/init.jsx
+++ b/apps/ledger-live-desktop/src/renderer/init.jsx
@@ -46,6 +46,7 @@ import {
 import ReactRoot from "~/renderer/ReactRoot";
 import AppError from "~/renderer/AppError";
 import { expectOperatingSystemSupportStatus } from "~/support/os";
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/init.jsx
```diff
@@ -132,6 +133,13 @@ async function init() {
 
   const isMainWindow = remote.getCurrentWindow().name === "MainWindow";
 
+  // TODO hydrate the store with the bridge/cache
+  listCachedCurrencyIds().then(ids => {
```
Justkant, line 137: From the code above it looks like `listCachedCurrencyIds` returns the array of ids and not a promise, right ?

apps/ledger-live-desktop/src/renderer/init.jsx
```diff
@@ -132,6 +133,13 @@ async function init() {
...
+    ids.forEach(id => {
+      hydrateCurrency(id);
+    });
+  });
+
```
lambertkevin, line 142:
```suggestion
  // hydrate the store with the bridge/cache
  await Promise.allSettled(
    listCachedCurrencyIds()
      .map(id => {
        const currency = findCryptoCurrencyById(id);
        return currency ? hydrateCurrency(currency) : null;
      }),
  );
```

lambertkevin: `listCachedCurrencyIds()` is not async, but `hydrateCurrency` is, so here is the working version. 
Also missing an import of `findCryptoCurrencyById` for my suggestion to work

lambertkevin (changes requested):
Not working as is unfortunately, but with this suggestion, problem is now solved! Thanks 🙏

lambertkevin:
Added here unless with merge this one before, with LLD & LLM support: 572df86852f0f8c909c46136bd4706fbe0a6218d 👌

apps/ledger-live-desktop/src/renderer/init.jsx
gre: ooooh

gre:
@lambertkevin yeah it will be find to land either of the PR. this is a bugfix so it may be good to document a changelog too

gre:
this work is going to get included as part of https://github.com/LedgerHQ/ledger-live/pull/2328

PR #1370, opened 2022-09-26
ETH/BTC explorers v4 + removal of json-bignumber

hedi-edelbloute:
### 📝 Description

- Removal of JSONBigNumber
- Implementation of ethereum/btc explorers v4
- CDN explorer config overrides removal
- Removal of v2/v3 btc explorers

### ❓ Context

- **Impacted projects**: `ledger-live-common` 
- **Linked resource(s)**:
  - https://ledgerhq.atlassian.net/browse/LIVE-4179
  - https://ledgerhq.atlassian.net/browse/LIVE-2248` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #2363, opened 2023-01-19
WIP exploration to replace axios by wretch

gre:
## Conclusion of this exploration

- `fetch` is not a lot "more performant" than XHR. a tiny bit. most importantly it's not worse.
- axios or wretch are two legit framework. but in light of node.js soon supporting fetch, it would make sense to use a lighter solution at some point.
- in any case, `bitcoin` networking complexity need to be simplified (this PR was making one and managed to make bitcoin family fully function with it. there is no need to have a "local" axios client nor to use https and a pool system as long as we also take the "promiseAllBatched" approach to not spam of creating parallel queries in the first place.
- `network.ts` simplification need to happen (logs, error management, retry strategies,..) and we need to precise who's responsable on providing networking, it's not consistent today as some coin decided to make their own networking / some don't. the first solution give more freedom to coins with the risk of having heavy JS bundle...

### regarding performance

Here is a very quick comparison of what was tested here and the conclusion are more interesting regarding the "renderer vs internal" execution rather than the networking.

| scenario                                                               | total time to sync all accounts (avg) | Chrome CPU time (how much seconds lost computing on renderer) |
|------------------------------------------------------------------------|---------------------------------------|---------------------------------------------------------------|
| production conditions (running coins in internal with axios)           | 500 seconds                           | 5s                                                            |
| experimental conditions (running coins in renderer with axios)         | 200 seconds                           | 17s                                                           |
| this branch conditions (running coins in renderer with wretch & fetch) | 200 seconds                           | 15s                                                           |

Here we can observe running coin sync on renderer side is overall FASTER.

The performance on running coins in renderer side is slower because we do more calculation on the renderer side, however there are not much frame drops so it's not perceived by user. However, the boot of the Live with accounts in case of running in internal is very bad, it freeze for 1-2 seconds when the renderer send a BIG object to internal (the "preload" of ethereum notably have tons of logs of ERC20 data). It's hard to put numbers on, but my computer was more intense when running internal (fan was heating), and indeed I didn't observe how much time we lose on CPU on INTERNAL side, but it's either the same or more than renderer. we likely lose time (that i didn't measure) on serialisation and sending objects around.

**to me, the conclusion is that moving to renderer is already a good performance benefits in current's condition**. We will have work of @hedi-edelbloute landing soon via #1370 which will increase even more the CPU situation here. We will have to run further audits when it happens, to see more accurately which parts remain to optimize but current profiles show it's the networking itself and i'm not sure we can do more but **reducing our overall networking usage by reworking coins**.

---

(rest of the notes)

### 📝 Description

Reasons why we would replace axios.

- it's a heavy framework. ~~we suspect bad performance relates to it~~ perf were proven not worse than `fetch`
- it stills uses XMLHttpRequest even though modern web browsers now have "fetch" available. wretch is a light wrapper on top of fetch.
- the bottleneck of performance in rework "dismantling internal process" shows the biggest cost of performance is lost in networking stack, notably the .send() method of XHR. this PR is to study if fetch() would be any better at this job.

Problem at replacing it.

- it's used quite heavily at some places, especially in wallet-btc 🤯 
  - notably there is a "Pool" system, https usage, .. all of these are useless for incoming work of running in renderer.
  - this pool & https usage was needed to throttle the queries, but instance we should just batch them in the first place, see https://github.com/LedgerHQ/ledger-live/pull/2257
- There are weird parts that DON'T use network() but directly will make "axios instances" which makes this transition even more painful..
- the previous network is used EVERYWHERE and depends on axios syntax, I tried to mimic the interface we had but it wasn't clearly defined. `wretch` have a strongly typed interface that will secure this in future.
- [ – idk yet what's the best pattern to offer the overridability of things like headers , or any global env that can still be changed dynamically. can wretch have some "dynamic resolution" of some config? example: nb of retries, X-Ledger headers

Other problem to fix

- the User-Agent isn't possible to set for web. So in context of transitioning to chrome/web, we have to replace the header. – i'll try to solve this in a separate PR too because we'll need this before this bigger work.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/init.jsx
+++ b/apps/ledger-live-desktop/src/renderer/init.jsx
@@ -49,7 +49,7 @@ import { expectOperatingSystemSupportStatus } from "~/support/os";
 
 logger.add(loggerInstance);
 
-if (process.env.DEV_TOOLS) {
+if (process.env.VERBOSE) {
   enableDebugLogger();
 }
 
--- a/libs/ledger-live-common/package.json
+++ b/libs/ledger-live-common/package.json
@@ -228,6 +228,7 @@
     "lru-cache": "5.1.1",
     "minimatch": "^5.1.0",
     "near-api-js": "^0.44.2",
+    "node-fetch": "^3.3.0",
     "numeral": "^2.0.6",
     "object-hash": "^2.2.0",
     "pako": "^2.0.4",
@@ -255,6 +256,7 @@
     "utility-types": "^3.10.0",
     "varuint-bitcoin": "1.1.2",
     "winston": "^3.4.0",
+    "wretch": "^2.3.2",
     "xstate": "^4.30.2",
     "zcash-bitcore-lib": "^0.13.20-rc3"
   },
--- a/libs/ledger-live-common/src/families/bitcoin/wallet-btc/__tests__/wallet.integration.test.ts
+++ b/libs/ledger-live-common/src/families/bitcoin/wallet-btc/__tests__/wallet.integration.test.ts
@@ -138,6 +138,7 @@ describe("testing wallet", () => {
     expect(Buffer.from(tx, "hex")).toHaveLength(20);
   });
 
+  /*
   it("should throw during sync if there is an error in explorer", async () => {
     const client = account.xpub.explorer.underlyingClient;
     // eslint-disable-next-line no-underscore-dangle
@@ -154,4 +155,5 @@ describe("testing wallet", () => {
     client.get = _get;
     expect(thrown).toEqual(true);
   });
+  */
 });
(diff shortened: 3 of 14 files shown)


gre:
**THIS PR** with `export EXPERIMENTAL_EXECUTION_ON_RENDERER=1` and running 1/14th of mere denis accounts ([methodology](https://ledgerhq.atlassian.net/wiki/spaces/~5570585546a11e40f64481b3c5af5e8f0034c0/pages/4032201626/recording+LLD+Performance+of+account+syncs+in+end2end+way) )

Screenshot 2023-01-24 at 13 38 45
Screenshot 2023-01-24 at 13 38 32

**develop branch** with `EXPERIMENTAL_EXECUTION_ON_RENDERER=1` and disabling enableDebugLogger() (https://github.com/LedgerHQ/ledger-live/pull/2390)

Screenshot 2023-01-24 at 13 51 09
Screenshot 2023-01-24 at 13 51 32

gre:
running on branch again seems to reach similar perf, but we need to dig the insights now

Screenshot 2023-01-24 at 14 00 40
Screenshot 2023-01-24 at 14 00 38

PR #2420, opened 2023-01-26
bugfix: opens modal on account page

cjordan-ledger:
### 📝 Description

It's risky how we are currently looking up a hook at runtime based on a selected account, as react expects a hook to always be called. 

Moving the hook lookup to the account screen mitigates this issue, as the account won't change, therefore the hook that is called won't change.

### ❓ Context

- **Impacted projects**: LLD
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5182

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/screens/stake/index.tsx
```diff
@@ -1,45 +1,19 @@
...
-
-const useStakeFlow = (props: Props) => {
+const useStakeFlow = () => {
   const history = useHistory();
-  const [accountSelection, setAccountSelection] = useState<Props>({});
   const stakeProgramsFeatureFlag = useFeature("stakePrograms");
-  const { params: paramsFlag, enabled: stakeProgramsEnabled } = stakeProgramsFeatureFlag || {};
+  const dispatch = useDispatch();
+  const { params: paramsFlag } = stakeProgramsFeatureFlag || {};
   const { list: listFlag } = paramsFlag || {};
```
sarneijim, line 16:
```suggestion
    const { params: paramsFlag, enabled: stakeProgramsEnabled } = stakeProgramsFeatureFlag || {};
  const { list: listFlag } = (stakeProgramsEnabled && paramsFlag) || {};
```

sarneijim (approved):

apps/ledger-live-desktop/src/renderer/screens/stake/index.tsx
cjordan-ledger: Not sure why we want to do this? This hook now just provides a callback for other buttons to use, and they could always be disabled by this `enabled` field if required.

sarneijim: I am not sure if set a list if the flag is disabled. 
An error in the code could be mixed with an error in the flag configuration.

cjordan-ledger: If the flag is disabled, I think the stake buttons should be disabled, otherwise we will open a drawer and display zero currencies.

PR #2432, opened 2023-01-26
Fix psbtv2 serializeMap function

0xBEEFCAF3:
### 📝 Description

The function `serializeMap` had some clear problems and because of it psbt serialization did not work. 
1) `map.keys` is a function, not a property of the map object
2) Should be using `for of` for a list of keys. Not `for in` 
3) `map.get(k)!;` I think this is trying to do some null check. Regardless I replaced it with an if and threw an error if the value did not exist

### ❓ Context

- **Impacted projects**: `` Any project interfacing with the new ledger 2.1.0 app using the js SDK
- **Linked resource(s)**: `` 

### ✅ Checklist

- ✅ **Test coverage** 
- ✅ **Atomic delivery** 
- ✅ **No breaking changes** 

### 📸 Demo
N/A

### 🚀 Expectations to reach

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203836703498839


libs/ledgerjs/packages/hw-app-btc/src/newops/psbtv2.ts
```diff
@@ -540,8 +540,11 @@ function createKey(buf: Buffer): Key {
   return new Key(buf.readUInt8(0), buf.slice(1));
 }
 function serializeMap(buf: BufferWriter, map: Map<string, Buffer>) {
-  for (const k in map.keys) {
-    const value = map.get(k)!;
+  for (const k of map.keys()) {
+    const value = map.get(k);
+    if (!value) {
+      throw new Error(`Failed to serialize map. Missing value for key ${k}`);
+    }
```
bigspider, line 547: Perhaps we can keep the code more compact:

```typescript
  for (const key of map.keys()) {
    const value = map.get(key)!;
```

The `!` only serves the purpose of keeping the Typescript compiler happy, as we know that it's not `Null`.

0xBEEFCAF3: Good point! I'll rebase and make this change in the next day

0xBEEFCAF3: Fixed!

haammar-ledger (approved):
@0xBEEFCAF3 thanks for your contribution!
Indeed that part of the lib lacks tests at the moment, any chance you could create one to cover that single issue? :pray:

0xBEEFCAF3:
Hi @haammar-ledger  Yes I can grab the same test from  https://github.com/LedgerHQ/app-bitcoin-new/pull/108/files and move it over here.

thanks for the feedback

0xBEEFCAF3:
@haammar-ledger added in [b0b1aab](https://github.com/LedgerHQ/ledger-live/pull/2432/commits/b0b1aab7bede7764c797b7bcff5fc7166ddc0c9d)

PR #2434, opened 2023-01-26
Feat: LIVE-4408 add market stake button

sarneijim:
### 📝 Description

Add stake button to market page

### ❓ Context

- **Impacted projects**: LLD
- **Linked resource(s)**: 
https://ledgerhq.atlassian.net/browse/LIVE-4408

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

image

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/screens/market/MarketRowItem.tsx
```diff
@@ -228,11 +245,23 @@ function MarketRowItem({
...
                   >
                     {t("accounts.contextMenu.swap")}
                   </Button>
                 )}
+                {availableOnStake && (
+                  <Box onClick={e => onStake(e)}>
+                    <Button
+                      data-test-id={`market-${currency?.ticker}-stake-button`}
+                      variant="color"
+                      onClick={stakeFlow({ currencies: [currency?.id] })}
```
cjordan-ledger, line 259: this is effectively calling a hook within a callback - which react recommends you don't do. If `stakeFlow` was called `useStakeFlow` I think this will fail linting with an error like:

```
React Hook "useStakeFlow" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function. (react-hooks/rules-of-hooks)eslint
```

sarneijim: If I use a callback, we will have the "hook" issue... I do not find a better solution.

cjordan-ledger: I think we can make the following work?

```
const startStakeFlow = useStakeFlow();

return <button onClick={() => startStateFlow({currencies: [currency?.id]) />}
```

Also, don't we want to show the account selection drawer when they click on this button? They don't need to select a single currency and then the account.

apps/ledger-live-desktop/src/renderer/families/polkadot/AccountHeaderManageActions.js
```diff
@@ -65,7 +65,7 @@ const AccountHeaderManageActions = ({ account }: Props) => {
 
   return [
     {
-      key: "polkadot",
+      key: "Stake",
```
cjordan-ledger, line 68: just out of interest, what's the reason for this change?

apps/ledger-live-desktop/src/renderer/screens/market/MarketList.tsx
```diff
@@ -219,6 +220,10 @@ const CurrencyRow = memo(function CurrencyRowItem({
   const isStarred = currency && starredMarketCoins.includes(currency.id);
   const availableOnBuy = currency && onRampAvailableTickers.includes(currency.ticker.toUpperCase());
   const availableOnSwap = internalCurrency && swapAvailableIds.includes(internalCurrency.id);
+  const stakeProgramsFeatureFlag = useFeature("stakePrograms");
+  const { params: paramsFlag, enabled: stakeProgramsEnabled } = stakeProgramsFeatureFlag || {};
```
cjordan-ledger, line 224: Is typescript passing this code? `listFlag` could be undefined but we're not checking for it on line 226. I think we'll have better type safety without the destructuring, so something like this instead:

```
const listFlag = stakeProgramsFeatureFlag?.params?.list ?? []
const stakeProgramsEnabled = stakeProgramsFeatureFlag?.enabled
```

cjordan-ledger (commented):
Looks good, couple of minor comments

apps/ledger-live-desktop/src/renderer/families/polkadot/AccountHeaderManageActions.js
cjordan-ledger: never mind - i can see this is added as we use this to find the right stakeAction.onClick to call

cjordan-ledger (approved):

brunospolidorio (approved):

PR #2435, opened 2023-01-26
Support - LLM performance improvements - LIVE-4965

ofreyssinet-ledger:
### 📝 Description

In the context of an audit of the performance in LLM, I did some profiling (mainly React profiling with the dev tools).

- This PR addresses one of the main issues observed: **high frequency of store updates causing re-renders of entire screens**.
  - If a user has N accounts, when the app syncs these N accounts, it results in N successive updates of the `store.getState().accounts.active` part of the store, they happen at a high frequency (in the order of 1 per second). **This frequency of updates was not directly addressed here, but ...**
  - How it impacts the UI performance:
  In many components of the app we have `useSelector(accountsSelector)` that points directly to that part of the store, so each of these updates in the store triggers a re-render of all these components. We are talking about **entire screens that get fully re-rendered very frequently** with a cost of ~200ms (in debug) per render, that's a lot of frames dropped.
  - How it is addressed:
  **In a lot of cases these components only consume data that is _derived from_ `accounts.active`**, not the entire data. For instance in a lot of cases we just count the number of accounts or whether some accounts have a positive balance.
  **The fix is then to create selectors that directly derive the data needed**, thus restricting re-renders to updates that cause this derived data to change.

- Another directly related issue is the **expensive computation of graph data**.
  - The components rendering the graph actually need the full data coming from `accounts.active` (and other frequently sources of data such as the countervalues state), so there, optimising the selectors is not be the right strategy (also it's more difficult because the data of countervalues comes from a context, not from the store).
  - It impacts the UI performance because each of these computation happens synchronously, can take ~50ms (in debug) and thus block the JS thread for several frames, making the app less responsive.
  - The solution here is to throttle the inputs of that computation so that a computation happens at most every 5000ms instead of at every single render.  For that purpose several `useThrottle*` hooks were implemented in ledger-live-common and allow throttling a value, an array of values, or the execution of a function along with its dependencies.

- Another important thing that was implemented here is to **break down big components into smaller components** so that re-renders only happen in parts that actually needs this re-render. The general idea is to eliminate as much state as possible from big components and pull it down when possible (when that state isn't shared between the smaller parts). This is especially possible on the main Portfolio screen and the WalletCentricAsset screen.

#### Scope of changes in UI

- Portfolio main screen (including ReadOnly)
- Asset main screen (including ReadOnly)
- NFT gallery
- Custom image NFT gallery selector
- Main list of operations
- Settings main screen (Accounts section)
- Settings debug: Toggle Service status incident section

#### Important points to fully understand the changes in this PR:

- [How `useSelector` works and when does it trigger re-renders](https://react-redux.js.org/api/hooks#useselector): 
> Screenshot 2023-02-03 at 10 28 04

- [How to create memoized selectors with `reselect` to avoid heavy computations and creation of new refs at every store update, in order to avoid unnecessary re-renders of components](https://github.com/reduxjs/reselect#createselectorinputselectors--inputselectors-resultfunc-selectoroptions).

- [How React.memo works](https://reactjs.org/docs/react-api.html#reactmemo)
> Screenshot 2023-02-03 at 10 30 09

### ❓ Context

- **Impacted projects**: `mobile` `common` 
- **Linked resource(s)**: [LIVE-4965] [LIVE-3986] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Some results of a simple test where I used [react-native-flipper-performance-monitor](https://github.com/bamlab/react-native-flipper-performance-monitor) to get a score on a simple test scenario: opening an instance of the app with many accounts and measuring the FPS for 20 seconds:

Screenshot 2023-02-03 at 18 13 25

### 🚀 Expectations to reach

[LIVE-4964]: https://ledgerhq.atlassian.net/browse/LIVE-4964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-3986]: https://ledgerhq.atlassian.net/browse/LIVE-3986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-4965]: https://ledgerhq.atlassian.net/browse/LIVE-4965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/components/TransactionsPendingConfirmationWarning.tsx
```diff
@@ -53,9 +47,27 @@ const TransactionsPendingConfirmationWarning = ({
...
 };
 
+export const TransactionsPendingConfirmationWarningAllAccounts = () => {
+  const areSomeAccountsBalanceUnconfirmed = useSelector(
+    areSomeAccountsBalanceUnconfirmedSelector,
+  );
+  if (!areSomeAccountsBalanceUnconfirmed) return null;
+  return <TransactionsPendingConfirmationWarningContent />;
+};
+
```
ofreyssinet-ledger, line 70: Here we split this component in 2 for 2 types of consumer:
- one that wants to use all accounts and thus needs a component that auto updates only when these accounts have/don't have some transactions pending, thus it's optimized to only re-render when that changes because it uses a selector made for that.
- one for a unique account

apps/ledger-live-mobile/src/logic/ratings.tsx
```diff
@@ -80,11 +79,9 @@ const useRatings = () => {
   const ratingsOldRoute = useSelector(ratingsCurrentRouteNameSelector);
   const ratingsHappyMoment = useSelector(ratingsHappyMomentSelector);
   const ratingsDataOfUser = useSelector(ratingsDataOfUserSelector);
-  const accounts: Account[] = useSelector(accountsSelector);
```
ofreyssinet-ledger, line 83: This useSelector will cause a re-render of the component integrating this hook every single time the `accounts` object changes in the store.
This can be optimised as this hook only needs to know the amount of accounts with a positive balance.

apps/ledger-live-mobile/src/logic/ratings.tsx
```diff
@@ -80,11 +79,9 @@ const useRatings = () => {
...
 
-  const accountsWithAmountCount = useMemo(
-    () => accounts.filter(account => account.balance?.gt(0)).length,
-    [accounts],
+  const accountsWithAmountCount = useSelector(
+    accountsWithPositiveBalanceCountSelector,
```
ofreyssinet-ledger, line 84: With this selector we now have only a re-render (caused by useSelector) every time the _number_ of accounts with a positive balance changes.

apps/ledger-live-mobile/src/analytics/index.ts
```diff
@@ -1,7 +1,5 @@
-// eslint-disable-next-line import/no-cycle
 import Track from "./Track";
-// eslint-disable-next-line import/no-cycle
 import TrackScreen from "./TrackScreen";
-// eslint-disable-next-line import/no-cycle
+
```
ofreyssinet-ledger, line 3: Unrelated to the goal of this PR but I managed to get rid of the require cycle 🧹

apps/ledger-live-mobile/src/components/Carousel/Slide.tsx
```diff
@@ -1,40 +1,17 @@
...
 import { useTranslation } from "react-i18next";
 import styled from "styled-components/native";
-import { NavigationProp, useNavigation } from "@react-navigation/native";
-
-// eslint-disable-next-line import/no-cycle
+import { useNavigation } from "@react-navigation/native";
 import Touchable from "../Touchable";
-// eslint-disable-next-line import/no-cycle
 import { track } from "../../analytics";
+import { SlideProps } from "./types";
```
ofreyssinet-ledger, line 9: all changes in the `Carousel` directory are unrelated to the general goal of this PR but there was a require cycle here that was itching me 🧹

apps/ledger-live-mobile/src/components/Carousel/defaultSlides.ts
```diff
@@ -0,0 +1,14 @@
...
+  SlideName.Swap,
+  SlideName.Lido,
+  SlideName.buyCrypto,
+  SlideName.Market,
+  SlideName.FamilyPack,
+];
+
+export const defaultSlidesVisibilityState = Object.fromEntries(
+  defaultSlides.map(slideName => [slideName, true]),
+);
```
ofreyssinet-ledger, line 14: all changes in the `Carousel` directory are unrelated to the general goal of this PR but there was a require cycle here that was itching me 🧹

apps/ledger-live-mobile/src/components/Carousel/Slide.tsx
```diff
@@ -1,109 +0,0 @@
-import React, { useCallback } from "react";
```
ofreyssinet-ledger, line 1: all of this was completely unused

apps/ledger-live-mobile/src/components/Carousel/shared.tsx
```diff
@@ -1,256 +0,0 @@
-import React from "react";
```
ofreyssinet-ledger, line 1: all of this was completely unused

apps/ledger-live-mobile/src/reducers/accounts.ts
```diff
@@ -4,6 +4,7 @@ import {
   createSelector,
```
ofreyssinet-ledger, line 4: In this file we implemented a bunch of memoized selectors (made with reselect) that return more granular data than just the basic `accountsSelector`.
It's the base for most of the optimisations made in this PR, as this part of the store is changing at a high frequency for a user with many accounts, so having selectors targeting more specific things is vital to having a good performance in some components by avoiding useless re-renders of these components.

apps/ledger-live-mobile/src/reducers/settings.ts
```diff
@@ -760,20 +748,6 @@ export const dismissedBannersSelector = (state: State) =>
   state.settings.dismissedBanners;
 export const hasAvailableUpdateSelector = (state: State) =>
   state.settings.hasAvailableUpdate;
-export const carouselVisibilitySelector = (state: State) => {
```
ofreyssinet-ledger, line 763: legacy unused code

apps/ledger-live-mobile/src/reducers/settings.ts
```diff
@@ -124,9 +121,6 @@ export const INITIAL_STATE: SettingsState = {
     size: 0,
     hash: "",
   },
-  carouselVisibility: Object.fromEntries(
-    SLIDES.map(slide => [slide.name, true]),
-  ),
```
ofreyssinet-ledger, line 129: legacy unused code

apps/ledger-live-mobile/src/reducers/settings.ts
```diff
@@ -399,12 +393,6 @@ const handlers: ReducerMap<SettingsState, SettingsPayload> = {
     osTheme: (action as Action<SettingsSetOsThemePayload>).payload.osTheme,
   }),
 
-  [SettingsActionTypes.SETTINGS_SET_CAROUSEL_VISIBILITY]: (state, action) => ({
-    ...state,
-    carouselVisibility: (action as Action<SettingsSetCarouselVisibilityPayload>)
-      .payload.carouselVisibility,
-  }),
-
```
ofreyssinet-ledger, line 407: legacy unused code

apps/ledger-live-mobile/src/reducers/types.ts
```diff
@@ -191,7 +191,6 @@ export type SettingsState = {
   hasAvailableUpdate: boolean;
   theme: Theme;
   osTheme: string | null | undefined;
-  carouselVisibility: number | Record<string, boolean>;
```
ofreyssinet-ledger, line 194: legacy unused code

apps/ledger-live-mobile/src/actions/settings.ts
```diff
@@ -266,14 +265,6 @@ export const dismissBanner = (bannerId: string) =>
...
   });
 
-const setCarouselVisibilityAction =
-  createAction<SettingsSetCarouselVisibilityPayload>(
-    SettingsActionTypes.SETTINGS_SET_CAROUSEL_VISIBILITY,
-  );
-export const setCarouselVisibility = (carouselVisibility: {
-  [key: string]: boolean;
-}) => setCarouselVisibilityAction({ carouselVisibility });
-
```
ofreyssinet-ledger, line 276: legacy unused code

apps/ledger-live-mobile/src/actions/types.ts
```diff
@@ -362,10 +362,6 @@ export type SettingsSetAvailableUpdatePayload = Pick<
 >;
 export type SettingsSetThemePayload = Pick<SettingsState, "theme">;
 export type SettingsSetOsThemePayload = Pick<SettingsState, "osTheme">;
-export type SettingsSetCarouselVisibilityPayload = Pick<
-  SettingsState,
-  "carouselVisibility"
->;
```
ofreyssinet-ledger, line 368: legacy unused code

apps/ledger-live-mobile/src/components/SectionHeader.tsx
```diff
@@ -1,12 +1,11 @@
 import React from "react";
 import styled from "styled-components/native";
 import { Flex, Text } from "@ledgerhq/native-ui";
+import { isEqual } from "lodash";
 import FormatDay from "./FormatDay";
 
 type Props = {
-  section: {
-    day: Date;
-  };
```
ofreyssinet-ledger, line 9: less deep props => easier deep equal

apps/ledger-live-mobile/src/screens/Portfolio/Header.tsx
```diff
@@ -17,14 +17,29 @@ import { track } from "../../analytics";
 import useDynamicContent from "../../dynamicContent/dynamicContent";
```
ofreyssinet-ledger, line 17: In this file we split the `Header` component in smaller components with their internal state (via the hooks they use) to avoid re-renders of the entire component for updates that are specific to one of its parts.

apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
```diff
@@ -96,89 +75,37 @@ function PortfolioScreen({ navigation }: NavigationProps) {
...
+            hideEmptyTokenAccount={hideEmptyTokenAccount}
+            openAddModal={openAddModal}
+          />
+        </Box>
+      ) : (
+        <TrackScreen
+          category="Wallet"
+          accountsLength={0}
+          discreet={discreetMode}
         />
```
ofreyssinet-ledger, line 107: In case `showAssets===true`, the `TrackScreen` that has the `accountsLength` has been moved inside `PortfolioAssets`. This avoids a dependency of this entire screen on useDistribution (which itself depends on `useSelector(accountsSelector)` so it is updated very frequently).

apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
```diff
@@ -96,89 +75,37 @@ function PortfolioScreen({ navigation }: NavigationProps) {
...
+          <PortfolioAssets
+            hideEmptyTokenAccount={hideEmptyTokenAccount}
+            openAddModal={openAddModal}
+          />
+        </Box>
+      ) : (
+        <TrackScreen
+          category="Wallet"
+          accountsLength={0}
+          discreet={discreetMode}
```
ofreyssinet-ledger, line 106: In case `showAssets===true`, the `TrackScreen` that has the `accountsLength` has been moved inside `PortfolioAssets`. This avoids a dependency of this entire screen on `useDistribution` (which itself depends on `useSelector(accountsSelector)` so it is updated very frequently).

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/ImportBridgeStreamData.tsx
```diff
@@ -11,25 +12,52 @@ type Props = {
...
+    <SettingsRow
+      title={title}
+      desc={
+        dataStr
+          ? "Import string detected. Press to import it. Long press to import the string from the clipboard."
+          : "No import string detected. Press to import the string from the clipboard."
+      }
+      iconLeft={<Icons.ImportMedium size={24} color="black" />}
+      onPress={handlePress}
+      onLongPress={importFromClipboard}
```
ofreyssinet-ledger, line 60: changes here are a bit unrelated to the goal of this PR but it was very helpful as the only way to import some data from LLD (with lots of accounts) on an emulator is to:
- either set the data in an env -> on build time if the data string is too long it gets sliced.
- copy paste the env data string to set the env dynamically in the env data debug screen -> it seems like the input field there has a length limit.

So here we get the data directly from the clipboard and even if it's very long it seems to work fine.

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/ToggleServiceStatus.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <Switch
+        value={incidents.length > 0}
+        onValueChange={() => {
+          toggleMockIncident();
+          updateData();
+        }}
+      />
+    </SettingsRow>
+  );
+}
```
ofreyssinet-ledger, line 24: moved the use of the `useServiceStatus` hook out of the main settings screen so that its update don't trigger a re-render of that whole screen

apps/ledger-live-mobile/src/screens/WalletCentricAsset/index.tsx
```diff
@@ -1,24 +1,19 @@
-import React, { useMemo, useState, useCallback, useEffect } from "react";
```
ofreyssinet-ledger, line 1: In this file we split a big component with a lot of hooks in smaller components that have their internal logic / internal state.
This helps us to prevent updates of an entire screen when only one part of that screen should be updated.

apps/ledger-live-mobile/src/screens/WalletCentricSections/Allocations.tsx
```diff
@@ -59,14 +59,17 @@ const Allocations = () => {
...
+      .filter(asset => {
         return (
           asset.currency.type !== "TokenCurrency" ||
           !blacklistedTokenIds.includes(asset.currency.id)
         );
-      },
-    );
+      })
+      .map(obj => {
+        const { accounts, ...other } = obj;
```
ofreyssinet-ledger, line 70: Return the type actually expected by `RingChart` in its props (`DistributionItem`)

apps/ledger-live-mobile/src/components/FormatDate.tsx
```diff
@@ -1,33 +1,44 @@
...
 
 function FormatDate({
   date,
   withHoursMinutes = false,
 }: Props): JSX.Element | null {
-  const locale = useSelector(localeSelector);
+  const defaultOptions = useSelector(defaultOptionsSelector);
+  const hoursAndMinutesOptions = useSelector(hoursAndMinutesOptionsSelector);
   const jsx =
     date && date.getTime()
```
lvndry, line 38: Just asking for opinion here but shouldn't we instead have date be a mandatory prop
```ts
type Props = {
  date: Date;
  withHoursMinutes?: boolean;
};
```

And only render this component is date exists ?

```ts
date && <FormatDate date={date} />
```
This will avoid this check here but also this would avoid calling `createSelector` twice and then returning null

apps/ledger-live-mobile/src/reducers/accounts.ts
```diff
@@ -391,6 +423,163 @@ export const hasLendEnabledAccountsSelector = createSelector(
...
+  makeHasAccountsWithPositiveBalanceSelector(tokenAccountsSelector);
+
+export const hasTokenAccountsNotBlackListedWithPositiveBalanceSelector =
+  makeHasAccountsWithPositiveBalanceSelector(
+    tokenAccountsNotBlacklistedSelector,
+  );
+
+/**
+ * NON-TOKEN-ACCOUNTS SELECTORS
+ */
```
lvndry, line 523: just a proposal but could we create a `accounts` folder here and split these selectors into smallers files and imports them in `index.ts` ? this file start to get quite huge

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/ImportBridgeStreamData.tsx
```diff
@@ -11,25 +12,52 @@ type Props = {
...
-      desc={dataStr ? "Import string detected" : "No import string detected"}
-      iconLeft={<Icons.ImportMedium size={24} color="black" />}
-      onPress={() => {
-        if (!dataStr) return;
 
-        const data = JSON.parse(
-          Buffer.from(dataStr, "base64").toString("utf8"),
-        );
+  const importString = useCallback(
+    (dataStr: Parameters<typeof Buffer.from>[0]) => {
```
lvndry, line 17: why not just set the type to `string` ?

apps/ledger-live-mobile/src/screens/WalletCentricSections/Allocations.tsx
```diff
@@ -88,7 +91,7 @@ const Allocations = () => {
         ticker: t("common.others"),
         color: colors.neutral.c70,
         type: "CryptoCurrency",
-      },
+      } as CryptoCurrency,
```
lvndry, line 94: no way to give the type `DistributionItem` to `othersAllocations` instead of using `as` here ?

```ts
const othersAllocations: DistributionItem = ...
```

apps/ledger-live-mobile/src/screens/WalletCentricSections/Allocations.tsx
```diff
@@ -121,28 +124,40 @@ const Allocations = () => {
             <Flex>
               {distributionListFormatted.length > 0 ? (
                 <AllocationCaption
-                  assetAllocation={distributionListFormatted[0]}
-                  colors={colors}
+                  currencyTicker={distributionListFormatted[0]?.currency.ticker}
```
lvndry, line 127: can't we just do a map here ?

```ts
{
     distributionListFormatted.map((listItem) => {
         return <AllocationCaption
                  currencyTicker={listItem.currency.ticker}
                  currencyColor={getCurrencyColorEnsureContrast(listItem.currency, colors)}
         />
     })
}
```

lvndry (commented):
It's really cool that you added these comments on the PR, made the changes way easier to understand!

apps/ledger-live-mobile/src/reducers/accounts.ts
ofreyssinet-ledger: True, I would split at least a file with all the reducers and a file with all the selectors, I think I'll do that in a separate PR as this one already has a lot of micro changes to review (and doing that means changing the imports in many many files...)

apps/ledger-live-mobile/src/components/FormatDay.ts
```diff
@@ -2,39 +2,48 @@ import { Component } from "react";
...
 type Props = {
   day: Date;
-  locale: string;
+  dateTimeFormat: Intl.DateTimeFormat;
 };
 
 class FormatDay extends Component<Props> {
-  shouldComponentUpdate({ day: nextDay, locale: nextLocale }: Props) {
-    const { day, locale } = this.props;
+  shouldComponentUpdate({
```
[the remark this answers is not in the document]
ofreyssinet-ledger: Good point I'll do that 👍

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/ImportBridgeStreamData.tsx
ofreyssinet-ledger: It seemed necessary at some point when I was reworking this component but yeah it's not actually 😅 thanks!

apps/ledger-live-mobile/src/components/FormatDay.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: done ✅

apps/ledger-live-mobile/src/screens/WalletCentricSections/Allocations.tsx
ofreyssinet-ledger: fixed

apps/ledger-live-mobile/src/screens/WalletCentricSections/Allocations.tsx
ofreyssinet-ledger: done

apps/ledger-live-mobile/src/components/FormatDate.tsx
lvndry: Nice! Didn't know

mcayuelas-ledger:
LGTM on the part i used to work ! :) 

Nice job 🚀

apps/ledger-live-mobile/src/screens/WalletCentricAsset/AssetRow.tsx
```diff
@@ -53,12 +55,20 @@ const AssetRow = ({
...
+   * render, and it's meant to be rendered in a list.
+   *  */
+  const balance = useMemo(() => BigNumber(asset.amount), [asset.amount]);
+
   return (
     <AssetRowLayout
       onPress={onAssetPress}
       currency={currency}
       currencyUnit={unit}
-      balance={new BigNumber(asset.amount)}
```
nparigi-ledger, line 61: not related to the PR, but I'm thinking about why we convert this number to BigNumber in the first place, as if it's a Number we already lost the BigNumber part, and if not then it's useless to convert :sneezing_face:

ofreyssinet-ledger: Yeah legacy logic I guess 😅

nabil-brn (approved):
Nice, I noticed changes and performance gain on Portfolio & Asset page.
Using tools but also "human feeling", would be nice to warn @alalmi-ledger about changes to ease the regression tests.

ofreyssinet-ledger:
@nabil-brn Thanks! I didn't know QA engineers were capable of manipulating "human feeling" 😄 :trollface:

apps/ledger-live-mobile/src/reducers/accounts.ts
```diff
@@ -169,6 +190,17 @@ export const accountsCountSelector = createSelector(
   accountsSelector,
   acc => acc.length,
 );
+/** Returns a boolean that is true if and only if all accounts are empty */
```
nparigi-ledger, line 193:
```suggestion
/** Returns a boolean that is true if and only there is no accounts */
```

ofreyssinet-ledger: nice catch merci

nparigi-ledger (approved):

apps/ledger-live-mobile/src/components/FormatDate.tsx
Justkant: A simpler solution would have been to use the `useMemo` hook

apps/ledger-live-mobile/src/screens/WalletCentricSections/Allocations.tsx
Justkant: If you want you can do the `filter` and `map` with just one `reduce`

Issue #2458, opened 2023-01-30
[Bug]: Ledger Nano X: `TransportStatusError` thrown if `LedgerEthTransactionResolution` is passed into signTransaction on other network than `homestead` for NFTs


xdzurman:
### Impacted Library name

@ledgerhq/hw-app-eth

### Impacted Library version

6.30.5

### Describe the bug

For any ERC721 or ERC1155 NFTs, if transactionResolution is passed into `signTransaction` (goerli network), an error is returned instantly, completely identical to the error you get if the user explicitly declines the transaction:
```
{
        "name": "TransportStatusError",
        "message": "Ledger device: Condition of use not satisfied (denied by the user?) (0x6985)",
        "stack": "Error\n    at new TransportStatusError (webpack-internal:///../node_modules/@ledgerhq/errors/lib-es/index.js:367:18)\n    at TransportWebHID.eval (webpack-internal:///../node_modules/@ledgerhq/hw-transport/lib-es/Transport.js:131:39)\n    at step (webpack-internal:///../node_modules/@ledgerhq/hw-transport/lib-es/Transport.js:45:23)\n    at Object.eval [as next] (webpack-internal:///../node_modules/@ledgerhq/hw-transport/lib-es/Transport.js:26:53)\n    at fulfilled (webpack-internal:///../node_modules/@ledgerhq/hw-transport/lib-es/Transport.js:17:58)\n    at patchedCallback (<anonymous>:132:45)",
        "statusCode": 27013,
        "statusText": "CONDITIONS_OF_USE_NOT_SATISFIED"
    }
```

The code for getting resolution is 
```
import {ledgerService} from '@ledgerhq/hw-app-eth'
const resolution = await ledgerService.resolveTransaction(
      serializedTransaction,
      {},
      {nft: true},
)

const {v, r, s} = await ledger.signTransaction(
        derivationPath,
        serializedTransaction,
        resolution,
)
```

### Expected behavior

The built-in resolution returns a normal looking response from `https://nft.api.live.ledger.com/v1/ethereum`, but it crashes somewhere down the road. If the resolution is okay, I'd expect the device to correctly show NFT data (clear signing, no blind signing enabled). If it is not valid, I'd expect it not to crash on this random error, but say something along the lines that current network is not supported for clear signing.

lambertkevin:
Hi @xdzurman ! 
Thanks for the report, I'll have a look ASAP 👌

lambertkevin:
[...]

PR #2465, opened 2023-01-30
[LIVE-5317] Feature: hw-app-eth default nft resolution

lambertkevin:
### 📝 Description

Make NFT resolution the new default option in `hw-app-eth` 

### ❓ Context

- **Impacted projects**: `hw-app-eth` 
- **Linked resource(s)**: [LIVE-5317](https://ledgerhq.atlassian.net/browse/LIVE-5317) 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-5317]: https://ledgerhq.atlassian.net/browse/LIVE-5317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/fuzzy-wasps-grin.md
+++ b/.changeset/fuzzy-wasps-grin.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/hw-app-eth": patch
+---
+
+Adding default NFT support for resolution
--- a/libs/ledgerjs/packages/hw-app-eth/src/Eth.ts
+++ b/libs/ledgerjs/packages/hw-app-eth/src/Eth.ts
@@ -205,6 +205,7 @@ export default class Eth {
         .resolveTransaction(rawTxHex, this.loadConfig, {
           externalPlugins: true,
           erc20: true,
+          nft: true,
         })
         .catch((e) => {
           console.warn(


gre:
I imagine we can't merge this without being sure all is working and non breaking on nft:true case? I know that there are failing cases and we may not want to turn this default for all third parties if it's the case.
https://github.com/LedgerHQ/ledger-live/issues/2458 notably, but also past events showed cases where backend were failing on these 🤔

lambertkevin:
> I imagine we can't merge this without being sure all is working and non breaking on nft:true case? I know that there are failing cases and we may not want to turn this default for all third parties if it's the case. #2458 notably, but also past events showed cases where backend were failing on these 🤔 any reason to enable this default on hw-app-eth and not in live-common?

Unfortunately, for now it's the only solution to help the Connect team. As of today, they have completely re-implemented the `hw-app-eth` for the sole reason of using the `nft: true` flag...
In `live-common` it's already the [default now](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/ethereum/modules/send.ts#L205).
Concerning the failing cases, I'm having a look at it, but they're really really rare now (except for the backend issues of last time, but that's impacting us as well so we'll be on the watch for this I guess). So between the very low number of implementations using the latest version of this lib + the rare cases of failures, I don't think it's too risky if it can help Connect as a result.

About the #2458 issue, I'm having a look a it right now to understand what's happening 👍

lambertkevin:
Ok so after discussing with the Firmware team, we can't merge this because of the Ethereum app chain ID check... 
If chainId !== 1, we'll have a failure... 

I'll program a dev to work on a strict mode to ignore resolutions errors and just go to blind sign instead, so we can activate it without it breaking for everybody...

PR #1959, opened 2022-11-24
Feat/live-4625 Fix btc swap

hzheng-ledger:
### 📝 Description

Currently, the ledger-live isn't compatible with bitcoin nano app 2.1.0. So, in this PR,
1.  Remove getBitcoinLikeInfo API call in Ledger Live which is deprecated in new nano app and isn't supported anymore
2. In case of new bitcoin nano app 2.1.0, dependencies change. As a consequence, bitcoin app doesn't depend on any app and altcoin app depend on legacy bitcoin app.
3. Launch the bot and test compatibility.
4. In hw-app-btc package, Btc APIs method signature cleanup and refactoring.

Nb. TODO: The current PR is compatible with old bitcoin nano app and new bitcoin nano app. After the new bitcoin releases, we don't need to support old bitcoin nano app any more and we can do a cleanup at that time to simplify the code

This PR disable tests/Btc.test.ts and tests/Btc.test.ts which are currently broken by the APDU modification due to exchange app and new/old btc app check. This check is no longer necessary after the release of new btc app. We will enable these two tests and remove such check then

### ❓ Context

- **Impacted projects**: `` LLC, LedgerJs
- **Linked resource(s)**:
  - https://ledgerhq.atlassian.net/browse/LIVE-4088
  - https://ledgerhq.atlassian.net/browse/LIVE-3310
  - https://ledgerhq.atlassian.net/browse/LIVE-4625

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [ ] **No breaking changes** :x: hw-app-btc bitcoin APIs change: Btc class (hw-app-btc/src/Btc.ts) constructor signature changes. Rename "signMessageNew" to "signMessage". Rename "createPaymentTransactionNew" to "createPaymentTransaction". (refer to hw-app-btc/src/Btc.ts) It would impact external app that use such APIs

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #2467, opened 2023-01-30
bugfix/LIVE-5293 Dependency resolution for newly standalone currency apps

juan-cortes:
### 📝 Description

Starting on version `2.1.0` some of our apps that used to depend on Bitcoin have become independent. However, the current way of resolving dependencies is based on the currency family and does not support this. A first approach was introduced on [feat/LIVE-4625](https://github.com/LedgerHQ/ledger-live/pull/1959) but it was flawed in the sense that the presence of the Bitcoin Legacy app does not necessarily mean that the clone app is independent.

This PR introduces a check based on the version and drops the presence check, as well as adding some tests to cover that functionality.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-5293` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo, only logic.

### 🚀 Expectations to reach
- We should be able to install and run apps that are Bitcoin clones on versions equal or higher than 2.1.0 without having Bitcoin installed. Uninstalling Bitcoin shouldn't impact these apps if installed. 
- Bitcoin clones on versions lower than 2.1.0 should still require Bitcoin app being installed as they've always done.

### DIFF
--- a/libs/ledger-live-common/src/apps/polyfill.ts
+++ b/libs/ledger-live-common/src/apps/polyfill.ts
@@ -68,6 +68,9 @@ export function declareDep(name: string, dep: string): void {
   ["Yearn", "Ethereum"],
 ].forEach(([name, dep]) => declareDep(name, dep));
 
+// Nb Starting on version 2.1.0 Bitcoin family apps no longer depend on
+// Bitcoin as a library. We will soon have the dependencies resolved from
+// back-end but until then we need to manually remove them.
 const versionBasedWhitelistDependencies = {
   Bitcoin: "2.1.0",
 };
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/apps/polyfill.ts
```diff
@@ -67,28 +67,39 @@ export function declareDep(name: string, dep: string): void {
...
+};
+
+export const getDependencies = (
+  appName: string,
+  appVersion?: string
+): string[] => {
+  const maybeDirectDep = directDep[appName];
+  if (!maybeDirectDep) return [];
+
+  // Nb Starting on version 2.1.0 Bitcoin family apps no longer depend on
```
gre, line 82: this comment belong to line 72 I think? before the `Bicoin: "2.1.0"`

gre (approved):

libs/ledger-live-common/src/apps/polyfill.ts
juan-cortes: Bitcoin is the only currency currently doing this standalone approach, but I guess maybe it will extend to other currencies 🤔. I'm ok with having it there for now and removing it once it's more than just Bitcoin. If it annoys you a lot I can change it though 🙏🏼

haammar-ledger (approved):
:pray:

PR #2473, opened 2023-01-30
Make the Bot freeze objects to detect possible mistakes

gre:
### 📝 Description

from Slack discussion, we suspect some coins have bug where they modify the original `Account` object, that is supposed to be preserved as immutable.
To detect such issue, we introduce in the bot an usage of `Object.freeze` which in combination with `"use strict"` would make the bot crash on case of mutating the existing account fields in case they get written (it makes field "read-only")

### ❓ Context

- **Impacted projects**: `bot` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/curly-papayas-itch.md
+++ b/.changeset/curly-papayas-itch.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Fixes Tron signOperation immutability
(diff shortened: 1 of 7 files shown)


gre:
```
⚠️ TypeError: Cannot add property _bot_context, object is not extensible
```

working on this tomorrow

gre:
coin to study more: cardano, crypto org, near, stellar, tron

haammar-ledger (approved):

PR #2478, opened 2023-01-31
Elrond: fixes immutability of syncESDTTokenAccountOperations

gre:
### 📝 Description

https://github.com/LedgerHQ/ledger-live/pull/2473 have revealed problem in Elrond implementation that alters the original object (input of the sync) which can cause subtle bugs all over our stack.

sync must embrace the immutability paradigm and not alter the incoming objects / create copy of objects when necessary.

There were a few mutating issue:
- `tokenAccounts.operations` getting mutated if the token account already exists and it doesn't recreate a ref for that token account. (possible consequence: UI may not always refresh to show the latest state of the account if user stay on the token account, notably on LLM)
- same for `tokenAccounts.balance`
- `operation.extra` mutated in the spec for the bot. this may hide real bug on that test

### ❓ Context

- **Impacted projects**: `*` 

### ✅ Checklist

- [x] **Test coverage** running the bot on this PR 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/elrond/js-buildSubAccounts.ts
+++ b/libs/ledger-live-common/src/families/elrond/js-buildSubAccounts.ts
@@ -75,10 +75,12 @@ async function syncESDTTokenAccountOperations(
   );
   const operations = mergeOps(oldOperations, newOperations);
 
-  tokenAccount.operations = operations;
-  tokenAccount.operationsCount = operations.length;
+  if (operations === oldOperations) return tokenAccount;
 
-  return tokenAccount;
+  const copy = { ...tokenAccount };
+  copy.operations = operations;
+  copy.operationsCount = operations.length;
+  return copy;
 }
 
 async function elrondBuildESDTTokenAccounts({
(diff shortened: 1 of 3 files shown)


libs/ledger-live-common/src/families/elrond/js-buildSubAccounts.ts
```diff
@@ -75,10 +75,12 @@ async function syncESDTTokenAccountOperations(
   );
   const operations = mergeOps(oldOperations, newOperations);
 
-  tokenAccount.operations = operations;
-  tokenAccount.operationsCount = operations.length;
+  if (operations === oldOperations) return tokenAccount;
 
-  return tokenAccount;
+  const copy = { ...tokenAccount };
```
haammar-ledger, line 80: This only does a shallow copy, right? Can't that change the behavior?

haammar-ledger (approved):

libs/ledger-live-common/src/families/elrond/js-buildSubAccounts.ts
gre: it should properly make the ref change when operations changes, so it behaves like all other coins 👍

libs/ledger-live-common/src/families/elrond/js-buildSubAccounts.ts
```diff
@@ -133,12 +135,20 @@ async function elrondBuildESDTTokenAccounts({
...
           accountAddress
         );
-        tokenAccount.balance = new BigNumber(esdt.balance);
-        tokenAccount.spendableBalance = new BigNumber(esdt.balance);
+        const balance = new BigNumber(esdt.balance);
+        if (!balance.eq(tokenAccount.balance)) {
+          // only recreate the object if balance changed
+          if (inputTokenAccount === tokenAccount) {
+            tokenAccount = { ...tokenAccount };
+          }
```
haammar-ledger, line 148: I don't get what this is for :thinking:

haammar-ledger: Ooh it just makes the ref change I guess. Got it.

haammar-ledger (approved):

PR #2511, opened 2023-02-02
LIVE-5363 - Added network fee E2E test steps and fixed network fee race condition

ggilchrist-ledger:
### 📝 Description

[Adding network fee tests](https://ledgerhq.atlassian.net/browse/LIVE-5363)

Adding steps to the swap Playwright E2E test that:
- opens the network tab
- checks the standard and advanced fees can be clicked
- The fee is updated to reflect a change

Also a bug has been fixed that caused a race condition on the network fee drawer. Basically the Transaction object in the Redux store was being set to `null` first and then the transaction. This meant that the drawer was trying to read a null object before it had been set to the transaction, which was only picked up on the fast moving PW tests. 

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/87659842/217205315-0595e86d-fbb5-4a00-84dc-94e8e28e619b.mov

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/reducers/swap.js
```diff
@@ -32,9 +32,10 @@ export type UPDATE_PROVIDERS_TYPE = {
   payload: $NonMaybeType<$PropertyType<SwapStateType, "providers">>,
 };
 const updateProviders = (state: SwapStateType, { payload: providers }: UPDATE_PROVIDERS_TYPE) => {
+  console.log("UPDATE PROVIDERS GETTING CALLED");
```
cjordan-ledger, line 35:
```suggestion
```

cjordan-ledger (commented):
Looks good, one tiny comment

apps/ledger-live-desktop/src/renderer/reducers/swap.js
ggilchrist-ledger: Ooops, missed that. Thanks

PR #2540, opened 2023-02-07
LLM - Auth vulnerability by date bypassing - fix.

mlegall-ledger:
### 📝 Description

When using LLM password prompt feature, we're currently logging current time when the application goes background. When coming back to the app, we compare the time the app went background to the current time. If the delay is big enough, we prompt the user for his password.
A vulnerability was reported in our bugbounty program : the password prompt can be defeated by setting the device time to a date before the app went background.
The proposed solution was the one retained : instead of comparing times, we use an unchangeable value, the device uptime. If the uptime retained when the app went background is smaller than the current uptime, we prompt for the password.

### ❓ Context

- **Impacted projects**: LLM
- **Linked resource(s)**: LIVE-5060

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

N/A

### 🚀 Expectations to reach


apps/ledger-live-mobile/ios/ledgerlivemobile/Info.plist
```diff
@@ -10,7 +10,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>$(APP_NAME)</string>
+	<string>LL [STAGING]</string>
```
mcayuelas-ledger, line 13: Is this needed?

mlegall-ledger: You're right, not sure I should commit that. Either way this is replaced when doing pods build (so in the CI too). Getting rid of that change.

mcayuelas-ledger (approved):

PR #2570, opened 2023-02-09
LLM - Feat/LIVE-5016 - Newfeed page

nparigi-ledger:
### 📝 Description

Implement [Cryptopanic API](https://cryptopanic.com/developers/api/), a crypto-related news aggregator, and create a new hook, useCryptopanicPosts, to easily use it on 

Add Explore/news tab navigator, with inside Learn page and a new Newsfeed page 
Add newsfeedPage Feature flag

#### New shared components : 

**FormatRelativeTime**
Add FormatRelativeTime component using Intl.RelativeTimeFormat to get a relative date to compare two date (ex: "3 minutes ago) and use it for newsfeed, cause the current date formatting was awful
Polyfill for Intl.RelativeTimeFormat  ( @formatjs/intl-relativetimeformat ) is really not that big, so that's ok.

Misc: Deprecated old Button component

#### New package.json modules : 

**[@formatjs/intl-relativetimeformat](https://formatjs.io/docs/polyfills/intl-relativetimeformat/)** for polyfilling  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat that we use to get relative time of the publish date of an article

**[react-native-inappbrowser-reborn](https://www.npmjs.com/package/react-native-inappbrowser-reborn)** to implement the native in-app browser we open to external URL when clicking an article

### ❓ Context

- **Impacted projects**: `` live-mobile
- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-5016
### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Group 2 (2)

https://user-images.githubusercontent.com/89014981/219688374-e03aa837-eace-4afc-a2df-9f61e9f8cd31.mp4

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/components/Button.tsx
```diff
@@ -50,6 +50,9 @@ type Props = BaseButtonProps & {
   isFocused: boolean;
 };
 
+/**
+ * @deprecated This button is a wrapper around native-ui, trying to translate the props of this component to those of the native-ui button. Please use directly Button from @ledgerhq/native-ui or the wrapped one from ./wrappedUi if you want to use event and eventProperties props.
+ */
```
ofreyssinet-ledger, line 55: noice

apps/ledger-live-mobile/src/hooks/newsfeed/useIsNewsfeedAvailable.ts
```diff
@@ -0,0 +1,20 @@
...
+  const { locale } = useLocale();
+  const newsfeedPageFeature = useFeature("newsfeedPage");
+
+  return (
+    newsfeedPageFeature?.enabled &&
+    newsfeedPageFeature?.params?.cryptopanicApiKey &&
+    cryptopanicAvailableRegions.includes(
+      locale as CryptopanicAvailableRegionsType,
+    ) &&
+    newsfeedPageFeature?.params?.whitelistedLocales?.includes(locale)
```
ofreyssinet-ledger, line 18: isn't there already a `languages_whitelisted` parameter for this kind of stuff?

ofreyssinet-ledger: and it's handled "natively" by useFeature() to return `enabled: false` in case a feature should be disabled for a given locale

mcayuelas-ledger (approved):

apps/ledger-live-mobile/src/hooks/newsfeed/useIsNewsfeedAvailable.ts
ofreyssinet-ledger: well to be exact: it's not handled directly by `useFeature`, it's implemented in the `getFeature` that is passed to the `FeatureFlagsProvider`

cgrellard-ledger (approved):

apps/ledger-live-mobile/src/hooks/newsfeed/useIsNewsfeedAvailable.ts
nparigi-ledger: Ah yes you are right, didn't know ... :face_exhaling: 
Don't know if i can switch this before code freeze...

apps/ledger-live-mobile/src/hooks/newsfeed/useCryptopanicPosts.ts
```diff
@@ -0,0 +1,101 @@
...
+  CryptopanicAvailableRegionsType,
+  CryptopanicGetParams,
+  CryptopanicNewsWithMetadata,
+  getPosts,
+} from "./cryptopanicApi";
+import { useLocale } from "../../context/Locale";
+
+type LoadingStateType = null | "initial" | "refreshing" | "loadingMore";
+
+export function useCryptopanicPosts(
```
ofreyssinet-ledger, line 14: several remarks about this hook:
we should avoid big "do-it-all" hooks like this for the sake of performance (especially that this one is used in a component that renders a list of items)
- I think the state should be kept in one or multiple Context providers at the top level of the app so that everything doesn't have to be re-initialised every time the hook is mounted (which will happen every time the user navigates to that page)
- then to avoid expensive re-renders of the list you might want to split it in several contexts:
  - one that just provides the list of posts that will be consumed and the bare minimum
  - one that provides loading state
  - another that provides access to the API (refresh, loadMore)
- then you split the big screen component so that each part (header, list of posts, footer) can use one of these contexts without being re-rendered uselessly and expensively every time the loading state changes)

... but yeah then you start see that using a Context for keeping frequently updating state like this is not ideal because (even if it's better than just a hook), so using redux here would make sense so that every little part of the screen can consume exactly what it needs in the store without being re-rendered too frequently

apps/ledger-live-mobile/src/screens/Newsfeed/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const CRYPTOPANIC_URL = "https://cryptopanic.com";
+const CRYPTOPANIC_DEFAULT_PARAMS = {
+  metadata: true,
+  approved: true,
+  public: true,
+};
+
+function NewsfeedPage() {
```
ofreyssinet-ledger, line 31: You would certainly benefit in terms of performance (because I saw in the code of the hook that you had perf issues) by splitting this component in several components that consume the state from one/several context (or possibly even better from the redux store).

For instance a header, footer, empty state and the main list, and each item would be its own memoized component. And if you do that properly the main component would have barely any state in it, each part would be updating just when needed.

libs/ui/packages/native/src/components/Cards/InformativeCard/index.tsx
```diff
@@ -31,14 +52,15 @@ const CardContainer = (props: CardProps): React.ReactElement => {
         </Text>
       </Flex>
 
-      <Flex width={100} height={75}>
+      {imageUrl && (
```
ofreyssinet-ledger, line 55: prefer a ternary `imageUrl ? (/*...*/) : null` here

mlegall-ledger: Isn't that syntax preferred for specific cases (eg. when evaluating an array length, or when there's need for fallback ?). In that case I can't see any scenario where this would be an issue.

ofreyssinet-ledger: Alternatively to the ternary you can use `!!imageUrl && whatever` or `Boolean(imageUrl) && whatever`

ofreyssinet-ledger: Basically when you do conditional rendering with a `&&` make sure that to the left of the `&&` you always have either `null` or a boolean.

ofreyssinet-ledger (approved):

apps/ledger-live-mobile/src/hooks/newsfeed/useIsNewsfeedAvailable.ts
nparigi-ledger: Fixed there : https://github.com/LedgerHQ/ledger-live/pull/2757 @ofreyssinet-ledger

nparigi-ledger: @ofreyssinet-ledger Fixed there https://github.com/LedgerHQ/ledger-live/pull/2758

PR #2577, opened 2023-02-09
fix: lld - redirect to dex live-app using state instead of params

cjordan-ledger:
### 📝 Description

Refactors how we pass params from swap screen to live-app. According to typescript, `history.push` does not take in a `params` value (even though it seems to work). 

### ❓ Context

- **Impacted projects**: LLD
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-3880

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/119950081/217870331-0e711051-d84f-4996-838f-9b1bed8715b5.mov

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.jsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.jsx
@@ -389,7 +389,7 @@ const SwapForm = () => {
         `/platform/${getProviderName(exchangeRate.provider).toLowerCase()}`;
       history.push({
         pathname: providerURL,
-        params: {
+        state: {
           returnTo: "/swap",
           accountId: fromAddress,
         },
--- a/apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
@@ -40,8 +40,8 @@ export default function PlatformApp({ match, appId: propsAppId, location }: Prop
 
   const returnTo = useMemo(() => {
     const params = new URLSearchParams(search);
-    return params.get("returnTo") || internalParams?.returnTo;
-  }, [search, internalParams]);
+    return urlParams?.returnTo || params.get("returnTo") || internalParams?.returnTo;
+  }, [search, urlParams?.returnTo, internalParams?.returnTo]);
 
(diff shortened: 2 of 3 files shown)


apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
```diff
@@ -40,8 +40,8 @@ export default function PlatformApp({ match, appId: propsAppId, location }: Prop
 
   const returnTo = useMemo(() => {
     const params = new URLSearchParams(search);
-    return params.get("returnTo") || internalParams?.returnTo;
-  }, [search, internalParams]);
+    return urlParams?.returnTo || params.get("returnTo") || internalParams?.returnTo;
```
cjordan-ledger, line 43: We may be able to delete `params.get("returnTo") || internalParams?.returnTo` as I don't see any params being passed around LLD that way... but I haven't deleted them as I'm unsure if any deep links would have them.

Justkant: I would go for that instead, don't see why we would need the state here
```suggestion
    return internalParams?.returnTo || params.get("returnTo");
```

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.jsx
```diff
@@ -389,7 +389,7 @@ const SwapForm = () => {
         `/platform/${getProviderName(exchangeRate.provider).toLowerCase()}`;
       history.push({
         pathname: providerURL,
-        params: {
+        state: {
```
Justkant, line 392: IMO we can revert that
```suggestion
        params: {
```

Justkant (changes requested):
I would keep the `params` instead of `state` for the `history.push` and check first the `internalParams` for the `returnTo` memo

apps/ledger-live-desktop/src/renderer/screens/platform/App.jsx
Justkant: We might want to still keep `params.get("returnTo")` but not sure about `internalParams`

Justkant (approved):

PR #2589, opened 2023-02-10
POC of a Detox E2E Test involving a Device Action in the new paradigm

grsoares21:
### 📝 Description
This PR has a proposal on how to create E2E tests involving Device Actions on LLM.

A few key premises of this proposal:
- It only involves the new Device Action paradigm. We can use this as a strategy to create E2E tests (i.e. each time we migrate a device action to the new paradigm, it has to include the associated E2E tests)
- It only tests the UI rendering of the device actions and not really their logic, those will be tested separately withing the Device SDK itself.
- It doesn't depend on running the app with a specific environment variable such as MOCK=1
- The state evolution of the device actions is directly controllable from the test spec itself

#### How it works
In the new Device Action (DA) paradigm, from the consumer point of view, a DA is only a function that receives instructions of what needs to be executed in the device and returns an Observable that emits events containing the current state of the DA according to the evolution of that execution. For usage withing React and React Native based apps, we also provide within the SDK a hook for each DA that triggers it (via useEffect) and stores the emitted states (via useState).

For usage within the context of E2E tests, we export from the hook file a mockable variable for the DA that is going to be used within the hook. If we initialize that variable, it's possible to directly control the full state that will be stored in the hook. This is why one of the key premises is that we're not actually testing the logic of a DA, but only the UI rendering of a DA within the context of a React / RN app. By initializing this variable from an E2E test context we're able to fully control the state of a given DA.

#### How to mock a Device Action in a Detox E2E tests
In order to communicate with the app JS execution and correctly initialize the mockable variable, we use the pre-existing Socket communication (it communicates between the Detox server and the app JS execution).

The first step to mock the DA you want to test using the the `mockDeviceAction` function on the server bridge on Detox side passing the name of the device action. This will send a message to the client side to replace the real DA by a mocked version which we can control and emit events for.
image

In order to change the state of a DA you call the `updateDeviceActionState` function on the server bridge on Detox side passing the name of the device action and the new state to be used. This will send a message to the client side to emit an event with a new state from the mocked DA to be used by the corresponding hook.
image

All the names of the device actions and their corresponding state structures are typed accordingly. If there's a DA that is not listed in the name types, it should be added along with the mocked Subject to be used, all of this can be done inside the `deviceActionMocks.ts` file.

### ❓ Context

- **Impacted projects**: `ledger-live-mobile`, `ledger-live-common`
- **Linked resource(s)**: [FAT-806]

### ✅ Checklist

- [x] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes** 

### 📸 Demo
N/A

[FAT-806]: https://ledgerhq.atlassian.net/browse/FAT-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/e2e/bridge/client.ts
```diff
@@ -22,20 +25,60 @@ export function init(port = 8099) {
...
+    case "mockDeviceAction":
+      mockDeviceActions(msg.payload.deviceActionToMock);
+      break;
+    case "newDeviceActionSate":
+      updateDeviceActionState(msg.payload);
+      break;
     case "add":
     case "open":
-      e2eBridgeSubject.next(msg);
+      e2eDeviceDescriptorSubject.next(msg);
```
ggilchrist-ledger, line 81: Probably best not to call it specifically `DeviceDescriptor` as this bridge is also used to do other stuff like load user data (the `setGlobals` one I believe)

PR #2592, opened 2023-02-10
Feat/poc detox device action

grsoares21:
### 📝 Description
This PR has a proposal on how to create E2E tests involving Device Actions on LLM.

A few key premises of this proposal:
- It only involves the new Device Action paradigm. We can use this as a strategy to create E2E tests (i.e. each time we migrate a device action to the new paradigm, it has to include the associated E2E tests)
- It only tests the UI rendering of the device actions and not really their logic, those will be tested separately withing the Device SDK itself.
- It doesn't depend on running the app with a specific environment variable such as MOCK=1
- The state evolution of the device actions is directly controllable from the test spec itself

#### How it works
In the new Device Action (DA) paradigm, from the consumer point of view, a DA is only a function that receives instructions of what needs to be executed in the device and returns an Observable that emits events containing the current state of the DA according to the evolution of that execution. For usage withing React and React Native based apps, we also provide within the SDK a hook for each DA that triggers it (via useEffect) and stores the emitted states (via useState).

For usage within the context of E2E tests, we export from the hook file a mockable variable for the DA that is going to be used within the hook. If we initialize that variable, it's possible to directly control the full state that will be stored in the hook. This is why one of the key premises is that we're not actually testing the logic of a DA, but only the UI rendering of a DA within the context of a React / RN app. By initializing this variable from an E2E test context we're able to fully control the state of a given DA.

#### How to mock a Device Action in a Detox E2E tests
In order to communicate with the app JS execution and correctly initialize the mockable variable, we use the pre-existing Socket communication (it communicates between the Detox server and the app JS execution).

The first step to mock the DA you want to test using the the `mockDeviceAction` function on the server bridge on Detox side passing the name of the device action. This will send a message to the client side to replace the real DA by a mocked version which we can control and emit events for.
image

In order to change the state of a DA you call the `updateDeviceActionState` function on the server bridge on Detox side passing the name of the device action and the new state to be used. This will send a message to the client side to emit an event with a new state from the mocked DA to be used by the corresponding hook.
image

All the names of the device actions and their corresponding state structures are typed accordingly. If there's a DA that is not listed in the name types, it should be added along with the mocked Subject to be used, all of this can be done inside the `deviceActionMocks.ts` file.

### ❓ Context

- **Impacted projects**: `ledger-live-mobile`, `ledger-live-common`
- **Linked resource(s)**: [FAT-806]

### ✅ Checklist

- [x] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/87659842/218435920-23ea20d1-5c3a-49a7-8a32-2db1cf22e889.mov

[FAT-806]: https://ledgerhq.atlassian.net/browse/FAT-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/e2e/bridge/server.ts
```diff
@@ -60,7 +62,8 @@ export function addDevices(
   deviceNames.forEach((name, i) => {
     postMessage({
       type: "add",
-      payload: { id: `mock_${i + 1}`, name, serviceUUID: `uuid_${i + 1}` },
+      payload: { id: `mock_${i + 1}`, name, serviceUUID: "13d63400-2c97-0004-0000-4c6564676572" },
```
ggilchrist-ledger, line 65: Was the previous `uuid_1`, `uuid_2`, etc not working? @grsoares21

apps/ledger-live-mobile/e2e/setup.ts
```diff
@@ -4,6 +4,11 @@ import * as bridge from "./bridge/server";
 
 beforeAll(async () => {
   bridge.init();
+
+  // Certain setups might need the following 2 lines to work in debug mode
+  await device.reverseTcpPort(8081);
+  await device.reverseTcpPort(8099);
```
ggilchrist-ledger, line 10: What process is running on `8099`?

apps/ledger-live-mobile/e2e/specs/firmwareUpdate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  onboarded: true,
+  pinValidated: true,
+  providerName: null,
+  seTargetId: 855638020,
+  seVersion: "2.0.0",
+  targetId: 855638020,
+  version: "2.0.0",
+};
+
+describe("Receive BTC", () => {
```
ggilchrist-ledger, line 38: Should this be Firmware Update?

ggilchrist-ledger:
Added demo video

apps/ledger-live-mobile/e2e/specs/firmwareUpdate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    await debugSettingsPage.tapFeaturesDebugFlow();
+    await debugSettingsPage.tapFirmwareUpdateFlow();
+  });
+
+  it("should start update", async () => {
+    await deviceSelectionPage.tapAddWithBluetooth();
+    await deviceSelectionPage.tapConnectAnExistingLedger();
+
+    bridge.addDevices();
```
ggilchrist-ledger, line 67: Why do the devices need added at this stage? @grsoares21

apps/ledger-live-mobile/e2e/specs/firmwareUpdate.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        error: null,
+      },
+    });
+
+    await expect(debugSettingsPage.getUpdateProgress()).toHaveText("0.5");
+    await expect(debugSettingsPage.getUpdateLockedDevice()).toHaveText("false");
+    await expect(debugSettingsPage.getUpdateStep()).toHaveText("installingOsu");
+    await expect(debugSettingsPage.getUpdateError()).toHaveText("null");
+
+    await new Promise((resolve) => setTimeout(() => resolve("a"), 10000));
```
ggilchrist-ledger, line 104: What's the purpose of this promise?

ggilchrist-ledger:
Once the new paradigm is fully in place, does that mean we won't need to use the `e2eDeviceDescriptorSubject` at all and all mocking will be done with the `mockDeviceActions` and `updateDeviceActionState` functions? @grsoares21 ?

apps/ledger-live-mobile/e2e/setup.ts
grsoares21: It's the websocket service that communicates between the Detox server and the App itself.

apps/ledger-live-mobile/e2e/specs/firmwareUpdate.spec.ts
grsoares21: Yes, good catch

apps/ledger-live-mobile/e2e/specs/firmwareUpdate.spec.ts
grsoares21: This is where we open the device selection screen and it starts bluetooth scanning, so it needs to be added to appear in that screen so we can select.

apps/ledger-live-mobile/e2e/specs/firmwareUpdate.spec.ts
grsoares21: This is a left-over. I'm removing it, it was there only to give me a pause after the test was over so I can see what was going on.

grsoares21:
> Once the new paradigm is fully in place, does that mean we won't need to use the `e2eDeviceDescriptorSubject` at all and all mocking will be done with the `mockDeviceActions` and `updateDeviceActionState` functions? @grsoares21 ?

We will still need it whenever we pass through a device selection screen.

PR #2596, opened 2023-02-10
In tools, add a utility to test a sync on the web, for any account id

gre:
### 📝 Description

Add in our deployed tools an easy way to test synchronisation of an account by its id. **This is mostly useful for developers in order to more precisely measure performance, profile JS, inspect Network activity (with moderns tools) of what a synchronisation is exactly doing.**

Screenshot 2023-02-13 at 14 42 06

Screenshot 2023-02-13 at 14 53 12

### ❓ Context

- **Impacted projects**: `none` 

### ✅ Checklist

- [ ] **Test coverage** not needed it's a tool, devs test it! 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/tools/common-tools/src/demos/index.js
+++ b/tools/common-tools/src/demos/index.js
@@ -2,10 +2,12 @@ import LogsViewer from "./logsviewer";
 import lldSignature from "./lld-signature";
 import repl from "./repl";
 import ethtxtools from "./eth-tx-tools";
+import sync from "./sync";
 
 export default {
   LogsViewer,
   lldSignature,
   repl,
   ethtxtools,
+  sync,
 };
--- a/tools/common-tools/src/demos/sync/index.js
+++ b/tools/common-tools/src/demos/sync/index.js
@@ -1,0 +1,220 @@
+import "./live-common-setup";
+import React, { useEffect, useState } from "react";
+import { BigNumber } from "bignumber.js";
+import {
+  encodeAccountId,
+  decodeAccountId,
+  shortAddressPreview,
+  emptyHistoryCache,
+} from "@ledgerhq/live-common/lib/account";
+import {
+  findCryptoCurrencyById,
+  getCryptoCurrencyById,
+} from "@ledgerhq/live-common/lib/currencies/index";
+import { getAccountBridge } from "@ledgerhq/live-common/lib/bridge";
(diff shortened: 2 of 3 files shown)


tools/common-tools/src/demos/index.js
```diff
@@ -1,11 +1,13 @@
...
-import repl from "./repl";
+//import repl from "./repl";
 import ethtxtools from "./eth-tx-tools";
+import sync from "./sync";
 
 export default {
   LogsViewer,
   lldSignature,
-  repl,
+  //repl,
```
gre, line 10: this makes absolutely no sense but if i don't comment repl after adding the sync folder. I now get this:

```

./src/demos/repl/live-common-setup.js
Cannot find module: '@ledgerhq/hw-transport-webusb'. Make sure this package is installed.

You can install this package by running: npm install @ledgerhq/hw-transport-webusb.

```

maybe we have some sort of circular dep issue? or is it because of webpack somehow? but it worked BEFORE adding the sync/ demo (which depends more strongly on live-common then 🤔  )

maybe you know @jelbaz-ledger ?

tools/common-tools/src/demos/sync/index.js
```diff
@@ -0,0 +1,220 @@
...
+
+function App() {
+  // synchronise account with an id that is input in a input text field
+  const [accountId, setAccountId] = useState("");
+  const [accountIdError, setAccountIdError] = useState("");
+
+  const [account, setAccount] = useState(null);
+  const [accountError, setAccountError] = useState("");
+
+  useEffect(() => {
```
hzheng-ledger, line 104: just a question, why can't we just merge these two "useEffect" hook into one?

gre: indeed it could be unified. they don't have exactly the same meaning and they could however be split into two parts, like one is doing the inference & the other is focusing on synching when it's valid.

fair feedback but not blocking this tools to land i think ;-)

gre: btw, i think a better design would actually be to offer some sort of `const accountId = useInferred(userInputText)` that is then fed into the last effect; i just did this very quick

haammar-ledger (approved):

PR #2598, opened 2023-02-10
[LIVE-5490] Fix eth sync and perf

hedi-edelbloute:
### 📝 Description

Fix eth perf problems

We were calling 20 times the same page

This time we call the same page maximum 2 times

If last tx of last request is the same as the current one we stop calling the backend

The real fix is using the token pagination mechanism

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5490 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/ethereum/synchronisation.ts
+++ b/libs/ledger-live-common/src/families/ethereum/synchronisation.ts
@@ -580,18 +580,27 @@ const fetchCurrentBlock = ((perCurrencyId) => (currency) => {
   return f();
 })({});
 
-// FIXME we need to figure out how to optimize this
-// but nothing can easily be done until we have a better api
+// FIXME : We need to use the pagination token from backend
 const fetchAllTransactions = async (api: API, address, blockHeight) => {
   let getTransactionsResult: Tx[];
   let txs: Tx[] = [];
   let maxIteration = 20; // safe limit
+  let lastRequestTxHash: string | undefined;
 
   do {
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/src/families/ethereum/synchronisation.ts
```diff
@@ -586,12 +586,22 @@ const fetchAllTransactions = async (api: API, address, blockHeight) => {
   let getTransactionsResult: Tx[];
   let txs: Tx[] = [];
   let maxIteration = 20; // safe limit
+  let lastCallTxs: Tx[] | undefined;
```
haammar-ledger, line 589: Just an optim proposal: maybe you don't need to store the full txs array but just the hash of the last one?

haammar-ledger (approved):
LGTM, good catch!

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
hedi-edelbloute: Indeed !

hedi-edelbloute: Done :)

gre (approved):

gre:
Test coverage is checked but i don't think it should have. there is no test that would have detected the bug that was added

PR #2601, opened 2023-02-13
fix: load envs at js boot

gre:
### 📝 Description

Problem: in some cases, Ledger Live Desktop don't properly set the env from process.env : when our JS code expected getEnv() value to be available and settled at boot time.

To fix this, we simply can load the module that do the setEnv (experimental.tsx in the current case of LLD) before the rest of our codebase (notably of coin implementations)

A way to reproduce the issue:

```
EXPERIMENTAL_EXECUTION_ON_RENDERER=1 API_STELLAR_HORIZON=https://stellar.coin.ledger-stg.com pnpm dev:lld
```

and when you sync a stellar account, it was still using the production one (see in Network panel) because the stellar impl have a getEnv() call at the module level: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/stellar/api/horizon.ts#L26 (which makes it impossible to have an experimental settings for it btw, but at least we need to support its 'loads'. this isn't how getEnv is supposed to be used and i think it's noted as improvement on blockchain side but note it is just one example among possibly other cases)

### ❓ Context

### ✅ Checklist

- [ ] **Test coverage** uneasy to automate a test for this but i'm going to check later 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/index.js
+++ b/apps/ledger-live-desktop/src/renderer/index.js
@@ -1,3 +1,4 @@
 // @flow
 import "~/env";
+import "~/renderer/experimental"; // NB we need to load this first because it loads things from process.env and will setEnv properly at boot
 import "~/renderer/init";


gre:
ok i hate this XD

PR #2618, opened 2023-02-14
feat: ethereum explorer tx pagination

hedi-edelbloute:
eth use token for pagination

### 📝 Description

Use token for pagination for ethereum transaction retrieval

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5603 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/api/Ethereum.ts
+++ b/libs/ledger-live-common/src/api/Ethereum.ts
@@ -116,9 +116,10 @@ export type BlockByHashOutput = {
 export type API = {
   getTransactions: (
     address: string,
-    blockHeight: number | null | undefined,
-    batchSize?: number
-  ) => Promise<Tx[]>;
+    blockHeight?: number | null,
+    batchSize?: number,
+    token?: string
+  ) => Promise<{ txs: Tx[]; nextPageToken: string | undefined }>;
   getCurrentBlock: () => Promise<Block>;
   getAccountNonce: (address: string) => Promise<number>;
   broadcastTransaction: (signedTransaction: string) => Promise<string>;
@@ -169,8 +170,9 @@ export const apiForCurrency = (currency: CryptoCurrency): API => {
     async getTransactions(
       address,
       blockHeight,
-      batchSize = 2000
-    ): Promise<Tx[]> {
+      batchSize = 2000,
+      token
+    ): Promise<{ txs: Tx[]; nextPageToken: string | undefined }> {
       const query: any = {
         batch_size: batchSize,
         filtering: true,
@@ -179,6 +181,9 @@ export const apiForCurrency = (currency: CryptoCurrency): API => {
         query.block_height = blockHeight;
         query.order = "descending";
       }
+      if (token) {
+        query.token = token;
+      }
       const txData = await network({
         method: "GET",
         url: URL.format({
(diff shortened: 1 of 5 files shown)


libs/ledger-live-common/src/families/ethereum/synchronisation.ts
```diff
@@ -581,35 +581,22 @@ const fetchCurrentBlock = ((perCurrencyId) => (currency) => {
...
-      log("ethereum", "block.height missing!");
-      return txs;
-    }
-  } while (--maxIteration);
-
-  return txs;
+    const { txs, nextPageToken } = await api.getTransactions(
+      address,
+      blockHeight,
+      2000,
```
lambertkevin, line 592: This arbitrary value for the batch size has been discussed with someone ?

lambertkevin (approved):
1 comment and 1 question, other than that 👍

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
hedi-edelbloute: It was like this in v3, I didn't change it so this hasn't been discussed :man_shrugging:

libs/ledger-live-common/src/api/Ethereum.ts
```diff
@@ -116,9 +116,10 @@ export type BlockByHashOutput = {
 export type API = {
   getTransactions: (
     address: string,
-    blockHeight: number | null | undefined,
-    batchSize?: number
-  ) => Promise<Tx[]>;
+    blockHeight?: number | null,
```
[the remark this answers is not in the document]
hedi-edelbloute: This is in the function input not in an object, this allows the function to be called without the input, it's different imo

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
hedi-edelbloute: Would it make sense to use a lower/higher value ?

libs/ledger-live-common/src/api/Ethereum.ts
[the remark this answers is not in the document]
lambertkevin: The arguments of a function is technically an array/object as well:
```typescript
function hello (a,b,c) {
    return arguments.length;
}
hello("a", "b"); // 2
hello("a", "b", undefined); // 3
```
So here if you were rellying on `arguments.length` you can have the same kind of "side effects" by implying that `anything?: string | null | undefined` is the same thing as `anything?: string | null` because the value is optional.

I'm just picky here because I know it's been a topic lately that we should remove all the `undefined` in those cases in the project in general, but in some cases, it creates some unwanted behaviours.

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
lambertkevin: Ok I thought it wasn't here with v3 because I didn't see that value directly in synchronization.ts, all good then 👌

libs/ledger-live-common/src/api/Ethereum.ts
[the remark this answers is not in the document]
hedi-edelbloute: Oh ok just overall remark then, here we don't use the args array

It's the same topic as not defining a key in an object vs setting the value as undefined

PR #2620, opened 2023-02-14
LC-553: Add swift bridge/wrapper packages

harrisonf-ledger:
### 📝 Description

Adds packages needed to bridge hw-app-eth and hw-app-solana to Swift. These are used in Ledger Connect in conjunction with these Swift packages:

https://github.com/LedgerHQ/ios-ble-wrapper-eth
https://github.com/LedgerHQ/ios-ble-wrapper-solana
https://github.com/LedgerHQ/hw-transport-ios-ble

### ❓ Context

- **Impacted projects**: Ledger Connect
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LC-553

### ✅ Checklist

- [ ] **Test coverage**
- [x] **Atomic delivery** 
- [x] **No breaking changes**


libs/ledgerjs/packages/swift-bridge-hw-transport-ble/src/TransportBLESwift.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  promisify(callback: Callback): Promise<any> {
+    return new Promise((resolve, reject) => {
+      callback(function (response, error) {
+        if (error == "") {
+          resolve(response);
+        } else {
+          reject(error);
+        }
```
chabroA, line 33: This feels a bit counterintuitive 🤔 

What is the reason to compare specifically `error == ""`?

Is it similar to this suggestion, or is there some specific cases where response can be nullish but we still want to resolve?

```suggestion
        if (response) {
          resolve(response);
        } else {
          reject(error);
        }
```

chabroA (changes requested):
Couple style feedbacks and a question / interrogation related to the `promisify` function

chabroA:
Also I guess the monorepo Readme would need to be updated to add references to the newly created packages -> https://github.com/LedgerHQ/ledger-live#libs---libraries

Unless there is a smart script that does that automatically? cc. @gre @elbywan?

libs/ledgerjs/packages/swift-bridge-hw-transport-ble/src/TransportBLESwift.ts
harrisonf-ledger: I'm not sure why this was written this way originally, but after testing this, checking for a response instead seems to be fine.

chabroA (approved):

PR #2624, opened 2023-02-14
chore: reduce the frequency of bot triggers

gre:
### 📝 Description

Facts

- the bot is running too often, and notably with the increase of merging in develop it's too fast to run one per merge. Instead we move to a daily
- the bot is now greatly being used "à la carte", PR per PR which reduces the need to do non reg on develop (instead it's run on PRs more and more)
- we're not yet at 100% and it's too fast & noisy to run that much bot when most fails anyway
- full sunset of mere denis (as it was for phosphore) in order to use it only when needed or use their accounts as possible frozen data set too

Proposal

- rotation of bots is now moving to a daily basis (and not every 6 hours)
- on the main non reg seed (7, nitrogen), no longer running per commit merge but instead once per day.

### ❓ Context

- **Impacted projects**: `CI` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


valpinkman (approved):

haammar-ledger (commented):
What about running non-reg bots during the night? Was that considered before?
Just an idea, to allow running the test bots without conflict with non-reg runs.

PR #2641, opened 2023-02-15
feat/LIVE-5657 DebugErrors tool and attempt at unified rendering

juan-cortes:
### 📝 Description
Part of a wider effort to align CS, product, design and engineering in the way we handle error cases in our flows. Allowing even non-tech savvy colleagues to check the rendering and suggest modifications to specific error cases. The tool can be accessed via the Debug menu as always and will prompt both fullscreen and modal versions of the selected errors.

### Potentially breaking changes
Before this gets merged we need to go over all the flows and screen that are impacted to be sure we are not blocking the user and still offer a way out of errors. In particular I am aware of breaking analytics in some flows but no fix makes sense until we know from a product perspective exactly what makes sense to track and whether this is the right place to do so.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-5657` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/4631227/219034269-1a81fe89-a754-4efa-959a-a23a15762239.mov

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/e2e/specs/errors.spec.ts
+++ b/apps/ledger-live-mobile/e2e/specs/errors.spec.ts
@@ -1,0 +1,80 @@
+import fs from "fs";
+import { device } from "detox";
+import PortfolioPage from "../models/wallet/portfolioPage";
+import SettingsPage from "../models/settings/settingsPage";
+import DebugSettingsPage from "../models/settings/debugSettingsPage";
+import { loadConfig } from "../bridge/server";
+
+/**
+ * Note that when addings new errors the exported class name may or may not match the
+ * internal name, because we are inconsistent. If the filter doesn't find anything check
+ * the internal name used when the error is created.
+ */
+const items: string[] = [
+  // // Empty to not disrupt the tests on CI
+  // "LockedDeviceError",
+  // "ManagerDeviceLocked",
+  // "UpdateYourApp",
+  // "UserRefusedDeviceNameChange",
+  // "UserRefusedFirmwareUpdate",
+  // "ImageLoadRefusedOnDevice",
+  // "ImageCommitRefusedOnDevice",
+  // "DeviceNotOnboarded",
+  // "FwUpdateBluetoothNotSupported",
+];
+
+// Nb, since this spec is not run automatically but rather locally, fill the export
+// path for the screenshots when you run it.
+let outPath = "";
+
+let portfolioPage: PortfolioPage;
+let settingsPage: SettingsPage;
+let debugSettingsPage: DebugSettingsPage;
+
+beforeAll(async () => {
+  await loadConfig("1AccountBTC1AccountETH", true);
+  portfolioPage = new PortfolioPage();
+  settingsPage = new SettingsPage();
+  debugSettingsPage = new DebugSettingsPage();
+});
+
+describe("Error rendering tool", () => {
+  it("Navigate to Debug Errors and generate screenshots", async () => {
+    await portfolioPage.navigateToSettings();
+    await settingsPage.navigateToDebugSettings();
+    await debugSettingsPage.navigateToDebugFeaturesSettings();
+    await debugSettingsPage.navigateToDebugErrors();
+
+    // Do this twice, once for modals, once for full-screen
+    for (const type of ["full", "modal"]) {
+      for (const name of items) {
+        // Open the selector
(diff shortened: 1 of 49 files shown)


apps/ledger-live-mobile/src/components/CustomImageDeviceAction.tsx
```diff
@@ -116,50 +103,21 @@ const CustomImageDeviceAction: React.FC<Props & { remountMe: () => void }> = ({
...
-                  : analyticsErrorTryAgainEventProps
-              }
-            >
-              {isRefusedOnStaxError
-                ? t("customImage.uploadAnotherImage")
-                : t("common.retry")}
-            </Button>
-            {isRefusedOnStaxError ? (
-              <Flex py={7}>
-                <Link
```
ofreyssinet-ledger, line 152: just writing this here so we don't forget to put back a "skip" button somehow once the discussion with product/design has advanced

juan-cortes: Likely we could have a conditional secondary action that is shown only when it receives the prop (like we discussed), it should be easy to implement and would be compatible with a device action version that doesn't have it.

apps/ledger-live-mobile/src/hooks/useErrorRenderData.ts
```diff
@@ -0,0 +1,132 @@
...
+      break;
+
+    case error instanceof ManagerDeviceLockedError:
+      data.Icon = Icons.LockAltMedium;
+      data.iconColor = "warning.c80";
+      data.hasExportLogs = false;
+      break;
+
+    default:
+      // Yaiza
```
alexandremgo, line 125: 😂

juan-cortes: TS wanted something there, again, this PR is not for review really, it should've been a draft :trollface:

apps/ledger-live-mobile/e2e/specs/errors.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        await debugSettingsPage.selectErrorByName(name);
+
+        if (type === "modal") {
+          // We want the modal view, so toggle it before the screenshot
+          await debugSettingsPage.toggleErrorRenderingType();
+        }
+
+        // We are in the tool now, create some sort of loop that goes over the errors and outputs
+        // screenshots of the device.
+        const dir = `/Users/juan.cortes-ext/Desktop/Screenshots/${type}/`;
```
alexandremgo, line 58: This won't work on other computers - is it the way to save screenshots ?

apps/ledger-live-mobile/e2e/specs/errors.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+beforeAll(async () => {
+  await loadConfig("1AccountBTC1AccountETH", true);
+  portfolioPage = new PortfolioPage();
+  settingsPage = new SettingsPage();
+  debugSettingsPage = new DebugSettingsPage();
+});
+
+describe("Error rendering tool", () => {
+  it("waits for Onboarding page to be ready", async () => {
```
alexandremgo, line 38: Not sure it's what the test is doing 🙃

apps/ledger-live-mobile/e2e/specs/errors.spec.ts
juan-cortes: Good catch, I forgot this.

apps/ledger-live-mobile/e2e/specs/errors.spec.ts
juan-cortes: hahahaha copy paste me

juan-cortes:
Rebased on develop, and addressed the issues @alexandremgo pointed out.
Still not mergable though.

juan-cortes:
Let us close this PR for now, as @alexandremgo will attempt to retrieve any recoverable work the branch. We can reassess the situation at a later time, after our Product, Customer Service, and Design teams have agreed on the most appropriate way to display errors, including relevant support links, tracked events, calls-to-action, and wording. 
Once we have established these guidelines, we can proceed with implementing the necessary changes.

PR #2651, opened 2023-02-15
Feat/LIVE-5401 : Add go to live app button inside webview topbar

brunospolidorio:
### 📝 Description

Add the option inside webview topbar to render a button to go back to live app when the app redirects the user to external url

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-5401` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

image

image

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
+++ b/apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
@@ -128,18 +128,17 @@ const WebPlatformTopBar = ({
   config = {},
   webviewRef,
 }: Props) => {
-  const { name, icon } = manifest;
+  const { name, icon, displayBackToManifestApp } = manifest;
 
   const {
     shouldDisplayName = true,
     shouldDisplayInfo = true,
     shouldDisplayClose = !!onClose,
     shouldDisplayNavigation = false,
-    shouldDisplayBackToLedgerLiveApp = false,
   } = config;
 
-  const lastLedgerLiveAppURL = useRef("");
-  const [isLedgerLiveAppURL, setIsLedgerLiveAppURL] = useState(true);
+  const lastManifestAppURL = useRef("");
+  const [isManifestAppURL, setIsManifestAppURL] = useState(true);
   const [canGoBack, setCanGoBack] = useState(false);
   const [canGoForward, setCanGoForward] = useState(false);
   const enablePlatformDevTools = useSelector(enablePlatformDevToolsSelector);
@@ -149,16 +148,16 @@ const WebPlatformTopBar = ({
     const webview = webviewRef.current;
 
     if (webview) {
-      if (shouldDisplayBackToLedgerLiveApp) {
+      if (displayBackToManifestApp) {
         const url = webview.getURL();
         const isLedgerLiveAppUrl = url.includes(manifest.url);
-        if (isLedgerLiveAppUrl) lastLedgerLiveAppURL.current = url;
-        setIsLedgerLiveAppURL(isLedgerLiveAppUrl);
+        if (isLedgerLiveAppUrl) lastManifestAppURL.current = url;
+        setIsManifestAppURL(isLedgerLiveAppUrl);
       }
       setCanGoBack(webview.canGoBack());
       setCanGoForward(webview.canGoForward());
     }
-  }, [webviewRef, shouldDisplayBackToLedgerLiveApp, manifest]);
+  }, [webviewRef, displayBackToManifestApp, manifest]);
 
   useEffect(() => {
     const webview = webviewRef.current;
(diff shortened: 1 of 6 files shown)


dreis-ledger (approved):

cjordan-ledger (approved):
Looks good 👍

sarneijim (approved):

sarneijim:
Happy, please wait to @IAmMorrow approval to merge

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
```diff
@@ -146,10 +149,16 @@ const WebPlatformTopBar = ({
     const webview = webviewRef.current;
 
     if (webview) {
+      if (shouldDisplayBackToLedgerLiveApp) {
+        const url = webview.getURL();
+        const isLedgerLiveAppUrl = url.includes(manifest.url);
```
cpellizzari-ledger, line 154: I would rather compare the domains for equality, no ?
It would be more robust to changes we may do like : 
- adding a query parameter to manifest.url : ...buy.app?sandboxes=true
- manifest.url not being the root url : ...buy.app/intro

brunospolidorio: Done!

dreis-ledger (approved):

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
```diff
@@ -128,7 +128,7 @@ const WebPlatformTopBar = ({
   config = {},
   webviewRef,
 }: Props) => {
-  const { name, icon } = manifest;
+  const { name, icon, displayBackToManifestApp } = manifest;
```
Justkant, line 131: I'm not a big fan of adding `displayBackToManifestApp` in the manifest and would actually prefer to have that behaviour by default if it makes sense

Justkant (changes requested):
Well done LGTM, I just don't think we have to change the manifest here wdyt ?

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
```diff
@@ -197,6 +206,13 @@ const WebPlatformTopBar = ({
     }
   }, [webviewRef]);
 
+  const onBackToLiveApp = useCallback(() => {
+    const webview = webviewRef.current;
+    if (webview) {
+      webview.loadURL(lastManifestAppURL.current);
```
IAmMorrow, line 212: How is "loadURL" behaving with the navigation history ? Is it erasing the history ? Pushing a new entry ? This could affect the overall user experience.

IAmMorrow (changes requested):

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
cpellizzari-ledger: I believe this is pushing a new entry by default. @brunospolidorio can you check ?
If easily fixed, I think the cleanest is to erase the history, but @IAmMorrow keep in mind this feature is not supposed to be used together with browser navigation, it is a workaround as browser navigation is broken on many of our providers apps.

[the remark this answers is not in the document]
cpellizzari-ledger: Good point, that would be a bug and developper should beware of it. @brunospolidorio can you add a warning comment here ? I think that will be enough.

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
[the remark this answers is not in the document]
cpellizzari-ledger: ok then... now you are aware of the feature in case the need arise at some point from other teams !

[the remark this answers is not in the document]
brunospolidorio: Should I remove the changes from manifest and include inside config like this:

`const {
    shouldDisplayName = true,
    shouldDisplayInfo = true,
    shouldDisplayClose = !!onClose,
    shouldDisplayNavigation = false,
    shouldDisplayBackToLiveApp = false
  } = config;`
  
 Is it ok @IAmMorrow ?

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/TopBar.jsx
IAmMorrow: Our team goal is to have navigation enabled for every single web3app soon so we can prepare for the web transition.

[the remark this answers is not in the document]
brunospolidorio: Added warn comment.

Justkant:
Closing as it should be replaced by https://github.com/LedgerHQ/ledger-live/pull/2891

PR #2653, opened 2023-02-15
fix(env.ts): add default non-undefined value for FEATURE_FLAGS json

ofreyssinet-ledger:
### 📝 Description

There was a warning at every start of LLD:
```
Invalid ENV value for FEATURE_FLAGS
```

Because of this:

```ts
if (value === undefined || value === null) {
    console.warn(`Invalid ENV value for ${name}`);
    return false;
}
```

and the fact that the `FEATURE_FLAGS` variable default was `""` which is parsed (with JSON.parse) as `undefined`.

So I replaced the default value `""` by `"{}"` which is parsed as a plain empty object, and that doesn't break the expected behaviour for overriding feature flags with the env variable:

```ts
function getFeature(key) {
    // ...
    const envFlags = getEnv("FEATURE_FLAGS") as
      | { [key in FeatureId]?: Feature }
      | undefined;

    if (allowOverride && envFlags) {
      const feature = envFlags[key];
      if (feature)
        return {
          ...feature,
          overridesRemote: true,
          overriddenByEnv: true,
        };
    }
}
```

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/rich-carpets-divide.md
+++ b/.changeset/rich-carpets-divide.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-framework": patch
+---
+
+env.ts: add default non-undefined value for FEATURE_FLAGS json
--- a/libs/coin-framework/src/env.ts
+++ b/libs/coin-framework/src/env.ts
@@ -711,7 +711,7 @@ const envDefinitions: Record<
     desc: "bucket S3 of the dynamic cryptoassets list",
   },
   FEATURE_FLAGS: {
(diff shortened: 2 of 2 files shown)


juan-cortes (approved):
Silences the terminal complaint, everybody wins.

PR #2670, opened 2023-02-16
bugfix/LIVE-5364: Fix bug in device language change prompt

juan-cortes:
### 📝 Description

We were not storing the changed language on the last seen device information leaving to a missed opportunity to prompt the user to also change the device language when they changed the app language and we can offer it. It's easier to understand with the test instructions really.

### ❓ Context

- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-5364 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/4631227/219421109-18a0d5a4-49ed-4aa3-a6b5-ec5660668b3f.mp4

### 🚀 Expectations to reach
- LLM should remember the last seen device language now.
- Changing LLM to a language supported by FW should prompt the user.
- If the last seen device language is already LLM's language, no prompt.

### DIFF
--- a/apps/ledger-live-mobile/src/components/ModalLock.tsx
+++ b/apps/ledger-live-mobile/src/components/ModalLock.tsx
@@ -6,7 +6,7 @@ import { setModalLock } from "../actions/appstate";
 /**
  * If mounted, lock all drawers using the modalLock redux app state (components/QueuedDrawer)
  *
- * The modal lock is reset when loosing the react-navigation screen focus.
+ * The modal lock is reset when losing the react-navigation screen focus.
  *
  * Used in some DeviceAction displayed in drawers:
  * All steps of a DeviceAction that have an ongoing operation on
--- a/apps/ledger-live-mobile/src/components/ModalLock.tsx
+++ b/apps/ledger-live-mobile/src/components/ModalLock.tsx
@@ -1,0 +1,24 @@
+import { useEffect } from "react";
+import { useDispatch } from "react-redux";
+import { setModalLock } from "../actions/appstate";
+
+/**
+ * All steps of a DeviceAction that have an ongoing operation on
+ * the device should disable the closing of the modal. If we are
+ * transfering something (installing an app, a language, CLS, etc)
+ * we don't want to break the flow.
+ */
+const LockModal = () => {
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(setModalLock(true));
+    return () => {
+      dispatch(setModalLock(false));
+    };
+  }, [dispatch]);
+
+  return null;
+};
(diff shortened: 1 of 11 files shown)


apps/ledger-live-mobile/src/reducers/settings.ts
```diff
@@ -473,6 +474,20 @@ const handlers: ReducerMap<SettingsState, SettingsPayload> = {
     },
   }),
 
+  [SettingsActionTypes.LAST_SEEN_DEVICE_LANGUAGE_ID]: (state, action) => {
+    if (!state.lastSeenDevice) return state; // Nb can't make you happy TS.
```
alexandremgo, line 478: I think it's normal to have this check on `lastSeenDevice`, because the action's `payload` is only a `Partial<DeviceInfo>` -> so if `lastSeenDevice` does not exist, we would have a `deviceInfo: Partial<DeviceInfo>` and not `deviceInfo: DeviceInfo` ;)

alexandremgo (approved):
👌

apps/ledger-live-mobile/src/reducers/settings.ts
juan-cortes: So it's not that I didn't want to make it happy, I just couldn't because of the nature of our relationship. Got it.

juan-cortes:
```suggestion
    if (!state.lastSeenDevice) return state;
```

ofreyssinet-ledger (approved):
lgtm 👍

apps/ledger-live-mobile/src/components/ModalLock.tsx
```diff
@@ -0,0 +1,24 @@
...
+    dispatch(setModalLock(true));
+    return () => {
+      dispatch(setModalLock(false));
+    };
+  }, [dispatch]);
+
+  return null;
+};
+
+export default LockModal;
```
alexandremgo, line 24: Should it be memoized to avoid re-rendering it when it's not necessary ? This will reduce the number of dispatch occurring

juan-cortes: I tested it (as per the video) and it didn't seem to cause any problem. Worst case scenario is we unlock and lock again.
I think we can live with it

alexandremgo: Ok !

juan-cortes:
Waiting for CI to green light this since it's QA'ed and reviewed

PR #2671, opened 2023-02-16
[LIVE-5080] Feature - Domain Service Resolution on LLD

henrily-ledger:
### 📝 Description

It's a naming service behind a feature flag (FeatureENS). it will help to convert something like "vitalik.eth" to his related eth address
Bonus we have a RTL test 👍 

### ❓ Context

- **Impacted projects**: `live-desktop` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5080 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Screenshot 2023-02-22 at 11 45 29
Screenshot 2023-02-22 at 11 45 33
Screenshot 2023-02-22 at 11 45 41

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.react.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+import { TransactionMode } from "~/../../../libs/ledger-live-common/lib/families/ethereum/modules";
+import RecipientField from "./RecipientField";
+import defaultTheme from "../../../styles/theme";
+import { InvalidAddress } from "@ledgerhq/errors";
+import { NamingServiceProvider } from "~/../../../libs/ledger-live-common/lib/naming-service";
+import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
+
+const eth = findCryptoCurrencyById("ethereum");
+
+const mockAccount = {
```
lvndry, line 26: can we type it ?

libs/ledger-live-common/src/naming-service/logic.ts
```diff
@@ -11,13 +12,12 @@ export function isOutdated(resource: NamingServiceStatus): boolean {
...
+      return now - resource.updatedAt > 30 * 1000; // 30 seconds
     }
   }
   return false;
 }
 
 export const isNameValid = (name: string | undefined): boolean => {
-  const regex = /[.*]\w{1,}/g;
-  return !!(name && regex.test(name));
+  return VALID_DOMAINS.some(str => name?.includes(str))
```
lvndry, line 22:
```suggestion
  return VALID_DOMAINS.some(str => name?.endsWith(str))
```

It's unlikely to happen but this function would return true if we have `foo.eth.xyz` or `foo.ethx`. not sure if we want it to be a valid case

apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientFieldNamingService.tsx
```diff
@@ -0,0 +1,69 @@
...
+  ]);
+
+  useEffect(() => {
+    if (
+      hasValidatedName &&
+      (transaction.recipient === transaction.recipientName || value !== transaction.recipientName)
+    ) {
+      onChangeTransaction(
+        bridge.updateTransaction(transaction, {
+          recipient: (namingServiceResponse as { address: string }).address,
```
lambertkevin, line 49: Don't know if you want to fix that, but I had a look about type narrowing in those cases and the solution would be to change:
```typescript
 const hasValidatedName = useMemo(() => namingServiceResponse.status === "loaded", [
    namingServiceResponse.status,
  ]);
```

into:
```typescript
 const hasValidatedName = useCallback((nsResponse: UseNamingServiceResponse): nsResponse is { status: "loaded"; address: string } => nsResponse.status === "loaded", []);
```

and use `hasValidatedName(namingServiceResponse)` in your condition instead. Might be worth to have that hook in LLC and create a specific type `UseNamingServiceResponseLoaded` ?

Might a bit of work for just avoiding a cast, so it's up to you 👍

libs/ledger-live-common/src/naming-service/logic.ts
```diff
@@ -11,13 +12,12 @@ export function isOutdated(resource: NamingServiceStatus): boolean {
       return now - resource.updatedAt > 60 * 1000; // 1 minute
     }
     case "error": {
-      return now - resource.updatedAt > 1 * 1000; // 1 second
+      return now - resource.updatedAt > 30 * 1000; // 30 seconds
```
lambertkevin, line 15: Not sure 30seconds cannot be an issue here in case of backend failure.
E.g. The backend is under heavy load or simply being updated for a sec so it's not responding, having to wait 30s to redo the call even if the user is changing the address might be an issue I think 🤔

lambertkevin (commented):
Just a few comments/questions, nothing major. 
Great job ! 🚀

apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientFieldNamingService.tsx
henrily-ledger: Nice thank you, will try to update it if it's help readability

libs/ledger-live-common/src/naming-service/logic.ts
henrily-ledger: Yeah good point I was more like, if someone got his address not found we doesn't need to refresh this, this quickly.
I can probably differ error and not found so we can modulate more easily this type of "error"

apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientFieldNamingService.tsx
henrily-ledger: Not so trivial still working on it :(

apps/ledger-live-desktop/package.json
```diff
@@ -40,6 +40,7 @@
     "release": "node ./tools/dist --release -v",
     "test": "pnpm test:jest && pnpm test:playwright",
     "test:codecheck": "node ./tools/ci.js",
+    "jest": "jest",
```
elGrogz, line 43: Is this needed? @henrily-ledger

henrily-ledger: I'll remove it, It was just a easier for me to test just one specific file with `pnpm desktop jest <filename>`

lambertkevin (approved):

PR #2672, opened 2023-02-16
feat: LLD - adds elrond stake banner to account screen

cjordan-ledger:
### 📝 Description

Adds elrond staking banner to account screen

Note: this is for the Elrond Multiverse account, and not the elrond ethereum token.

### ❓ Context

- **Impacted projects**:  LLD
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5090

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

#### Redelegation 
https://user-images.githubusercontent.com/119950081/220151570-7212ccb6-4a01-4180-b735-655d173d7325.mov

#### Delegation

https://user-images.githubusercontent.com/119950081/220151641-63b85766-333b-49be-9051-b9f53b6ccc33.mov

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
@@ -23,10 +23,8 @@ export const StakeBanner: React.FC<{ account: ElrondAccount }> = ({ account }) =
   const elrondPreloadData = useElrondPreloadData();
   const dispatch = useDispatch();
   const bannerState: AccountBannerState = getElrondBannerState(account, elrondPreloadData);
-  if (
-    stakeAccountBannerParams?.elrond?.delegate === false &&
-    bannerState.bannerType === "delegate"
-  ) {
+  const showDelegationBanner = bannerState.bannerType === "delegate";
+  if (stakeAccountBannerParams?.elrond?.delegate === false && showDelegationBanner) {
     return null;
   }
 
@@ -37,9 +35,21 @@ export const StakeBanner: React.FC<{ account: ElrondAccount }> = ({ account }) =
     return null;
   }
 
-  const title = "TODO";
-  const description = "DESCRIPTION";
-  const cta = bannerState.bannerType === "delegate" ? "DELEGATE" : "REDELEGATE";
+  const title = showDelegationBanner
+    ? t("account.banner.delegation.title")
+    : t("account.banner.redelegation.elrond.title");
+  const description = showDelegationBanner
+    ? t("account.banner.delegation.elrond.description")
+    : t("account.banner.redelegation.elrond.description");
+  const cta = showDelegationBanner
+    ? t("account.banner.delegation.cta")
+    : t("account.banner.redelegation.cta");
+  const linkUrl = showDelegationBanner
+    ? "https://www.ledger.com/staking"
+    : "https://support.ledger.com/hc/en-us/articles/7228337345693-Staking-MultiversX-EGLD-through-Ledger-Live-?support=true";
+  const linkText = showDelegationBanner
+    ? t("account.banner.delegation.linkText")
+    : t("account.banner.redelegation.linkText");
   const onClick = () => {
     track("button_clicked", {
       ...stakeDefaultTrack,
@@ -78,8 +88,8 @@ export const StakeBanner: React.FC<{ account: ElrondAccount }> = ({ account }) =
       cta={cta}
       onClick={onClick}
       display={bannerState.bannerType !== "hidden"}
-      linkText={t("account.banner.delegation.linkText")}
-      linkUrl={"https://www.ledger.com/staking-ethereum"}
+      linkText={linkText}
+      linkUrl={linkUrl}
     />
   );
 };
--- a/apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
@@ -24,14 +24,11 @@ export const StakeBanner: React.FC<{ account: ElrondAccount }> = ({ account }) =
   const dispatch = useDispatch();
   const bannerState: AccountBannerState = getElrondBannerState(account, elrondPreloadData);
   const showDelegationBanner = bannerState.bannerType === "delegate";
-  if (stakeAccountBannerParams?.elrond?.delegate === false && showDelegationBanner) {
+  if (!stakeAccountBannerParams?.elrond?.delegate && showDelegationBanner) {
     return null;
   }
 
-  if (
-    stakeAccountBannerParams?.elrond?.redelegate === false &&
-    bannerState.bannerType === "redelegate"
-  ) {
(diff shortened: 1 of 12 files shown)


libs/ledger-live-common/src/families/elrond/banner.ts
```diff
@@ -0,0 +1,106 @@
...
+    return {
+      bannerType: "delegate",
+    };
+  }
+
+  if (delegationAddresses.length > 0 && account.balance.isZero()) {
+    return {
+      bannerType: "hidden",
+    };
+  }
```
sarneijim, line 52: Can we use the same criteria than AccountHeaderManageAction is using? 
```
earnRewardEnabled && hasDelegations
```
(We can move those function to common and use them in LLM in the future.)

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const StakeBanner: React.FC<{ account: ElrondAccount }> = ({ account }) => {
+  const { t } = useTranslation();
+  const stakeAccountBanner = useFeature("stakeAccountBanner");
+  const stakeAccountBannerParams: StakeAccountBannerParams | null =
+    stakeAccountBanner?.params ?? null;
+
+  const elrondPreloadData = useElrondPreloadData();
+  const dispatch = useDispatch();
+  const bannerState: AccountBannerState = getElrondBannerState(account, elrondPreloadData);
+  const showDelegationBanner = bannerState.bannerType === "delegate";
```
sarneijim, line 26: Can we unify the criteria across all the banners?
(I don't mind use 'bannerType' and 'bannerState', but I think what it is more important is have a "common" approach)
```
const { redelegate, display, ...} = state;

File: apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+      token: "ELROND",
+    });
+
+    if (bannerState.bannerType === "redelegate") {
+      dispatch(
+        openModal(modals.unstake, {
+          account,
+          contract: bannerState.selectedDelegation.contract,
+          validators: elrondPreloadData.validators,
+          delegations: bannerState.mappedDelegations,
```
sarneijim, line 66: do we need to pass ```mappedDelegations``` from state? Is it not an "account" attribute?

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
cjordan-ledger: Yeah I agree, a common approach would be nice.

I did it this way as you get better type safety with a discriminated union type, i.e. you don't need optional params on `bannerState` for the redelegate fields. I'd be happy to refactor the others to keep it consistent if you're happy with the approach

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
cjordan-ledger: It is an account attribute that has been transformed. Annoyingly the modal doesn't open properly otherwise. This is duplicating this logic here:

https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/families/elrond/EarnRewards.jsx?plain=1#L86-L89

libs/ledger-live-common/src/families/elrond/banner.ts
```diff
@@ -0,0 +1,106 @@
...
+      bannerType: "redelegate",
+      mappedDelegations: account.elrondResources.delegations.map(
+        (delegation: ElrondDelegation) => ({
+          ...delegation,
+          validator: findValidator(delegation.contract),
+        })
+      ),
+      selectedDelegation,
+    };
+  } else if (!account.balance.isZero()) {
```
sarneijim, line 97: Idem

libs/ledger-live-common/src/families/elrond/banner.ts
cjordan-ledger: Yes, happy to use those functions. So will this then look like this?

```
if (!(earnRewardsEnabled && hasDelegation)) { return { bannerType: "hidden" } }
```

sarneijim (commented):
Changeset is missing :)

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
sarneijim: (But please use the destructuring syntax)

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
sarneijim: Should we "obtain" "mappedDelegations" directly in StakeBanner.tsx as it is independent of the banner state?

libs/ledger-live-common/src/families/elrond/banner.ts
cjordan-ledger: idem?

libs/ledger-live-common/src/families/elrond/banner.ts
sarneijim: As "redelegation" has preference...I am not sure if we must move this logic to the button, to simplify it.

```
 if (hasDelegation) {
{...}
  if (worstDelegator !== ledger) {
     return redelegate
  }
  if (earnRewardsEnabled) {
    return delegate
  } else {
    return hidden
  }
} 
```

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
cjordan-ledger: Yes we can do. I had it that way originally but abstracted to the helper - but I actually think it's better in StakeBanner as it seems like we need this just for LLD to work properly.

cjordan-ledger: https://github.com/LedgerHQ/ledger-live/pull/2672/commits/e2109f0e00214e2c5465f6008f16f9c71c3e0c5f

libs/ledger-live-common/src/families/elrond/banner.ts
cjordan-ledger: https://github.com/LedgerHQ/ledger-live/pull/2672/commits/e2109f0e00214e2c5465f6008f16f9c71c3e0c5f

apps/ledger-live-desktop/src/renderer/families/elrond/StakeBanner.tsx
cjordan-ledger: If it's okay with you, I'm going to refactor in a separate PR to minimise the changes in this PR.

PR #2677, opened 2023-02-17
Feat: lld show near staking banner

sarneijim:
### 📝 Description

Add stake banner to near account

### ❓ Context

- **Impacted projects**: LLD | Common
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-5091

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/near/banner.ts
+++ b/libs/ledger-live-common/src/families/near/banner.ts
@@ -1,0 +1,85 @@
+import { getCurrentNearPreloadData } from "./preload";
+import { canStake, canUnstake, FIGMENT_NEAR_VALIDATOR_ADDRESS } from "./logic";
+import type { NearAccount, NearValidatorItem } from "./types";
+
+export interface AccountBannerState {
+  display: boolean;
+  redelegate: boolean;
+  validatorId: string;
+  ledgerValidator: NearValidatorItem | undefined;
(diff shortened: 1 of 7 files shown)


libs/ledger-live-common/src/families/near/banner.ts
```diff
@@ -0,0 +1,85 @@
...
+  // Get ledger validator data
+  const { validators } = getCurrentNearPreloadData() ?? {
+    validators: [],
+  };
+
+  const ledgerValidator = validators.find(
+    (validator) => validator.validatorAddress === FIGMENT_NEAR_VALIDATOR_ADDRESS
+  );
+
+  // If Ledger doesn't provide validator, we don't display banner
```
cjordan-ledger, line 33: Do we need to hide the banner if `ledgerValidator` is not found? I can see that we're doing this for the other banners, but I don't see this in the acceptance criteria for these tickets 🤔

sarneijim: Not really, but in case that ledger validator address change (without change common variable) or close, we are covered. Theoretically, this is not going to happen.

PR #2691, opened 2023-02-20
fix: empty USER_ID on LLM

Justkant:
### 📝 Description

The `USER_ID` env var is empty on LLM so we're using `getOrCreateUser` that is async on LLM instead.
Also avoids using the package.json to get the name and version of the app/wallet.

### ❓ Context

- **Impacted projects**: `ledger-live-mobile` `ledger-live-desktop` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** Tested manually with wallet-api-tools (https://github.com/LedgerHQ/wallet-api/blob/main/apps/wallet-api-tools/manifest.json) 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Screenshot 2023-02-20 at 14 13 25
Screenshot 2023-02-20 at 14 19 50

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
+++ b/apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
@@ -13,6 +13,7 @@ import {
   TouchableOpacity,
   SafeAreaView,
 } from "react-native";
+import VersionNumber from "react-native-version-number";
 import { WebView as RNWebView } from "react-native-webview";
 import { useNavigation } from "@react-navigation/native";
 import { SignedOperation } from "@ledgerhq/types-live";
@@ -37,9 +38,12 @@ import prepareSignTransaction from "./liveSDKLogic";
 import { StackNavigatorNavigation } from "../RootNavigator/types/helpers";
 import { BaseNavigatorStackParamList } from "../RootNavigator/types/BaseNavigator";
 import { analyticsEnabledSelector } from "../../reducers/settings";
-import packageInfo from "../../../package.json";
+import getOrCreateUser from "../../user";
 
-const wallet = { name: packageInfo.name, version: packageInfo.version };
+const wallet = {
+  name: "ledger-live-mobile",
+  version: VersionNumber.appVersion,
+};
 const tracking = trackingWrapper(track);
 
 function useUiHook(): Partial<UiHook> {
@@ -159,6 +163,22 @@ function useUiHook(): Partial<UiHook> {
   );
 }
 
+const useGetUserId = () => {
+  const [userId, setUserId] = useState("");
+
+  useEffect(() => {
+    let mounted = true;
+    getOrCreateUser().then(({ user }) => {
+      if (mounted) setUserId(user.id);
+    });
+    return () => {
+      mounted = false;
+    };
+  }, []);
+
+  return userId;
+};
+
 function useWebView({ manifest, inputs }: Pick<Props, "manifest" | "inputs">) {
   const accounts = useSelector(flattenAccountsSelector);
   const navigation = useNavigation();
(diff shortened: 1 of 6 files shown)


apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
```diff
@@ -159,6 +163,22 @@ function useUiHook(): Partial<UiHook> {
...
+
+  useEffect(() => {
+    let mounted = true;
+    getOrCreateUser().then(({ user }) => {
+      if (mounted) setUserId(user.id);
+    });
+    return () => {
+      mounted = false;
+    };
+  }, []);
```
JunichiSugiura, line 177: If I understand collectedly, `mounted` flag is to prevent setting `userId` when the component is already unmounted?

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
```diff
@@ -37,9 +38,12 @@ import prepareSignTransaction from "./liveSDKLogic";
 import { StackNavigatorNavigation } from "../RootNavigator/types/helpers";
 import { BaseNavigatorStackParamList } from "../RootNavigator/types/BaseNavigator";
 import { analyticsEnabledSelector } from "../../reducers/settings";
-import packageInfo from "../../../package.json";
+import getOrCreateUser from "../../user";
 
-const wallet = { name: packageInfo.name, version: packageInfo.version };
+const wallet = {
+  name: "ledger-live-mobile",
```
JunichiSugiura, line 44: Doesn't want to re-use name from `package.json` anymore?

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
JunichiSugiura: Why this logic is different from desktop where simply calling `getEnv("USER_ID)` ?

JunichiSugiura (commented):
Some questions.

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
Justkant: We don't want to import `package.json` in the code and the name in it is not the same

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
Justkant: Because `USER_ID` is only set in LLD and not LLM

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
JunichiSugiura: ah, I forgot about the inconsistency between desktop and mobile (`ledger-live-desktop` vs `live-mobile` 😑)

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
JunichiSugiura: Could you remind me of why the diff exists (only set in desktop)?

JunichiSugiura (approved):

apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
Justkant: I don't really know the why, just that it was not present and `getOrCreateUser` is used in every analytics file of LLM

PR #2733, opened 2023-02-21
support/nowarnings Address some of the warnings on LLM dev builds

juan-cortes:
### 📝 Description

Today I sat down and actually read the warnings that get thrown when we launch Ledger Live Mobile instead of dismissing them. How hard can it be to have an app without warnings or at least without errors? Let's go deeper.

- `'deprecated-react-native-prop-types'` warning, it doesn't seem to be coming from us and according to [some guy on github](https://github.com/facebook/react-native/issues/33557#issuecomment-1093083115) it's safe to silence them.
- Memory leak on `remoteliveappprovider` because we launch some async task and the component gets unmounted before it completes yet we try to update the state afterwards. Simply using the `useIsMounted` hook that was already available we can prevent this leak.
- `Failed to fetch Firebase remote config..` it's kinda similar, fetching is async, and it gets cancelled because it gets unmounted. Checking that it's not unmounted before logging the error, no error.
- Circular dependencies on `native-ui` package. They seem to come from components importing from the `index` of a parent instead of the specific component they rely on, so `TimelineItem` which is under the `Layout` folder, importing `Flex` from `../..` instead of from `../../Flex`. No idea if there's a better way but changing the import removed the warning.
- There's a circular dependency between `react-native-crypto -> react-native-randombytes -> sjcl -> react-native-crypto` but that's no on our codebase, not sure how to tackle it, any help on this one would be much appreciated.
- `null is not a valid color or brush`. I actually looked into this a while ago and remember precisely what it was. It's usage of "null" inside svg files for some of our icons. Luckily I didn't have to spend time because I have good memory. Essentially `stroke="null"` is invalid yet present in `SGB.svg` and `FTM.svg` This gave a warning when listing all currencies, ie add accounts.
- Bunch of styling warnings in the Portfolio NFTs tab, we should've caught these during review probably. For styled components you need to provide the pxs now. So that should silence the warning.

So I couldn't catch them all, but together maybe?
🙏🏼 

### ❓ Context

- **Impacted projects**: `ledger-live-mobile, ledger-live-desktop, ledger-live-common, native-ui` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
App should behave the same as far as QA is concerned, but for developers it should show less warnings.


ofreyssinet-ledger:
I also just found this one that depends on react-native-ble-plx
Screenshot_1677060495

ofreyssinet-ledger (approved):
LGTM thanks for taking care of all these warnings 🙏

lambertkevin (approved):
You rock. 🚀

Justkant (approved):
LGTM, btw you don't need to separate the changesets anymore

PR #2744, opened 2023-02-22
[LIVE-5682] Feature - Address Poisoning Filter

lambertkevin:
### 📝 Description

This PR aims at temporarly remove ERC20 transactions with an amount = 0 from the UI to prevent address poisoning. 
Once possible, this fix should be directly applied to the respective coin integrations.

### ❓ Context

- **Impacted projects**: `coin-framework`, `live-common`, `ledger-live-desktop`, `ledger-live-mobile` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5682 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/44363395/220674281-47019685-b193-42db-be10-d7ac53e4fc2c.mov

https://user-images.githubusercontent.com/44363395/220674823-d5124ed3-7e16-4f5c-b419-7728a88b4a1f.mp4

### 🚀 Expectations to reach


.changeset/fuzzy-lemons-cough.md
```diff
@@ -0,0 +1,8 @@
+---
+"ledger-live-desktop": minor
```
hedi-edelbloute, line 2: Do we still need to split this into multiple changesets since we now deliver both at once ?

hedi-edelbloute (approved):
Looks fine overall, but I wonder... when do we need to add unit tests ?

libs/coin-framework/src/operation.ts
```diff
@@ -210,3 +210,17 @@ export const isConfirmedOperation = (
   operation.blockHeight
     ? account.blockHeight - operation.blockHeight + 1 >= confirmationsNb
     : false;
+
+export const isAddressPoisoningOperation = (
+  operation: Operation,
+  account: AccountLike
+): boolean => {
+  const familiesImpacted = ["ethereum", "evm", "tron"];
```
haammar-ledger, line 218: I don't consider it a blocker for this fix, but this isn't going in the right direction regarding coin modularization.

Potentially proper pattern could be to add an `isAddressPoisoningOperation` to the `AccountBridge`.
I'm reluctant to "pollute" the bridges, but at the same time anything that's isn't relevant / identical for _all_ coins should go through them.

hzheng-ledger (approved):
LGTM

.changeset/fuzzy-lemons-cough.md
lambertkevin: Nope

haammar-ledger (commented):
Just one comment, didn't have a close look at the rest.

libs/coin-framework/src/operation.ts
lambertkevin: Yup agreed, and I also have mixed feelings about polluting the bridge with too many things... We also had this discussion regarding NFTs related features and I'm still disturbed a year later about having some methods being 100% specific to only 1 familly... 
Might be a discussion to have with Stephane so we can figure out a "good looking" way to maybe provide a second layer of functionnalities that are not "basic" stuff like this one ? 🤔

lambertkevin: One change that I can already do, is to move that array as an env ?

haammar-ledger: :+1: we should have a discussion at some point with gre & Stephane about a proper strategy of when something should/shouldn't be added to bridges.

:+1: for the env too, I mean it's not a general solution but here it works because the behavior is exactly the same for all.

PR #2779, opened 2023-02-27
LIVE-5065 Add edit transaction helper and env variable

hzheng-ledger:
### 📝 Description

Add edit transaction helper and env variable (ETHEREUM_STUCK_TRANSACTION_TIMEOUT and isEditableOperation)

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


libs/coin-framework/src/operation.ts
```diff
@@ -225,3 +225,20 @@ export const isAddressPoisoningOperation = (
     operation.value.isZero()
   );
 };
+
+export const isEditableOperation = (
```
lvndry, line 229: Something we haven't discussed yet: when should we implement functions in the coin framework vs in live-common ?
Is it the role of the coin-framework the say if an operation is editable or not ?

hzheng-ledger: I suggest that we can put it in coin framework for this moment for two reasons.
1. The other helpers related to operation such as "isConfirmedOperation" and "isAddressPoisoningOperation" are implemented in coin framework. So we are doing something similar.
2. The methods in live-common can only be used in LLM and LLD but the methods in coin-framework can also be used by coin specific code in the future. e.g. if we have ethereum specific logic or UI code. (i.e. the "coin-evm" package in the future.) this part of evm specific code can use these helpers("isEditableOperation", ...)

Feel free to ask @sprohaszka-ledger for opinion and feel free to add more commit if you think it is necessary. Thanks

libs/coin-framework/src/operation.ts
```diff
@@ -225,3 +225,20 @@ export const isAddressPoisoningOperation = (
...
+  let isEthFamily = false;
+  if (account.type === "Account") {
+    isEthFamily = account.currency.family === "ethereum";
+  } else if (account.type === "TokenAccount") {
+    isEthFamily = account.token.parentCurrency.family === "ethereum";
+  }
+  return (
+    isEthFamily &&
+    operation.blockHeight === null &&
+    (operation.type === "OUT" || operation.type === "NFT_OUT")
```
lambertkevin, line 242: Don't think that line is really necessary, or at least too specific, I think you should probably just use the `operation.transactionRaw` parameter as a way to know if that transaction was crafter in the live

libs/coin-framework/src/operation.ts
sprohaszka-ledger: My thoughts are:
* if the function has a business logic impact, it can be added to `coin-framework`
* if the function has a UX/UI impact, it should not be in `coin-framework`

We all need to discuss about the scope of `coin-framework` :)

hzheng-ledger: ok, thanks a lot for the info.
As this method is UX/UI related for this moment. I move it out of coin-framework

libs/coin-framework/src/operation.ts
hzheng-ledger: fixed

PR #2407, opened 2023-01-25
feat/LIVE-4374 Updated wallet-api-server on LLD

IAmMorrow:
… in constructor

### 📝 Description

This pull request bump the wallet-api-server package to the latest version and provide wallet name, wallet version and user equipment id to the wallet-api-server instance.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

image

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #2781, opened 2023-02-27
feat: add wallet api storage support

cksanders:
### 📝 Description

- Updated wallet-api-server and wallet-api-core version
- Deprecate buggy custom LLD 'db' `getKey` and `setKey`(local file storage) in favour of electron-store module which reduces boilerplate code.
- Add Next storage & remove  obsolete `AsyncStorage_db_size_in_MB` & update Kotlin version as its a Next storage min requirement
- Added wallet api storage support
- Allows for 10MB per value in Android 
- NB: Next storage is flaky on Android emulator and causes app to crash/stall

```
const longString = (new Array(10*1024*1024)).join("p");
const bytes = new Blob([longString]).size
const sizeInMB = (bytes / 1048576).toFixed(2)

console.log(`${sizeInMB}MB`) => 10MB
```

### ❓ Context

- **Linked resource(s)**: replaces https://github.com/LedgerHQ/ledger-live/pull/2407

### Still To Do

As we have deprecated the [custom lld db](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/main/db/index.js) in favour of `electron-store` we will need to raise a tech debt ticket to refactor all usage of the custom lld db - ensuring there is only one local storage file for LLD - users aren't confused with multiple storage offerings. 

Recommended to refactor encrypted strings to use [Safe Storage](https://www.electronjs.org/docs/latest/api/safe-storage)


apps/ledger-live-mobile/android/build.gradle
```diff
@@ -16,15 +16,15 @@ buildscript {
...
+            ndkVersion = "23.1.7779620"
         }
     }
     repositories {
         google()
         mavenCentral()
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:4.2.2")
+        classpath("com.android.tools.build:gradle:7.4.1")
```
IAmMorrow, line 27: Sounds like a huge upgrade (v4 -> v7), maybe we don't want to handle this on our own only since this could perhaps have other consequences outside the scope of this team ?

libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -182,6 +182,12 @@ export interface UiHook {
     onError: (error: Error) => void;
     onCancel: () => void;
   }) => void;
+  "storage.get": (params: { key: string; storeId?: string }) => Promise<string>;
+  "storage.set": (params: {
+    key: string;
+    value: string;
+    storeId?: string;
+  }) => Promise<void>;
```
Justkant, line 190:
```suggestion
  "storage.get": WalletHandlers["storage.get"];
  "storage.set": WalletHandlers["storage.set"];
```

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/WebViewV2.tsx
```diff
@@ -89,6 +90,14 @@ function useUiHook(manifest: AppManifest): Partial<UiHook> {
           }),
         );
       },
+      "storage.get": async ({ key, storeId }) => {
+        const namespace = `wallet-api/${storeId}`;
+        return await getKey(namespace, key);
+      },
```
Justkant, line 96:
```suggestion
      "storage.get": ({ key, storeId }) => {
        const namespace = `wallet-api/${storeId}`;
        return getKey(namespace, key);
      },
```

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/WebViewV2.tsx
```diff
@@ -89,6 +90,14 @@ function useUiHook(manifest: AppManifest): Partial<UiHook> {
...
+      "storage.set": async ({ key, value, storeId }) => {
+        const namespace = `wallet-api/${storeId}`;
+        setKey(namespace, key, value);
+      },
```
Justkant, line 100:
```suggestion
      "storage.set": ({ key, value, storeId }) => {
        const namespace = `wallet-api/${storeId}`;
        setKey(namespace, key, value);
      },
```

Justkant (changes requested):
Some small suggestions around promises

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/WebViewV2.tsx
cksanders: get typescript errors doing this as expects return type to be `Promise<void>`

PR #2814, opened 2023-03-01
feat/FAT-793 UX Improvement for install set of apps

juan-cortes:
### 📝 Description

The original functionality of inline app installation was tightly coupled to a flow that had a hard dependency on a set of applications. For instance, accessing a swap flow for BTC-ETH would **need** all of the apps installed in order to succeed. However, with the use that we are giving it on the sync onboarding flow this dependency is no longer as strong. 

#### Install set of apps
This step of the onboarding will attempt to install a list of applications that can change between users since it's either something we provide from firebase or a "recover" list of apps from another device the user has interacted with. There is an odd chance that these apps are not available for installation in the newly onboarded device either because they're not developed, or not compatible, or even already installed.

#### More granular exposed state
The way inline app installations worked was very closed in terms of progress publishing, adapting it to `install-set-of-apps` required exposing a few flags, like dependency resolution, partial failures with reasons, etc. This means we can know what will be installed and what won't and why. This can also be used for other flows that attempt to install an app if we wanted, and LLD will also have to take advantage of this soon.

#### Tooling
Apart from the feature itself which can be hard to access, this PR provides a tool in the debugging menu to facilitate access to it. It has two modes, one of which takes the list from firebase and one which takes a known invalid list to debug the UI of failures.

### ❓ Context

- **Impacted projects**: `ledger-live-mobile, live-common` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/FAT-793` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/4631227/222235948-9376fba1-a217-4da5-bc2b-7cc369a0c24e.mov

### 🚀 Expectations to reach
There are some discrepancies in the design that come from an outdated design system. Components no longer matching what our designers are working with, and it's out of the scope of this feature to update them. Everyone is aware of this fact and the UI will adapt once the components are updated.

### DIFF
--- a/libs/ledger-live-common/src/apps/hw.ts
+++ b/libs/ledger-live-common/src/apps/hw.ts
@@ -87,7 +87,7 @@ export type InlineAppInstallEvent =
  * @param transport Transport instance
  * @param appNames List of app names to install
  * @param onSuccessObs Optional observable to run after the installation
- * @param skipAppInstallIfNotFound If true, skip the installation of any app that were not found from the provider. Default to false.
+ * @param allowPartialDependencies If true, keep installing apps even if some are missing
  * @returns Observable of InlineAppInstallEvent or ConnectAppEvent
  * - Event "inline-install" contains a global progress of the installation
  */
@@ -95,12 +95,12 @@ export const inlineAppInstall = ({
   transport,
   appNames,
   onSuccessObs,
-  skipAppInstallIfNotFound = false,
+  allowPartialDependencies = false,
 }: {
   transport: Transport;
   appNames: string[];
   onSuccessObs?: () => Observable<any>;
-  skipAppInstallIfNotFound?: boolean;
+  allowPartialDependencies?: boolean;
 }): Observable<InlineAppInstallEvent | ConnectAppEvent> =>
   concat(
     of({
@@ -124,7 +124,7 @@ export const inlineAppInstall = ({
               reducer(state, {
                 type: "install",
                 name,
-                skip: skipAppInstallIfNotFound,
+                allowPartialDependencies,
               }),
             initState(e.result)
           );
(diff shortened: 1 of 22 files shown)


libs/ledger-live-common/src/apps/hw.ts
```diff
@@ -228,26 +245,7 @@ export const listApps = (
           )
           .catch((e) => {
             log("hw", "failed to HSM list apps " + String(e) + "\n" + e.stack);
-
-            if (getEnv("EXPERIMENTAL_FALLBACK_APDU_LISTAPPS")) {
```
gre, line 232: is there an active reason to drop this? if yes, should you also drop the env related to it as well as the hwListApps implementation? not sure if it would therefore used at all?

juan-cortes: I'm dropping it because it was introduced 4 years ago and it has never been used or exposed to the UI since.
We maintain the command in case we want to introduce it as part of our flow and **[it's now documented](https://ledgerhq.atlassian.net/wiki/spaces/~5be5d09253edba78d2964720/pages/3753377984/An+attempt+at+APDU+specs)**  on our confluence. I see no reason to maintain that complexity in the hw.ts file in the meantime.

gre: then i suggest to drop the unused env 👍

juan-cortes: sure, you're right. Just did

alexandremgo (approved):
Perfect, saw nothing wrong 🔥

PR #2820, opened 2023-03-02
Support/use playwright http mocking for swap tests

ggilchrist-ledger:
### 📝 Description

In order to more properly E2E the swap flow, we will use the HTTP mocking that comes with Playwright, instead of relying on the `MOCK` logic in the LLD codebase. 

This is simple to implement - it's just adding the following `route.fulfill` code into the test and adding the mocked body you want. Eg:
```js
    await page.route("https://swap.ledger.com/v4/providers**", async route => {
      console.log("Mocking provider HTTP response");
      const mockProvidersResponse = getProvidersMock();
      route.fulfill({ body: mockProvidersResponse });
    });
```

The `route` object has lots of other ways of manipulating requests and responses that could be useful for E2E testing.

Doing this means we can:
- control the mocked data easier
- properly test that API calls have been made 
- test that LLD handles 'real' API responses correctly
- potentially we can remove some of the `MOCK` logic in Ledger Live Common (out of scope for this PR)

Also fixed some flakiness:
- Asserts on the `terms` spec returning the wrong  type to assert on
- Added wait for market search bar to be empty before assert 
- Added specific element to wait for in account drawer in case there are multiple matches

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/87659842/223438677-d5f443b2-a436-408f-a090-75c788cb30f4.mov

### 🚀 Expectations to reach


apps/ledger-live-desktop/tests/models/Drawer.ts
```diff
@@ -18,7 +18,7 @@ export class Drawer {
     this.currencyButton = (currency: string) =>
       page.locator(`data-test-id=currency-row-${currency.toLowerCase()}`);
     this.accountButton = (accountName: string, index: number) =>
-      page.locator(`data-test-id=account-row-${accountName.toLowerCase()}-${index}`);
+      page.locator(`data-test-id=account-row-${accountName.toLowerCase()}-${index}`).first();
```
ggilchrist-ledger, line 21: I saw one instance of a test failing because it found two of these. Since it's just one time it's probably flakiness, but this will make sure the element we want is clicked

gre (approved):

PR #2861, opened 2023-03-06
feat: add wallet api storage support

cksanders:
### 📝 Description

The storage feature will be used for the following use cases:

Allow inter-web3app data sharing
Allow ledger live to read data stored from web3app (facilitate using tokens in the recover case)
In the future, easily "cloud syncing" data from web3app between multiple ledger live clients
It's totally sandboxed from one app to another by default. The sandboxing logic reside in the WalletAPIServer package. For now, we only allow an app to fetch data from its own namespace. In the future, special permissions in the manifest will allow a web3app to fetch data from a namespace it doesn't own.

- Updated wallet-api-server and wallet-api-core version
- Add Next storage & remove  obsolete `AsyncStorage_db_size_in_MB` & update Kotlin version as its a Next storage min requirement
- Added wallet api storage support
- Allows for 10MB per value in Android, iOS & Electron
- NB: Next storage is flaky on Android emulator and causes app to crash/stall

### ❓ Context

- **Impacted projects**: replaces https://github.com/LedgerHQ/ledger-live/pull/2407

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
@@ -31,6 +31,7 @@ import { shareAnalyticsSelector } from "~/renderer/reducers/settings";
 import { Loader } from "./styled";
 import { WebviewAPI, WebviewProps, WebviewTag } from "./types";
 import { useWebviewState } from "./helpers";
+import { getStoreValue, setStoreValue } from "~/renderer/store";
 
 const wallet = { name: "ledger-live-desktop", version: __APP_VERSION__ };
 const tracking = trackingWrapper(track);
@@ -84,6 +85,12 @@ function useUiHook(manifest: AppManifest): Partial<UiHook> {
             onClose: onCancel,
           }),
         );
+      },
+      "storage.get": ({ key, storeId }) => {
+        return getStoreValue(key, storeId);
+      },
+      "storage.set": ({ key, value, storeId }) => {
+        setStoreValue(key, value, storeId);
       },
       "transaction.sign": ({
         account,
--- a/.changeset/strange-bears-complain.md
+++ b/.changeset/strange-bears-complain.md
@@ -1,0 +1,11 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
+"@ledgerhq/live-common": minor
+---
+
+- Update Kotlin version
+- Updated `wallet-api-server` and `wallet-api-core` version
+- Add Next storage & remove obsolete `AsyncStorage_db_size_in_MB`
+- Added wallet api storage support
+- Add stable & tested LLD storage support
--- a/apps/ledger-live-desktop/package.json
+++ b/apps/ledger-live-desktop/package.json
@@ -72,6 +72,7 @@
     "color": "^3.1.3",
     "dotenv": "^16.0.3",
     "electron-context-menu": "^3.6.0",
+    "electron-store": "^8.1.0",
     "firebase": "^9.6.6",
     "focus-trap": "^6.6.1",
     "fuse.js": "^3.6.1",
--- a/apps/ledger-live-desktop/src/main/db/index.ts
+++ b/apps/ledger-live-desktop/src/main/db/index.ts
@@ -56,15 +56,6 @@ function init(_DBPath: string) {
 }
 
 /**
- * Register a transformation for a given namespace and keyPath
- * it will be used when reading/writing from/to file
- */
-function registerTransform(ns: string, keyPath: string, transform: Transform) {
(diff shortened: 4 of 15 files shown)


apps/ledger-live-mobile/src/components/WebPlatformPlayer/WebViewV2.tsx
```diff
@@ -105,6 +106,14 @@ function useUiHook(): Partial<UiHook> {
...
         });
       },
+      "storage.get": async ({ key, storeId }) => {
+        const storageKey = storeId ? `${storeId}-${key}` : key;
+        return (await deviceStorage.get(storageKey)) as string;
+      },
+      "storage.set": ({ key, value, storeId }) => {
+        const storageKey = storeId ? `${storeId}-${key}` : key;
+        deviceStorage.save(storageKey, value);
+      },
```
IAmMorrow, line 116: storeId will always be defined here, in all cases. If the client didn't specify a custom storeId, the appId will be inferred instead

apps/ledger-live-desktop/src/main/index.js
```diff
@@ -111,10 +112,26 @@ app.on("ready", async () => {
...
+
+  /** @deprecated (Use "getStoreValue" instead) */
   ipcMain.handle("getKey", (event, { ns, keyPath, defaultValue }) => {
     return db.getKey(ns, keyPath, defaultValue);
   });
 
+  /** @deprecated (Use "setStoreValue" instead) */
   ipcMain.handle("setKey", (event, { ns, keyPath, value }) => {
     return db.setKey(ns, keyPath, value);
   });
```
IAmMorrow, line 137: Can we really consider this deprecated ? All user data are currently stored by the current lib. Moving away from it will require a migration process.

IAmMorrow (commented):
I'm afraid that by adding electron store, we'll end up indefinitely with two storage libraries at once. LLD's current storage library is doing a lot of encryption logic using the (optional) user's password. It need to support adding a password, removing it, changing it ... Unless those requirement are also met by electron-store I'm not sure it's a great move

apps/ledger-live-desktop/src/main/index.js
gre: i miss a lot of context on what we wanted to do here. do we have documentation on the part that you "noticed buggy" ?

On our side we know we want to consider better solution than app.json with tons of data in it but we would need to brainstorm a bit all solutions & what each would mean.

side note: electron-store used to be a library we were using in the past, and i'm not sure why we moved away from it.

Justkant (approved):
LGTM, there is only one mismatch between LLD & LLM when you get a value undefined
Screenshot 2023-03-21 at 14 50 51
Screenshot 2023-03-21 at 14 51 16

cksanders:
> LGTM, there is only one mismatch between LLD & LLM when you get a value undefined Screenshot 2023-03-21 at 14 50 51 Screenshot 2023-03-21 at 14 51 16

Updated to return `null` given empty.

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -79,6 +80,12 @@ function useUiHook(manifest: AppManifest): Partial<UiHook> {
           }),
         );
       },
+      "storage.get": ({ key, storeId }) => {
+        return getStoreValue(key, storeId);
+      },
+      "storage.set": ({ key, value, storeId }) => {
+        setStoreValue(key, value, storeId);
```
gre, line 87: do we have a more precise spec of what is `storage` and why it is needed for a given web3 app? I'm not super fan of the fact any web3 can query any other sibling web3 app as it doesn't seem to be sandboxed per app (or i missed something?)

cksanders: _"The storage feature will be used for the following use cases:
1. Allow inter-web3app data sharing
2. Allow ledger live to read data stored from web3app (facilitate using tokens in the recover case)
3. In the future, easily "cloud syncing" data from web3app between multiple ledger live clients

It's totally sandboxed from one app to another by default. The sandboxing logic reside in the WalletAPIServer package. For now, we only allow an app to fetch data from its own namespace. In the future, special permissions in the manifest will allow a web3app to fetch data from a namespace it doesn't own."_

c.c. @IAmMorrow

gre: ok thanks! this should be in the PR description btw

gre (approved):

PR #2864, opened 2023-03-07
Support LIVE-4440 btc fix integration test and cleanup

hzheng-ledger:
### 📝 Description

Btc fix integration test and cleanup. Due to the bitcoin nano app 2.1.0 migration, we had some APDU to check the bitcoin nano app version firstly and call the corresponding bitcoin API. Now we update the min required version for bitcoin app to 2.1.0 and we don't need this check any more on live side.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledger-live-common/src/apps/support.ts
```diff
@@ -37,8 +37,11 @@ const appVersionsRequired = {
   Solana: ">= 1.2.0",
   Celo: ">= 1.1.8",
   "Cardano ADA": ">= 4.1.0",
-  Zcash: "> 2.0.6",
+  Bitcoin: ">= 2.1.0",
```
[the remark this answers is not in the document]
hzheng-ledger: good point.
but I think we don't have to be worried about that. because since bitcoin (and other bitcoin family apps) 2.1.0
none of them depends on bitcoin nano app at this moment. (refer to https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/apps/polyfill.ts#L73-L78)
I've just now put all the bitcoin family apps in the appVersionsRequired list. So users should not have any problem about it.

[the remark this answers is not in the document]
haammar-ledger: Agree with Huiqi, not dependency anymore between bitcoin apps.

However for some coins in the list, this may force a firmware update? I don't remember why but this seemed to be an issue in the past, maybe for users on LLM?

libs/ledger-live-common/src/exchange/index.ts
```diff
@@ -11,7 +11,7 @@ import { findTestExchangeCurrencyConfig } from "./testCurrencyConfig";
 const exchangeSupportAppVersions = {
   bitcoin_cash: "1.5.0",
   bitcoin_gold: "1.5.0",
-  bitcoin: "1.5.0",
+  bitcoin: "2.1.0",
```
haammar-ledger, line 14: No need to update the version for the other ones than bitcoin?

haammar-ledger (commented):
In `specs.ts`, you can remove the `appVersion: "2.1.0-rc"` param for all specs, so the bot tests again with latest app versions.

Besides this LGTM but it will need to be tested properly:
- integration tests
- bot on the whole bitcoin family
- exchange flows

henrily-ledger (commented):
LGTM but snapshot seems to differ a lot, is this expected ?

libs/ledger-live-common/src/exchange/index.ts
hzheng-ledger: in fact, it doesn't matter as the exchange app 2.0.12 supports both old and new altcoin apps well. old and new altcoin app use the same APIs.
but anyway, as I've updated the required versions in support.ts. I update the versions for all the other ones now.

hzheng-ledger:
> LGTM but snapshot seems to differ a lot, is this expected ?

I just use the -u parameter to update the snapshot. I think it should be OK

libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
```diff
@@ -267,15 +260,4 @@ export default class Btc {
...
-  async changeImplIfNecessary(): Promise<BtcOld | BtcNew> {
-    const appAndVersion = await getAppAndVersion(this._transport);
-    if (appAndVersion.name === "Exchange" && this._impl instanceof BtcNew) {
-      const isBtcLegacy = await checkIsBtcLegacy(this._transport);
-      if (isBtcLegacy) {
-        this._impl = new BtcOld(this._transport);
-      }
-    }
-    return this._impl;
-  }
```
gre, line 280: what's the status of this PR change vs https://github.com/LedgerHQ/ledger-live/pull/2952

PR #2881, opened 2023-03-08
Feature - Domain-Service library

lambertkevin:
### 📝 Description

This PR creates a new lib in charge of handling all domains related features.
For now it's mostly focusing on providing an API to resolve and sign domains and addresses in reverse resolving.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


libs/domain-service/src/registries/index.ts
```diff
@@ -0,0 +1,57 @@
+import { Registry } from "../types";
+
+const REGISTRIES: Registry[] = [
```
henrily-ledger, line 3: Should we import this one later in an other file ?

henrily-ledger (approved):

libs/domain-service/src/registries/index.ts
lambertkevin: Yup, the goal is to have this thing from a backend instead once they will release it. 
That's why the `getRegistries` method is async, so we don't have to add async/awaits everywhere afterwards once it's coming from that backend. 👌

PR #2891, opened 2023-03-08
Feat/live-5401 Refactor webviews + implement PTX custom player

IAmMorrow:
### 📝 Description

1. Fully refactored the WebPlatformPlayer component on both LLD & LLM to allow easy custom implementations
2. Added a WebPTXPlayer component that implement a custom "BackToLastMatchingURL" button (LLD only)

### ❓ Context

- **Impacted projects**: `LLD`, `LLM` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-5401` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

image

image

### 🚀 Expectations to reach


ggilchrist-ledger:
Can confirm this branch is working with the previous Buy app (1.7) and the upcoming one (2.0) with will incorporate the Back button and get rid of the provider iframe.

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
```diff
@@ -0,0 +1,21 @@
...
+    refs.forEach(ref => {
+      if (!ref) return;
+
+      if (typeof ref === "function") {
+        ref(targetRef.current);
+      } else {
+        ref.current = targetRef.current;
+      }
+    });
+  }, [refs]);
```
Justkant, line 18: `refs` will always change here if I'm not mistaken and re-run the effect on every render, if you still want this solution there is another usage here https://mtsknn.fi/blog/react-fallback-ref/#merge-refs

Justkant (changes requested):
Some little things here and there

IAmMorrow:
Just pushed a major update, this PR is now covering both LLM & LLD webview rework.

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { WALLET_API_VERSION } from "@ledgerhq/live-common/wallet-api/constants";
+import React, { forwardRef } from "react";
+import { WalletAPIWebview } from "./WalletAPIWebview";
+import { PlatformAPIWebview } from "./PlatformAPIWebview";
+import TrackPage from "~/renderer/analytics/TrackPage";
+import { WebviewAPI, WebviewProps } from "./types";
+
+export const Web3AppWebview = forwardRef<WebviewAPI, WebviewProps>(
+  ({ manifest, inputs, onStateChange }, ref) => {
+    <TrackPage category="Platform" name="App" appId={manifest.id} params={inputs} />;
```
Justkant, line 13: I don't think this will work as you expect, because we don't return it in the react tree.
Also shouldn't we track "Platform" only on the `PlatformAPIWebview` ? or do we consider "Platform" the same as "WalletAPI" for tracking ?

libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -53,51 +60,11 @@ import { MessageData } from "../hw/signMessage/types";
...
-    manifest.url,
-    manifest.params,
-    inputs,
-    params.background,
-    params.text,
-    params.loadDate,
-  ]);
+export function safeGetRefValue<T>(ref: RefObject<T>): T {
+  if (!ref.current) {
+    throw new Error("Ref objects doesn't have a current value");
```
Justkant, line 65: Is it present only for development purposes ?

apps/ledger-live-mobile/src/components/Web3AppWebview/helpers.ts
```diff
@@ -0,0 +1,178 @@
...
+      setState({
+        title: nativeEvent.title,
+        url: nativeEvent.url,
+        canGoBack: nativeEvent.canGoBack,
+        canGoForward: nativeEvent.canGoForward,
+        loading: nativeEvent.loading,
+      });
+    },
+    [],
+  );
```
Justkant, line 141: We can probably define only 1 function and reuse it for every `useCallback` because they are all the same

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
```diff
@@ -0,0 +1,225 @@
...
+      loading: webview.isLoading(),
+    });
+  }, [webviewRef]);
+
+  useEffect(() => {
+    const webview = webviewRef.current;
+
+    if (!isMounted || !webview) {
+      return;
+    }
```
Justkant, line 188: Not a big fan of this `isMounted` solution to wait and do a second render.
I suppose we can start with that and look for something better later.

Justkant (approved):
Really nice, I still need to test it locally
Also one concern I have here is that we'll have to be careful whenever we fix a bug in one implementation to check the other ones for fixes also

libs/ledger-live-common/src/wallet-api/react.ts
IAmMorrow: Nope, it's mostly to keep typescript happy. This should never happen

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/index.tsx
IAmMorrow: What did you assume i'm expecting here ? 😂 

For the events, at this point I don't even know if those events are being used. We'll need to re-evaluate because we are tracking some very random stuff and sometime I feel we are trying to do sentry's job.

For this reason I took the decision to let it as-is

apps/ledger-live-mobile/src/components/Web3AppWebview/helpers.ts
IAmMorrow: True, and I thought about it.

But then I arbitrarily had to use one of the types for every single function and it felt wrong 😂 
Feels more readable like that since there aren't so many :)

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
IAmMorrow: Agreed

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/index.tsx
Justkant: Okay because if we actually want that component to work we'll have to use a const and probably use fragment to wrap the webview using the const in the fragment too

Justkant (approved):

PR #2917, opened 2023-03-12
feat: introduce the first set of destructors on debug menu

juan-cortes:
### 📝 Description
During a recent app refactoring, I encountered a challenge with resetting the application's state. It required either uninstalling the application or manually resetting the settings to their expected values, which was a tedious and time-consuming process. However, by including destructors in the code, we can streamline this process, benefiting both development and quality assurance teams. Destructors allow for the automatic reset of an application's state, reducing the need for manual intervention and saving time during testing cycles. Thus, utilizing destructors can improve the efficiency and accuracy of debugging tools for both development and QA purposes.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6732` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
https://user-images.githubusercontent.com/4631227/224574727-2f8fbb07-4622-4e61-beee-bef74bcb89c4.mov

### 🚀 Expectations to reach
- You should now be able to wipe the accounts, the settings, the known devices in a few clicks without uninstalling. More destructors could be created if needed, we are not limited by the normal destruction.

### DIFF
--- a/apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/index.tsx
@@ -3,8 +3,7 @@ import config from "react-native-config";
 import { getEnv } from "@ledgerhq/live-common/env";
 import { Alert as Confirmation } from "react-native";
 import { Alert, Icons } from "@ledgerhq/native-ui";
-import { useDispatch, useSelector } from "react-redux";
-import { State } from "../../../../reducers/types";
+import { useDispatch } from "react-redux";
 import GenerateMockAccounts from "./GenerateMockAccounts";
 import GenerateMockAccountsNft from "./GenerateMockAccountsNFTs";
 import ImportBridgeStreamData from "./ImportBridgeStreamData";
@@ -21,7 +20,6 @@ import { INITIAL_STATE as INITIAL_ACCOUNTS_STATE } from "../../../../reducers/ac
 import { INITIAL_STATE as INITIAL_BLE_STATE } from "../../../../reducers/ble";
 
 export default function Generators() {
-  const state = useSelector<State, State>(s => s);
   const dispatch = useDispatch();
   const reboot = useReboot();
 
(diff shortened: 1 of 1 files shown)


apps/ledger-live-mobile/src/components/RootNavigator/SettingsNavigator.tsx
```diff
@@ -221,7 +221,7 @@ export default function SettingsNavigator() {
         name={ScreenName.DebugGenerators}
         component={DebugGenerators}
         options={{
-          title: "Generators",
+          title: "Generators and Destructors",
```
ofreyssinet-ledger, line 224: https://www.youtube.com/watch?v=lb13ynu3Iac

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/index.tsx
```diff
@@ -29,6 +102,34 @@ export default function Generators() {
...
+      />
+      <SettingsRow
+        title="Refresh?"
+        desc="You may want to reload the app after wiping data"
+        iconLeft={<Icons.RefreshMedium size={24} color="black" />}
+        onPress={onForceRefresh}
+      />
+      <SettingsRow
+        title="Settings"
+        desc="Restores all settings to their default values"
```
ofreyssinet-ledger, line 117: Maybe use imperative here "Restore all settings" like in other descriptions

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/index.tsx
```diff
@@ -1,15 +1,88 @@
...
+        cancelable: true,
+      },
+    );
+  };
+
+  const onWipeSettings = useCallback(() => {
+    onCallbackWithConfirmation(() => {
+      dispatch(
+        dangerouslyOverrideState({
+          ...state,
```
ofreyssinet-ledger, line 53: do you actually have to pass the current state here? 🤔
the reducers of this action already do that in a safer way

ofreyssinet-ledger (approved):
great addition!

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/index.tsx
juan-cortes: Suggest the edit here and we can merge it

apps/ledger-live-mobile/src/screens/Settings/Debug/Generators/index.tsx
juan-cortes: Honestly I'm not sure, it's `dangerouslyOverride` why do it safer?

juan-cortes: but you are probably right, passing a partial state should be enough, in the debug state screen I passed the whole thing to not have to keep track of what I modified, but here it could be just passing in the settings/accounts/ble and it _should_ work

ofreyssinet-ledger: I think this is just a bad practice with redux, especially if the reducers already handle this. What I meant by "safer" is in terms of synchronisation with the current state of the store. The current state of the store should be handle at the reducer level to avoid race conditions.
Changing this will also be better in terms of perf because right now with the `useSelector(s => s)` this screen is getting unnecessarily re-rendered at every single dispatch of an action that modifies the store.

juan-cortes: Good point, I'll change it 👍🏼

ofreyssinet-ledger (approved):
so_good

PR #2928, opened 2023-03-13
fix: handle failed cases on genuine check - sync onboarding

juan-cortes:
### 📝 Description

Errors thrown during the genuine check step of the sync onboarding were only shown as skipped step without any more information. This PR introduces the failed state that shows the error, allows a retry, and a skip. As well as exporting the logs for further debugging.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6553` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

image

https://user-images.githubusercontent.com/4631227/224707877-1a07df21-2134-45ec-85b8-9225037fd789.mov

### 🚀 Expectations to reach
- Failed genuine checks (either by the device not being genuine or any other failure case) should prompt a modal instead of skipping automatically the step.


apps/ledger-live-mobile/src/screens/SyncOnboarding/SoftwareChecksStep.tsx
```diff
@@ -210,12 +203,14 @@ const SoftwareChecksStep = ({ device, isDisplayed, onComplete }: Props) => {
         } else if (devicePermissionState === "refused") {
           nextDrawerToDisplay = "cancelled";
         }
+      } else if (genuineCheckStatus === "failed") {
+        nextDrawerToDisplay = "cancelled";
```
alexandremgo, line 207: Should we rename the value `"cancelled"` to `"failed"` in `GenuineCheckUiDrawerStatus` ? I don't think `cancelled` makes sense anymore - the other time this value is used is just above on:

```
  } else if (devicePermissionState === "refused") {
          nextDrawerToDisplay = "cancelled";
        }
```

and if it was `"failed"` it would still make sense

apps/ledger-live-mobile/src/screens/SyncOnboarding/SoftwareChecksStep.tsx
```diff
@@ -376,7 +372,7 @@ const SoftwareChecksStep = ({ device, isDisplayed, onComplete }: Props) => {
                 button: "check if hardware genuine later",
                 drawer: "Failed Stax hardware check",
               });
-              setGenuineCheckStatus("failed");
+              setGenuineCheckStatus("skipped");
```
alexandremgo, line 375: Just above on

```
          <UnlockDeviceDrawer
            isOpen={nextDrawerToDisplay === "unlock-needed"}
            onClose={() => {
              // Closing because the user pressed on close button, and the genuine check is ongoing
              if (genuineCheckStatus === "ongoing") {
                // Fails the genuine check entirely
                setGenuineCheckStatus("failed");
              }
``` 

the `setGenuineCheckStatus("failed");` needs to be set to `skipped` too <- it's if the user does not unlock their device and close the drawer

apps/ledger-live-mobile/src/screens/SyncOnboarding/SoftwareChecksStep.tsx
```diff
@@ -190,15 +187,11 @@ const SoftwareChecksStep = ({ device, isDisplayed, onComplete }: Props) => {
...
-            "Failed to perform genuine check with error:",
-            genuineCheckError.message,
-          );
           setGenuineCheckStatus("failed");
         } else if (genuineState === "genuine") {
           setGenuineCheckStatus("completed");
         } else if (genuineState === "non-genuine") {
-          // FIXME: if the device is non-genuine, we should display something else
+          nextDrawerToDisplay = "cancelled";
           setGenuineCheckStatus("failed");
```
alexandremgo, line 195: You can remove the case `"failed"` now for the `GenuineCheckStatus`, it is not needed as now we have the `"skipped"` value.
And so you can remove the `else if` check where i've put a 🧹 in the PR comment

alexandremgo, line 195: Ok actually: (i removed some comments i've put): we are keeping this `setGenuineCheckStatus("failed");` here, but can you remove the `nextDrawerToDisplay = "cancelled";` you've added ?

Here we want to only update the `genuineCheckStatus` (to keep a bit of organization in the code), see the comments just above the first `if` of this block

The `nextDrawerToDisplay` is set correctly just after where you added:
```
} else if (genuineCheckStatus === "failed") {
        nextDrawerToDisplay = "cancelled";
```

apps/ledger-live-mobile/src/screens/SyncOnboarding/GenuineCheckFailedDrawer.tsx
```diff
@@ -0,0 +1,105 @@
...
+export type Props = {
+  isOpen: boolean;
+  onRetry?: () => void;
+  onSkip?: () => void;
+  onClose?: () => void;
+  productName: string;
+  error: Error | null | undefined;
+};
+
+const GenuineCheckFailedDrawer = ({
```
alexandremgo, line 18: Maybe add a comment to explain that there is 2 cases: an `error` is provided and no `error` is provided

apps/ledger-live-mobile/src/screens/SyncOnboarding/GenuineCheckFailedDrawer.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { TrackScreen } from "../../analytics";
+import GenericErrorView from "../../components/GenericErrorView";
+
+export type Props = {
+  isOpen: boolean;
+  onRetry?: () => void;
+  onSkip?: () => void;
+  onClose?: () => void;
+  productName: string;
+  error: Error | null | undefined;
```
alexandremgo, line 15: could just be `error?: Error | null` to keep consistency with above 🙏

apps/ledger-live-mobile/src/screens/SyncOnboarding/SoftwareChecksStep.tsx
alexandremgo: or actually we can keep it to `failed`, so the drawer you added is displayed ! It's actually better this way

alexandremgo (approved):

PR #2933, opened 2023-03-13
feat: lld/earn deep links

cjordan-ledger:
### 📝 Description

Adds deep link integration for new earn live-app. The following deep links are now possible to access, and are used by the earn dashboard to launch LLD functionality:
1. `ledgerlive://earn?action=get-funds&currencyId=ethereum` - this will open the Get Funds modal in LLD
2. `ledgerlive://earn?action=stake` - this will launch the staking flow in LLD

Related: https://github.com/LedgerHQ/earn-live-app/pull/20/files#diff-be7f625fb703c201664dc12705ca870b8e26a253fa1992616c1cf70cff54f65bR48

We are passing all query params from the deep link to the earn screen so we can easily add more query params in future without changing `useDeepLinkHandler`.

### ❓ Context

- **Impacted projects**:  desktop
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-5945 / https://ledgerhq.atlassian.net/browse/LIVE-5949

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/screens/stake/index.tsx
```diff
@@ -12,58 +12,63 @@ import { openModal } from "~/renderer/actions/modals";
...
+              ...stakeDefaultTrack,
+              button: "asset",
+              page: history.location.pathname,
+              currency: account?.currency?.family,
+              account,
+              parentAccount,
+              drawer: "Select Account And Currency Drawer",
+            });
+            setDrawer();
+            dispatch(openModal("MODAL_START_STAKE", { account, parentAccount, alwaysShowNoFunds }));
```
sarneijim, line 50: We could dispatch here directly to MODAL_NO_FUNDS_STAKE, avoiding move that logic too deep (MODAL_START_STAKE)
```suggestion
            dispatch(openModal("MODAL_NO_FUNDS_STAKE", { account, parentAccount }));
```
On the other hand, what do you think if instead use "alwaysShowNoFunds" we use "mode" (Thinking if tomorrow we need to create more modes....for instance "claimRewards"
```
mode: "stake"
mode: "noFunds"
```

apps/ledger-live-desktop/src/renderer/screens/earn/useDeepLinkListener.ts
```diff
@@ -0,0 +1,60 @@
...
+        shouldRedirect: false,
+        currencies: [decodedQueryParams.currencyId],
+        alwaysShowNoFunds: true,
+      });
+    }
+
+    // reset the query params so we don't trigger the effect again
+    history.replace({
+      search: "",
+    });
```
sarneijim, line 34: Not a big fan of this approach, although without changing the architecture I do not find anything better

apps/ledger-live-desktop/src/renderer/modals/StartStake/index.tsx
```diff
@@ -30,12 +31,17 @@ const ModalStartStake: FC<ModalStartStakeProps> = ({ account, parentAccount }) =
     const stakeAction = manageList.find(item => item.key === "Stake");
     dispatch(closeModal("MODAL_START_STAKE"));
 
+    if (alwaysShowNoFunds) {
+      dispatch(openModal("MODAL_NO_FUNDS_STAKE", { account, parentAccount }));
+      return;
+    }
```
sarneijim, line 37: I think move this logic to the drawer will be better

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.js
```diff
@@ -43,10 +43,12 @@ export function useDeepLinkHandler() {
 
   const navigate = useCallback(
     (url: string, state?: any, search?: string) => {
-      if (url !== location.pathname) {
-        setTrackingSource("deeplink");
-        history.push({ pathname: url, state, search });
+      if (url === location.pathname && !search) {
```
sarneijim, line 46: Can you add here a comment explaining this logic? (we need to check that we do not break any existing deeplink)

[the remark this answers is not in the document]
cjordan-ledger: Currently, we are not navigating to the deep link if the pathname doesn't change. Therefore if we just change the query params, then the screen component never updates.

So I'm changing it to say that if the URL hasn't changed AND there's no search provided, then don't do the navigation... otherwise we can navigate.

apps/ledger-live-desktop/src/renderer/screens/earn/useDeepLinkListener.ts
```diff
@@ -2,52 +2,28 @@ import useStakeFlow from "~/renderer/screens/stake";
 import { useHistory, useLocation } from "react-router-dom";
 import { useEffect } from "react";
 
-type QueryParamAction =
```
sarneijim: Can we add the valid url links as a comment?

apps/ledger-live-desktop/src/renderer/screens/earn/useDeepLinkListener.ts
cjordan-ledger: Yep, not ideal.. but without changing the messaging from webview to parent this is the least intrusive option.

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.js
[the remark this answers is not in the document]
cjordan-ledger: I've changed so it now reads as:

```
      if (url === location.pathname && search === location.search) {
        return;
      }
```

Hopefully this is a bit clearer with what's going on now?

brunospolidorio (approved):
LGTM

ggilchrist-ledger:
If we do the `get-funds` flow it looks like this copy will be there which is slightly misleading (what if they do have funds?). Do you know if this has been considered? @cjordan-ledger

cjordan-ledger:
> If we do the `get-funds` flow it looks like this copy will be there which is slightly misleading (what if they do have funds?). Do you know if this has been considered? @cjordan-ledger image

@ggilchrist-ledger agreed this copy isn't quite right... the idea is to use this modal for now to get something working for demo 1, and in a later iteration we can change the copy

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.js
```diff
@@ -42,18 +42,20 @@ export function useDeepLinkHandler() {
   const walletConnectLiveApp = useFeature("walletConnectLiveApp");
 
   const navigate = useCallback(
-    (url: string, state?: any, search?: string) => {
-      if (url !== location.pathname) {
+    (pathname: string, state?: any, search?: string) => {
+      const hasPathnameChanged = pathname !== location.pathname;
+      const hasSearchChanged = typeof search === "string" && search !== location.search;
```
sarneijim, line 47: Could it be ```hasNewSearch``` && ```hasNewPathname``` a better name as it is not rewriting the history when "clean" the param?

sarneijim (approved):
I am not a big fan of polluting the routes bad always that you do not need "back" button...

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.js
cjordan-ledger: https://github.com/LedgerHQ/ledger-live/pull/2933/commits/5a99ebb5866a3547f8701ee28761feebde728127

PR #2952, opened 2023-03-15
Improve app version detection of Btc::changeImplIfNecessary

bigspider:
### 📝 Description

`changeImplIfNecessary` only switches to the legacy implementation for bitcoin during the Exchange flow. This works well enough in Ledger Live, but is not optimal for external integrations relying on `hw-app-btc`, which would break if using a recent `hw-app-btc` on an older app version.

This does not change the behaviour for any case where `BtcOld` was already instantiated, but it makes sure to use the legacy protocol for versions below 2.1.0 of `"Bitcoin"`/`"Bitcoin Test"`.

I also added legacy default for altcoins, ` "Bitcoin Legacy"` or `"Bitcoin Test Legacy"`; that's redundant within Ledger Live since the constructor of `Btc` would already use `BtcOld` thanks to the `currency` parameter, so it's only added to be defensive in this function (and possibly useful for external integrations).

### ❓ Context

- **Impacted projects**: external integrations using `hw-app-btc`.

### ✅ Checklist

- [ ] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes**


haammar-ledger:
@bigspider Just a lint issue spotted by the CI: https://github.com/LedgerHQ/ledger-live/actions/runs/4425537198/jobs/7760723706#step:9:1638

PR #2859, opened 2023-03-06
LLD: Fully sunset "internal commands" + remove tech debt & dead code

gre:
### 📝 Description

This PR ships another step in simplifying main/internal scripts and making internal process even more minimal (sunset of commands, only remains the transport).

**Clean up related to internal process**

- all commands are sunset. **internal process is now exclusively used for hardware wallet transport to HID.** (We will later look at possibility to do this in a web worker)
  - transportListen was added in the transport channel.
  - all the renderer usage of commands has been moved to using the JS code directly. this removes a lot of remaining serialisation/deserialisation usage.
- logger is entirely moving to renderer.
- serialization.ts is entirely removed because it is unused since #2721 
  - `AccountBridge#applyReconciliation` is therefore dropped.
- main.bundle.js was greatly optimized. **Moving from 20MB to 1MB**. This should optimize the app load time.
  - This was made possible by moving more logic to renderer side. Notably the logger logic.
  - add a test that guarantee we stay < 5mb in future.
  - renderer.bundle.js on the other side, is growing a bit (49.2 -> 53.7 mb), taking a bit more of the work of internal commands. managed to reduce a bit the bundle with the other code clean up documented here. but this will be a major focus in a near future.

**introduce Web Workers**

- @elbywan added the possibility to create web worker scripts.
- `secp256k1` Bitcoin elliptic curve is executed in a web worker (`publicKeyTweakAdd`). with some parallel web worker to allow even faster parallel derivations.

**general libraries and other code clean up**

- minor Electron upgrade 23.1.0 -> 23.1.3 to benefit latest chromium.
- remove 16 libraries that weren't even used!
- remove the HTTP 404 fetch effect on LLD which was the only direct dependent of axios + also was useless. we will schedule bigger rework of that remote config system
- remove EXPERIMENTAL_WS_EXPORT which was depending on "ip"
- remove HSMStatusBanner system which was never used
- remove EXPERIMENTAL_MARKET_INDICATOR_SETTINGS

### ❓ Context

- **Impacted projects**: `LLD` 
- **Linked resource(s)**: LIVE-5697 + LIVE-5698 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

no demo. negative coding 😆 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

[...]

PR #2959, opened 2023-03-16
Add LLD architecture documentation for IPCTransport and web workers

gre:
### 📝 Description

document even more the PR #2859 

### ❓ Context

- **Impacted projects**: `doc` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/IPCTransport.md
+++ b/apps/ledger-live-desktop/src/renderer/IPCTransport.md
@@ -1,0 +1,141 @@
+# IPC Transport
+
+IPCTransport is an implementation of a [Transport](https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/hw-transport) specific to Ledger Live Desktop.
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/IPCTransport.md
```diff
@@ -0,0 +1,141 @@
...
+  }
+  subject.next({ type: "exchange", apduHex: data.apduHex, requestId });
+};
+```
+
+Now, this part is interesting: we are actually expecting the transport to have been opened before, and kept in a `transports` map. This is what the transportOpen has done and it will plug all the events coming on the global subject to do effect on the transport.
+
+### `exchangeBulk` Observable example
+
+The exact same idea applies now for the Observable. The main difference is that we can then more than one event on an observable, and we also need to be able to unsubscribe to the Observable, so we need to have a way to send an event to the internal to unsubscribe.
```
alexandremgo, line 82: `The main difference is that we can then **receive** more than one event on an observable` <- i think it's missing a "receive" here

alexandremgo (approved):
🔥 thanks a lot for the documentation 🙏

PR #2960, opened 2023-03-16
bugfix(usb-otg): LIVE-6771 multiple fixes to USB OTG on LLM

juan-cortes:
### 📝 Description

While addressing the original bug reported on https://ledgerhq.atlassian.net/browse/LIVE-6275 I encountered several issues with the USB OTG integration on LLM so while QA'ing this, please take into account the following bugs and topics:
- Duplicated devices when plugging/removing on the legacy device selector.
- Flashing device row on the select list when entering exiting apps in any flow.
- Entering an app breaks the device action on the new device selector (open app BTC, modal closes).
- Text overflow to the right on the empty state for USB on the legacy device selector screen.
- Disconnect during add accounts shows a "Device disconnected" error modal that can't be closed with the X.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6771` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach

Go over several flows with USB OTG on Android, while being in the wrong app, play with connecting and disconnecting and see if there's a bad behaviour somewhere. The disconnect event is currently debounced to two seconds in order to prevent the flashing so it's expected that after you unplug there's a delay on the device selector screen until you see the device disappear.

### DIFF
--- a/apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
+++ b/apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
@@ -101,12 +101,18 @@ export default function SelectDevice({
       if (e.type === "remove") setDevice(undefined);
       if (e.type === "add") {
         const { name, deviceModel, id, wired } = e;
-        setDevice({
-          deviceName: name,
-          modelId: deviceModel?.id,
-          deviceId: id,
-          wired,
-        } as Device);
+
+        setDevice((maybeDevice: Device | undefined) => {
+          return (
+            maybeDevice ||
+            ({
+              deviceName: name,
+              modelId: deviceModel?.id,
+              deviceId: id,
+              wired,
+            } as Device)
+          );
+        });
       }
     });
     return () => sub.unsubscribe();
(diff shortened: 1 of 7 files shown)


apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
```diff
@@ -101,12 +101,18 @@ export default function SelectDevice({
...
-        setDevice({
-          deviceName: name,
-          modelId: deviceModel?.id,
-          deviceId: id,
-          wired,
-        } as Device);
+
+        setDevice((maybeDevice: Device | undefined) => {
+          return (
+            maybeDevice ||
```
alexandremgo, line 107: Maybe just add a comment to explain why we're not replacing the (hid or httpdebug) device if one has already been set and it was not explictly removed (with a `remove` event)

alexandremgo (approved):
👨‍🔧🔥

PR #2969, opened 2023-03-16
[LIVE-5082] Feature: ENS / domain service for LLM

henrily-ledger:
### 📝 Description

Enable domain service for LLM, UI part only

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5082 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Screenshot 2023-03-16 at 16 11 03
Screenshot 2023-03-16 at 16 11 12
Screenshot 2023-03-16 at 16 11 20
Screenshot 2023-03-16 at 16 11 30

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/screens/SendFunds/02-SelectRecipient.tsx
+++ b/apps/ledger-live-mobile/src/screens/SendFunds/02-SelectRecipient.tsx
@@ -7,15 +7,15 @@ import {
 } from "@ledgerhq/live-common/bridge/react/index";
 import useBridgeTransaction from "@ledgerhq/live-common/bridge/useBridgeTransaction";
 import { isNftTransaction } from "@ledgerhq/live-common/nft/index";
-import Clipboard from "@react-native-community/clipboard";
 import { useTheme } from "@react-navigation/native";
 import invariant from "invariant";
 import React, { useCallback, useEffect, useRef, useState } from "react";
 import { Trans, useTranslation } from "react-i18next";
-import { Platform, StyleSheet, View } from "react-native";
+import { StyleSheet, View } from "react-native";
 import { SafeAreaView } from "react-native-safe-area-context";
 import Icon from "react-native-vector-icons/FontAwesome";
 import { useSelector } from "react-redux";
+import { DomainServiceProvider } from "@ledgerhq/domain-service/hooks/index";
 import { track, TrackScreen } from "../../analytics";
 import Alert from "../../components/Alert";
 import Button from "../../components/Button";
@@ -24,17 +24,15 @@ import GenericErrorBottomModal from "../../components/GenericErrorBottomModal";
 import KeyboardView from "../../components/KeyboardView";
 import LText from "../../components/LText";
 import NavigationScrollView from "../../components/NavigationScrollView";
-import RecipientInput from "../../components/RecipientInput";
 import RetryButton from "../../components/RetryButton";
 import {
   BaseComposite,
   StackNavigatorProps,
 } from "../../components/RootNavigator/types/helpers";
 import { SendFundsNavigatorStackParamList } from "../../components/RootNavigator/types/SendFundsNavigator";
-import SupportLinkError from "../../components/SupportLinkError";
-import TranslatedError from "../../components/TranslatedError";
 import { ScreenName } from "../../const";
 import { accountScreenSelector } from "../../reducers/accounts";
+import { RecipientRow } from "./RecipientRow";
 
 const withoutHiddenError = (error: Error): Error | null =>
   error instanceof RecipientRequired ? null : error;
@@ -58,6 +56,7 @@ export default function SendSelectRecipient({ navigation, route }: Props) {
   const shouldSkipAmount =
     transaction?.family === "ethereum" &&
     transaction?.mode === "erc721.transfer";
+  const [value, setValue] = useState<string>("");
 
   const isNftSend = isNftTransaction(transaction);
   // handle changes from camera qr code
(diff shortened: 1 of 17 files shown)


apps/ledger-live-mobile/src/screens/SendFunds/02-SelectRecipient.tsx
```diff
@@ -92,8 +91,9 @@ export default function SendSelectRecipient({ navigation, route }: Props) {
           recipient,
         }),
       );
+      setValue(recipient);
     },
-    [account, parentAccount, setTransaction, transaction],
+    [account, parentAccount, setTransaction, transaction, setValue],
```
lvndry, line 96: `setValue` as a dependency ? Why not `value` ?

libs/domain-service/src/resolvers/index.ts
```diff
@@ -8,6 +8,23 @@ import {
   getRegistriesForDomain,
 } from "../registries";
 
+Promise.allSettled =
```
lvndry, line 11: Why is it needed ?

henrily-ledger: allSettled doesn't exist in react-native env but only on node

apps/ledger-live-mobile/src/screens/SendFunds/02-SelectRecipient.tsx
lambertkevin: it's a `useCallback` so value is the argument of that method, no need to have it as dep

lambertkevin (changes requested):
All good but shouldn't be merge before #2671

lambertkevin (approved):

libs/domain-service/src/resolvers/index.ts
```diff
@@ -8,6 +9,8 @@ import {
   getRegistriesForDomain,
 } from "../registries";
 
+allSettled.shim(); // Makes it compatible with React Native when Promise.allSettled isn't avaiblable
```
gre, line 12: this is not to be done in the lib but in React Native side then.

lambertkevin: This lib will be used externally, so no it should be there as well to make sure any external wallet can use it independently of the platform.

apps/ledger-live-mobile/src/screens/SendFunds/DomainServiceRecipientRow.tsx
```diff
@@ -0,0 +1,181 @@
...
+            recipient: value,
+            recipientDomain: undefined,
+          }),
+        );
+      }
+      return;
+    }
+
+    // verify if domain is present in transaction
+    const domainSet = recipientDomain === ensResolution;
```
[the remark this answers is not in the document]
lambertkevin: - ENS is not "only" Ethereum, it could resolve every coin possible if configured the right way. It's not in this step though backend wise. 
- In our case, ENS is the only possible solution as it's the only one we *want* to support. Since the product hasn't started working on "what should happen if we had multiple resolutions/collisions", we can't predict what the code should look like when we don't have this very specific use case of ENS. I think we should update the names only once we have multiple domain providers and we have a product decision on what should happen in those cases. 
- `recipientDomain` is generic as we could very well implement it on all chains with different providers. E.g.:
  - Cosmos chains -> Stargaze names, DENS.sh, Starname. 
  - EVM -> Lens Protocol.
  - All chains -> ENS, Unstoppable Domains

libs/domain-service/src/resolvers/index.ts
gre: a library must not have a global shim side effect. it is either on user land to fix OR just don't use allSettled in this library.

apps/ledger-live-mobile/src/screens/SendFunds/DomainServiceRecipientRow.tsx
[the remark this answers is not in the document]
gre: by Ethereum I meant, ethereum as a family, are you saying the wording will remain as "ens" even for other blockchains?

libs/domain-service/src/resolvers/index.ts
gre: it's for the end projects to deal with this, typically if we upgrade React Native we may not need it anymore and therefore it kinda belongs to the polyfills the same way we didn't polyfill live-common "in place" just because React Native needed tons of polyfill to make it run.

apps/ledger-live-mobile/src/screens/SendFunds/DomainServiceRecipientRow.tsx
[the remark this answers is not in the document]
lambertkevin: Yes it would, you need to see ENS as a provider and its database is the Ethereum blockchain, but it can resolve anything (again, in theory, right now it's only resolving EVM addresses by backend design). 
So you could type "greweb.eth" and resolve a BTC address for example.

That's why the domain-service is an external lib, as technically most of them are not directly linked to a single family. But as said before I guess the addition of new providers will tell us how we'll have to make that thing evolve 🤞

libs/domain-service/src/resolvers/index.ts
lambertkevin: Done, shim moved to LLM + Error thrown in domain-service lib

PR #2977, opened 2023-03-17
feat(Firmware Update LLM): Implement new firmware update UX on LLM

grsoares21:
### 📝 Description
This PR implements a new User Experience for the firmware update feature on LLM.
It uses the newly created "Firmware Update" device action from the device SDK in order to implement this UX. I also enables Firmware Updates via BLE for Stax devices.

This feature is being implemented in two parts: first the update only (this PR) and then the back-up and restore of apps and settings from the update. So this PR is not, in fact, standalone. This isn't a problem because the developments being done here are under the feature flag `fxUpdateNewUx`, we should however make sure that there are no regressions when the feature flag is disabled.

### ❓ Context

- **Linked resource(s)**: [LIVE-6286], [LIVE-6285]

### ✅ Checklist

- [ ] **Test coverage** Not yet, since Device Action related Detox Tests are still in the early stages.
- [ ] **Atomic delivery** - As explained in the description, this feature is not complete without the back up and restore of device settings, however this is under feature flag, so it shouldn't impact the user while the feature flag is disabled
- [x] **No breaking changes** 

### 📸 Demo
https://user-images.githubusercontent.com/6013294/225872933-70f957a3-fa96-4a1d-84e9-26f59948e5a6.mp4

### Test plan
We should be able to update the firmware via BLE or USB for Stax devices in any of the directions il0 <-> il1 / il2 / il3. 
We should be able to update the firmware only via USB for other devices (S, S+, X), beware that Nano X il firmwares are seen as Stax devices, and it might be possible to try the update via BLE in this case and it probably won't work, this isn't a bug in the feature.
No regressions on the firmware update feature should exist if the feature flag is disabled. Some of the changes in this PR are not under feature flag (e.g. the design of the firmware update banner), but this shouldn't create any regressions on the firmware update feature itself.

[LIVE-6286]: https://ledgerhq.atlassian.net/browse/LIVE-6286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-6285]: https://ledgerhq.atlassian.net/browse/LIVE-6285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/components/FirmwareUpdateBanner.tsx
```diff
@@ -40,15 +49,81 @@ const FirmwareUpdateBanner = ({
...
+    if (lastConnectedDevice?.modelId === DeviceModelId.stax) {
+      const sub = concat(
+        withDevice(lastConnectedDevice.deviceId)(transport =>
+          from(
+            getBatteryStatus(transport, BatteryStatusTypes.BATTERY_PERCENTAGE),
+          ),
+        ),
+        withDevice(lastConnectedDevice.deviceId)(transport =>
+          from(getBatteryStatus(transport, BatteryStatusTypes.BATTERY_FLAGS)),
+        ),
```
juan-cortes, line 68: Considering we will most likely need the tuple of information `battery %, isCharging` you should extend the command to accommodate this. Make it do the two apdus, get the combined information in one line and not this mess.

juan-cortes (changes requested):

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -1190,3 +1191,251 @@ export const renderImageCommitRequested = ({
     />
   );
 };
+
+// NEW DEVICE ACTION UI
+// area reserved to the project of redefining the UX of the device actions
+// new renderings should be define here in order to progressively move towards then new UX
+
```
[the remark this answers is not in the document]
grsoares21: Done

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -1190,3 +1191,251 @@ export const renderImageCommitRequested = ({
...
+          ? firmwareUpdateErrorTitle
+          : genericErrorTitle,
+      errorDescription:
+        firmwareUpdateErrorDescription !== firmwareUpdateErrorDescriptionId
+          ? firmwareUpdateErrorDescription
+          : genericErrorDescription !== genericErrorDescriptionId
+          ? genericErrorDescription
+          : null,
+    };
+  }, [errorName, t]);
```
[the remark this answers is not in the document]
grsoares21: I've made this generic by adding an optional context. I'm not sure this is the same logic that exists in the `TranslatedError` component though, it might make it even harder to understand the logic of `TranslatedError` to accept my use case.

apps/ledger-live-mobile/src/components/FirmwareUpdateBanner.tsx
```diff
@@ -40,15 +49,81 @@ const FirmwareUpdateBanner = ({
...
+              charging: batteryFlags.charging === 1,
+            });
+          },
+          error: err =>
+            log(
+              "firmwareUpdateBanner",
+              "Error while retrieving stax battery:",
+              err,
+            ),
+        });
```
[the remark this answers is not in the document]
grsoares21: Done

apps/ledger-live-mobile/src/components/FirmwareUpdateBanner.tsx
```diff
@@ -88,36 +165,74 @@ const FirmwareUpdateBanner = ({
...
+        </Flex>
+        <InvertTheme>
+          <Flex flexDirection="row" flex={1}>
+            <Button
+              flex={1}
+              outline
+              event="FirmwareUpdateBannerClick"
+              type="main"
+              title={t("common.learnMore")}
+              onPress={() => console.log("TODO: implement learn more click")}
```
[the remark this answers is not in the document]
grsoares21: I'm missing some product input for this. I was thinking of doing it in a supplementary PR with product feedback if I don't get the input for this in time. wdyt?

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -1190,3 +1191,251 @@ export const renderImageCommitRequested = ({
...
+
+  return (
+    <Flex alignItems="center" justifyContent="center" px={1}>
+      <IconBadge
+        iconColor="warning.c100"
+        iconSize={32}
+        Icon={Icons.WarningSolidMedium}
+      />
+      <Text fontSize={24} fontWeight="semiBold" textAlign="center" mt={6}>
```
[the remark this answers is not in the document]
grsoares21: Done

libs/ledger-live-common/src/hw/getBatteryStatus.ts
```diff
@@ -22,38 +22,55 @@ enum FlagMasks {
   ISSUE_TEMPERATURE = 0x00000020,
 }
 
-const getBatteryStatus = async (
+// conditional type used to infer what will be the tuple of returned values of the getBatteryStatus
+// command. According to the tuple that of StatusTypes that is requested, it will infer the corresponding
```
alexandremgo, line 26: `According to the tuple of BatteryStatusTypes that is requested` no ?

libs/ledger-live-common/src/hw/getBatteryStatus.ts
```diff
@@ -22,38 +22,55 @@ enum FlagMasks {
...
+// tupple of either number or BatteryStatusFlags for the result
+type BatteryStatusTuple<Statuses extends ReadonlyArray<BatteryStatusTypes>> = {
+  [index in keyof Statuses]: Statuses[index] extends BatteryStatusTypes.BATTERY_FLAGS
+    ? BatteryStatusFlags
+    : number;
+};
+
+const getBatteryStatus = async <
+  StatusesType extends ReadonlyArray<BatteryStatusTypes>
+>(
```
alexandremgo, line 36: I think the associated unit tests are failing

```
pnpm common jest libs/ledger-live-common/src/hw/getBatteryStatus.test.ts
```

apps/ledger-live-mobile/src/screens/Settings/Debug/Features/FirmwareUpdate.tsx
```diff
@@ -1,48 +1,286 @@
...
+  version: "2.0.2-il0",
+};
+
+const mockDevice: Device = {
+  deviceId: "it doesn't matter",
+  modelId: DeviceModelId.stax,
+  wired: false,
+};
 
 export default function DebugFirmwareUpdate() {
```
alexandremgo, line 87: It's really useful ! Nice !

We do agree that's it's only to test the UI components of the firmware update ? If so, update also the `<SettingRow desc="..">` of
`apps/ledger-live-mobile/src/screens/Settings/Debug/Features/index.tsx` because it's saying `"USE AT YOUR OWN RISK!"` ahaha

apps/ledger-live-mobile/src/components/RootNavigator/ManagerNavigator.tsx
```diff
@@ -59,6 +60,16 @@ export default function ManagerNavigator() {
...
       }}
     >
+      <Stack.Screen
+        name={ScreenName.FirmwareUpdate}
+        component={FirmwareUpdateScreen}
+        options={{
+          gestureEnabled: false,
+          headerTitle: () => null,
+          headerLeft: () => null,
+          headerRight: () => <Button Icon={Icons.CloseMedium} />,
```
alexandremgo, line 70: Settings the header here works well with the error/device actions (we can always access the close button) ? Just asking, i think it's better to set the closing/back arrow in the react-navigation header

Though: does this button do anything ? I don't see any navigation back etc.

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -0,0 +1,446 @@
...
+        <CloseWarning
+          onPressContinue={() => setIsCloseWarningOpen(false)}
+          onPressQuit={quitUpdate}
+        />
+      </QueuedDrawer>
+    </>
+  );
+};
+
+const FirmwareUpdateScreen = () => {
```
alexandremgo, line 427: But wait: so we're removing directly all the current/existing version of the firmware update screen/component ? We don't use this version with a feature flag for now ?

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -0,0 +1,446 @@
...
+      ),
+    });
+  }, [navigation, quitUpdate, isAllowedToClose]);
+
+  // this will depend on the steps we go through during the update
+  const [totalNumberOfSteps, setTotalNumberOfSteps] = useState(2);
+
+  useEffect(() => {
+    // TODO: instead of time out, trigger when all the backups have been made
+    triggerUpdate();
```
alexandremgo, line 229: No timeout nor waiting for backup for now ;)

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -0,0 +1,446 @@
...
+            t("FirmwareUpdate.steps.installUpdate.titleActive") +
+            ` ${totalNumberOfSteps}/${totalNumberOfSteps}`;
+        } else {
+          updateInstallStepProgress();
+          newSteps.installUpdate.title =
+            t("FirmwareUpdate.steps.installUpdate.titleActive") +
+            ` ${totalNumberOfSteps - 1}/${totalNumberOfSteps}`;
+        }
+        break;
+      // eslint-disable-next-line no-fallthrough
```
alexandremgo, line 300: Not needed ?

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -0,0 +1,446 @@
...
+        </Flex>
+        <Flex mx={6} mb={8}>
+          <Alert type="info" title={t("FirmwareUpdate.doNotLeaveLedgerLive")} />
+        </Flex>
+      </Flex>
+      <QueuedDrawer
+        isRequestingToBeOpened={Boolean(deviceInteractionDisplay)}
+        noCloseButton
+      >
+        {deviceInteractionDisplay}
```
alexandremgo, line 415: Only changing the content of the `QueuedDrawer` does not: add a new drawer to the queue, close the current drawer, display the new drawer - it just changes the content of the current displayed drawer

Is it rendering correctly/as you expect ?

As long as you don't enable a close button and do something on the `onClose` of the QueuedDrawer that would do different thing depending on the `updateState.step` you're good

apps/ledger-live-mobile/src/components/RootNavigator/ManagerNavigator.tsx
grsoares21: It is only set here too to avoid "flickering" (i.e. a brief moment where there's a back button and no close button that gets replaced, if we set here, the change in the header done in the component is in visible to the user). That's also the reason why there's a button with no action here, it is replaced inside the component later.

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
grsoares21: This is a very old screen that was never used actually, this isn't the old firmware update, it's dead code. The old firmware update is elsewhere.

libs/ledger-live-common/src/deviceSDK/tasks/updateFirmware.ts
```diff
@@ -75,44 +75,44 @@ function internalUpdateFirmwareTask({
...
-            }).pipe(
-              map((e) => {
-                if (e.type === "unresponsive") {
-                  return {
-                    type: "error" as const,
-                    error: new LockedDeviceError(),
-                  };
-                }
+            return subscriber.closed
+              ? EMPTY
```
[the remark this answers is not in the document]
grsoares21: It's not really about emitting an event but rather about preventing API calls to be made and APDUs to be sent in case the observable gets unsubscribed. It's a way to stop the side-effects of the observable.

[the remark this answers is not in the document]
grsoares21: It's there for when the unsubscribe happens exactly at the point we're already inside this function

apps/ledger-live-mobile/src/screens/Settings/Debug/Features/FirmwareUpdate.tsx
grsoares21: Hahaha initially it was the actual update, but then I've changed to be UI-only and forgot to update that description 🙈 . Updated it now 👌

libs/ledger-live-common/src/deviceSDK/actions/updateFirmware.ts
```diff
@@ -62,64 +62,67 @@ export const initialState: UpdateFirmwareActionState = {
...
+            return of(event);
+          } else {
+            return updateFirmwareTask({
+              deviceId,
+              updateContext: event.firmwareUpdateContext,
+            });
+          }
+        })
+      )
+      .pipe(
```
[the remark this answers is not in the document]
grsoares21: Good catch! Done!

libs/ledger-live-common/src/deviceSDK/hooks/useUpdateFirmware.ts
```diff
@@ -35,7 +35,7 @@ export const useUpdateFirmware = ({
     }
   }, [deviceId, updateFirmwareAction, nonce]);
 
-  const triggerUpdate = () => setNonce(nonce + 1);
+  const triggerUpdate = useCallback(() => setNonce((nonce) => nonce + 1), []);
```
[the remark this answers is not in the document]
grsoares21: good point. added some, wdyt?

libs/ledger-live-common/src/hw/getBatteryStatus.ts
grsoares21: yep, good catch, done.

libs/ledger-live-common/src/hw/getBatteryStatus.ts
grsoares21: Yeah, I had forgotten to adapt the unit tests to the changes I've made. Good catch. Done

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
grsoares21: Good catch, removed the comment

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -0,0 +1,446 @@
...
+      newSteps.prepareUpdate.status = VerticalStepper.ItemStatus.completed;
+      newSteps.installUpdate.status = VerticalStepper.ItemStatus.completed;
+      newSteps.restoreAppsAndSettings.status =
+        VerticalStepper.ItemStatus.active;
+    };
+
+    switch (updateState.step) {
+      case "installingOsu":
+        updatePrepareStepProgress();
+      // eslint-disable-next-line no-fallthrough
```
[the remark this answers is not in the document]
grsoares21: It was saving only 1 repeated line of code, so I actually removed it. I was easier to repeat the line than to explain the need for the fall through.

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
grsoares21: It's ok, there shouldn't be a queue here. We can review that if we end up needing a close button.

libs/ledger-live-common/src/deviceSDK/hooks/useUpdateFirmware.ts
```diff
@@ -7,26 +7,33 @@ import {
 
 export type UseUpdateFirmwareArgs = {
   updateFirmwareAction?: typeof defaultUpdateFirmwareAction;
-  deviceId?: string;
+  deviceId: string;
 };
 
+/**
+ * Hooke used to trigger the update of a firmware, it isn't triggered right away but rather returns
```
alexandremgo, line 14: hooke ❤️

libs/ledger-live-common/src/deviceSDK/hooks/useUpdateFirmware.ts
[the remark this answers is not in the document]
alexandremgo: Thanks :)

libs/ledger-live-common/src/deviceSDK/tasks/updateFirmware.ts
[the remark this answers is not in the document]
alexandremgo: Ok i actually agree with you, it's necessary !

I tested with:
```JS
import { Observable, interval, from } from "rxjs";
import { switchMap } from "rxjs/operators";

function asyncDelay(ms) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

const subscription = new Observable((subscriber) => {
  interval(2000)
    .pipe(
      switchMap((value) => {
        return from(
          (async () => {
            console.log(`Start switch: ${value}`);
            await asyncDelay(1000);
            console.log(`End switch: ${value}`);
            return Promise.resolve(value);
          })()
        );
      })
    )
    .subscribe(subscriber);
}).subscribe({
  next: (value) => {
    console.log(`➡️ next: ${value}`);
    setTimeout(() => {
      console.log(`🚨 unsubscribing from ${value}`);
      subscription.unsubscribe();
    }, 1001);
  },
  complete: () => console.log(`completed`)
});
```

and got the following result:
```bash
Start switch: 0 
End switch: 0 
➡️ next: 0 
Start switch: 1 
🚨 unsubscribing from 0 
End switch: 1 
```

Showing that even if an unsubscription occurred, a block inside a `switchMap` is not stopped directly but is finished until the end (just nothing is emitted after)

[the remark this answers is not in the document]
grsoares21: It's really cool that you did the test, that's awesome! Gonna save it for future refrence hahaha

apps/ledger-live-mobile/src/components/FirmwareUpdateBanner.tsx
```diff
@@ -250,7 +250,14 @@ const FirmwareUpdateBanner = () => {
...
             outline={false}
-            onPress={onExperimentalFirmwareUpdate}
+            onPress={
+              newFwUpdateUxFeatureFlag?.enabled
+                ? onExperimentalFirmwareUpdate
+                : () => {
+                  setShowBatteryWarningDrawer(false);
+                  setShowUnsupportedUpdateDrawer(true);
+                }
+            }
```
juan-cortes, line 260: I would personally move this to a callback with dependencies on the enabled flag and other function, but I don't care enough anymore

grsoares21: Done

libs/ledger-live-common/src/featureFlags/groupedFeatures.ts
```diff
@@ -11,11 +11,11 @@ export const groupedFeatures: Record<
     featureIds: [
       "customImage",
       "deviceInitialApps",
-      "llmNewDeviceSelection",
```
abdurrahman-ledger, line 14: Why did we removed this ?

juan-cortes (approved):

PR #2978, opened 2023-03-17
Feat/lld/add common stake flow 

sarneijim:
### 📝 Description

Common stake flow integrated in the main menu

Ghost page which use select account flow to redirect to stake programs

### ❓ Context

- **Impacted projects**: LLM
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-5229

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/38540290/225891198-f493cd90-b987-4e9e-b5fc-dbd90d36146e.mov

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/Stake/index.tsx
+++ b/apps/ledger-live-mobile/src/components/Stake/index.tsx
@@ -15,7 +15,8 @@ import logger from "../../logger";
 const StakeFlow = () => {
   const featureFlag = useFeature("stakePrograms");
   const list = featureFlag?.params?.list;
-  const navigation = useNavigation();
+  const navigation =
+    useNavigation<StackNavigationProp<{ [key: string]: object | undefined }>>();
   const cryptoCurrencies = useMemo(() => {
     return filterCurrencies(listCurrencies(true), {
       currencies: list || [],
@@ -55,11 +56,7 @@ const StakeFlow = () => {
           },
         });
       } else {
-        (
-          navigation as StackNavigationProp<{
-            [key: string]: object | undefined;
-          }>
-        ).navigate(NavigatorName.Base, {
+        navigation.navigate(NavigatorName.Base, {
           screen: name,
           params: {
             screen: options.screen,
@@ -79,9 +76,7 @@ const StakeFlow = () => {
   };
 
   const requestAccount = useCallback(() => {
-    (
-      navigation as StackNavigationProp<{ [key: string]: object | undefined }>
-    ).replace(NavigatorName.RequestAccount, {
+    navigation.replace(NavigatorName.RequestAccount, {
       screen: ScreenName.RequestAccountsSelectCrypto,
       params: {
         currencies: cryptoCurrencies,
--- a/apps/ledger-live-mobile/src/components/Stake/index.tsx
+++ b/apps/ledger-live-mobile/src/components/Stake/index.tsx
@@ -1,0 +1,92 @@
+import { useMemo, useLayoutEffect, useCallback } from "react";
+import { useNavigation } from "@react-navigation/native";
+import { Account, AccountLike } from "@ledgerhq/types-live";
+import {
+  listCurrencies,
+  filterCurrencies,
+} from "@ledgerhq/live-common/currencies/helpers";
+import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
+import { NavigatorName, ScreenName } from "../../const";
+import perFamilyAccountActions from "../../generated/accountActions";
+
+const StakeFlow = { currencies } => {
+  const { params: { list } = { list: undefined } } =
+    useFeature("stakePrograms") || {};
+  const navigation = useNavigation();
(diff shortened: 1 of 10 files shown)


apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+import {
+  listCurrencies,
+  filterCurrencies,
+} from "@ledgerhq/live-common/currencies/helpers";
+import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
+import { NavigatorName, ScreenName } from "../../const";
+import perFamilyAccountActions from "../../generated/accountActions";
+
+const StakeFlow = ({ currencies }) => {
+  const { params: { list } = { list: undefined } } =
```
cjordan-ledger, line 13: Up to you but this seems a bit simpler without the destructuring:
```
const featureFlag = useFeature("stakePrograms")
const list = featureFlag?.params?.list;
```

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+            },
+          },
+        });
+      }
+    },
+    [navigation],
+  );
+  const onError = error => {
+    // eslint-disable-next-line no-console
+    console.log("RequestAccountsSelectCrypto in Stake Flow", error);
```
cjordan-ledger, line 74: should we log an error to sentry here?

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+          decorators.getMainActions({
+            account,
+            parentAccount,
+            colors: {},
+          })) ||
+        [];
+      const stakeFlow = familySpecificMainActions?.find(
+        action => action.id === "stake",
+      )?.navigationParams;
+      const [name, options] = stakeFlow;
```
cjordan-ledger, line 37: `stakeFlow` could be undefined here - what should happen if we can't find it? early return maybe?

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+      const familySpecificMainActions =
+        (decorators &&
+          decorators.getMainActions &&
+          decorators.getMainActions({
+            account,
+            parentAccount,
+            colors: {},
+          })) ||
+        [];
+      const stakeFlow = familySpecificMainActions?.find(
```
cjordan-ledger, line 34: minor: i don't think you need the null check here as you default to an empty array on line 33, does this work? `familySpecificMainActions.find(`

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+        action => action.id === "stake",
+      )?.navigationParams;
+      const [name, options] = stakeFlow;
+
+      // TODO: Remove after Kiln stake implementation was be done
+      if (
+        account?.currency?.family === "ethereum" &&
+        name === NavigatorName.Base
+      ) {
+        navigation.navigate(name, {
```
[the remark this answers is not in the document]
sarneijim: Now, when we click the stake button in eth we go directly to discovery app, we are not getting the same structure that we have in the other modals. When we implement Kiln, we will create a modal and then we will remove this if.

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+        navigation.navigate(NavigatorName.Base, {
+          screen: name,
+          params: {
+            screen: options.screen,
+            params: {
+              ...(options?.params || {}),
+              account,
+              parentAccount,
+              accountId: account?.id,
+              parentId: parentAccount?.id,
```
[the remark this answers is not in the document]
sarneijim: You are right, parentId and accountId is only needed for platform (eth)

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          screen: options.screen,
+          params: {
+            ...(options?.params || {}),
+            accountId: account?.id,
+            parentId: parentAccount?.id,
+          },
+        });
+      } else {
+        (
+          navigation as StackNavigationProp<{
```
cjordan-ledger, line 59: I think it's best to avoid casting `navigation`, it could hide some possible issues. I assume typescript is complaining about the params passed to `navigation.navigate`?

You might be able to do this instead:

```
const navigation = useNavigation<StackNavigationProp<{ [key: string]: object | undefined }>>();
```

cjordan-ledger (approved):

PR #3002, opened 2023-03-20
fix: LIVE-6749 Handle the device full case on language pack install

juan-cortes:
### 📝 Description

Currently if a device is full and we try to install a language pack the error is an unexpected response from the device, we had no reference to `5102` meaning the device was full in this context. I'm taking a guess that it is and mapping the error accordingly. Also tackles the https://ledgerhq.atlassian.net/browse/LIVE-6830 topic covering CLS on full device, same issue.

### ❓ Context

- **Impacted projects**: `ledger-live-common errors` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6749` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/4631227/226449251-c90ba04e-e7c5-4b9c-a0dd-9050a67859f1.mov

### 🚀 Expectations to reach
This can be tested on either LLD or LLM, on a full device, installing a language pack.

### DIFF
--- a/libs/ledger-live-common/src/hw/installLanguage.ts
+++ b/libs/ledger-live-common/src/hw/installLanguage.ts
@@ -5,6 +5,8 @@ import {
   TransportError,
   TransportStatusError,
   LanguageNotFound,
+  StatusCodes,
+  ManagerNotEnoughSpaceError,
 } from "@ledgerhq/errors";
 
 import ManagerAPI from "../api/Manager";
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/hw/installLanguage.ts
```diff
@@ -103,11 +105,13 @@ export default function installLanguage({
                 const statusStr = status.toString(16);
 
                 // Some error handling
-                if (status === 0x5501) {
+                if (status === StatusCodes.USER_REFUSED_ON_DEVICE) {
```
alexandremgo, line 108: thanks 🙏

alexandremgo (approved):
nice ! 🔥

PR #3026, opened 2023-03-22
chore: make exported logs formatted by default

juan-cortes:
### 📝 Description
Two minute chore, makes the exported logs readable by default so I can read them on my phone and also open them without having to worry about formatting them. I can also refer to a specific line without needing the other party to have formatted the log the same way my editor did. Longer to write this than implement it.

### ❓ Context

- **Impacted projects**: `dekstop, mobile` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6856` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
image

### 🚀 Expectations to reach
Merge it

### DIFF
--- a/.changeset/odd-kangaroos-give.md
+++ b/.changeset/odd-kangaroos-give.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
+Make exported logs formatted by default
--- a/apps/ledger-live-desktop/src/renderer/components/ExportLogsButton.jsx
+++ b/apps/ledger-live-desktop/src/renderer/components/ExportLogsButton.jsx
@@ -15,7 +15,11 @@ import type { Props as ButtonProps } from "~/renderer/components/Button";
 import { accountsSelector } from "~/renderer/reducers/accounts";
 
 const saveLogs = async (path: { canceled: boolean, filePath: string }) => {
-  await ipcRenderer.invoke("save-logs", path, JSON.stringify(memoryLogger.getMemoryLogs()));
+  await ipcRenderer.invoke(
+    "save-logs",
+    path,
+    JSON.stringify(memoryLogger.getMemoryLogs(), null, 4),
+  );
 };
 
 type RestProps = ButtonProps & {|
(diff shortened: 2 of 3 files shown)


apps/ledger-live-desktop/src/renderer/components/ExportLogsButton.jsx
```diff
@@ -15,7 +15,11 @@ import type { Props as ButtonProps } from "~/renderer/components/Button";
 import { accountsSelector } from "~/renderer/reducers/accounts";
 
 const saveLogs = async (path: { canceled: boolean, filePath: string }) => {
-  await ipcRenderer.invoke("save-logs", path, JSON.stringify(memoryLogger.getMemoryLogs()));
+  await ipcRenderer.invoke(
+    "save-logs",
+    path,
+    JSON.stringify(memoryLogger.getMemoryLogs(), null, 4),
```
Justkant, line 21: I like 2 spaces instead of 4, wdyt ?
```suggestion
    JSON.stringify(memoryLogger.getMemoryLogs(), null, 2),
```

apps/ledger-live-mobile/src/components/useExportLogs.tsx
```diff
@@ -8,7 +8,9 @@ export default function useExportLogs() {
   return useCallback(() => {
     const exportLogs = async () => {
       const logs = logReport.getLogs();
-      const base64 = Buffer.from(JSON.stringify(logs)).toString("base64");
+      const base64 = Buffer.from(JSON.stringify(logs, null, 4)).toString(
```
Justkant, line 11: Just here if you like my preference so that you can commit from UI
```suggestion
      const base64 = Buffer.from(JSON.stringify(logs, null, 2)).toString(
```

.changeset/odd-kangaroos-give.md
```diff
@@ -0,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"live-mobile": patch
```
alexandremgo, line 3: Just we should still separate LLM and LLD changeset i would say

Justkant (approved):
approved but would still like to know if we can make it 2 spaces

alexandremgo (approved):
:beautiful:

.changeset/odd-kangaroos-give.md
Justkant: We don't need to anymore as we version both at the same time now

ofreyssinet-ledger (approved):

valpinkman (approved):

PR #3035, opened 2023-03-23
feat(LLM/Recover): Upsell

mlegall-ledger:
### 📝 Description

New Recover upsell after onboarding (opening live-app), prevented to open if no internet connection. Removed unused component. Updated react network library.

### ❓ Context

- **Impacted projects**: LLM
- **Linked resource(s)**: [LIVE-5857]

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/39890664/227160992-573c49ee-2b70-4a8b-a1d1-b1fa63005901.mp4

### 🚀 Expectations to reach

[LIVE-5857]: https://ledgerhq.atlassian.net/browse/LIVE-5857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
```diff
@@ -58,9 +65,41 @@ function PortfolioScreen({ navigation }: NavigationProps) {
...
+  const lastConnectedDevice = useSelector(lastConnectedDeviceSelector);
   const [isAddModalOpened, setAddModalOpened] = useState(false);
   const { colors } = useTheme();
   const { isAWalletCardDisplayed } = useDynamicContent();
+  const protectFeature = useFeature("protectServicesMobile");
+  const postOnboardingURL =
+    protectFeature?.params?.onboardingRestore?.postOnboardingURI;
+  const dispatch = useDispatch();
+
+  const internetReachable = async () => {
```
ofreyssinet-ledger, line 78: can you extract this function and put it in `src/logic/*` or somewhere similar

apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
```diff
@@ -58,9 +65,41 @@ function PortfolioScreen({ navigation }: NavigationProps) {
...
+    // workaround : on iOS, netInfo returns null on the first call, see ; https://github.com/react-native-netinfo/react-native-netinfo/issues/572
+    let isInternetReachable = false;
+
+    isInternetReachable = !!(await NetInfo.fetch()).isInternetReachable;
+
+    if (isInternetReachable) {
+      return isInternetReachable;
+    }
```
ofreyssinet-ledger, line 86: if it returns false you still try it again, maybe just do that: if `result===null` retry, else return `result`

apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
```diff
@@ -58,9 +65,41 @@ function PortfolioScreen({ navigation }: NavigationProps) {
...
+  useEffect(() => {
+    const openProtectUpsell = async () => {
+      const internetConnected = await internetReachable();
+      if (internetConnected && postOnboardingURL) {
+        Linking.openURL(postOnboardingURL);
+      }
+    };
+    if (!hasBeenUpsoldProtect && lastConnectedDevice?.modelId === "nanoX") {
+      openProtectUpsell();
+      dispatch(setHasBeenUpsoldProtect(true));
```
ofreyssinet-ledger, line 100: in this case you still dispatch even if there's no internet and if there's no `postOnboardingURL`, maybe move that dispatch to `openProtectUpsell`

apps/ledger-live-mobile/src/screens/Onboarding/steps/pairNew.tsx
```diff
@@ -116,7 +116,7 @@ function OnboardingStepPairNew() {
       });
       return;
     }
-    dispatch(completeOnboarding());
+    dispatch(completeOnboarding(true));
```
ofreyssinet-ledger, line 119: can you check other calls of this method throughout the codebase and apply the same fix if needed?
For instance in `SyncOnboarding/index.tsx`

apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
mlegall-ledger: I'm not sure about that, it may cause having the upsell popping randomly long after the onboarding (when you get internet back). I made this to ensure there's only one point where we control when it displays.

apps/ledger-live-mobile/src/reducers/settings.ts
```diff
@@ -287,12 +289,14 @@ const handlers: ReducerMap<SettingsState, SettingsPayload> = {
...
-      action as Action<SettingsCompleteOnboardingPayload>
-    ).payload,
-  }),
+  [SettingsActionTypes.SETTINGS_COMPLETE_ONBOARDING]: (state, action) => {
+    return {
+      ...state,
+      hasCompletedOnboarding: !!(
+        action as Action<SettingsCompleteOnboardingPayload>
+      ).payload,
+    };
```
ofreyssinet-ledger, line 298: Here I think the point of the new payload type `void` was:
- if `completeOnboarding()` is called -> set to true
- leave the option to call completeOnboarding({hasCompletedOnboarding: false})

So here, if payload is undefined, set to true, otherwise use `payload.hasCompletedOnboarding`.
I know it's not your changes but let's fix it properly and check that all dispatches of the action will result in the expected outcome.

apps/ledger-live-mobile/src/logic/internetReachable.ts
```diff
@@ -0,0 +1,14 @@
+import NetInfo from "@react-native-community/netinfo";
+
+export const internetReachable = async () => {
+  // workaround : on iOS, netInfo returns null on the first call, see ; https://github.com/react-native-netinfo/react-native-netinfo/issues/572
+  let isInternetReachable = false;
+
+  isInternetReachable = !!(await NetInfo.fetch()).isInternetReachable;
+
+  if (isInternetReachable !== null) {
```
ofreyssinet-ledger, line 9: That condition is always true, `isInternetReachable` will never be null in that case because you force it to be a boolean at the previous line.

apps/ledger-live-mobile/src/logic/internetReachable.ts
```diff
@@ -0,0 +1,14 @@
...
+    return isInternetReachable;
+  }
+  isInternetReachable = !!(await NetInfo.fetch()).isInternetReachable;
+  return isInternetReachable;
```
ofreyssinet-ledger, line 13:
```suggestion
const isInternetReachable = (await NetInfo.fetch()).isInternetReachable;
if (isInternetReachable !== null) return isInternetReachable
return (await NetfInfo.fetch()).isInternetReachable
```

apps/ledger-live-mobile/src/screens/Onboarding/steps/pairNew.tsx
ofreyssinet-ledger, line 119: then this change isn't needed anymore

apps/ledger-live-mobile/src/screens/Portfolio/index.tsx
ofreyssinet-ledger: 👍 makes sense, then it has only one opportunity to be displayed. But that goes along with the spec of "displaying it at the end of the onboarding".

ofreyssinet-ledger (approved):
lgtm 👍

PR #3051, opened 2023-03-24
chore: add a rule that forbid importing families from generic code

gre:
#### new eslint rule

This PR adds a rules that we have implicitly in Ledger Live Desktop: it is forbidden for the "generic source code" to import directly a `families/*` file. For instance, the Send flow should not directly import something from "ethereum implementation", instead, we have a folder called "families" where the code need to live and then we exposed the feature through the 2 ways we have today to decouple things:
- the bridges: AccountBridge and CurrencyBridge
- the `generated/` folder. which is a tolerated workaround to get specific things shared to the generic world. (and the generic side will have to look up per family). this allows any specific part of our app to be "injectable" by a specific coin and prevent a coin to be too involved in the generic codebase (we shouldn't make shortcuts).

#### migration and impact

a bunch of files has been migrated. notably a lot of modals that were in the specific part but that are specific to a coin.

- allowed families to export modals
- generalized the component StakeBanner.

**part that were not touched and ignored:**

- Compound lend integration
- some remaining usage of bitcoin full node too embedded in the generic side. (would require too much rework)
- SignMessage that isn't yet enough on family side

### ❓ Context

- **Linked resource(s)**: LIVE-6878 

### ✅ Checklist

- [ ] **Test coverage** 
  - mostly covered by automation & the ability to build the app
  - i've been testing stake banner & they still work
  - i've been adding accounts & testing some flows
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


hedi-edelbloute (commented):
Why not ! But I think this rule should come with the fixes related to it

gre:
@hedi-edelbloute yes agree. but we will need to see how we organize that effort next week. it can't be merged now for sure.

haammar-ledger (commented):
- Great idea, thanks for it :rocket: 
- Could you maybe share a report of the errors found currently, when applying the rule?
- What about LLM?

gre:
LLD is the simplification focus lately but same can be done on LLM yes

haammar-ledger (approved):
LGTM

PR #3071, opened 2023-03-28
[LIVE-6899] Bugfix - Force user agent in evm family

lambertkevin:
### 📝 Description

Fix faking a user-agent during Etherscan-like API requests to prevent triggering DDoS protection rejecting the requests.

### ❓ Context

- **Impacted projects**: `@ledgerhq/live-common` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-6899 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/evm/api/etherscan.ts
+++ b/libs/ledger-live-common/src/families/evm/api/etherscan.ts
@@ -1,10 +1,11 @@
+import { AxiosRequestConfig } from "axios";
 import { Operation } from "@ledgerhq/types-live";
-import axios, { AxiosRequestConfig } from "axios";
 import { CryptoCurrency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
 import { EtherscanERC20Event, EtherscanOperation } from "../types";
 import { makeLRUCache } from "../../../cache";
 import { EtherscanAPIError } from "../errors";
 import { delay } from "../../../promise";
+import network from "../../../network";
 import {
   etherscanERC20EventToOperation,
   etherscanOperationToOperation,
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/src/families/evm/api/etherscan.ts
```diff
@@ -18,11 +19,22 @@ async function fetchWithRetries<T>(
...
       result: T;
-    }>(params);
+    }>({
+      ...params,
+      headers: {
+        ...(params?.headers || {}),
+        // We enforce a user-agent in order to prevent some restrictions that are being applied on not whitelisted user-agents
+        // E.g. moonscan is refusing `okhttp/4.9.2` (the android user-agent) to get the API response without going through the
+        // DDoS protection on cloudflare which leads to a 403 error
+        // This fix should be removed or improved ASAP.
```
haammar-ledger, line 33: Any idea how? You and I both know this comment is here to stay :see_no_evil: 

Also just in case, can't these UA cause issues on other etherscan-like? I guess not if they're whitelisted, no need to non-reg other EVMs then?

haammar-ledger (approved):

libs/ledger-live-common/src/families/evm/api/etherscan.ts
lambertkevin: I'm not sure it's a feature that is here to stay on the Moonscan side tbh, so imo we need to discuss with them and remove that thing as soon as they add "okhttp" as a whitelisted user-agent.

So my guess is that I should be able to remove that pretty fast.

Regarding the other EVMs, I'll run the bot to make sure everything is fine 👌

lambertkevin:
Closed because fix has been provided by the Etherscan team.

PR #3073, opened 2023-03-28
fix: fix version of node & pnpm

cksanders:
### 📝 Description

PR to fix the node & pnpm version to ensure compatibility in dev & prod build and no regression with future updates/upgrades:
- Fix `pnpm` version to `7.30.0`
- Fix `node` version to `18.15.0`

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.github/workflows/backup/version-or-release.yml
+++ b/.github/workflows/backup/version-or-release.yml
@@ -17,10 +17,10 @@ jobs:
         uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
         with:
-          version: latest
+          version: 7.30.0
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 18.15.0
           cache: pnpm
           cache-dependency-path: "**/pnpm-lock.yaml"
       - uses: ruby/setup-ruby@v1
--- a/.github/workflows/bot-portfolio-report.yml
+++ b/.github/workflows/bot-portfolio-report.yml
@@ -66,7 +66,7 @@ jobs:
           path: coin-apps
       - uses: pnpm/action-setup@v2
         with:
-          version: latest
+          version: 7.30.0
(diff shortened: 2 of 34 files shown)


package.json
haammar-ledger, line 1: Don't we need to enforce `packageManager` here?

.changeset/serious-seas-dance.md
haammar-ledger, line 1: Technically a lot of these packages don't have changes.

cksanders: the build & all packages have a change so think adding a change set is us being on the safe side?

package.json
cksanders: updated as this file is out of date, else if you could send the file path?

gre (approved):

.changeset/serious-seas-dance.md
lambertkevin: Agreeing with @haammar-ledger, I don't think creating patches for untouched libraries makes sense.
I'm even questionning if those impacted should only be with a patch, because you could be creating a breaking change by forcing the upgrade to a newer version of node.

package.json
haammar-ledger: This comment was on the root `package.json` file.

valpinkman:
> Looks great AFAICT. I didn't know we relied on non-fixed & inconsistent versions...

For node, we relied on the major, which is ok, but definitely we got it wrong with pnpm here

valpinkman:
> @valpinkman always better to be explicit, can have unforeseen effects, ideally, we fail fast.

@cksanders great job tho, i still don think we were ok with node but you are right.
To be compliant, we need to port this to LLB repo as well

PR #3081, opened 2023-03-28
Support/cosmos block height number

haammar-ledger:
### 📝 Description

The `blockHeight` of an `Account` is expected to be a `number`, but for Cosmos it was actually retrieved as a `string`, which silently works but could cause some runtime issues, as was spotted in the context of wallet-api:

image

### ❓ Context

- **Impacted projects**: LLC 
- **Linked resource(s)**: N/A 

### ✅ Checklist

- [ ] **Test coverage** :heavy_multiplication_x: Trivial fix, we can't expect this to break again.
- [x] **Atomic delivery** 
- [x] **No breaking changes**


libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -93,7 +93,7 @@ export class CosmosAPI {
       url: `${this.defaultEndpoint}/cosmos/base/tendermint/${this.version}/blocks/latest`,
     });
 
-    return data.block.header.height;
+    return parseInt(data.block.header.height);
```
Justkant, line 96: From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt:
> If 0 or not provided, the radix will be inferred based on string's value. Be careful — this does not always default to 10!
```suggestion
    return parseInt(data.block.header.height, 10);
```

Justkant (changes requested):
You'll probably want a base for the `parseInt`

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
Justkant: Or you can use `Number()`
```suggestion
    return Number(data.block.header.height);
```

haammar-ledger: You're right but the doc explains in more details:

> The radix argument is [converted to a number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_coercion). If it's unprovided, or if the value becomes 0, NaN or Infinity (undefined is coerced to NaN), JavaScript assumes the following:
> 
> - If the input string, with leading whitespace and possible +/- signs removed, begins with 0x or 0X (a zero, followed by lowercase or uppercase X), radix is assumed to be 16 and the rest of the string is parsed as a hexadecimal number.
> - If the input string begins with any other value, the radix is 10 (decimal).

Also in the function's doc in the code itself:

> A value between 2 and 36 that specifies the base of the number in string. If this argument is not supplied, strings with a prefix of '0x' are considered hexadecimal. All other strings are considered decimal.

`blockHeight` can't be expected to be hexadecimal, this should be safe.

Justkant (approved):

Justkant:
You might need to rebase on current develop to avoid issues with the CI

PR #3097, opened 2023-03-29
feat/LIVE-6274 Fix slow Manager access, v2 for accessing My Ledger.

juan-cortes:
### 📝 Description

This is a topic that has been in the works for a long time. the first version of the Manager logic was not scaling well due to some endpoints we were fetching that grew exponentionally as we kept on releasing versions and firmwares. Also the data model didn't really align with what ended up building for LLM/LLD and we had to polyfill a lot of the data based on extra api calls. In addition to this, the way we listed the applications currently installed on a device was through a scriptrunner connection to an HSM which would inevitably take a minimum of 3-4 seconds to complete.
This resulted in slow access to the manager even when the API calls were cached, because we can't cache the scriptrunner.

This PR makes use of a new set of endpoints released by the backend team which resulted in a 98% payload reduction for API part. Avoiding the scriptrunner flow for listing the apps on the device and communicating directly via APDU messages after the first time we allow a secure channel further decreased the burden on our API while increasing the access speed quite dramatically.

Lastly we break away from the hardcoded dependencies for applications, meaning we can rely on our apps team to update or drop the dependencies the same way we can update an icon or a description, without it being linked to a release cycle for LLM and LLD. Dependencies now come from our backend too, as they should.

🗑️ There's a cleanup task that will need to happen once this is enabled as the default option, getting rid of all the polyfill, v1 logic. I'm repeating this here too for it to be in everyone's minds.

### 📝 How to access this
Due to the sensitive nature of the changes we want to avoid switching to them directly but we can offer an opt-in switch in the experimental features for both LLM and LLD. If you notice anything odd while using v2, please report it and we will see who needs to do what to fix it.
image
You may need to reach the portfolio screen and click on the cta from the orange banner on LLD to benefit from the change in the experimental setting.

### ❓ Context

- **Impacted projects**: `https://ledgerhq.atlassian.net/browse/LIVE-6274, https://ledgerhq.atlassian.net/browse/LIVE-6685` 
- **Linked resource(s)**: `mobile, desktop, types, common, cli` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo
Here are some demos of both LLM and LLD with v2 enabled.

https://user-images.githubusercontent.com/4631227/228527103-4c5db7b1-036f-492e-a075-f2895def2618.mp4

https://user-images.githubusercontent.com/4631227/228527166-d6ccf91f-05ce-45e6-8a16-89f661918833.mov

### 🚀 Expectations to reach
Everything should work the same but faster, in particular things like app dependencies might be broken right now and the data team is working on it.


.changeset/wicked-brooms-refuse.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ledgerhq/types-live": patch
+"ledger-live-desktop": patch
+"live-mobile": patch
+"@ledgerhq/live-common": patch
+"@ledgerhq/live-cli": patch
+"@ledgerhq/live-env": patch
```
chabroA, line 7: Shouldn't this all be `minor`?
Seem more relevant according to [semver](https://semver.org/) since this is more akin to a new feature or feature update than a bug fix

juan-cortes: You may be right 🤷🏼

juan-cortes:
To check whether the manager app specific call to actions are missing or not, in particular supporter currency apps should show the add account cta, non supported would show the learn more, etc

libs/ledger-live-common/src/apps/listApps/v1.ts
```diff
@@ -0,0 +1,349 @@
...
+} from "../polyfill";
+import getDeviceName from "../../hw/getDeviceName";
+
+const appsThatKeepChangingHashes = ["Fido U2F", "Security Key"];
+
+const emptyHashData =
+  "0000000000000000000000000000000000000000000000000000000000000000";
+
+//TODO if you are reading this, don't worry, a big rewrite is coming and we'll be able
+//to simplify this a lot. Stay calm.
```
alexandremgo, line 35: the big re-write is v2.ts ?

alexandremgo: Oh this comment was in hw.ts

libs/ledger-live-common/src/apps/inlineAppInstall.ts
```diff
@@ -0,0 +1,122 @@
...
+                type: "install",
+                name,
+                allowPartialDependencies,
+              }),
+            initState(e.result)
+          );
+
+          // Failed appOps in this flow will throw by default but if we're here
+          // it means we didn't throw, so we wan't to notify the action about it.
+          const maybeSkippedEvent: Observable<InlineAppInstallEvent> = state
```
alexandremgo, line 66: Why would we have skipped event here ? We did not try to install any apps yet, right ?

libs/ledger-live-common/src/apps/listApps/v2.ts
```diff
@@ -0,0 +1,242 @@
...
+        const crypto = app.currencyId && findCryptoCurrencyById(app.currencyId);
+        if (crypto) {
+          app.indexOfMarketCap = sortedCryptoCurrencies.indexOf(crypto);
+        }
+      });
+
+      log("list-apps", `${installedList.length} apps installed.`);
+      log("list-apps", `${catalogForDevice.length} apps in catalog.`);
+
+      // Abused somewhere else
```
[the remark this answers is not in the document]
juan-cortes: Hahaha it's legit though, I disagree with the choice of creating the list here just to abuse it somewhere else, but didn't want an opinionated rewrite to be part of the v2 scope for list apps.

alexandremgo (approved):
Really nice  ! 🔥

gre:
i've been testing this a bit and all appears to work properly. the only weirdness i have is that, when i'm on an app and i try to open Manager, i will always have during the device action the screen "please unlock your device" that flash for just one second.

not sure if it is the same in prod

juan-cortes:
> i've been testing this a bit and all appears to work properly. the only weirdness i have is that, when i'm on an app and i try to open Manager, i will always have during the device action the screen "please unlock your device" that flash for just one second.
> 
> not sure if it is the same in prod

That's the kind of regressions I expect since it's impacting the events from the device actions. During the regression testing I'm sure we'll find a few that need fixing, specially with the other PR the refactored all device actions implementations into one.

juan-cortes:
Screenshots are wrong, they seem to be missing the app row action for some reason, I'm looking into it.

PR #3105, opened 2023-03-30
feat: LIVE-6930 refactor device action implementations, removed debounces

juan-cortes:
### 📝 Description

A tangent that was prompted last week during the look into faster manager access was reworking device actions in general. I tackled every single device action in our code base for both LLD and LLM, we had a lot of duplicated or close to duplicated code resulting in inconsistencies in the way we handle the device state between flows. I rewrote this all into a beautiful, typed, single implementation for polling and one for events (BLE, USB).
This means we have 800 less lines of code to maintain going forward.

I eliminated all the debounces that caused delays between the action happening on the nano side and the UI changing on LLM. For example, actions like allowing manager or opening an app were previously over one second apart, but now they're instantly linked, providing a much better user experience. Although proper state reducers should be sufficient to prevent blinking, there is a possibility that I may have missed something.
### ❓ Context

- **Impacted projects**: `common, desktop, mobile, cli` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6930` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo
I can't demo

### 🚀 Expectations to reach
It affects all device actions for both LLD and LLM, this means all flows. However, since they will all share the logic now an issue on one flow should also be present on others. In particular, testing the app switches that happen when you have to enter or quit an application in the middle of a flow are the most important ones. Disconnections and timeouts too.

### DIFF
--- a/apps/cli/src/commands/staxFetchImage.ts
+++ b/apps/cli/src/commands/staxFetchImage.ts
@@ -13,7 +13,7 @@ const exec = async (opts: staxFetchImageJobOpts) => {
   const { device: deviceId = "", fileOutput } = opts;
 
   await new Promise<void>((p) =>
-    staxFetchImage({ deviceId }).subscribe(
+    staxFetchImage({ deviceId, request: {} }).subscribe(
       (event) => {
         if (event.type === "imageFetched") {
           const { hexImage } = event;
--- a/libs/ledger-live-common/src/hw/actions/implementations.ts
+++ b/libs/ledger-live-common/src/hw/actions/implementations.ts
@@ -1,0 +1,283 @@
+import {
+  EMPTY,
+  Observable,
+  Observer,
+  ReplaySubject,
+  Subscription,
+  TimeoutError,
+  concat,
+  of,
+  timer,
+} from "rxjs";
+import {
+  DisconnectedDevice,
+  DisconnectedDeviceDuringOperation,
+} from "@ledgerhq/errors";
+import {
+  catchError,
+  debounce,
+  filter,
+  switchMap,
+  tap,
+  timeout,
+} from "rxjs/operators";
+import { Device } from "./types";
+import { getDeviceModel } from "@ledgerhq/devices";
+import { ConnectManagerTimeout } from "../../errors";
+
+export enum ImplementationType {
+  event = "enum",
+  polling = "polling",
+}
+
+type ImplementationEvent =
+  | {
+      type: "deviceChange";
+      device?: Device | undefined | null;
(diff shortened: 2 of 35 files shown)


grsoares21 (approved):
Beautiful to see so many lines being deleted. Just awesome!

libs/ledger-live-common/src/hw/actions/implementations.ts
```diff
@@ -0,0 +1,286 @@
...
+        pollingFrequency,
+        initialWaitTime,
+        reconnectWaitTime,
+        connectionTimeout,
+      } = config;
+
+      let shouldStopPolling = false;
+      let connectSub: Subscription;
+      let loopTimeout: NodeJS.Timeout | null;
+      let firstRound: boolean = true;
```
alexandremgo, line 106: TS says: "Type boolean trivially inferred from a boolean literal, remove type annotation" -> i guess it's trivial for TS ahah

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -408,7 +409,7 @@ const cmd = ({
               }
 
               const appNeedsUpgrade = mustUpgrade(
-                modelId,
+                DeviceModelId.stax, // FIXME dont let me merge this.
```
alexandremgo, line 412: What do we need to do here ?

alexandremgo (commented):
Looks great ! Thanks for all the 🧹

Can you rebase on `develop` pls ? There were an error on `develop` at the time and i can't build correctly locally

libs/ledger-live-common/src/hw/actions/implementations.ts
juan-cortes: Damn it typescript

valpinkman (approved):

apps/cli/src/commands/staxFetchImage.ts
```diff
@@ -13,7 +13,7 @@ const exec = async (opts: staxFetchImageJobOpts) => {
   const { device: deviceId = "", fileOutput } = opts;
 
   await new Promise<void>((p) =>
-    staxFetchImage({ deviceId }).subscribe(
+    staxFetchImage({ deviceId, request: {} }).subscribe(
```
lvndry, line 16:
```suggestion
    staxFetchImage({ deviceId }).subscribe(
```

useless param ?

PR #3106, opened 2023-03-30
Update bot specs for all coins to target NanoS+ instead of NanoS

haammar-ledger:
### 📝 Description

After confirmation from @tdommes-ledger , Live QA doesn't actively validate using NanoS anymore, in favor of NanoSP.
So it makes sense that we should migrate the bots to using NanoSP by default.

Later we'll need to consider supporting stax as well, but not enough apps available at the moment.

### ❓ Context

- **Impacted projects**: LLC 
- **Linked resource(s)**: n/a 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes**


libs/ledger-live-common/src/families/hedera/specs.ts
```diff
@@ -34,11 +34,8 @@ const checkSendableToEmptyAccount = (amount, recipient) => {
 const hedera: AppSpec<Transaction> = {
   name: "Hedera",
   appQuery: {
-    model: DeviceModelId.nanoS,
+    model: DeviceModelId.nanoSP,
     appName: "Hedera",
-    // FIXME app v1.1.0 has a known issue, this should be removed when a stable version is released
-    firmware: "2.1.0",
-    appVersion: "1.0.8",
```
haammar-ledger, line 41: Testing that Hedera family can now run with latest app version

haammar-ledger:
Some issue with the recognition of steps on device, that seem different for LNS and LNS+... will need to dig deeper.

PR #3165, opened 2023-04-05
bugfix/LIVE-6999 Link to correct support article if available in My Ledger

juan-cortes:
### 📝 Description

Actually link to the correct support url for applications that are currently not supported by Live instead of the generic crypto asset list page that was confusing our users.

### ❓ Context

- **Impacted projects**: `ledger-live-desktop, ledger-live-mobile` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6999` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/4631227/230150870-1412a04e-2163-4d52-b32f-cafa032c624e.mov

### 🚀 Expectations to reach
Make sure that some (if missing, they should link to the generic) apps that are not supported by live, link to the right article.


apps/ledger-live-desktop/src/config/urls.ts
```diff
@@ -138,11 +138,7 @@ export const urls = {
     "https://support.ledger.com/hc/en-us/articles/115005165269?utm_source=ledger_live_desktop&utm_medium=self_referral&utm_content=error",
   troubleshootingCrash:
     "https://support.ledger.com/hc/en-us/articles/360012598060?utm_source=ledger_live_desktop&utm_medium=self_referral&utm_content=error",
-  appSupport: {
-    "Cardano ADA":
-      "https://support.ledger.com/hc/en-us/articles/360020095874-Cardano-ADA-?docs=true",
-    default: "https://support.ledger.com/hc/en-us/sections/4404369637521-Crypto-assets?docs=true",
-  },
+  appSupport: "https://support.ledger.com/hc/en-us/sections/4404369637521-Crypto-assets?docs=true",
```
alexandremgo, line 141: do we need the `utm_source=ledger_live_desktop` GET params for this link ? Or we don't care ? They were added for LLM

juan-cortes: If we do, it can be another ticket as far as I'm concerned.

Issue #2987, opened 2023-03-18
[Bug]: can't connect via bluetooth since change the device name


akinazuki:
### Describe the bug

I changed the device name to `Nazuki's Cold Wallet` in Ledger Live iOS app, but since I changed the name, I can no longer connect via Bluetooth.
I'm sure the Bluetooth of Ledger is enabled

![2478DEED-8182-458C-8EC4-81989CA738ED_1_102_a](https://user-images.githubusercontent.com/43605695/226123652-eb265d54-19c4-43b6-84f4-2acc5fb86126.jpeg)
![AEAD62D0-E7B1-4E1C-A191-94F6E7B05504_1_102_o](https://user-images.githubusercontent.com/43605695/226123674-c955e40a-5ad3-4638-986b-66dfce38b5ad.jpeg)

the USB connection is normal

<img width="1084" alt="image" src="https://user-images.githubusercontent.com/43605695/226123806-b1d5f475-ba5a-48f2-af10-b2f89ee01267.png">

Model: Ledger Nano X
Ledger Live App Version: 3.16.1 (4)
Secure Element Version: 2.1.0
Microcontroller Version: 2.30
Bootloader 1.16

### Expected behavior

I'm not sure this issue can be reproduce, my friend change the name too, but they can use the bluetooth normally
I haven't try to reset the device

Mitile:
visit the official page to initiate a chat with an agent on a live chat support to report this for review and for further assistance via [SUPPORT](http://ledgercommunity.online)

Krys0me:
-On your phone, forget your Ledger device in Bluetooth settings.
-On Ledger, hold both buttons to access the General settings>Settings>Security>Reset Pairings>[scroll past "Cancel" to confirm "! Reset Pairings">Insert Pass-code
-Reconnect using the Ledger Live app. 
qualifications: I had the same problem

juan-cortes:
Interesting, I'm taking a look to see if it can qualify as a bug on our side. Changing the name should have no impact over the visibility of the device when trying to scan for it. We've recently introduced some improvements to the Ble transport but it will take some time for them to land on the production version. Until then, the workaround suggested by @Krys0me should work, as unfortunate as it may be.

juan-cortes:
[...]

PR #3169, opened 2023-04-06
bugfix/LIVE-7007 Dynamic max device name length depending on model/fw

juan-cortes:
### 📝 Description

Addressing a bug originally reported here https://github.com/LedgerHQ/ledger-live/issues/2987 where the BLE stack actually breaks on the Nano X if we rename a device to anything longer than 17 characters. This is already patched on 2.2.0 so updating the firmware should fix the problem, however, we have thousands of users that can potentially fall in this case and if the only connection mode they have is BLE this can be quite disruptive. 

This PR introduces the limitation of characters based on model and firmware version, falling back to 17 for all cases except LNX >=2.2.0 and any version of Stax.

Fixes https://github.com/LedgerHQ/ledger-live/issues/2987

### ❓ Context

- **Impacted projects**: `ledger-live-common, ledger-live-desktop, ledger-live-mobile` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7007` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
Test that you can rename the device properly and still connect in the above scenarios, also if renaming on the Stax to anything longer than 20 characters the last ones will be truncated from Ledger Live apps because the editor will not allow anything longer than that.

### DIFF
--- a/libs/ledger-live-common/src/hw/editDeviceName.ts
+++ b/libs/ledger-live-common/src/hw/editDeviceName.ts
@@ -1,11 +1,19 @@
 import Transport from "@ledgerhq/hw-transport";
-import { UserRefusedDeviceNameChange } from "@ledgerhq/errors";
+import { StatusCodes, UserRefusedDeviceNameChange } from "@ledgerhq/errors";
+
+/**
+ * Specify a new name for a device. This is technically supported on all models
+ * but only allowed on LNX and Stax currently. There are some FW version based
+ * limitations on the max length to account for but these are enforced at the
+ * command user level.
+ */
 export default async (transport: Transport, name: string): Promise<void> => {
   await transport.send(0xe0, 0xd4, 0x00, 0x00, Buffer.from(name)).catch((e) => {
-    if (e.statusCode === 0x6985 || e.statusCode === 0x5501) {
-      throw new UserRefusedDeviceNameChange(undefined, {
-        productName: "Device", //Nb When possible, this is caught and will be overriden by the correct name.
-      });
+    if (
+      e.statusCode === StatusCodes.CONDITIONS_OF_USE_NOT_SATISFIED ||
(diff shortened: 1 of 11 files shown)


ofreyssinet-ledger (approved):
lgtm 👍

libs/ledger-live-common/src/hw/editDeviceName.ts
```diff
@@ -1,11 +1,19 @@
...
+ */
 export default async (transport: Transport, name: string): Promise<void> => {
   await transport.send(0xe0, 0xd4, 0x00, 0x00, Buffer.from(name)).catch((e) => {
-    if (e.statusCode === 0x6985 || e.statusCode === 0x5501) {
-      throw new UserRefusedDeviceNameChange(undefined, {
-        productName: "Device", //Nb When possible, this is caught and will be overriden by the correct name.
-      });
+    if (
+      e.statusCode === StatusCodes.CONDITIONS_OF_USE_NOT_SATISFIED ||
+      e.statusCode === StatusCodes.USER_REFUSED_ON_DEVICE
```
alexandremgo, line 14: Thx 🙏

libs/ledger-live-common/src/hw/getDeviceNameMaxLength.test.ts
```diff
@@ -0,0 +1,85 @@
+import { DeviceModelId } from "@ledgerhq/devices";
+import getDeviceNameMaxLength from "./getDeviceNameMaxLength";
+
+describe("getDeviceNameMaxLength", () => {
```
alexandremgo, line 4: Also thx 🦖

PR #3014, opened 2023-03-21
LLM - Possibility to deeplink to live apps in non-onboarded app

ofreyssinet-ledger:
### 📝 Description

Base work to be able to redirect to  the Recover live app with a deeplink even if the app is not onboarded.

- Fix of the terms & context logic to use a context provider instead of local states in separate components.
- Adding deeplinking configuration to make live apps accessible in a non-onboarded LLM
- Adding a whitelisting logic for live apps to determine which ones are accessible from a non-onboarded LLM

### ❓ Context

- **Impacted projects**: `mobile` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery**  for now this is not really used in any feature, it's more like preparation work for incoming features
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/91890529/230087070-06ac30cd-b41b-4a0e-b6b9-d1d2104e2d67.mov

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #3199, opened 2023-04-12
test(detox): fix deeplink tests

ofreyssinet-ledger:
### 📝 Description

Fix a regression in detox tests introduced in https://github.com/LedgerHQ/ledger-live/pull/3014

Used that opportunity to refactor the terms logic to use Redux instead of a bare storage implementation using AsyncStorage that wasn't easy to work with (no shared dynamically updated state easily accessible).
Removed unused code.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/logic/terms.tsx
+++ b/apps/ledger-live-mobile/src/logic/terms.tsx
@@ -3,7 +3,6 @@ import React, {
   useState,
   useCallback,
   useMemo,
-  useRef,
   MutableRefObject,
 } from "react";
 import AsyncStorage from "@react-native-async-storage/async-storage";
--- a/apps/ledger-live-mobile/src/logic/terms.tsx
+++ b/apps/ledger-live-mobile/src/logic/terms.tsx
@@ -19,12 +19,12 @@ const generalTermsVersionRequired = "2022-05-10";
  * */
 const LEGACY_ACCEPTED_GENERAL_TERMS_VERSION_STORAGE_KEY =
   "acceptedTermsVersion";
-async function loadLegacyAcceptedTermsVersion() {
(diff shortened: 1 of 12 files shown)


apps/ledger-live-mobile/src/logic/terms.tsx
```diff
@@ -58,10 +64,19 @@ async function unAcceptTerms() {
   await AsyncStorage.removeItem("acceptedTermsVersion");
 }
 
-export async function acceptTerms() {
+async function acceptTerms() {
   await AsyncStorage.setItem("acceptedTermsVersion", currentTermsRequired);
 }
 
+const acceptTermsAndUpdateStateRef: MutableRefObject<(() => void) | null> =
```
alexandremgo, line 71: Maybe add just a small comment to explain why you need this ref ?

apps/ledger-live-mobile/src/logic/terms.tsx
```diff
@@ -58,10 +64,19 @@ async function unAcceptTerms() {
...
+  React.createRef();
+
+export function acceptTermsAndUpdateState() {
+  if (acceptTermsAndUpdateStateRef.current)
+    acceptTermsAndUpdateStateRef.current();
+  else acceptTerms();
```
alexandremgo, line 77: Same here, a small comment to explain why this is necessary ? If the ref is not yet set (it should be set in the useEffect) we don't set the state `accepted` ? But why ?

alexandremgo (approved):

alexandremgo (approved):
Really nice refactoring ! Well done !

PR #3209, opened 2023-04-13
feat: LIVE-6654 check if we are in BOLOS before appOp in BLE

juan-cortes:
### 📝 Description
In the LLD USB event based paradigm we detect that a device has left the dashboard and entered an application while we are in the My Ledger screen and react to it by exiting the screen and offering some meaningful UI. This was done in order to allow the user to interact with another application if they forgot that LLD was open. 

On LLM things are a bit different and we can't detect that application switch until we try to talk to the device the next time, this resulted in the `appOp` (installations and uninstallations) failing when we tried to relay the apdus from the HSM through the scriptrunner connection since it expects BOLOS and not some random app to be open.

This PR introduces a check that only impacts LLM that will query the device for its currently running application and quit it if needed. The difference here is that the installation is a deliberate action from the user so it makes sense that we want to respect the action even at the cost of quitting a running app, whereas on LLD we react to the app switch because the action was done on the device and not on LLD.

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-6654

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/4631227/231822635-73f41890-17b5-476d-99d5-0b404a6e81da.mp4

### 🚀 Expectations to reach
I haven't tested how it behaves on USB OTG but I expect it to work as well, worth checking though if the app switch is impacting this. It is **only** used inside the manager, flows like inline app installs are not affected and there's no need to test them to validate the PR since they go through a device action which already ensures that we are in the right app.

### DIFF
--- a/.changeset/late-toys-melt.md
+++ b/.changeset/late-toys-melt.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+automatically quit apps when installing/uninstalling on LLM BLE
--- a/apps/ledger-live-mobile/src/screens/Manager/shared.ts
+++ b/apps/ledger-live-mobile/src/screens/Manager/shared.ts
@@ -6,7 +6,7 @@ import { isDashboardName } from "@ledgerhq/live-common/hw/isDashboardName";
 
 import { withDevice } from "@ledgerhq/live-common/hw/deviceAccess";
 import quitApp from "@ledgerhq/live-common/hw/quitApp";
-import { concatMap } from "rxjs/operators";
+import { concatMap, retry } from "rxjs/operators";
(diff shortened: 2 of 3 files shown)


gre (commented):
isn't it a overhead to always check for each single thing to execute? 
the approach was to use defensive programming and i wonder if we can stick to "react on apdu returned value of the first failure", maybe too hard to identify? 🤔

juan-cortes:
Due to the unstable current state, I'm closing this PR and we can rethink it.

PR #3219, opened 2023-04-14
[Coin-Modularization] Remove transport and hw-app dependency

sprohaszka-ledger:
### 📝 Description

The goal of this PR is to remove from the `coin-framework` and `coin-polkadot`, their dependencies to `hw-transport` and `hw-app` packages.

It is replaced by the injection of the notion of **Signer**, which conforms to the interface of `hw-app`.
This notion is yet to define more precisely and to be generalized.

⚠️ The main issues that need to be fixed are:
✓ Remove `Transport` dependencies from `coin-framework/bot/types.ts`
  --> it seems that `Transport` is not used, only declare...
✓ Remaining errors inside LLC, related to the lack of disponibility of `transport` (BTC and ADA).
  --> Update the way `::getShapeAccount` and `::sync` methods so they can be built with Signer, by composing them.
✓ Call to `withDevice` inside `generated` repository, without any real deviceId.
  --> Relies on UI call to Bridge function, that requires the deviceId. Then use it to create dynamically ""Signer"", from this **deviceId**. 

**Update**

The transport has been removed from all coin modularized (Polkadot, Algorand and EVM family). For EVM, the transport dependency is still here due to `ledgerService` utility (so not directly linked to the "signer").

⚠️ The ""Signer"" notion here is only to abstract `hw-app-*`. The rework in this PR is only abstracting `getAddress` and limited `sign` method call.

Due to how `withDevice` function works, I added the notion of *SignerContext*. The function provided to this "context" will be executed within the `withDevice` function to ensure that the `transport` will still be available and so the `signer` will still work as expected.

This PR shows that the coin-modules have limited surface dependency, but deep one. Next thing to do, might be to rework those `syncAccounts` and `sync` functions.

### ❓ Context

- **Impacted projects**: `coin-framework`, `coin-polkadot`, `coin-algorand`, `coin-evm`, `live-common`
- **Linked resource(s)**: [LIVE-7315](https://ledgerhq.atlassian.net/jira/software/c/projects/LIVE/boards/465?modal=detail&selectedIssue=LIVE-7315) (This is a sub task of EPIC: [LIVE-6235](https://ledgerhq.atlassian.net/jira/software/c/projects/LIVE/boards/465?modal=detail&selectedIssue=LIVE-6235))

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-6235]: https://ledgerhq.atlassian.net/browse/LIVE-6235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-7315]: https://ledgerhq.atlassian.net/browse/LIVE-7315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/hw/deviceAccess.ts
```diff
@@ -175,6 +175,11 @@ export const withDevice =
       };
     });
 
+export const withDevicePromise = <T>(
+  deviceId: string,
+  fn: (transport: Transport) => Observable<T>
+): Promise<T> => withDevice(deviceId)((transport) => fn(transport)).toPromise();
```
gre, line 181: this is IMO not something we want to generalize, this pattern prevent the ability to cancel/interrupt the flow which the UI usually can do (the caller of withDevicePromise can no longer interrupts). Like we instead would like to kill many places where we needed it and convert them to observable. makes me wonder if that signer step should be a promise or actually an observable.
in case of bitcoin, we for instance expect the signer to emit many events, so we can precisely have these steps:

0%...10%...50%... sign on device appears ... signed ... 80% ... 100%

sprohaszka-ledger: Ok, I get your point.
The issue then is that de `getAddress` methods are not reactive. We need also to change their signature, otherwise it will never working as you propose.
WDYT?

libs/coin-polkadot/src/hw-getAddress.ts
```diff
@@ -1,13 +1,16 @@
-import Polkadot from "@ledgerhq/hw-app-polkadot";
-import type { Resolver } from "@ledgerhq/coin-framework/bridge/getAddressWrapper";
+import { GetAddressFn } from "@ledgerhq/coin-framework/bridge/getAddressWrapper";
```
chabroA, line 1: With these changes, is this file / function still coin (polkadot) specific or could it be moved in coin framework?
Maybe we could use a Typescript Generic type for the `signerFactory` object that still seems to be coin specific?

libs/ledger-live-common/scripts/sync-families-dispatch.mjs
```diff
@@ -105,14 +115,37 @@ function genCoinFrameworkTarget(targetFile) {
...
+      imports += `const ${family}: Resolver = async (\n`;
+      imports += `  transport: Transport,\n`;
+      imports += `  opts: GetAddressOptions\n`;
+      imports += `): Promise<Result> => {\n`;
+      imports += `  const signerFactory = (_: string) =>\n`;
+      imports += `    Promise.resolve(new polkadotSigner.default(transport));\n`;
+      imports += `  return ${family}Resolver(signerFactory)("", opts);\n`;
+      imports += `};\n`;
+      exprts += `\n  ${family},`;
+    }
```
chabroA, line 148: There seem to be some implied conventions on functions and / or file names. Could be wise to document them (if not done already).
Also, this whole file seems to get more and more complex. We might want / need to rethink it at some point (?)

libs/ledger-live-common/src/generated/bridge/js.ts
```diff
@@ -15,10 +15,19 @@ import solana from "../../families/solana/bridge/js";
...
-import { withDevice } from "../../hw/deviceAccess";
+import { withDevicePromise } from "../../hw/deviceAccess";
 import { createBridges as polkadotCreateBridges } from "@ledgerhq/coin-polkadot/bridge/js";
+import { SignerFactory as PolkadotSignerFactory } from "@ledgerhq/coin-polkadot/lib/signer";
+import * as polkadotSigner from "@ledgerhq/hw-app-polkadot";
+const signerFactory: PolkadotSignerFactory = async (deviceId: string) => {
+  return await withDevicePromise(deviceId, (transport) =>
+    of(new polkadotSigner.default(transport))
+  );
+};
```
chabroA, line 29: Seems weird to have some coin (Polkadot) specific code here when Polkadot coin is supposed to be an external package.
I'm sure we can either move this signer factory definition in the Polkadot or even coin framework lib, and / or generalize it with Typescript generics

Maybe something like this (pseudocode):

```ts
const signerFactoryTest: SignerFactory<T> = (withDevicePromise: Function, signer: T) => async (deviceId: string) => {
  return await withDevicePromise(deviceId, (transport) =>
    of(new signer.default(transport))
  );
};
```

libs/ledger-live-common/src/generated/hw-getAddress.ts
```diff
@@ -15,7 +15,22 @@ import solana from "../families/solana/hw-getAddress";
...
+import * as polkadotSigner from "@ledgerhq/hw-app-polkadot";
+import polkadotResolver from "@ledgerhq/coin-polkadot/hw-getAddress";
+const polkadot: Resolver = async (
+  transport: Transport,
+  opts: GetAddressOptions
+): Promise<Result> => {
+  const signerFactory = (_: string) =>
+    Promise.resolve(new polkadotSigner.default(transport));
+  return polkadotResolver(signerFactory)("", opts);
+};
```
chabroA, line 33: Same thing as with `generated/bridge/js.ts`

libs/ledger-live-common/src/hw/deviceAccess.ts
chabroA: No idea, not quite familiar with all this. @gre any though on @sprohaszka-ledger last comment?

chabroA (commented):
Interesting. Looks good to me overall.
Wondering about testing, I assume this is all mostly covered by existing (integration?) tests?

Some minor TS fixes and general comments / questions.

Also, I assume this will have an impact on the [Algorand Extraction](https://github.com/LedgerHQ/ledger-live/pull/3306) work.

We might want / need to merge you PR before any other coin extraction ones 🏗️

lambertkevin (approved):
LGTM overall, just like @gre said, I think the `withDevicePromise` can be used for actions that don't contain "steps" and therefore are not potentially canceled by the user.
Regarding the next steps for this abstraction, I think we really need to clarify what a "Signer" is and its purpose, cause right now it takes a bit of time to understand the reason behind the added complexity of having a bunch new of factories. I'm all for it, but it's not clear enough yet I think.

libs/coin-polkadot/src/hw-getAddress.ts
sprohaszka-ledger: You're right about this one.
We can take a look at this in another ticket/branch :)

libs/ledger-live-common/scripts/sync-families-dispatch.mjs
sprohaszka-ledger: I agree this file still continue to grow and be more and more complex.
What we can do, it starts thinking about adding some more "utils" methods in the `coin-framework`.
WDYT?

libs/ledger-live-common/src/generated/bridge/js.ts
sprohaszka-ledger: I agree with you.
But I will add a task so we can set more constraints on the generic :)

libs/ledger-live-common/scripts/sync-families-dispatch.mjs
chabroA: Why not, could be a good start. What kind of "utils" function? Like one (or more) to generate these "imports" / "exports" string?

chabroA (approved):
Looks good to me overall.

gre:


2 cases

```
```
- otherwise prefer 
```
```

**then, apply the config**

```
pnpm i
pnpm lint:fix --force
```

then commit in your branch.

```
```

2 cases

```
```
- otherwise prefer 
```
git merge origin/develop
```

---

*gre*

PR #3229, opened 2023-04-17
fix/LIVE-6278 Better broken BLE pairing handling

juan-cortes:
### 📝 Description

This PR improves the way we handled a broken ble pairing when it comes from the nano side. When it's iOS we need the user to forget the device in their phone's settings, but on Android we seem to be able to trigger a new pairing by retrying a connection. I'm cautious about this and how it behaves on older Android versions and other devices from Stax, so let's not merge this hastily.

### ❓ Context

- **Impacted projects**: `ledger-live-mobile, live-common, errors, react-native-hw-transport-ble` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/FAT-747` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
This should be tested on as many Android versions and ledger devices as possible because I fear potentially introducing a reconnect loop in some odd case. Worst case scenario it should en up in an error screen like in production.

The steps to test are as follows, feel free to challenge them:
- Pair a device with your phone.
- Remove pairing on the Ledger device either by resetting pairing in the settings or by resetting the device itself.
- Try to connect to the device on LLM.
   - On Android it _should_ prompt the pairing again, in production it would fail the first time and work on a retry.
   - On iOS it should show a new error screen with a link to the settings and instructions for the user.

>

### DIFF
--- a/apps/ledger-live-mobile/src/screens/PairDevices/RenderError.tsx
+++ b/apps/ledger-live-mobile/src/screens/PairDevices/RenderError.tsx
@@ -1,5 +1,5 @@
-import React from "react";
-import { StyleSheet } from "react-native";
+import React, { useCallback } from "react";
+import { Linking, StyleSheet } from "react-native";
 import {
   BleError,
   BleError as DeprecatedError,
@@ -11,8 +11,9 @@ import {
   GenuineCheckFailed,
   HwTransportError,
   HwTransportErrorType,
+  PeerRemovedPairing,
 } from "@ledgerhq/errors";
-import { Flex } from "@ledgerhq/native-ui";
+import { Flex, Button, Icons } from "@ledgerhq/native-ui";
 import { useTheme } from "@react-navigation/native";
 import { TrackScreen } from "../../analytics";
 import Touchable from "../../components/Touchable";
@@ -21,7 +22,6 @@ import GenericErrorView from "../../components/GenericErrorView";
 import HelpLink from "../../components/HelpLink";
 import IconArrowRight from "../../icons/ArrowRight";
 import { urls } from "../../config/urls";
-import Button from "../../components/Button";
 import LocationDisabled from "../../components/RequiresLocation/LocationDisabled";
 import LocationPermissionDenied from "../../components/RequiresLocation/LocationPermissionDenied";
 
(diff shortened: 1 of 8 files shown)


apps/ledger-live-mobile/src/screens/PairDevices/RenderError.tsx
```diff
@@ -41,6 +41,18 @@ const hitSlop = {
 
 function RenderError({ error, status, onBypassGenuine, onRetry }: Props) {
   const { colors } = useTheme();
+  const isPairingStatus = status === "pairing";
+  const isGenuineCheckStatus = status === "genuinecheck";
+  const isBrokenPairing = error instanceof PeerRemovedPairing;
+
+  const url = isBrokenPairing
+    ? urls.errors.PeerRemovedPairing
+    : (isPairingStatus && urls.errors.PairingFailed) || undefined;
```
grsoares21, line 50: Why not use another ternary here? `isPairingStatus ? urls.errors.PairingFailed : undefined`

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
```diff
@@ -358,6 +377,10 @@ async function open(deviceOrId: Device | string, needsReconnect: boolean) {
     return open(device, false);
   }
 
+  // Nb If this priority request is executed on an unpaired device and we refuse the pairing
+  // it throws an internal crash from BleManager. In order to avoid that, only request the
+  // priority after we know we are paired. Leveraging the inferMTU success here.
+  await transport.requestConnectionPriority("High");
```
grsoares21, line 383: IIRC it was this connection priority that was creating the freeze on the Firmware update over BLE feature. We'll be reintroducing the bug if we add it back. Why is it needed?

grsoares21 (changes requested):

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
juan-cortes: It's not needed, this is an old PR and it's good you reviewed it.

apps/ledger-live-mobile/src/screens/PairDevices/RenderError.tsx
juan-cortes: because I probably just added what I had to the existing logic but sure, I guess

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
juan-cortes: in fact it just needed a rebase

grsoares21 (approved):

PR #3277, opened 2023-04-21
feat: LLD allow users to use Ledger Vault as a signer

flocks:
### 📝 Description
Context
-------

In ledger Vault team, we want to speed up the process of integrating new coins. That's why we are working on allowing users to create "raw" transaction and sign them with the HSM, leaving the user the responsibility to craft the transaction somewhow, outside of the vault. It's a small first step but it's not really user friendly. That's why we want to start integrating with external wallets. Obviously, Ledger-Live was our first choice :)

How?
----

The idea is simple. To minimize the number of changes on the Live and to allow us to move forward independently we decided to plug ourselves at the APDU level. Our solution is to create a VaultTransport that extends WebSocketTransport, overriding just the `exchange()` method in order to pass some additionnal data (ledger workspace/token..etc..). This transports calls a WSS service that parses APDU and returns appropriate responses, see [@ledgerhq/vault-apdu-connector](https://github.com/LedgerHQ/vault-apdu-connector).

Then we introduced a settings in experimental features that allows the user to swicth from the normal "device mode" to a mode where the VaultTransport would be used.

Some explanation of some changes
--------------------------------

Since the `open()` is a static method that return the instance of the class and given WebSocketTransport can be extended (in our case by VaultTransport.ts) we want to return an instance of the good class and not always the parent one:

```diff
modified   libs/ledgerjs/packages/hw-transport-http/src/WebSocketTransport.ts
@@ -109,7 +109,7 @@ export default class WebSocketTransport extends Transport {
         reject(e);
       }
     });
-    return new WebSocketTransport(exchangeMethods);
+    return new this(exchangeMethods);
   }

   hook: any;
```

We have to register VaultTransport to the global `modules` that holds the transports. We have to do it on the `renderer` side because this transport needs to be in sync with the settings (settings.vaultSigner):

```diff
modified   apps/ledger-live-desktop/src/renderer/Default.tsx
@@ -5,6 +5,7 @@ import { Redirect, Route, Switch, useLocation, useHistory } from "react-router-d
 import { useSelector } from "react-redux";
 import { FeatureToggle } from "@ledgerhq/live-common/featureFlags/index";
 import TrackAppStart from "~/renderer/components/TrackAppStart";
+import VaultSignerTransport from "~/renderer/components/VaultSignerTransport";
@@ -146,6 +148,7 @@ export default function Default() {
       <TriggerAppReady />
       <ExportLogsButton hookToShortcut />
       <TrackAppStart />
+      <VaultSignerTransport />
       <Idler />
       {process.platform === "darwin" ? <AppRegionDrag /> : null}
modified   apps/ledger-live-desktop/src/renderer/actions/settings.ts
```

It's done in a useEffect() because we want to update the transport everytime the settings change. In the open function we have to make sure the id is vault-transport, because there is also IPC transport. That's why I also updated the `registerTransportModule` of `IPC`:

```js
+  const { token, enabled, host, workspace } = useSelector(vaultSignerSelector);
+  useEffect(() => {
+    if (enabled && currentMode !== "polling") {
+      setDeviceMode("polling");
+    } else if (!enabled) {
+      setDeviceMode(originalDeviceMode);
+    }
+    registerTransportModule({
+      id: "vault-transport",
+      open: (id: string) => {
+        if (id !== "vault-transport") return;
+        return retry(() =>
+          VaultTransport.open(host).then(transport => {
+            transport.setData({ token, workspace });
+            return Promise.resolve(transport);
+          }),
+        );
+      },
+      disconnect: () => Promise.resolve(),
+    });
+  }, [host, token, workspace, enabled]);
```

```diff
modified   apps/ledger-live-desktop/src/renderer/live-common-setup.ts
@@ -17,6 +17,9 @@ setEnvOnAllThreads("USER_ID", getUserId());
 // This defines our IPC Transport that will proxy to an internal process (we shouldn't use node-hid on renderer)
 registerTransportModule({
   id: "ipc",
-  open: id => retry(() => IPCTransport.open(id)),
+  open: (id: string) => {
+    if (id !== "ipc") return;
+    return retry(() => IPCTransport.open(id));
+  },
   disconnect: () => Promise.resolve(),
 });
```

Because the `useEffect()` is run everytime the `settings.vaultSigner` change, I updated the `registerTransportModules` to not always `push()` inside the array but only update the transport if it's already there.

```diff
modified   libs/ledger-live-common/src/hw/index.ts
@@ -42,7 +42,12 @@ export type TransportModule = {
 };
 const modules: TransportModule[] = [];
 export const registerTransportModule = (module: TransportModule): void => {
-  modules.push(module);
+  const index = modules.findIndex((m) => m.id === module.id);
+  if (index > -1) {
+    modules[index] = module;
+  } else {
+    modules.push(module);
+  }
 };
 export const discoverDevices = (
   accept: (module: TransportModule) => boolean = () => true
```

_Replace this text by a clear and concise description of what this pull request is about and why it is needed._

### ❓ Context

- **Impacted projects**: `ledger-live-desktop` `ledger-live-common` `ledgerjs/hw-transport-http`
- **Linked resource(s)**: [VG-14376](https://ledgerhq.atlassian.net/browse/VG-14376)

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery**
- [x] **No breaking changes** 

### 📸 Demo
cosmoslive

### 🚀 Expectations to reach

[VG-14376]: https://ledgerhq.atlassian.net/browse/VG-14376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/nasty-rats-raise.md
+++ b/.changeset/nasty-rats-raise.md
@@ -1,0 +1,7 @@
+---
+"@ledgerhq/hw-transport-http": minor
+"ledger-live-desktop": minor
+"@ledgerhq/live-common": minor
+---
+
+feat: LDD allow user to use vault as a signer
--- a/apps/ledger-live-desktop/src/renderer/Default.tsx
+++ b/apps/ledger-live-desktop/src/renderer/Default.tsx
@@ -5,6 +5,7 @@ import { Redirect, Route, Switch, useHistory } from "react-router-dom";
 import { useSelector } from "react-redux";
 import { FeatureToggle } from "@ledgerhq/live-common/featureFlags/index";
 import TrackAppStart from "~/renderer/components/TrackAppStart";
+import VaultSignerTransport from "~/renderer/components/VaultSignerTransport";
 import { BridgeSyncProvider } from "~/renderer/bridge/BridgeSyncContext";
 import { SyncNewAccounts } from "~/renderer/bridge/SyncNewAccounts";
 import Dashboard from "~/renderer/screens/dashboard";
@@ -51,6 +52,7 @@ import { ToastOverlay } from "~/renderer/components/ToastOverlay";
 import Drawer from "~/renderer/drawers/Drawer";
 import UpdateBanner from "~/renderer/components/Updater/Banner";
 import FirmwareUpdateBanner from "~/renderer/components/FirmwareUpdateBanner";
+import VaultSignerBanner from "~/renderer/components/VaultSignerBanner";
 import Onboarding from "~/renderer/components/Onboarding";
 import PostOnboardingScreen from "~/renderer/components/PostOnboardingScreen";
 import { hasCompletedOnboardingSelector } from "~/renderer/reducers/settings";
@@ -146,6 +148,7 @@ export default function Default() {
       <TriggerAppReady />
       <ExportLogsButton hookToShortcut />
       <TrackAppStart />
+      <VaultSignerTransport />
       <Idler />
       {process.platform === "darwin" ? <AppRegionDrag /> : null}
 
@@ -205,6 +208,7 @@ export default function Default() {
                           <TopBannerContainer>
                             <UpdateBanner />
                             <FirmwareUpdateBanner />
+                            <VaultSignerBanner />
                           </TopBannerContainer>
                           <Switch>
                             <Route path="/" exact component={Dashboard} />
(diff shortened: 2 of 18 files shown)


flocks:
PR recreated [here](https://github.com/LedgerHQ/ledger-live/pull/3287)

PR #3286, opened 2023-04-24
fix: update react-ui alert component

sarneijim:
### 📝 Description

Update react-ui alert component align with native to be used in web and lld

### ❓ Context

- **Impacted projects**: libs/ui/react
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-6890
- Associated Vercel deployment : https://react-ui-storybook-git-bugfix-update-react-ui-a-54405c-ledgerhq.vercel.app/?path=/story/messages-alerts--default&globals=backgrounds.value:!hex(131214)

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://user-images.githubusercontent.com/38540290/233985345-3bfdef3f-0859-4e48-90f6-7fc4a23e1c87.mov

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/polite-eyes-complain.md
+++ b/.changeset/polite-eyes-complain.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/react-ui": patch
+---
+
+Update alert component
--- a/.changeset/polite-eyes-complain.md
+++ b/.changeset/polite-eyes-complain.md
@@ -1,5 +1,5 @@
 ---
-"@ledgerhq/react-ui": patch
+"@ledgerhq/react-ui": minor
 ---
 
-Update alert component
+Update alert component. Add new secondary and sucess types. Change border-radius, colors and other css details.
--- a/libs/ui/packages/react/src/components/message/Alert/index.tsx
+++ b/libs/ui/packages/react/src/components/message/Alert/index.tsx
@@ -45,17 +45,17 @@ export interface AlertProps {
 }
 
 const StyledIconContainer = styled.div`
-  margin-right: 12px;
+  margin-right: 8px;
   display: flex;
   align-items: center;
 `;
 
 const icons = {
-  info: <InfoAltFillMedium size={20} />,
-  secondary: <InfoAltFillMedium size={20} />,
-  success: <CircledCheckSolidMedium size={20} />,
-  warning: <WarningSolidMedium size={20} />,
-  error: <CircledCrossSolidMedium size={20} />,
+  info: <InfoAltFillMedium size={24} />,
+  secondary: <InfoAltFillMedium size={24} />,
+  success: <CircledCheckSolidMedium size={24} />,
+  warning: <WarningSolidMedium size={24} />,
+  error: <CircledCrossSolidMedium size={24} />,
 };
 
 const getColors = ({ theme, type }: { theme: DefaultTheme; type?: AlertType }) => {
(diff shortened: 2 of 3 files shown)


mlegall-ledger (commented):
LGTM

libs/ui/packages/react/src/components/message/Alert/index.tsx
```diff
@@ -91,28 +103,33 @@ export default function Alert({
...
   const theme = useTheme();
-  const { color, background } = getColors({ theme, type });
+  const { iconColor, background } = getColors({ theme, type });
   const textProps: { variant?: TextVariants; fontWeight?: string } = {
     variant: "paragraph",
     fontWeight: "medium",
   };
+  const textColor = "neutral.c100";
   return (
     <StyledAlertContainer
```
ofreyssinet-ledger, line 113: The icon and the other content are supposed to be aligned at the top. You can change StyledAlertContainer and put `align-items: flex-start;`

ofreyssinet-ledger: Screenshot 2023-04-25 at 10 40 51

.changeset/polite-eyes-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/react-ui": patch
+---
+
+Update alert component
```
ofreyssinet-ledger, line 5: Please add a bit more info here regarding what was updated, also there is a breaking change with the `renderContent` prop typing changing so it's not retro-compatible, you might want to include that and use a `minor` bump. (Ideally it would be a `major` but for now we're not ready for 1.0.0)

ofreyssinet-ledger (changes requested):
Thanks for taking care of that!
Just two minor changes to do on the changeset and the UI.

libs/ui/packages/react/src/components/message/Alert/index.tsx
ofreyssinet-ledger, line 113: One more thing: the border radius has also changed in the design.

libs/ui/packages/react/src/components/message/Alert/index.tsx
```diff
@@ -91,28 +103,33 @@ export default function Alert({
   containerProps,
 }: AlertProps): JSX.Element {
   const theme = useTheme();
-  const { color, background } = getColors({ theme, type });
+  const { iconColor, background } = getColors({ theme, type });
   const textProps: { variant?: TextVariants; fontWeight?: string } = {
     variant: "paragraph",
     fontWeight: "medium",
   };
```
ofreyssinet-ledger, line 110: variant should now be `bodyLineHeight` and fontWeight `semiBold`

sarneijim: I have changed lineHeight to 21px too.

.changeset/polite-eyes-complain.md
sarneijim: I have updated the comment. Please let me know if you want to add some specific text.

sarneijim:
> In my opinion it would also be good to expose components `Alert.BodyText` and `Alert.UnderlinedText` [like I did here](https://github.com/LedgerHQ/ledger-live/pull/3251/files#diff-8ac7c79eb63e12962a7b569609a6190fe924917c04e4e6884e8d0af4e5a03399R113-R114), and maybe remove the `renderContent` prop, and remove the `textProps` object now that `textColor` and `textProps` are not dependent on the `type` prop.

Added

libs/ui/packages/react/src/components/message/Alert/index.tsx
ofreyssinet-ledger: let's check with designers if line height variants should all adopt this rather than making it a particular rule here for this specific case ...

ofreyssinet-ledger (approved):
lgtm 👍

libs/ui/packages/react/src/components/message/Alert/index.tsx
ofreyssinet-ledger: so it's back to 170% in the design, you can remove the places where you put 1.5
https://www.figma.com/file/WXCEmRMwW47ELrFonzPTGg?node-id=9509:42966#428079324

PR #3320, opened 2023-04-28
Rework of feature flagging on Add Account flows

lvndry:
### 📝 Description

Rework add account feature flagging in manager app.
We fetch all existing flags and check if any currency app is impacted by the feature flag

### ❓ Context

- **Impacted projects**: `LLD` `LLC`
- **Linked resource(s)**:
  - https://ledgerhq.atlassian.net/browse/LIVE-7344
  - https://ledgerhq.atlassian.net/browse/LIVE-5099
  - https://ledgerhq.atlassian.net/browse/LIVE-5100

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
@@ -35,21 +35,21 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
   const localOverrides = useSelector(overriddenFeatureFlagsSelector);
   const dispatch = useDispatch();
 
-  const getAllFlags = () => {
+  const getAllFlags = useCallback(() => {
     if (remoteConfig) {
       return getAll(remoteConfig);
     }
 
     return null;
-  };
+  }, [remoteConfig]);
 
-  const getAllCurrencyFlags = (): Record<string, Feature> | null => {
+  const getAllCurrencyFlags = useCallback((): Record<string, Feature> => {
     const allFlags = getAllFlags();
 
     if (allFlags) {
       const entries = Object.entries(allFlags)
         .filter(([key]) => {
-          return key.includes("_currency_");
+          return key.includes("feature_currency_");
         })
         .map(([key, value]) => {
           return [key, JSON.parse(value.asString())];
@@ -59,25 +59,28 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
     }
 
     return {};
-  };
+  }, [getAllFlags]);
 
-  const isFeature = (key: string): boolean => {
-    if (!remoteConfig) {
-      return false;
-    }
-
-    try {
-      const value = getValue(remoteConfig, formatToFirebaseFeatureId(key));
-
-      if (!value || !value.asString()) {
+  const isFeature = useCallback(
+    (key: string): boolean => {
+      if (!remoteConfig) {
         return false;
       }
-      return true;
-    } catch (error) {
-      console.error(`Failed to check if feature "${key}" exists`);
-      return false;
-    }
-  };
+
+      try {
+        const value = getValue(remoteConfig, formatToFirebaseFeatureId(key));
+
+        if (!value || !value.asString()) {
+          return false;
+        }
+        return true;
+      } catch (error) {
+        console.error(`Failed to check if feature "${key}" exists`);
+        return false;
+      }
+    },
+    [remoteConfig],
+  );
 
   const getFeature = useCallback(
     (key: FeatureId, allowOverride = true): Feature | null => {
--- a/apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
@@ -4,7 +4,7 @@ import semver from "semver";
 import { useDispatch, useSelector } from "react-redux";
 import { FeatureFlagsProvider } from "@ledgerhq/live-common/featureFlags/index";
 import { Feature, FeatureId } from "@ledgerhq/types-live";
-import { getValue } from "firebase/remote-config";
+import { getAll, getValue } from "firebase/remote-config";
 import { getEnv } from "@ledgerhq/live-common/env";
 import { formatToFirebaseFeatureId, useFirebaseRemoteConfig } from "./FirebaseRemoteConfig";
 import { overriddenFeatureFlagsSelector } from "../reducers/settings";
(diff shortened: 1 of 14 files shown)


apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -35,6 +35,32 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
   const localOverrides = useSelector(overriddenFeatureFlagsSelector);
   const dispatch = useDispatch();
 
+  const getAllFlags = () => {
```
lambertkevin, line 38: Please add a `useCallback`

apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -35,6 +35,32 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
...
+    if (remoteConfig) {
+      return getAll(remoteConfig);
+    }
+
+    return null;
+  };
+
+  const getAllCurrencyFlags = (): Record<string, Feature> | null => {
```
lambertkevin, line 46: Please add a `useCallback`

apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -35,6 +35,32 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
...
+          return [key, JSON.parse(value.asString())];
+        });
+
+      return Object.fromEntries(entries);
+    }
+
+    return {};
+  };
+
   const isFeature = (key: string): boolean => {
```
lambertkevin, line 64: I know it's not part of your modification, but add it here as well please 🙏

apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -35,6 +35,32 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
...
+    return null;
+  };
+
+  const getAllCurrencyFlags = (): Record<string, Feature> | null => {
+    const allFlags = getAllFlags();
+
+    if (allFlags) {
+      const entries = Object.entries(allFlags)
+        .filter(([key]) => {
+          return key.includes("_currency_");
```
lambertkevin, line 52: Would it work with `feature_currency_` instead ? To me, the more specific we are, the better.

(Example that I have in mind, if we decide to do the remote config with a `config_currency_XXX` 🤷‍♂️ )

lambertkevin (changes requested):
Only some remarks regarding the React memoization part, wait for another review anyway from someone who knows a little bit more about feature flags than me 😝

gre:
👋  we just merged https://github.com/LedgerHQ/ledger-live/pull/3359 which greatly improved the TypeScript checks on Ledger Live Desktop projects (components, modals, screens) and effectively make it type check on the CI. Could you rebase with it and ensure there is no issue on `pnpm desktop typecheck`? Thanks and don't hesitate to ask for any TS help.

apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -35,23 +35,52 @@ export const FirebaseFeatureFlagsProvider = ({ children }: Props): JSX.Element =
...
+      return getAll(remoteConfig);
     }
 
-    try {
-      const value = getValue(remoteConfig, formatToFirebaseFeatureId(key));
+    return null;
+  }, [remoteConfig]);
 
-      if (!value || !value.asString()) {
+  const getAllCurrencyFlags = useCallback((): Record<string, Feature> => {
```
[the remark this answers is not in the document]
lvndry: Yes I could use only only `getAllFlags`, however we will just loop over every feature flags in the manager app even though they are not currency feature flags

haammar-ledger (approved):
LGTM

apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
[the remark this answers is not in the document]
haammar-ledger: You're right, but I don't think there will ever be so many feature flags that there is a real perf impact.

apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
```diff
@@ -85,9 +83,21 @@ const Item = ({
     state.apps,
   ]);
 
-  const ffKey = currencyId ? formatCurrencyIdToFeatureKey(currencyId) : "";
-  const feature = useFeature(ffKey as FeatureId);
-  const currencyFlagEnabled = !feature || feature.enabled;
+  const { getAllFlags } = useFeatureFlags();
+  const currencyFlags = getAllFlags();
```
gre, line 87: ⚠️  i'm not familiar enough with firebase framework but are we sure this is not costly?
actually i'm really worried about the performance of this. 
- getAllFlags is implementing with a bunch of JSON.parse and it's fetching ALL flags where you instead want to only ask "is my currency concerned" (and you don't care about parsing the others).
- this is not secured by a useMemo, so this will run on each single re-render
- this will be running for each item in the AppsList then!🤔  

so we have AppsCount x FlagsCount number of JSON parse on each re-render.

gre (changes requested):
on the algorithmic code review of this, it's perf nogo to me.

also do we have ways to cover this code by test?

apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
lvndry: We can't add query parameters to only fetch currency flags on firebase. These are the only available methods: https://firebase.google.com/docs/reference/js/v8/firebase.remoteconfig.RemoteConfig#methods

My initial solution was to only parse the currency flags but an other PR comment recommended to do otherwise (https://github.com/LedgerHQ/ledger-live/pull/3320#discussion_r1184696787)

Indeed I should pass the flags to the items and not do the fetch from there I'll change that

gre (approved):
overall i'm ok with the way this remains contained but i have some reserve on the approach to `JSON.parse` a lot of things

PR #3194, opened 2023-04-11
Feat LIVE-4869 bitcoin sync perf

hzheng-ledger:
### 📝 Description
In order to improve btc sync performance. I've done follows
1. Remove generic pool lib for http connection pool which is not necessary
2. Pending / Non pending transaction retrieval code refactoring
3. Conditional reorg. (Reorg check only when current block heigh - account tx block heigh > threshold)
4. Promise.all to parallelise more http call

Bitcoin account sync performance comparison summary

1. http call from Ledger-live

for each bitcoin address: 
we had 3 https call from v4 explorer:
-.  /address/${address.address}/txs/pending
-.  /address/${address.address}/txs
-.  /block/${height} (only for incremental sync to check reorg)

During the PR, we check whether the reorg occurs globally once per account. 
We get the highest block of the account in the cache and calling /block/${height} only once per account.
As the reorg occurs rarely, in most cases of large account incremental sync, we can reduce 1/3 http calls.

2. Bots Portfolio
CPU user time optimisation
https://github.com/LedgerHQ/ledger-live/actions/runs/4960955067
6 total seeds of Portfolio bot
bitcoin testnet(53s)
bitcoin cash(34s)
bitcoin gold(32s)
zcash(20s)

vs. develop branch
https://github.com/LedgerHQ/ledger-live/actions/runs/4976884743
bitcoin testnet(68s)
bitcoin cash(40s)
bitcoin gold(37s)
zcash(23s)

sync time:
bitcoin testnet(318s)
bitcoin cash(201s)
bitcoin gold(200s)
zcash(97s)

vs. develop branch
https://github.com/LedgerHQ/ledger-live/actions/runs/4976884743
bitcoin testnet(575s)
bitcoin cash(219s)
bitcoin gold(249s)
zcash(133s)

### ❓ Context

- **Impacted projects**: `` LLC
- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-4869

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #3338, opened 2023-05-02
[LIVE-6970] Restore full network logs under env var + exp setting

chabroA:
### 📝 Description

Restore full network logs under env var + exp setting to help debugging and investigation issues, including in production.

Most of it is a revert of [this PR](https://github.com/LedgerHQ/ledger-live/pull/2419)

### ❓ Context

- **Impacted projects**: `LLC LLM LLD`
- **Linked resource(s)**: [LIVE-6970](https://ledgerhq.atlassian.net/browse/LIVE-6970)

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-6970]: https://ledgerhq.atlassian.net/browse/LIVE-6970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/families/bitcoin/wallet-btc/explorer/index.ts
+++ b/libs/ledger-live-common/src/families/bitcoin/wallet-btc/explorer/index.ts
@@ -1,12 +1,16 @@
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
 import type { AxiosRequestConfig } from "axios";
 import axios, { AxiosInstance } from "axios";
 import axiosRetry, { isNetworkOrIdempotentRequestError } from "axios-retry";
 import genericPool, { Pool } from "generic-pool";
+import { blockchainBaseURL } from "../../../../explorer";
+import {
+  errorInterceptor,
+  requestInterceptor,
+  responseInterceptor,
+} from "../../../../network";
 import { Address, Block, TX } from "../storage/types";
 import { IExplorer } from "./types";
-import { errorInterceptor } from "../../../../network";
-import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
-import { blockchainBaseURL } from "../../../../explorer";
 
 type ExplorerParams = {
   batch_size?: number;
@@ -74,7 +78,8 @@ class BitcoinLikeExplorer implements IExplorer {
     }
 
     // Logging
-    client.interceptors.response.use(undefined, errorInterceptor);
+    client.interceptors.request.use(requestInterceptor);
+    client.interceptors.response.use(responseInterceptor, errorInterceptor);
   }
 
   async broadcast(tx: string): Promise<{ data: { result: string } }> {
--- a/libs/ledger-live-common/src/families/stellar/api/horizon.ts
+++ b/libs/ledger-live-common/src/families/stellar/api/horizon.ts
@@ -1,21 +1,21 @@
+import { LedgerAPI4xx, LedgerAPI5xx, NetworkDown } from "@ledgerhq/errors";
+import type { Account, Operation } from "@ledgerhq/types-live";
 import { BigNumber } from "bignumber.js";
 import StellarSdk, {
   // @ts-expect-error stellar-sdk ts definition missing?
   AccountRecord,
-  NotFoundError,
   NetworkError,
+  NotFoundError,
 } from "stellar-sdk";
-import { getEnv } from "../../../env";
 import { getCryptoCurrencyById, parseCurrencyUnit } from "../../../currencies";
-import type { Account, Operation } from "@ledgerhq/types-live";
-import type { NetworkInfo } from "../types";
+import { getEnv } from "../../../env";
+import { requestInterceptor, responseInterceptor } from "../../../network";
 import {
   getAccountSpendableBalance,
-  rawOperationsToOperations,
   getReservedBalance,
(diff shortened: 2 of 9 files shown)


libs/ledger-live-common/src/families/bitcoin/wallet-btc/explorer/index.ts
haammar-ledger, line 1: Changes in this file will unfortunately conflict and won't be needed after https://github.com/LedgerHQ/ledger-live/pull/3194

libs/ledger-live-common/src/network.ts
```diff
@@ -1,15 +1,62 @@
...
+
+  const { baseURL, url, method = "", metadata } = response.config;
+  const { startTime = 0 } = metadata || {};
+
+  log(
+    "network-success",
+    `${response.status} ${method} ${baseURL || ""}${url} (${(
+      Date.now() - startTime
+    ).toFixed(0)}ms)`,
+    getEnv("DEBUG_HTTP_RESPONSE") ? { data: response.data } : undefined
```
haammar-ledger, line 49: I think you can remove this line (keep the env var as it's used by solana somehow) and log data too by default with the rest, when `ENABLE_NETWORK_LOGS` is true.

libs/ledger-live-common/src/network.ts
```diff
@@ -34,18 +81,28 @@ export const errorInterceptor = (error: AxiosError<any>): AxiosError<any> => {
...
+    );
     throw new NetworkDown();
   }
   throw error;
 };
 
-axios.interceptors.response.use(undefined, errorInterceptor);
+axios.interceptors.request.use(requestInterceptor);
+
+// $FlowFixMe LLD raise issues here
```
haammar-ledger, line 104: Is that comment needed anymore?
(Maybe not after https://github.com/LedgerHQ/ledger-live/pull/3359 ?)

libs/ledger-live-common/src/network.ts
chabroA: Done here 2dc0313cd591649a1646409af5a12eb7cc57c7e1

libs/ledger-live-common/src/network.ts
```diff
@@ -34,18 +81,28 @@ export const errorInterceptor = (error: AxiosError<any>): AxiosError<any> => {
     }
     log(
       "network-error",
-      `${status} ${method} ${baseURL || ""}${url}: ${errorToThrow.message}`,
+      `${status} ${method} ${baseURL || ""}${url} (${(
+        Date.now() - startTime
+      ).toFixed(0)}ms): ${errorToThrow.message}`,
```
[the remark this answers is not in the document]
chabroA: duration calculation and string generation extracted here af722555cd3ecfed60227086fbdbd9505f257ba2

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -30,7 +30,10 @@ export const BASE_RESERVE = 0.5;
 export const BASE_RESERVE_MIN_COUNT = 2;
 export const MIN_BALANCE = 1;
 
+StellarSdk.HorizonAxiosClient.interceptors.request.use(requestInterceptor);
+
 StellarSdk.HorizonAxiosClient.interceptors.response.use((response) => {
+  responseInterceptor(response);
```
henrily-ledger, line 36: nice catch on doing it in Stellar 👍

libs/ledger-live-common/src/network.ts
```diff
@@ -1,17 +1,65 @@
...
+    ).toFixed(0)}ms)`,
+    { data: response.data }
+  );
+
+  return response;
+};
 
 export const errorInterceptor = (error: AxiosError<any>): AxiosError<any> => {
-  const config = error?.response?.config || null;
+  const config = error?.response?.config as ExtendedXHRConfig | null;
```
henrily-ledger, line 55: not sure but if `|` is same as `||` ? I though `|` was more for typing

hedi-edelbloute: This doesn't look equivalent :grimacing: 

error?.response?.config returns undefined if config is not defined

Which means config before would be equal to null, now it's equal to undefined

image

Maybe the change has no impact though, but it needs to be checked

chabroA: It's not the same indeed. Here `|` is used because we are in the context of TS types (we are casting `error?.response?.config` as a `ExtendedXHRConfig | null` type).

Here, the type of `config` is `ExtendedXHRConfig | null`.

An alternative, although not identical, would be:
```ts
const config = (error?.response?.config as ExtendedXHRConfig) || null;
```
In this case, `error?.response?.config` is casted as a `ExtendedXHRConfig` type. The `|| null` is superflus since `config` will always have type `ExtendedXHRConfig` (because TS casting juste changes the type of a value, without checking if it works or not. This is not an operation).

chabroA: I fixed the typing here 59e26dabfdc2c003d256771c344a9437842bfac6 to get rid of this casting and improve clarity :+1:

PR #3371, opened 2023-05-05
chore: establishes LLD coin families typing

gre:
- [ ] properly document each method/component of LLDCoinFamily

in light of incoming fixing the TypeScript of src/renderer/families of LLD, we are establishing the base type of what is a LLD Coin Family.

### 📝 Description

- **`generated/` is gone!** in the process of fixing typescript of coin families folder, we remove completely this complex folder. We only need one `generated.ts` which is the result of all the `families/*/index.ts`
- each family export one **`index.ts`** and only exports a module object of type `LLDCoinFamily`.
- this unified type will allow us to more easily maintain the "Coin Family interface" to implement in LLD, easily deprecate things if we have to, as well as detect typescript issues.
- this unified type is self documenting the API you have to implement with comments on each component that you can define (instead of the current status quo of copying things from another family!)

#### Better typing with parametric binding specifics' Account, Transaction and TransactionStatus

The type is reinforced in that you can define your own specific account/transaction/status and it's about the only place where you are allowed to do these (inside your family). This does not involve casting at all!

```typescript
import { CeloAccount, Transaction, TransactionStatus } from "@ledgerhq/live-common/families/celo/types";
const family: LLDCoinFamily<CeloAccount, Transaction, TransactionStatus> = {
```

this will allow you to drop any code like `invariant(account.family === "celo")` and `account as CeloAccount`, you have typesafety on this and all components will directly be able to define `({ account: CeloAccount })` as props.

let's hope this gives idea to eventually generalize the same on the bridge interface!

this essentially works by the fact the type is defined as:

```
type LLDCoinFamily<
  A extends Account,
  T extends TransactionCommon,
  TS extends TransactionStatus,
  CoinModalsData = {}
>
```

#### NO BREAKING CHANGES from a family perspective...

an important choice of this rework was to NOT break any signature and existing design choices.

#### ...BUT just one file to add, and types to fix to it:

The impact is very limited. You just need to add a index.ts file. you can start with this template and remove the files you don't have:

```typescript
import "./live-common-setup";
import {
  _FAMILY_Account,
  Transaction,
  TransactionStatus,
} from "@ledgerhq/live-common/families/_FAMILY_/types";
import { LLDCoinFamily } from "../types";
import operationDetails from "./operationDetails";
import modals, { ModalsData } from "./modals";
import accountActions from "./accountActions";
import accountHeaderManageActions from "./AccountHeaderManageActions";
import transactionConfirmFields from "./TransactionConfirmFields";
import AccountBodyHeader from "./AccountBodyHeader";
import AccountSubHeader from "./AccountSubHeader";
import sendAmountFields from "./SendAmountFields";
import sendRecipientFields from "./SendRecipientFields";
import sendWarning from "./SendWarning";
import receiveWarning from "./ReceiveWarning";
import AccountBalanceSummaryFooter from "./AccountBalanceSummaryFooter";
import tokenList from "./TokenList";
import StepReceiveFunds from "./StepReceiveFunds";
import StepReceiveFundsPostAlert from "./StepReceiveFundsPostAlert";
import NoAssociatedAccounts from "./NoAssociatedAccounts";
import StakeBanner from "./StakeBanner";

const family: LLDCoinFamily<_FAMILY_Account, Transaction, TransactionStatus, ModalsData> = {
  operationDetails,
  accountActions,
  accountHeaderManageActions,
  modals,
  transactionConfirmFields,
  AccountBodyHeader,
  AccountSubHeader,
  sendAmountFields,
  sendRecipientFields,
  sendWarning,
  receiveWarning,
  AccountBalanceSummaryFooter,
  tokenList,
  StepReceiveFunds,
  StepReceiveFundsPostAlert,
  NoAssociatedAccounts,
  StakeBanner,
};

export default family;
```

#### Notes for future

the tradeoff of no breaking change is that, we should probably rework some parts. Here are a few ideas to consider in future:

- A bunch of components takes `{ account: A | TokenAccount | SubAccount, parentAccount: A | undefined | null }` props but it's possible they could just need `{ account: A }` where account would be the "parentAccount" in that case. For now, this code isn't touching this aspect.
  - in the meantime, `getMainAccount()` was improved to have a parametric `A` so we can preserve that A, where A extends Account.
- StepProps would be better if reworked with a parametric type from perspective of a coin.
- `export type AllCoinFamilies = Record<string, LLDCoinFamily<any, any, any>>;` is currently necessary due to the need to implement `getLLDCoinFamily<A extends Account, T extends TransactionCommon,  TS extends TransactionStatusCommon>(name: string): LLDCoinFamily<A, T, TS>`. In future, we could make this function return `LLDCoinFamily<Account,TransactionCommon,TransactionStatusCommon>` but it wouldn't work due to heavy usage of union type `Transaction`. **overall this is the only single place we allow the type to become loosely, but on both side we have typechecks working**

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7308 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveStakingFlow.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveStakingFlow.tsx
@@ -10,9 +10,10 @@ import { openURL } from "~/renderer/linking";
 import { withV3StyleProvider } from "~/renderer/styles/StyleProviderV3";
 import Button from "~/renderer/components/ButtonV3";
 import CheckBox from "~/renderer/components/CheckBox";
-import perFamilyManageActions from "~/renderer/generated/AccountHeaderManageActions";
 import { getAccountName } from "@ledgerhq/live-common/account/index";
 import { Account } from "@ledgerhq/types-live";
+import { getLLDCoinFamily } from "~/renderer/families";
+import { ManageAction } from "~/renderer/families/types";
 
 export const LOCAL_STORAGE_KEY_PREFIX = "receive_staking_";
 
@@ -35,27 +36,30 @@ const StepReceiveStakingFlow = (props: StepProps) => {
   const { t } = useTranslation();
   const receiveStakingFlowConfig = useFeature("receiveStakingFlowConfigDesktop");
   const [doNotShowAgain, setDoNotShowAgain] = useState<boolean>(false);
-  const [action, setAction] = useState<object>({});
+  // FIXME action is not used?
+  const [action, setAction] = useState<ManageAction | undefined>();
   const [title, setTitle] = useState<string>("");
   const [description, setDescription] = useState<string>("");
   const { account, parentAccount } = props;
 
   const id = account && "currency" in account ? account.currency?.id : undefined;
   const supportLink = id ? receiveStakingFlowConfig?.params[id]?.supportLink : undefined;
+
   const manage =
-    account && "currency" in account
-      ? perFamilyManageActions[account.currency.family as keyof typeof perFamilyManageActions]
-      : undefined;
+    account && account.type === "Account"
+      ? getLLDCoinFamily(account.currency.family).accountHeaderManageActions
+      : null;
+
   const familyManageActions =
     account && manage ? manage({ account: account as Account, parentAccount }) : undefined;
 
   useEffect(() => {
     const manageList =
       familyManageActions && familyManageActions.length > 0 ? familyManageActions : [];
-    // @ts-expect-error ts seems to have issues with union of arrays of different types
     const newAction = manageList && manageList.find(item => item.key === "Stake");
     const newTitle = t(`receive.steps.staking.${id}.title`);
     const newDescription = t(`receive.steps.staking.${id}.description`);
+    // FIXME fix this code. https://ledgerhq.atlassian.net/browse/LIVE-7343
     if (JSON.stringify(title) !== JSON.stringify(newTitle)) {
       setTitle(newTitle);
     }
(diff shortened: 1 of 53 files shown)


apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveStakingFlow.tsx
```diff
@@ -119,47 +123,41 @@ const StepReceiveStakingFlow = (props: StepProps) => {
...
-  );
+  const getTrackProperties = useCallback(() => {
+    return {
+      page: window.location.hash
+        .split("/")
+        .filter(e => e !== "#")
+        .join("/"),
+      flow: "stake",
+      currency: account && "currency" in account ? account?.currency?.name : undefined,
+      provider: providerName,
```
sarneijim, line 156: @gre we can remove it 
```suggestion
```

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveStakingFlow.tsx
```diff
@@ -119,47 +123,41 @@ const StepReceiveStakingFlow = (props: StepProps) => {
...
-    "currency" in account &&
-    perFamilyManageActions[account.currency.family as keyof typeof perFamilyManageActions];
+  const specific =
+    account && account.type === "Account" ? getLLDCoinFamily(account.currency.family) : null;
+  const manage = specific?.accountHeaderManageActions;
   const familyManageActions =
     manage && account && manage({ account: account as Account, parentAccount });
 
+  // @ts-expect-error FIXME: during our coin rework, we need to confirm this: i can't find a single case where provider is set
+  const providerName = (action && "provider" in action && action?.provider?.name) || "Ledger";
```
sarneijim, line 137:
```suggestion
```

PR #3380, opened 2023-05-09
fix(React Native BLE Transport): fix remapping of disconnected device ble errors

grsoares21:
### 📝 Description
Fix the remapping of errors in the React Native BLE Transport. Due to a recent change in the remapping code, the "was disconnected" and "not found" type of errors were not being correctly remapped into the DisconnectedDevice error class. 
This PR fixes this issue.

### ❓ Context

- **Impacted projects**: `ledgerjs/react-native-hw-transport-ble`
- **Linked resource(s)**: None

### ✅ Checklist

- [ ] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes** -- there will only be breaking changes if a dependent package was relying on that wrong behavior. Which is most probably not the case as the problem was only introduced very recently.

### 📸 Demo
N/A

### DIFF
--- a/.changeset/popular-nails-laugh.md
+++ b/.changeset/popular-nails-laugh.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/react-native-hw-transport-ble": patch
+---
+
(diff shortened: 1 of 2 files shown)


alexandremgo (approved):
Thanks !

PR #3423, opened 2023-05-12
feat: add deeplink clicked track

sarneijim:
### 📝 Description

Add deeplink clicked track

### ❓ Context

- **Impacted projects**: LLD
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7003
### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
+++ b/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
@@ -95,7 +95,7 @@ export function useDeepLinkHandler() {
       const fullUrl = pathname.replace(/(^\/+|\/+$)/g, "");
       const [url, path] = fullUrl.split("/");
       // Track deeplink only when ajsPropSource attribute exists.
-      const { ajs_prop_source: ajsPropSource } = query;
+      const ajsPropSource = searchParams.get("ajs_prop_source");
       if (ajsPropSource) {
         const { currency, installApp, appName } = query;
         track("deeplink_clicked", {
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
```diff
@@ -93,6 +94,19 @@ export function useDeepLinkHandler() {
       const query = Object.fromEntries(searchParams);
       const fullUrl = pathname.replace(/(^\/+|\/+$)/g, "");
       const [url, path] = fullUrl.split("/");
+      // Track a deeplink only when ajsPropSource attribute exists.
```
sarneijim, line 97:
```suggestion
      // Track deeplink only when ajsPropSource attribute exists.
```

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
```diff
@@ -93,6 +94,18 @@ export function useDeepLinkHandler() {
       const query = Object.fromEntries(searchParams);
```
kallen-ledger, line 94: I know this is unrelated so feel free to ignore for this PR. But searchParams looks to be an instance of [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Could we do something like the following below instead of using `Object.fromEntries`.


```
const ajsPropSource = searchParams.get("ajs_prop_source")
```

kallen-ledger: Or maybe even use the `has` method so theres no need to store `ajs_prop_source` in  a const.

```
if (searchParams.has("ajs_prop_source")) {
```

kallen-ledger (approved):
Minor questions but nothing to stop a merge of this ticket. 👍

PR #3444, opened 2023-05-15
Unify Tools (ESLint + Prettier)

valpinkman:
### 📝 Description

This PR aims to mimic https://github.com/LedgerHQ/ledger-live/pull/3195 but for ESLint + Prettier config throughout the whole projects in the monorepo. 

### ❓ Context

- **Impacted projects**: `monorepo`, `tools` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] **No breaking changes**: Some breaking changes will probably happen as we are centralizing tools to a single version, some things might need to be fixed (will be fixed when the PR is ready)

### 🚀 Expectations to reach


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "printWidth": 100,
```
chabroA, line 2: Curious about the rational behind changing the default value of 80 (which seems a sensible default value to me)?
cf. https://prettier.io/docs/en/options.html#print-width

.prettierrc
```diff
@@ -0,0 +1,5 @@
...
+  "trailingComma": "all",
+  "arrowParens": "avoid"
```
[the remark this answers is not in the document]
elbywan: I would say that this is a highly subjective point of view. For instance I do find it clearer without parenthesis, and I find the reasons in paragraph that you quoted dubious.

>  it becomes harder to add type annotations, extra arguments or default values as well as making other changes

Is it that hard to add a parenthesis manually when needed? And remove the visual noise otherwise automatically?

But in the end this kind of nitpicking over a tiny formatting style is completely unimportant to me. I would glady take whatever setting the whole team find best (even though I am sure most people actually do not care at all).

.prettierrc
elbywan: _Please note that the .prettierrc files existed before this PR with the width set to 100 already. So I'm not sure who changed the setting originally._

Anyway I can try to formulate an argument. Since we are in 2023, with our shiny retina screens our favourite code editors can display at least 130 characters / line even with the sidebar open (and with a small 13" screen). 80 seems really contrived and leave of lot of unused space.

But ultimately (and just as the other thing you raised) this is really not important to me. So whatever, if we vote in the team and people are happy with using 1/3 of their screen space I'm fine with it.

aboissiere-ledger: So great, we should get an agreement quickly then.
One thing sure: We can't push code modifying other developers' code without consulting them. 

If you don't want this PR to be blocked, let's override at first prettier rules for the projects that used other conventions so there is no impact at all on them and then we can review the shared rules with everyone.

gre:
=> we need to have 2 commits that we merge as is
- a commit that provide the config change
- a commit that apply it on all projects

=> we need to provide instruction on how to apply on a branch the changes so it limit the impact of conflicts.

- cherry-pick config
- apply lint + commit
- git rebase origin develop

gre:
template proposal
---

👋 ESLint config has been unified on develop after we merged https://github.com/LedgerHQ/ledger-live/pull/3444 – this will improve the project consistency, remove eslint dependencies duplicates, and overall simplify the maintenance. This comes with a prettier library upgrade and unification of the prettierrc too, so you likely have conflicts with develop.

**Step 1: cherry-pick the config commit and apply it to your branch**

take the config
```
git cherry-pick 89ca4e75af9fa0b72c07b9739069841dd8f5ff34
```

apply the config

```
pnpm lint:fix --force
```

- if your branch were continuously rebased with develop and don't have any merge commit, you can do `git rebase origin develop`
- otherwise prefer `git merge origin/develop`

haammar-ledger (approved):

PR #3445, opened 2023-05-15
feat(llm/synconboarding/analytics): track seeding

ofreyssinet-ledger:
### 📝 Description

Tracking of the properties of the seeding step (number of words and type of seeding).
This one is a tiny bit more complex than other tracking logic because this information cannot be obtained directly from the device once it is seeded, so _during_ the onboarding we have to keep track of the onboarding state returned by the device and use the values we kept to track the event with the correct properties.

For Recover, the implementation will have to be completed separately, as specified in comments on the ticket.

### ❓ Context

- **Linked resource(s)**: [LIVE-6993] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-6993]: https://ledgerhq.atlassian.net/browse/LIVE-6993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/analytics/segment.ts
+++ b/apps/ledger-live-mobile/src/analytics/segment.ts
@@ -338,13 +338,44 @@ export const useAnalytics = () => {
   };
 };
 export const screen = async (
+  /**
+   * First part of the event name string
+   */
   category?: string,
+  /**
+   * Second part of the event name string, will be concatenated to `category`
+   * after a whitespace if defined.
+   */
   name?: string | null,
+  /**
+   * Event properties
+   */
   properties?: Record<string, unknown> | null | undefined,
+  /**
+   * Should this function call update the previous & current route names, route names are used to track:
+   * - the `screen` property in non-screen events (for instance `button_clicked` events)
+   * - the `source` property in further screen events
+   */
+  updateRoutes?: boolean,
+  /**
+   * Should this function call update the current route name.
+   * If true, it means that the full screen name (`category` + " " + `name`) will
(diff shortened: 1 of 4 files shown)


alexandremgo (approved):
Looks good !

mlegall-ledger (approved):

apps/ledger-live-mobile/src/analytics/segment.ts
```diff
@@ -337,14 +337,53 @@ export const useAnalytics = () => {
...
   };
 };
+
+/**
+ * Track an event which will have the name `Page ${category}${name ? " " + name : ""}`.
+ * Extra logic to update the route names used in "screen" and "source"
+ * properties of further events can be optionally enabled with the parameters
+ * `updateRoutes` and `refreshSource`.
+ */
 export const screen = async (
```
alexandremgo, line 347: If the parameters of this `screen` function are all optional, it can be interesting to put them in an object, so the consumer can choose which parameter to set and not set:

```
screen({category, name, properties, etc. }: { category?: string; name?: string | null; properties?: etc. }
```

WDYT ?

ofreyssinet-ledger: Usages of this are pretty rare/exceptional so I think it's not too inconvenient to keep the signature like this for now.

PR #3456, opened 2023-05-16
fix: correct logic for showing the provider warning ui on LLM

juan-cortes:
### 📝 Description

When you have either an odd provider or developer mode enabled on LLM it should be showing the warning while inside My Ledger. In the first implementation which I hastily merged, I was only showing the banner if the provider was not 1. Meaning that in the specific case of the provider being 1, but the developer mode being ON, there would be no warning. This PR addresses that.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7301` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
Group 1

### 🚀 Expectations to reach
Testing with any combination of the settings should match the expected screenshots.

### DIFF
--- a/.changeset/curly-bottles-pump.md
+++ b/.changeset/curly-bottles-pump.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Fix bad logic for ProviderWarning visibility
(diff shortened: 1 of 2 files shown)


ofreyssinet-ledger (approved):
👍

PR #3457, opened 2023-05-16
fix: allow closing device action modals when no blocking action

juan-cortes:
### 📝 Description

For the sake of this task, the only thing that we need to check is that upon selecting a device from the device selection screen, during the Loading step or any other step that is not blocking, we should still see the X on the top right and be allowed to dismiss the modal. A regression introduced a few months ago meant that we had to wait until a timeout error in order to close the modal, leading to a sub-par user experience if they chose the wrong device.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7435` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/4631227/2f278717-7f45-4b64-8650-1aed319606dd

### 🚀 Expectations to reach

The easiest way to test this is to follow what the video does, select a device that is not available and see if you can close the modal while it _tries_ to connect to said device. A future task that covers all steps would be useful but for now, most blocking steps are already blocking and this should introduce no changes to them.

### DIFF
--- a/apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
+++ b/apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
@@ -531,6 +531,7 @@ const AllowOpeningApp = ({
           })}
         </CenteredText>
       ) : null}
+      <ModalLock />
     </Wrapper>
   );
 };
@@ -922,8 +923,10 @@ export function renderConnectYourDevice({
 export function renderLoading({
   t,
   description,
+  lockModal = false,
 }: RawProps & {
   description?: string;
+  lockModal?: boolean;
 }) {
   return (
     <Wrapper>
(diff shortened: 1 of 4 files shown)


apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -922,16 +923,18 @@ export function renderConnectYourDevice({
 export function renderLoading({
   t,
   description,
+  lockModal = false,
```
ofreyssinet-ledger, line 926: As a precaution I would have made the default to true and turned it off for specific cases where we're sure it doesn't break, it would be enough to cover the scope of the ticket.
But seeing that you've made changes in other specific cases like the swap confirmation modal, I trust you to have tested all usages of this 👍

ofreyssinet-ledger (approved):

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
juan-cortes: I trust me too

PR #3460, opened 2023-05-16
feat: add network error component and error fallback for webview

mitchellv-ledger:
### 📝 Description

Adds an error handler and error screen for webview when the external web application fails to load properly.  When try again is pressed it imperatively refreshes the webview via its ref.

### ❓ Context

- **Impacted projects**: `PTX`
- **Linked resource(s)**: `LIVE-7063` 

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 📸 Demo

Screenshot 2023-05-16 at 16 36 48

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
@@ -1,0 +1,53 @@
+import React, { useCallback } from "react";
+import { Button, Flex, Icon, Link, Text } from "@ledgerhq/react-ui";
+import { useTranslation } from "react-i18next";
+import ErrorNoBorder from "~/renderer/icons/ErrorNoBorder";
+import { urls } from "~/config/urls";
+import { openURL } from "~/renderer/linking";
+
+export const NetworkErrorScreen = ({ refresh }: { refresh: () => void }) => {
+  const { t } = useTranslation();
+
+  const handleContactSupport = useCallback(() => openURL(urls.contactSupportWebview), []);
+
+  return (
+    <Flex alignItems="center" justifyContent="center" flexGrow={1} flexDirection="column">
+      <ErrorNoBorder size={64} />
+      <Text
+        variant="h4Inter"
+        marginTop="16px"
+        fontSize={24}
+        fontWeight="semiBold"
+        fontFamily="Inter"
+      >
+        {t("webview.networkError.title")}
+      </Text>
+      <Text variant="body" color="neutral.c70" marginTop="24px" fontSize={14} fontFamily="Inter">
+        {t("webview.networkError.subtitle")}
+      </Text>
+      <Text variant="body" color="neutral.c70" marginTop="0px" fontSize={14} fontFamily="Inter">
+        {t("webview.networkError.subtitleTwo")}
+      </Text>
+      <Button
+        variant="main"
+        width={143}
+        height={40}
+        marginTop="24px"
+        onClick={refresh}
+        borderRadius="44px"
+      >
+        {t("webview.networkError.tryAgain")}
+      </Button>
+      <Link
+        size="large"
+        Icon={() => <Icon name="ExternalLink" />}
+        marginTop="17px"
+        onClick={handleContactSupport}
+      >
+        <Text fontSize={14} fontWeight="bold" fontFamily="Inter">
+          {t("webview.networkError.contactSupport")}
+        </Text>
+      </Link>
+    </Flex>
+  );
+};
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
@@ -61,7 +61,7 @@ export const NetworkErrorScreen = ({ refresh }: { refresh: () => void }) => {
         customComponent={(handleSaveLogPress: () => Promise<void>) => (
           <Link
             size="large"
-            Icon={() => <Icon name="Download" />}
+            Icon={() => <Icon name="Import" />}
             marginTop="17px"
             onClick={handleSaveLogPress}
           >
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
@@ -4,11 +4,22 @@ import { useTranslation } from "react-i18next";
 import ErrorNoBorder from "~/renderer/icons/ErrorNoBorder";
 import { urls } from "~/config/urls";
 import { openURL } from "~/renderer/linking";
+import { languageSelector } from "~/renderer/reducers/settings";
+import { useSelector } from "react-redux";
 
 export const NetworkErrorScreen = ({ refresh }: { refresh: () => void }) => {
   const { t } = useTranslation();
+  const locale = useSelector(languageSelector) || "en";
 
-  const handleContactSupport = useCallback(() => openURL(urls.contactSupportWebview), []);
+  const handleContactSupport = useCallback(
+    () =>
+      openURL(
+        urls.contactSupportWebview[
+          locale in urls.terms ? (locale as keyof typeof urls.terms) : "en"
+        ],
+      ),
+    [locale],
+  );
 
   return (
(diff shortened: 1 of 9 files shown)


apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
```diff
@@ -193,6 +209,8 @@ export function useWebviewState(
     webview.addEventListener("did-start-loading", handleDidStartLoading);
     webview.addEventListener("did-stop-loading", handleDidStopLoading);
     webview.addEventListener("dom-ready", handleDomReady);
+    webview.addEventListener("did-fail-load", handleFail);
+    webview.addEventListener("crashed", handleFail);
```
cpellizzari-ledger, line 213: with did-fail-load, we could have more data : https://www.electronjs.org/docs/latest/api/webview-tag#event-did-fail-load. Also I believe you need to filter on isMainFrame property, even if not really well documented... We do not want to have global error if an iframe fails. So I would handle both events with 2 separate handlers. On both events, I would like to have a specific log with the specific event fields, so the CS may get data.

Please check what is the behavior when you get a 403 error or a 500 error. Does it trigger did-fail-load on 4XX/5XX ?

cpellizzari-ledger (changes requested):

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
mitchellv-ledger: Will look into this 👌

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Button, Flex, Icon, Link, Text } from "@ledgerhq/react-ui";
+import { useTranslation } from "react-i18next";
+import ErrorNoBorder from "~/renderer/icons/ErrorNoBorder";
+import { urls } from "~/config/urls";
+import { openURL } from "~/renderer/linking";
+
+export const NetworkErrorScreen = ({ refresh }: { refresh: () => void }) => {
+  const { t } = useTranslation();
+
+  const handleContactSupport = useCallback(() => openURL(urls.contactSupportWebview), []);
```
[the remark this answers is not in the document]
mitchellv-ledger: 😆 that was true, but now have added language support which depends on redux

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const NetworkErrorScreen = ({ refresh }: { refresh: () => void }) => {
+  const { t } = useTranslation();
+  const locale = useSelector(languageSelector) || "en";
+
+  const handleContactSupport = useCallback(
+    () =>
+      openURL(
+        urls.contactSupportWebview[
+          locale in urls.terms ? (locale as keyof typeof urls.terms) : "en"
```
kallen-ledger, line 18: Should this be?

```suggestion
          locale in urls.contactSupportWebview ? (locale as keyof typeof urls.contactSupportWebview) : "en"
```

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
```diff
@@ -166,18 +170,49 @@ export function useWebviewState(
...
+      isAppUnavailable: oldState.isAppUnavailable,
+    }));
+  }, [webviewRef]);
+
+  const handleFailLoad = (useCallback(
+    (errorCode: number, errorDescription: string, validatedURL: string, isMainFrame: boolean) => {
+      const errorInfo = { errorCode, errorDescription, validatedURL, isMainFrame };
+      console.error("Web3AppView handleFailLoad", errorInfo);
+
+      track("useWebviewState", errorInfo);
```
cpellizzari-ledger, line 192: Avoid sending too much "raw" data to tracker, for privacy concerns.

errorCode = no pb
errorDescription = we don't know what websites will put there, should be removed
validatedURL = we know there are wallet addresses for instance, please retain only the domain & subdomain (drop path, query params...)

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useTranslation } from "react-i18next";
+import ErrorNoBorder from "~/renderer/icons/ErrorNoBorder";
+import { urls } from "~/config/urls";
+import { openURL } from "~/renderer/linking";
+import { languageSelector } from "~/renderer/reducers/settings";
+import { useSelector } from "react-redux";
+
+export const NetworkErrorScreen = ({ refresh }: { refresh: () => void }) => {
+  const { t } = useTranslation();
+  const locale = useSelector(languageSelector) || "en";
```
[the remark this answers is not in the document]
mitchellv-ledger: this is a separate fallback to the one on line 18

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
mitchellv-ledger: I've addressed this issue in new commit

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -291,6 +294,9 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
 
     return (
       <>
+        {!webviewState.loading && webviewState.isAppUnavailable && (
+          <NetworkErrorScreen refresh={handleRefresh} />
+        )}
```
kallen-ledger, line 299: [Question]: Are we still rendering the webview component in this case? If so do we want to do that or should we only render the webview if there is no network error?

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <Flex alignItems="center" justifyContent="center" flexGrow={1} flexDirection="column">
+      <ErrorNoBorder size={64} />
+      <Text
+        variant="h4Inter"
+        marginTop="16px"
+        fontSize={24}
+        fontWeight="semiBold"
+        fontFamily="Inter"
```
kallen-ledger, line 32: [Question] Do we need to define Inter for both fontFamily and in the variant?

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
```diff
@@ -0,0 +1,64 @@
...
+      >
+        {t("webview.networkError.title")}
+      </Text>
+      <Text variant="body" color="neutral.c70" marginTop="24px" fontSize={14} fontFamily="Inter">
+        {t("webview.networkError.subtitle")}
+      </Text>
+      <Text variant="body" color="neutral.c70" marginTop="0px" fontSize={14} fontFamily="Inter">
+        {t("webview.networkError.subtitleTwo")}
+      </Text>
```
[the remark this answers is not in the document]
mitchellv-ledger: yes, to make sure the spacing is correct

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/NetworkError.tsx
mitchellv-ledger: I'm not entirely sure tbh.  IIRC it wasn't displaying with the correct font family unless I did this

kallen-ledger (approved):

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
kallen-ledger: Thanks 👍

cpellizzari-ledger (approved):

cpellizzari-ledger (approved):

PR #3488, opened 2023-05-22
fix: bad rendering of hash verification on old fw update step

juan-cortes:
### 📝 Description

Fix overflow of the fw update verification hash for old firmware update flows.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7523` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo
image

### 🚀 Expectations to reach
Follow a firmware update of an old path (1.4.2 on lns for instance) check that the identifier doesn't overflow. Depending on the version there may be some ellipsis, some multi-line, etc. It should match what the device shows in any case.

### DIFF
--- a/.changeset/clever-dolls-lie.md
+++ b/.changeset/clever-dolls-lie.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Fix bad rendering for old firmware update hash veritification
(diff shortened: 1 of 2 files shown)


alexandremgo (approved):
👌

PR #3491, opened 2023-05-23
[LIVE-7531]: update Manifest types, add helper hook and util funcs

kallen-ledger:
### 📝 Description

Noticed during the development of a separate feature that the type for `LiveAppManifest["params"]` does not line up with what I was seeing in manifest files.

* Update type to include all combinations found. Maintain the current type of `Array<string>` so as not to need a major version bump.
* Add helper hook `useManifest` to encapsulate logic that was happening on both LLD and LLM repeatedly.
* Add helper utils in new `util` dir.
  * `appendQueryParamsToDappURL`: Takes a manifest and a queryParams object and appends it to the manifests dappUrl if it exists
  * `extractDappURLFromManifest`: Extracts the dappUrl from the manifest if it exists. 
### ❓ Context

- **Impacted projects**: `libs/ledger-live-common`

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach
https://ledgerhq.atlassian.net/browse/LIVE-7531

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/platform/App.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/platform/App.tsx
@@ -53,14 +53,11 @@ export default function PlatformApp({ match, appId: propsAppId, location }: Prop
   const localManifest = useLocalLiveAppManifest(appId);
   const remoteManifest = useRemoteLiveAppManifest(appId);
   let manifest = localManifest || remoteManifest;
-  if (customDappUrl) {
+  if (customDappUrl && manifest && manifest.params && "dappUrl" in manifest.params) {
     manifest = {
       ...manifest,
       params: {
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/src/platform/utils/appendQueryParamsToDappURL.ts
```diff
@@ -0,0 +1,29 @@
+import { LiveAppManifest } from "../types";
+import { extractDappURLFromManifest } from "./extractDappURLFromManifest";
+
+export function appendQueryParamsToDappURL(
```
sarneijim, line 4: Where do we plan export these utils? Is this the scope part of a different pr?

PR #3497, opened 2023-05-24
Feat: LIVE-7290: toggle onboarding early checks apdu

alexandremgo:
### 📝 Description

Introducing a new device action (implemented in the device SDK): `toggleOnboardingEarlyCheckAction` with
- its associated new command and task
- its associated new onboarding state
- a hook `useToggleOnboardingEarlyCheck` for simple usage on LLM and LLD
- unit tests
- its associated cli command `deviceSDKToggleOnboardingEarlyCheck`

This new action uses a new APDU to enter and exit the "early security check" blocking step during the onboarding of Stax.

Also implementing the new architecture on LLM for the sync onboarding:
- a new `SyncOnboardingCompanion` component getting (almost) all the logic that existed directly in the screen
- an empty `EarlySecurityCheck` component that directly notifies (for now) that everything is good
- a screen that orchestrate which of the 2 components should be rendered by checking the onboarding state of the device, and calling the new device action  `toggleOnboardingEarlyCheckAction` when needed

### ❓ Context

- **Impacted projects**: `common`, `mobile`, `cli`
- **Linked resource(s)**: [LIVE-7290](https://ledgerhq.atlassian.net/browse/LIVE-7290) & [LIVE-7593](https://ledgerhq.atlassian.net/browse/LIVE-7593)

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

<details>
  <summary>Demo on Android : toggling ESC and trying to force desync</summary>

https://github.com/LedgerHQ/ledger-live/assets/15096913/47b0bb6b-a848-46e3-8b2a-468bdc877c0a

</details>
### 🚀 Expectations to reach

[LIVE-7290]: https://ledgerhq.atlassian.net/browse/LIVE-7290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/screens/SyncOnboarding/SyncOnboardingCompanion.tsx
+++ b/apps/ledger-live-mobile/src/screens/SyncOnboarding/SyncOnboardingCompanion.tsx
@@ -1,0 +1,729 @@
+import React, {
+  useEffect,
+  useState,
+  useCallback,
+  useMemo,
+  ReactNode,
+  useRef,
+  useLayoutEffect,
+} from "react";
+import {
+  Flex,
+  VerticalTimeline,
+  Text,
+  ContinueOnDevice,
+  Divider,
+} from "@ledgerhq/native-ui";
+import { useSafeAreaInsets } from "react-native-safe-area-context";
+import { useOnboardingStatePolling } from "@ledgerhq/live-common/onboarding/hooks/useOnboardingStatePolling";
+import {
+  OnboardingStep as DeviceOnboardingStep,
+  fromSeedPhraseTypeToNbOfSeedWords,
+} from "@ledgerhq/live-common/hw/extractOnboardingState";
+import { useTranslation } from "react-i18next";
+import { getDeviceModel } from "@ledgerhq/devices";
+import { Device } from "@ledgerhq/live-common/hw/actions/types";
+import { useDispatch } from "react-redux";
+import useFeature from "@ledgerhq/live-common/featureFlags/useFeature";
+
+import { SeedPhraseType, StorylyInstanceID } from "@ledgerhq/types-live";
+import { DeviceModelId } from "@ledgerhq/types-devices";
+import { addKnownDevice } from "../../actions/ble";
+import { ScreenName } from "../../const";
+import HelpDrawer from "./HelpDrawer";
+import ResyncOverlay from "./ResyncOverlay";
+import SoftwareChecksStep from "./SoftwareChecksStep";
+import {
+  completeOnboarding,
+  setHasOrderedNano,
+  setLastConnectedDevice,
+  setReadOnlyMode,
+} from "../../actions/settings";
+import InstallSetOfApps from "../../components/DeviceAction/InstallSetOfApps";
+import Stories from "../../components/StorylyStories";
+import { TrackScreen, screen } from "../../analytics";
+import ContinueOnStax from "./assets/ContinueOnStax";
+import type { SyncOnboardingScreenProps } from ".";
+
+const { BodyText, SubtitleText } = VerticalTimeline;
+
+type StepStatus = "completed" | "active" | "inactive";
+
+type Step = {
+  key: CompanionStepKey;
+  status: StepStatus;
+  title: string;
+  doneTitle?: string;
+  estimatedTime?: number;
+  renderBody?: (isDisplayed?: boolean) => ReactNode;
+};
+
+export type SyncOnboardingCompanionProps = {
+  /**
+   * A `Device` object
+   */
+  device: Device;
+  /**
+   * A react-navigation instance to handle navigation once the device is ready, when there is a desync
+   * issue, and to update the react-navigation header with available languages
+   */
+  navigation: SyncOnboardingScreenProps["navigation"];
+  /**
+   * Called when the polling from the companion component has definitely lost/is desync with the device
+   */
+  onLostDevice: () => void;
+
+  /**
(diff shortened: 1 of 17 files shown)


libs/ledger-live-common/src/deviceSDK/commands/toggleOnboardingEarlyCheck.ts
```diff
@@ -0,0 +1,42 @@
...
+  transport,
+  p2,
+}: {
+  transport: Transport;
+  p2: ToggleTypeP2;
+}): Observable<ToggleOnboardingEarlyCheckCmdEvent> {
+  return new Observable((subscriber) => {
+    transport
+      .send(0xe0, 0x03, 0x00, p2)
+      .then(() => {
```
grsoares21, line 34: Doesn't the answer matter here? Are we sure that this promise will throw in case of an error and not just resolve with a specific error code or something like that?

grsoares21 (approved):
Just a small question, but really well done! Congrats, specially on the tests ;)

libs/ledger-live-common/src/deviceSDK/commands/toggleOnboardingEarlyCheck.ts
alexandremgo: It's true that looking at it, we can't know that it would always fail/throw if the response is not `0x9000`

But there is a check in `send` on a list of status code `statusList`, and by default this list only contains `0x9000`: 

```TS
  send = async (
    cla: number,
    ins: number,
    p1: number,
    p2: number,
    data: Buffer = Buffer.alloc(0),
    statusList: Array<number> = [StatusCodes.OK]
  ): Promise<Buffer> => { }
```

So we're sure here in the `then` that it's a success.

Also the tests in `libs/ledger-live-common/src/deviceSDK/commands/toggleOnboardingEarlyCheck.test.ts` double check this ;)

apps/ledger-live-mobile/src/screens/SyncOnboarding/SyncOnboardingCompanion.tsx
```diff
@@ -0,0 +1,729 @@
...
+   */
+  onShouldHeaderBeOverlaid: (shouldBeOverlaid: boolean) => void;
+
+  /**
+   * Updates any existing delay before displaying the hiding the header below an overlay
+   */
+  updateHeaderOverlayDelay: (delayMs: number) => void;
+
+  /**
+   * Called when the companion component thinks the device is not in a correct state anymore
```
ofreyssinet-ledger, line 88: breaking news: we have AGI!

libs/ledger-live-common/src/deviceSDK/hooks/useToggleOnboardingEarlyChecks.ts
```diff
@@ -0,0 +1,71 @@
...
+ *  Useful to enable/disable this hook.
+ * @param toggleOnboardingEarlyCheckAction dependency injected action. A default implementation is provided.
+ * @returns an object containing the state of the onboarding early check toggling with possible error.
+ *   The resulting state is reset to its initial state on each new triggering of the toggle action (when `toggleType` is updated):
+ * {
+ *   toggleStatus: "none",
+ *   ...initialSharedActionState,
+ * }
+ */
+export const useToggleOnboardingEarlyCheck = ({
```
ofreyssinet-ledger, line 39: With this approach, the command is only sent to the device (for a fixed device and implementation) if the `"toggleType"` changes, is that something that is ok? 
For instance consider this scenario:
- in my component I want to use this command once by setting `enter` in an internal state and passing that state to this hook
- the command fails and the hook returns a "failure" status (for instance because the device wasn't in the correct state)
- what can I do to re-trigger the command if I want ?

2 solutions I see:
- Could there be at least a nonce parameter or something to "force re-trigger" the command?
- Another approach for this hook would be: instead of triggering the command based on a state, just make this hook return a function that does the triggering. Then it's up to the consumer to implement the way they want to trigger that command (they can use an internal state and a useEffect in a similar fashion if they want, they can also just use the function returned by the hook and call it whenever they want)

But maybe the scenario above is not likely to happen in our use cases so we don't care 🤷‍♂️

apps/ledger-live-mobile/src/screens/SyncOnboarding/index.tsx
```diff
@@ -77,660 +31,225 @@ export type SyncOnboardingCompanionProps = CompositeScreenProps<
...
   }, []);
 
-  const handleDesyncRetry = useCallback(() => {
-    // handleDesyncClose is then called
-    track("button_clicked", {
-      button: "Try again",
-      drawer: "Could not connect to Stax",
-    });
-    setDesyncDrawerOpen(false);
+  // Called when the companion component thinks the device is not in a correct state anymore
```
ofreyssinet-ledger, line 108: is it a sentient component? :trollface:

libs/ledger-live-common/src/deviceSDK/hooks/useToggleOnboardingEarlyChecks.ts
ofreyssinet-ledger: actually I can recover by setting to `null` and then back to the same value so maybe it's fine like that

ofreyssinet-ledger (approved):
lgtm, great job! 🎉

PR #3509, opened 2023-05-24
[LIVE-7401] Feat - Add NFTs to EVM family sync

lambertkevin:
### 📝 Description

draft

### ❓ Context

- **Impacted projects**: `@ledgerhq/ledger-live-common` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7401 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/44363395/48a8314d-79f4-4229-b4a5-6b7b55184cd9

### 🚀 Expectations to reach

Screenshot 2023-05-24 at 23 58 37


libs/ledger-live-common/src/families/evm/__tests__/fixtures/common.fixtures.ts
```diff
@@ -83,7 +87,7 @@ export const makeAccount = (
     },
   };
 
-  return account;
+  return Object.freeze(account);
```
hedi-edelbloute, line 90: Can you explain why we need all those Object.freeze ?

libs/ledger-live-common/src/families/evm/__tests__/fixtures/synchronization.fixtures.ts
```diff
@@ -0,0 +1,220 @@
...
+  makeAccount,
+  makeNft,
+  makeOperation,
+  makeTokenAccount,
+} from "./common.fixtures";
+
+jest.useFakeTimers().setSystemTime(new Date("2014-04-21"));
+
+export const currency: CryptoCurrency = Object.freeze({
+  ...getCryptoCurrencyById("ethereum"),
```
hedi-edelbloute, line 18: Shouldn't this be mocked too ?

libs/ledger-live-common/src/families/evm/__tests__/integration/bridge.integration.test.ts
```diff
@@ -1,21 +1,20 @@
...
 
 const dataset: DatasetTest<Transaction> = {
   implementations: ["mock", "js"],
   currencies: {
-    ethereum: {
-      scanAccounts: [ethereumScanAccounts1],
+    ethereum_as_evm_test_only: {
       accounts: [
         {
+          FIXME_tests: ["balance is sum of ops"],
```
hedi-edelbloute, line 17: By curiosity, is it failing ? If yes, do we have an idea why ?

libs/ledger-live-common/src/families/evm/api/etherscan.ts
```diff
@@ -109,16 +117,184 @@ export const getLastTokenOperations = makeLRUCache<
...
+      .filter(Boolean) as Operation[];
+  },
+  (currency, address, accountId, fromBlock) => accountId + fromBlock,
+  { ttl: 5 * 1000 }
+);
+
+/**
+ * Get all the last ERC721 transactions
+ */
+export const getLastERC721Operations = makeLRUCache<
```
hedi-edelbloute, line 129: We could cover getLastERC721Operations, getLastERC1155Operations and getLastNftOperations with unit tests

libs/ledger-live-common/src/families/evm/synchronization.ts
```diff
@@ -68,33 +72,38 @@ export const getAccountShape: GetAccountShape = async (infos) => {
...
+      try {
+        const { getLastOperations } = getExplorerApi(currency);
+        return await getLastOperations(
+          currency,
+          address,
+          accountId,
+          latestSyncedOperation?.blockHeight || 0
+        );
+      } catch (e) {
+        log("EVM Family", "Failed to get latest transactions", {
```
hedi-edelbloute, line 86: We could unit test if the error is thrown when either getExplorerApi or getLastOperations fails

hedi-edelbloute (approved):
Looks good overall, just minor unit testing remarks and Object.freeze magic that I want to understand

libs/ledger-live-common/src/families/evm/__tests__/fixtures/common.fixtures.ts
lambertkevin: Mutating an object that is going to be used by our methods can be very problematic as the execution of a test can change the object for the next tests.
Object.freeze make sure that you can't mutate that Object (which is a good thing with React), so everytime you're using it with a function that mutates it, you'll have to use a copy, and all the tests are now safe to always use the Object you intented.

libs/ledger-live-common/src/families/evm/__tests__/integration/bridge.integration.test.ts
lambertkevin: Yes it is, no idea why but that's a very difficult test to have correct anyway. I'll have a look later to find out what is missing exactly (could be an unsupported operation type, could be some fee not accounted correctly), but that test is deactivated for Ethereum/Ripple/Celo/Cosmos/Filecoin/Near & Stellar, so I guess either it doesn't make sense anymore, or we need to rework it globally.

libs/ledger-live-common/src/families/evm/synchronization.ts
lambertkevin: It's covered here: https://github.com/LedgerHQ/ledger-live/blob/170bcd405fc999f25fc4cf8a1946a48cc53dc3aa/libs/ledger-live-common/src/families/evm/__tests__/unit/synchronization.unit.test.ts#L80

hedi-edelbloute: Indeed my bad, this is the idea :) but then technically speaking it's the sub method that is failing ?

libs/ledger-live-common/src/families/evm/api/etherscan.ts
hedi-edelbloute: It's not only a network call, for example apiDomain is null, you don't call the network and return []

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
```diff
@@ -3604,6 +3604,70 @@ export const cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency> = {
     ],
   },
   // Light Integrations are at the end of the list until we figure out a way to fix the ticker/managerApp collisions
+  ethereum_as_evm_test_only: {
```
gre, line 3607: I'm a bit concerned by this idea as it's the first time we actually break the 1 currency = 1 ledger currency relashionship. the coin model is not designed to be duplicated per network (afaik it's the first proposal doing this) and this can be a root cause of future obscure issues. for instance, the reasons we still need "findTokenByTicker" may be impacted ❓ 

To fix this, we should move toward the idea to have multiple backends among the same family. in the past, this was made possible, i think it still is: `libcore` and `js` was living at same time, and it was possible to make a switch between bridges, and we even had ways for user to switch the implementations. this system is still in place as we have `mock` vs `js`.

we may need to make `evm` & `ethereum` family one again & then it could internally change based on the `js:*` prefix of the account id (this is the way it was working for libcore)

I imagine we have such plan in mind for the future, in that case it would be good to be sure we mention it in the code, because this is going to add some tech debt.

lambertkevin: Those currencies are for test purpose only and hidden behind the "developer" mode since they're added as "testnets of" the original currency. I'm not sure to understand what kind of issues could arise from such "clone" currencies ? 🤔
Also tokens for "evm" currencies are only added as preload of a currency (so they're not as default in the @ledgerhq/cryptoassets lib), so *if* you had any issue with `findTokenByTicker` you'd have to be a dev to have it I guess ? (I can only find 1 `findTokenByTicker` usage in the whole codebase — the cli of Solana in this case — so I'm not sure when/how it's used, but "byTicker" to me would probably be used for counter values, where the Tokens would be identical then I guess ? Let me know if you have a specific case in mind.)

The plan is 100% to have only 1 family at the end, this PR is part of the process of adding every functionnalities of the "Ethereum" family to the EVM one, to be able to merge the 2, and have the "multiple backends among the same family" you're talking about 👍

chabroA (approved):
Looks good to me.
Libraries tests are failing, seems to be due to a snapshot not updated on the bot

libs/ledger-live-common/src/families/evm/api/rpc.common.ts
```diff
@@ -13,8 +13,9 @@ import { getSerializedTransaction } from "../transaction";
 import { makeLRUCache } from "../../../cache";
 import ERC20Abi from "../abis/erc20.abi.json";
 
-export const RPC_TIMEOUT = 5000; // wait 5 sec after a fail
-export const DEFAULT_RETRIES_RPC_METHODS = 3;
+export const RPC_TIMEOUT = process.env.NODE_ENV === "test" ? 100 : 5000; // wait 5 sec after a fail
```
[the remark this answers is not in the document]
lambertkevin: No it's actually changed only for the "integration tests" to either succeed faster or fail faster. With the 5s timeout, for a human there is no problem except "it's a bit long" (but necessary for some nodes like Optimism that have a very strict ratelimit), but for the bridge integration test which is trying to compare 2 "prepared" transactions, the 5s timeout can change completely the fees estimations because the network will have changed. So end result is transaction1 & transaction2 are different because of the network, but not because of our implementation, so it triggers a false negative 🤷‍♂️ 

So in fine, this timeout in "test" mode is just here to retry if it fails and wait the bare minimum amount of time, and if it fails then it's still fine as integration tests are flaky and optional (for now at least).

gre:


2 cases

```
```
- otherwise prefer 
```
```

**then, apply the config**

```
pnpm i
pnpm lint:fix --force
```

then commit in your branch.

```
```

2 cases

```
```
- otherwise prefer 
```
git merge origin/develop
```

---

*gre*

libs/coin-framework/src/nft/helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+        //
+        // E.g.: I'm creating 10 ERC1155 on OpenSea. It's not going to create anything on-chain.
+        // Then I (bob) decide to transfer 5 to someone (kvn). OpenSea is going to mint the NFT in its
+        // collection (`OpenSea Shared Storefront` / `OpenSea Collections`) and transfer it to kvn.
+        // But the event fired by the Smart Contract is going to be `bob transfered 5 NFT to kvn` which is false.
+        // It would then result in bob have -5 NFTs since he never received them first.
+        // If kvn send 2 back to bob, based on the events we would think that bob has -3 NFTs.
+        //
+        // To mitigate that we put a minimum value of 0 when an account is transferring some NFTs.
+        nft.amount = newAmount.isNegative() ? new BigNumber(0) : newAmount;
```
chabroA, line 53: If we take the same example from the above comment, wouldn't this line result in bob having 0 NFTs instead of 2 (the 2 that kvn sent back to bob)?

libs/coin-evm/src/adapters.ts
```diff
@@ -57,40 +66,52 @@ export const transactionToEthersTransaction = (
 export const etherscanOperationToOperation = (
   accountId: string,
   tx: EtherscanOperation
-): Operation => {
+): Operation[] => {
```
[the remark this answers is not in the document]
lambertkevin: In the condition of a self send. If I'm sending something to my own address, I have 2 operations in return, 1 IN, 1 OUT.

libs/coin-framework/src/nft/helpers.ts
lambertkevin: This method as just been copy pasted to the coin-framework, so it's not something new, but to answer your question: 
Bob's balance cannot go below 0 with this method so when he sends 5 NFTs, his balance went from 0 to 0 insteand of -5. When kvn is sending 2 back, he has the 2 NFTs and not -3 because we made sure he wouldn't get a negative balance.
If you're asking Open Sea though, his balance should be 7 because he still has 5 "lazy minted" that only exist on Open  Sea's database & 2 that are really on chain.

libs/coin-evm/src/adapters.ts
```diff
@@ -57,40 +66,52 @@ export const transactionToEthersTransaction = (
 export const etherscanOperationToOperation = (
```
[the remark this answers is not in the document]
lambertkevin: Done

libs/coin-evm/src/adapters.ts
```diff
@@ -99,50 +120,145 @@ export const etherscanOperationToOperation = (
...
+        extra: {},
+      } as Operation)
+  );
+};
+
+/**
+ * Adapter to convert an ERC721 transaction received
+ * on etherscan-like APIs into an Operation
+ */
+export const etherscanERC721EventToOperation = (
```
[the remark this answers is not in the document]
lambertkevin: Done

libs/coin-evm/src/adapters.ts
```diff
@@ -99,50 +120,145 @@ export const etherscanOperationToOperation = (
...
+        extra: {},
+      } as Operation)
+  );
+};
+
+/**
+ * Adapter to convert an ERC1155 transaction received
+ * on etherscan-like APIs into an Operation
+ */
+export const etherscanERC1155EventToOperation = (
```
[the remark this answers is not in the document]
lambertkevin: Done

PR #3515, opened 2023-05-25
[SUPPORT][LIVE-7577] extract network and cache in own lib

chabroA:
### 📝 Description

Extract network and cache logic from LLC to a new `live-network` package

- move `cache` and `network` files from LLC to new `live-network` package
- update paths for `cache` and `network` usage in LLC, LLD, LLM, cli
- add newly created package to `tools/common-tools/config-overrides.js`

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7577

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] ~**No breaking changes**~  || New package created and dependencies paths updated throughout projects

### 🚀 Expectations to reach


henrily-ledger (approved):
I don't have a lot of experience on package building, but it's LGTM.

libs/ledger-live-common/src/families/near/constants.ts
lambertkevin, line 1: I'm not sure this change is necessary for this PR ? 
If not, I would advise not adding this into a jungle of other changes, that would make it difficult to track.

libs/live-network/package.json
```diff
@@ -0,0 +1,84 @@
+{
+  "name": "@ledgerhq/live-network",
+  "version": "0.0.1",
```
lambertkevin, line 3: Version should probably be `1.0.0` right away as this version would mean we're in so early state that it could be breaking at every minor. 
I think this one is important enough for us to precise on the versionning so we know when a breaking change is really hapenning or not

lambertkevin (approved):
LGTM except for 2 not super important comments 👌 

And just as general comment, next time we have to update so many imports at the same time, it would probably be better to deactivate your plugin moving every imports by name 😆 
It makes it harder to review in this case 😋

libs/ledger-live-common/src/families/near/constants.ts
chabroA: It's needed to avoid circular dependency in imports (preventing the app from building). cf. this commit 1252726b4cf24beb46008bad22abb66e20853eb0

> remove circular dependency in near familly
circular dependency between logic and preload, broken by extracting
constants in a dedicated file

lambertkevin: Alright, makes sense 👌

gre:
+ undocumented impact on near family

PR #3523, opened 2023-05-26
feat: LLD allow users to use Ledger Vault as a signer

drakoFukayu:
### 📝 Description

Reopening the PR 3287 https://github.com/LedgerHQ/ledger-live/pull/3287

### ❓ Context

- **Impacted projects**: `ledger-live-desktop` `ledger-live-common` `ledgerjs/hw-transport-http`
- **Linked resource(s)**: [VG-14376 ](https://ledgerhq.atlassian.net/browse/VG-14376)

### ✅ Checklist

- [ ] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes**

### 📸 Demo

Demo

Between the record of this GIF, I updated some wording and added some Alerts:

Demo

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/components/VaultSignerTransport.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const VaultSignerTransport = () => {
+  const { token, enabled, host, workspace } = useSelector(vaultSignerSelector);
+  useEffect(() => {
+    if (enabled && currentMode !== "polling") {
+      setDeviceMode("polling");
+    } else if (!enabled) {
+      setDeviceMode(originalDeviceMode);
+    }
+    registerTransportModule({
```
gre, line 34: `registerTransportModule` was not yet designed to be dynamically used. it probably works the first time, but the problem is that we don't have a way to "unregister" which will end up being registering multiple time each time the effect runs and could create race conditions.

I would suggest:
- to move `registerTransportModule` as a module definition level again, next to the others registerTransportModule (like the one we do in the live common setup logic). => it shouldn't be a problem as you will filter on the id
- as you need some extra information (host, token, workspace) in that transport, you may also need to have it part of the id. for instance `vault-transport?host=foo&token=bar&workspace=42`

libs/ledgerjs/packages/hw-transport-http/src/VaultTransport.ts
```diff
@@ -0,0 +1,48 @@
+import { log } from "@ledgerhq/logs";
+import WebSocketTransport from "./WebSocketTransport";
+
+type VaultData = {
+  token: string;
+  workspace: string;
+};
+
+export default class VaultTransport extends WebSocketTransport {
```
gre, line 9: it would be better to make `VaultTransport` a dedicated package because it is not the scope of hw-transport-http i think.

an alternative idea is to add feature directly on the WebSocketTransport class, but it should not be breaking. it might be good to have an independent package that you can maintain independently (and that depends on hw-transport-http)

libs/ledger-live-common/src/hw/index.ts
```diff
@@ -42,7 +42,12 @@ export type TransportModule = {
 };
 const modules: TransportModule[] = [];
 export const registerTransportModule = (module: TransportModule): void => {
-  modules.push(module);
+  const index = modules.findIndex((m) => m.id === module.id);
+  if (index > -1) {
+    modules[index] = module;
+  } else {
+    modules.push(module);
+  }
```
gre, line 50: i would prefer not to do this (see first comment) or if we want to go to such approach, the function should instead return a "unregister" function and it's user land responsability not to register twice.

gre (changes requested):

gre:
👋  we have recently merged https://github.com/LedgerHQ/ledger-live/pull/3455 in develop branch which reinforces the TypeScript checks, mostly related to modals and coin families. Could you rebase and check that everything still pass?

in the PR, i could see there is an impact on the way modals was defined, it is likely you will have to define a new modal in `modals/types.ts` and type the data it takes, then you should no longer expect a `name` generic prop from the modal code, but inline the name="YOUR_NAME".

Thanks, gre.

apps/ledger-live-desktop/src/renderer/components/VaultSignerTransport.tsx
flocks: The problem is I need the transport to be registered each time `host/token/workspace` changes, that's why it's inside the `useEffect()`

flocks:
Thx @gre I will have a look next week

apps/ledger-live-desktop/src/renderer/components/VaultSignerTransport.tsx
gre: but then it could work if it's part of the id?

flocks: I struggle to see what you mean tbh, if I register the vault transport [here](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/live-common-setup.ts#L24) how can make it re-register each time `host/token/workspace` change?

libs/ledgerjs/packages/hw-transport-http/src/VaultTransport.ts
flocks: I think you are right, vault transport should be a dedicated package!
I will work on that

apps/ledger-live-desktop/src/renderer/components/VaultSignerTransport.tsx
gre: example https://github.com/LedgerHQ/ledger-live/blob/main/libs/ledger-live-common/src/load/speculos.ts#L471-L496

gre:


2 cases

```
```
- otherwise prefer 
```
```

**then, apply the config**

```
pnpm i
pnpm lint:fix --force
```

then commit in your branch.

```
```

2 cases

```
```
- otherwise prefer 
```
git merge origin/develop
```

---

*gre*

apps/ledger-live-desktop/src/renderer/live-common-setup.ts
```diff
@@ -27,3 +29,26 @@ registerTransportModule({
...
   disconnect: () => Promise.resolve(),
 });
+
+registerTransportModule({
+  id: "vault-transport",
+  open: (id: string) => {
+    const prefixID = "vault-transport:";
+    if (!id.startsWith(prefixID)) return;
+    // FIXME how to restore initial device mode value?
+    setDeviceMode("polling");
```
gre, line 39: in future, the device mode should be provided by the transport module because it seems related to the transport used anyway. i think you can go with a hack that setDeviceMode it back in the usb transport module's open() but it's something we will need to track somewhere ( cc @juan-cortes @alexandremgo )

flocks: Yep agree it feels weird to set it back in the `open` of an another transport, but I can think of a better solution  :/

gre (approved):

apps/ledger-live-desktop/src/renderer/live-common-setup.ts
flocks: @gre just pushed this commit because i realized it was not working with a real device. It seems we actually never open a transport with `id=ipc` so the module was never picked up. May be it changed because I remember testing toggling on and off vault transport when I initially worked on that (weeks ago now).

flocks:
@gre just in case you missed my [previous comment](https://github.com/LedgerHQ/ledger-live/pull/3523#discussion_r1226796494). And if you agree with that and re-approve, do you mind if I rebase and squash all commits on this branch into 1 single commit so it will ease potential future rebase onto develop, in case this PR is not merged right away?

gre:
@flocks you can squash all commits.

gre (approved):
✔️ i did a smoke test on Mac and can still properly use a nano device.

PR #3542, opened 2023-05-30
feat: LIVE-7473 add 800ms delay before failures for better UX

juan-cortes:
### 📝 Description

The rapid transition between steps during error cases for device actions may result in a glitchy appearance. For example, when attempting to retry a failed device action on a locked device, the completion time is just a few milliseconds. However, this duration is sufficient for the device action to briefly display a flashing indication of the previously attempted step before the failure.

To address this issue we have implemented a compromise solution. We introduce an `800ms` delay for error notifications, allowing users to observe a visual indication before the error is thrown again and the user interface (UI) changes accordingly. It is important to note that this delay also affects user refusals, but the product team has agreed that this delay is acceptable.

If there are specific errors for which we do not want to introduce a delay, we can incorporate provisions to whitelist user refusals from the beginning. Implementing this provision should be relatively straightforward.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7473` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
Failed cases for device actions should be delayed, an easy way of testing this is accessing the manager with a locked device and hitting retry but technically all device action flows should benefit from this improvement.

### DIFF
--- a/.changeset/nasty-papayas-peel.md
+++ b/.changeset/nasty-papayas-peel.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Add fake delay before device action failures for better UX
--- a/libs/ledger-live-common/src/hw/actions/implementations.ts
+++ b/libs/ledger-live-common/src/hw/actions/implementations.ts
@@ -6,6 +6,7 @@ import {
   Subscription,
   TimeoutError,
   concat,
+  interval,
   of,
   timer,
 } from "rxjs";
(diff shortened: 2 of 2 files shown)


alexandremgo (approved):
Thanks for the doc comments ! 👌

PR #3570, opened 2023-06-02
fix(Firmware Update LLM): fix bluetooth restriction bypass

grsoares21:
### 📝 Description
Since https://github.com/LedgerHQ/ledger-live/pull/2742 it has been possible to bypass the USB-only restriction on the firmware update for LNX and start the firmware update using a Bluetooth transport. However, Bluetooth firmware updates are not supported by LNX and can lead to unpredictable behaviour. This PR fixes this and keeps the fix that was made on https://github.com/LedgerHQ/ledger-live/pull/2742 for the firmware update banner. It does that by updating the transport to be considered for the firmware update when we plug a device via USB. If the user is on the manager, the manager gets reloaded and starts using the USB transport for all operations (Firmware update, app installations, etc).

### ❓ Context

- **Linked resource(s)**: [LIVE-7048]

### ✅ Checklist

- [N/A] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes**

### 📸 Demo
https://github.com/LedgerHQ/ledger-live/assets/6013294/61546a2e-0f73-4bf8-9672-502708cc8eff

[LIVE-7048]: https://ledgerhq.atlassian.net/browse/LIVE-7048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/hooks/useListenToHidDevices.ts
+++ b/apps/ledger-live-mobile/src/hooks/useListenToHidDevices.ts
@@ -5,9 +5,8 @@ import { DescriptorEvent, DeviceModelId } from "@ledgerhq/types-devices";
 import HIDTransport from "@ledgerhq/react-native-hid";
 import { map } from "rxjs/operators";
 import useIsMounted from "@ledgerhq/live-common/hooks/useIsMounted";
-import { setWiredDevice } from "../actions/appstate";
 import { DeviceLike } from "../reducers/types";
-import { AppStateSetWiredDevicePayload } from "../actions/types";
+import { setLastConnectedDevice } from "../actions/settings";
 
 /**
  * Allows LLM to be aware of USB OTG connections on Android as they happen.
@@ -46,8 +45,17 @@ export const useListenToHidDevices = () => {
           ),
         )
         .subscribe({
-          next: (wiredDevice: AppStateSetWiredDevicePayload) => {
-            dispatch(setWiredDevice(wiredDevice));
+          next: (wiredDevice: DeviceLike | null) => {
+            if (wiredDevice) {
+              dispatch(
+                setLastConnectedDevice({
+                  deviceId: wiredDevice.id,
+                  modelId: wiredDevice.modelId,
+                  deviceName: wiredDevice.name,
+                  wired: true,
+                }),
+              );
+            }
           },
           error: () => {
             onScheduleNewListen();
--- a/apps/ledger-live-mobile/src/screens/Manager/Manager.tsx
+++ b/apps/ledger-live-mobile/src/screens/Manager/Manager.tsx
@@ -22,10 +22,7 @@ import { setLastSeenDeviceInfo } from "../../actions/settings";
 import { ScreenName } from "../../const";
 import FirmwareUpdateScreen from "../../components/FirmwareUpdate";
 import { ManagerNavigatorStackParamList } from "../../components/RootNavigator/types/ManagerNavigator";
-import {
-  BaseComposite,
-  StackNavigatorProps,
-} from "../../components/RootNavigator/types/helpers";
+import { BaseComposite, StackNavigatorProps } from "../../components/RootNavigator/types/helpers";
 import { lastConnectedDeviceSelector } from "../../reducers/settings";
 
 type NavigationProps = BaseComposite<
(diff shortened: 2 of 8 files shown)


apps/ledger-live-mobile/src/screens/Manager/Manager.tsx
```diff
@@ -47,6 +48,17 @@ const Manager = ({ navigation, route }: NavigationProps) => {
   const [state, dispatch] = useApps(result, deviceId, appsToRestore);
   const reduxDispatch = useDispatch();
 
+  const lastConnectedDevice = useSelector(lastConnectedDeviceSelector);
+  useEffect(() => {
+    // refresh the manager if a new device gets connected
+    // (happes only when we plug a new device via USB)
+    if (lastConnectedDevice?.deviceId !== device.deviceId) {
```
gre, line 55: which part of the code is going to setting `lastConnectedDevice` back to device? it probably should be the same place?

because when only looking at this portion of the added code, nothing show me that lastConnectedDevice is going to converge to the same device. if it's not done in the same effect, there is a risk that worse case scenario the value never is set (so the effect runs indefinitely?) or also that it is set in another cycle (so it `replace` a few times before settling 🤔 )

grsoares21: We don't change `lastConnectedDevice` anywhere here, we're just reacting to a change on it (it is a global state).

This converges because we're renavigating to a previous page (this one is `ScreenName.ManagerMain`, the previous one, where we select the device, is the `ScreenName.Manager` one), with the new device, so the next time we get to this useEffect the new `device.deviceId` will be equal to `lastConnectedDevice.deviceId`.

If `lastConnectedDevice` is set to null anywhere (it shouldn't really happen in the app, but if it does), we'll just renavigate to the manager device selection screen with no device in the params, so the user will have manually to select a device to connect to.

grsoares21: So, the places where we update these values when selecting a device in the manager are the following:
* https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/components/SelectDevice/index.tsx#L107 (for the old device selection component)
* https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx#L160 (for the new device selection component)

On both of them we dispatch the action that updates the global state before calling the `onSelect` callback which is the callback that is navigating to this open manager screen. However, I'm not sure if this is enough to guarantee that we will have the global state up to date before entering this page 🤔. Do you have any ideas of how we could guarantee that?

grsoares21: And the other place where this value is updated is on https://github.com/LedgerHQ/ledger-live/pull/3570/files#diff-7323cf8e09a7abc28bab7b04c44c296398310994e71edccdd604386682628e95
where we update it when the user plugs a new USB device, and this is the one we want to be able to react to and reload the manager.

gre:


2 cases

```
```
- otherwise prefer 
```
```

**then, apply the config**

```
pnpm i
pnpm lint:fix --force
```

then commit in your branch.

```
```

2 cases

```
```
- otherwise prefer 
```
git merge origin/develop
```

---

*gre*

apps/ledger-live-mobile/src/screens/Manager/Manager.tsx
grsoares21: We update it in the connect manager flow. the big question here is if there's a way for it to not be updated by the time we declare the useEffect that reacts to it. This in theory, would cause us to enter the manager, then, right after it, reload it again because the device was updated after we entered the manager.

apps/ledger-live-mobile/src/hooks/useListenToHidDevices.ts
```diff
@@ -48,8 +47,17 @@ export const useListenToHidDevices = () => {
...
+            if (wiredDevice) {
+              dispatch(
+                setLastConnectedDevice({
+                  deviceId: wiredDevice.id,
+                  modelId: wiredDevice.modelId,
+                  deviceName: wiredDevice.name,
+                  wired: true,
+                }),
+              );
+            }
```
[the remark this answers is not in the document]
grsoares21: Yeah, not really, if we try to connect to an USB device without it being connected the app will only ask you to connect it.

PR #3594, opened 2023-06-06
feat: added address query and navigation to account route for deeplinks

ggilchrist-ledger:
### 📝 Description

Added address query and navigation to account route for deeplinks

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7770

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Uploading Screen Recording 2023-06-06 at 17.50.39.mov…

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/warm-adults-clap.md
+++ b/.changeset/warm-adults-clap.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Added ability for navigation to specific account via the account deeplink if a valid currency and account address are provided as params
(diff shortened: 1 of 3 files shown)


sarneijim (commented):
It look great!
Before merge It will be nice confirm with @mcayuelas-ledger , who did the precious implementation that we are not going to have any "incompatibility" between networks. Could we have 2 different accounts with the same address?

https://media.giphy.com/media/OJCx3bKHMNEAkY0Dkz/giphy.gif

ggilchrist-ledger:
Yeah good point. I wanted to avoid having to check the currency type matches with the account but it would probably be a good idea to do so. Currently if you put `bitcoin` as the currency and a valid eth address it will navigate to the eth address which isn't the best. Any thoughts @mcayuelas-ledger ?

sarneijim (approved):

PR #3600, opened 2023-06-07
bugfix/LIVE-7728 Memoize device action requests to prevent loops

juan-cortes:
### 📝 Description

The recent refactor in device action implementation meant that rerenders of the screen that renders the `DeviceAction` component would trigger a new execution of the action itself **if the request was set inline**, if we memoize the request this doesn't happen.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7728` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
image

### 🚀 Expectations to reach

In particular it should impact flows starting on the discover section, these were found out to be looping in the regression testing. This should address that.

### DIFF
--- a/apps/ledger-live-mobile/src/screens/DeviceConnect/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/DeviceConnect/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from "react";
+import React, { useCallback, useMemo, useState } from "react";
 import { StyleSheet } from "react-native";
 import { SafeAreaView } from "react-native-safe-area-context";
 import { TFunction, useTranslation } from "react-i18next";
@@ -94,6 +94,13 @@ export default function DeviceConnect({ navigation, route }: NavigationProps) {
     [navigation, t],
   );
 
+  const request = useMemo(
+    () => ({
+      appName,
+    }),
+    [appName],
+  );
+
   return (
     <SafeAreaView
       edges={["bottom"]}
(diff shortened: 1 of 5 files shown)


apps/ledger-live-mobile/src/screens/DeviceConnect/index.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { useCallback, useState } from "react";
-import { StyleSheet } from "react-native";
+import React, { useCallback, useMemo, useState } from "react";
+import { StyleSheet, Text } from "react-native";
```
alexandremgo, line 2: Looks you used `Text` but forgot to remove it ;)

apps/ledger-live-mobile/src/screens/SignTransaction/02-ConnectDevice.tsx
```diff
@@ -57,16 +77,8 @@ function ConnectDevice({
...
-          parentAccount,
-          appName,
-          transaction,
-          status,
-          tokenCurrency,
-          dependencies,
-          requireLatestFirmware: true,
-        }}
+        // @ts-expect-error Wrong types?
+        request={request}
```
alexandremgo, line 81: It looks like it's because `transaction` could be `undefined` or `null` while it should not for the `request` type -> but i don't understand why it was not complaining before ?

alexandremgo (approved):
Except the small import to remove, looks good !

apps/ledger-live-mobile/src/screens/SignTransaction/02-ConnectDevice.tsx
juan-cortes: Me neither, and since this is a hot fix, I dont want to understand it now.
It's actually the same on LLD, bunch of expect errors until we figure it out

alexandremgo: Yep agree !

PR #3611, opened 2023-06-07
[LIVE-7405] Create GasTracker abstraction

chabroA:
### 📝 Description

Add `gasOption` and `gasTracker` in types
Add `gasOptions` in evm tx type
Add `gasTracker` in `ethereumLikeInfo` in `CryptoCurrency` type
Use `gasOption` in `prepareTransaction` to get appropriate `feeData`

### ❓ Context

- **Impacted projects**: `coin-evm` `llc`
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7405

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


libs/coin-evm/src/prepareTransaction.ts
```diff
@@ -127,7 +127,24 @@ export const prepareTransaction = async (
 ): Promise<EvmTransaction> => {
   const { currency } = account;
   // Get the current network status fees
-  const feeData = await getFeesEstimation(currency);
+  const feeData = await (async () => {
+    if (!transaction.feesStrategy) {
+      return await getFeesEstimation(currency);
```
lambertkevin, line 132: Except in try/catch, return await is not necessary I think.
Also is this early return really necessary ? I think not having it is having the same result at line 143 because you'll get an undefined `gasOption` and return the `getFeesEstimation` instead ?

libs/coin-evm/src/prepareTransaction.ts
```diff
@@ -127,7 +127,24 @@ export const prepareTransaction = async (
...
+      return {
+        gasPrice: transaction.gasPrice ?? null,
+        maxFeePerGas: transaction.maxFeePerGas ?? null,
+        maxPriorityFeePerGas: transaction.maxPriorityFeePerGas ?? null,
+      };
+    }
+
+    const gasOption = transaction.gasOptions?.[transaction.feesStrategy];
+
+    return gasOption ?? (await getFeesEstimation(currency));
```
lambertkevin, line 145: Same here, I think `return gasOption ?? getFeesEstimation(currency)` is equivalent ?

libs/coin-evm/src/api/gasTracker/index.ts
```diff
@@ -0,0 +1,19 @@
...
+} => {
+  switch (currency.ethereumLikeInfo?.gasTracker?.type) {
+    case "ledger":
+      return {
+        getGasOptions: ledgerGetGasOptions,
+      };
+
+    default:
+      throw new Error(`No gas tracker found for currency "${currency.id}"`);
+  }
```
lambertkevin, line 18: That might be problematic with the React hook, as that means you'd need to have a conditional hook or a try/catch in it for all the currencies that won't have a gasTracker 🤔

libs/coin-evm/src/api/gasTracker/ledger.ts
lambertkevin, line 1: Globally it feels like this file is trying too much to be close to what it was in the Ethereum family. 
We have the freedom to rework it completely to become something a bit more readable 👌

Let's try to simplify it as much as possible:

- Don't use `EIP1559ShouldBeUsed`, transactions are typed correctly in the EVM family, the need for EIP1559 can be found simply by looking at the `transaction.type === 2`
- We create a very simple map of `CurrencyId => ExplorerId` (like `polygon: "matic"`) to be able to build the URI in a simple manner. No need for 10 abstractions like we used to have to build a simple URI.
- Let's not waste time and readability by having 2 methods for a unique need: `getGasTrackerBarometer` can go away and `getGasOptions` can do the request directly and return the `{ gasPrice, maxFeePerGas, maxPriorityFeePerGas }`

At the end, I think this file should only be a Map & a `getGasOptions` method. Wdyt ?

libs/coin-evm/src/api/gasTracker/ledger.unit.test.ts
lambertkevin, line 1: Move this file in the `__tests__` folder please

lambertkevin (changes requested):
Just a few things, but the most important one is the Ledger GasTracker that can be really simplified 🙏

libs/ledger-live-common/src/families/evm/react.ts
```diff
@@ -0,0 +1,40 @@
...
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { useEffect, useState } from "react";
+
+/**
+ *
+ * React hook to get gas options for a given currency
+ * Used in the send flow to get the gas options for the currency,
+ * for example in SelectFeeStrategy.tsx file in the renderer/families/{currency} folder in LLD
+ */
+export const useGasOptions = ({
```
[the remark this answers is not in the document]
lambertkevin: It will be used in the frontend once we recreate the new UI's for the EVM family send flow to add the gas tracker

libs/coin-evm/src/prepareTransaction.ts
chabroA: > Also is this early return really necessary ? I think not having it is having the same result at line 143 because you'll get an undefined gasOption and return the getFeesEstimation instead ?

That would be true in JS (or loosely typed TS) code, but here the `GasOptions` type only has `slow`, `medium` and `fast` valid keys. `null` or `undefined` are not valid keys (and I don't think they should be).

This if statement basically acts as a type guard / type narrowing logic, ensuring that when we try to access a `gasOption` field, we are using a valid key (i.e: not a `null` or `undefined` one).

For example, here is what TS says when we remove this if statement:
TS error

An alternative could be to remove this if statement and update line 143 to something like:
```ts
    const gasOption = transaction.feesStrategy
      ? transaction.gasOptions?.[transaction.feesStrategy]
      : undefined;
```

I'm not a big fan of this here though, I feel like it makes the code less readable / straightforward

libs/coin-evm/src/api/gasTracker/ledger.ts
```diff
@@ -0,0 +1,123 @@
...
+import invariant from "invariant";
+
+//#region Ledger Explorer URL
+type LedgerExplorer = {
+  version: string;
+  id: string;
+  endpoint: string;
+};
+
+const findCurrencyExplorer = (
```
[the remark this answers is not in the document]
chabroA: This should actually be removed by a rework of this file. cf. https://github.com/LedgerHQ/ledger-live/pull/3611#discussion_r1223298777

libs/coin-evm/src/prepareTransaction.ts
lambertkevin: I think you can fix this by casting this way:

```typescript
const gasOption = transaction.gasOptions?.[transaction.feeStrategy as keyof typeof transaction.gasOptions];
```
Not super good looking by hey, typescript right ? 🤷‍♂️

libs/coin-evm/src/api/gasTracker/ledger.ts
chabroA: I agree with your overall feedback / idea of simplification. But I'm a bit sceptic regarding the EIP1559 / legacy differentiation logic handling though. With this proposal, the choice (and logic) of getting EIP1559 or legacy like gas options is outsourced to the client (here, the UI: LLD / LLM through a new param to the react hook). Does it really make sense / is it what we want?

Interesting idea of using a Map to link a currencyId with an explorerId, but it seems redondant with the `cryptocurrenciesById` record that already associate a currencyId with a currency definition, which includes (at least for some currencies), and explorerId.
I created such a map in my commit (57e6a18c70588f9558548e7228705ecc52c076f4) with a proposal rework, but it seems over engineered (and redondant) compared to a simple `currency.explorerId ?? currency.id` as was present in `findCurrencyExplorer`. Or maybe I didn't get your proposed solution quite right?

lambertkevin: > the choice (and logic) of getting EIP1559 or legacy like gas options is outsourced to the client

Not really, the typing is done at the `prepareTransaction` level once we find out which `FeeData` has been provided. It makes sense in our Bridge pipeline:
- We create a transaction without knowing anything basically (`createTransaction`)
- We receive new parameters from the users (`updateTransaction`) 
- We "prepare" the transaction with what the coin integration find (`prepareTransaction` that's where we're typing the transaction depending on the FeeData we've got either by the user or the node itself)
- ... (doesn't matter)
- We return the transaction to the client

And that loop will be done X times until the user stops triggering it with changes and reaches the end of the flow.

So since the clients are using the `useGasOption` hook, in the context of the live, "live-common" is still in control of whatever is gonna be provided to the transaction and the `useGasOption` hook should only intervene after several Bridge loops already meaning we will already have figured out with the node if the currency is compatible or not so the typing should already be correct.

Also I think the client should indeed be able to provide as much info about a transaction as possible without the coin integration inteferring. A currency supporting EIP1559 (type 2 transactions) doesn't mean it's incapable of working with type 1 or type 0 (Legacy). If a client wants to enforce a type 0, we should not interfere with it just because "we know it's compatible with type 2". And that should "retype" the transaction accordingly. 

> Interesting idea of using a Map to link a currencyId with an explorerId, but it seems redondant with the cryptocurrenciesById record that already associate a currencyId with a currency definition, which includes (at least for some currencies), and explorerId.

Ok I didn't know it was coming from the currencies list. Problem is just the name of the property then, as an "explorer" isn't always a Ledger explorer with our new paradigm. Anyway, indeed there no need for a map then, as long as the currency is provided we should be able to get that thing. 👍 

> I created such a map in my commit (https://github.com/LedgerHQ/ledger-live/commit/57e6a18c70588f9558548e7228705ecc52c076f4) with a proposal rework, but it seems over engineered (and redondant) compared to a simple currency.explorerId ?? currency.id as was present in findCurrencyExplorer.

Seems like we don't need the Map at all then, we can get rid of it and just throw if the currency provided isn't having an "explorerId" even if I really dislike the property name 😆
 

Didn't yet had a look a the full proposal commit but it looks better ! Imo, we can even get rid of the `buildExplorerURL` which is useless in it's current state and replace it with a very simple string template in the `getGasOptions`:
```typescript 
if (!currency.explorerId) throw new Error("blablabla");

const response = await network(`${getEnv(EXPLORER)}/blockchain/v4/${currency.explorerId}/gastracker/barometer`)
```

libs/coin-evm/src/api/gasTracker/index.ts
```diff
@@ -0,0 +1,31 @@
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { GasOptions } from "../../types";
+import { getGasOptions as ledgerGetGasOptions } from "./ledger";
+
+type GasTrackerApi = {
+  getGasOptions: ({
+    currency,
+    shouldUseEip1559,
```
lambertkevin, line 8: I would rather provide the full transaction rather than something that specific. 
You can then infer whatever you want from it.
Here it would be "If transaction.type === 2 -> display=eip1559"

chabroA:
@lambertkevin updated gas tracker abstraction, ledger implem and react hook following previous feedbacks -> ef2b182ab1d092be7ae3df93fe47c79c30241ee0

Though regarding our new `currencyGasTrackerMap` used to hold GasTracker config in the Ledger gas tracker implem' (cf. [here](https://github.com/LedgerHQ/ledger-live/pull/3611/files#diff-8254fc4a9ebeed4e85a3cdfb06b503cd83e8271bfc06baa023ee253ccb7888ffR18-R39)). It is static for now but we might want to to make it dynamic at some point?

gre:


2 cases

```
```
- otherwise prefer 
```
```

**then, apply the config**

```
pnpm i
pnpm lint:fix --force
```

then commit in your branch.

```
```

2 cases

```
```
- otherwise prefer 
```
git merge origin/develop
```

---

*gre*

lambertkevin (approved):
Great job, LGTM 🚀

PR #3643, opened 2023-06-09
[LIVE-7402] Feature - Add support for NFT transaction in EVM family

lambertkevin:
### 📝 Description

Adding support for sending NFT transactions in the EVM Family.

Part of the feature branch: #3655

### ❓ Context

- **Impacted projects**: `@ledgerhq/coin-framework`, `@ledgerhq/coin-evm`, `@ledegrhq/ledger-live-common`, `@ledgerhq/cryptoassets` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7402 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach
Screenshot 2023-06-12 at 13 08 10


gre (commented):
👍  nice tests

libs/coin-evm/src/api/etherscan.ts
```diff
@@ -102,7 +102,28 @@ export const getLastTokenOperations = async (
     url,
   });
 
-  return ops.map((event, index) => etherscanERC20EventToOperations(accountId, event, index)).flat();
+  // Why this thing ?
```
hedi-edelbloute, line 105: Nice comment, we could just add that event <=> operation is 1/1 correlation

Something like "We want one operation per event, but one transaction triggers multiple events, that's why ..."

Not important and can be ignored

libs/coin-evm/src/transaction.ts
```diff
@@ -118,22 +129,63 @@ export const toTransactionRaw = (tx: EvmTransaction): EvmTransactionRaw => {
...
+): Buffer | undefined => {
+  switch (transaction.mode) {
+    case "send": {
+      const contract = new ethers.utils.Interface(ERC20ABI);
+      const data = contract.encodeFunctionData("transfer", [
+        transaction.recipient,
+        transaction.amount.toFixed(),
+      ]);
+
+      // removing 0x prefix
```
hedi-edelbloute, line 160: Maybe add why we need to do this

libs/coin-framework/src/batcher/index.ts
```diff
@@ -0,0 +1,65 @@
+import { Batch, BatchElement, Batcher } from "./types";
+
+/**
+ * Factory to make an API call batcher.
```
hedi-edelbloute, line 4: I understand that it is used to kind of do a "Promise.all" between multiple files, it it to "compensate" for some architecture problem of our coin framework ?

Would we have another way to do it ?

hedi-edelbloute (approved):
Nice job on comments 👍 

Only thing that I want to understand is why we need to use this api call batcher, rest is minor

Lgtm

libs/coin-framework/src/batcher/index.ts
lambertkevin: It's not compensating for the coin-framework, it's a solution like https://github.com/graphql/dataloader. 
Basically, it allows multiple frontend components to ask in their own promise "fetch this param" while on the other end all the params are going to be grouped into a single "batched" request.
(so a `POST /endpoint ([param1, param2, param3]): [response1, response2, response3]`) 
And the responses will be returned as a resolution to each original promise individually. 

```
                                                                           
                    +----------+  +----------+                             
                    |          |  |          |                             
                    |  comp 1  |  |  comp 2  |                             
                    |          |  |          |                             
                    +----------+  +----------+                             
            +--+ promise1|            |promise2 +--+                       
            |    param   |            |  param2    |                       
            |            |            |            |                       
 resolve(A) |            +            +            | resolve(B)            
            |           +--------------+           |                       
            |           |              |           |                       
            +-----------|    batcher   |-----------+                       
                        |              |                                   
                        +--------------+                                   
                     params  |     + returns                               
                     [1,2]   |     | [A,B]                                 
                             +     |                                       
                         +------------+                                    
                         |            |                                    
                         |   server   |                                    
                         |            |                                    
                         +------------+                                    
                                                                           
                                                                           
                                                                           
```

libs/coin-evm/src/api/etherscan.ts
```diff
@@ -102,7 +102,28 @@ export const getLastTokenOperations = async (
...
+  // Multiple events can be fired by the same transactions and
+  // those transfer events can go from anyone to anyone, which
+  // means that multiple events could be sent to or from the
+  // same address during the same transaction.
+  //
+  // To make sure every event (transformed into an operation here)
+  // has a unique id, we're groupping them by transaction hash
```
chabroA, line 112:
```suggestion
  // has a unique id, we're grouping them by transaction hash
```

chabroA (approved):

libs/coin-evm/src/api/etherscan.ts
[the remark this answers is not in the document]
lambertkevin: I'll update this during a last rebase on the feature branch, otherwise I'm in rebase loop again :p

libs/coin-evm/src/transaction.ts
lambertkevin: Don't think that's necessary here, removing the 0x prefix is like a 101 thing when you work with Buffers and hexadecimals, so I don't see what I could really add to this 🤷‍♂️

libs/coin-evm/src/api/etherscan.ts
lambertkevin: Like said to Hedi, I'll do a last update/rebase once it's on the main branch, I can't stand the idea of doing yet another rebase, that will force me to rebase also the UI PR just for 1 letter haha

PR #3661, opened 2023-06-13
feat/LIVE-7872 Uninstall all feature + CLI implementation

juan-cortes:
### 📝 Description

With the 1.2.0 version of Stax we receive for the first time a firmware that allows us to bulk uninstall with user consent, allowing us to skip the uninstall queue completely, prevent HSM connections, speeding up the flow, improving the UX. We should use that when available, falling back to the normal uninstall flow when it’s not. Luckily, when the APDU is not available we get a unified `6d02` response, so leverage that in the implementation so that we always try that first.

This pr introduces the device action, the command and logic used, test coverage and a CLI implementation of the command, testable with `FORCE_PROVIDER=3 pnpm run:cli appUninstallAll` on a 1.2.0 Stax. Running the command (without the forced provider) on any other device should still uninstall the applications, but slowly, one, by, one.

### ❓ Context

- **Impacted projects**: `ledger-live-common, cli` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7872` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
1.2.0 Stax should uninstall apps very quickly, other devices should uninstall apps very slowly in comparison.
Always in the scenario where we are uninstalling everything, of course.

### DIFF
--- a/apps/cli/src/commands/appUninstallAll.ts
+++ b/apps/cli/src/commands/appUninstallAll.ts
@@ -5,7 +5,9 @@ import { withDevice } from "@ledgerhq/live-common/hw/deviceAccess";
 import getDeviceInfo from "@ledgerhq/live-common/hw/getDeviceInfo";
 import { reducer, runAll } from "@ledgerhq/live-common/apps/index";
 import { listApps, execWithTransport } from "@ledgerhq/live-common/apps/hw";
+import { command as uninstallAllApps } from "@ledgerhq/live-common/hw/uninstallAllApps";
 import { deviceOpt } from "../scan";
+
 export default {
   description: "uninstall all apps in the device",
   args: [deviceOpt],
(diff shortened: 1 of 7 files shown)


libs/ledger-live-common/src/hw/uninstallAllApps.test.ts
```diff
@@ -0,0 +1,37 @@
+import { StatusCodes, UserRefusedOnDevice } from "@ledgerhq/errors";
+import { command as uninstallAllApps } from "./uninstallAllApps";
+import Transport from "@ledgerhq/hw-transport";
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore next-line
+const mockTransportGenerator = out => ({ send: () => out } as Transport);
+
+describe("uninstallAllApps", () => {
```
alexandremgo, line 9: Nice :)

libs/ledger-live-common/src/apps/logic.ts
```diff
@@ -191,7 +191,12 @@ export const reducer = (state: State, action: Action): State => {
 
     case "recover":
       return { ...state, currentError: null };
-
+    case "wiped":
+      // Reconciliate an already achieved wipe via single uninstall apdu
+      return {
+        ...state,
+        installed: [],
+      }
```
alexandremgo, line 199:
```suggestion
      };
```

It's missing a last `;` ;)

apps/cli/src/commands/appUninstallAll.ts
```diff
@@ -16,20 +18,27 @@ export default {
...
-              (e: any) =>
-                reducer(e.result, {
-                  type: "wipe",
-                }),
-              exec,
-            ),
-            mergeMap(s => runAll(s, exec)),
-          ),
-        ),
+      return from(t.exchange(Buffer.from("e0dd000000", "hex"))).pipe(
```
alexandremgo, line 21: It's ok to have a "raw" APDU here ? I don't find it on the list of known apdu, could you add a comment above it pls 🙏

juan-cortes: It's actually stupid because it should be using the command instead which is the whole point, good catch

alexandremgo (approved):
LGTM 🐬

PR #3664, opened 2023-06-13
Fix/LIVE-7801: add optional shouldNotPopNavigationOnSuccess param to account selection

kallen-ledger:
### 📝 Description
Add optional `shouldNotPopNavigationOnSuccess` parameter that can be passed to the account selection. In cases where this is undefined or false then the account screen will work as it has previously.
In cases where the value is true then we will not pop the navigation on the account page. At the moment this is only useful for the staking flows as most handle their own navigation and poping from the account selection page has caused issues where the user is unable to access the staking flows.

### ❓ Context

- **Linked resource(s)**: `LIVE-7801` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/RootNavigator/types/RequestAccountNavigator.ts
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/types/RequestAccountNavigator.ts
@@ -20,6 +20,7 @@ export type RequestAccountNavigatorParamList = {
     currency: CryptoOrTokenCurrency;
     allowAddAccount?: boolean;
     onSuccess?: (account: AccountLike, parentAccount?: Account) => void;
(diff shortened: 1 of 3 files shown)


beths-ledger (approved):
Nice. A variable name occurred to me whilst reading this, `preventNavigateOnSuccess`, which to me echoes the familiar "e.preventDefault". I thought I'd share in case you find it more readable than `shouldNotPopNavigationOnSuccess`, but naming double negatives is difficult the other is also descriptive.

kallen-ledger:
Great suggestion with the variable name, but after pairing with @sarneijim I will close this and re-raise a very another PR.

PR #3677, opened 2023-06-14
fix: LIVE-7878 inline [] causing manager state rerender - Release

juan-cortes:
### 📝 Description

During the typescript migration of LLD we changed the way we passed the `appsToRestore` parameter around turning `appsToRestore={appsToRestore}` to `appsToRestore={appsToRestore || []}`. A seemingly harmless change to make typescript happy but with the added consequence of making anything that depends on that variable having to be recalculated **every time there's a render.** This had the side effect of resetting the app state every time we opened or closed a drawer.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7878` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
- Enter the manager
- Install some apps
- Start some flow (rename, cls, language)
- It is expected that the state is respected and not reset.

### DIFF
--- a/.changeset/red-falcons-train.md
+++ b/.changeset/red-falcons-train.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Prevent render loop that dismisses app state in my ledger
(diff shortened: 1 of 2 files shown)


alexandremgo (approved):

gre (commented):
good catch. sorry for this.

PR #3679, opened 2023-06-14
Bugfix/LIVE-7801: fix non-eth staking flows, generic drawer for eth and future liquid staking programs.

kallen-ledger:
### 📝 Description

* Created `StakingDrawer` component. This will encompass future staking drawers.
* Added id to root navigation to easily get with `navigation.getParent(id: string)`
* Reverted account selection to code before kiln pooling ticket was merged.
* Use `stakingDrawer` instead of just `drawer` as a parameter name.
* Remove eth staking drawer components where no longer needed.
### ❓ Context

- **Linked resource(s)**: `LIVE-7801` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


sarneijim (approved):
I love it! 
The only small detail that I would change is the name of the drawer making it a little more generic and the param used. I think we do not have the need to particularise it to stake, everyone could add their own banner if it is going to be called from multiple places in the app.

```
 <RootDrawer drawer={route.params?.drawer} />
 ```
 I have been testing, everything looks great, only I could not test with the "menu" button, I guess it is a problem with my simulator (please be sure you have test that part)

PR #3686, opened 2023-06-14
feat:  pagination for cosmos nodes

hedi-edelbloute:
### 📝 Description

This PR adds support for pagination on cosmos nodes and remove the orderBy DESCENDING dirty fix for accounts that have more than 200 transactions

I tried to use the next_key parameter but it does not work...

see
https://cosmoshub4.coin.ledger.com/cosmos/tx/v1beta1/txs?events=transfer.recipient=%27cosmos1e533m29emngs7camh5rjzpxkr0ngyuekw95jeg%27&pagination.limit=10

Total is higher than 10 and next_key is still not defined

Doc says otherwise... but oh well
https://docs.cosmos.network/swagger/#/Service/GetTxsEvent

So i implemented the pagination with pagination.limit and pagination.offset, example :

https://cosmoshub4.coin.ledger.com/cosmos/tx/v1beta1/txs?events=transfer.recipient=%27cosmos1e533m29emngs7camh5rjzpxkr0ngyuekw95jeg%27&pagination.limit=50&pagination.offset=50

### ❓ Context

- **Linked resource(s)**: [https://ledgerhq.atlassian.net/browse/LIVE-7660](https://ledgerhq.atlassian.net/browse/LIVE-7660) 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
+++ b/libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
@@ -39,7 +39,7 @@ export class CosmosAPI {
         await Promise.all([
           this.getAllBalances(address, currency),
           this.getHeight(),
-          this.getTransactions(address),
+          this.getTransactions(address, 50),
           this.getDelegations(address, currency),
           this.getRedelegations(address),
           this.getUnbondings(address),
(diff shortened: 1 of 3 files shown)


libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -229,49 +229,94 @@ export class CosmosAPI {
...
-  private async getSendingTransactions(
+  private async fetchTransactions(
+    nodeUrl: string,
+    filterOn: string,
     address: string,
-    orderBy?: "ORDER_BY_DESC" | "ORDER_BY_ASC",
-  ) {
-    return await network({
+    options: {
+      "pagination.limit"?: number;
```
lvndry, line 289: is there any possibility there are undefined ? From what I see in `fetchAllTransactions` it doesn't seem to be the case

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -229,49 +229,94 @@ export class CosmosAPI {
...
+      "pagination.reverse"?: boolean;
+    },
+  ): Promise<{
+    txs: CosmosTx[];
+    nextPageToken: string | undefined;
+    total: number;
+  }> {
+    let serializedOptions = "";
+    for (const key of Object.keys(options)) {
+      serializedOptions += options[key] != null ? `&${key}=${options[key]}` : "";
```
lvndry, line 300: Typing says it's `number | undefined` but here you check for `null`

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -229,49 +229,94 @@ export class CosmosAPI {
...
 
-  private async getSendingTransactions(
+  private async fetchTransactions(
+    nodeUrl: string,
+    filterOn: string,
     address: string,
-    orderBy?: "ORDER_BY_DESC" | "ORDER_BY_ASC",
-  ) {
-    return await network({
+    options: {
```
lvndry, line 288: could be nice to create a type object for it

```ts
type FetchTransactionOptions = {
      "pagination.limit"?: number;
       "pagination.offset"?: number;
       "pagination.reverse"?: boolean;
}
```

could you add a comment on what "pagination.reverse" does please ?

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: Yep because it's an optional parameter on the node endpoint

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: We check for null and undefined (not !==)

I want to include both conditions

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: This type is only used here so I don't think we need to create a type just for one iteration

WDYT

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
henrily-ledger: is !options[key] is the same as options[key] != null ? 
I find is less disturbing to have !options[key] than options[key] != null 
`!=` is weird for me

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -229,49 +229,94 @@ export class CosmosAPI {
...
+    const recipientTxs = await this.fetchAllTransactions(
+      address,
+      "transfer.recipient",
+      paginationSize,
+    );
 
+    return [...senderTxs, ...recipientTxs];
+  };
+
+  private async fetchAllTransactions(address: string, filterOn: string, paginationSize: number) {
```
henrily-ledger, line 243: If we know what is parameter on `filterOn` shouldn't we just specify it with : `"message.sender" || "transfert.recipient"`
Being less permissive sounds better to me

henrily-ledger:
Looks good to me but shouldn't it be better if we do this as recursive loop instead of iteration ? I don't know the benchmark performance iteration vs recursive
It's up to you I wont block this PR for this choice

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: In our app we use those two values but any filter on a transaction could be used, I don't know what is the best way to do it :shrug: 

I can do as you said, restrict on what we know is used on the app

henrily-ledger: If we can be more precise it's better for typing

hedi-edelbloute: Done

PR #3696, opened 2023-06-15
feat: LIVE-7904 make listAppsV2 feature flagged

juan-cortes:
### 📝 Description

The development of the faster access to My Ledger feature was done a few months ago but has been hidden behind an experimental setting on LLD/LLM meaning we couldn't enable it remotely or roll it to a percentage of our userbase to gather feedback. This PR addresses that issue by making it rely on a feature flag instead and actively listening to changes in its value to trigger a switch consumed on live-common to determine which version to use.

There are still some inconsistencies in the data returned by v2 versus v1 but this should make it easier to work with. Also changing the value no longer requires a restart of the application.

### ❓ Context

- **Impacted projects**: `desktop, mobile, common, types` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7904` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo here.

### 🚀 Expectations to reach
App should remain the same unless you override the value in the feature flags section of the developer menu.


apps/cli/src/commands/managerListApps.ts
```diff
@@ -46,8 +55,8 @@ export default {
           format === "raw"
             ? r
             : format === "json"
-            ? JSON.stringify(r)
-            : r.appsListNames
+              ? JSON.stringify(r)
+              : r.appsListNames
```
alexandremgo, line 59: I think there is just a small formatting pb here ?

alexandremgo (approved):
🏎️

libs/ledger-live-common/src/apps/hw.ts
```diff
@@ -24,8 +23,9 @@ export const execWithTransport =
  *   currency and parent application. And therefor we no longer need the version
  *   check that broke that dependency after a certain version for ETH and BTC.
  * - Remove all the legacy v1 code, and tests.
- * - Remove the env definition and all the forks based on it here.
+ * - Cleanup the feature flag that governs this.
  */
-// Nb Written this way to respect runtime changes to the env.
+let listAppsV2Enabled = false;
```
gre, line 28: is there another way not to have a local state living in the module? couldn't an env still be used but the featureFlag effect will just `setEnv` it? 🤔

juan-cortes: Considering the idea is to drop the fork in the logic between v1 and v2 pretty soon and not leave this permanently, I'd argue it's acceptable to have a local state for a couple of weeks.

gre: ok, could a jira ticket be created about this + with a comment of that ticket link? 👍

gre (approved):

PR #3704, opened 2023-06-16
[FEAT][LIVE-7406] create evm send flow (LLD)

chabroA:
### 📝 Description

Create send flow for EVM family, using GasTrackerAbstraction introduced in https://github.com/LedgerHQ/ledger-live/pull/3611

Notes:
- Added Ethereum and Polygon as evm family currencies for tests. Inspired by https://github.com/LedgerHQ/ledger-live/pull/3714
- I don't know where [this doc update](https://github.com/LedgerHQ/ledger-live/pull/3704/commits/0b3df7eda7cb29ccd062447e9ccd1a6a8036b485) come from but it is generated by `pnpm doc:ljs` and if not included, the _Test docs_ CI job fails (cf. [here](https://github.com/LedgerHQ/ledger-live/actions/runs/5291680940/jobs/9577549093) for example)

### ❓ Context

- **Impacted projects**: LLD LLC coin-evm
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7406 - https://ledgerhq.atlassian.net/browse/LIVE-6260

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/GasPriceField.tsx
```diff
@@ -0,0 +1,60 @@
...
+import FeeSliderField from "~/renderer/components/FeeSliderField";
+import { Range, inferDynamicRange } from "@ledgerhq/live-common/range";
+type Props = {
+  account: Account;
+  transaction: Transaction;
+  status: TransactionStatus;
+  updateTransaction: (updater: (_: Transaction) => Transaction) => void;
+};
+const fallbackGasPrice = inferDynamicRange(BigNumber(10e9));
+let lastNetworkGasPrice: Range | undefined; // local cache of last value to prevent extra blinks
```
lambertkevin, line 16: If I remember correctly, this is directly coming from the old implementation of the Ethereum family, and I'm wondering if you tried testing if it was really necessary ? 
I think it's still very unclear what it actually does, the random `10e9` (1 gwei) is coming out of nowhere with no good reason for it, etc etc...

I would really love if we could challenge this and make sure this is useful for us, especially since the `gasOptions` is supposed to always exist if we're seeing this component.

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/GasPriceField.tsx
```diff
@@ -0,0 +1,60 @@
...
+          feesStrategy: "custom",
+        }),
+      );
+    },
+    [updateTransaction, bridge],
+  );
+
+  const { gasOptions } = transaction;
+
+  // FIXME: gasPrice should always be defined if gasOptions is defined
```
lambertkevin, line 35: Why the need for a FIXME here ? 
We're creating thing from scratch here, we can't allow any FIXME/TODO imo

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/MaxFeeField.tsx
```diff
@@ -0,0 +1,128 @@
...
+`;
+
+const WhiteSpacedLabel = styled(Label)`
+  white-space: pre;
+  color: ${p => p.theme.colors.neutral.c60};
+`;
+
+const FeesField: NonNullable<EvmFamily["sendAmountFields"]>["component"] = ({
+  account,
+  parentAccount,
```
lambertkevin, line 50: I'm seeing an alert here about `parentAccount` being deprecated ? 🤔

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/GasLimitField.tsx
```diff
@@ -0,0 +1,95 @@
...
+              {t("send.steps.details.edit")}
+            </Box>
+          </Button>
+        </Box>
+      )}
+    </Box>
+  );
+};
+
+export default AdvancedOptions;
```
lambertkevin, line 95: Please memo this 🙏

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/GasPriceField.tsx
```diff
@@ -0,0 +1,60 @@
...
+      range={range}
+      defaultValue={range.initial}
+      value={gasPrice}
+      onChange={onGasPriceChange}
+      unit={units.length > 1 ? units[1] : units[0]}
+      error={status.errors.gasPrice}
+    />
+  );
+};
+export default FeesField;
```
lambertkevin, line 60: Please memo this 🙏

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/MaxFeeField.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </ErrorContainer>
+      <WhiteSpacedLabel>
+        {`${t("send.steps.details.nextBlock")} : `}
+        <FeesValues>{nextBaseFeeValue}</FeesValues>
+      </WhiteSpacedLabel>
+    </Box>
+  );
+};
+
+export default FeesField;
```
lambertkevin, line 128: Please memo this 🙏

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/PriorityFeeField.tsx
```diff
@@ -0,0 +1,145 @@
...
+        {`${t("send.steps.details.suggested")} : `}
+        <FeesValues>
+          {lowPriorityFeeValue} - {highPriorityFeeValue} {unitName}
+        </FeesValues>
+      </WhiteSpacedLabel>
+    </Box>
+  );
+};
+
+export default FeesField;
```
lambertkevin, line 145: Please memo this 🙏

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/PriorityFeeField.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+type Props = {
+  account: AccountLike;
+  parentAccount: Account | null | undefined;
+  transaction: Transaction;
+  status: TransactionStatus;
+  updateTransaction: Result<Transaction>["updateTransaction"];
+};
+
+const fallbackMaxPriorityFeePerGas = inferDynamicRange(new BigNumber(10e9));
```
lambertkevin, line 57: Same thing here, this value is a bit random and I would really like if we could avoid having it without proper justfication of why it's here, and why this 1 Gwei value

lambertkevin: Same thing here, this value is a bit random and I would really like if we could avoid having it without proper justfication of why it's here, and why this 10 Gwei value.
With the `gasOptions` being mandatory to see this component, I'm not sure we need any fallback

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/SelectFeeStrategy.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const feesCurrency = getAccountCurrency(mainAccount);
+
+  return (
+    <Box horizontal justifyContent="center" flexWrap="wrap" gap="16px">
+      {strategies.map(strategy => {
+        const selected = transaction.feesStrategy === strategy;
+
+        const gasOption = gasOptions[strategy];
+
+        const estimatedFees = getEstimatedFees(getTypedTransaction(transaction, gasOption));
```
lambertkevin, line 88: Maybe a `useMemo` just to maintain the ref of this variable ?

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const onFeeStrategyClick = useCallback(
+    ({ feesStrategy }) => {
+      updateTransaction((tx: EvmTransaction) =>
+        bridge.updateTransaction(tx, {
+          feesStrategy,
+        }),
+      );
+      if (drawerState.open) setDrawer(undefined);
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
lambertkevin, line 48: Is this still necessary ?

libs/coin-evm/src/api/rpc/rpc.common.ts
```diff
@@ -130,6 +130,15 @@ export const getGasEstimation = (
...
   withApi(account.currency, async api => {
+    /**
+     * NOTE: either we do this here, or we duplicate this logic in all functions
+     * using getGasEstimation (in order not to update the tx gasLimit when in
+     * "custom" mode)
+     */
+    if (transaction.feesStrategy === "custom") {
+      return transaction.gasLimit;
+    }
+
```
lambertkevin, line 141: Ok so you made this for the `gasLimit`. Let's discuss this when you come back so we can assess the differences between the 2 strategies of `useGasLimit` vs custom mode, but I think the custom mode has more drawbacks

libs/coin-evm/src/bridge/js.ts
```diff
@@ -20,7 +21,8 @@ const updateTransaction: AccountBridge<EvmTransaction>["updateTransaction"] = (
   transaction,
   patch,
 ) => {
-  return { ...transaction, ...patch } as EvmTransaction;
+  const newTransaction = { ...transaction, ...patch } as EvmTransaction;
+  return isEqual(transaction, newTransaction) ? transaction : newTransaction;
```
lambertkevin, line 25: Please cover this with tests 🙏

libs/coin-evm/src/prepareTransaction.ts
```diff
@@ -5,7 +5,8 @@ import { getFeesEstimation, getGasEstimation, getTransactionCount } from "./api/
 import { validateRecipient } from "./getTransactionStatus";
 import { getAdditionalLayer2Fees, getEstimatedFees } from "./logic";
 import { getTransactionData, getTypedTransaction } from "./transaction";
-import { Transaction as EvmTransaction } from "./types";
+import { Transaction as EvmTransaction, FeeData } from "./types";
+import { isEqual } from "lodash";
```
lambertkevin, line 9: Should be with the other non relative imports at the top

lambertkevin (changes requested):
Nothing major, but still some things I would like to get rid of if they're unecessary.

Also would be cool to cover the new UIs with React integrations test now that it's possible to make them :pray: 
(You can find examples in the @ledgerhq/domain-service lib)

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/index.tsx
chabroA: Indeed the `drawerState` does not seem to be needed or used in this component

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/SelectFeeStrategy.tsx
chabroA: Not possible here since this `estimatedFees` is defined inside a callback and for each strategy in `strategies`. Another possibility would be to manually and individually define an estimated fee value for each strategy before the return statement and use a switch case or if cascade inside the callback to select the correct estimated fee value every time. Not really worth the added complexity and reduced readability IMHO.

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/GasPriceField.tsx
chabroA: Indeed, no need for a FIXME here, forgot to remove that after I finished the implementation.
But I still added an `invariant` on `gasOptions` for good mesures since `gasOptions` should always be defined at this stage, even though according to TS, `transaction. gasOptions` is of type `GasOptions | undefined`

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/PriorityFeeField.tsx
chabroA: Indeed, with the use of `gasOptions` and since it should always be defined at this stage, there is no need for an hardcoded fallback value anymore.

chabroA:
> Nothing major, but still some things I would like to get rid of if they're unecessary.
> 
> Also would be cool to cover the new UIs with React integrations test now that it's possible to make them 🙏 (You can find examples in the @ledgerhq/domain-service lib)

Fixed all comments and handled user defined custom gas limit. Will update LLM UI on another PR branching from this one

libs/coin-evm/src/prepareTransaction.ts
```diff
@@ -1,13 +1,13 @@
...
-import { validateRecipient } from "./getTransactionStatus";
+import { Account, TokenAccount } from "@ledgerhq/types-live";
+import BigNumber from "bignumber.js";
+import { isEqual } from "lodash";
 import { getNftCollectionMetadata } from "./api/nft";
 import { getFeesEstimation, getGasEstimation, getTransactionCount } from "./api/rpc";
+import { validateRecipient } from "./getTransactionStatus";
 import { getAdditionalLayer2Fees, getEstimatedFees, isNftTransaction } from "./logic";
+import { getTransactionData, getTypedTransaction } from "./transaction";
+import { EvmNftTransaction, Transaction as EvmTransaction, FeeData } from "./types";
```
lambertkevin, line 11: My beautiful pyramidal order 😢

libs/coin-framework/src/bridge/jsHelpers.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        recipient: "0x17733CAb76d9A2112576443F21735789733B1ca3",
+        amount: new BigNumber("10000000000000"),
+      };
+
+      const updatedTransaction = defaultUpdateTransaction(transaction, {
+        recipient: "0x17733CAb76d9A2112576443F21735789733B1ca3",
+        amount: new BigNumber("10000000000000"),
+      });
+
+      expect(Object.is(transaction, updatedTransaction)).toBe(true);
```
lambertkevin, line 18: You can use `.toBe` instead of `Object.is` because `.toBe` is already using `Object.is` under the hood.

libs/coin-framework/src/bridge/jsHelpers.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        recipient: "0x17733CAb76d9A2112576443F21735789733B1ca3",
+        amount: new BigNumber("10000000000000"),
+      };
+
+      const updatedTransaction = defaultUpdateTransaction(transaction, {
+        recipient: "0x17733CAb76d9A2112576443F21735789733B1ca3",
+        amount: new BigNumber("20000000000000"),
+      });
+
+      expect(Object.is(transaction, updatedTransaction)).toBe(false);
```
lambertkevin, line 32: Same heere

libs/ledger-live-common/src/families/evm/react.unit.test.ts
lambertkevin, line 1: Component UI testing should be named `*.react.test.ts` 🙏

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
```diff
@@ -3614,6 +3614,7 @@ export const cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency> = {
         uri: "https://api.etherscan.io",
         type: "etherscan",
       },
+      gasTracker: { uri: "", type: "ledger" },
```
lambertkevin, line 3617: Please add this as well for `polygon_as_evm_test_only" 🙏

lambertkevin (commented):
Just a few things, nothing major again (except maybe the condition in the `getTransactionStatus`) and nice to have tests for the hooks 👍

If you have the time & will, would be cool to have UI component testing for the new family components (`GasLimitField`, `GasPriceField` etc)

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/SelectFeeStrategy.tsx
lambertkevin: Nevermind just realised that it's not helping for re-renders here, the solution would be to memoize directly the result for each strategy, either with a subcomponent or with a `useMemo` of the expected JSX for each `gasOption` 🤔

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/GasPriceField.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+
+  const { gasOptions } = transaction;
+
+  invariant(gasOptions, "GasPriceField: 'transaction.gasOptions' should be defined");
+
+  const networkGasPrice = inferDynamicRange(gasOptions.medium.gasPrice!, {
+    minValue: gasOptions.slow.gasPrice!,
+    maxValue: gasOptions.fast.gasPrice!,
+  });
```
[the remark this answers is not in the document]
chabroA: Done here ae06ffea406c839f994731f7f607c8d46a8dfb91

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/MaxFeeField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      updateTransaction(() =>
+        bridge.updateTransaction(transaction, {
+          maxFeePerGas,
+          feesStrategy: "custom",
+        }),
+      ),
+    [bridge, transaction, updateTransaction],
+  );
+
+  const maxFeePerGas = transaction.maxFeePerGas || transaction.gasOptions?.medium.maxFeePerGas;
```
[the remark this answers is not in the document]
chabroA: Done here be3c1922f2d7beb2c932af75d029ab91def2ed12

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/PriorityFeeField.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  strategies.forEach(strategy => {
+    invariant(
+      gasOptions[strategy].maxPriorityFeePerGas,
+      `PriorityFeeField: 'gasOptions[${strategy}].maxPriorityFeePerGas' should be defined`,
+    );
+  });
+
+  const maxPriorityFee =
+    transaction.maxPriorityFeePerGas ?? gasOptions.medium.maxPriorityFeePerGas!;
```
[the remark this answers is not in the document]
chabroA: Done here be3c1922f2d7beb2c932af75d029ab91def2ed12

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  if (error) {
+    log("error", error.message);
+  }
+
+  useEffect(() => {
+    updateTransaction((tx: EvmTransaction) => bridge.updateTransaction(tx, { ...tx, gasOptions }));
+  }, [bridge, updateTransaction, gasOptions]);
+
+  const [isAdvanceMode, setAdvanceMode] = useState(!transaction.feesStrategy);
```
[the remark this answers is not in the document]
chabroA: Indeed, `!transaction.feesStrategy` will always be false, and that's actually what we want `isAdvanceMode` default value to be. Since it's juste a dumb UI toggle.
Has been replaced by `const [isAdvanceMode, setAdvanceMode] = useState(false);` here 12e5e0e2331d0728197343f23f273683f9021295

libs/coin-framework/src/bridge/jsHelpers.test.ts
chabroA: Done here c1cfc6d076467b2b15b699e50b3fecb29cba1dbd
Got confused by the last paragraph of the [Jest `toBe` doc](https://jestjs.io/docs/expect#tobevalue) that is actually only related to debug

libs/ledger-live-common/src/families/evm/react.unit.test.ts
chabroA: Done here -> 64e3836fd892785bbf18cf5dc1da72111f47573d

`react.react.test.ts` 🤡 😅 🤷‍♂️

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
chabroA: Done here a4ce505d893dd6ebdd7b1b9753f883029d3d1306

libs/coin-evm/src/prepareTransaction.ts
chabroA: Indeed, but VSCode automation and shortcuts https://code.visualstudio.com/docs/typescript/typescript-refactoring#_organize-imports feels much better and powerful than manual imports ordering 👨‍💻

lambertkevin: But it's not... PYRAMIDAL. 📐🎨

libs/coin-evm/src/getTransactionStatus.ts
```diff
@@ -129,6 +139,24 @@ export const validateGas = (
     errors.maxPriorityFee = new PriorityFeeTooLow();
   }
 
+  if (maximalMaxFeePerGas && tx.maxFeePerGas?.gt(maximalMaxFeePerGas)) {
+    // priority fee is more than max fee (total fee for the transaction) which doesn't make sense
+    errors.maxPriorityFee = new PriorityFeeHigherThanMaxFee();
+  }
```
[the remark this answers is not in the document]
chabroA: Indeed, the correct version is:
```
if (tx.type === 2 && tx.maxPriorityFeePerGas.gt(tx.maxFeePerGas)) {
  errors.maxPriorityFee = new PriorityFeeHigherThanMaxFee();
}
```

I wanted to migrate this condition from the Ethereum family, but it looks like I messed up somewhere:
```
  if (tx.maxPriorityFeePerGas?.isGreaterThan(tx.maxFeePerGas || 0)) {
    // priority fee is more than max fee (total fee for the transaction) which doesn't make sense
    errors.maxPriorityFee = new PriorityFeeHigherThanMaxFee();
  }
```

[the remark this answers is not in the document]
chabroA: Fixed here 816f83009a37c84f9ed9d0d93e8155b045859a12

lambertkevin (approved):
🚀  Great work !

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/SelectFeeStrategy.tsx
```diff
@@ -0,0 +1,163 @@
...
+import TachometerHigh from "~/renderer/icons/TachometerHigh";
+import TachometerLow from "~/renderer/icons/TachometerLow";
+import TachometerMedium from "~/renderer/icons/TachometerMedium";
+
+type OnClickType = {
+  feesStrategy: keyof GasOptions;
+};
+
+type Props = {
+  onClick: (arg: OnClickType) => void;
```
lvndry, line 23:
```suggestion
  onClick: (arg: keyof GasOptions) => void;
```

Are these two options the same ? If yes could you go with the latter please ?

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const bridge: AccountBridge<EvmTransaction> = getAccountBridge(account);
+
+  const [gasOptions, error] = useGasOptions({
+    currency: account.currency,
+    transaction,
+    interval: account.currency.blockAvgTime ? account.currency.blockAvgTime * 1000 : undefined,
+  });
+
+  if (error) {
+    log("error", error.message);
```
lvndry, line 26:
```suggestion
    log("SendAmountFields/index.tsx", error.message);
```

Could be worth being more specific to help with future debugging

libs/coin-evm/src/getTransactionStatus.ts
```diff
@@ -104,12 +108,25 @@ export const validateAmount = (
...
+  gasLimit: BigNumber,
+  customGasLimit?: BigNumber,
 ): Array<ValidationIssues> => {
   const errors: ValidationIssues = {};
   const warnings: ValidationIssues = {};
 
+  // Gas Limit
+  if (gasLimit.isZero()) {
+    errors.gasLimit = new FeeNotLoaded(); // "Could not load fee rates. Please set manual fees"
+  } else if (gasLimit.isLessThan(21000)) {
```
lvndry, line 121: could you make this a constant ?

libs/coin-evm/src/getTransactionStatus.ts
```diff
@@ -118,15 +135,32 @@ export const validateGas = (
...
-    // minimum gas for a tx is 21000
-    errors.gasLimit = new GasLessThanEstimate(); // "This may be too low. Please increase"
   } else if (tx.recipient && estimatedFees.gt(account.balance)) {
     errors.gasPrice = new NotEnoughGas(); // "The parent account balance is insufficient for network fees"
-  } else if (tx.maxPriorityFeePerGas && tx.maxPriorityFeePerGas.isZero()) {
-    errors.maxPriorityFee = new PriorityFeeTooLow();
+  }
+
+  // Gas Price for EIP-1559
+  if (tx.type === 2) {
```
lvndry, line 143: Don't know how big of a change it would be but could be nice to have type be an enum instead:

```ts
enum TransactionType = {
 LEGACY = 0;
 EIP2718 = 1;
 EIP1559 = 2;
}
```

or something like that

libs/coin-evm/src/logic.ts
```diff
@@ -27,14 +27,21 @@ export const legacyTransactionHasFees = (tx: EvmTransactionLegacy): boolean =>
 export const eip1559TransactionHasFees = (tx: EvmTransactionEIP1559): boolean =>
   Boolean(tx.type === 2 && tx.maxFeePerGas && tx.maxPriorityFeePerGas);
 
+/**
```
lvndry, line 30: Out of the scope of this PR but `logic.ts` means a bit nothing to me. Seems like we only talk about fees in this file could we give it a more relevant name ?

lvndry, line 30: Out of the scope of this PR but `logic.ts` means a bit nothing to me. Seems like we only talk about fees in this file could we give it a more relevant name ?

chabroA: Naming could probably be improved, not strong opinion on that.
As of today it looks like this `logic.ts` file is part of the _coin integration framework_ and should hold the "coin specific business logic"

cf. https://github.com/LedgerHQ/developer-portal/pull/773/files#r1255806102

libs/coin-evm/src/getTransactionStatus.ts
chabroA: Good point, will take a look at that (will probably do it in a separate dedicated PR to avoid off topic changes)
cc. @lambertkevin  FYI

lambertkevin: Why not, but I'm not 100% sure about how useful it would be at the end for devs, because now that forces everyone to have an understanding and knowledge about what transaction's type = what EIP. And in the case of 1559, it's a famous one so people should know, but for type 1 for example, I'm not so sure. 

Anyway, that can be up to you both to decide if you think it's a better DX or not, but indeed, let's take this in another PR 👌

libs/coin-evm/src/logic.ts
lambertkevin: Yup agreed, `logic.ts` isn't an amazing name, but that's kinda standard today in the coin integration. 
Naming can surely be improved, but I'm afraid we'll end up with something like `helpers.ts`, and I'm not sure it's really  better 🤷‍♂️

apps/ledger-live-desktop/src/renderer/families/evm/SendAmountFields/SelectFeeStrategy.tsx
chabroA: Done here 9c3313126b343e8815c7c7733aa81e528824df1f

libs/coin-evm/src/getTransactionStatus.ts
lambertkevin: Already exists in transaction.ts if you need want it

lambertkevin: Already exists in transaction.ts if you need want it

libs/coin-evm/src/getTransactionStatus.ts
lvndry: Yeah I mean if any developer is making change in our code they better know what they are doing 😅 I think it's a good thing they can be aware of the change they make and not do anything blindly.
Sounds good for doing on an other PR

lambertkevin: Knowing our code base != knowing all EIPs numbers even the ones we don't use yet (EIP-2718 is an example of that), so I'm not sure it's improving the readability and how easy it is to understand vs type 0,1,2 🤷‍♂️
But maybe we can do a small POC to test and ask the team's opinion on that 👍

PR #3713, opened 2023-06-16
fix: unable to close root drawer without navigation

kallen-ledger:
### 📝 Description

Unable to close root drawer without navigation.

* Created `RootDrawerProvider` to handle the logic of showing or hiding the drawer.
* In the onClose function create a one time listener that when triggered will call the onClose callback.
* Emit an event when the onModalHide component finishes. This is to allow the animation to complete fully.

### ❓ Context

- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/context/RootDrawerContext.tsx
+++ b/apps/ledger-live-mobile/src/context/RootDrawerContext.tsx
@@ -1,0 +1,81 @@
+import EventEmitter from "events";
+import React, { PropsWithChildren, createContext, useCallback, useContext, useState } from "react";
+import { RootDrawerProps } from "../components/RootDrawer";
+import { ParamListBase, useNavigation } from "@react-navigation/native";
+import { StackNavigationProp } from "@react-navigation/stack";
+import { NavigatorName } from "../const";
+
+const rootDrawerEmitter = new EventEmitter();
+
+type Props = {
+  drawer?: RootDrawerProps;
+};
+
+type RootDrawerContext = {
+  drawer: RootDrawerProps;
+  isOpen: boolean;
+  onModalHide(): void;
+  onClose(cb?: () => void): void;
+  openDrawer(): void;
+};
+
+export const RootDrawerContext = createContext<RootDrawerContext>({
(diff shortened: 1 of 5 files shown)


apps/ledger-live-mobile/src/context/RootDrawerContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+  openDrawer(): void;
+};
+
+export const RootDrawerContext = createContext<RootDrawerContext>({
+  drawer: {} as RootDrawerProps,
+  isOpen: false,
+  onModalHide: () => undefined,
+  onClose: cb => {
+    if (cb) cb();
+  },
```
sarneijim, line 28: I find confusing the difference between `onModalHide` and `onClose`. Are they different buttons? If not could we unify this in the same method `closeDrawer`?

kallen-ledger: I was confused too 😅 

On close is called when we initiate the close on the modal. The onModalHide callback is called once close animation has completed.
If we were to set drawer to undefined in the onClose the drawer would "pop" out of view instead of animate.

sarneijim: Would someone change `onClose` default action in that case? Or could we remove this prop? (I am thinking that it is expected have a common behaviour across the app?)

kallen-ledger: One solution is to have a timeout in the onClose so we will set drawer to undefined after a certain time. So the animation can complete.

sarneijim: Thanks @kallen-ledger for the explanation. As`onClose` and `onModalHide` are props in `QueuedDrawer` component. This pr is approved by my side :)

sarneijim (approved):

PR #3717, opened 2023-06-16
[support] Cleanup unused dependencies in coin-* libs

haammar-ledger:
### 📝 Description

Remove some unused dependencies in our `coin-*` libs.
Found using [depcheck](https://github.com/depcheck/depcheck).

### ❓ Context

- **Impacted projects**: `coin-framework` `coin-polkadot` `coin-evm` 
- **Linked resource(s)**: N/A 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Nothing to demo.

### 🚀 Expectations to reach

Live still builds. 100% covered by CI.


sprohaszka-ledger (approved):
LGTM :)

gre:
> Found using [depcheck](https://github.com/depcheck/depcheck).

interesting tool :notbad:

today, we also have `unimported` that can detect not only dependencies, but also dead code, we have secured this on LLD and LLM, would you recon it be worth doing the same for literally all our libraries then?

(good thing is we can even configure it finely like this https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/.unimportedrc.json and then the CI will guarantee we don't leave deadcode behind )

gre (approved):

haammar-ledger:
> today, we also have unimported that can detect not only dependencies, but also dead code, we have secured this on LLD and LLM, would you recon it be worth doing the same for literally all our libraries then?

@gre definitely, actually I've been trying to use it from a script, using [its API](https://github.com/depcheck/depcheck#api), it worked but I just wasn't able to find the right conf to provide, it always had either false positives or false negatives... I didn't want to spend too much time right now, but with the right config maybe we could get a reliable list for all of our projects, yeah.

PR #3739, opened 2023-06-19
fix: llm - loading between webviews & flickering header text

cksanders:
### 📝 Description

PR to:

1. Fix flickering in the header by removing the buggy onLoadProgress hook. `isWhitelistedDomain` was being updated by multiple WebView state hooks, including `onLoadProgress` which has the previous `nativeEvent.url` rather than current ... meaning header text flickered due to `isWhitelistedDomain` transitioning from `false` to `true` to `false`.

2. Add a loading spinner when navigating between WebViews in WebPTXPLayer

### ❓ Context

- **Impacted projects**: Closes https://ledgerhq.atlassian.net/browse/LIVE-7930 & https://ledgerhq.atlassian.net/browse/LIVE-7927
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/grumpy-cameras-bake.md
+++ b/.changeset/grumpy-cameras-bake.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Fix flickering in header by removing buggy onLoadProgress hook
--- a/.changeset/stupid-sloths-explain.md
+++ b/.changeset/stupid-sloths-explain.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Add loading spinner when navigating between web pages in WebPTXPLayer
--- a/apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
+++ b/apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
@@ -24,6 +24,7 @@ import { initialWebviewState } from "../Web3AppWebview/helpers";
 import { track } from "../../analytics";
 import { NavigationHeaderCloseButtonAdvanced } from "../NavigationHeaderCloseButton";
 import { NavigatorName } from "../../const";
+import { Loading } from "../Loading";
 
 type BackToWhitelistedDomainProps = {
   manifest: AppManifest;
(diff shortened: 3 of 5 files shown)


gre (approved):

.changeset/stupid-sloths-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Add loading spinner when navigating between WebViews in WebPTXPLayer
```
cpellizzari-ledger, line 5: just "between webpages" actually

cpellizzari-ledger (approved):

.changeset/grumpy-cameras-bake.md
```diff
@@ -0,0 +1,6 @@
+
```
Justkant, line 1: We might want to remove that first empty line, I have no idea if it affects changeset but I wouldn't risk it
```suggestion
```

IAmMorrow:
As of today I can't guarantee that removing `onLoadProgress` will have no undesired side effect on other features / the general webview ux. I feel that we probably added it for a reason. I might be wrong but I remember getting this webview thing to work right was kind of painful 😅 

If there is a regression, another solution that can be explored is not syncing the whole state when the `onLoadProgress` event fire !

apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
```diff
@@ -185,6 +191,7 @@ export const WebPTXPlayer = ({ manifest, inputs }: Props) => {
         inputs={inputs}
         onStateChange={setWebviewState}
       />
+      {isLoading ? <Loading /> : null}
```
Justkant, line 194: You can just use `webviewState.loading` IMO
```suggestion
      {webviewState.loading ? <Loading /> : null}
```

Justkant (changes requested):
Clean, I just have 2 small changes I would make

apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
cksanders: @cpellizzari-ledger this is what you were talking about - thanks!

Justkant (approved):

PR #3767, opened 2023-06-21
remove critical error on navigation

hedi-edelbloute:
### 📝 Description

Attempt to remove sentry navigation error that occur every prerelease

We shouldn't have a critical error when user stops his navigation flow

### ❓ Context

- **Linked resource(s)**: Sentry 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/pink-dogs-know.md
+++ b/.changeset/pink-dogs-know.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Remove critical error on navigation
--- a/apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
@@ -264,8 +264,8 @@ export default function BaseNavigator() {
               const onError =
                 route.params?.onError || (route.params as unknown as typeof route)?.params?.onError;
               // @TODO replace with correct error
-              if (onError && typeof onError === "function")
-                onError(route.params.error || new Error("Request account interrupted by user"));
+              if (onError && typeof onError === "function" && route.params.error)
+                onError(route.params.error);
             },
           })}
         />
--- a/apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
@@ -263,7 +263,8 @@ export default function BaseNavigator() {
             */
               const onError =
                 route.params?.onError || (route.params as unknown as typeof route)?.params?.onError;
-              // @TODO replace with correct error
+              // @TODO This route.params.error mechanism is deprecated, no part of the code is currently using it
+              // WalletAPI are suggesting they will rework that at some point
               if (onError && typeof onError === "function" && route.params.error)
                 onError(route.params.error);
             },
--- a/apps/ledger-live-mobile/src/components/Stake/index.tsx
+++ b/apps/ledger-live-mobile/src/components/Stake/index.tsx
@@ -7,7 +7,6 @@ import { listCurrencies, filterCurrencies } from "@ledgerhq/live-common/currenci
 import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
 import { NavigatorName, ScreenName } from "../../const";
 import perFamilyAccountActions from "../../generated/accountActions";
-import logger from "../../logger";
 import type { StackNavigatorProps, BaseComposite } from "../RootNavigator/types/helpers";
 import type { StakeNavigatorParamList } from "../RootNavigator/types/StakeNavigator";
 
@@ -59,9 +58,6 @@ const StakeFlow = ({ route }: Props) => {
     },
     [navigation, parentRoute],
   );
-  const onError = (error: Error) => {
-    logger.critical(error);
-  };
 
   const requestAccount = useCallback(() => {
     if (cryptoCurrencies.length === 1) {
@@ -72,7 +68,6 @@ const StakeFlow = ({ route }: Props) => {
           currency: cryptoCurrencies[0],
           onSuccess,
         },
-        onError,
       });
     } else {
       navigation.replace(NavigatorName.RequestAccount, {
@@ -82,7 +77,6 @@ const StakeFlow = ({ route }: Props) => {
           allowAddAccount: true,
           onSuccess,
         },
-        onError,
       });
     }
   }, [cryptoCurrencies, navigation, onSuccess]);


.changeset/pink-dogs-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"live-mobile": minor
```
Justkant, line 2: Can be a patch IMO

Justkant (approved):

.changeset/pink-dogs-know.md
hedi-edelbloute: Done, thanks

sarneijim:
Not really happy with the solution, we will be hidden any error.
Solutions that I have in mind.
Happy to modify the error level from critical to warning or silent that error for that specific component.
Or even better remove filter this 'fake error' in the origin
image

sarneijim (approved):

mcayuelas-ledger (approved):

Justkant:
> Not really happy with this solution, we will be hidden any error. Solutions that I have in mind:
> 
> * Modify the error level from critical to warning.
> * Filter this 'fake error' in the origin
>   image

IMO we should not rely on `beforeRemove` at all because as it is now it can and probably will call `onError` if present even if there is no error and that the user never cancelled but probably just finished

Also I don't think we will be hiding any errors because I couldn't find a place where we set error on the params of the navigator, if we can find one place setting error on the params we should migrate to another solution like calling onError directly

We need to keep in mind that we depend on `onError` in the platform and wallet-api currently

hedi-edelbloute:
> > Not really happy with this solution, we will be hidden any error. Solutions that I have in mind:
> > 
> > * Modify the error level from critical to warning.
> > * Filter this 'fake error' in the origin
> >   image
> 
> IMO we should not rely on `beforeRemove` at all because as it is now it can and probably will call `onError` if present even if there is no error and that the user never cancelled but probably just finished
> 
> Also I don't think we will be hiding any errors because I couldn't find a place where we set error on the params of the navigator, if we can find one place setting error on the params we should migrate to another solution like calling onError directly
> 
> We need to keep in mind that we depend on `onError` in the platform and wallet-api currently

If I understood well, we should then just remove the onError call in StakeFlow

hedi-edelbloute:
Can you guys confirm current changes are what we want ? thanks !

apps/ledger-live-mobile/src/components/Stake/index.tsx
```diff
@@ -59,9 +59,6 @@ const StakeFlow = ({ route }: Props) => {
     },
     [navigation, parentRoute],
   );
-  const onError = (error: Error) => {
-    logger.critical(error);
-  };
```
sarneijim, line 64: I think we need restore this.

Justkant: IMO you don't and that's the whole point of the PR, if you are able to find any other message than the one we want to remove I'm happy to do this PR another way but currently there is only on error message sent here and it's `Request account interrupted by user` nothing else.

Did you check my comment above ? (https://github.com/LedgerHQ/ledger-live/pull/3767#issuecomment-1600839757)

I've checked all usage of `NavigatorName.RequestAccount` and we never set `params.error`, so there is no chance of any other error landing in your onError fonction, I've also check all other navigator setting `beforeRemove` and none is using the `params.error` so again no possibility of any other message.

I've also seen that your implementation in `Stake/index.tsx` is the only one setting the `onError` callback outside the `params` of the screen, it's okay and worked because the object itself represents the params of the navigator but it was not matching the other places in our codebase.

Justkant:
> Can you guys confirm current changes are what we want ? thanks !

IMO yes this is exactly what we want, I've already spent to much time on that non subject, we need to move on instead of discussing that for ages

Justkant:
For the tests you probably need to rebase the PR

apps/ledger-live-mobile/src/components/Stake/index.tsx
sarneijim: @Justkant , what is it the problem to listen here them?

Justkant: I've explained the problem multiple times already

> currently there is only on error message sent here and it's `Request account interrupted by user` nothing else

I've not implemented this logic, it's here from a long time ago, no idea who started that pattern.
No one but the platform and wallet-api is actually using this (like I said already too) and I've said that we want to change that in the future as it's not a good solution to send errors back

hedi-edelbloute:
Linter has been fixed

apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
```diff
@@ -264,8 +264,8 @@ export default function BaseNavigator() {
               const onError =
                 route.params?.onError || (route.params as unknown as typeof route)?.params?.onError;
               // @TODO replace with correct error
-              if (onError && typeof onError === "function")
-                onError(route.params.error || new Error("Request account interrupted by user"));
+              if (onError && typeof onError === "function" && route.params.error)
+                onError(route.params.error);
```
sarneijim, line 268: @Justkant and @hedi-edelbloute, what do you think about this?
```suggestion
              if (onError && typeof onError === "function") {
                onError(route.params.error);
              } else {
                onClose(new Error("Request account interrupted by user"));
              }```

apps/ledger-live-mobile/src/components/Stake/index.tsx
hedi-edelbloute: @sarneijim Can you please list all message that are triggered here other than the one listed in sentry notifications ?

If there is nothing else, it's dead code. Dead code should be removed

sarneijim: If onError(route.params.error) has not been removed, onError callback is not "dead code".

apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
haammar-ledger: `&& route.params.error` seems the more generic approach to me.

haammar-ledger (approved):
Settling the debate :handshake: 
- It is correct that at the moment the filtering on route.params.error likely has no effet as no onError is ever passed as any props anywhere.
- However WalletAPI are mentioning they will go over that error handling mechanism at some point, so it doesn't hurt to merge that low-impact temporary state of the code

apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
Justkant: This change broken the wallet-api

PR #3772, opened 2023-06-21
[LIVE-8061] Bugfix - Prevent operation duplicates with rate limited explorers

lambertkevin:
### 📝 Description

This PR add a `toBlock` parameter to the explorer API of the EVM family. 
Why ? 
Because due to how rate limited some explorers (Etherscan/Blockscout) are, there might be a discrepency between calls concerning how recent the blocks are.

Example: I want to sync a Polygon (2s blocktime) account with Etherscan as an explorer. Etherscan has a 1 call / 5s limit and we need 4 calls to get the full state. So:
- Get the coin operations (latest block 10)
- wait 5 seconds
- Get the erc20 operations (latest block 12)
- wait 5 seconds
- Get the erc721 operations (latest block 15)
- wait 5 seconds
- Get the erc1155 operations (latest block 17)

Which means that when sending an ERC20/ERC721/ERC1155 transaction, you can be too early to get a coin operation and 15 seconds later find an erc1155 event without being able to link it to the coin operation. 

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8061 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-evm/src/__tests__/unit/synchronization.unit.test.ts
+++ b/libs/coin-evm/src/__tests__/unit/synchronization.unit.test.ts
@@ -39,7 +39,7 @@ describe("EVM Family", () => {
         // Mocking getAccount to prevent network calls
         jest.spyOn(rpcAPI, "getBalanceAndBlock").mockImplementation(() =>
           Promise.resolve({
-            blockHeight: 10,
+            blockHeight: 6969,
             balance: new BigNumber(100),
           }),
         );
@@ -161,7 +161,7 @@ describe("EVM Family", () => {
             getAccountShapeParameters,
             {} as any,
           );
-          expect(account.blockHeight).toBe(10);
+          expect(account.blockHeight).toBe(6969);
         });
 
         it("should keep the operations from a sync to another", async () => {
@@ -202,6 +202,7 @@ describe("EVM Family", () => {
             getAccountShapeParameters.address,
             account.id,
             0,
+            6969,
           );
         });
 
@@ -223,6 +224,7 @@ describe("EVM Family", () => {
             getAccountShapeParameters.address,
             account.id,
             coinOperations[2].blockHeight! - synchronization.SAFE_REORG_THRESHOLD,
+            6969,
           );
         });
       });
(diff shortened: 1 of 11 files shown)


libs/coin-evm/src/api/explorer/etherscan.ts
```diff
@@ -69,6 +70,9 @@ export const getLastCoinOperations = async (
   if (fromBlock) {
     url += `&startBlock=${fromBlock}`;
   }
+  if (toBlock) {
```
hedi-edelbloute, line 73: More of a personal pref but i'd personally compare with undefined or null value, if(0) doesn't execute instruction, same for ''

hedi-edelbloute (approved):
lgtm

libs/coin-evm/src/__tests__/unit/synchronization.unit.test.ts
```diff
@@ -523,12 +525,32 @@ describe("EVM Family", () => {
...
           transactionSequenceNumber: 123,
+        };
+
+        expect(
+          await synchronization.getOperationStatus(currency, {
+            ...coinOperations[0],
+            subOperations: [tokenOperations[0]],
+            nftOperations: [erc721Operations[0], erc1155Operations[0]],
+          }),
+        ).toEqual({
```
henrily-ledger, line 541:
```suggestion
        const operationStatus =  await synchronization.getOperationStatus(currency, {
            ...coinOperations[0],
            subOperations: [tokenOperations[0]],
            nftOperations: [erc721Operations[0], erc1155Operations[0]],
          });
        expect(operationStatus).toEqual({
```

not mandatory but I think it's clearer if we don't have an await inside a function

henrily-ledger (approved):

libs/coin-evm/src/api/explorer/etherscan.ts
lambertkevin: In this case I also want to "ignore" 0 actually, because it could break the request completely depending on the implementation:

"get block from 100 to 0" ->  doesn't make sense, we should ignore the "to 0"
"get block from 0 to 100" -> uncessary to add as default behaviour is already "null | undefined = 0" for the "startBlock" option
"get block from 0 to 0" -> this one is tricker because you would expect nothing as a response but will get everything from block 0 to latest instead, but I guess this is really an edge case that don't need to be considered right now as nothing in the coin integration would use it this way without everything being broken before this request :p

libs/coin-evm/src/__tests__/unit/synchronization.unit.test.ts
lambertkevin: Updated 👍

PR #3774, opened 2023-06-21
Bugfix/stacks

lawRathod:
### 📝 Description

Fix bot tests for stacks integration.

### ❓ Context

- **Impacted projects**:  ledger-live-common
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
+++ b/libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
@@ -17,7 +17,7 @@ import {
   fetchFullTxs,
   fetchNonce,
 } from "../../bridge/utils/api";
-import { MempoolTransaction, StacksNetwork, TransactionResponse } from "./api.types";
+import { StacksNetwork, TransactionResponse } from "./api.types";
 import { getCryptoCurrencyById } from "../../../../currencies";
 import { encodeOperationId } from "../../../../operation";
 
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
```diff
@@ -135,22 +149,28 @@ export const getAccountShape: GetAccountShape = async info => {
...
 
-  const result = {
+  const result: Partial<Account> = {
     id: accountId,
     xpub: publicKey,
     freshAddress: address,
     balance,
     spendableBalance,
     operations: flatMap(rawTxs, mapTxToOps(accountId, info)),
+    pendingOperations: flatMap(pendingTxs, mapTxToOps(accountId, info)),
```
haammar-ledger, line 173: Sorry, I don't think it's the good approach to edit `pendingOperations` yourself, because unlike `operations` it has a fully generic lifecycle and that would interfere with it (pending operations are only added after a broadcast, and only removed by a generic mechanism to reconcile them with confirmed operations found during sync)

Do you think you could just call `fetchFullMempoolTxs` during `findNextNonce`? That would shorten the loop between when you need the info and when you get it.

lawRathod: I see what you mean. I didn't know this, but I was thinking if I change between devices the pending txns are lost and now I get it, it's by design. This can be reverted quickly as we now know the problem and we fix it the way you suggest. The `fetchFullMempoolTxs` doesn't take long so can be done in `findNextNonce`.

lawRathod: Fixed @haammar-ledger, please have another look 😄

PR #3778, opened 2023-06-22
feat: LIVE-6280 allow device pairing with open app on LLM (new selector)

juan-cortes:
### 📝 Description

The new device selector (the old one also fails) fails to complete a pairing flow when there's an open application on the device. The reason for this is the use of the `getVersion` apdu instead of the `getAppAndVersion` apdu since one is available only on the dashboard and the other is available virtually everywhere.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6280` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
- With the new device selector feature flag enabled
- Make sure your device is not paired with your phone, delete if from Live if necessary
- Enter BTC/ETH or some other application on your nano
- Attempt to pair the device on LLM

It is expected that the pairing will succeed (if you agree to the code of course). Entering the manager afterwards should automatically quit the app and access the screen.  Doing the same without the feature flag enabled should fail since the logic is relying on `getVersion` which fails if not on the dashboard.

### DIFF
--- a/libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.ts
+++ b/libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.ts
@@ -1,9 +1,8 @@
 import { useState, useEffect } from "react";
 import { from } from "rxjs";
 import { first } from "rxjs/operators";
-import type { FirmwareInfo } from "@ledgerhq/types-live";
 import { withDevice } from "../../hw/deviceAccess";
-import getVersion from "../../hw/getVersion";
+import getAppAndVersion from "../../hw/getAppAndVersion";
 import { BleError } from "../types";
 
 export type useBleDevicePairingArgs = {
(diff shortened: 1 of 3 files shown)


libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.ts
```diff
@@ -31,10 +30,10 @@ export const useBleDevicePairing = ({
   const [pairingError, setPairingError] = useState<PairingError>(null);
 
   useEffect(() => {
-    const requestObservable = withDevice(deviceId)(t => from(getVersion(t))).pipe(first());
+    const requestObservable = withDevice(deviceId)(t => from(getAppAndVersion(t))).pipe(first());
```
alexandremgo, line 33: It's a great idea ! Though i think you need to update `libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.test.ts` so it mocks `getAppAndVersion` now (and not `getVersion`) ;)

juan-cortes: Good catch, pushed.

alexandremgo (approved):
Great fix, thx ! 🐬

juan-cortes:
Not doable for now until we figure out something involving FW
https://ledger.slack.com/archives/C02KVKX8GCF/p1687504327915769

PR #3783, opened 2023-06-22
feat: extend expiration time of tron transaction

drakoFukayu:
Extend expiration time of tron transactions.

### 📝 Description

This build skip onboarding since client do not need it

### ✅ Checklist

- [x] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes** 

### 📸 Demo

same as tron 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/reducers/settings.ts
```diff
@@ -121,7 +121,7 @@ export const getInitialLocale = () => {
 };
 
 const INITIAL_STATE: SettingsState = {
-  hasCompletedOnboarding: false,
+  hasCompletedOnboarding: true,
```
gre, line 124: probably a mistake

libs/ledger-live-common/src/families/tron/api/index.ts
```diff
@@ -140,10 +141,26 @@ export const createTronTransaction = async (
...
+  const VAULT_EXPIRATION_TIME = 600;
+  const HttpProvider = TronWeb.providers.HttpProvider;
+  const fullNode = new HttpProvider(getBaseApiUrl());
+  const solidityNode = new HttpProvider(getBaseApiUrl());
+  const eventServer = new HttpProvider(getBaseApiUrl());
+  const tronWeb = new TronWeb(fullNode, solidityNode, eventServer);
+  // tronWeb.setHeader({
+  //   "TRON-PRO-API-KEY": "ec320f10-8bd8-4ac9-a6cf-eef9bd346e25",
+  // });
+  return tronWeb.transactionBuilder.extendExpiration(preparedTransaction, VAULT_EXPIRATION_TIME);
```
gre, line 161: is this essentially doing a http call?
since the tron implementation generally didn't use "tronweb" i think it would be good to try to do the same without having to depends on tronweb (if it's possible to do it with api calls) so it stays consistent with the coin implementation.

it's on @LedgerHQ/live-blockchain-support  to review / advise further now 🙏

gre (dismissed):

libs/ledger-live-common/src/families/tron/api/index.ts
haammar-ledger: Very good point but we had a look and at the moment there doesn't seem to be a way to do that with the HTTP API, so TronWeb is the most straightforward to do it...

apps/ledger-live-desktop/src/renderer/reducers/settings.ts
haammar-ledger: Maybe relevant to bypass onboarding I assume, which is Nano-specific.
OK for a custom build, but won't likely be mergeable on Live codebase.

YBadiss: Yep that's correct: it's a bypass that we will remove before asking for a merge into LL

YBadiss: Removed 👍

libs/ledger-live-common/src/families/tron/api/index.ts
haammar-ledger: Actually after a deeper look in the tronweb lib, it looks like the only part that's useful to you is [this](https://github.com/tronprotocol/tronweb/blob/4b131bbcaa94f9d70fffd1dc36577703d42eefa1/src/lib/transactionBuilder.js#L2690C1-L2693C66).

@drakoFukayu could you please try using these few lines instead of the whole tronweb lib?

Sorry for asking for more work, the rationale is that the bundle size for Live is already very large, causing long build and loading time, which will be even more of an issue when we want to move to a web environment.
So we are trying to add new dependencies only when they're absolutely necessary.

libs/ledger-live-common/package.json
```diff
@@ -123,6 +123,12 @@
     "@elrondnetwork/erdjs-network-providers": "^1.1.2",
     "@ethereumjs/common": "^2.6.2",
     "@ethereumjs/tx": "^3.5.0",
+    "@ethersproject/bignumber": "^5.7.0",
+    "@ethersproject/bytes": "^5.7.0",
+    "@ethersproject/keccak256": "^5.7.0",
+    "@ethersproject/logger": "^5.7.0",
+    "@ethersproject/properties": "^5.7.0",
+    "@ethersproject/strings": "^5.7.0",
```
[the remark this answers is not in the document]
drakoFukayu: unfortunately it does not build with pnpm. We open a github issue there https://github.com/tronprotocol/tronweb/issues/385
As the issue mention, it should be fix in the next release of tronweb and thus we should be able to remove theses lib

libs/ledger-live-common/src/families/tron/api/index.ts
haammar-ledger: Needs more than the few lines I mentioned, doesn't seem there is an alternative to using tronweb at the moment.

For context: I have contacted trongrid who are willing to add support for changing the expiration through an HTTP call, so we may consider an evolution soon to remove the tronweb lib and just use the HTTP API.

haammar-ledger (approved):

PR #3786, opened 2023-06-23
Feature - Add `setLoadConfig` method to signerContext & support for env var backend URIs

lambertkevin:
### 📝 Description

This PR adds support for the `setLoadConfig` method of `hw-app-eth` in the `signerContext` and adds the usage of the environment variables `DYNAMIC_CAL_BASE_URL` & `NFT_ETH_METADATA_SERVICE`.

### ❓ Context

- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-evm/src/signOperation.ts
+++ b/libs/coin-evm/src/signOperation.ts
@@ -1,18 +1,19 @@
-import { Observable } from "rxjs";
 import {
   Account,
   SignOperationFnSignature,
   SignOperationEvent,
   DeviceId,
 } from "@ledgerhq/types-live";
+import { Observable } from "rxjs";
+import { getEnv } from "@ledgerhq/live-env";
 import { ledgerService } from "@ledgerhq/hw-app-eth";
-import { ResolutionConfig } from "@ledgerhq/hw-app-eth/lib/services/types";
 import { SignerContext } from "@ledgerhq/coin-framework/signer";
+import { LoadConfig, ResolutionConfig } from "@ledgerhq/hw-app-eth/lib/services/types";
 import { buildOptimisticOperation } from "./buildOptimisticOperation";
+import { EvmAddress, EvmSignature, EvmSigner } from "./signer";
 import { prepareForSignOperation } from "./prepareTransaction";
 import { getSerializedTransaction } from "./transaction";
 import { Transaction } from "./types";
-import { EvmAddress, EvmSignature, EvmSigner } from "./signer";
 
 /**
  * Transforms the ECDSA signature paremeter v hexadecimal string received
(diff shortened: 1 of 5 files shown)


libs/coin-evm/src/signer.ts
```diff
@@ -25,7 +25,22 @@ type LedgerEthTransactionResolution = {
   domains: DomainServiceResolution[];
 };
 
+// Duplicate type definition from hw-app-eth.
```
haammar-ledger, line 28: So that's a `FIXME` comment, out of scope for the migration but to address with the signer abstraction topic maybe.

henrily-ledger (approved):

libs/coin-evm/src/signOperation.ts
```diff
@@ -67,21 +68,30 @@ export const buildSignOperation =
           erc20: true,
           domains: transaction.recipientDomain ? [transaction.recipientDomain] : [],
         };
+        const loadConfig: LoadConfig = {
+          cryptoassetsBaseURL: getEnv("DYNAMIC_CAL_BASE_URL"),
+          nftExplorerBaseURL: getEnv("NFT_ETH_METADATA_SERVICE") + "/v1/ethereum",
```
haammar-ledger, line 73: I'm seeing a `if (isNFTActive(account.currency))` condition in the ethereum family, don't we need that here too?

lambertkevin: Nope, providing the URI or not won't change anything as long as the `nft: true` is not added to the `ResolutionConfig` (it will be added properly in the NFT PR right here: https://github.com/LedgerHQ/ledger-live/blob/e1d75bc992b707add781369aa1c93464f675c8c1/libs/coin-evm/src/signOperation.ts#L69)

libs/coin-evm/src/signer.ts
lambertkevin: Yeah I followed what Stephane did above, I think a quick fix could be to move those types in the `types-live` lib if we want to avoid the dups 🤷‍♂️

haammar-ledger: Not quite a Live type... I'd say for now `hw-app-eth` is the legitimate owner of that type?
So either importing the type from there, or a FIXME comment.

lambertkevin: Yes, the legitimate owner is `hw-app-eth` for both `LoadConfig` & `LedgerEthTransactionResolution` above.
I'm not sure creating a separate "types-XXX" for every lib is really the way to scale this, specifically when it's purely for DX so it's a dev dependency anyway.
A good old "@ledgerhq/types` would be fine to me, with proper folders to represent our libs and every "common" things to share can be there 🤷‍♂️ 
So yeah, in the end, I'm either it's a FIXME, either I add a new "types-hw-app-eth", or it's in `types-live`

haammar-ledger: I meant just importing them from `hw-app-eth`, as it is now?
Neither `@ledgerhq/types` nor `types-hw-app-eth`, just keeping it simple but avoiding types duplication.

lambertkevin: The whole signer abstraction makes it kinda antipattern to import 'hw-app-eth' unfortunately :(

haammar-ledger (approved):

PR #3790, opened 2023-06-23
feat: LIVE-6419 Improvements to pairing flow on Stax

juan-cortes:
### 📝 Description

The Ble transport has a forced disconnect / reconnect process that is in place to address a bug in the BLE firmware for the LNX device. Since this doesn't impact Stax, there's no need to introduce that delay in the first place. This optimization should mean that first connections and pairing flows should be faster on Stax, around 4-5 seconds.

### ❓ Context

- **Impacted projects**: `@ledgerhq/react-native-hw-transport-ble` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-6419` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
- With an unpaired Stax device launch LLM.
- Make sure it's not listed in the known devices, if so, delete it first.
- Start a pairing flow and time the delay between accepting on Stax and seeing the success screen on LLM.
- It **should** be around 4-5 seconds less than on a develop build.

The same scenario when tested on a LNX should be slower, since it has the delay mentioned.


libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
```diff
@@ -317,15 +318,18 @@ async function open(deviceOrId: Device | string, needsReconnect: boolean) {
     const afterMTUTime = Date.now();
 
     if (reconnectionConfig) {
-      // workaround for #279: we need to open() again if we come the first time here,
-      // to make sure we do a disconnect() after the first pairing time
-      // because of a firmware bug
+      // Refer to ledgerjs archived repo issue #279
+      // All HW .v1 LNX have a bug that prevents us from communicating with the device right after pairing.
+      // When we connect for the first time we issue a disconnect and reconnect, this guarantees that we are
+      // in a good state. This is avoidable in some key scenarios ↓
```
alexandremgo, line 324: Thanks for the documentation 🙏

alexandremgo (approved):

PR #3796, opened 2023-06-23
test(lld): first iteration of testing deeplink

gre:
### 📝 Description

proposal to cover deeplinks in e2e tests. also based on some work from @Justkant .

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/B2CQA-1721 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
```diff
@@ -258,7 +258,7 @@ export function useDeepLinkHandler() {
           break;
         }
         case "discover":
-          if (path.startsWith("protect")) {
+          if (path?.startsWith("protect")) {
```
gre, line 261: apparently we have this real bug in `develop` btw, `ledgerlive://discover` will make a crash at boot with "startsWith" error, after i fix it the test pass, it was a great way for me to test but i think there is a bug on develop🥵 

cc @Justkant @ggilchrist-ledger

Justkant: Nice catch, I think typescript should catch that no ?

Justkant (approved):

PR #3804, opened 2023-06-26
Fix: LIVE-7873 Fix Eth tx order in the account page

hzheng-ledger:
### 📝 Description

Fix pending transaction order problem. 
Explanation: 
The transactions history in eth account are sorted by date. When a eth pending transaction is confirmed or user speed up a transaction, the transaction date is overwritten. So the order of transactions history is changed. In order to fix the problem, we should keep the initial date to keep the order.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-7873

### ✅ Checklist

- [x] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledger-live-common/src/families/ethereum/synchronisation.ts
```diff
@@ -77,6 +77,27 @@ export const getAccountShape: GetAccountShape = async (infoInput, { blacklistedT
 
   // transform transactions into operations
   let newOps = flatMap(txs, txToOps({ address, id: accountId, currency }));
+
+  // keep the date of the pending operations that appears in newOps
+  const pendingOperationsDate = {};
+
+  (initialAccount?.pendingOperations || []).concat(initialAccount?.operations || []).forEach(op => {
```
hedi-edelbloute, line 84: Wouldn't just a global if initialAccount do the same as those `|| []` ?

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
```diff
@@ -77,6 +77,27 @@ export const getAccountShape: GetAccountShape = async (infoInput, { blacklistedT
 
   // transform transactions into operations
   let newOps = flatMap(txs, txToOps({ address, id: accountId, currency }));
+
+  // keep the date of the pending operations that appears in newOps
+  const pendingOperationsDate = {};
```
hedi-edelbloute, line 82: I don't understand why we need to do this :(

The pendingOperationsDate map is reset on each sync, right ?

hedi-edelbloute (commented):
We need to add unit tests !

haammar-ledger (changes requested):
Thanks for the fix.
- Please add unit tests covering at least the initial problem
- Please clean up the PR's title so a little better readable :)

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
hzheng-ledger: yes, good point. I've done a little refactoring now

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
hzheng-ledger: yes, the pendingOperationsDate map is reset on each sync. (I rename the "pendingOperationsDate" variable now)
I don't see any better solution. To scan all the operations in initialAccount, it is not very time consuming. 
In fact, we have already another place to scan all operations on each sync. "mergeOps".
I wanted to reuse it but it seems to be very difficult.

libs/ledger-live-common/src/families/ethereum/signOperation.ts
```diff
@@ -167,7 +170,7 @@ export const signOperation = ({
                 senders,
                 recipients,
                 accountId,
-                date: new Date(),
+                date: editingOp ? editingOp.date : new Date(),
```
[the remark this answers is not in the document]
hzheng-ledger: Good point. 
but as this fix is related to edit eth transaction and the edit eth transaction feature is only available with ethereum family so far.
in EVM family, the transactions are not supposed to be stuck for a long time (only eth transactions can be stuck for longtime) so the tx order problem rarely occurs today.
we will migrate the edit eth transaction feature along with this fix to EVM family of course.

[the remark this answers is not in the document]
hedi-edelbloute: Nice catch, is EDIT eth transaction available on evm family ?

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
```diff
@@ -677,3 +682,31 @@ function reconciliateSubAccounts(
 
   return subAccounts;
 }
+
+// keep the initial dates of the operations to avoid the dates are overwrited by the ones from the blockchain
+export function preserveInitialOperationsDate(
```
hedi-edelbloute, line 687: We could unit test this method

haammar-ledger: That's exactly what the test added in `synchronisation.test.ts` does, right?

libs/ledger-live-common/src/families/ethereum/synchronisation.test.ts
```diff
@@ -48,3 +54,74 @@ describe("fetchAllTransactions", () => {
     expect(txs.length).toEqual(3);
   });
 });
+
+describe("preserveInitialOperationsDate", () => {
+  it("should preserve initial operations date in newOps", () => {
```
haammar-ledger, line 59: @hzheng-ledger Don't we need to cover several different scenarios? (different date orders, etc...)

libs/ledger-live-common/src/families/ethereum/synchronisation.test.ts
haammar-ledger, line 1: Can you please rename this file to `synchronisation.unit.test.ts`

libs/ledger-live-common/src/families/ethereum/synchronisation.ts
hzheng-ledger: yes

hedi-edelbloute: After discussing it together, we might need to split it in multiple test (one for each call) to cover every use case

cc @haammar-ledger @hzheng-ledger

hedi-edelbloute (approved):

libs/ledger-live-common/src/families/ethereum/synchronisation.test.ts
hzheng-ledger: fixed

libs/ledger-live-common/src/families/ethereum/synchronisation.test.ts
hzheng-ledger: fixed

PR #3812, opened 2023-06-26
feat: add scroll to top button to nft gallery

mderrick:
### 📝 Description

Scroll to top button to the NFT gallery page

### ❓ Context

- **Linked resource(s)**: `LIVE-7815` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/1044686/b8fd969c-9d2c-4c37-99eb-1c73e51491ad

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/Nft/NftGallery/ScrollToTopButton.tsx
+++ b/apps/ledger-live-mobile/src/components/Nft/NftGallery/ScrollToTopButton.tsx
@@ -1,0 +1,86 @@
+import React, { FC, useContext } from "react";
+import { Animated, FlatList, StyleSheet } from "react-native";
+import { WalletTabNavigatorScrollContext } from "../../WalletTab/WalletTabNavigatorScrollManager";
+import { useTranslation } from "react-i18next";
+import { Button } from "@ledgerhq/native-ui";
+import { useSafeAreaInsets } from "react-native-safe-area-context";
+
+interface Props {}
+
+const ScrollToTopButton: FC<Props> = () => {
+  const { t } = useTranslation();
+  const { scrollY, headerHeight, scrollableRefArray } = useContext(WalletTabNavigatorScrollContext);
+
+  const translateY = scrollY.interpolate({
+    inputRange: [800, 850],
+    outputRange: [-40, 10],
+    extrapolate: "clamp",
+  });
+  const opacity = scrollY.interpolate({
+    inputRange: [800, 850],
+    outputRange: [0, 1],
+    extrapolate: "clamp",
+  });
+
+  const goToTop = () => {
+    if (scrollableRefArray.current) {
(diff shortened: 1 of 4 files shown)


apps/ledger-live-mobile/src/components/Nft/NftGallery/ScrollToTopButton.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { FC, useContext } from "react";
+import { Animated, FlatList, StyleSheet } from "react-native";
+import { WalletTabNavigatorScrollContext } from "../../WalletTab/WalletTabNavigatorScrollManager";
+import { useTranslation } from "react-i18next";
+import { Button } from "@ledgerhq/native-ui";
+import { useSafeAreaInsets } from "react-native-safe-area-context";
+
+const ScrollToTopButton: FC = () => {
```
juan-cortes, line 8: Unless I'm missing something it seems like you are reinventing the wheel here, we have `scrollToTop` util which should take care of all the shenanigans of scrolling, I think you can just use that on your button instead of rewriting it.

juan-cortes (changes requested):

apps/ledger-live-mobile/src/components/Nft/NftGallery/ScrollToTopButton.tsx
mderrick: I had a look at this and it seems that it works if your view is wrapped in `NavigationScrollView` but the wallet/gallery isn't, so not sure i can use that util.

My understanding of how the wallet view works so far is that the NFT and Crypto tab share scroll behaviour in an attempt to keep them in sync (I assume for the tab and header animation) all driven by this context. I think I have to scroll this view manually using the ref.

juan-cortes: I'd ask for someone from live-hub to review this then 🙏🏼 but you have my ✔️

mcayuelas-ledger: I'm not expert with scrolls behaviour.
But yes Crypto Tab and NFTs tab are in sync when scrolling down.

So for the button i'm ok, i don't know if we have done already something similar. 

My concern is just about the sync  when scrolling up? what happen to the crypto tab? Is it in sync? :) If yes so LGTM !

mderrick: I've confirmed that the header animation is kept in sync when using this scroll to top button and navigating back tot he "Crypto" tab.

PR #3814, opened 2023-06-26
feat: add earn enabled to tracking

mitchellv-ledger:
### 📝 Description

Add earn dashboard feature flag as an extra property to track users who have it enabled that complete the staking flow.

### ❓ Context

- **Impacted projects**: `ledger-live, ledger-mobile` 
- **Linked resource(s)**: `LIVE-6906` 

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
-  [X] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/analytics/segment.ts
+++ b/apps/ledger-live-desktop/src/renderer/analytics/segment.ts
@@ -26,6 +26,7 @@ import createStore from "../createStore";
 import { currentRouteNameRef, previousRouteNameRef } from "./screenRefs";
 import { useCallback, useContext } from "react";
 import { analyticsDrawerContext } from "../drawers/Provider";
+import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
 invariant(typeof window !== "undefined", "analytics/segment must be called on renderer thread");
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const os = require("os");
@@ -84,6 +85,7 @@ const extraProperties = (store: ReduxStore) => {
     : [];
   const hasGenesisPass = hasNftInAccounts(GENESIS_PASS_COLLECTION_CONTRACT, accounts);
   const hasInfinityPass = hasNftInAccounts(INFINITY_PASS_COLLECTION_CONTRACT, accounts);
+
   return {
     appVersion: __APP_VERSION__,
     language,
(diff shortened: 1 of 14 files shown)


apps/ledger-live-desktop/src/renderer/analytics/segment.ts
```diff
@@ -233,12 +237,13 @@ export function useTrack() {
         eventName,
         {
           ...(analyticsDrawerName ? { drawer: analyticsDrawerName } : {}),
+          ptxEarnEnabled: !!ptxEarnFeatureFlag?.enabled,
```
sarneijim, line 240: Not sure about this solution.
This method is common across all the app, ptxEarnEnabled  is very specific, I do not think we must include this info in tracks which are not relative to "swap" flow.

Options that I have in mind:
1. Create useSharedStakeTracking method and pass FF value each time that you use it:
`useSharedStakeTracking(ptxEarnEnabled) = {flow: "stake", ptxEarnEnabled }`
`track(...,{..., useSharedStakeTracking(ptxEarnEnabled)}`

2. Create sharedStakeTracking object and keep FF in memory when you call it to the menu:
`sharedStakeTracking = {flow: "stake", ptxEarnEnabled: window.localStorage.getItem("ptxEarnEnabled"); }

mitchellv-ledger: I went for this solution as it's okay to add extra details that aren't specifically relevant to other flows, eg. tracking adds `hasGenesisPass` to all tracking.  I'll check with Martin to be sure as otherwise we'd end up in a situation where we're trying to add it everywhere for all different scenarios which would be more complex imo than just adding it here

mitchellv-ledger: Also, checking the usage for the useTrack hook it's only used in a couple of other places besides where I've added it so actually this would suit us well.

sarneijim: Which track not relative to `flow: "stake"` do you want to track? The person responsable of track is Amine BELGHAZI

sarneijim: Do we must change any "track" to "useTrack"? if not, you wont have the property always.

PR #3829, opened 2023-06-27
Support/react dev tools

KVNLS:
### 📝 Description

Today we can't load React Developer Tools from LLD because the latest version uses Manifest V3 and Electron doesn't support it yet. https://github.com/electron/electron/issues/36545
Developers need to have access to this tool to debug the application (performance / re-rendering issues).
While Electron doesn't support Manifest V3 extension the only solution that I see is to provide a version of the extension prior to V3 migration in the repository (built from React sources) and to load the extension in LLD from it instead of the one installed in Chrome.
Everything under react-dev-tools comes from the build from react repo.

### ❓ Context

- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo
Inspect Element => You need to have two tabs "Components" & "Profiler"
image

### 🚀 Expectations to reach


gre:
i won't have enough time to fully review this, but are we sure that committing the whole plugin in our source code is the only way? like it isn't yet a library?

KVNLS:
> this week i won't have enough time to fully review this, but are we sure that committing the whole plugin in our source code is the only way? like it isn't yet a library?

@gre Another solution would be to use this library: https://www.npmjs.com/package/react-devtools but it will open devtools in a standalone and the DX will not be great.
It raises other questions such as: 
- How will you be able to reopen the devtools if you close the standalone
- Do we want to open the standalone programmatically or through the console ? 
- And so on...

I understand your concern but IMO the solution to embed the result of the build directly here is the costless one without degrading DX + it's a temporary solution until Electron fix manifest V3 support.
Last option is to fork the project, build the extension, expose a npm package and add a dependency on it but as it's a temporary solution I don't think that it worth the cost.
cc @valpinkman, what do you think ?

valpinkman:
@KVNLS I think what you propose is a pretty elegant temporary solution until Electron supports V3 manifests. One thing I could think of, is maybe we could move the react-devtool source files to `tools` folder. 
We are looking with @elbywan if we can remove this code from the repo, or at least move it outside of LLD. 

Thanks for the great fix !

valpinkman:
@KVNLS I pushed a proposal on the branch directly. Tell me if you are OK with the changes. Also, the PR needs to be rebased and I don't want to overwrite your branch myself.

KVNLS:
> @KVNLS I pushed a proposal on the branch directly. Tell me if you are OK with the changes. Also, the PR needs to be rebased and I don't want to overwrite your branch myself.

Thanks a lot @valpinkman looks great and work like a charm!
I rebased the branch

valpinkman (approved):

PR #3830, opened 2023-06-27
[support] Fix reference stability tests in bridge integration tests

haammar-ledger:
### 📝 Description

Account bridge functions `updateTransaction` and `prepareTransaction` must respect a ~~obscure~~ little-known rule (documented in the [wiki](https://github.com/LedgerHQ/ledger-live/wiki/LLC:AccountBridge#preparetransaction)), that the reference of their input transaction must _not_ change if they bring no change to the data.

This rule is strongly coupled to React, which will basically re-render everytime the ref of the transaction changes. So changing the ref when the transaction's content has not changed is bad for perf.

Ideally we'd like to remove that dependency to React from our design, but for now it makes sense to enforce the rule. Previous attempts were done using the bridge integration tests, but:
- It only covered `prepareTransaction`, not `updateTransaction`
- The test was wrong and didn't actually check for ref stability, it only checked that the data was the same

### ❓ Context

- **Impacted projects**:
  - `coin-framework`
  - `coin-algorand`
  - `coin-evm`
  - `coin-polkadot`
  - `ledger-live-common`

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8237

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

- All ref stability tests on `updateTransaction` and `prepareTransaction` pass for all families
- Hopefully the rest of the tests aren't any more broken than they are now


gre:
> must respect a an obscure, little-known rule

yet it was documented and it's one of the most important property of prepareTransaction in the original design (section "Reference stability") https://github.com/LedgerHQ/ledger-live/wiki/LLC:AccountBridge#preparetransaction



> The test was wrong and didn't actually check for ref stability, it only checked that the data was the same

indeed, my bad on this (as original writer of this test), the confusion came from `expect(t2).toStrictEqual(t3);` , toStrictEqual is actually not a strict equal `===` 😭  TiL

can you double check that, instead of,

```
            expect(Object.is(t2, t3)).toBe(true);
```

that you could do 

```
            expect(t2).toBe(t3);
```

if i'm correct, toBe is actually that `===` we are looking for. and the benefit is that if it's not you would see the object diff in the console. (to be double/triple check)

**Cheers for fixing this,**
gre

libs/ledger-live-common/src/families/cosmos/js-updateTransaction.ts
```diff
@@ -1,15 +1,18 @@
...
+    ("mode" in patch && patch.mode !== tx.mode) ||
+    ("validators" in patch && patch.validators?.length !== tx.validators.length)
+  ) {
+    updatedTx.gas = null;
+    updatedTx.fees = null;
   }
 
-  return { ...t, ...patch };
+  return updatedTx;
 };
```
lambertkevin, line 16: I think this can lead to unexpected mutation of the inital `tx`. Maybe we should rather create a new `patch` instead and use a spread still have `defaultUpdateTransaction` as the last instruction with the 2 patches ?

libs/ledger-live-common/src/families/near/js-transaction.ts
```diff
@@ -30,5 +25,6 @@ export const prepareTransaction = async (a: Account, t: Transaction): Promise<Tr
       })
     : t.amount;
 
-  return { ...t, fees, amount };
+  const newTx = { ...t, fees, amount };
+  return isEqual(t, newTx) ? t : newTx;
```
lambertkevin, line 29: Is there a reason to not use the `defaultUpdateTransaction` here with `{ fees, amount }` as patch ?

libs/ledger-live-common/src/families/ethereum/updateTransaction.ts
```diff
@@ -1,12 +1,16 @@
...
+import { defaultUpdateTransaction } from "@ledgerhq/coin-framework/bridge/jsHelpers";
 import { Transaction } from "./types";
 
 export const updateTransaction: AccountBridge<Transaction>["updateTransaction"] = (tx, patch) => {
+  const updatedTx = defaultUpdateTransaction(tx, patch);
+
   if ("recipient" in patch && patch.recipient !== tx.recipient) {
-    return { ...tx, ...patch, userGasLimit: null, estimatedGasLimit: null };
+    updatedTx.userGasLimit = null;
+    updatedTx.estimatedGasLimit = null;
```
lambertkevin, line 10: Same thing here, I think this can lead to unexpected mutations of `tx`

libs/ledger-live-common/src/families/filecoin/bridge.integration.test.ts
```diff
@@ -43,7 +42,8 @@ const filecoin: CurrenciesData<Transaction> = {
   ],
   accounts: [
     {
-      FIXME_tests: ["balance is sum of ops"],
+      // TODO works or not?
+      //FIXME_tests: ["balance is sum of ops"],
```
lambertkevin, line 46: Are we good reactivating this test ?

libs/ledger-live-common/src/families/stellar/js-prepareTransaction.ts
```diff
@@ -15,7 +16,8 @@ const prepareTransaction = async (a: Account, t: Transaction): Promise<Transacti
     return { ...t, networkInfo, fees, baseReserve, assetCode, assetIssuer };
   }
 
-  return { ...t, assetCode, assetIssuer };
+  const newTx = { ...t, assetCode, assetIssuer };
+  return isEqual(t, newTx) ? t : newTx;
```
lambertkevin, line 20: Same here, shouldn't we use `defaultUpdateTransaction`?

lambertkevin (commented):
Just a few comments regarding harmonization of the solution + some questions regarding some potential mutations

libs/ledger-live-common/src/families/solana/js-prepareTransaction.ts
```diff
@@ -105,15 +106,14 @@ const prepareTransaction = async (
...
 
-  const preparedTx: Transaction = {
-    ...tx,
+  const patch: Partial<Transaction> = {
     model,
   };
 
   // LLM requires this field to be truthy to show fees
-  (preparedTx as any).networkInfo = true;
+  (patch as any).networkInfo = true;
```
lvndry, line 114: Why not adding a optional `networkInfo` field inside `Transaction` instead ?

libs/ledger-live-common/src/families/cosmos/js-updateTransaction.ts
haammar-ledger: You're right there an issue here, but I think it's actually the opposite :smile: 

What matters is that _reference_ doesn't change when data doesn't change, which is actually different from _immutability_.

```
const a = { foo: 1 };
const b = a;

a.foo = 2; // mutate
console.log(a === b); // true

const c = { ...a, foo: 3 }; // change ref
console.log(a === c); // false
```

However the issue currently, is that the code in the `if` can actually mutate the tx _without_ changing its ref, which is also not what we want.

libs/ledger-live-common/src/families/filecoin/bridge.integration.test.ts
haammar-ledger: Apparently we are \o/

libs/ledger-live-common/src/__tests__/test-helpers/bridge.ts
```diff
@@ -175,15 +175,15 @@ export function testBridge<T extends TransactionCommon>(data: DatasetTest<T>): v
       });
 
       if (scanAccounts) {
-        if (FIXME_ignoreOperationFields) {
+        if (FIXME_ignoreOperationFields && FIXME_ignoreOperationFields.length) {
```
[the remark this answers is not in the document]
haammar-ledger: All the `FIXME_` prefixed stuff in the bridge integration tests are used to disable some of the default tests for a given coin. The idea of the "fixme" prefix is that if some tests are disabled it's likely not normal and should be fixed at some point, but this is not always true actually, as some tests just can't pass with some blockchains...

For instance the "balance is sum of ops" test can't possibly pass for some blockchain where the balance includes some on-chain data that is not part of the operations, e.g pending rewards.

The FIXME_ prefix is a bit confusing indeed, we should likely remove it and just add FIXME comments next to disabled tests in each coin's dataset.

libs/ledger-live-common/src/families/solana/js-prepareTransaction.ts
haammar-ledger: Because there's actually no need for that field in the transaction, except to make the UI happy.
In the end this is a hack, I don't know why we let it pass the review back then, but I feel it's better that it's obvious than to hide it deeper by artificially putting it in the Transaction type.

libs/ledger-live-common/src/families/cosmos/js-updateTransaction.ts
lambertkevin: Yup that's what I meant, in the case where `defaultUpdateTransaction` sees no changes from the patch, it will return `tx` (and its ref) and that means that you're going to update the content of `tx` (now `updatedTx`) without changing the ref 💥 🧠

haammar-ledger: Haha ok didn't get that's what you meant.
This should be fixes in the last commits.

lambertkevin (approved):
🚀

PR #3844, opened 2023-06-28
bugfix/LIVE-8196 Firmware update regression in LLD for old devices

juan-cortes:
### 📝 Description

Longer explanation in the linked ticket.
We broke the automatic firmware update process for users on old versions, but we have found a solution. Instead of asking them to download an old version which was the temporary solution, we have this PR. The proposed fix, which opens a drawer instead of a modal, solves the original issue.

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-8196` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo
No demo

### 🚀 Expectations to reach
A nano S on 1.4.2 should be able to update its firmware on LLD without complications.

### DIFF
--- a/.changeset/real-adults-double.md
+++ b/.changeset/real-adults-double.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": minor
+---
+
+Fixed regression in firmware update for old devices on LLD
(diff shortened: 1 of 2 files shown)


grsoares21 (approved):

alexandremgo (approved):
Thanks for the fix !

PR #3859, opened 2023-06-29
Fix: LIVE-7419 OpReturn fix + bitcoin bot fix

hzheng-ledger:
### 📝 Description

1. Fix Bitcoin Op Return output
2. Fix Bitcoin Bot
3. Increase dust threshold for some coins in bitcoin family

### ❓ Context

- **Linked resource(s)**: [`https://ledgerhq.atlassian.net/browse/LIVE-7419`](https://ledgerhq.atlassian.net/browse/LIVE-7419) 

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach


lvndry (approved):

hedi-edelbloute (changes requested):
We need to add unit tests on libs/ledger-live-common/src/families/bitcoin/logic.ts or explain why we don't

Changeset is missing

Code looks fine and commented

libs/ledger-live-common/src/families/bitcoin/logic.ts
```diff
@@ -216,7 +216,8 @@ export const mapTxToOperations = (
       : tx.outputs.map(o => o.output_index).reduce((p, c) => (p > c ? p : c));
 
   for (const output of tx.outputs) {
-    if (output.address) {
+    // ledger explorer returns "unknown" as recipient for OP_RETURN outputs, we don't want to display it in our UI
+    if (output.address && !output.address.includes("unknown")) {
```
haammar-ledger, line 220: Is this the only way to identify an OP_RETURN output?
Best way would be from blockchain data instead of some specific response from Ledger explorer, but otherwise it's fine.

libs/ledger-live-common/src/families/bitcoin/specs.ts
```diff
@@ -446,6 +442,7 @@ const pivx: AppSpec<Transaction> = {
   mutations: bitcoinLikeMutations(),
   minViableAmount: genericMinimalAmount,
 };
+const qtumViacoin = parseCurrencyUnit(getCryptoCurrencyById("qtum").units[0], "0.001");
```
haammar-ledger, line 445: Do you mean `minQtum` instead of `qtumViacoin`?

haammar-ledger (commented):
@hedi-edelbloute In this case, the bot was seeing the issue so technically there is already coverage.
This is the third acceptable case actually: adding new test coverage OR no coverage possible with justification OR **coverage was already there** (typically when fixing tests, just like this PR)

hedi-edelbloute:
> @hedi-edelbloute In this case, the bot was seeing the issue so technically there is already coverage. This is the third acceptable case actually: adding new test coverage OR no coverage possible with justification OR **coverage was already there** (typically when fixing tests, just like this PR)

I thought we needed unit tests on every PR and changed code

We could take this as an opportunity to add coverage on the mapTxToOperations method here, especially on the part that filters out outputs that either have no address, or an address that includes "unknown"

describe(mapTxToOperations) {
it("should filter out outputs that have no address")
it("should filter out outputs that have an address that includes 'unknown')
}

Those tests description are to adapt if needed, i don't have the whole knowledge of what it does

Shouldn't be too long IMO, but as you guys wish :)

hedi-edelbloute (approved):
LGTM

Let's not forget changeset before merging :+1:

libs/ledger-live-common/src/families/bitcoin/logic.ts
hzheng-ledger: We can also check the "script_hex" field, but it is more complicated. Also, the OP_RETURN output is not the only "unknown" address output, (bitcoin spec has some more complicated output with "unknown" address) 
but we should filter out all those output as we can't support such operations in LL so far. (i.e. we can broadcast a transaction with OP_RETURN output but we can't display such outputs without address in LLD/LLM )

haammar-ledger (approved):

PR #3892, opened 2023-07-03
fix: LIVE-8236 Missing pending transaction

hzheng-ledger:
### 📝 Description

When nonce = 0, pending transaction is filtered out by mistake because of (op.transactionSequenceNumber is false when transactionSequenceNumber = 0)

### ❓ Context

- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** (I didn't add unit test for this PR as it is a bug in prod and the fix is obvious. We are going to migrate the eth family to evm family and this piece of code will not be used any more in the near future. We have a different logic to filter pending transactions )

- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/violet-kids-arrive.md
+++ b/.changeset/violet-kids-arrive.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+fix missing pending transaction bug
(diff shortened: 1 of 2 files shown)


haammar-ledger (approved):
I've started a couple bots on the PR let's run them a few times.

haammar-ledger (changes requested):
@hzheng-ledger See bot reports, all failing with
```
⚠️ TypeError: Cannot add property nonce, object is not extensible
```

PR #3894, opened 2023-07-04
Smoke Test + CI rework

nabil-brn:
### 📝 Description

This PR brings a few things;

First of all, we've tagged the playwright specs to have a lighter test suites in PR

On CI, we've changed the way e2e tests are run.
Now:
- Smoke tests are run on each PR
- Full tests are run on push on `main`|`develop`|`hotfix`|`release`
- Possible to run full test manually on a PR
- Fails on full suite will be reported to slack

### ❓ Context

- **Impacted projects**: `Tests, CI Workflow`
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7748

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

N/A

### 🚀 Expectations to reach

When PR is in draft, only a subset of tests should be executed, if PR is open then we run all the tests.


ggilchrist-ledger (approved):
LGTM - nice work

ggilchrist-ledger:
Just to add something we should keep and eye on: let's make sure that the feedback loop doesn't become too long or frustrating - there's a danger we merge things that break non-linux, non-critical tests that then need fixed at a later date (meaning context switching for devs and lost time when we could have fixed something before a PR merge).

apps/ledger-live-desktop/package.json
```diff
@@ -38,7 +38,9 @@
     "test:jest": "jest src --silent --ci --coverage --json --testLocationInResults --outputFile=report.json",
     "test:playwright:setup": "playwright install chromium",
     "test:playwright": "playwright test --config=tests/playwright.config.ts",
```
alexandremgo, line 40: Just a question: if someone runs only `pnpm test:playwright` locally, which behavior is it going to use by default ? It's going to test all playwright tests ?

valpinkman: I think i tried it myself and it takes the first config described in the project key in thep laywright config file

alexandremgo: Ok thx for the info ! :)

alexandremgo (approved):

valpinkman (approved):

haammar-ledger (approved):

PR #3909, opened 2023-07-04
LLD onboarding early security checks

ofreyssinet-ledger:
### 📝 Description

The genuine check and firmware update check are now happening before the sync onboarding (before we ask the user to seed their device).

For the firmware update part, it can be easily tested with the IL (infinite loop) firmware. However since the Early Security Checks don’t let the user move forward until they have the most up to date firmware, it will result in an infinite loop (because IL firmware…).
So implemented in there is a hack that pretends that (in the ESC only) the upgrade path il2 to il0 does not  exist.

Therefore:

- you can test the ESC with a single firmware upgrade by using il0 and [provider 12](https://ledger.slack.com/archives/C0389M739ML/p1687782285481249).
- you can test the ESC with no fw upgrade by using il2 and provider 12.
- you can test the ESC with several fw upgrades in a row by using il0 and provider 10 or 13, but it will never complete (always a new update available)
- you can test the ESC with a “entity not found” bug with 1.2.1 and an odd provider (e.g 4)

### ❓ Context

- **Impacted projects**: `lld` 
- **Linked resource(s)**: [LIVE-8264] [LIVE-8390] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

- firmware up to date (thanks Benoît for the video):

https://github.com/LedgerHQ/ledger-live/assets/91890529/17583a5c-5014-410b-9d37-b24db35d21bd

- firmware update available:

https://github.com/LedgerHQ/ledger-live/assets/91890529/e533d471-9bfb-4561-9ced-9f3e0a073731

### 🚀 Expectations to reach

[LIVE-8264]: https://ledgerhq.atlassian.net/browse/LIVE-8264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-8390]: https://ledgerhq.atlassian.net/browse/LIVE-8390


apps/ledger-live-desktop/src/renderer/components/Onboarding/index.tsx
```diff
@@ -106,15 +106,13 @@ export function Onboarding() {
...
                 path={`${path}/select-use-case`}
-                render={props =>
-                  setUseCase && (
-                    <SelectUseCase
-                      {...props}
-                      setOpenedPedagogyModal={setOpenedPedagogyModal}
-                      setUseCase={setUseCase}
-                    />
-                  )
-                }
```
ofreyssinet-ledger, line 117: setUseCase is always truthy

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+    resetGenuineCheckState,
+  } = useGenuineCheck({
+    getGenuineCheckFromDeviceId,
+    isHookEnabled: genuineCheckActive,
+    deviceId,
+  });
+
+  const {
+    state: { deviceInfo, firmwareUpdateContext: latestFirmware, status, lockedDevice },
+  } = useGetLatestAvailableFirmware({
```
alexandremgo, line 94: You can also retrieve an `error` object from this hook: it can be good to set `firmwareUpdateStatus` to `failed` if an error occurred, and the user has a button to continue

There is one thing to pay attention to: some error are retried by the inner device action, and so the action has not yet failed completely
You can check for this with the following:

```TS
      // If the current error triggered a retry attempt, does not display failure
      if (
        getLatestAvailableFirmwareError &&
        !(
          getLatestAvailableFirmwareError.type === "SharedError" &&
          getLatestAvailableFirmwareError.retrying
        )
      ) {
        log(
          "EarlySecurityCheck",
          "Failed to retrieve latest firmware version with error:",
          getLatestAvailableFirmwareError.name,
        );

        // Or equivalent
        setFirmwareUpdateCheckStatus("failed");
      }
```

If the error has a `type` of `SharedError`, then you can check if it has triggered a retry with the `retrying` property ;)

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/index.tsx
```diff
@@ -44,11 +44,17 @@ export type StepProps = {
   nonce: number;
   setNonce: (arg0: number) => void;
   setFirmwareUpdateCompleted: (arg0: boolean) => void;
+
+  finalStepSuccessDescription?: string;
+  finalStepSuccessButtonLabel?: string;
+  finalStepSuccessButtonOnClick?: () => void;
+  shouldReloadManagerOnCloseIfUpdateRefused?: boolean;
+  deviceHasPin?: boolean;
```
alexandremgo, line 52: Good idea !

apps/ledger-live-desktop/src/renderer/screens/manager/FirmwareUpdate/index.tsx
```diff
@@ -88,35 +91,33 @@ const FirmwareUpdate = (props: Props) => {
...
+      stepId: stepId,
+      installed: installed,
+      firmware: firmware,
+      deviceInfo: deviceInfo,
+      device: device,
+      error: error,
+      deviceModelId: deviceSpecs.id,
+      setFirmwareUpdateOpened,
+      setFirmwareUpdateCompleted,
+      shouldReloadManagerOnCloseIfUpdateRefused: true,
```
alexandremgo, line 114: Make sure QA will test this scenario with a normal fw update from the manager to be safe

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+      setAvailableFirmwareVersion("");
+    } else if (status === "available-firmware") {
+      setAvailableFirmwareVersion(latestFirmware?.final.name || "");
+      setFirmwareUpdateStatus(SoftwareCheckStatus.updateAvailable);
+    }
+  }, [firmwareUpdateStatus, status, genuineCheckStatus, latestFirmware, deviceInfo?.version]);
+
+  const lockedDeviceModalIsOpen =
+    (devicePermissionState === "unlock-needed" && genuineCheckActive) ||
+    (lockedDevice && firmwareUpdateStatus === SoftwareCheckStatus.active);
```
alexandremgo, line 207: You should && another condition: 

```TS
        // The device has no PIN in the ESC - and having an unresponsive device here probably means the user
        // cancelled a firmware update previously and came back to the ESC (with a device in an incorrect state).
        // -> Not displaying a drawer in the case of unresponsive, and letting the device coming back to a normal state.
        (lockedDevice && firmwareUpdateStatus === SoftwareCheckStatus.active && getLatestAvailableFirmwareError?.name !== "UnresponsiveDeviceError") {
```

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/GenuineCheckErrorDrawer.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { setDrawer } from "~/renderer/drawers/Provider";
+
+export type Props = {
+  error: Error;
+  onClickRetry: () => void;
+};
+
+const NotFoundEntityError = createCustomErrorClass("NotFoundEntityError");
+
+const GenuineCheckErrorDrawer: React.FC<Props> = ({ error, onClickRetry }) => {
```
alexandremgo, line 20: Nice i should add one on the LLM ESC 🦥

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/index.tsx
```diff
@@ -127,7 +142,7 @@ const SyncOnboardingScreen: React.FC<SyncOnboardingScreenProps> = ({
   useEffect(() => {
     let timeout: ReturnType<typeof setTimeout>;
 
-    if (allowedError) {
+    if (allowedError && !(allowedError instanceof LockedDeviceError)) {
```
alexandremgo, line 145: Good point

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/index.tsx
```diff
@@ -182,33 +215,46 @@ const SyncOnboardingScreen: React.FC<SyncOnboardingScreenProps> = ({
...
+          device,
+          error,
+          onRetry: isPollingOn ? undefined : notifyOnboardingEarlyCheckShouldReset,
+        })}
+      </Flex>
+    );
+  } else if (currentStep === "early-security-check" && lastSeenDevice) {
     stepContent = (
-      <EarlySecurityCheck
+      <EarlySecurityChecks
```
alexandremgo, line 232: I guess it's plural ahah

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/index.tsx
alexandremgo: I mean: i'm thinking right now on LLM: maybe we should only log that the device was considered locked, but not display anything ? Because for the early security check, locking a device should be impossible (except if it's considered "unresponsive" like i was saying above, but it's different than locked in this specific case)

alexandremgo: And we should move this locked device drawer inside the screen where the polling occurs (to know if the ESC or the sync companion should be rendered): in this case it could happen that the device is locked, and you can check on the `allowedError`

alexandremgo: Actually, i already return an `lockedDevice` from `useOnboardingStatePolling` ;)

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/ErrorDrawer.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onClickRetry: () => void;
+  closeable?: boolean;
+};
+
+const NotFoundEntityError = createCustomErrorClass("NotFoundEntityError");
+
+const ErrorDrawer: React.FC<Props> = ({ error, onClickRetry, closeable = false }) => {
+  const { t } = useTranslation();
+  const history = useHistory();
+  const isNotFoundEntityError = error.message === "not found entity";
```
alexandremgo, line 26: The error is not an instance of `FirmwareNotRecognized` ?

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/ErrorDrawer.tsx
```diff
@@ -0,0 +1,99 @@
...
+import TrackPage from "~/renderer/analytics/TrackPage";
+import { track } from "~/renderer/analytics/segment";
+
+export type Props = {
+  error: Error;
+  onClickRetry: () => void;
+  closeable?: boolean;
+};
+
+const NotFoundEntityError = createCustomErrorClass("NotFoundEntityError");
```
alexandremgo, line 21: It's really weird you need to create an error for this (i know it's to get the `ErrorDisplay` to work). But you are sure that in this case `error instanceof FirmwareNotRecognized` is not true ? The base class is `Error`, but you can still check if in the prototype chain (thx Javascript :pain:) you have this `FirmwareNotRecognized`

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/DeviceNotGenuineDrawer.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { localeSelector } from "~/renderer/reducers/settings";
+import { urls } from "~/config/urls";
+import TrackPage from "~/renderer/analytics/TrackPage";
+import { track } from "~/renderer/analytics/segment";
+
+export type Props = {
+  productName: string;
+};
+
+const DeviceNotGenuineError = createCustomErrorClass("DeviceNotGenuineError");
```
alexandremgo, line 22: Indeed there is no "not genuine" error existing for this case. But do you need to use `ErrorDisplay` here ? You can just display some translated text directly. 
Otherwise let's put this `DeviceNotGenuineError` in the errors package (but as it is not thrown from anywhere, i'm less a fan of that)

apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/ExitChecksDrawer.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { createCustomErrorClass } from "@ledgerhq/errors";
+import TrackPage from "~/renderer/analytics/TrackPage";
+import { track } from "~/renderer/analytics/segment";
+
+export type Props = {
+  onClose: () => void;
+  onClickExit: () => void;
+};
+
+const ExitError = createCustomErrorClass("ExitEarlySecurityChecksWarning");
```
alexandremgo, line 16: Idk if it's a good idea to create errors like this just to use `ErrorDisplay`. 
We should keep this for errors that are really thrown from somewhere on our libs

If it's coming from the logic of components (from LLM or LLD), and no error was thrown from our libs/hooks, we should just display the contents directly, without creating an error.

WDYT ?

alexandremgo (approved):
All good 🔥

ivp5:
This change breaks "Genuine" check, at least on Intel Macs, making Ledge Live Desktop hang on "Loading..." screen, BEFORE requesting to connect and unlock a device (so not USB related).

After testing versions available on old versions/signatures page, the last working build was 2.64.2 and the issue starts appearing from version 2.66.0 and persists until the currently-latest version (2.69.0).

For people encountering this issue (Ledger Live hangs showing spinning circle and "Loading..." at start of Genuine check) try downgrading to 2.64.2 where the "Loading..." screen is only seen for 0.5 second before showing the next screen ("Connect and unlock the device.." animation)

PR #3926, opened 2023-07-06
feat(recover): restore check seeded [PROTECT-1711]

Justkant:
### 📝 Description

Check if the device is already seeded/onboarded when going for a restore flow with recover

### ❓ Context

- **Linked resource(s)**: [`PROTECT-1711`](https://ledgerhq.atlassian.net/browse/PROTECT-1711) 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/components/RecoverRestore/index.tsx
```diff
@@ -1,34 +1,100 @@
...
+  }
+
+  if (state?.isOnboarded) {
+    return (
+      <Flex width="100%" height="100%" position="relative">
+        <Flex position="relative" height="100%" width="100%" flexDirection="column">
+          <OnboardingNavHeader onClickPrevious={() => history.push("/onboarding/select-device")} />
+          {renderError({
+            t,
+            error: new DeviceOnboarded("Your device is already setup with a Secret Recover Phrase"),
```
haammar-ledger, line 91: All user-facing errors should define a translation key, rather than define the message in the constructor, [here](https://github.com/LedgerHQ/ledger-live/blob/178d65887d7e6d67ddda8b83181c9fc1a3b0a13b/apps/ledger-live-desktop/static/i18n/en/app.json#L5026) for LLD ([here](https://github.com/LedgerHQ/ledger-live/blob/178d65887d7e6d67ddda8b83181c9fc1a3b0a13b/apps/ledger-live-mobile/src/locales/en/common.json) for LLM)

haammar-ledger (changes requested):
LGTM save for one comment

apps/ledger-live-desktop/src/renderer/components/RecoverRestore/index.tsx
```diff
@@ -1,34 +1,100 @@
...
+  }
+
+  if (state?.isOnboarded) {
+    return (
+      <Flex width="100%" height="100%" position="relative">
+        <Flex position="relative" height="100%" width="100%" flexDirection="column">
+          <OnboardingNavHeader onClickPrevious={() => history.push("/onboarding/select-device")} />
+          {renderError({
+            t,
+            error: new DeviceOnboarded(t("errors.DeviceAlreadySetup.title")),
```
haammar-ledger, line 91: The translation isn't required here, empty constructor works.

haammar-ledger (approved):

apps/ledger-live-desktop/src/renderer/components/RecoverRestore/index.tsx
stephane-lieumont-ledger: Are you sure that is not required ?

```
(alias) renderError({ error, t, withOpenManager, onRetry, withExportLogs, list, supportLink, warning, info, managerAppName, requireFirmwareUpdate, withOnboardingCTA, device, inlineRetry, }: {
    error: ErrorConstructor | Error;
    t: TFunction<"translation", undefined>;
    withOpenManager?: boolean | undefined;
    onRetry?: (() => void) | null | undefined;
    ... 9 more ...;
    inlineRetry?: boolean | undefined;
}): JSX.Element
import renderError
```

Justkant: In this case I don't think it was present by default because I just created this error for our use case

PR #3945, opened 2023-07-07
[FEAT][LIVE-8289] create evm send flow (LLM)

chabroA:
### 📝 Description

Create send flow for EVM family, using GasTrackerAbstraction introduced in https://github.com/LedgerHQ/ledger-live/pull/3611.

🔗 https://github.com/LedgerHQ/ledger-live/pull/3704

Can be tested using the `ethereum_as_evm_test_only` crypto available when enabling `Developer mode` under `Experimental features`  

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8289

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/families/evm/EvmCustomFees/utils.tsx
```diff
@@ -0,0 +1,51 @@
+import BigNumber from "bignumber.js";
+import { getEnv } from "@ledgerhq/live-env";
+import { inferDynamicRange, Range } from "@ledgerhq/live-common/range";
+import { GasOptions } from "@ledgerhq/coin-evm/lib/types";
+import invariant from "invariant";
+
+// This is 10 Gwei in wei. 20 Gwei is 2e8 or 20e9
+// Should we rename the variable to TEN_GWEI or update the value to 20e9?
+const TWENTY_GWEI = new BigNumber(10e9);
```
chabroA, line 9: This is 10 Gwei in wei. 20 Gwei is `2e8` or `20e9`
Should we rename the variable to TEN_GWEI or update the value to 20e9?

apps/ledger-live-mobile/src/components/SendRowsFee.tsx
```diff
@@ -37,9 +37,29 @@ export default ({
   ...props
 }: Props) => {
   const mainAccount = getMainAccount(account, parentAccount);
+
   // eslint-disable-next-line no-prototype-builtins
   if (perFamily.hasOwnProperty(mainAccount.currency.family)) {
     const C = perFamily[mainAccount.currency.family as keyof typeof perFamily];
+
+    /**
+     * Evm familly type does not have networkInfo property but we still want to
```
hedi-edelbloute, line 46: Typo on family

apps/ledger-live-mobile/src/components/SendRowsFee.tsx
```diff
@@ -37,9 +37,29 @@ export default ({
...
+     * display the fee selection logic
+     */
+    if (mainAccount.currency.family === "evm") {
```
hedi-edelbloute, line 49: Isn't there a better way to do it ? Shouldn't adding a SendRowFee component in evm family be enough ? Looks like it should be in perFamily map

apps/ledger-live-mobile/src/families/evm/EvmCustomFees/utils.tsx
hedi-edelbloute: I lean more for the TEN_GWEI name, and then use 2*TEN_GWEI if needed

libs/coin-evm/src/logic.ts
```diff
@@ -44,6 +44,12 @@ export const getEstimatedFees = (tx: EvmTransaction): BigNumber => {
   return tx.maxFeePerGas?.multipliedBy(gasLimit) || new BigNumber(0);
 };
 
+/**
+ * Helper to get the currency unit to be used for the fee field
+ */
+export const getDefaultFeeUnit = (currency: CryptoCurrency): Unit =>
+  currency.units.length > 1 ? currency.units[1] : currency.units[0];
```
hedi-edelbloute, line 51: Correct me if i'm wrong, but i understand we suppose there is always either 2 units or 1, shouldn't this be the last element then ?

hedi-edelbloute (changes requested):
ChangesRequested because i don't like how it's done for the SendRowsFee component, maybe move to the evm folder if possible, why has it been done like this for bitcoin ? :thinking: 

Also, don't we need to add UI tests ? (question)

apps/ledger-live-mobile/src/components/SendRowsFee.tsx
chabroA, line 46:
```suggestion
     * Evm family type does not have networkInfo property but we still want to
```

apps/ledger-live-mobile/src/components/SendRowsFee.tsx
chabroA: > Isn't there a better way to do it ?

Probably, but not that I am aware of

> Shouldn't adding a SendRowFee component in evm family be enough ?

We already have a `SendRowFee` component in the evm family folder (as well as in other families folders). But this component here seems to be the "root" component that will display the correct family specific component from the `perFamily` generated file based on the account family.

> why has it been done like this for bitcoin ?

I don't know. But it seems to be legacy code and might not be what we want.

```ts
  if (perFamily.hasOwnProperty(mainAccount.currency.family)) {
    const C = perFamily[mainAccount.currency.family as keyof typeof perFamily];
    // FIXME: looks like a hack, need to find how to handle networkInfo properly
    return (transaction as BitcoinTransaction)?.networkInfo ? (
      <C
        {...props}
        setTransaction={setTransaction}
        transaction={transaction}
        account={account}
        parentAccount={parentAccount}
        navigation={navigation}
        route={route}
      />
    ) : null;
  }
``` 

Indeed, here we will only display the `SendRowsFee` component for families that have such component defined in their own specific family folder (this is logic) **AND** that have a `networkInfo` property in their transaction.
This second condition seems odd because not all families have a `networkInfo` property in their transaction. For example `algorand`, but also the new `evm`. But these families have their own `SendRowsFee` component defined in their family folder. Same thing for `celo`, `polkadot`, `solana` (here we have an hardcoded patch, cf. [this](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/solana/js-prepareTransaction.ts#L113-L114)).
So, based on this logic, as of today the `SendRowsFee` for `algorand`, `celo` and `polkadot` would not be displayed.

I would rather get rid of the second condition altogether to have something like that.
Then each specific coin has the responsibility, if need be, to check that they have all the info needed to display their own `SendRowsFee` component, and return `null` otherwise (which is what is done in the `evm` one for example, cf. [here](https://github.com/LedgerHQ/ledger-live/pull/3945/files#diff-06721b193f06f6bec81920428dbc92a0a4778dae78488729526832544a038d01R91-R93)) 

```ts
  if (perFamily.hasOwnProperty(mainAccount.currency.family)) {
    const C = perFamily[mainAccount.currency.family as keyof typeof perFamily];
    return (
      <C
        {...props}
        setTransaction={setTransaction}
        transaction={transaction}
        account={account}
        parentAccount={parentAccount}
        navigation={navigation}
        route={route}
      />
    );
  }
```  

But I think this (little) rework should not be handled in this PR but rather in it's own PR to avoid unexpected side effects not related with the "create evm send flow" topic.

cc. @haammar-ledger FYI
cc. @gre maybe you have some clues on where this `(transaction as BitcoinTransaction)?.networkInfo` condition comes from? From the history, I see it is quite old (before the monorepo), hence my guess on it's possible legacy status.

chabroA:
> Also, don't we need to add UI tests ? (question)

Yes we do, but this will be handled as a broader topic under [this epic](https://ledgerhq.atlassian.net/browse/LIVE-8491) in another PR, as part of next PI focussed on Testing and QA strategies (AFAIK, no component UI unit tests have been setup for anything in the `families/**` solders and subfolders)

apps/ledger-live-mobile/src/components/SendRowsFee.tsx
chabroA: FYI, proposed solution implemented in this PR https://github.com/LedgerHQ/ledger-live/pull/4010

hedi-edelbloute: PR #4010 looks good ! Let's just check if there is no regression especially on btc

apps/ledger-live-mobile/src/families/evm/EvmCustomFees/EvmLegacyCustomFees.tsx
```diff
@@ -0,0 +1,117 @@
...
+  invariant(gasOptions, "GasPriceField: 'transaction.gasOptions' should be defined");
+
+  const networkGasPrice = inferDynamicRange(gasOptions.medium.gasPrice!, {
+    minValue: gasOptions.slow.gasPrice!,
+    maxValue: gasOptions.fast.gasPrice!,
+  });
+
+  if (!lastNetworkGasPrice) {
+    lastNetworkGasPrice = networkGasPrice;
+  }
```
lvndry, line 42: Why not use useMemo instead ?

apps/ledger-live-mobile/src/families/evm/SelectFeesStrategy.tsx
```diff
@@ -0,0 +1,255 @@
...
+import TachometerFast from "../../icons/TachometerFast";
+import TachometerMedium from "../../icons/TachometerMedium";
+import TachometerSlow from "../../icons/TachometerSlow";
+import SummaryRow from "../../screens/SendFunds/SummaryRow";
+import { sharedSwapTracking } from "../../screens/Swap/utils";
+import BigNumber from "bignumber.js";
+import invariant from "invariant";
+
+type Strategy = keyof GasOptions;
+type StrategyWithCustom = Strategy | "custom";
```
lvndry, line 35: Seems like we use `keyof GasOptions` a lot
I would put this type in `coin-evm` as well

libs/coin-evm/src/logic.ts
lvndry: Also it happens there's more than 2 elements. For exemple ethereum has 4.

https://github.com/LedgerHQ/ledger-live/blob/fd35196740b2baa0247be755043225300688024e/libs/ledgerjs/packages/cryptoassets/src/currencies.ts#L51

hedi-edelbloute (approved):
SendRowFee will be merged in another PR

apps/ledger-live-mobile/src/families/evm/EvmCustomFees/EvmLegacyCustomFees.tsx
chabroA: Good point

PR #3997, opened 2023-07-11
fix: do not remove peerDependencies of react-redux

KVNLS:
### 📝 Description

I'm working on a POC to introduce Integration tests on LLD (and LLM after).
When I try to test a component that needs redux I wrap it into a Redux Provider but I have an error regarding react-redux that can't find react-dom module.
Looks like we remove react-dom peer dependencies of react-redux in pnpfile.cjs.
The initial issue that introduced this change was when we started to use react-redux in live-common and consumes it in live-mobile.
I think that the initial issue is not valid anymore as we declare correctly react-redux as a peerDependency and it's declared as a devDependencies too.

Moreover, If we have an issue with react-redux peer dependencies declaration I think that we could use pnpm.packageExtensions to fix these issues:
https://pnpm.io/package_json#pnpmpackageextensions

### ❓ Context

- **Impacted projects**: `llm` `lld` `llc` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.pnpmfile.cjs
+++ b/.pnpmfile.cjs
@@ -52,13 +52,6 @@ function readPackage(pkg, context) {
         So we are going to patch these until the maintainers fix their own stuff…
         Feel free to make PRs if you feel like it :).
       */
-      /*
-        Remove react-native/react-dom from react-redux optional peer dependencies.
-        Without this, using react-redux code in LLM from LLC will fail because the package will get duplicated.
-      */
-      removeDependencies("react-redux", ["react-native", "react-dom"], {
-        kind: "peerDependencies",
-      }),
       /* Storybook packages */
       addDependencies("@storybook/webpack-config", { "resolve-from": "*" }),
       addDependencies("@storybook/addon-knobs", {


elbywan (approved):
> I think that the initial issue is not valid anymore as we declare correctly react-redux as a peerDependency and it's declared as a devDependencies too.

I think the only reason why it works for LLM is because we are forcing the react-redux resolution in the metro config file now:

```
// Dependencies that are forcefully resolved from the LLM folder.
const forcedDependencies = [
  "react-redux",
  "react-native",
  "react-native-svg",
  "styled-components",
];
```

> Moreover, If we have an issue with react-redux peer dependencies declaration I think that we could use pnpm.packageExtensions to fix these issues:
https://pnpm.io/package_json#pnpmpackageextensions

It is basically the same as using a pnpmfile, albeit far less powerful. There is no way to remove dependencies with package extensions AFAIK. I also advise against using multiple places to configure package resolution, keeping a single source of truth seems better to me.

KVNLS:
@elbywan thanks for the explanation it's way more clear for me what's happening. I was not aware of the `forcedDependencies` declaration. I will find a more appropriate solution to make it works

KVNLS:
@elbywan, so I had an extra look following the information you provided.
The initial problem comes from the module resolution with .pnpm folder. We have two react-redux one from llm and one from lld ( `node_modules/.pnpm/react-redux@7.2.9_@types+react@17.0.53_react@17.0.2` and `node_modules/.pnpm/react-redux@7.2.9_@types+react@17.0.53_react@18.0.0`) and looks like metro doesn't support well monorepos so that's why we have a custom module resolution that embeds both of them by default.
It crashes in LLM because we use two different react-redux.
The fix is to force react-redux resolution for LLM as already done with what you explained with `forcedDependencies` but removing peerDependencies on react-redux has no impact on this issue.
Another solution would be to avoid hoisting react-redux in node_modules/.pnpm folder.

PR #4000, opened 2023-07-11
Bugfix/LIVE-7413-7415-7416 Cosmos tx2op multiple event fix

hzheng-ledger:
### 📝 Description
A cosmos transactions can contain multiple events. 
E.g. a claim_reward cosmos transaction can contain two claim_reward events.
We handle only one of them (variable mainMessage in the code) at this moment.
With this PR, now, we can support multiple events by using correspondingMessages array instead of mainMessage.

### ❓ Context

- **Linked resource(s)**: `` 
https://ledgerhq.atlassian.net/browse/LIVE-7413
https://ledgerhq.atlassian.net/browse/LIVE-7414
https://ledgerhq.atlassian.net/browse/LIVE-7415
https://ledgerhq.atlassian.net/browse/LIVE-7416

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -54,130 +54,110 @@ const txToOps = (info: AccountShapeInfo, accountId: string, txs: CosmosTx[]): Op
 
     const correspondingMessages = messages.filter(m => m.type === mainMessage.type);
 
-    if (correspondingMessages.length === 0) {
```
hzheng-ledger, line 57: this check is not needed. I think. It never happens.

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -54,130 +54,110 @@ const txToOps = (info: AccountShapeInfo, accountId: string, txs: CosmosTx[]): Op
...
-          if (!op.type && attributes.sender === address) {
-            op.type = "OUT";
-            op.value = op.value.plus(fees);
-          } else if (!op.type && attributes.recipient === address) {
-            op.type = "IN";
+        // TODO: handle IBC transfers here
+        for (const message of correspondingMessages) {
+          const amount = message.attributes.find(attr => attr.key === "amount")?.value;
+          const sender = message.attributes.find(attr => attr.key === "sender")?.value;
+          const recipient = message.attributes.find(attr => attr.key === "recipient")?.value;
```
hedi-edelbloute, line 63: Are we sure there is always one recipient ? If we try to send to two recipient, will there be two message or two attributes ?

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -54,130 +54,110 @@ const txToOps = (info: AccountShapeInfo, accountId: string, txs: CosmosTx[]): Op
...
+          if (amount && sender && recipient && amount.indexOf(currency.units[1].code) != -1) {
+            if (op.senders.indexOf(sender) === -1) op.senders.push(sender);
```
hedi-edelbloute, line 65: Why do we need to do this check ? I don't understand in what case we fill the operation and then ignore other senders, same for other similar checks

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -54,130 +54,110 @@ const txToOps = (info: AccountShapeInfo, accountId: string, txs: CosmosTx[]): Op
...
+            if (op.recipients.indexOf(recipient) === -1) op.recipients.push(recipient);
+            op.value = op.value.plus(amount.replace(currency.units[1].code, ""));
+            if (!op.type && sender === address) {
```
hedi-edelbloute, line 68: Why do we need to check if op type is already set ?

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -54,130 +54,110 @@ const txToOps = (info: AccountShapeInfo, accountId: string, txs: CosmosTx[]): Op
...
-          });
+      }
+      case "unbond": {
+        op.type = "UNDELEGATE";
+        op.value = new BigNumber(fees);
+        const unbondShards: { amount: BigNumber; address: string }[] = [];
+        for (const message of correspondingMessages) {
+          const amount = message.attributes.find(attr => attr.key === "amount")?.value;
+          const validator = message.attributes.find(attr => attr.key === "validator")?.value;
+          if (amount && validator && amount.indexOf(currency.units[1].code) != -1) {
```
hedi-edelbloute, line 147: We could add a variable named `const validAmountUnit = amount.indexOf(currency.units[1].code) !== -1`

Minor suggestion

Same in other op type

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hedi-edelbloute: I trust your analysis, let's remove it ! If i remember correctly it was more as a safety net, same as the loop in reward section

hedi-edelbloute (approved):
Lgtm, just some things to look out for in transfer operation types

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hzheng-ledger: In one "CosmosMessage" (event), in a send event, we have only one "sender" sender and one "recipient". 
If you want to send funds to multiple recipients, it is possible that have multiple events in one transactions.
we can refer to 
the cosmos sdk source code: https://github.com/cosmos/cosmos-sdk/blob/1b0fcdc9f059419328ed2b6b2393320fb042270f/x/bank/keeper/send.go#L141
to have a look at the source code of multi send transaction and make sure that one event can have only one sender and one recipient.

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hzheng-ledger: if (op.senders.indexOf(sender) === -1) op.senders.push(sender);
-> we can make sure that there is no duplicated sender in the op.senders array.
if (op.recipients.indexOf(recipient) === -1) op.recipients.push(recipient);
-> we can make sure that there is no duplicated sender in the op.recipients array.

In the case one transaction contains multiple "transfer" events, e.g. a same sender sends funds to multiple different recipients, we don't want to display the same sender address multiple time in UI

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hzheng-ledger: yes, you are right, it is not necessary.
but I don't think it breaks anything.
anyway, I copy the same code from 
https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts#L76

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hzheng-ledger: Thanks a lot
I've done some minor refactoring to make the code more concise

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hedi-edelbloute: I think we should remove useless code

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hedi-edelbloute: Thanks, perfect

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hedi-edelbloute: Why do we need to make sure there aren't duplicated sender ? let's check if it happens. We have to remove useless code

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hzheng-ledger: fixed. op.type check removed

hedi-edelbloute: Thanks !

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -54,130 +55,105 @@ const txToOps = (info: AccountShapeInfo, accountId: string, txs: CosmosTx[]): Op
...
-          op.extra.sourceValidator = attributes.source_validator;
+      }
+      case "redelegate": {
+        op.type = "REDELEGATE";
+        op.value = new BigNumber(fees);
+        const redelegateShards: { amount: BigNumber; address: string }[] = [];
+        for (const message of correspondingMessages) {
+          const amount = message.attributes.find(attr => attr.key === "amount")?.value;
+          const validatorDst = message.attributes.find(attr => attr.key === "validator_dst")?.value;
+          const validatorSrc = message.attributes.find(attr => attr.key === "validator_src")?.value;
```
haammar-ledger, line 124: @hzheng-ledger This introduced a regression: https://ledgerhq.atlassian.net/browse/LIVE-8787

Indeed the attribute keys were changed from `source_validator` and `destination_validator` to `validator_src` and `validator_dst`, was this intentional?

I'm seeing that at least for ATOM the messages still have this structure:
```
[
  {
    key: "source_validator",
    value: "cosmosvaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7",
  },
  {
    key: "destination_validator",
    value: "cosmosvaloper10wljxpl03053h9690apmyeakly3ylhejrucvtm",
  },
  {
    key: "amount",
    value: "5031uatom",
  },
  {
    key: "completion_time",
    value: "2022-12-07T16:41:08Z",
  },
]
```

PR #4010, opened 2023-07-12
[SUPPORT] remove unnecessary check on tx networkInfo to render SendRowsFee

chabroA:
### 📝 Description

Following this comment https://github.com/LedgerHQ/ledger-live/pull/3945#discussion_r1259334119 and the related discussion, proposal to remove what seem to be an unnecessary condition on the existence of a `networkInfo` property on the `transaction` object to render (or not) a family custom `SendRowsFee` component.

Not all crypto transaction types have a `networkInfo` property and some of them have some custom `SendRowsFee` component.
If a family `SendRowsFee` component display is dependent on the presence of a `networkInfo` property in the `transaction` object, these families own `SendRowsFee` component will never be displayed.

This condition seems odd because not all families have a `networkInfo` property in their `transaction`. For example `algorand`, but also the new `evm`. But these families have their own `SendRowsFee` component defined in their family folder. Same thing for `celo`, `polkadot`, `solana` (here we even have some hardcoded patch, cf. [this](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/solana/js-prepareTransaction.ts#L113-L114)).
So, based on this logic, as of today the `SendRowsFee` for `algorand`, `celo` and `polkadot` would not be displayed.

Furthermore, we don't seem to have similar logic on LLD (cf. [here](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/modals/Send/SendAmountFields.tsx) and [here](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/modals/SignTransaction/SendAmountFields.tsx)) where we don't have an condition on `networkInfo`. 

The only place where I found such condition on LLD are these 2 files ([here](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FeesDrawer/index.tsx#L55) and [here](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FormSummary/SectionFees.tsx#L84-L85)) related to swap, which might probably be some legacy logic from original swap implementation since these 2 conditions control the display of the `SendAmountFields` file listed just above, and it would not make sense to always display this component on a normal send flow but not on a swap flow (if I can edit fees on a normal send, why could I not do it in a Swap flow?)

### ❓ Context

- **Impacted projects**: `LLM` `LLD`
- **Linked resource(s)**: https://github.com/LedgerHQ/ledger-live/pull/3945#discussion_r1259334119

### ✅ Checklist

- [ ] **Test coverage**
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FeesDrawer/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FeesDrawer/index.tsx
@@ -52,7 +52,7 @@ export default function FeesDrawer({
       />
       <DrawerTitle i18nKey="swap2.form.details.label.fees" />
       <Box mt={3} flow={4} mx={3}>
-        {transaction && "networkInfo" in transaction && transaction.networkInfo && mainAccount && (
+        {transaction && mainAccount && (
           <SendAmountFields
             account={mainAccount as Account}
             parentAccount={parentAccount}
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FormSummary/SectionFees.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FormSummary/SectionFees.tsx
@@ -78,14 +78,7 @@ const SectionFees = ({
   const family = mainFromAccount?.currency.family;
   const sendAmountSpecific = account && family && getLLDCoinFamily(family)?.sendAmountFields;
   const canEdit =
-    hasRates &&
-    showSummaryValue &&
-    transaction &&
-    "networkInfo" in transaction &&
-    transaction.networkInfo &&
-    account &&
-    family &&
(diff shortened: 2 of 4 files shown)


hedi-edelbloute (approved):

hedi-edelbloute (approved):

chabroA:
cherry-picked on develop and moved in this PR https://github.com/LedgerHQ/ledger-live/pull/4020 to make it independent from gas tracker work

chabroA:
cherry-picked on develop and moved in this PR https://github.com/LedgerHQ/ledger-live/pull/4020 to make it independent from gas tracker work

PR #4016, opened 2023-07-12
Feat/llm/Implement and harmonise tracking for the entry point of each staking flow

sarneijim:
### 📝 Description

Implement and harmonise tracking for the entry point of each staking flow (delegation) LLM
Harmonise algorand claim rewards with other currencies. LLD

### ❓ Context

- **Impacted projects**: LLM LLD
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-7836

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/perfect-cheetahs-hide.md
+++ b/.changeset/perfect-cheetahs-hide.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Add attributes in stake screen tracks
--- a/.changeset/witty-turtles-search.md
+++ b/.changeset/witty-turtles-search.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Add algo claim reward track attributes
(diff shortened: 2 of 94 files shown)


apps/ledger-live-mobile/src/reducers/accounts.ts
```diff
@@ -262,7 +263,7 @@ export const parentAccountSelector = createSelector(
 export const accountScreenSelector =
   (route?: {
     params?: {
-      account?: AccountLike;
+      account?: AccountLike | CosmosAccount;
```
kallen-ledger, line 266: [Question]: Why the need to add `CosmosAccount` here?

sarneijim: I think `AccountLike` it is the common type provided by any coin, it is failing because we need to use specific CosmosAccount  properties.

kallen-ledger: Perhaps it might be better to leave this as is since it seems quite generalised.

One idea is to use `getAccountCurrency` helper along with the `accountScreenSelector`?

kallen-ledger (changes requested):
Perhaps it may be better to use the `getAccountCurrency` helper along with account selector to get the currency id for cosmos.

sarneijim:
> > Perhaps it may be better to use the `getAccountCurrency` helper along with account selector to get the currency id for cosmos.
> 
> I am going to retest the type, but if it is CosmosAccount, I think we must improve the type, not change the method.

@kallen-ledger , you are totally right. Your suggested changes have been applied.

kallen-ledger (approved):

PR #4022, opened 2023-07-13
fix: wallet tabs scroll to top and scroll sync

mderrick:
### 📝 Description

This fixes the wallet/nft tab scroll sync behaviour that breaks under two different scenarios:

**Bug 1:**
1. Scroll down on one tab so the header is out of view.
2. Switch tab and scroll up so the header is in view (not pull to refresh, just up to top)
3. Switch back to the other tab and it will be broken

NOTE: This seems to only break when the scroll views are re-rendered (which happens a lot when hot reloading on dev - in prod, i'm sure this can still happen just not as common?).

**Fix:** I believe that when `onGetRef` is called it never updated the scroll ref if it already existed. This meant that none of the scrolling features that uses these refs (including the "scroll to top" on the NFT gallery) would work and instead would be trying to scroll a scrollview that no longer existed in the view.

**Bug 2:**
1. Pull to refresh on any tab
2. Navigate to another tab before it finishes loading

**Fix:** Trigger the sync of the scroll views even when scrollY is negative (pulled to refresh).

### ❓ Context

- **Linked resource(s)**: `LIVE-4741` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/components/WalletTab/WalletTabNavigatorScrollManager.tsx
```diff
@@ -55,7 +55,7 @@ export default function WalletTabNavigatorScrollManager({
           if (
             scrollYValue !== null &&
             item.value &&
-            ((scrollYValue < headerHeight && scrollYValue >= 0) ||
+            (scrollYValue < headerHeight ||
```
mderrick, line 58: I'm uncertain of the reason this existed but this keeps the view sin sync when pulling to refresh. Is there another use case I may have broken doing this?

mcayuelas-ledger: @nparigi-ledger

nparigi-ledger (approved):

PR #4030, opened 2023-07-14
fix(LIVE-8530): quotes are not readable in light mode

cng-ledger:
### 📝 Description

Swap quotes are not readable in light mode. `Price` component has an optional `color` prop which determines the text color. If value is not provided text color seemed to be defaulted to white. 

Code changes to use `theme.colors.palette.text.shade100` as default value the `color` is not provided.

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8530 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

Screenshot 2023-07-14 at 09 28 03
Screenshot 2023-07-14 at 09 28 17

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/new-planets-raise.md
+++ b/.changeset/new-planets-raise.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+quotes are not readable in light mode
--- a/apps/ledger-live-desktop/src/renderer/components/Price.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Price.tsx
@@ -67,7 +67,10 @@ export default function Price({
     ? BigNumber(rawCounterValue)
     : rawCounterValue;
   const theme = useTheme();
-  const textColor = useMemo(() => color ?? theme.colors.palette.text.shade100, [color, theme]);
+  const textColor = useMemo(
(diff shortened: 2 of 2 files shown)


sarneijim (changes requested):
This is a generic component (check the path) the text color component is provided by color property. I think you need to change the parent not the component.

cng-ledger:
> This is a generic component (check the path) the text color component is provided by color property. I think you need to change the parent not the component itself (the color props property).

The `color` prop in the `Price` component is key of the `colors` from `~/renderer/styles/theme`, rather than a string. By passing the value from parent component, it will require an extra condition to check the current theme mode (Light / Dark) to pass the right value to the `Price` component.

Code change in this PR use the `theme.colors.palette.text.shade100` from the current theme, which is switched by the depending on the mode. (Light / Dark)

Which means, if elsewhere in the code base we pass a `color` to this `Price` component, the behaviour should not change. If no color is passed, it will use the color from the theme itself. Switching between dark / light color text.

Let me know what you think of this @sarneijim !!!

sarneijim (approved):

PR #4034, opened 2023-07-17
Fix: LIVE-8553 polkadot not enough balance error

hzheng-ledger:
### 📝 Description

-. Steps to Reproduce the bug:

An account of Polkadot with available balance between 0-1 DOT and enough bonded balance (or unbonded balance) >1 DOT

Try to send funds (e.g. 0.5 DOT) from this account.

-. Expected behavior:

We should be able to send funds from this account because it is lower than “send max” balance.

-. Actual behavior:

"Not enough balance" error.

In our old code, we had such check that raises the NotEnoughBalance error which is not correct.
```
if (
      !errors.amount &&
      !t.useAllAmount &&
      a.spendableBalance.lte(EXISTENTIAL_DEPOSIT.plus(EXISTENTIAL_DEPOSIT_RECOMMENDED_MARGIN))
    ) {
      errors.amount = new NotEnoughBalance();
    }
```

In this PR, I've removed this check and added getTransactionStatus.unit.test.ts to cover this bug.
Also, I've done some refactoring for polkadot family.

### ❓ Context

- **Impacted projects**:  LLC
- **Linked resource(s)**:  https://ledgerhq.atlassian.net/browse/LIVE-8553

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach


libs/coin-polkadot/src/__tests__/getTransactionStatus.unit.test.ts
haammar-ledger, line 1: This unit test is literally doing the same that can be done a bit easier as a bridge integration test :)
But I guess the debate is still on, about pros and cons of both...

haammar-ledger (commented):
LGTM but please run the bot a couple times to be safe. Thanks :pray:

PR #4053, opened 2023-07-18
feat(LLM/Ratings): New rating flow based on a NPS input

mle-gall:
### 📝 Description

New ratings flow based on a typeform containing a NPS input. NPS value is set in user properties in segment. Inputs from the typeform are tracked into segment.
The feature is based on a new feature flag : npsRatingsPromt

### ❓ Context

- **Impacted projects**: LLM/LLC/LLT
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-7518 
- 
### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/39890664/b294e10c-89d9-43bf-a359-e2b6f82c31c0

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/logic/npsRatings.tsx
```diff
@@ -0,0 +1,346 @@
...
+  alreadyClosedFromInitStep?: boolean;
+  /** Whether or not the user already rated the app */
+  alreadyRated?: boolean;
+  /** If true, we will not prompt the rating flow again unless the user triggers it manually from the settings */
+  doNotAskAgain?: boolean;
+  /** "satisfied" if user clicked on the "Satisfied" cta in the ratings flow, "disappointed" if user clicked on the "I'm disappointed" in the ratings flow */
+  satisfaction?: string;
+};
+
+const ratingsDataOfUserAsyncStorageKey = "ratingsDataOfUser";
```
cgrellard-ledger, line 56: Should we update this storage key to be different than the one of the original ratings flow ?

apps/ledger-live-mobile/src/logic/npsRatings.tsx
ofreyssinet-ledger, line 1: given the very minor differences between this and the original hook in `ratings.tsx` is it possible to reuse that hook with just slight modifications?
It's quite heavy logic so we should be careful to not duplicate code like this if possible, it will just make everything harder to maintain.

The only differences are:
- the feature flag being used
- the initial step
- the tracking

(these could be passed as params to a `useRatingsBaseLogic()` or something like that)

- `updateNpsTracking` function

Also since these files only export a hook maybe rename them with the name of the hook 🤷

apps/ledger-live-mobile/src/screens/NpsRatingsModal/DisappointedDone.tsx
```diff
@@ -0,0 +1,73 @@
...
+      page: "NPS Step 3 Thank you for feedback",
+      button: "mailto",
+    });
+  }, [ratingsFeatureParams, ratingsHappyMoment?.route_name]);
+
+  return (
+    <Flex flex={1} alignItems="center" justifyContent="center" mt={3}>
+      <TrackScreen
+        flow="NPS"
+        page="3NPS Step 3 Thank you for feedback"
```
ofreyssinet-ledger, line 36: typo here?

libs/ledgerjs/packages/types-live/src/feature.ts
```diff
@@ -75,7 +76,8 @@ export type FeatureId =
   | "protectServicesDesktop"
   | "transactionsAlerts"
   | "listAppsV2"
-  | "editEthTx";
+  | "editEthTx"
+  | "listAppsV2";
```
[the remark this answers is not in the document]
ofreyssinet-ledger: 🦅 👁️

libs/ledger-live-common/src/featureFlags/defaultFeatures.ts
```diff
@@ -122,6 +122,63 @@ export const defaultFeatures = {
   editEthTx: {
     enabled: false,
   },
+  npsRatingsPrompt: {
```
ofreyssinet-ledger, line 125: instead of creating a new feature flag maybe there should be only 1 flag that configures whether we use the NPS flow or the old flow?

apps/ledger-live-mobile/src/screens/NpsRatingsModal/DisappointedDone.tsx
```diff
@@ -0,0 +1,73 @@
...
+        source={ratingsHappyMoment?.route_name}
+        params={ratingsFeatureParams}
+      />
+      <Text
+        variant="h4"
+        fontWeight="semiBold"
+        color="neutral.c100"
+        textAlign="center"
+        lineHeight="34.8px"
```
ofreyssinet-ledger, line 45: this shouldn't be necessary, same for the other instance below

apps/ledger-live-mobile/src/screens/NpsRatingsModal/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+    formUrlSplitted[0] +
+    `#app_version=${appVersion}` +
+    `&app_language=${language}` +
+    `&platform_os=${Platform.OS}` +
+    `&platform_version=${Platform.Version}` +
+    `&model_id=${lastDevice?.modelId}` +
+    `&firmware_version=${lastDevice?.deviceInfo?.version}` +
+    `&notifications_allowed=${notificationsAllowed}` +
+    `&notifications_blacklisted=${notificationsBlacklisted}` +
+    `&done?${formUrlSplitted[1] || ""}`;
```
ofreyssinet-ledger, line 106: better to use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) for this type of URL manipulation, rather than direct string manipulation which is more error prone

ofreyssinet-ledger (commented):
Overall it seems good but it would be wise to avoid duplicating so much code with almost 0 difference in the core logic, see my comments above.

mle-gall:
> Overall it seems good but it would be wise to avoid duplicating so much code with almost 0 difference in the core logic, see my comments above.

I generally agree on that and that's the way I've initially done the feature. But considering the future removal of the original ratings flow, I've splitted the components and logic, creating duplicates (which is not optimal I agree), in order to simplify the work when removing the old feature.

cgrellard-ledger (approved):

PR #4065, opened 2023-07-19
bugfix: LIVE-8299 osmo naming error

hzheng-ledger:
### 📝 Description

when the method cryptoFactory("osmosis") is called and if cosmosChainParams["osmo"] is not null, currently, the method returns cosmosChainParams["osmosis"] and it is null.
but we are supposed to return cosmosChainParams["osmo"]

the null cause the bug that is mentioned in https://ledgerhq.atlassian.net/browse/LIVE-8299

### ❓ Context

- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-8299

### ✅ Checklist

- [x] **Test coverage** The fix is inside the "cryptoFactory" method and it uses a global variable cosmosChainParams. It is hard to write unit test.
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach


hedi-edelbloute (approved):
We could actually check in unit test that we have something defined in the map for each currency, something like :
`
const cases = ["osmo", "osmosis", "axelar", "cosmos",...]
it.each(cases)(
    "should not return null with currency id %p",
    (currencyId) => {
      const api = cryptoFactory(currencyId);
      expect(api).toBeTruthy();
    }
  );
  
  it("should throw an error when currency id is unknown", () => {
       const factoryCall = () => cryptoFactory("definitelyUnknownCurrency");
      expect(factoryCall).toThrowError();
  })
  
    it("should specify which currency is unknown when an error is thrown", () => {
      const currencyId = "i am a currency name";
       const factoryCall = () => cryptoFactory("i am a currency name");
      expect(factoryCall).toThrowError(currencyId);
  })`

PR #4068, opened 2023-07-19
[BUGFIX][HOTFIX] don't reset data in for fallback tx in prepareTransaction

chabroA:
### 📝 Description

Don't reset data in for fallback tx in prepareTransaction, which could cause a smart contact interaction to be transformed to a send operation.

TODO:

- [ ] Tests
- [ ] Changeset

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** --- **In progress**
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/ethereum/prepareTransaction.ts
+++ b/libs/ledger-live-common/src/families/ethereum/prepareTransaction.ts
@@ -51,7 +51,7 @@ export const prepareTransaction: AccountBridge<Transaction>["prepareTransaction"
         return {
           from: account.freshAddress,
           to: recipient,
-          data: undefined,
+          data: tx.data,
           value: tx.amount,
         };
       }
--- a/libs/ledger-live-common/src/families/ethereum/prepareTransaction.ts
+++ b/libs/ledger-live-common/src/families/ethereum/prepareTransaction.ts
@@ -1,12 +1,18 @@
-import invariant from "invariant";
+import { log } from "@ledgerhq/logs";
 import { AccountBridge } from "@ledgerhq/types-live";
-import { getNetworkInfo, inferGasPrice, inferMaxFeePerGas, inferMaxPriorityFeePerGas } from "./gas";
-import { estimateGasLimit } from "./gas";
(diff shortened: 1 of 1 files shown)


libs/ledger-live-common/src/families/ethereum/prepareTransaction.ts
```diff
@@ -44,14 +50,15 @@ export const prepareTransaction: AccountBridge<Transaction>["prepareTransaction"
...
+        log("error", "prepareTransaction: buildEthereumTx failed", e);
         // fallback in case @ethereumjs/tx breaks on building a potentially partial transaction
         const subAccount = tx.subAccountId ? findSubAccountById(account, tx.subAccountId) : null;
         const recipient =
           subAccount?.type === "TokenAccount" ? subAccount.token.contractAddress : tx.recipient;
         return {
           from: account.freshAddress,
           to: recipient,
-          data: undefined,
+          data: tx.data,
```
aboissiere-ledger, line 61: Can it be feature flagged?

chabroA:
This is actually not the issue

Why is that:
- the protoTransaction (defined by a return from buildEthereumTx  or the fallback returned in the catch) is only used to get an [estimatedGasLimit](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/ethereum/prepareTransaction.ts#L61-L73). It’s not passed down the flow for signature
- the buildEIP1559Transaction [log](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/ethereum/transaction.ts#L238), with the tx object missing data is called in the buildEthereumTx function (before we reach the catch clause)
- adding the data to the transaction in the catch might/will trigger problems as this step is supposed to prevent some errors being visible in the send flow because of the estimateGasLimit backend call which is a simulation of the transaction and therefore would fail if the transaction was incorrect or incomplete.

PR #4108, opened 2023-07-21
Bugfix Live-8247 Cosmos undelegate/redelegate LLM error when available balance is not enough

hzheng-ledger:
### 📝 Description

In LLM cosmos account, when user available balance is too low to pay transaction fee, we don't have any check on UI side. So user can continue the transaction and finally get an error from Rpc node. We should give user balance error message on UI side before the transaction broadcast and the "continue button" should be disabled.
The LLD has such balance check, so the LLM should have the same behaviour.

### ❓ Context

- **Impacted projects**: `` LLM
- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-8247

### ✅ Checklist

- [ ] **Test coverage** The fix is on LLM UI part, I add some balance check on UI part and raise error if available balance is low. It is very hard to write unit test for this bug.
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 📸 Demo
On LLM, when available balance is too low, this error appears:
Screenshot 2023-07-22 at 01 20 37

for reference, on LLD,

Screenshot 2023-07-21 at 18 03 35

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/families/cosmos/shared/02-SelectAmount.tsx
```diff
@@ -138,7 +144,7 @@ function DelegationAmount({ navigation, route }: Props) {
                 value={value}
                 onChange={setValue}
                 inputStyle={styles.inputStyle}
-                hasError={error}
+                hasError={isAmountOutOfRange || isNotEnoughBalance}
```
haammar-ledger, line 147: The UI should contain as little blockchain logic as possible, these components must keep relying on `error`, and the changes must be done in the `getTransactionStatus` function instead.

haammar-ledger (changes requested):

hedi-edelbloute (changes requested):
This should come from getTransactionStatus imo

haammar-ledger (approved):
After discussion, this PR doesn't introduce new tech debt, as the whole component should be reworked, and this is out of scope for this bugfix.
A ticket should be created to rework the `getTransactionStatus` function and the whole component to use it.

PR #4113, opened 2023-07-24
[LIVE-8528] Add retry mechanism for app restoration errors during FW update

grsoares21:
### 📝 Description
This PR adds two improvements to improve User Experience in case an error happens during the app restoration phase of a firmware update. 

The first is an automatic retry mechanism in the app installation operation in order to recover from quick transient errors (e.g. errors that might happen due to an unstable network).

The second one is a manual retry mechanism in case the app restoration fails to recover automatically. It works using the same manual retry mechanism as the other restoration phases of the firmware update. 

### ❓ Context
- **Impacted projects**: `ledger-live-common`,`ledger-live-mobile`
- **Linked resource(s)**: [LIVE-8528]

### ✅ Checklist

- [ ] **Test coverage**
- [x] **Atomic delivery**
- [x] **No breaking changes**

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/6013294/f4afbc16-dfdc-414e-a796-aef76b87d0b2

### 🚀 Expectations to reach
Test plan:
The part concerning the automatic recovery might be hard to test. But in theory, one could disable and re-enabe their network so the first try fails but a subsequent one succeeds. In this case, no drawer will appear indicating any error and the installation will continue seamlessly.

The part concerning the manual retry is easier to test. First you need to disable the network data until an error drawer appears during the app installation, then you can re-enable the network and click on the retry button. Live will take some time to realize which apps have already been installed, but it should continue where it stopped and not reinstall already installed apps. Note that if you disable the network in the middle of an app installation, the installation will still work, since by the time the installation begins we have already downloaded all the data needed from the network, so you'll need to wait until the next app starts being installed to see any errors (see the demo video).

[LIVE-8528]: https://ledgerhq.atlassian.net/browse/LIVE-8528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/hw/installApp.ts
+++ b/libs/ledger-live-common/src/hw/installApp.ts
@@ -5,6 +5,11 @@ import Transport from "@ledgerhq/hw-transport";
 import type { ApplicationVersion, App } from "@ledgerhq/types-live";
 import ManagerAPI from "../manager/api";
 import { getDependencies } from "../apps/polyfill";
+import { retryWithDelay } from "../rxjs/operators/retryWithDelay";
+
+const APP_INSTALL_RETRY_DELAY = 500;
+const APP_INSTALL_RETRY_LIMIT = 5;
+
 export default function installApp(
   transport: Transport,
   targetId: string | number,
@@ -20,6 +25,7 @@ export default function installApp(
     firmwareKey: app.firmware_key,
     hash: app.hash,
   }).pipe(
+    retryWithDelay(APP_INSTALL_RETRY_DELAY, APP_INSTALL_RETRY_LIMIT),
     filter((e: any) => e.type === "bulk-progress"), // only bulk progress interests the UI
     throttleTime(100), // throttle to only emit 10 event/s max, to not spam the UI
     map((e: any) => ({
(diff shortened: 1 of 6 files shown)


libs/ledger-live-common/src/hw/installApp.ts
```diff
@@ -20,6 +25,7 @@ export default function installApp(
     firmwareKey: app.firmware_key,
     hash: app.hash,
   }).pipe(
+    retryWithDelay(APP_INSTALL_RETRY_DELAY, APP_INSTALL_RETRY_LIMIT),
```
alexandremgo, line 28: Maybe we should make sure QA (or you did ?) also tests the installation of an app from the manager ? I guess now it's also retrying 5 times if it fails (which is good 🔥)

alexandremgo (approved):
Nice addition ! 🦖

libs/ledger-live-common/src/hw/installApp.ts
grsoares21: We tried, but it is really hard to properly QA the retries, we have to manually cause an error (e.g. by cutting the network from the device) and then remove the cause of the error before the retries are over. I did test it locally by increasing the interval and adding logs to the retries, it all seems to work as it should, but it's hard to test in staging. We did check that it didn't break anything on the installations though.

PR #4127, opened 2023-07-25
Live-8737: Remove dead FTX/Wyre code from Swap - LLD

ggilchrist-ledger:
- feat: initial commit - tidying up code formatting and removing redundant exports
- fix: form component imports
- feat: more whitespacing and comments
- chore: more whitespacing and tidying up
- chore: more whitespacing before removing code

Scenarios to test:
- [ ] Full E2E swap
- [ ] Modals and drawers working
  - [ ] Transactions drawers
  - [ ] Fee change drawer
  - [ ] Provider drawer 
- [ ] Add account banner working
- [ ] No more KYC available in `ledger` object in devtools
 

### 📝 Description

Remove FTX and Wyre related code from swap (LLD in this PR)
- Removed MFA, KYC and login LLD components
- Removed redundant imports and auth hooks/state in swap components
- Removed auth related code from Redux state
- Added some whitespace to swap files
- Removed Wyre icon

### ❓ Context

- **Impacted projects**: LLD
- **Linked resource(s)**: [Main task](https://ledgerhq.atlassian.net/browse/LIVE-8503) / [Subtask](https://ledgerhq.atlassian.net/browse/LIVE-8737)

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FormSummary/index.tsx
```diff
@@ -43,11 +47,14 @@ const SwapFormSummary = ({ swapTransaction, provider }: SwapFormSummaryProps) =>
   } = swapTransaction.swap.from;
   const { currency: toCurrency, account: toAccount } = swapTransaction.swap.to;
   const ratesState = swapTransaction.swap.rates;
-  const hasRates = ratesState?.value?.length && ratesState?.value?.length > 0;
+  const hasRates = ratesState?.value?.length && ratesState?.value?.length > 0; // TODO why do we need this and the below line? They appear to be the same thing
```
chabroA, line 50: That's quite odd indeed.

We should be able to do without one of them.

As I understand it, the `hasRates` is a shorthand for the default value of the `hasFetchedRates` state value and this `hasFetchedRates` state value is updated every time `hasRates` changes (through the `useEffect`).

Two possibilities:
- Remove the `hasFetchedRates` line and `useEffect` update logic entirely and only use `hasRates`.
This should work fine if `ratesState?.value?.length` is not unassigned or updated to 0 between renders. Because with today's logic, once `hasRates` is `true` once, `hasFetchedRates` is set to `true` and not updated anymore.
Because `setHasFetchedRates(v => (!v ? hasRates : v))` updates the `hasFetchedRates` value (to `hasRates`) only if the previous value of `hasFetchedRates` was falsy. If the previous value was truthy, it reassigns it to the same value (therefore not updating the actual value).

- Keep the `hasRates` and `hasFetchedRates` logic (and maybe rename `hasFetchedRates` to `hasFetchedRatesOnce` to be more explicit) and only use `hasFetchedRates` in the returned component. 
In the returned component, both `hasFetchedRates` and `hasRates` are used. This seems odd and, based on current logic, only `hasFetchedRates` should be used (since it is derived from `hasRates`). Also, `hasFetchedRates` and `hasRates` *may* have different values at any point in time (cf. explanation above)

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.tsx
```diff
@@ -98,29 +77,29 @@ export const useProviders = () => {
...
 };
+
 const SwapForm = () => {
-  // FIXME: should use enums for Flow and Banner values
-  const [currentFlow, setCurrentFlow] = useState<string | null>(null);
-  const [currentBanner, setCurrentBanner] = useState<string | null>(null);
   const swapDefaultTrack = useGetSwapTrackingProperties();
   const [idleState, setIdleState] = useState(false);
-  const [error, setError] = useState<ValidCheckQuoteErrorCodes | undefined>();
+  const [error, setError] = useState<ValidCheckQuoteErrorCodes | undefined>(); // TODO: delete?
```
chabroA, line 84: It seems to only be linked to the `checkQuote` logic that was used for FTX. Could probably be cleaned up / removed.

cf.
- https://github.com/LedgerHQ/ledger-live/blob/support/live-8503-remove-ftx-wyre-code-lld/libs/ledger-live-common/src/exchange/swap/types.ts#L104-L133
- https://github.com/LedgerHQ/ledger-live/blob/support/live-8503-remove-ftx-wyre-code-lld/libs/ledger-live-common/src/exchange/swap/checkQuote.ts

Also, you should get rid of all the KYC related logic in LLC as well
cf.
- https://github.com/LedgerHQ/ledger-live/blob/support/live-8503-remove-ftx-wyre-code-lld/libs/ledger-live-common/src/exchange/swap/types.ts#L8-L24

One possibility would be to start by removing the types (cf. links above) and pull up the thread to the actual functions 🧹

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/SwapConnectWidget/index.tsx
```diff
@@ -1,105 +0,0 @@
-import * as remote from "@electron/remote";
```
chabroA, line 1: if you remove the `SwapConnectWidget`, then the `swapConnectWebviewPreloader` related files should be removed as well:
- `apps/ledger-live-desktop/src/swapConnectWebviewPreloader/index.ts`
- `apps/ledger-live-desktop/tools/config/swapConnectWebviewPreloader.esbuild.js`

chabroA (commented):
Nice! Let's get rid of all this dead code! 👍 

Just make sure not to forget some dust behind 🧹 

I saw that this whole cleanup was split in 3 parts (LLD, LLM, LLC).

Can make sense / feel safer to split it (maybe easier to test), but here is how I would go about it:
- have a "feature branch" for this cleanup, that will hold the final cleanup before merging it to `develop`
- proceed to cleanup step by step (LLD, LLM, LLC) separately as you see fit, starting with the edges (LLD and LLM), as you started to do with this PR
- you can merge LLD and LLM work in the "feature branch" independently when each part seems done
- once both LLD and LLM "seems done" proceed with cleaning up the whole logic in LLC (as part of this, you will probably find code in LLD and LLM that you left in the previous steps now irrelevant, fell free to delete it as part of this step, even if it's not part of LLC)
- merge the LLC PR in feature branch (including probable dust from LLD and LLM that you removed there as well)
- you can now fully QA / non reg this feature branch including the whole cleanup

This will avoid leaving dust behind and having develop in a "half cleaned" state with some dead code still present in between each steps. This will also reduce QA burden since you only have to QA / non reg deeply once the feature branch, and can "light QA" the LLD and LLM PRs along the way, since you have a safety net with the feature branch because these PRs are not merged directly to `develop`.

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/SwapConnectWidget/index.tsx
ggilchrist-ledger: Yep got it noted 👍

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/FormSummary/index.tsx
ggilchrist-ledger: It sounds like keeping the useEffect might be safer but keeping it is potentially redundant. I will try the first suggestion and test it to see if there are any obvious issues

ggilchrist-ledger: I've tested both keeping just `hasRates` and keeping the useEffect that responds to the rates changes and both seem to be the exact same behaviour. It's just controlling the account selection and fee selection buttons and the main swap form is the main control whether it displays or not, so I think having the simplest solution (`hasRates`) is fine

PR #4199, opened 2023-08-02
[Feat/LIVE-7863]: Swap Form - Edit Rates tracking updates

kallen-ledger:
### 📝 Description

- Add `Swap Form - Edit Rates` tracking event to LLM.
- Add the following to this event on both LLD and LLM:

```ts
partnersList: string[];
exchangeRatesState: string[];
sourceCurrency: string;
targetCurrency: string;
```

### ❓ Context

- **Impacted projects**: `LLD, LLM` 
- **Linked resource(s)**: [LIVE-7863](https://ledgerhq.atlassian.net/browse/LIVE-7863) 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-7863]: https://ledgerhq.atlassian.net/browse/LIVE-7863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/Rates/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/Rates/index.tsx
@@ -22,8 +22,6 @@ import IconInfoCircle from "~/renderer/icons/InfoCircle";
 import { filterRates } from "./filterRates";
 import { getFeesUnit } from "@ledgerhq/live-common/account/index";
 import { formatCurrencyUnit } from "@ledgerhq/live-common/currencies/index";
-import BigNumber from "bignumber.js";
-import C from "~/renderer/modals/Receive/Body";
 
 type Props = {
   fromCurrency: SwapSelectorStateType["currency"];
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/Rates/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/Rates/index.tsx
@@ -20,6 +20,10 @@ import styled from "styled-components";
 import Tooltip from "~/renderer/components/Tooltip";
 import IconInfoCircle from "~/renderer/icons/InfoCircle";
 import { filterRates } from "./filterRates";
+import { getFeesUnit } from "@ledgerhq/live-common/account/index";
+import { formatCurrencyUnit } from "@ledgerhq/live-common/currencies/index";
+import BigNumber from "bignumber.js";
+import C from "~/renderer/modals/Receive/Body";
 
 type Props = {
   fromCurrency: SwapSelectorStateType["currency"];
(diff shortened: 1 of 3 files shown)


apps/ledger-live-mobile/src/screens/Swap/Form/index.tsx
```diff
@@ -148,13 +151,58 @@ export function SwapForm({
...
+        ...sharedSwapTracking,
+        provider,
+        partnersList,
+        exchangeRateList,
+        sourceCurrency: swapTransaction.swap.from.currency?.id,
+        targetCurrency: swapTransaction.swap.from.currency?.id,
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pageState, editRatesTrackingProps]);
```
sarneijim, line 204: What is the benefit of stringify the track object vs add here directly dependencies?

kallen-ledger: By stringifying it here we are guaranteeing that the useEffect will only be rerun if there is a change in the editRatesTrackingProps. If we were to pass an object here the useEffect would re-run on each new creation of that object even if all the keys and values are the same.

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/Rates/index.tsx
```diff
@@ -60,7 +62,11 @@ export default function ProviderRate({
   const [defaultPartner, setDefaultPartner] = useState<string | null>(null);
   const selectedRate = useSelector(rateSelector);
   const filteredRates = useMemo(() => filterRates(rates, filter), [rates, filter]);
-
+  const providers = rates?.map(rate => rate.provider) ?? [];
```
sarneijim, line 65: 2 rates can have the same provider (float and fix)

apps/ledger-live-mobile/src/screens/Swap/Form/index.tsx
sarneijim: Are you sure if it running twice because of this and not because we are in dev mode?

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/Rates/index.tsx
kallen-ledger: Thank you, will get unique providers here 🙌

kallen-ledger: Sorry, forgot to push these changes yesterday. Updated now.

cng-ledger (approved):

PR #4206, opened 2023-08-03
[bugfix] Polygon zk evm feature flag

haammar-ledger:
### 📝 Description

Feature flag wasn't correctly defined for Polygon Zk EVM and Polygon Zk EVM testnet, causing them to always be available even with feature flag off.

### ❓ Context

- **Impacted projects**: `LLD` `LLM` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8844 

### ✅ Checklist

- [ ] **Test coverage** :x: No framework to auto-test feature flags, to my knowledge
- [x] **Atomic delivery** 
- [x] **No breaking changes**

### DIFF
--- a/.changeset/eight-rockets-behave.md
+++ b/.changeset/eight-rockets-behave.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
+Fix feature flag for Polygon Zk EVM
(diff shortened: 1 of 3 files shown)


hedi-edelbloute (approved):
TIL about this requirement

Idk how hard it would be to add a test that checks that each key is also found in currencies.ts

lgtm :+1:

PR #4279, opened 2023-08-10
[bugfix] Secret Network feature flag

haammar-ledger:
### 📝 Description

Feature flag wasn't correctly defined for Secret Network, causing it to always be available even with feature flag off.

### ❓ Context

- **Impacted projects**: `LLD` `LLM` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8916 

### ✅ Checklist

- [x] **Test coverage**
- [x] **Atomic delivery** 
- [x] **No breaking changes**

### DIFF
--- a/apps/ledger-live-mobile/src/screens/AddAccounts/01-SelectCrypto.tsx
+++ b/apps/ledger-live-mobile/src/screens/AddAccounts/01-SelectCrypto.tsx
@@ -89,7 +89,7 @@ export default function AddAccountsSelectCrypto({ navigation, route }: Props) {
   const klaytn = useFeature("currencyKlaytn");
 
   const featureFlaggedCurrencies = useMemo(
-    (): Partial<Record<CryptoCurrencyId, Feature<any> | null>> => ({
+    (): Partial<Record<CryptoCurrencyId, Feature<unknown> | null>> => ({
       axelar,
       stargaze,
       umee,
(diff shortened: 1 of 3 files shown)


hedi-edelbloute (approved):

hzheng-ledger (approved):

apps/ledger-live-mobile/src/screens/AddAccounts/01-SelectCrypto.tsx
```diff
@@ -94,7 +94,7 @@ export default function AddAccountsSelectCrypto({ navigation, route }: Props) {
       stargaze,
       umee,
       desmos,
-      secretNetwork,
+      secret_network: secretNetwork,
```
gre, line 97: it seems that we have just suffered from poor string typing here, could would add this?
it seems not easy to provide a non reg test, but adding a type will be that way to non regress this.
Screenshot 2023-08-10 at 10 59 18

gre (changes requested):
bugfix needs to offer a non regression test. for this fix, you can just do a typecheck that will be that non regression test

apps/ledger-live-mobile/src/screens/AddAccounts/01-SelectCrypto.tsx
gre: by doing so, you can also remove the 2 comments that become useless since we prove by typing this "pay attention to the case" 👍

gre: as side note, it could also be a good idea to move this logic out of the UI side, 01-SelectCrypto.tsx is a generic screen, what if another screen need this, 
not even talking about the non modular aspect of this, but at least this should be moved in a hooks/ helper maybe? 
`function useFeatureFlaggedCurrencies`

gre (approved):

PR #4281, opened 2023-08-10
Fix: LIVE-8269 cosmos amount string parser

hzheng-ledger:
### 📝 Description
In the response of cosmos events from explorer. The "amount" field string can contain prefix. Our parser should be able to remove this prefix and get the correct amount string. Otherwise, it will cause a bug in the portfolio graph because of the incorrect amount.

It is a problem due to the multi chain rewards update of cosmos. Before the update, we didn't have this prefix as the follows:

```
{
                  "key": "amount",
                  "value": "11717uosmo"
},

```
Now we can have prefix of the amount value since multi chain rewards update
```
{
                  "key": "amount",
                  "value": "56ibc/0025F8A87464A471E66B234C4F93AEC5B4DA3D42D7986451A059273426290DD5,512ibc/6B8A3F5C2AD51CD6171FA41A7E8C35AD594AB69226438DB94450436EA57B3A89,11717uosmo"
},

```
we should remove the prefix to get the correct value

### ❓ Context

- **Impacted projects**: LLC
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8269

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/families/cosmos/logic.ts
+++ b/libs/ledger-live-common/src/families/cosmos/logic.ts
@@ -190,3 +190,7 @@ export function getRedelegationCompletionDate(
   const currentRedelegation = getRedelegation(account, delegation);
   return currentRedelegation ? currentRedelegation.completionDate : null;
 }
+
+export function parseAmountStringToNumber(amountString: string, unitCode): string {
+  return amountString.slice(amountString.lastIndexOf(",") + 1).replace(unitCode, "");
+}
--- a/libs/ledger-live-common/src/families/cosmos/logic.ts
+++ b/libs/ledger-live-common/src/families/cosmos/logic.ts
@@ -191,6 +191,6 @@ export function getRedelegationCompletionDate(
   return currentRedelegation ? currentRedelegation.completionDate : null;
 }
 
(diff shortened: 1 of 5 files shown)


libs/ledger-live-common/src/families/cosmos/logic.ts
```diff
@@ -190,3 +190,7 @@ export function getRedelegationCompletionDate(
   const currentRedelegation = getRedelegation(account, delegation);
   return currentRedelegation ? currentRedelegation.completionDate : null;
 }
+
+export function parseAmountStringToNumber(amountString: string, unitCode): string {
```
lvndry, line 194:
```suggestion
export function parseAmountStringToNumber(amountString: string, unitCode: string): string {
```

hzheng-ledger: fixed

haammar-ledger (changes requested):
Just needs a changeset. Thanks :)

hzheng-ledger:
changeset added

haammar-ledger (approved):

PR #4307, opened 2023-08-10
chore: add resiliency on importing accounts that fail to be decoded

gre:
### 📝 Description

This add resiliency on importing accounts that would fail to be decoded in the future on both LLM and LLD.

There are a few possible reasons why an old account could be failed to be imported. For instance, if one of the field including `currencyId`, `derivationMode`, ...other field that have strict assertion (on the `fromAccountRaw` call) are bad (altered, modified, removed).

Some real world examples:
- a derivationMode becomes obsolete and is removed
- a currency becomes obsolete and is removed (example: Ethereum Ropsten)
- one of these field is altered by the user somehow (app.json modified / corrupted)
- exporting/importing an app.json manually from the past / through various backup services

**The previous behavior:**

One bad account data and it was badly crashing the app at boot.

**The new behavior:**

The problematic account that fails to import is evicted from the list of accounts. So, if a currencyId (like ethereum ropsten) becomes obsolete, it will be transparent for the user and account will disappear without a crash at boot.

**Note that a report is also sent to Sentry (if enabled) so we still can have visibility over these failures** because this acts more as a failsafe (resiliency) than a behavior we want to regularly leverage, removing currency is not something we will do often.

### ❓ Context

- **Impacted projects**: `LLD/LLM` 
- **Linked resource(s)**: N/A yet 

### ✅ Checklist

- [x] **Test coverage** tests both added for LLM and LLD. the app no longer crashes and correctly land on empty portfolio when all data was altered with unsupported currency id data. 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/rude-years-flow.md
+++ b/.changeset/rude-years-flow.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
+Implement resiliency on importing accounts that fail to be decoded (bad currencyId, bad derivationMode,...)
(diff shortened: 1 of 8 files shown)


haammar-ledger (approved):
LGTM

PR #4322, opened 2023-08-11
feat: changing locale prop on the earn app to use the language instead

ggilchrist-ledger:
### 📝 Description

LLM doesn't have locales like LLD does so we need to use the `language` to determine the formatting for LLM as before. This PR sets the `language` prop as `locale` for Earn in LLM so we don't need to have specific logic in Earn for LLM and LLD regarding the currency formatting. It also means in future if locales are supported in LLM, it just means changing `locale` prop in the Earn web player from `language` to `locale`

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach


kallen-ledger (approved):
Nice 👍

beths-ledger (approved):

PR #4339, opened 2023-08-11
[bugfix] Avoid conflicts in currency feature flag names in manager

haammar-ledger:
### 📝 Description

If a feature flag name for a currency contains another existing currency (e.g `feature_currency_polygon_zk_evm` containing `polygon`) then the two currencies will conflict in the manager, causing https://ledgerhq.atlassian.net/browse/LIVE-8884

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8884 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes**

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
@@ -86,7 +86,7 @@ const Item = ({
 
   const flag = currencyId
     ? Object.entries(currencyFlags || {}).find(([flagName]) => {
-        return flagName.includes(currencyId);
+        return flagName.replace("feature_currency_", "") === currencyId;
       })
     : undefined;
 
--- a/apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
@@ -1,7 +1,8 @@
 import React, { useMemo, memo, useCallback } from "react";
+import { camelCase } from "lodash";
 import { useNotEnoughMemoryToInstall } from "@ledgerhq/live-common/apps/react";
 import { getCryptoCurrencyById, isCurrencySupported } from "@ledgerhq/live-common/currencies/index";
-import { App, Feature } from "@ledgerhq/types-live";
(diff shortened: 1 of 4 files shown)


apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx
```diff
@@ -86,7 +86,7 @@ const Item = ({
 
   const flag = currencyId
     ? Object.entries(currencyFlags || {}).find(([flagName]) => {
-        return flagName.includes(currencyId);
+        return flagName.replace("feature_currency_", "") === currencyId;
```
ofreyssinet-ledger, line 89: Then you might as well directly use
```ts
useFeature(camelCase(`currency_${currencyId}`))
```
I have a suspicion that it might fix [your other issue](https://ledgerhq.atlassian.net/browse/LIVE-8489).

The problem might be with `getAllFeatures`, if you check how it's defined in `FirebaseRemoteConfigProvider` it captures the value of `remoteConfig`, which is only set once at app start, so `getAllFeatures` might be returning only the values that the feature flags had at that first load.
It also completely ignores the local overrides.

ofreyssinet-ledger: If this is right we should either get rid of getAllFeatures or fix in both apps it so that it is aligned with the way the rest of the feature flags work in LL (support for overrides and also use the same flag name formatting without feature_ and in camelCase).

gre: i'm not exactly sure if we can go with @ofreyssinet-ledger 's suggestion, but the current change is indeed no super performant (the current approach to have to "find" through entries isn't performant already, and it makes it worse. and we do this for each item of the manager list without any caching :o

haammar-ledger: @ofreyssinet-ledger Thanks for the suggestion, I used it and it works. However I don't want to refactor `getAllFlags` in this PR, neither create a full mapping between `FeatureId` and `Currency.id` because this will be reworked properly when we work on our Coin Config feature.

KVNLS (approved):

PR #4353, opened 2023-08-15
Support/LIVE-7931 Evm mock test

hzheng-ledger:
### 📝 Description

Add evm mock tests.

### ❓ Context

- **Impacted projects**: `` LLC, LLM, LLD
- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-7931

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #4376, opened 2023-08-17
Cardano Bridge Mock testing

lvndry:
### 📝 Description

Add Cardano Bridge Mock to make it easier to do E2E tests

### ❓ Context

- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-7932` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-desktop/src/renderer/modals/AddAccounts/steps/StepChooseCurrency.tsx
```diff
@@ -258,7 +277,12 @@ export const StepChooseCurrencyFooter = ({
         <Button
           primary
           disabled={!currency || fullNodeNotReady || !navigator.onLine}
-          onClick={() => transitionTo("connectDevice")}
+          onClick={() => {
+            if (mock) {
```
hzheng-ledger, line 281: remove this piece of code according to the discussion https://ledger.slack.com/archives/C0308JT8WS2/p1692373876296589?thread_ts=1691507827.835819&cid=C0308JT8WS2

apps/ledger-live-desktop/tests/userdata/1AccountADA.json
```diff
@@ -0,0 +1,221 @@
+{
```
hzheng-ledger, line 1: I think this file is not used. It can be removed

hzheng-ledger (changes requested):
See my comments. Thanks a lot

libs/env/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "@types/jest": "^29.2.4",
+    "@types/lodash": "^4.14.197",
```
haammar-ledger, line 29: Is this required?

libs/ledger-live-common/src/mock/account.ts
```diff
@@ -108,6 +109,199 @@ export function genAccount(id: number | string, opts: GenAccountOptions = {}): A
...
+            protocolParams: {
+              minFeeA: "",
+              minFeeB: "",
+              stakeKeyDeposit: "",
+              lovelacePerUtxoWord: "",
+              collateralPercent: "",
+              priceMem: "",
+              priceSteps: "",
+              languageView: {
+                PlutusScriptV1: {
```
haammar-ledger, line 133: Ouch :sweat_smile: so there was no way to use an empty object here? I think all this stuff is only useful is you need to actually sign and broadcast a valid tx.

But anyway now that it's done no problem.

apps/ledger-live-desktop/src/renderer/modals/AddAccounts/steps/StepChooseCurrency.tsx
```diff
@@ -63,6 +66,7 @@ const StepChooseCurrency = ({ currency, setCurrency }: StepProps) => {
   const base = useFeature("currencyBase");
   const baseGoerli = useFeature("currencyBaseGoerli");
   const klaytn = useFeature("currencyKlaytn");
+  const mock = useEnv("MOCK");
```
haammar-ledger, line 69: It seems we want to limit as much as possible the use of the env var directly in the app's code, see https://ledger.slack.com/archives/C0308JT8WS2/p1692373491532989?thread_ts=1691507827.835819&cid=C0308JT8WS2

But there's a debug tool available in the app to set it anyway.

apps/ledger-live-desktop/src/renderer/modals/AddAccounts/steps/StepChooseCurrency.tsx
```diff
@@ -136,19 +140,26 @@ const StepChooseCurrency = ({ currency, setCurrency }: StepProps) => {
...
-    const deactivatedCurrencies = Object.entries(featureFlaggedCurrencies)
-      .filter(([, feature]) => !feature?.enabled)
-      .map(([name]) => name);
+
+    const deactivatedCurrencies = mock
+      ? []
+      : Object.entries(featureFlaggedCurrencies)
+          .filter(([, feature]) => !feature?.enabled)
+          .map(([name]) => name);
+
```
haammar-ledger, line 154: I don't think we need to ignore feature flagged coins, as they just need to be enabled for the same environment that playwright is running on, which should be only dev or staging (never runs on prod builds I guess), so it just means we need to have the feature flag enabled on dev or staging envs, which should be the case anyway before we want to run the UI tests for that coin.

Saying all that just so we can avoid using the `MOCK` env var from within the code :slightly_smiling_face: it seems we want to limit this as much as possible, see discussion here https://ledger.slack.com/archives/C0308JT8WS2/p1692373491532989?thread_ts=1691507827.835819&cid=C0308JT8WS2

libs/ledger-live-common/src/families/cardano/bridge/mock.ts
```diff
@@ -0,0 +1,169 @@
...
+    errors,
+    warnings,
+    estimatedFees,
+    amount,
+    totalSpent,
+  });
+};
+
+const prepareTransaction = async (account: Account, transaction: CardanoTransaction) => {
+  return transaction;
```
haammar-ledger, line 145: should maybe just set the `fees` field?

hzheng-ledger (approved):
looks good to me

apps/ledger-live-desktop/src/renderer/modals/AddAccounts/steps/StepChooseCurrency.tsx
lvndry: Tbf I just took it from this PR because I didn't want to deal with merge conflicts later: https://github.com/LedgerHQ/ledger-live/pull/4353/files#diff-e05343023abba025ba4898ebc9f04f698c171b29d3efdb2fcbb07c95545bf98e

I can remove though. Is it okay @hzheng-ledger ?

hzheng-ledger: @haammar-ledger What you are saying is correct. We can remove this piece of code in theory.
I've just checked the firebase configuration. It seems that in the project "ledger-live-development" all the currencies in dev environment are enable.
but when we launch the playwright test, it uses the configuration of "ledger-live-testing" project, all the currencies flags are missing there. Therefore, it takes the default value which is false and the evm coins are not enabled. It causes evm playwright test case failure.
So I suggest we add the currencies feature flags in "ledger-live-testing" project firstly and remove this piece of code.

haammar-ledger (approved):

PR #4390, opened 2023-08-18
fix: don't pop from live app when verifying address

kallen-ledger:
### 📝 Description

Verifying addresses within a live app would cause a redirection out of the live app for users. In cases where the user was already in a live app like "buy/sell" and was then directed to a provider like "BTCDirect" then the user would be redirected back to the buy/sell app and they were unable to continue with their purchase.

In cases where a user loaded from the discover section they would be redirected back to the discover homepage after they verified their address.

The fix:
- check if we are about to pop from a live app screen. If we are then do not pop, otherwise continue as normal.

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-8928 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/screens/VerifyAccount/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/VerifyAccount/index.tsx
@@ -49,7 +49,19 @@ export default function VerifyAccount({ navigation, route }: NavigationProps) {
   const onDone = useCallback(() => {
     const n = navigation.getParent<StackNavigatorNavigation<RootStackParamList>>();
 
-    if (n) {
+    // get the route at the given index on navigation
+    const { index, routes } = navigation.getState();
+    const { name: routeAtIndexName } = routes[index];
+
+    // if this route is a live app we do not want to pop
+    // as doing so would take the user out of the live app.
+    // an example of this is the BTCDirect app.
+    const isRouteAtIndexALiveApp = [ScreenName.Exchange, ScreenName.PlatformApp].includes(
+      routeAtIndexName as ScreenName,
+    );
+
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/screens/VerifyAccount/index.tsx
```diff
@@ -49,7 +49,19 @@ export default function VerifyAccount({ navigation, route }: NavigationProps) {
...
 
-    if (n) {
+    // get the route at the given index on navigation
+    const { index, routes } = navigation.getState();
+    const { name: routeAtIndexName } = routes[index];
+
+    // if this route is a live app we do not want to pop
+    // as doing so would take the user out of the live app.
+    // an example of this is the BTCDirect app.
+    const isRouteAtIndexALiveApp = [ScreenName.Exchange, ScreenName.PlatformApp].includes(
```
sarneijim, line 59: If you want get only live apps I think this filter could be incorrect. ScreenName.Exchange include sell and swap. 

image

kallen-ledger: Oh ok, I wasn't aware that it also included swap, so will have to double check that!

Yes, will update the comment also and try to better explain it.

I don't quite understand what you mean by this sentence 😄 
"On the other hand, do we need really filter "buy" app here?"

sarneijim: Confirmed. After navigate to BTC Direct we continue in Buy context.

PR #4391, opened 2023-08-18
fix: typings on live-env

valpinkman:
### 📝 Description

Fix typings on live-env so we cannot get or set unknown env names

### ❓ Context

- **Impacted projects**: `live-env` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/screens/Settings/Experimental/FeatureRow.tsx
```diff
@@ -44,7 +44,7 @@ const FeatureRow = ({ feature }: Props) => {
         onChange={setEnvUnsafe}
         isDefault={isEnvDefault(feature.name) || getEnv(feature.name) === undefined}
         {...rest}
-        value={getEnv(feature.name)}
+        value={getEnv(feature.name) as number}
```
haammar-ledger, line 47: same here

libs/ledger-live-common/src/env.test.ts
```diff
@@ -0,0 +1,8 @@
+import { getEnv } from "@ledgerhq/live-env";
+
+// @ts-expect-error sanity check on getEnv() not accepting incorrect key names
+getEnv("yolo");
+
+test("typecheck env", () => {
+  expect(true).toBe(true);
+});
```
haammar-ledger, line 8: I don't get what these are testing?

haammar-ledger (commented):
:100: for updating the import paths to the lib directly, please now kill [the old env.ts](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/env.ts) !

.prototools
```diff
@@ -1,3 +1,3 @@
-node = "18"
-npm = "9"
-pnpm = "8.6"
+node = "18.17.1"
+npm = "9.8.1"
+pnpm = "8.6.12"
```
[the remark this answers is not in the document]
valpinkman: it should.... but i updated `proto` on my side and now it cannot resolve uncomplete version... so i had to bump....

apps/ledger-live-desktop/src/renderer/screens/settings/sections/Experimental/index.tsx
```diff
@@ -48,11 +48,11 @@ const ExperimentalFeatureRow = ({
   return Children ? (
     <Row title={feature.title} desc={feature.description}>
       <Children
-        value={envValue}
         readOnly={isReadOnlyEnv(feature.name)}
         isDefault={isDefault}
         onChange={onChange}
         {...rest}
+        value={envValue as number}
```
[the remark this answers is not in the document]
valpinkman: It's not blind. It's just that we have 3 components with different props. Two of them accept a `value` props and both of them use `number`

apps/ledger-live-mobile/src/screens/Settings/Experimental/FeatureRow.tsx
valpinkman: answered above ^

libs/ledger-live-common/src/env.test.ts
valpinkman: we are using `// @ts-expect-error` on the first line, which basically EXPECT typescript to throw an error. If, by some changes, we break the type inference on `getEnv("")` and `yolo` now works, the comment itself will throw an error during typechking

gre:
vercel CI seems to fail to build?
```
live-common-tools:build: ./src/demos/repl/components/LiveEnvEditor.js
--
10:28:47.375 | live-common-tools:build: Cannot find module: '@ledgerhq/live-env'. Make sure this package is installed.
```

libs/ledger-live-common/src/env.test.ts
gre: I think you can instead move this file to `src/typescript-tests/env.ts`

the goal is to do a typescript test, this one:

```
// @ts-expect-error
getEnv("yolo");
```

to actually non regress on what we just fixed here ;)

sprohaszka-ledger (approved):
It would be great if we could have something that cast the value in its correct type when calling `getEnv`.

I don't understand your comment about LLC dead code as you remove `env.ts` from LLC 😅

valpinkman:
> It would be great if we could have something that cast the value in its correct type when calling `getEnv`.
> 
> I don't understand your comment about LLC dead code as you remove `env.ts` from LLC 😅

I removed it after Hakim suggested it, so do not take it into consideration ^^

For the correct casting type, I believe we should restart the typings from scratch cause some stuff are way too complicated and I think we can definitely type it better than that

PR #3827, opened 2023-06-27
[LIVE-6263] Feature - Ledger infrastructure abstraction for the EVM Family

lambertkevin:
### 📝 Description

Add support for Ledger infrastructure inside the EVM family. This will allow to have fully configurable currencies for:
1. RPC node
2. Explorer (not mandatory)
3. Gas Tracker (not mandatory)

Any combination of those 3 services should be functional.

### ❓ Context

- **Impacted projects**: `@ledgerhq/live-common` `@ledgerhq/coin-evm` `@ledgerhq/coin-framework` `@ledgerhq/cryptoassets` `@ledgerhq/types-cryptoassets` `ledger-live-desktop` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-6263 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/44363395/a5fba28d-94f5-4442-af07-c3d4ad14995a

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

#### `coin-evm` code coverage:
Screenshot 2023-07-11 at 13 59 01

PR #4446, opened 2023-08-23
[BUGFIX][LIVE-9035] fix unexpected gas tracker error

chabroA:
### 📝 Description
When EIP-1559 is not supported by a currency but the useEIP1559 option is provided as true, gracefully fallback to legacy option instead of throwing error.

[This PR](https://github.com/LedgerHQ/ledger-live/pull/3827) about ledger infrastructure abstraction updated the get estimation fees logic when relying on ledger infrastructure, calling the `getGasOptions` function with `useEIP1559` option set to `true` if the transaction is of type `2`. cf. [here](https://github.com/LedgerHQ/ledger-live/blob/c14fb81e82f6e0e3d1d6d335f5b72604a4c5fa34/libs/coin-evm/src/api/node/ledger.ts#L216-L225).

But in the `prepareTransaction` logic, this call is done before we have narrowed down the transaction type, by calling `getTypedTransaction` (cf. [here](https://github.com/LedgerHQ/ledger-live/blob/c14fb81e82f6e0e3d1d6d335f5b72604a4c5fa34/libs/coin-evm/src/prepareTransaction.ts#L201-L205)). Since a new evm transaction is of type 2 by default (cf. [here](https://github.com/LedgerHQ/ledger-live/blob/c14fb81e82f6e0e3d1d6d335f5b72604a4c5fa34/libs/coin-evm/src/createTransaction.ts#L23)), this results in an error for all evm currencies that don't support eip1559.

This is solved by gracefully falling back to the "legacy" option (i.e: not using eip1559) in the `getGasOptions` function, instead of throwing error.

Therefore, when requesting an estimation of fees on the network (through `getFeeData` in `libs/coin-evm/src/api/node/ledger.ts`), we get the appropriate estimation, without error, whether the currency supports eip1559 or not.

Lastly, it's better to gracefully fallback than crash with an error when possible, which is the case here.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach


libs/coin-evm/src/api/gasTracker/ledger.ts
```diff
@@ -49,13 +45,8 @@ export const getGasOptions = async ({
...
 
-  const { useEIP1559 = false } = options || /* istanbul ignore next */ {};
-
-  if (useEIP1559 && !gasTrackerConfig.compatibilty.eip1559) {
-    throw new GasTrackerDoesNotSupportEIP1559(
-      `Gas tracker does not support EIP1559 for ${currency.id}`,
-    );
-  }
+  // We use the eip1559 display parameter only if requested AND the currency supports it
+  const useEIP1559 = (options?.useEIP1559 ?? false) && gasTrackerConfig.compatibilty.eip1559;
```
haammar-ledger, line 49: Why not:
```suggestion
  const useEIP1559 = options?.useEIP1559 && gasTrackerConfig.compatibilty.eip1559;
```

haammar-ledger (approved):

PR #4453, opened 2023-08-23
Bugfix/live 8464 llm contexts improvements

ofreyssinet-ledger:
### 📝 Description

Small refactorings focused on killing unused contexts:
- Removed `DelayedTrackingProvider`: dead code. Only way this was useful is if `pushDelayedTrackingEvent` had been used anywhere once, it has never been actually used.
- `AdjustProvider` was not actually a provider so it was renamed into `AdjustSetup` to make identifying actual providers easier.
- `AnalyticsProvider` was using segment's AnalyticsProvider but that one was not actually of any use in our codebase, so it was killed. Then this component is no longer a provider -> renamed it `SegmentSetup` as it's just used to trigger the initialisation of the segment SDK.
  `SegmentAnalyticsProvider` is not actually used at all because we never use anywhere the `useAnalytics` function from `"@segment/analytics-react-native"`.
  Then what this component does is this:
  | `loaded` | `segmentClient` | result                               |
  |--------|---------------|--------------------------------------|
  | truthy | truthy        | `return <SegmentAnalyticsProvider {segmentClient}>{children}</SegmentAnalyticsProvider>`</br>and if you check the source of `SegmentAnalyticsProvider`,</br>if `segmentClient` is falsy it returns `null` (doesn't happen in this row), otherwise it returns its children wrapped in a provider which we currently have no use of |
  | truthy | falsy         | `return <>{children}</>`              |
  | falsy  | truthy        | `return <>{children}</>`              |
  | falsy  | falsy         | `return <>{children}</>`               |

  So in terms of rendering this component is equivalent to a component that just always renders its children (wrapped in a provider we don't use).
  So we might as well make it a component that returns null and not pass it children.
- LedgerStore had an argument `store` in its `children` render function prop. This argument is unnecessary as the store can be obtained simply with useStore() (which is actually the point of the react-redux Provider rendered by that `LedgerStore` component)
- Renamed `ButtonUseTouchable` into `ButtonUseTouchableContext` as it's a react context
- Terms and conditions context provider:
  - all the data necessary lives in the redux store so the context is unnecessary, so it was killed.
  - some hooks were defined in order to still have some helpers necessary to easily get the current state and update it.
  - the old component that performs migration from some legacy storage to the new way of storage (`settings` part of the store) was kept as a childless component that just needs to be mounted at the root of the app.
- `AnalyticsContext`: the state passed to the provider was defined inside the RootNavigator component, which would cause re-renders of this component even though no part of the component makes use of that state. So the `AnalyticsContext.Provider` was moved into a standalone component `AnalyticsContextProvider` (in `AnalyticsContext.tsx`) that has its own internal state.

### ❓ Context

- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/components/DelayedTrackingProvider.tsx
ofreyssinet-ledger, line 1: dead code: only way this was useful is if `pushDelayedTrackingEvent` had been used anywhere once, it has never been actually used

apps/ledger-live-mobile/src/analytics/SegmentSetup.tsx
ofreyssinet-ledger, line 1: this is the renamed `AnalyticsProvider`.
cf. comment on `AnalyticsProvider` above for explanation

apps/ledger-live-mobile/src/components/AdjustSetup.tsx
```diff
@@ -10,7 +10,7 @@ import Config from "react-native-config";
 import { useSelector } from "react-redux";
 import { analyticsEnabledSelector } from "../reducers/settings";
 
-export default function AdjustProvider() {
+export default function AdjustSetup() {
```
ofreyssinet-ledger, line 13: renamed because we should probably keep "Provider" for context providers

apps/ledger-live-mobile/src/context/LedgerStore.tsx
```diff
@@ -121,6 +117,6 @@ export default class LedgerStoreProvider extends Component<
   render() {
     const { children } = this.props;
     const { ready, initialCountervalues } = this.state;
-    return <Provider store={store}>{children(ready, store, initialCountervalues)}</Provider>;
```
ofreyssinet-ledger, line 124: `store` can be obtained with `useStore` so it's longer needed in the component using this render function (cf. `src/index.tsx`)

apps/ledger-live-mobile/src/index.tsx
```diff
@@ -231,9 +230,7 @@ const StylesProvider = ({ children }: { children: React.ReactNode }) => {
 
   return (
     <StyleProvider selectedPalette={resolvedTheme}>
-      <GeneralTermsContextProvider>
```
ofreyssinet-ledger, line 234: this provider was killed because all the data needed live in the redux store (cf. changes in `logic/terms.tsx`)

apps/ledger-live-mobile/e2e/bridge/client.ts
```diff
@@ -9,8 +9,7 @@ import { ConnectManagerEvent } from "@ledgerhq/live-common/hw/connectManager";
 import { store } from "../../src/context/LedgerStore";
 import { importSettings } from "../../src/actions/settings";
 import { importStore as importAccounts } from "../../src/actions/accounts";
-import { acceptGeneralTermsLastVersion } from "../../src/logic/terms";
-import accountModel from "../../src/logic/accountModel";
```
ofreyssinet-ledger, line 13: unused var

apps/ledger-live-mobile/src/analytics/SegmentSetup.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { start } from "./segment";
+import type { AppStore } from "../reducers";
+import { useStore } from "react-redux";
+
+const SegmentSetup = (): null => {
+  const store = useStore();
+  const loadSegment = useCallback(async (store: AppStore) => {
+    try {
+      await start(store);
```
sshmaxime, line 11: Why don't we use `hooks` instead of functions ? I've seen that it was already like that before but just asking about the rationale behind that.

sshmaxime (approved):

PR #4477, opened 2023-08-25
[BUGFIX][LIVE-9028] fix fees rounding issue

chabroA:
### 📝 Description

<details><summary>The issue 📸  </summary>

IMG_6718-20230821-154934 Large

</details> 

> Since our use of BigNumber implies only using integers, we need to round up to integer value with `integerValue()` after doing some math (like multiplying with a possible float value in this case)

Make sure bignumber fee values are integer to prevent rounding error between what is displayed on LL screens and what is displayed on nano device screen. Especially when interacting with `EIP1559_BASE_FEE_MULTIPLIER` that can be a float value.
Similar to what was done here https://github.com/LedgerHQ/ledger-live/blob/305f97b/libs/ledger-live-common/src/families/ethereum/react.ts
Could be removed once we remove the use of `EIP1559_BASE_FEE_MULTIPLIER` with https://ledgerhq.atlassian.net/browse/LIVE-7356

Add [`jest-expect-message` ](https://www.npmjs.com/package/jest-expect-message) to `coin-evm` to improve readability of tests involving for loops (like the added tests) in case of error.

Also (re)enabled the `disableRounding` prop for `FeesField` (estimated fees) on LLD tx confirm screen otherwise the fees are not similar between what is shown on LLD and on the nano device, because the value is not rounded on the nano device.
Not sure why rounding for this filed was enabled on LLD. It is not on LLM and makes the whole flow not very consistent, since fees are not rounded in the "Fees" or "Summary" steps, but are rounded here at the "Device confirmation" step, whereas they are not on the actual nano device either.

### ❓ Context

- **Impacted projects**: `LLD`, `LLM`, `coin-evm`
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9028

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

| Before | After |
|--------|--------|
| before | after | 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-evm/src/__tests__/unit/api/gasTracker/ledger.unit.test.ts
+++ b/libs/coin-evm/src/__tests__/unit/api/gasTracker/ledger.unit.test.ts
@@ -1,13 +1,14 @@
-import BigNumber from "bignumber.js";
-import { AssertionError } from "assert";
 import { getEnv, setEnv } from "@ledgerhq/live-env";
 import network from "@ledgerhq/live-network/network";
 import { CryptoCurrency, CryptoCurrencyId } from "@ledgerhq/types-cryptoassets";
+import { AssertionError } from "assert";
+import BigNumber from "bignumber.js";
 import { getGasOptions } from "../../../../api/gasTracker/ledger";
-import { GasOptions } from "../../../../types";
 import { LedgerGasTrackerUsedIncorrectly, NoGasTrackerFound } from "../../../../errors";
+import { GasOptions, Strategy } from "../../../../types";
 
 jest.mock("@ledgerhq/live-network/network");
+const mockedNetwork = jest.mocked(network);
 
 const fakeCurrency: Partial<CryptoCurrency> = {
   id: "ethereum" as CryptoCurrencyId,
@@ -25,19 +26,21 @@ const fakeCurrency: Partial<CryptoCurrency> = {
   units: [{ code: "ETH", name: "ETH", magnitude: 18 }],
 };
 
+const TEST_EIP1559_BASE_FEE_MULTIPLIER = 2;
+
 describe("EVM Family", () => {
   describe("api/gasTracker/index.ts", () => {
     const originalEIP1559_BASE_FEE_MULTIPLIER: number = getEnv("EIP1559_BASE_FEE_MULTIPLIER");
 
     beforeAll(() => {
-      setEnv("EIP1559_BASE_FEE_MULTIPLIER", 2);
+      setEnv("EIP1559_BASE_FEE_MULTIPLIER", TEST_EIP1559_BASE_FEE_MULTIPLIER);
     });
     afterAll(() => {
       setEnv("EIP1559_BASE_FEE_MULTIPLIER", originalEIP1559_BASE_FEE_MULTIPLIER);
     });
 
     beforeEach(() => {
-      const gastrackerBarometerMock = new Promise((resolve, _) => {
+      const gastrackerBarometerMock: any = new Promise((resolve, _) => {
         resolve({
           data: {
             low: "1",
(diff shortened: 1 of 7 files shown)


lvndry (commented):
Was it actually related to the merge ? It feels like it could be merged to develop and have its own changeset

lvndry (approved):

libs/coin-evm/src/__tests__/unit/api/gasTracker/ledger.unit.test.ts
```diff
@@ -53,100 +55,223 @@ describe("EVM Family", () => {
...
-            gasPrice: null,
-            nextBaseFee: new BigNumber(4),
-          },
-        };
+        it("should return integer values when EIP1559_BASE_FEE_MULTIPLIER is a float", async () => {
+          setEnv("EIP1559_BASE_FEE_MULTIPLIER", 1.5);
 
-        expect(gasOptions).toEqual(expectedGasOptions);
-      });
+          // @ts-expect-error method is mocked
```
hedi-edelbloute, line 94: Minor but I think you could use `const mockedNetwork = jest.mocked(network) to have right typings`

It's already used somewhere else and will fix all those ts expect error

libs/coin-evm/src/api/gasTracker/ledger.ts
```diff
@@ -54,30 +54,40 @@ export const getGasOptions = async ({
...
       slow: {
-        maxFeePerGas: next_base.times(EIP1559_BASE_FEE_MULTIPLIER).plus(low),
+        maxFeePerGas: next_base.times(EIP1559_BASE_FEE_MULTIPLIER).plus(low).integerValue(),
         maxPriorityFeePerGas: low,
         gasPrice: null,
         nextBaseFee: next_base,
       },
       medium: {
-        maxFeePerGas: next_base.times(EIP1559_BASE_FEE_MULTIPLIER).plus(medium),
+        maxFeePerGas: next_base.times(EIP1559_BASE_FEE_MULTIPLIER).plus(medium).integerValue(),
```
hedi-edelbloute, line 84: Aren't we already multiplying an integer at this point ? next_base is an integer right ?

hedi-edelbloute (approved):
Minor improvements

libs/coin-evm/src/api/gasTracker/ledger.ts
hedi-edelbloute: NVM "EIP1559_BASE_FEE_MULTIPLIER" could be a float value as you pointed

libs/coin-evm/src/api/gasTracker/ledger.ts
```diff
@@ -54,30 +54,40 @@ export const getGasOptions = async ({
...
+    low: new BigNumber(data.low).integerValue(),
+    medium: new BigNumber(data.medium).integerValue(),
+    high: new BigNumber(data.high).integerValue(),
+    next_base: new BigNumber(data.next_base).integerValue(),
   }));
 
   const EIP1559_BASE_FEE_MULTIPLIER: number = getEnv("EIP1559_BASE_FEE_MULTIPLIER");
 
+  /**
+   * Since our use of BigNumber implies only using integers, we need to round up to
```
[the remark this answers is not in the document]
gre: we do when we parse an amount from the user / format an amount to the user (so on the two extremes of the spectrum)
but most of the time, unless there are intermediary calculations (like done here, we can multiply/divide, in the realm of floating numbers) we will eventually need to go back to integers since we usually store everything on the atomic level (we can't divide a satoshi / a wei)

gre (approved):

PR #4510, opened 2023-08-29
Support/Live 8495 cosmos chains bot fix

hzheng-ledger:
### 📝 Description

1. Currently, We check the empty cosmos account by using the balance = 0 and operation array is empty. But it is not correct for the currencies that do support operation history since operation arrays are always empty for them. So we should use sequence = 0 to check whether an account is empty.
2. Ignore "nftoperations" field that is not used by cosmos family but breaks the bot tests.
3. Fix cosmos family bot, especially for the errors related to currencies that do support operation history.

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-framework/src/account/helpers.ts
+++ b/libs/coin-framework/src/account/helpers.ts
@@ -106,6 +106,17 @@ export const getAccountSpendableBalance = (account: AccountLike): BigNumber => {
 };
 
 export const isAccountEmpty = (a: AccountLike): boolean => {
+  // FIXME LIVE-5966 why do we need this? also this shouldn't be implemented here / this part must be removed back to the coin specifics
+  if (a.type === "Account" && a.currency.family === "tron") {
+    return (a as any).tronResources && (a as any).tronResources.bandwidth.freeLimit.eq(0);
+  }
+  if (a.type === "Account" && a.currency.family === "cosmos") {
+    return (
+      (a as any).cosmosResources &&
+      (a as any).cosmosResources.sequence === 0 &&
+      (a as any).balance.isZero()
+    );
+  }
   const hasSubAccounts = a.type === "Account" && a.subAccounts && a.subAccounts.length;
   return a.operationsCount === 0 && a.balance.isZero() && !hasSubAccounts;
 };
--- a/libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
+++ b/libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
@@ -178,8 +178,16 @@ export const getAccountShape: GetAccountShape = async info => {
     derivationMode,
   });
 
-  const { balances, blockHeight, txs, delegations, redelegations, unbondings, withdrawAddress } =
-    await new CosmosAPI(currency.id).getAccountInfo(address, currency);
+  const {
+    accountInfo,
+    balances,
+    blockHeight,
+    txs,
+    delegations,
+    redelegations,
+    unbondings,
+    withdrawAddress,
+  } = await new CosmosAPI(currency.id).getAccountInfo(address, currency);
   const oldOperations = initialAccount?.operations || [];
   const newOperations = txToOps(info, accountId, txs);
   const operations = mergeOps(oldOperations, newOperations);
@@ -221,6 +229,7 @@ export const getAccountShape: GetAccountShape = async info => {
       pendingRewardsBalance,
       unbondingBalance,
       withdrawAddress,
+      sequence: accountInfo.sequence,
     },
   };
 
--- a/.changeset/fifty-bats-leave.md
+++ b/.changeset/fifty-bats-leave.md
@@ -1,0 +1,6 @@
+---
+"@ledgerhq/live-common": patch
+"@ledgerhq/coin-framework": patch
+---
+
+fix cosmos family bot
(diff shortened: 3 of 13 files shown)


libs/ledger-live-common/src/families/cosmos/chain/Osmosis.ts
```diff
@@ -15,7 +14,6 @@ class Osmosis extends CosmosBase {
     this.unbondingPeriod = 14;
     this.stakingDocUrl =
       "https://support.ledger.com/hc/en-us/articles/6235986236957-Earn-Osmosis-OSMO-staking-rewards-in-Ledger-Live?docs=true";
-    this.minimalTransactionAmount = new BigNumber(10);
```
hedi-edelbloute, line 18: We don't want any for osmosis ?

libs/ledger-live-common/src/families/cosmos/chain/cosmosBase.ts
```diff
@@ -9,7 +9,8 @@ abstract class cosmosBase {
   abstract prefix: string;
   defaultGas = 100000;
   minGasPrice = 0.0025;
-  minimalTransactionAmount = new BigNumber(10000);
+  // minimalTransactionAmount is only used in bot tests. When we make a transaction ,we should make sure that the spendable balance is greater than minimalTransactionAmount. We usually use the upper limit of send transaction fee as the minimalTransactionAmount. e.g. 5000 uatom for cosmos.
```
hedi-edelbloute, line 12: Shouldn't this be an input of "generateGenericCosmosTest" method instead ? We don't want test logic in our app imo

In specs.ts we do :
 const cosmos = {
  ...generateGenericCosmosTest("cosmos"),
};

libs/coin-framework/src/account/helpers.ts
```diff
@@ -106,6 +106,17 @@ export const getAccountSpendableBalance = (account: AccountLike): BigNumber => {
 };
 
 export const isAccountEmpty = (a: AccountLike): boolean => {
+  // FIXME LIVE-5966 why do we need this? also this shouldn't be implemented here / this part must be removed back to the coin specifics
+  if (a.type === "Account" && a.currency.family === "tron") {
+    return (a as any).tronResources && (a as any).tronResources.bandwidth.freeLimit.eq(0);
+  }
+  if (a.type === "Account" && a.currency.family === "cosmos") {
```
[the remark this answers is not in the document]
haammar-ledger: Yes, this should stay in LLC for now, you can add Cosmos there along with Tron.
The part in coin-framework remains 100% generic, and the part in LLC is for "the rest" that we couldn't modularize correctly yet.

The long-term idea here is that `isAccountEmpty` should become a function in the `AccountBridge`, so every family can define its implementation.

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
```diff
@@ -221,6 +229,7 @@ export const getAccountShape: GetAccountShape = async info => {
       pendingRewardsBalance,
       unbondingBalance,
       withdrawAddress,
+      sequence: accountInfo?.sequence ?? 0,
```
haammar-ledger, line 232: I think a fallback to a 0 value isn't necessarily what we want, it may cause bugs.
Fortunately I don't think `accountInfo` or `sequence` can be undefined?
```suggestion
      sequence: accountInfo.sequence,
```

libs/coin-framework/src/account/helpers.ts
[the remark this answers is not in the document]
haammar-ledger: Following discussions, you can keep this here for now, since this function is directly used in places in coin-framework.

libs/ledger-live-common/src/families/cosmos/chain/Osmosis.ts
hzheng-ledger: not relevant any more

libs/coin-framework/src/account/helpers.ts
```diff
@@ -106,6 +106,17 @@ export const getAccountSpendableBalance = (account: AccountLike): BigNumber => {
...
+    return (
+      (a as any).cosmosResources &&
```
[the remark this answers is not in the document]
hzheng-ledger: we will fix it later after some offline discussion

libs/ledger-live-common/src/families/cosmos/chain/Osmosis.ts
hzheng-ledger: I removed that because it would use the value from the parent class

libs/ledger-live-common/src/families/cosmos/chain/cosmosBase.ts
hzheng-ledger: fixed

libs/ledger-live-common/src/families/cosmos/js-synchronisation.ts
hzheng-ledger: fixed

hedi-edelbloute (approved):
lgtm

PR #4540, opened 2023-08-30
Hardcode Ledger validator details for Solana

haammar-ledger:
### 📝 Description

The external backend we use to provide Solana validators details is flaky and sometimes doesn't provide details for the Ledger validator.

Long-term fix will consists in either fixing the external backend (unlikely, as validators.app has been unreliable for a long time now), or changing the provider we use to retrieve validators details for Solana.
In the meantime this PR mitigates this by using harcoded data as fallback if we don't get it from the backend.

### ❓ Context

- **Impacted projects**: `LLC` `LLD` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9101 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/proud-kangaroos-bake.md
+++ b/.changeset/proud-kangaroos-bake.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"@ledgerhq/live-common": patch
+---
+
+Hardcode Ledger validator details for Solana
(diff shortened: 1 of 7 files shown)


hedi-edelbloute (approved):
small linter + import problem but lgtm

sarneijim (approved):

chabroA (commented):
What happens if the backend suddenly works / is fixed?
Will LL display the hardcoded values introduced by this PR, instead of the "actual" values?
Of are they just a fallback, in case the backend fails?

PR #4546, opened 2023-08-31
chore: update RXJS to v7

valpinkman:
### 📝 Description

- Remove rxjs-compat compatibility layer between rxjs 5 and rxjs 6 in preparation of migration to rxjs 7
- Update to RXJS 7

#### Technical changes

We have migrated some rxjs calls to the new API style of RXJS 7. We also noticed the TypeScript typing is improved so new cases can be detected.

here are a few typical cases that may help other developers having conflict after the PR is merged:

- `obs.toPromise()` can be mostly changed to `firstValueFrom(obs)`. but we always need to check that "obs" always emits only one event, since the new firstValueFrom will throw in case obs terminates with no events (unlike toPromise)
- `throwError(err)` becomes `throwError(() => err)`
- `timeout(ms, err)` becomes `timeout({ each: ms, with: () => throwError(() => err) })`

note: typechecker will detect any API that wasn't migrated yet, so it's not possible to left one behind as long as things are properly typed.

complete doc: https://rxjs.dev/6-to-7-change-summary

#### Tradeoffs

What was not addressed yet:

- RXJS 7 have deprecated **some** API. They still work but they are meant to be reworked to another API. to limit the impact and the risk of our first 6->7 migration, we only changed the one that was trivial to us but we left others unaddressed. It may lead to some TypeScript warnings and we will address them incrementally in the future.

#### Functional Impact

RXJS impacts mostly the device interactions since it's what uses the most this Observable paradigm. We also use it for the bridge interface, typically the `scanAccounts`. @haammar-ledger confirmed me all coins use shared helpers like makeAccountBridgeReceive, makeScanAccounts, makeSync,... which will only requires us to test on one coin, because it's where most of the rxjs lives when it comes to coin logic <> rxjs.
This is visible by looking at the files we changed. Therefore, here is a list of minimal features it is mandatory to test for sanity checks, and it should cover 99% of the impact here:

- [x] Manager access (device connection, hsm connection, list apps)
- [x] Installing apps
- [x] Adding accounts
- [x] doing a Receive Flow
- [x] doing a Send Flow

on platforms

- [x] LLD on Mac OSX
- [x] LLM on iOS

(check the checkbox if it was manually tested by us)

### ❓ Context

- **Impacted projects**: `common`, `llm`, `lld` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9093 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


.pnpmfile.cjs
```diff
@@ -189,9 +189,9 @@ function readPackage(pkg, context) {
       addPeerDependencies("any-observable", {
         rxjs: "*",
       }),
-      addPeerDependencies("rxjs-compat", {
-        rxjs: "*",
-      }),
+      // addPeerDependencies("rxjs-compat", {
+      //   rxjs: "*",
+      // }),
```
gre, line 194: you can remove the code instead of commenting

gre:
peer programming notes after debugging an issue with @valpinkman: unexpected regression of rxjs v7 on the behavior of `events.pipe(() => debounce(() => EMPTY))` : in older version of RXJS it was letting events going through. In v7, it is no longer the case and we need to do `debounce(() => of(null))` for instance.

gre (approved):

PR #4564, opened 2023-08-31
feat: add v5 to mobile and desktop

kallen-ledger:
### 📝 Description

Add v5 endpoints to desktop and mobile applications

### ❓ Context

- **Impacted projects**: `LLM,LLD,LLC` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledger-live-common/src/exchange/swap/utils/inferRatesError.ts
```diff
@@ -0,0 +1,57 @@
...
+      ),
+      amount: new BigNumber(minOrMaxError.amount),
+    });
+  }
+
+  if (isAnErrorCodeMessageError) {
+    return getSwapAPIError(response.errorCode, response.errorMessage);
+  }
+
+  return new SwapGenericAPIError();
```
sarneijim, line 56: There is a case where we are having a limit error but we do not know the limit value, neither if it is min or max.
Are we taking this into account? 
example: use a super small quantity with a swap that it is supported by DEX providers, for instance: eth vs tether, this limit is always smaller or bigger than the CEX ones, although there is not a proper amount.

kallen-ledger: So I investigated this a bit, and it did exist, but there was no error message defined for that particular error. So the user would never get feedback as to why their quotes had errored. Not sure if we want to add one as part of this ticket?

sarneijim: The behaviour is correct.
https://github.com/LedgerHQ/ledger-live/pull/2189

libs/ledger-live-common/src/exchange/swap/index.ts
```diff
@@ -17,10 +17,11 @@ import {
   ValidationError,
 } from "../../errors";
 import getCompleteSwapHistory from "./getCompleteSwapHistory";
-import getExchangeRates from "./getExchangeRates";
-import getProviders from "./getProviders";
 import initSwap from "./initSwap";
 import { postSwapAccepted, postSwapCancelled } from "./postSwapState";
+import getExchangeRates from "./getExchangeRates";
+import getProviders from "./getProviders";
+import { isIntegrationTestEnv } from "./utils/isIntegrationTestEnv";
```
sarneijim, line 24: Not now, but when we move this code to web....I think we must try avoid use "isIntegrationTestEnv" in the logic, we could mock the calls or run a fake server...what do you think?

libs/ledger-live-common/src/exchange/swap/hooks/useSwapTransaction.ts
```diff
@@ -1,23 +1,22 @@
 import { AmountRequired, NotEnoughGas, NotEnoughGasSwap } from "@ledgerhq/errors";
 import { useMemo } from "react";
 import {
-  ExchangeRate,
   SwapSelectorStateType,
   OnNoRatesCallback,
   SwapTransactionType,
-  AvailableProviderV3,
-  OnBeforeTransaction,
+  OnBeforeFetchRates,
```
sarneijim, line 7: Can we remove "onBeforeFetchRates" & "onBeforeFetch"? Do you have something in mind? (Not sure if we really need it...)

libs/ledger-live-common/src/exchange/swap/hooks/v5/useFetchCurrencyAll.ts
```diff
@@ -0,0 +1,18 @@
...
+import { fetchCurrencyAll } from "../../api/v5";
+
+export function useFetchCurrencyAll() {
+  const { data, ...rest } = useAPI({
+    queryFn: fetchCurrencyAll,
+    queryProps: {
+      providers: getAvailableProviders(),
+    },
+    // assume the all currency list for the given props won't change during a users session.
+    staleTimeout: Infinity,
```
sarneijim, line 12: How long does it take? (only curiosity) ;)

libs/ledger-live-common/src/exchange/swap/types.ts
```diff
@@ -27,41 +27,70 @@ export type ExchangeRate = {
...
   error?: Error;
   providerURL?: string;
   expirationTime?: number;
 };
 
-type ExchangeRateV5ProviderType = "DEX" | "CEX" | "DISABLED";
+type ExchangeProviderType = "CEX" | "DEX";
 
-type ExchangeRateV5CommonProperties = {
+type ExchangeRateV5CommonPropertiesRaw = {
```
sarneijim, line 39: As we are not using "properties" in the other types we could rename it as `ExchangeRateV5CommonRaw` (suggestion, not required)

sarneijim (changes requested):
Minor changes
https://giphy.com/clips/viralhog-viral-hog-otter-claps-and-taps-for-clams-vfI3uwl4FMuDaLbZah

apps/ledger-live-desktop/src/renderer/actions/swap.ts
```diff
@@ -37,20 +35,7 @@ export const filterAvailableToAssets = (pairs: Pair[] | null | undefined, fromId
...
-  return flattenAccounts(allAccounts).map(account => {
-    const id = getAccountCurrency(account).id;
-    const isAccountAvailable = !!pairs.find(pair => pair.from === id);
-    return {
-      ...account,
-      disabled: !isAccountAvailable,
-    };
-  }) as (Account & { disabled: boolean })[];
-};
+
```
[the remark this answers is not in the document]
kallen-ledger: This is no longer used in LLD and removed as part of this PR, but the currency/to/from now return a unique list of currency ids.

libs/ledger-live-common/src/exchange/swap/index.ts
kallen-ledger: As you said, one for the future, but I'd recommend this when we move to a live app for swap. https://mswjs.io/

apps/ledger-live-mobile/src/screens/Swap/useSwapableAccounts.tsx
```diff
@@ -0,0 +1,21 @@
+import { getAccountCurrency } from "@ledgerhq/coin-framework/account/helpers";
+import { useFetchCurrencyFrom } from "@ledgerhq/live-common/exchange/swap/hooks/index";
+import { useMemo } from "react";
+import { useSelector } from "react-redux";
+import { flattenAccountsSelector } from "../../reducers/accounts";
+
+export function useSwapableAccounts() {
```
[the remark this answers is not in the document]
kallen-ledger: Yes 👍 Great suggestion.

libs/ledger-live-common/src/exchange/swap/hooks/useSwapTransaction.ts
kallen-ledger: Will check to see if it's used somewhere, but yes, happy to remove if not used.

libs/ledger-live-common/src/exchange/swap/hooks/v5/useFetchCurrencyAll.ts
kallen-ledger: Wasn't more than a few ms to a second during development.

libs/ledger-live-common/src/exchange/swap/types.ts
kallen-ledger: Yes, good suggestion.

apps/ledger-live-desktop/src/renderer/actions/swap.ts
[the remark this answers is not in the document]
sarneijim: ?
image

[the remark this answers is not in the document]
kallen-ledger: Ah yes, right you are! Yes makes sense to only return unique.

.tool-versions
```diff
@@ -0,0 +1 @@
+pnpm 8.5.1
```
sarneijim, line 1: Can we unlock the version?

kallen-ledger: Good catch, shold not have this in the PR. Thanks.

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/index.tsx
```diff
@@ -401,14 +366,6 @@ const SwapForm = () => {
     );
   }
 
-  // TODO: ensure that the error is catch by Sentry in this case
-  if (storedProviders?.length === 0 || providersError) {
```
sarneijim, line 405: I am not sure to understand this logic know...if providers.length === 0. Are we going to show a loading state?
What happens if backend is failing and we do not have providers?

kallen-ledger: So I believe that this is now managed by the usePageState. So if calls were to fail, or go to a loading state, then that hook would manage it.

apps/ledger-live-desktop/tests/models/SwapPage.ts
```diff
@@ -99,6 +99,21 @@ export class SwapPage {
     await this.destinationCurrencyDropdown.click();
   }
 
+  async waitForCurrenciesToExist() {
+    await this.page.waitForTimeout(500);
```
sarneijim, line 103: What is this 500?
- Waiting for a call....then I assume this logic is not secure.
- Waiting for a transition, can we use a global variable that makes this logic secure if we apply a change in the transition.

apps/ledger-live-mobile/src/screens/Swap/Form/index.tsx
```diff
@@ -428,10 +386,5 @@ export function SwapForm({
     );
   }
 
-  // Sentry?
```
sarneijim, line 431: Idem here

apps/ledger-live-desktop/tests/models/SwapPage.ts
kallen-ledger: Agreed, not super. So will take another look at this.

Was caused by a slight delay in the calling of the /to endpoint.

apps/ledger-live-mobile/src/screens/Swap/Form/index.tsx
kallen-ledger: Since we are no longer using the providers endpoint to get the providers this bit of code seemed redundant.

sarneijim (commented):
I know this change has a huge impact across different part of the app, but next time please, could you split this changes in multiple prs? (we are changing multiple logics here).

With the best effort coding a reviewing, I am scary that we can have multiple regressions.

apps/ledger-live-mobile/src/screens/Swap/Form/index.tsx
sarneijim: Ok, I think then you need to change the logic in the top:

```
if (storedProviders?.length) {....} <Loading>
```

apps/ledger-live-desktop/tests/models/SwapPage.ts
sarneijim: Happy if we mock those call in the unit tests.

libs/ledger-live-common/src/exchange/swap/getStatus.ts
```diff
@@ -1,15 +1,15 @@
...
 import { isIntegrationTestEnv } from "./utils/isIntegrationTestEnv";
+import { getEnv } from "@ledgerhq/live-env";
 
 export const getMultipleStatus: GetMultipleStatus = async statusList => {
   if (isIntegrationTestEnv()) return mockGetStatus(statusList);
 
   const res = await network({
     method: "POST",
-    url: `${getSwapAPIBaseURL()}/swap/status`,
+    url: `${getEnv("SWAP_API_BASE_V5")}/swap/status`,
```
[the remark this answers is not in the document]
kallen-ledger: Yes 👍

libs/ledger-live-common/src/hooks/useAPI.ts
```diff
@@ -71,7 +71,8 @@ export function useAPI<T, P extends Record<PropertyKey, unknown> | undefined>({
         fetch();
       }
     }
-  }, [cacheKey, staleTimeout, enabled, fetch, onSuccess]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [cacheKey, staleTimeout, enabled]);
```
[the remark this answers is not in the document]
kallen-ledger: Will leave a comment here as to why

libs/ledger-live-common/src/hooks/useAPI.ts
```diff
@@ -27,7 +27,7 @@ export function useAPI<T, P extends Record<PropertyKey, unknown> | undefined>({
   queryFn,
   queryProps = {},
   enabled = true,
-  staleTimeout = 5000,
+  staleTimeout = 60000,
```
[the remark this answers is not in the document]
kallen-ledger: This is just a default staleTimeout since at the moment quotes are only refreshed every 60 seconds. But will put back to 5seconds

beths-ledger (approved):
big so there's probably stuff that I cannot fully test or understand, but it looks great!

apps/ledger-live-mobile/src/screens/Swap/SubScreens/SelectCurrency.tsx
```diff
@@ -32,22 +43,35 @@ export function SelectCurrency({
       // @ts-expect-error navigation type is only partially declared
       navigation.navigate(ScreenName.SwapForm, { currency });
     },
-    [track, navigation],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
   );
+
+  const RenderItem = useCallback(({ item }: { item: CryptoCurrency | TokenCurrency }) => {
```
beths-ledger, line 50: does renderItem need to be memoized in the callback? I usually just use `renderItem = (item) => <Row currency={item} ... />` So it looks to me like we could avoid the useCallback and eslint-disable.

If there is a need, `renderItem(onSelect)` could also be a function outside of the SelectCurrency component, which returns a function that takes the flatlist data item props, i.e. 
```
function renderItem(onSelect) {
  return ({currency}: { currency: CryptoCurrency | TokenCurrency }) => 
    <CurrencyRow currency={currency} onPress={onSelect} />
}
```
but I don't know if that's really necessary.

apps/ledger-live-mobile/src/screens/Swap/SubScreens/SelectCurrency.tsx
```diff
@@ -32,22 +43,35 @@ export function SelectCurrency({
...
   const renderList = useCallback(
     items => (
       <FlatList
         contentContainerStyle={styles.list}
+        removeClippedSubviews={true}
         data={items}
-        renderItem={({ item }) => <CurrencyRow currency={item} onPress={onSelect} />}
-        keyExtractor={currency => currency.id}
+        renderItem={RenderItem}
+        keyExtractor={keyExtractor}
```
beths-ledger, line 64: is currency.id not unique in this list?

kallen-ledger: Ah, this was for testing ( had 1000's of the same currency ). Can revert this.

apps/ledger-live-mobile/src/screens/Swap/SubScreens/SelectCurrency.tsx
```diff
@@ -23,7 +34,7 @@ export function SelectCurrency({
   const { track } = useAnalytics();
 
   const onSelect = useCallback(
-    (currency: CryptoCurrency | TokenCurrency) => {
+    (currency: Currency) => {
```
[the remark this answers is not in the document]
kallen-ledger: I believe Currency is just a union of Crypto and Token currency types. But will double check.

sarneijim (approved):

PR #4583, opened 2023-09-05
feat: injective integration

hedi-edelbloute:
### 📝 Description

Integrate injective

### ❓ Context

- **Impacted projects**: `LLC,LLD,LLM` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-8106` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -72,16 +76,26 @@ export class CosmosAPI {
...
-        response.sequence = parseInt(data.account.sequence);
+      if (srcAccount.pub_key) {
+        accountData.pubKey = srcAccount.pub_key.key;
+        accountData.pubKeyType = srcAccount.pub_key["@type"];
       }
-      // eslint-disable-next-line no-empty
-    } catch (e) {}
-    return response;
+    } catch (e) {
+      console.log("Could not fetch account info, using default values instead");
```
haammar-ledger, line 95: Is this really what we want? Just asking, because sometimes using default values could actually cause bugs and just throwing an error is cleaner.
Let's review how the returned fields are used and what the impact is with default values, to assess if they are ok or not.

(Also, `console.log` being only available in dev env, for logs that we actually want to add in prod `log` should be used instead)

libs/ledger-live-common/src/families/cosmos/helpers.ts
```diff
@@ -13,3 +13,26 @@ export const getMainMessage = (messages: CosmosMessage[]): CosmosMessage => {
...
+      for (const element of value) {
+        newArray.push(sortObjectKeysDeeply(element));
+      }
+      object[key] = newArray;
+    } else if (typeof value === "object") {
+      object[key] = sortObjectKeysDeeply(value);
+    }
+  }
+  return order(object);
+};
```
haammar-ledger, line 38: No clever lodash that does this?
Maybe for `order` at least?

libs/ledger-live-common/src/families/cosmos/js-buildTransaction.ts
```diff
@@ -255,180 +251,21 @@ export const buildTransaction = async (
...
+export const buildTransaction = async (
   protoMsgs: Array<ProtoMsg>,
+  memo: string,
+  pubKeyType: string,
+  pubKey: string,
+  feeAmount: any,
+  gasLimit: any,
+  sequence: string,
+  signature: Uint8Array,
 ): Promise<Uint8Array> => {
```
haammar-ledger, line 263: Just a good practice: when starting to have a lot of params it's better to use a single object as param, so that each field is handled explicitely.
```suggestion
export const buildTransaction = async ({
  protoMsgs,
  memo,
  pubKeyType,
  pubKey,
  feeAmount,
  gasLimit,
  sequence,
  signature,
}:{
  protoMsgs: Array<ProtoMsg>;
  memo: string;
  pubKeyType: string;
  pubKey: string;
  feeAmount: any;
  gasLimit: any;
  sequence: string;
  signature: Uint8Array;
}): Promise<Uint8Array> => {
```

libs/ledger-live-common/src/families/cosmos/js-buildTransaction.ts
```diff
@@ -21,17 +21,13 @@ import { PubKey } from "@keplr-wallet/proto-types/cosmos/crypto/secp256k1/keys";
...
-import { EncodeObject, GeneratedType, makeAuthInfoBytes, Registry } from "@cosmjs/proto-signing";
-import { CosmosAPI } from "./api/Cosmos";
 
 type ProtoMsg = {
   typeUrl: string;
   value: Uint8Array;
 };
 
-export const buildTransaction = async (
+export const txToMessages = async (
```
haammar-ledger, line 30: TBC, but I think this may not need to be `async`

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
```diff
@@ -50,35 +47,30 @@ export const getEstimatedFees = async (
...
-      account,
-      transaction,
-      pubkey,
-      unsignedPayload,
+  const { protoMsgs } = await txToMessages(account, transaction);
+  const { sequence, pubKeyType, pubKey } = await cosmosAPI.getAccount(account.freshAddress);
+  const signature = new Uint8Array(Buffer.from(account.seedIdentifier, "hex"));
+
+  if (protoMsgs && protoMsgs.length > 0) {
+    const tx_bytes = await buildTransaction(
```
haammar-ledger, line 59: You can use snakeCase here as well

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
```diff
@@ -50,35 +47,30 @@ export const getEstimatedFees = async (
...
+      protoMsgs,
+      transaction.memo || "",
+      pubKeyType,
+      pubKey,
+      undefined,
+      undefined,
+      sequence + "",
```
haammar-ledger, line 66: Beware, I don't think that works
```
undefined + "" === "undefined"
null + "" === "null"
```

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -42,21 +45,44 @@ const signOperation: SignOperationFnSignature<Transaction> = ({ account, deviceI
...
+          const path = account.freshAddressPath.split("/").map(p => parseInt(p.replace("'", "")));
+
+          const { compressed_pk } = await app.getAddressAndPubKey(path, chainInstance.prefix);
+
+          const signResponseApp =
+            path[1] === 60
+              ? await app.sign(path, tx, chainInstance.prefix)
+              : await app.sign(path, tx);
+
+          const pubKey = Buffer.from(compressed_pk).toString("base64");
```
haammar-ledger, line 70: Just for readability: I'd move this right after `compressed_pk`  above

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -128,6 +154,7 @@ const signOperation: SignOperationFnSignature<Transaction> = ({ account, deviceI
             signedOperation: {
               operation,
               signature: signed,
+              expirationDate: undefined,
```
haammar-ledger, line 157: This is isn't required anymore since recently, you can remove

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -42,21 +45,44 @@ const signOperation: SignOperationFnSignature<Transaction> = ({ account, deviceI
...
+          const tx = Buffer.from(JSON.stringify(signDoc), "utf-8");
+
+          const app = new CosmosApp(transport);
+
+          const path = account.freshAddressPath.split("/").map(p => parseInt(p.replace("'", "")));
+
+          const { compressed_pk } = await app.getAddressAndPubKey(path, chainInstance.prefix);
+
+          const signResponseApp =
+            path[1] === 60
```
haammar-ledger, line 66: Just for my curiosity: can you please explain why we need this condition?

haammar-ledger (commented):
Really cool code cleanup around `buildTransaction` :slightly_smiling_face:

Let's just make sure this introduces no regression :crossed_fingers: since it changes a core part of the cosmos integration.

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -72,16 +76,26 @@ export class CosmosAPI {
...
+
+      if (srcAccount.sequence) {
+        accountData.sequence = parseInt(srcAccount.sequence);
       }
 
-      if (data.account.sequence) {
-        response.sequence = parseInt(data.account.sequence);
+      if (srcAccount.pub_key) {
+        accountData.pubKey = srcAccount.pub_key.key;
+        accountData.pubKeyType = srcAccount.pub_key["@type"];
```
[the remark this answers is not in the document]
hedi-edelbloute: Yes :sweat_smile: See https://injective-api.polkachu.com/cosmos/auth/v1beta1/accounts/inj1d2kmwygyqh2vf658yns8ytqdffhsr54urpuesj

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: To be honest, I don't know if it's necessary, I already changed a lot so I didn't want to do any breaking change here...

I agree throwing would be cleaner

Will add log instead while we discuss this

libs/ledger-live-common/src/featureFlags/defaultFeatures.ts
```diff
@@ -113,6 +113,9 @@ export const defaultFeatures = {
   currencyCoreum: {
     enabled: false,
   },
+  currencyInjective: {
+    enabled: true,
```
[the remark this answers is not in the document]
hedi-edelbloute: This is temporary

libs/ledger-live-common/src/families/cosmos/js-buildTransaction.ts
```diff
@@ -255,180 +251,21 @@ export const buildTransaction = async (
...
-  if (!isComplete) {
-    return [];
-  }
-
-  return messages;
-};
-
-export const postBuildTransaction = async (
-  signResponse: AminoSignResponse,
+export const buildTransaction = async (
```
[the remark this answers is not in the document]
hedi-edelbloute: same

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
haammar-ledger: Ha, I don't know why I assumed it was a string, but it's better that it's a number.
And you're right that here the typing prevents it, so we're good nevermind.

Just to note that the typing doesn't make it impossible though to assign null or undefined at _runtime_:
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#strictnullchecks-off

And this can especially happen for data that's coming from outside the app, typically from backends like here.
If some backend API decides to stop providing a field that we're expecting, TS won't protect us from having undefined in the expected field.
Just something to remember, but this risk is already here for 99% of the API implementations of our coins, so no difference here.

libs/ledger-live-common/src/families/cosmos/js-buildTransaction.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
hedi-edelbloute: Yes, it's typed as number as we always treat it like this, it comes from backend as a number, and its default value is a number

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -42,21 +45,44 @@ const signOperation: SignOperationFnSignature<Transaction> = ({ account, deviceI
...
             chainId,
             transaction.memo || "",
             accountNumber.toString(),
             sequence.toString(),
           );
-          const ledgerSigner = new LedgerSigner(transport, {
-            hdPaths: [stringToPath("m/" + account.freshAddressPath)],
-            prefix: cryptoFactory(account.currency.id).prefix,
-          });
+          signDoc = sortObjectKeysDeeply(signDoc) as StdSignDoc;
```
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
hedi-edelbloute: removed

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -42,21 +45,44 @@ const signOperation: SignOperationFnSignature<Transaction> = ({ account, deviceI
...
+            path[1] === 60
+              ? await app.sign(path, tx, chainInstance.prefix)
+              : await app.sign(path, tx);
+
+          const pubKey = Buffer.from(compressed_pk).toString("base64");
+          const signature = Buffer.from(
+            Secp256k1Signature.fromDer(signResponseApp.signature).toFixedLength(),
+          );
+
+          const tx_bytes = await buildTransaction(
```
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/package.json
```diff
@@ -121,9 +121,9 @@
     "@celo/wallet-base": "^3.0.1",
     "@celo/wallet-ledger": "^3.0.1",
     "@cosmjs/amino": "^0.28.4",
-    "@cosmjs/crypto": "^0.26.5",
+    "@cosmjs/crypto": "^0.31.0",
     "@cosmjs/launchpad": "^0.26.5",
-    "@cosmjs/ledger-amino": "^0.26.5",
+    "@cosmjs/ledger-amino": "^0.31.0",
     "@cosmjs/proto-signing": "^0.26.5",
```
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: Confirmed : https://injective-api.polkachu.com/cosmos/auth/v1beta1/accounts/inj1ddscefl2mnv5sfnduesz7tu0x3fdrttven4f9m

It is needed for a pristine account, we need a 0 sequence in that case else we are stuck

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -72,16 +76,26 @@ export class CosmosAPI {
         url: `${this.defaultEndpoint}/cosmos/auth/${this.version}/accounts/${address}`,
       });
 
-      if (data.account.account_number) {
-        response.accountNumber = parseInt(data.account.account_number);
+      const srcAccount =
+        data.account.base_account != null ? data.account.base_account : data.account;
```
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/helpers.ts
hedi-edelbloute: Found
```
var ordered = {};   
_(Categories).keys().sort().each(function (key) {
  ordered[key] = Categories[key];
});

Categories = ordered;
```

```
_(obj).toPairs().sortBy(0).fromPairs().value()
```

Lib [lodash-keyarrange](https://www.npmjs.com/package/lodash-keyarrange)

Only the lib cover every sublevels

Do you have any preference ? I find my current solution to be ok-tier

Feel free to search more than me

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
hedi-edelbloute: Backend returns a number
Nano returns a string

buildTransaction behavior requires a Long

For now we do Long.fromString in the method, we take nano output as is, and we convert to String on this line

I feel like it is the least error prone way, wdyt ?

libs/ledger-live-common/src/families/cosmos/helpers.ts
```diff
@@ -13,3 +13,27 @@ export const getMainMessage = (messages: CosmosMessage[]): CosmosMessage => {
...
+      .sort((a, b) => a.localeCompare(b))
+      .forEach(function (key) {
+        out[key] = sortObjectKeysDeeply(src[key]);
+      });
+
+    return out;
+  }
+
+  return src;
+};
```
haammar-ledger, line 39: I'm thinking we should set a safety limit on the number of recursions. A bit paranoid, but just to be safe, since JS doesn't forbid recursive objects.
We can put something large enough, like `20` or idk, just to prevent infinite loops.

libs/ledger-live-common/src/families/cosmos/helpers.ts
haammar-ledger: I had a quick look and didn't find anything great either, so your current version is ok to me :+1: 
Just see other my comment about safety on recursion.

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
haammar-ledger: :ok_hand: Thanks for checking :raised_hands:

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
haammar-ledger: Hmm I thought about using `String` as well, but it seems discouraged by the doc itself...
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String
It creates a `String` instance instead of `string` primitive, which indeed may be weird.

Also apparently `String(undefined) === "undefined"` anyway :sweat_smile:

In the end I think I would go with:
```
sequence ? sequence + "" : "0",
```
Uglier but the safest, I think.

libs/ledger-live-common/src/families/cosmos/js-buildTransaction.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-buildTransaction.ts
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/featureFlags/defaultFeatures.ts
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/package.json
```diff
@@ -121,10 +121,8 @@
     "@celo/wallet-base": "^3.0.1",
     "@celo/wallet-ledger": "^3.0.1",
     "@cosmjs/amino": "^0.28.4",
```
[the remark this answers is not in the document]
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -2,15 +2,15 @@ import type { Transaction } from "./types";
 import { Observable } from "rxjs";
 import { withDevice } from "../../hw/deviceAccess";
 import { encodeOperationId } from "../../operation";
-import { LedgerSigner } from "@cosmjs/ledger-amino";
-import { stringToPath } from "@cosmjs/crypto";
-import { buildTransaction, postBuildTransaction } from "./js-buildTransaction";
+import { txToMessages, buildTransaction } from "./js-buildTransaction";
 import BigNumber from "bignumber.js";
 import { makeSignDoc } from "@cosmjs/launchpad";
```
hzheng-ledger, line 7: I think the "@cosmjs/launchpad" is legacy.
you can do
```
import { serializeSignDoc, makeSignDoc } from "@cosmjs/amino";
```
and you can replace the "import { sortObjectKeysDeeply } from "./helpers";" with "serializeSignDoc" 

and don't have to maintain sortObjectKeysDeeply by ourselves

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
```diff
@@ -50,14 +50,35 @@ const signOperation: SignOperationFnSignature<Transaction> = ({ account, deviceI
             accountNumber.toString(),
             sequence.toString(),
           );
-          const ledgerSigner = new LedgerSigner(transport, {
-            hdPaths: [stringToPath("m/" + account.freshAddressPath)],
-            prefix: cryptoFactory(account.currency.id).prefix,
+          const tx = Buffer.from(JSON.stringify(sortObjectKeysDeeply(signDoc)), "utf-8");
```
hzheng-ledger, line 53: we can replace the sortObjectKeysDeeply with serializeSignDoc as I mentioned above

libs/ledger-live-common/src/families/cosmos/specs.ts
```diff
@@ -437,10 +438,18 @@ const generateGenericCosmosTest = (currencyId: string, config?: Partial<AppSpec<
...
       model: DeviceModelId.nanoS,
       appName: "Cosmos",
+      appVersion: "2.34.11",
     },
     genericDeviceAction: acceptTransaction,
     testTimeout: 2 * 60 * 1000,
     test: cosmosLikeTest,
+    onSpeculosDeviceCreated:
+      currencyId === "injective"
+        ? async ({ transport }) => {
```
hzheng-ledger, line 448: may be we should add an inline comment here to explain the expert mode for injective here

libs/coin-framework/src/bot/types.ts
```diff
@@ -147,6 +151,11 @@ export type AppSpec<T extends TransactionCommon> = {
   allowEmptyAccounts?: boolean;
   // do not keep operations in accounts (Cosmos family case)
   skipOperationHistory?: boolean;
+  onSpeculosDeviceCreated?: (device: {
```
hzheng-ledger, line 154: add an inline comment as other methods in this files plz

libs/ledger-live-common/src/config/defaultConfig.ts
```diff
@@ -57,6 +57,10 @@ const defaultConfig = {
         lcd: "https://full-node.mainnet-1.coreum.dev:1317",
         minGasPrice: 0.1,
       },
+      injective: {
+        lcd: "https://injective-api.polkachu.com",
+        minGasPrice: 900000000,
```
hzheng-ledger, line 62: I find the the low gas price for keplr is
higher than ours. Can you confirm that this value is enough?
https://github.com/chainapsis/keplr-wallet/blob/04f394ac30a8c5923f01dad0a4fb5ff6f8faee47/packages/extension/src/config.ts#L1918

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
```diff
@@ -42,58 +39,54 @@ export const calculateFees: CacheRes<
...
   transaction: Transaction,
-): Promise<{ estimatedFees: BigNumber; estimatedGas: BigNumber }> => {
-  const cosmosCurrency = cryptoFactory(account.currency.id);
-  let estimatedGas = new BigNumber(cosmosCurrency.defaultGas);
+): Promise<{
+  gasWanted: BigNumber;
+  gasWantedFees: BigNumber;
+}> => {
+  const chainInstance = cryptoFactory(account.currency.id);
+  let gasWanted = new BigNumber(chainInstance.defaultGas);
```
hzheng-ledger, line 54: no need to give an initial value to gasWanted.
line 83, you can just write:
const gasWanted = gasUsed.times(getEnv("COSMOS_GAS_AMPLIFIER")).integerValue(BigNumber.ROUND_CEIL);

hzheng-ledger (changes requested):
see my comments plz

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
hedi-edelbloute: Nice catch

https://github.com/cosmos/cosmjs/blob/main/packages/amino/src/signdoc.ts#L35

libs/ledger-live-common/src/config/defaultConfig.ts
hedi-edelbloute: I took it from here : https://github.com/cosmos/chain-registry/blob/master/injective/chain.json#L23

libs/ledger-live-common/src/families/cosmos/helpers.ts
hedi-edelbloute: No longer relevant as we use cosmjs lib now

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/js-signOperation.ts
hedi-edelbloute: Done

libs/ledger-live-common/src/families/cosmos/specs.ts
hedi-edelbloute: please check updated version

libs/coin-framework/src/bot/types.ts
hedi-edelbloute: done

libs/ledger-live-common/src/families/cosmos/js-prepareTransaction.ts
hedi-edelbloute: indeed, done

hzheng-ledger (approved):
looks good to me

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -46,7 +50,7 @@ export class CosmosAPI {
         unbondings,
         withdrawAddress,
       ] = await Promise.all([
-        this.getAccount(address),
+        this.getAccount(address, this.chainInstance.defaultPubKeyType),
```
haammar-ledger, line 53: It doesn't seem necessary to provide the `defaultPubKeyType` to the `getAccount` function, since it already has access to it

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -83,16 +92,30 @@ export class CosmosAPI {
...
+        accountData.accountNumber = parseInt(srcAccount.account_number);
+      }
+
+      if (srcAccount.sequence) {
+        accountData.sequence = parseInt(srcAccount.sequence);
       }
 
-      if (data.account.sequence) {
-        response.sequence = parseInt(data.account.sequence);
+      if (srcAccount.pub_key) {
```
haammar-ledger, line 106: Just curious, I don't remember if I asked already: did you see cases where the `pub_key` field isn't provided?
If it wasn't optional, then we wouldn't need `defaultPubKeyType` at all, but I assume it is optional?

libs/ledger-live-common/src/families/cosmos/specs.ts
```diff
@@ -430,17 +431,28 @@ function cosmosLikeMutations(minimalTransactionAmount: BigNumber): MutationSpec<
...
+  isExpertModeRequired: boolean,
+  config?: Partial<AppSpec<Transaction>>,
+) => {
   return {
     name: currencyId,
     currency: getCryptoCurrencyById(currencyId),
     appQuery: {
       model: DeviceModelId.nanoS,
       appName: "Cosmos",
+      appVersion: "2.34.11",
```
haammar-ledger, line 445: We want to remove this now, right?

libs/ledger-live-common/src/families/cosmos/specs.ts
```diff
@@ -430,17 +431,28 @@ function cosmosLikeMutations(minimalTransactionAmount: BigNumber): MutationSpec<
   ];
 }
 
-const generateGenericCosmosTest = (currencyId: string, config?: Partial<AppSpec<Transaction>>) => {
+const generateGenericCosmosTest = (
+  currencyId: string,
+  isExpertModeRequired: boolean,
```
haammar-ledger, line 436: I'm thinking `isExpertModeRequired` should also belong to `AppSpec`. There are already a couple custom config params there, and I believe it's better to keep them all centralized.

libs/ledger-live-common/src/config/defaultConfig.ts
haammar-ledger: Any definitive answer here?
What do you think is the better source of truth between chain-registry and kepler?

haammar-ledger (approved):
Nothing blocking a merge, but a couple possible improvements, could be done in a second step if needed.

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: pub_key is null when account is pristine, we want to use default value else we end up on the bug i resolved last week

libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
hedi-edelbloute: indeed

libs/ledger-live-common/src/families/cosmos/specs.ts
hedi-edelbloute: i'm going to try while it's getting qaed

libs/ledger-live-common/src/families/cosmos/specs.ts
hedi-edelbloute: I am not sure since it's cosmos specific

libs/ledger-live-common/src/config/defaultConfig.ts
hedi-edelbloute: I wouldn't change it as I had failing txs with 700000000

haammar-ledger:
Another future improvement, from the bots results:

> Spec injective:
>     There are not enough accounts to cover all mutations. Please increase the account target to at least 7 accounts

libs/ledger-live-common/src/families/cosmos/specs.ts
hedi-edelbloute: Done

PR #4612, opened 2023-09-06
fix: add react-native-webview isLoading state on loadEnd

mitchellv-ledger:
### 📝 Description

A bug was observed on Earn dashboard where switching tabs caused the webview to display an infinitely loading icon.  After an investigation this appears to be due to onLoadEnd not setting the `isLoading` property on react-native-webview component as detailed https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#onloadend [here](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#onloadend).  This fix should hopefully avoid this from occuring.

### ❓ Context

- **Linked resource(s)**: `LIVE-9185` 

### ✅ Checklist

- [X] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
+++ b/apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
@@ -8,7 +8,12 @@ import { NetworkError } from "./NetworkError";
 
 export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
   ({ manifest, inputs = {}, onStateChange, allowsBackForwardNavigationGestures = true }, ref) => {
-    const { onMessage, onLoadError, webviewProps, webviewRef } = useWebView(
+    const {
+      onMessage,
+      onLoadError,
+      webviewProps: { onLoadEnd, ...webviewStateProps },
+      webviewRef,
+    } = useWebView(
       {
         manifest,
         inputs,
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -38,7 +43,16 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
         allowsUnsecureHttps={__DEV__ && !!Config.IGNORE_CERTIFICATE_ERRORS}
-        {...webviewProps}
+        {...webviewStateProps}
+        onLoadEnd={event => {
+          if (onLoadEnd) {
+            onLoadEnd(event);
+
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore next-line
+            this.isLoading = event.nativeEvent.loading;
```
Justkant, line 53: TS is correct here IMO
You don't have a `this` in this context
The exemple present in the documentation you shared is probably just showing a use case in a class component
Also there is no `isLoading` in the codebase of `react-native-webview` so I don't think this code does anything

Justkant (commented):
Not sure about this but I don't think the changes here are needed, I would need to reproduce the issue locally to check a bit more

Also the loading state used by the UI and the webview are updated here https://github.com/LedgerHQ/ledger-live/blob/19a559dfccda66dfddb432c1c9b54249b52fc2e2/apps/ledger-live-mobile/src/components/Web3AppWebview/helpers.ts#L182-L223

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
mitchellv-ledger: I thought there would be more to the underlying component as I checked and also couldn't see the isLoading declared anywhere.  `this` is available but `isLoading` is undefined so can see that it will make no actual difference.  When I was initially testing the differences I could reproduce the infinite loading only when this line was not being set, but now it is not being reproduced even when it is not.

I suspect it may be a memory issue or race condition for when `onLoadEnd` and `onNavigationStateChange` run.

PR #4616, opened 2023-09-07
fix: add isDeepLinking state to prevent deeplink lock

beths-ledger:
### 📝 Description

EARN dashboard on Android deep links (e.g. on tapping "Get [coin]" button) were triggering the LLM password lock.`<SkipLock />` does not work in this case because the background state is triggered more than once.

- Adds `isDeepLinking` to state
- Set on when deep link initiated
- Reset to false when earn screen is returned to, or drawer is closed

### ❓ Context

- **Impacted projects**: `` ledger-live-mobile
- **Linked resource(s)**: `` https://ledgerhq.atlassian.net/browse/LIVE-9180

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/components/RootNavigator/EarnLiveAppNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/EarnLiveAppNavigator.tsx
@@ -1,20 +1,21 @@
-import React, { useEffect, useMemo } from "react";
+import React, { useCallback, useEffect, useMemo } from "react";
 import { createStackNavigator } from "@react-navigation/stack";
 
 import { useTheme } from "styled-components/native";
 import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
+import { useFocusEffect, useRoute } from "@react-navigation/native";
+import { useDispatch, useSelector } from "react-redux";
+import { getAccountIdFromWalletAccountId } from "@ledgerhq/live-common/wallet-api/converters";
+
 import { ScreenName, NavigatorName } from "../../const";
 import { getStackNavigatorConfig } from "../../navigation/navigatorConfig";
-import { useRoute } from "@react-navigation/native";
-import { useSelector } from "react-redux";
-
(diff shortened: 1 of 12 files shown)


apps/ledger-live-mobile/src/reducers/types.ts
```diff
@@ -268,6 +268,8 @@ export type EarnState = {
     message?: string;
     messageTitle?: string;
   };
+  /** Earn web player deeplinks need to skip privacy lock. Resets on return to Earn dashboard. If other deeplinks also need to skip lock, could move this to AppState or somewhere more global. */
```
mitchellv-ledger, line 271: I think best to move directly to appstate

mitchellv-ledger (changes requested):
Looks great, I think the redux state should live in appstate instead of earn as it's not earn's responsibility to maintain this feature

apps/ledger-live-mobile/src/reducers/types.ts
beths-ledger: i actually agree with this, thanks for the feedback!

apps/ledger-live-mobile/src/components/RootNavigator/EarnLiveAppNavigator.tsx
```diff
@@ -23,11 +24,12 @@ type NavigationProps = BaseComposite<
 >;
 
 const Earn = (props: NavigationProps) => {
+  const dispatch = useDispatch();
   // Earn dashboard feature flag
   const ptxEarn = useFeature("ptxEarn");
   const paramAction = props.route.params?.action;
   const navigation = props.navigation;
-  const accounts = useSelector(accountsSelector);
+  const accounts = useSelector(shallowAccountsSelector);
```
ofreyssinet-ledger, line 32: is this change intentional in the scope of this PR?

mitchellv-ledger (approved):
awesome 🏆

PR #4689, opened 2023-09-14
Feat/live 8933 add new drawer before cancelling firmware update

aussedatlo:
### 📝 Description

Add a new drawer view to ask user before cancelling current firmware update.

### ❓ Context

- **Linked resource(s)**: [LIVE-8933] ; [Figma](https://www.figma.com/file/Le3nxsc4L8ApwHXFbMbPsA/Onboarding-%2F-Stax?type=design&mode=design&t=3OwVpKXsRvp6IGiL-0)

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

image

### 🚀 Expectations to reach

[LIVE-8933]: https://ledgerhq.atlassian.net/browse/LIVE-8933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/Cancel.tsx
```diff
@@ -0,0 +1,47 @@
...
+        px={12}
+        alignSelf="stretch"
+        flexDirection="row"
+        justifyContent="space-between"
+        paddingTop={4}
+        paddingBottom={1}
+      >
+        <Flex alignItems="center" justifyContent="center">
+          <Link onClick={onCancel}>
+            <Text fontWeight="600">{t("manager.firmware.cancelUpdate")}</Text>
```
alexandremgo, line 35: We don't use the `font-weight` property often -> is there a `variant` (it's a prop on `Text`) that is closed to what you need ? (you can choose `body`, `h4` etc. it's too have some consistency with the codebase)

apps/ledger-live-desktop/src/renderer/screens/manager/FirmwareUpdate/index.tsx
```diff
@@ -61,13 +57,13 @@ const FirmwareUpdate = (props: Props) => {
     error,
     onReset,
   } = props;
-  const { setDrawer } = useContext(context);
```
alexandremgo, line 64: Oh ! I guess you were blocked when you wanted to add a `cancel` drawer, while the update drawer was still opened ?

ofreyssinet-ledger:
To make your PR easier to review you can add `[LIVE-8933]` (but not in code formatting like this) in the description (in linked resources) and some magic bot from jira will pick it up and create the link to the jira ticket automatically.
Also, optional but you can directly add the Figma link when there's one (makes it slightly easier than just a Jira link)

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/Cancel.tsx
ofreyssinet-ledger: in that case you should use `import { Link } from "react-ui"` and it's stylised as you want.

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/index.tsx
```diff
@@ -55,6 +58,8 @@ export type StepProps = {
 export type StepId = "idCheck" | "updateMCU" | "updating" | "finish" | "resetDevice" | "restore";
 
 export type Props = {
+  isOpen: boolean;
+  onRequestClose: () => void;
```
alexandremgo, line 62: Also this update on the `Props` definition is causing [this TS error](https://github.com/LedgerHQ/ledger-live/actions/runs/6183570971/job/16785550172#step:6:31)

If we keep this change, i think you need to also update `apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/EarlySecurityChecks/index.tsx` ;)

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/Cancel.tsx
ofreyssinet-ledger: https://react-ui-storybook.vercel.app/?path=/story/cta-link--default

ofreyssinet-ledger: As for regular text (doesn't apply here but just so you know), you should always use the `variant` and `fontWeight` props and never hardcode `fontSize`, `fontFamily` or `fontWeight`.

For instance in the following example, `variant` will be `"bodyLineHeight"` and `fontWeight` will be `"medium"`. You can find all the existing combinations here: https://react-ui-storybook.vercel.app/?path=/story/asorted-typography-text--overview
In case that info is missing on Figma, it's a mistake in the design so ping the designer responsible (or a product manager if you don't know who's the design).
Screenshot 2023-09-14 at 15 35 26

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/index.tsx
```diff
@@ -222,72 +239,86 @@ const UpdateModal = ({
...
-    <Flex
-      key={`${nonce}_fwUpdate`}
-      flexDirection="column"
-      rowGap={5}
-      height="100%"
-      overflowY="hidden"
-      width="100%"
-      flex={1}
-      data-test-id="firmware-update-container"
+    <SideDrawer
```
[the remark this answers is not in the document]
ofreyssinet-ledger: Nope normally on LLD the stable way to go is to use 
```ts
import { context } from "~/renderer/drawers/Provider";
```
and then
```ts
const { setDrawer } = useContext(context);
```
Then you can use a `useEffect` that will open/close the drawer based on your `isOpen` state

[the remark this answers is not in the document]
ofreyssinet-ledger: this will ensure that only 1 drawer at a time can be rendered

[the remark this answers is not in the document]
ofreyssinet-ledger: Ah wait but you're already inside a drawer here 🤔

apps/ledger-live-desktop/src/renderer/drawers/Drawer.tsx
```diff
@@ -18,23 +18,23 @@ const transitionStyles = {
   },
 };
 const DURATION = 200;
-const Bar = styled.div.attrs<{ state: TransitionStatus }>(props => ({
+const Bar = styled.div.attrs<{ state: TransitionStatus; padding: boolean }>(props => ({
```
ofreyssinet-ledger, line 21: just a suggestion, it's up to you: maybe `withPadding` would be a tiny bit clearer (with `padding` I would expect a `number` value)

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/Header.tsx
ofreyssinet-ledger, line 1: Instead of duplicating the code for the Header, is it possible to extract the one from `SideDrawer` into its own purely presentational component `SideDrawerHeader` ? Basically a generic header component for side drawers. Then it would be use in `UpdateFirmwareModal` and in `SideDrawer` and we avoid potential divergences between two duplicated implementations in the future.

it would take these props:
```
onClickBack?: () => void; // if undefined, no back button displayed
title?: string; // if undefined, no title displayed
onClickClose?: () => void; // if undefined, no close button displayed
```

ofreyssinet-ledger (changes requested):
Nice!
Just one request: if it's possible to remove the new Header component and instead extract the existing one into a generic presentational `SideDrawerHeader` component.
Up to debate of course if you disagree 👍

apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/Header.tsx
aussedatlo: I was also thinking about this

ofreyssinet-ledger (approved):
perfect thanks 👍

PR #4709, opened 2023-09-15
feat: setup new log system and record logs from LLD internal thread

alexandremgo:
#### `ledgerjs/logs`
- new `trace` and `LocalTracer` definitions, **without breaking the behavior** of the existing `log` function

#### `hw-transport`, `hw-transport-node-hid-noevents`, `hw-transport-node-hid-singleton`, `react-native-hw-transport-ble`, `ledger-live-common` 

- Usage of new tracing system in some transports
- The tracing helps keeping a context (for ex a `job id`) that is propagated to other logs, creating a (simple) tracing span

Note on the "tracing": for now added a simple "context" that is propagated from `withDevice` to called functions, mainly from Transport. This is a preview of what could be done to have a real tracing system in the DeviceSDK

Note: the optional `openTimeoutMs` on the different `open` function is necessary for legacy (used by one of the Transport), and will actually be used in another PR. Don't mind it during the code review. But it should only by passed to a couple of other functions, not directly used (in this PR).

#### LLD:
- Logs from the `internal` thread are forwarded to the `main` thread
- The `main` thread records them, and can export them
  - on export, recorded logs from the `renderer` thread are sent to the `main` thread, and the `main` thread expand (not a merge, just pushed at the end) them with the logs it recorded from the `internal` thread
- If `VERBOSE` env var is set, filtered logs can be stdout from the main thread
  - Same from the `renderer` thread
- Setup simple tracing system (context) on LLD

Usage with `VERBOSE`:
```
VERBOSE="apdu,hw,hid-verbose,usb-detection" pnpm dev:lld
```

#### LLM:
- Setup simple tracing system on LLM with context
- If `VERBOSE` env var is set, filtered logs can be stdout from the main thread

Usage with `VERBOSE`: in a `.env` file, add (for ex)
```
VERBOSE="apdu,hw,ble-verbose,ble-error,transport"
```
You will need to re-build fully the app so the new `.env` file is transfered.

### ❓ Context

- **Impacted projects**: `hw-transport`, `hw-transport-node-hid-noevents`, `hw-transport-node-hid-singleton`, `react-native-hw-transport-ble`, `ledger-live-common` , `ledgerjs/logs`, LLM and LLD
- **Linked resource(s)**:  [LIVE-9405](https://ledgerhq.atlassian.net/browse/LIVE-9405)

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] **No breaking changes** 

#### LLD performance

Non-scientific manual measurements (on dev build): beginning of the fw update until the end of the fw transfer:

<details>
  <summary>Measurements fw update transfer</summary>

With new log system (without `VERBOSE`):
- 47.909 s
- 48.228 s
- 46.515 s
- 47.181 s
- 47.807 s

Without new log system:
- 47.311 s
- 48.783 s
- 48.184 s
- 47.092 s
- 47.667 s

-> no obvious performance decline
</details>

#### Exported logs from LLD

<details>
  <summary>Example separation `renderer` and `internal` thread</summary>

LLD_export_logs_separation

</details>

#### Stdout debug logs on LLD with `VERBOSE`

<details>
  <summary>LLD logs on terminal</summary>

LLD_stdout_logs_VERBOSE

</details>

#### Stdout debug logs on LLM with `VERBOSE`

<details>
  <summary>LLM logs on terminal</summary>

LLM_stdout_ex

</details>

### 🚀 Expectations to reach

[LIVE-9405]: https://ledgerhq.atlassian.net/browse/LIVE-9405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/main/setup.ts
+++ b/apps/ledger-live-desktop/src/main/setup.ts
@@ -8,6 +8,17 @@ import updater from "./updater";
 import resolveUserDataDirectory from "~/helpers/resolveUserDataDirectory";
 import path from "path";
 import { InMemoryLogger } from "./logger";
+import { setEnvUnsafe } from "@ledgerhq/live-env";
+
+/**
+ * Sets env variables for the main thread.
+ *
+ * The renderer thread will also set some env variables via the `setEnv` IPC channel
+ * but we might need some envs before the renderer thread is spawned.
+ */
+for (const k in process.env) {
+  setEnvUnsafe(k, process.env[k]);
+}
 
 ipcMain.on("mainCrashTest", () => {
   captureException(new Error("CrashTestMain"));
--- a/apps/ledger-live-desktop/src/main/setup.ts
+++ b/apps/ledger-live-desktop/src/main/setup.ts
@@ -7,17 +7,44 @@ import fs from "fs/promises";
 import updater from "./updater";
 import resolveUserDataDirectory from "~/helpers/resolveUserDataDirectory";
 import path from "path";
+import { InMemoryLogger } from "./logger";
+
 ipcMain.on("mainCrashTest", () => {
   captureException(new Error("CrashTestMain"));
 });
+
 ipcMain.on("updater", (e, type) => {
   updater(type);
 });
+
+/**
+ * Saves logs from the renderer thread and logs recorded from the internal thread to a file.
+ */
 ipcMain.handle(
   "save-logs",
-  async (event, path: Electron.SaveDialogReturnValue, experimentalLogs: string) =>
-    !path.canceled && path.filePath && fs.writeFile(path.filePath, experimentalLogs),
+  async (_event, path: Electron.SaveDialogReturnValue, rendererLogs: unknown[]) => {
+    if (!path.canceled && path.filePath) {
+      const inMemoryLogger = InMemoryLogger.getLogger();
+      const internalLogs = inMemoryLogger.getLogs();
+      console.log(
+        `Saving ${rendererLogs.length} logs from the renderer thread and ${internalLogs.length} logs from the internal thread`,
+      );
+
+      // Merging according to a `date` (internal logs) / `timestamp` (most of renderer logs) does not seem necessary.
+      // Simply pushes all the internal logs after the renderer logs.
+      // Note: this is not respecting the `EXPORT_MAX_LOGS` env var, but this is fine.
+      rendererLogs.push(
+        { type: "logs-separator", message: "Logs coming from the internal thread" },
+        ...internalLogs,
+      );
+
+      fs.writeFile(path.filePath, JSON.stringify(rendererLogs, null, 2));
+    } else {
+      console.warn("No path given to save logs");
+    }
+  },
 );
+
 ipcMain.handle(
   "export-operations",
   async (
@@ -39,6 +66,7 @@ ipcMain.handle(
     return false;
   },
 );
+
 const lssFileName = "lss.json";
 ipcMain.handle("generate-lss-config", async (event, data: string): Promise<boolean> => {
   const userDataDirectory = resolveUserDataDirectory();
@@ -52,6 +80,7 @@ ipcMain.handle("generate-lss-config", async (event, data: string): Promise<boole
   }
   return false;
 });
+
 ipcMain.handle("delete-lss-config", async (): Promise<boolean> => {
   const userDataDirectory = resolveUserDataDirectory();
   const filePath = path.resolve(userDataDirectory, lssFileName);
@@ -62,6 +91,7 @@ ipcMain.handle("delete-lss-config", async (): Promise<boolean> => {
   }
   return false;
 });
+
 ipcMain.handle("load-lss-config", async (): Promise<string | undefined | null> => {
   try {
     const userDataDirectory = resolveUserDataDirectory();
@@ -76,6 +106,7 @@ ipcMain.handle("load-lss-config", async (): Promise<string | undefined | null> =
   }
   return undefined;
 });
+
 process.setMaxListeners(0);
 
 // eslint-disable-next-line no-console
--- a/apps/ledger-live-desktop/src/main/setup.ts
+++ b/apps/ledger-live-desktop/src/main/setup.ts
@@ -33,10 +33,21 @@ ipcMain.on("updater", (e, type) => {
  */
 ipcMain.handle(
   "save-logs",
-  async (_event, path: Electron.SaveDialogReturnValue, rendererLogs: unknown[]) => {
+  async (_event, path: Electron.SaveDialogReturnValue, rendererLogsStr: string) => {
     if (!path.canceled && path.filePath) {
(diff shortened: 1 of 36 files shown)


apps/ledger-live-desktop/src/main/setup.ts
```diff
@@ -7,17 +7,44 @@ import fs from "fs/promises";
...
+      const internalLogs = inMemoryLogger.getLogs();
+      console.log(
+        `Saving ${rendererLogs.length} logs from the renderer thread and ${internalLogs.length} logs from the internal thread`,
+      );
+
+      // Merging according to a `date` (internal logs) / `timestamp` (most of renderer logs) does not seem necessary.
+      // Simply pushes all the internal logs after the renderer logs.
+      // Note: this is not respecting the `EXPORT_MAX_LOGS` env var, but this is fine.
+      rendererLogs.push(
+        { type: "logs-separator", message: "Logs coming from the internal thread" },
```
gre, line 37: can you double check that such log file with only "type" and "message" is going to not break the logsviewer?

apps/ledger-live-desktop/src/main/setup.ts
```diff
@@ -7,17 +7,44 @@ import fs from "fs/promises";
...
-    !path.canceled && path.filePath && fs.writeFile(path.filePath, experimentalLogs),
+  async (_event, path: Electron.SaveDialogReturnValue, rendererLogs: unknown[]) => {
+    if (!path.canceled && path.filePath) {
+      const inMemoryLogger = InMemoryLogger.getLogger();
+      const internalLogs = inMemoryLogger.getLogs();
+      console.log(
+        `Saving ${rendererLogs.length} logs from the renderer thread and ${internalLogs.length} logs from the internal thread`,
+      );
+
+      // Merging according to a `date` (internal logs) / `timestamp` (most of renderer logs) does not seem necessary.
```
gre, line 33: the only tradeoff is we may miss context of things by doing so, like we can't understand the relation because user actions (renderer) and the internal process behavior.

do we at least still have a correct date set on the logs from internal?

apps/ledger-live-desktop/src/renderer/IPCTransport.ts
```diff
@@ -87,18 +88,18 @@ export class IPCTransport extends Transport {
...
   async exchange(apdu: Buffer): Promise<Buffer> {
     const apduHex = apdu.toString("hex");
-    log("apdu", "=> " + apduHex);
+    log("ipc-apdu", "=> " + apduHex);
     const responseHex = await rendererRequest(transportExchangeChannel, {
       descriptor: this.id,
       apduHex,
     });
-    log("apdu", "<= " + responseHex);
+    log("ipc-apdu", "<= " + responseHex);
```
gre, line 96: i remember the logsviewer have a specific code have an "extract apdus" button that will recreate a apdu.txt on the logs of type "apdu", but i imagine they were now duplicated and so you had to do the ipc-adpu one?

apps/ledger-live-desktop/src/renderer/live-common-setup.ts
```diff
@@ -11,11 +11,19 @@ import { setEnvOnAllThreads } from "./../helpers/env";
...
+import { ipcRenderer } from "electron";
+
 listenLogs(({ id, date, ...log }) => {
+  // console.log(`🐬 listenLogs: ${JSON.stringify(log)}`);
   if (log.type === "hid-frame") return;
+
   logger.debug(log);
 });
+
+ipcRenderer.on("log:internal", event => {
```
gre, line 23: is this event coming? i'm not seeing where it gets emitted.
i'm a bit scared on the performance of this one. (whereas if we only log on main, it's ok, but here it's going to make the UI a bit more busy in heavy log times)
could you record in the performance tab the impact of this part when you install an app?

libs/ledger-live-common/src/hw/index.ts
```diff
@@ -17,30 +21,47 @@ export type Discovery = Observable<DeviceEvent>;
...
+   * you can typically prefix it with `something|` that identify it globally
+   * returns falsy if the transport module can't handle this id
+   * here, open means we want to START doing something with the transport
+   *
+   * @param deviceId
+   * @param timeoutMs Optional timeout that can be used by the Transport implementation when opening a communication channel
+   * @param context Optional context to be used in logs/tracing
+   */
+  open: (
+    deviceId: string,
```
gre, line 35: we may want to keep calling it "id" because it's not a "deviceId" as in "device id of the transport" but it's the userland that defines whatever the semantic of this "id" is. (LLM/LLD/cli/tools,.. can all decide how they decouple the id and the transport instance)

gre (commented):
great rework overall. some minor comments

apps/ledger-live-desktop/src/renderer/live-common-setup.ts
alexandremgo: sorry - it was an experiment, i removed it but you reviewed it too quickly ahah thx ! No logs are transfered from the `internal` (or `main`) threads to the `renderer` thread. There is no need.

The only transfers happening are: logs from `internal` to `main` so they are recorded

apps/ledger-live-desktop/src/renderer/IPCTransport.ts
alexandremgo: Exactly, we would get `apdu` from the IPCTransport and `apdu` from the actual `Transport`

libs/ledger-live-common/src/hw/index.ts
alexandremgo: You're right ! I put back `id` ;)

apps/ledger-live-desktop/src/main/setup.ts
alexandremgo: Right ! I'm going to test this

apps/ledger-live-desktop/src/main/setup.ts
alexandremgo: We have a correct data set on the logs from internal

But there is a possibility that a log coming from `renderer` has the same timestamp than a log coming from the `internal` thread: in that case, how do we decide which logs happened first ?

apps/ledger-live-desktop/src/main/setup.ts
alexandremgo: Screenshot 2023-09-18 at 15 44 46

It's still working with it ;)

apps/ledger-live-desktop/src/main/setup.ts
gre: i think it's ok for the time being, in future it wouldn't be too hard to sort by date if we want to, or even make it done on the logsviewer instead of inplace anyway. 
cheers

gre (approved):

PR #4705, opened 2023-09-15
Bugfix/live 8677 esc infinite loader

ofreyssinet-ledger:
### 📝 Description

wip

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #4777, opened 2023-09-20
Bugfix Stax LLD early security checks race condition

ofreyssinet-ledger:
### 📝 Description

**Issue**: unseeded Stax device, device not on the "early security checks" screen
When starting the early security checks, there is a race condition that manifests with an error drawer:

https://github.com/LedgerHQ/ledger-live/assets/91890529/a50fb33e-bb3f-4b6d-81c7-6bc33b2304bd

This issue is not blocking but we would rather avoid this.
It is caused by what I added in #4705 (the error handler in the `getDeviceInfo` that is used to determine if the device is in bootloader mode).

**The fix**:

This `getDeviceInfo` was necessary in case the user interrupted the fw update by unplugging the device at the wrong moment, resulting in the device being in bootloader mode. We would perform this check when the `device` object changes (so for instance when the device gets plugged back in). It was resulting in a race condition because the first time the `device` object changes we also send an APDU to put the device on the "early security checks" screen (`toggleEarlyOnboardingCheck`).

Now, we remove this check so the race condition cannot happen anymore.
And to handle the case in which the user cancels the firmware update and it results in the device being in bootloader mode, when the fw update drawer gets closed, we simply restart the onboarding state polling mechanism which will itself detect if the device is in bootloader mode, and then we get a `fatalError` which we use to trigger the "repair" mechanism (`isBootloader` is set to true and we mount a device action that puts the device out of bootloader mode).

### ❓ Context

- **Impacted projects**: `lld` 
- **Linked resource(s)**: [LIVE-9495] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-9495]: https://ledgerhq.atlassian.net/browse/LIVE-9495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/index.tsx
@@ -15,12 +15,8 @@ import { RecoverState } from "~/renderer/screens/recover/Player";
 import SyncOnboardingCompanion from "./SyncOnboardingCompanion";
 import EarlySecurityChecks from "./EarlySecurityChecks";
 import { setDrawer } from "~/renderer/drawers/Provider";
-import { withDevice } from "@ledgerhq/live-common/hw/deviceAccess";
-import { DeviceInfo } from "@ledgerhq/types-live";
 import connectManager from "@ledgerhq/live-common/hw/connectManager";
 import { mockedEventEmitter } from "~/renderer/components/debug/DebugMock";
-import { from } from "rxjs";
-import getDeviceInfo from "@ledgerhq/live-common/hw/getDeviceInfo";
 import { getEnv } from "@ledgerhq/live-env";
 import ExitChecksDrawer, {
   Props as ExitChecksDrawerProps,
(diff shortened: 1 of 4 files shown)


apps/ledger-live-desktop/src/renderer/components/SyncOnboarding/Manual/index.tsx
```diff
@@ -87,11 +82,18 @@ const SyncOnboardingScreen: React.FC<SyncOnboardingScreenProps> = ({
   const [toggleOnboardingEarlyCheckType, setToggleOnboardingEarlyCheckType] = useState<
     null | "enter" | "exit"
   >(null);
+  const [fwUpdateInterrupted, setFwUpdateInterrupted] = useState<FinalFirmware | null>(null);
```
ofreyssinet-ledger, line 85: Pulling this state out of the early security check component because it gets unmounted when we restart the polling after the user restarts an update.

alexandremgo (approved):
LGTM !

PR #4782, opened 2023-09-21
fix(LLD/LLM CounterValues): Changed supported CounterValues source to API

mle-gall:
### 📝 Description

Changed the CounterValue source to support only countervalues supported by our provider.

### ❓ Context

- Impacted projects: LLD, LLM, coin-framework
- Linked resource(s): [LIVE-2328]

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

N/A

### 🚀 Expectations to reach

[LIVE-2328]: https://ledgerhq.atlassian.net/browse/LIVE-2328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/reducers/settings.ts
+++ b/apps/ledger-live-desktop/src/renderer/reducers/settings.ts
@@ -105,6 +105,7 @@ export type SettingsState = {
   };
   featureFlagsButtonVisible: boolean;
   vaultSigner: VaultSigner;
+  supportedCounterValues: SupportedCountervaluesData[];
 };
 
 export const getInitialLanguageAndLocale = (): { language: Language; locale: Locale } => {
@@ -189,6 +190,7 @@ const INITIAL_STATE: SettingsState = {
 
   // Vault
   vaultSigner: { enabled: false, host: "", token: "", workspace: "" },
+  supportedCounterValues: [],
 };
 
 /* Handlers */
@@ -237,6 +239,7 @@ type HandlersPayloads = {
     featureFlagsButtonVisible: boolean;
   };
   SET_VAULT_SIGNER: VaultSigner;
+  SET_SUPPORTED_COUNTER_VALUES: SupportedCountervaluesData[];
 };
 type SettingsHandlers<PreciseKey = true> = Handlers<SettingsState, HandlersPayloads, PreciseKey>;
 
@@ -265,12 +268,6 @@ const handlers: SettingsHandlers = {
     };
   },
   FETCH_SETTINGS: (state, { payload: settings }) => {
-    if (
-      settings.counterValue &&
-      !supportedCountervalues.find(({ currency }) => currency.ticker === settings.counterValue)
-    ) {
-      settings.counterValue = INITIAL_STATE.counterValue;
-    }
     return {
       ...state,
       ...settings,
@@ -396,6 +393,10 @@ const handlers: SettingsHandlers = {
     ...state,
     vaultSigner: payload,
   }),
+  SET_SUPPORTED_COUNTER_VALUES: (state: SettingsState, { payload }) => ({
+    ...state,
+    supportedCounterValues: payload,
+  }),
 };
 export default handleActions<SettingsState, HandlersPayloads[keyof HandlersPayloads]>(
   handlers as unknown as SettingsHandlers<false>,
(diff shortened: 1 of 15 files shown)


mcayuelas-ledger (commented):
LGMT

apps/ledger-live-desktop/src/renderer/reducers/settings.ts
```diff
@@ -460,22 +469,23 @@ const defaultsForCurrency: (a: Currency) => CurrencySettings = crypto => {
...
-  ...possibleIntermediaries,
-]
-  .map(currency => ({
-    value: currency.ticker,
-    label: `${currency.name} - ${currency.ticker}`,
-    currency,
-  }))
-  .sort((a, b) => (a.currency.name < b.currency.name ? -1 : 1));
 
+export const getsupportedCountervalues = async (): Promise<supportedCountervaluesData[]> => {
```
gre, line 478: typo `s` -> `S` on the get**s**upportedCountervalues function and **s**upportedCountervaluesData type

libs/coin-framework/src/currencies/support.ts
```diff
@@ -63,15 +61,62 @@ setSupportedFiats([
...
-export function listSupportedFiats(): FiatCurrency[] {
+
+export async function listSupportedFiats(): Promise<FiatCurrency[]> {
+  if (userSupportedFiats === null) {
+    // Handle case where userSupportedFiats is not yet populated (e.g., by calling initializeUserSupportedFiats)
+    try {
+      await initializeUserSupportedFiats();
+    } catch (error) {
+      // Handle initialization error
+      console.error("Failed to initialize userSupportedFiats:", error);
```
gre, line 115: this is not really tracked like this. we should use `log` so it's included in exported logs at least.

gre (changes requested):
as far as I could read in the code, unfortunately the current behaviour of this implementation is that it is going to block the app to load before the data is fetched. and to me it's not acceptable to block the app load (it is not acceptable behaviour imo. today we never block on a API to be available to load the app)

basically the fact we do await the listSupportedFiats() before we even display the app, will only slow down the appearing of the app. if backend is slow, all LLD will be hanging very long at boot. basically we need to be more resilient especially for such "minor" feature, i think we could allow it to happen asynchronously.

on top of this, I also see that each time we call `listSupportedFiats()` (which used to be a sync call without side effect) , it's going to do an API call => this should probably be cached, because i don't think this data is going to change a lot. if Live could restore the latest loaded data (e.g. in localStorage, like we did for coins' preload() implementation), it would makes this even faster for the app to have the information (regardless if it blocks the load or not)

gre (changes requested):
also we need to test our PRs 🙏 .
it's possible to test and assert that a http call is done and is successful, see for example `analytics/segment.spec.ts` that tests that segment analytics is connected properly.

apps/ledger-live-desktop/src/renderer/reducers/settings.ts
```diff
@@ -265,12 +268,6 @@ const handlers: SettingsHandlers = {
     };
   },
   FETCH_SETTINGS: (state, { payload: settings }) => {
-    if (
-      settings.counterValue &&
-      !supportedCountervalues.find(({ currency }) => currency.ticker === settings.counterValue)
-    ) {
-      settings.counterValue = INITIAL_STATE.counterValue;
```
gre, line 272: this mechanism that was in place was to secure the fact a user, coming with a Countervalues from the past app state, may no longer have that countervalue available (supported) anymore by us.
In the new implementation, I don't see where the same mecanism is preserved. 

A way would be to move this previous logic into the handling of `SET_SUPPORTED_COUNTER_VALUES ` that receives the fetched list and the logic can now possibly re-set the counterValue to a default in case user's countervalue is no longer possible.

gre (changes requested):

apps/ledger-live-desktop/src/renderer/reducers/settings.ts
mle-gall: From what I saw that mechanism is actually handled elsewhere, by the way we get the active counter value. I've tested by removing the selected one from the response locally and it falls back as expected.

PR #4790, opened 2023-09-21
[BUGFIX][LLM] wrong fee value for evm custom fee

chabroA:
### 📝 Description

Update custom strategy saving logic for evm on LLM to display the correct fee value to the user when he selects the saved custom strategy.

```
   * When the user edits the custom fees, we save the related transaction patch
   * in order to reapply it the next time the user selects the saved custom
   * strategy entry in the strategy list.
   * This is done in order to display the correct fees value for the saved custom
   * entry in the fees strategy list.
   * We don't need to save the transaction patch for other fees strategies since
   * the transaction is already updated with the correct fee data retrieved from
   * the gasTracker in `prepareTransaction`.
   * cf. (as of 21-09-2023) https://github.com/LedgerHQ/ledger-live/blob/e87c4c2d879fb9322cd2849f2d34c3be9e500a1e/libs/coin-evm/src/prepareTransaction.ts#L190-L202
   * But since the custom strategy is handled locally by the client, if we don't
   * save the patch (including updated fee data) and only update the `feesStrategy`
   * field, the fees value displayed in the fees strategy for the custom entry
   * will be the one from the last selected strategy (e.g. slow, medium, fast).
```

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9507

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

| Before | After |
|--------|--------|
| <video src="https://github.com/LedgerHQ/ledger-live/assets/9203826/7178b629-2ee0-40f3-ba69-8e2ec2599bbd">  | <video src="https://github.com/LedgerHQ/ledger-live/assets/9203826/aa1dcc93-10bc-4028-9094-e7b3122f3834"> | 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
+++ b/apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
@@ -11,9 +11,9 @@ import React, { useCallback, useEffect, useState } from "react";
 import { ScreenName } from "../../const";
 import { EvmNetworkFeeInfo } from "./EvmNetworkFeesInfo";
 import SelectFeesStrategy from "./SelectFeesStrategy";
-import { SendRowsFeeProps as Props } from "./types";
+import { SendRowsFeeProps as Props, StrategyWithCustom } from "./types";
 
-const getCustomStrategy = (transaction: Transaction): BigNumber | null => {
+const getCustomStrategyFees = (transaction: Transaction): BigNumber | null => {
   if (transaction.feesStrategy === "custom") {
     return getEstimatedFees(transaction);
   }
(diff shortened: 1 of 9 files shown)


apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
```diff
@@ -54,26 +54,53 @@ export default function EvmFeesStrategy({
...
-    const newCustomStrategy = getCustomStrategy(transaction);
-
-    if (newCustomStrategy) {
-      setCustomStrategy(newCustomStrategy);
+    if (customStrategyTransactionPatch && transaction.feesStrategy === "custom") {
+      const newCustomStrategyFees = getCustomStrategyFees(transaction);
+      setCustomStrategyFees(newCustomStrategyFees);
     }
-  }, [transaction, setCustomStrategy]);
+  }, [transaction, customStrategyTransactionPatch]);
```
chabroA, line 81: We could remove the use of this `useState` and `useEffect` here if we go with the patch proposed bellow

apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
```diff
@@ -104,7 +132,7 @@ export default function EvmFeesStrategy({
     <SelectFeesStrategy
       {...props}
       gasOptions={gasOptions}
-      customFees={customStrategy}
+      customFees={customStrategyFees}
```
chabroA, line 135: We could simplify a bit the logic and derive the `customFees` value from the saved patch without having to store it in a new state variable by going with this suggestion.
The only minor issue is that we have to cast the object as a `Transaction` because it looks like TS complains about a compatibility issue between `transaction` and `customStrategyTransactionPatch` modes. Which is weird since `transaction` is of type `Transaction` and `customStrategyTransactionPatch` is of type `Partial<Transaction>` 🤔 

```
          Types of property 'mode' are incompatible.
            Type '"send" | "erc721" | "erc1155"' is not assignable to type '"erc721" | "erc1155"'.
              Type '"send"' is not assignable to type '"erc721" | "erc1155"'.
```

```suggestion
      customFees={getCustomStrategyFees({
        ...transaction,
        ...customStrategyTransactionPatch,
      } as Transaction)}
```

hedi-edelbloute: Calling a method like this is viewed as bad practice in Angular because it causes rerenders for no reason, is it fine in React ?

To answer to your question maybe it is just a type error, we can see on one side that the type isn't complete and misses 'send' value

since one overrides the other i think it should be fine but you should check why it's happening

hedi-edelbloute (approved):
Looks fine to me but I don't have the global "picture" of it

apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
lvndry: if you use `Pick<Transaction, "feesStrategy">` insead of `Partial<Transaction>` this will work. I agree the error shouldn't be raise but it kinda make sense

lvndry (approved):

apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
chabroA: > Calling a method like this is viewed as bad practice in Angular because it causes rerenders for no reason, is it fine in React ?

I don't see how it would cause rerenders.
It would cause unnecessary rerenders if it was an action function updating state value. Which is not the case here.

This function is just formatting value (more precisely, returning a computation of 2 values from `transaction`, or `null`).

One could decide to put this computation in a `useMemo` like so:

```ts
  const customFees = useMemo(
    () =>
      getCustomStrategyFees({
        ...transaction,
        ...customStrategyTransactionPatch,
      } as Transaction),
    [transaction, customStrategyTransactionPatch],
  );

....

  return (
    <SelectFeesStrategy
      customFees={customFees}
      ....
    />
```

But this is not necessary in this case since the memoised computation is not expensive. The cost of memoisation would probably exceed any "gain" (if any, since the memoised value would be invalidated and trigger a rerender each time `transaction` or `customStrategyTransactionPatch` anyway)
cf. https://react.dev/reference/react/useMemo#should-you-add-usememo-everywhere

> if you use Pick<Transaction, "feesStrategy"> insead of Partial<Transaction> this will work. I agree the error shouldn't be raise but it kinda make sense

I guess you are talking about this type:

```ts
  const [customStrategyTransactionPatch, setCustomStrategyTransactionPatch] =
    useState<Partial<Transaction>>();
```

But `customStrategyTransactionPatch` is not of type `feesStrategy`, it also includes other props of a transaction  related to fees (cf. where `setCustomStrategyTransactionPatch` is used under `apps/ledger-live-mobile/src/families/evm/EvmCustomFees`

lambertkevin (approved):

PR #4792, opened 2023-09-21
fix: add account wallet id conversion for buy deeplinks

sarneijim:
### 📝 Description

Add account wallet id conversion for buy deeplinks

### ❓ Context

- **Impacted projects**: LLM LLD
- **Linked resource(s)**:https://ledgerhq.atlassian.net/browse/LIVE-9482

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
@@ -4,11 +4,14 @@ import { useSelector } from "react-redux";
 import { RampCatalog } from "@ledgerhq/live-common/platform/providers/RampCatalogProvider/types";
 import Card from "~/renderer/components/Box/Card";
 import { languageSelector } from "~/renderer/reducers/settings";
+import { accountsSelector } from "~/renderer/reducers/accounts";
 import { useRemoteLiveAppManifest } from "@ledgerhq/live-common/platform/providers/RemoteLiveAppProvider/index";
 import useTheme from "~/renderer/hooks/useTheme";
 import { useLocalLiveAppManifest } from "@ledgerhq/live-common/platform/providers/LocalLiveAppProvider/index";
 import WebPTXPlayer from "~/renderer/components/WebPTXPlayer";
+import { getParentAccount, isTokenAccount } from "@ledgerhq/live-common/account/index";
 import { LiveAppManifest, Loadable } from "@ledgerhq/live-common/platform/types";
+import { accountToWalletAPIAccount } from "@ledgerhq/live-common/wallet-api/converters";
 import {
   DEFAULT_MULTIBUY_APP_ID,
   INTERNAL_APP_IDS,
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
```diff
@@ -33,6 +38,19 @@ const LiveAppExchange = ({ appId }: { appId: string }) => {
   const manifest = localManifest || mockManifest || remoteManifest;
   const themeType = useTheme().colors.palette.type;
 
+  // Use wallet-api ids if it is using v2 multibuy manifest
+  if (urlParams && manifest?.id === DEFAULT_MULTIBUY_APP_ID) {
```
cpellizzari-ledger, line 42: This would not work : we do the switch between wallet api and platform sdk inside the manifest, changing its content but keeping the same id.
ex : if we want to rollback to platform sdk, we will modify the manifest that has this same id, to point to old url and use platform sdk.

You should use another manifest field, I suppose the api version : https://github.com/LedgerHQ/manifest-api/blob/main/manifests/multibuy-v2/manifest.json#L10

apps/ledger-live-mobile/src/screens/PTX/BuyAndSell/index.tsx
```diff
@@ -43,6 +51,18 @@ export function BuyAndSellScreen({ route }: Props) {
   useEffect(
     () => {
       (async () => {
+        // Use wallet-api ids if it is using v2 multibuy manifest
+        if (manifest?.id === DEFAULT_MULTIBUY_APP_ID) {
```
cpellizzari-ledger, line 55: (same comment than desktop)

cpellizzari-ledger (changes requested):

cpellizzari-ledger (approved):

apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
```diff
@@ -33,6 +38,24 @@ const LiveAppExchange = ({ appId }: { appId: string }) => {
...
+    const account = accounts.find(a => a.id === accountId);
+    if (account) {
+      const parentAccount = isTokenAccount(account)
+        ? getParentAccount(account, accounts)
+        : undefined;
+      const walletApiId = accountToWalletAPIAccount(account, parentAccount)?.id;
+      urlParams.account = walletApiId;
+    }
+  }
+
```
Justkant, line 58: I would have put this in a `useEffect`

Also it would be better to use our constant for `WALLET_API_VERSION` and check if the manifest satisfies the version using the semver package like we do here: https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/components/Web3AppWebview/index.tsx#L10-L11
```tsx
import semver from "semver";

if (semver.satisfies(WALLET_API_VERSION, manifest.apiVersion)) {
  // your logic
}
```

PR #4801, opened 2023-09-21
[bugfix] Remove legacy hack for fees of Decred / Qtum / Viacoin

haammar-ledger:
### 📝 Description

Until now a hack was required on Live side to get around an issue with the `/fees` endpoint of the Bitcoin explorer, only for Decred, Qtum and Viacoin.
It appears the /fees endpoint has been fixed at some point and the hack is now redundant, causing the issue in the linked tickets.
Removing the hack solves the issue.

### ❓ Context

- **Linked resource(s)**:
  - https://ledgerhq.atlassian.net/browse/LIVE-10165
  - https://ledgerhq.atlassian.net/browse/LIVE-9497
  - https://ledgerhq.atlassian.net/browse/LIVE-9500

### ✅ Checklist

- [ ] **Test coverage** :x: depending on correctness of backend-provided data, which we don't validate at the moment  
- [x] **Atomic delivery** 
- [x] **No breaking changes**


hzheng-ledger (approved):
looks good

haammar-ledger:
On hold until further notice, as we need to clarify the migration plan with backend.
As a quick fix backend team are reverting to their previous (buggy) implem to unlock users, then we'll need to plan a proper, seamless migration on both backend and LL side.

lvndry (approved):

haammar-ledger:
Following discussions with backend they provided a fix on staging -- this PR can be validated using `https://explorers.api-01.live.ledger-stg.com`

:warning: HOWEVER the PR cannot be merged unless the fix on backend side is compatible with both before and after this fix...

Proper migration plan can only be one of:
- backend implements the fix on a **new** endpoint, different from `/network`
- we implement some temporary hack to support both, instead of just removing the legacy logic

chabroA (approved):

PR #4841, opened 2023-09-26
fix: refactor ptx adding correct account to query string

cksanders:
### 📝 Description

 refactor ptx adding correct account to query string

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
@@ -4,11 +4,14 @@ import { useSelector } from "react-redux";
 import { RampCatalog } from "@ledgerhq/live-common/platform/providers/RampCatalogProvider/types";
 import Card from "~/renderer/components/Box/Card";
 import { languageSelector } from "~/renderer/reducers/settings";
+import { accountsSelector } from "~/renderer/reducers/accounts";
 import { useRemoteLiveAppManifest } from "@ledgerhq/live-common/platform/providers/RemoteLiveAppProvider/index";
 import useTheme from "~/renderer/hooks/useTheme";
 import { useLocalLiveAppManifest } from "@ledgerhq/live-common/platform/providers/LocalLiveAppProvider/index";
 import WebPTXPlayer from "~/renderer/components/WebPTXPlayer";
+import { getParentAccount, isTokenAccount } from "@ledgerhq/live-common/account/index";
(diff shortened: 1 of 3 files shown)


sarneijim (approved):

apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
```diff
@@ -33,6 +42,28 @@ const LiveAppExchange = ({ appId }: { appId: string }) => {
...
 
+  /**
+   * Pass correct account ID
+   * Due to Platform SDK account ID not being equivalent to Wallet API account ID
+   */
+  const customUrlParams = useMemo(() => {
+    if (
+      urlParams?.account &&
+      manifest?.apiVersion &&
+      semver.satisfies(WALLET_API_VERSION, manifest.apiVersion)
```
sarneijim, line 53: I am not sure if the validation is the correct one.

semver.satisfies(manifest.apiVersion, "2.x" )

ggilchrist-ledger (approved):

apps/ledger-live-desktop/src/renderer/screens/exchange/index.tsx
cksanders: tested and works ok, we can extend it later if needed.

Justkant:
```tsx
      semver.satisfies(WALLET_API_VERSION, manifest.apiVersion)
```
This code is correct and `sermver.satisfies` allows for more versions to be checked thanks to ranges, this is why every wallet-api app is setup with `apiVersion: "^2.0.0"` meaning that if we update the constant `WALLET_API_VERSION` in LL, your app will still work

Justkant: If you want to have a manifest only for a specific wallet-api version it is possible if you omit the syntax for ranges like `^`, but this need more testing as we never had this use case before

cksanders: We can have a chat tomorrow and get a ticket created but will have to consider its priority.

PR #2499, opened 2023-02-01
Edit/Cancel Ethereum Transaction - LLM part

lvndry:
### 📝 Description

- Addition of a resend flow to speedup an ethereum transaction
- Addition of flow to cancel a transaction
 
### ❓ Context

- **Impacted projects**: `LLM`
- **Linked resource(s)**:
  - https://ledgerhq.atlassian.net/browse/LIVE-4317
  - https://ledgerhq.atlassian.net/browse/LIVE-4318
  - https://ledgerhq.atlassian.net/browse/LIVE-5066

### ✅ Checklist

- [x] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 📸 Demo

N/A

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #2441, opened 2023-01-27
Edit/Cancel Ethereum Transaction - LLD part

hzheng-ledger:
### 📝 Description

Add the feature to speedup/cancel ongoing transaction. See jira tickets for more details

### ❓ Context

- **Impacted projects**: LLD
- **Linked resource(s)**:
- https://ledgerhq.atlassian.net/browse/LIVE-4295
- https://ledgerhq.atlassian.net/browse/LIVE-4297
- https://ledgerhq.atlassian.net/browse/LIVE-5065
- https://ledgerhq.atlassian.net/browse/LIVE-5067
- https://ledgerhq.atlassian.net/browse/LIVE-5068
- https://ledgerhq.atlassian.net/browse/LIVE-5069
- https://ledgerhq.atlassian.net/browse/LIVE-5070

### ✅ Checklist

- [ ] **Test coverage** 
- [X] **Atomic delivery** 
- [X] **No breaking changes** 

### 📸 Demo

Pending transaction (less than 5 minutes) account page
截屏2023-01-27 10 20 08

Stuck transaction (more than 5 minutes) account page

Stuck transaction panel in account page

Transaction detail page (pending transaction more than 5 minutes)

Transaction detail page (pending transaction less than 5 minutes)

Adjust transaction fee to speed up transaction
截屏2023-01-27 10 20 43

Cancel transaction summary screen
截屏2023-01-27 10 20 30

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #4886, opened 2023-09-29
[FEATURE][LIVE-8957] migrate speedup cancel to evm

chabroA:
### 📝 Description

Once a transaction is broadcasted to the network, if a user wants to speed it up or cancel it, this transaction needs to be replaced by another one with higher fees.
- If the user wants to speed up a stuck transaction, because he broadcasted it with low fees compared to network activity and / or because network activity surged since the time he crafted his transaction, he will need to broadcast the exact same transaction (with the same nonce) but with higher fees. And optimally fees high enough compared to current network conditions.
- If the user wants to cancel a pending transaction, he will need to send a new transaction, sending 0 ETH to his own address, with the same nonce as the original transaction and with higher fees.

There is a non-standard minimum 10% bump in fees that need to be applied to the original fees in order for nodes to pick up the new transaction. If the fees are increased by less than 10%, a _ReplacementTransactionUnderpriced_ error will be returned by the node upon broadcast.

Some references on this topic:
- https://docs.alchemy.com/docs/retrying-an-eip-1559-transaction
- https://support.metamask.io/hc/en-us/articles/360015489251

----

- Migrate the original implementation of the speedup / cancel feature for evm transactions (only evm currencies with `gasTracker` and `explorer` at the moment)
- Rework speedup and cancel flows to have common UX. Sensible default fees are selected depending on original tx fees and current network conditions but user can update fees if he so chooses, as long as they are still higher than the minimum amount required to replace a tx.

#### 🗒️ Notes

Integration tests are failing due to an unrelated error on the cardano bridge tests (`src/families/cardano/bridge.integration.test.ts`).
This error also happens on `develop` at the time of writing.

#### 🧪 How to test

In order to more easily test the different flows, without using real funds nor depending on network conditions (for example you want to test what happen if the tx you want to speedup get validated while you are in the middle of the "speedup" or "cancel" flow), you can use [`anvil`](https://book.getfoundry.sh/reference/anvil/) and [`cast`](https://book.getfoundry.sh/reference/cast/cast) from [foundry](https://book.getfoundry.sh/).

Here is a quick overview / how to / setup for the purpose of this feature (but these tools are quite interesting, versatile and powerful so I let you dig in their respective docs if you want to learn more):

<details><summary>Details</summary>

The first step is to install the foundry toolchain which will install the `anvil` and `cast` tools we will be using here.
cf. https://book.getfoundry.sh/getting-started/installation 
```sh
curl -L https://foundry.paradigm.xyz | bash
```

With these tools installed, you can now start `anvil`, which we will use to create a local testnet node (cf. [`anvil` doc](https://book.getfoundry.sh/reference/anvil/) for more info)

```sh
anvil -f https://eth.llamarpc.com --no-storage-caching --steps-tracing --host 0.0.0.0 --no-mining
```
This will create a local testnet node with a state forked from the public node `https://eth.llamarpc.com`, and the `--no-mining` option disable auto and interval mining, and mine on demand instead (the default behaviour of `anvil` is to mine a block every time a tx is broadcasted, which is not what we want here). We will see bellow how to manually request a block to be mined.

The nice thing with forking the state of the blockchain from a public node is that you get all your accounts with their respective balances, associated tokens and NFTs, on this testnet (so no need to manually create new accounts or other testing data).

Before moving on, you need to get you local IP address. You should be able to get it with this command:
```sh
ifconfig | grep "inet 192.168.1."
```
Which should yield a result like:
```sh
inet 192.168.1.26 netmask 0xffffff00 broadcast 192.168.1.255
```
In this example, the IP address is `192.168.1.26`

With this info in hand, you now need to tell LL to use this network instead of the prod one for the `ethereum` currency.
To do that, you just need to replace the `node` prop of the `ethereumLikeInfo` object of the `ethereum` currency definition under `libs/ledgerjs/packages/cryptoassets/src/currencies.ts`

This:
https://github.com/LedgerHQ/ledger-live/blob/cd0736ea0d5262073ca63435a3646e211d57878b/libs/ledgerjs/packages/cryptoassets/src/currencies.ts#L1108

Would become this:
```ts
      node: { type: "external", uri: "http://[LOCAL_IP_FOUND_ABOVE]:8545" },
```

With the example `192.168.1.26` IP address found above, it would get
```ts
      node: { type: "external", uri: "http://192.168.1.26:8545" },
```

You should now be able to start LL, which should run on the local testnet network.
To be sure of this, you can increase the balance of one of your ethereum account on the local test network, and see if it gets reflected in LL.

This is done using the `cast` command (more info [here](https://book.getfoundry.sh/reference/cast/cast)).

When the test network is created with the `anvil` command as done above, 10 test accounts with 10000 ETH each are created.
We will use one of these test account to send some funds to one of our own account on the local test network.

To get YOUR_ETH_ADDRESS, go in LL and take the ETH address of one of your account.

```sh
cast send YOUR_ETH_ADDRESS "" --value 10ether --from 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
```

Since we started our local network with the `--no-mining` option, the above `cast` command will hang until we manually ask for a block to be mined.

To do this, open a new terminal and run the following command:

```sh
cast rpc evm_mine
```

Go back to LL, wait a bit (or re-sync your accounts), if the account corresponding to `YOUR_ETH_ADDRESS` gets credited of 10 ETH, you connected to the local testnet and can start testing and sending transactions.

PS: 
- keep in mind that you will need to run `cast rpc evm_mine` each time you broadcast a tx for it to take effect on the testnet. But you can also use this to test various behaviours linked to the feature, like what happens if a tx is pending and I send another one on top? does a tx indeed become stuck after a specific delay? what happen if the tx I want to speedup / cancel gets validated while I'm in the edit flow (i.e: you run `cast rpc evm_mine` while in the edit flow in LL), etc...
- you can also change the `ETHEREUM_STUCK_TRANSACTION_TIMEOUT` env var value (default value is `5 * 60 * 1000` ms, or 5 minutes) to a lower value (like `0.5 * 60 * 1000`ms or 30 seconds) to more easily test / witness a pending tx becoming stuck.

</details> 

### ❓ Context

- #2441 
- #2499
- #5245
- #5258 
- https://github.com/LedgerHQ/ledger-live/pull/5331
- https://github.com/LedgerHQ/ledger-live/pull/5405

- **Impacted projects**: `LLD` `LLM` `LLC`
- **Linked resource(s)**:
	- https://ledgerhq.atlassian.net/browse/LIVE-9263 
    - https://ledgerhq.atlassian.net/browse/LIVE-8957
    - https://ledgerhq.atlassian.net/browse/LIVE-8863
    - https://ledgerhq.atlassian.net/browse/LIVE-8820

### ✅ Checklist

- [x] **Test coverage** - Tests added for the added and updated logic
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

<details><summary>LLD - stuck operation + banner</summary>

stuck operation + banner

</details> 

#### 🖥️  LLD flows

<details><summary>LLD - cancel ETH send tx</summary>

https://github.com/LedgerHQ/ledger-live/assets/9203826/bab5c4c3-accb-403e-8e15-071da9f9a97f

</details> 

<details><summary>LLD - speed-up ETH send tx</summary>

https://github.com/LedgerHQ/ledger-live/assets/9203826/72f36b2d-6acd-4f2f-ae3f-032b14eb8c6d

</details> 

<details><summary>LLD - cancel ERC20 send tx</summary>
<p>

https://github.com/LedgerHQ/ledger-live/assets/9203826/261328ad-7415-465e-bb88-da7a603eaf97

</p>
</details> 

<details><summary>LLD - speed-up ERC20 send tx</summary>
<p>

https://github.com/LedgerHQ/ledger-live/assets/9203826/908ff09d-542a-40fe-b251-e884fc2f07e2

</p>
</details> 

<details><summary>LLD - cancel ERC721 send tx</summary>
<p>

https://github.com/LedgerHQ/ledger-live/assets/9203826/97560d02-9a2f-4274-9646-d4ed2319e5a3

</p>
</details> 

<details><summary>LLD - speedup ERC721 send tx</summary>
<p>

https://github.com/LedgerHQ/ledger-live/assets/9203826/3b8cf545-37e4-4c5d-8651-eb3737648bf9

</p>
</details> 

<details><summary>LLD - cancel ERC1155 send tx</summary>
<p>

https://github.com/LedgerHQ/ledger-live/assets/9203826/6dd9e3e0-f626-402f-9470-6bb325c12e44

</p>
</details> 

<details><summary>LLD - speedup ERC1155 send tx</summary>
<p>

https://github.com/LedgerHQ/ledger-live/assets/9203826/57b9eb11-df40-4e4b-a3c3-b2dea1d7963e

</p>
</details> 

#### 📱   LLM flows

_Running the dev app on LLM emulator being incredibly slow and painful, only the cancel and speedup of ETH send flows have been recorded_

<details><summary>LLM - cancel ETH send tx</summary>
<p>

[LLM - cancel ETH send tx.webm](https://github.com/LedgerHQ/ledger-live/assets/9203826/14968ae1-0dc7-4879-b73c-4d90d1096842)

</p>
</details> 

<details><summary>LLM - speedup ETH send tx</summary>
<p>

[LLM - speedup ETH send tx.webm](https://github.com/LedgerHQ/ledger-live/assets/9203826/22f6a88f-ccb6-4625-9c2e-c338434629bb)

</p>
</details> 

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+  const onContinue = useCallback(() => {
+    if (warnings.feeTooHigh) {
+      setHighFeesOpen(true);
+      return;
+    }
+    navigateToNext();
+  }, [navigateToNext, warnings.feeTooHigh]);
+
+  // FIXME: why is recipient sometimes empty?
```
lvndry, line 144: is it still the case here ?

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
```diff
@@ -0,0 +1,342 @@
...
+          <LText style={styles.warning} color="orange">
+            <TranslatedError error={warnings.recipient} />
+          </LText>
+        ) : null}
+        <SendRowsCustom
+          transaction={transaction}
+          account={mainAccount}
+          // FIXME: fix typing
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          navigation={navigation as any}
```
lvndry, line 192: yeah typing on LLM is quite a mess, you need to update 5+ files to add 1 props

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/MethodSelection.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+      setSelectedMethod(option);
+    },
+    [mainAccount, transaction, transactionToEdit, bridge, setTransaction],
+  );
+
+  /**
+   * Poll for transaction confirmation
+   * If the transaction has been confirmed, we navigate to the error screen
+   * PS: this still runs even if the user has navigated forward in the steps
```
lvndry, line 118: When does it stop then ? It doesn't run after you leave the flow I guess ? So after broadcast ?
Maybe polling every n seconds (30 seconds by default) is a bit overkill as users will rarely stay on the step for minutes, if you ask once at every step if the transaction is confirmed or not it should be enough to catch.

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/MethodSelection.tsx
```diff
@@ -0,0 +1,240 @@
...
+   * "Cannot update a component while rendering a different component" error
+   * But it works for now and avoid code duplication 🤷‍♂️
+   * If you have a better way to do this, please do!
+   */
+  useEffect(() => {
+    const setTransactionHasBeenValidatedCallback = async () => {
+      const tx = await getTransactionByHash(mainAccount.currency, operation.hash);
+      if (tx?.confirmations) {
+        // stop polling as soon as we have a confirmation
+        clearInterval(intervalId);
```
lvndry, line 129: shouldn't we call this function after intervalId is defined ?

apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
```diff
@@ -7,7 +7,7 @@ import { log } from "@ledgerhq/logs";
 import { InfiniteLoader } from "@ledgerhq/native-ui";
 import { AccountBridge } from "@ledgerhq/types-live";
 import BigNumber from "bignumber.js";
-import React, { useCallback, useEffect, useState } from "react";
+import { default as React, useCallback, useEffect, useState } from "react";
```
lvndry, line 10: Does it change anything ? The former way of importing seems more what you would expect

libs/coin-evm/src/api/transaction/index.ts
```diff
@@ -0,0 +1,15 @@
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { getExplorerApi } from "../explorer";
+
+// TODO: implement getTransactionByHash for etherscan explorer
```
lvndry, line 4: can't be done in this PR ?

libs/coin-evm/src/api/explorer/ledger.ts
```diff
@@ -126,8 +127,29 @@ export const getLastOperations: ExplorerApi["getLastOperations"] = async (
   };
 };
 
+const getTransactionByHash = async (
+  currency: CryptoCurrency,
+  transactionHash: string,
+): Promise<{ confirmations?: number }> => {
+  const { explorer } = currency.ethereumLikeInfo || /* istanbul ignore next */ {};
```
lvndry, line 134:
```suggestion
  const { explorer } = currency.ethereumLikeInfo!;
```
I could be wrong but both have the same result

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+  };
+};
+
+const ERROR_NOT_ENOUGH_NFT_OWNED = new NotEnoughNftOwned();
+const ERROR_NOT_OWNED_NFT = new NotOwnedNft();
+const ERROR_AMOUNT_REQUIRED = new AmountRequired();
+
+// NOTE: this logic should ideally be done in getTransactionStatus if possible
+// (not sure it is)
+export const getEditTransactionStatus = ({
```
lvndry, line 112: shouldn't this be in `coin-evm` ?

libs/ledger-live-common/src/families/evm/editTransaction/getMinFees.ts
```diff
@@ -0,0 +1,68 @@
...
+export const getMinFees = ({
+  transaction,
+}: {
+  transaction: Transaction;
+}): {
+  gasPrice?: BigNumber;
+  maxFeePerGas?: BigNumber;
+  maxPriorityFeePerGas?: BigNumber;
+} => {
+  if (transaction.type === 2) {
```
lvndry, line 61: Yeah I'm personally a bit confused what means type 1,2,3,...
I feel like an enum would make it simpler to know what we're talking about

libs/coin-evm/src/api/transaction/index.ts
lambertkevin: Unecessary, a transaction can be fetched from a node with the rpc method `eth_gettransactionbyhash` which is provided by `ethers` with the `provider.getTransaction( hash )` helper

libs/coin-evm/src/api/explorer/ledger.ts
lambertkevin: Nope, the non-null assertion will result in an error thrown if `ethereumLikeInfo` is undefined, while the `|| {}` allow to throw (or not) in an expected manner later on the flow.

libs/coin-evm/src/api/transaction/index.ts
lambertkevin: It's also provided by the `node` api with [this](https://github.com/LedgerHQ/ledger-live/blob/bbad6146cee4fa55a8d0e2fec47579c945e6e59b/libs/coin-evm/src/api/node/types.ts#L7)

libs/ledger-live-common/src/families/evm/editTransaction/getMinFees.ts
lambertkevin: We already discussed this, but unfortunatley that's some necessary knowledge in order to contribute to the evm family. You can find some documentation [here](https://ethereum.org/fr/developers/docs/transactions/#typed-transaction-envelope) if needed listing the different types.
We can add an enum that will say that type 0 = legacy, type 1 = EIP-2718, type 2 = EIP-1559, but to me, it's not going to help anyone, as, if you don't know what type 1 or 2 are, then you don't know what EIP-2718 and EIP-1559 are.

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/MethodSelection.tsx
lambertkevin: clearInterval never throws, no matter the param you give it.

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -0,0 +1,340 @@
...
+    id: "confirmation",
+    label: <Trans i18nKey="operation.edit.steps.confirmation.title" />,
+    excludeFromBreadcrumb: true,
+    component: StepConfirmation,
+    footer: StepConfirmationFooter,
+    onBack: null,
+  },
+];
+
+const mapStateToProps = createStructuredSelector({
```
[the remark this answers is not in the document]
chabroA: It's a function from `reselect`, cf. https://github.com/reduxjs/reselect#createstructuredselectorinputselectors-selectorcreator--createselector
It seems to be used to create the `mapStateToProps` selector used by `react-redux` `connect` decorator (cf. bottom of same file)

FYI it's a pattern introduced more than 4 years ago in the codebase (cf. https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/modals/Send/Body.tsx#L121-L124)

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -0,0 +1,340 @@
...
+import { StepSummaryFooter } from "./steps/StepSummaryFooter";
+import { St, StepId, StepProps } from "./types";
+
+export type Params = {
+  account: AccountLike | undefined | null;
+  parentAccount: Account | undefined | null;
+  recipient?: string;
+  amount?: BigNumber;
+  transaction?: Transaction;
+  transactionRaw: TransactionRaw;
```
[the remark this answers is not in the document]
chabroA: Turns out we don't need `transaction` and only need the following params:
```
account
parentAccount
transactionRaw
transactionHash
```

cf. usage of this component here
https://github.com/LedgerHQ/ledger-live/blob/1f711610bb0e620067e8c7b457c5ba99c6b3b73b/apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Modal.tsx#L34-L44

Will remove the extra unnecessary params

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepFees.tsx
```diff
@@ -0,0 +1,158 @@
...
+  parentAccount,
+  transaction,
+  transactionToUpdate,
+  status,
+  bridgePending,
+  t,
+  onChangeTransaction,
+  updateTransaction,
+}: StepProps) => {
+  const mainAccount = account ? getMainAccount(account, parentAccount) : null;
```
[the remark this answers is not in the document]
chabroA: Yes. Actually the typing of `account` under `StepProps` is wrong
https://github.com/LedgerHQ/ledger-live/blob/1babf55a1232e89711ac69575805496e8615f58b/apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/types.ts#L12

`account` is always defined at this stage since it's existence is already tested in `Body`
https://github.com/LedgerHQ/ledger-live/blob/d2145d48d0ccf502bd905d181bcbcd8218c4e712/apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx#L179

Will fix the typing and remove the unnecessary tests

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
chabroA: No idea, this is a copy of `src/screens/SendFunds/04-Summary.tsx` file (with some changes)
https://github.com/LedgerHQ/ledger-live/blob/70b4603a6fa34453c6fa6dd89fb3ed12a1038366/apps/ledger-live-mobile/src/screens/SendFunds/04-Summary.tsx#L154

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/MethodSelection.tsx
chabroA: The polling stops when:
- the tx has been confirmed on the blockchain (`tx?.confirmations` is truthy)
- the component unmount (cf. `return` in `useEffect`), i.e the users exits the "EditTransactionFlow" stack

> Maybe polling every n seconds (30 seconds by default) is a bit overkill as users will rarely stay on the step for minutes, if you ask once at every step if the transaction is confirmed or not it should be enough to catch.

This was the original implementation but it's (IMHO) a naive and suboptimal implementation. Here is why:
Most EVM blockchains have pretty fast block times (i.e: time between each new block).
For ethereum for example, this is between 12- 15 seconds (probably more for other blockchains). A user is likely to stay more than 12-15 seconds on a step (mainly the fee step if he wants to update them, or the summary step if he is diligent in his verification).
In addition to that, verifying only once at each step, as originally done, would imply an (excessive) code duplication of this check logic at every step (with associated complexity in maintenance and potential issues / oversight if the flow is updated and a step is added for example).

The idea here is to centralize this verification process in one place, at the first step.
And since this flow is a stack navigation, this logic still runs when the user goes past this method selection step since what React Native do under the hood is adding a new screen (for the next step) on top of the previous once (hence the "stack" navigation name).

apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
chabroA: It does not change anything, seems like an auto formatting related change. Will revert it to stay consistent with the rest of the codebase 👍

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionStatus.ts
chabroA: Yes probably, could be moved in `libs/coin-evm/src/getTransactionStatus.ts`. Which would imply moving `libs/ledger-live-common/src/families/evm/editTransaction/types.ts` and `libs/ledger-live-common/src/families/evm/editTransaction/getMinFees.ts` under `coin-evm` as well since it depends on them as well and we can't have a dependency to `ledger-live-common` in `coin-evm`

Will take a look at that and propose a patch, see which solution is better

apps/ledger-live-desktop/src/renderer/components/OperationsList/DateCell.tsx
```diff
@@ -16,31 +17,33 @@ const Cell = styled(Box).attrs(() => ({
...
+        <WarningSolidMedium size={12} color={"#FFBD42"} />
+      </Box>
+    );
+  }
+
+  return <InfiniteLoader size={12} style={{ verticalAlign: "middle" }} />;
+};
+
 class DateCell extends PureComponent<Props> {
+  // FIXME: still needed? seems unused
```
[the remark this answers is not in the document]
chabroA: 5730077065fb128b9035b0e5d7e8b4828b217a86

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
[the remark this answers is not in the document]
chabroA: 63c770ea2583f91d8cc562a3e703e498c480f0de

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -0,0 +1,340 @@
...
+    transaction,
+    setTransaction,
+    updateTransaction,
+    account: accountFromBridge,
+    parentAccount: parentAccountFromBridge,
+    status,
+    bridgeError,
+    bridgePending,
+  } = useBridgeTransaction<Transaction>(() => {
+    const parentAccount = params && params.parentAccount;
```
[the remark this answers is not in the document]
chabroA: dbeef08db33f0c0ac41050e4e5a581e854861bb1

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -0,0 +1,340 @@
...
+    const account = (params && params.account) || accounts[0];
```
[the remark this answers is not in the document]
chabroA: dbeef08db33f0c0ac41050e4e5a581e854861bb1

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -0,0 +1,340 @@
...
+      updateAccountWithUpdater(mainAccount.id, account =>
+        addPendingOperation(account, optimisticOperation),
+      );
+      setOptimisticOperation(optimisticOperation);
+      setTransactionError(null);
+    },
+    [account, parentAccount, updateAccountWithUpdater],
+  );
+
+  const handleStepChange = useCallback(e => onChangeStepId(e.id), [onChangeStepId]);
```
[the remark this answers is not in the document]
chabroA: dbeef08db33f0c0ac41050e4e5a581e854861bb1

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepFees.tsx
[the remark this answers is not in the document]
chabroA: 63c770ea2583f91d8cc562a3e703e498c480f0de

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepFees.tsx
```diff
@@ -0,0 +1,158 @@
...
+    formattedFeeValue,
+    formattedMaxPriorityFeePerGas,
+    formattedMaxFeePerGas,
+    formattedGasPrice,
+  } = getFormattedFeeFields({ transaction: transactionToUpdate, mainAccount, locale });
+
+  return (
+    <Box flow={4}>
+      <CurrencyDownStatusAlert currencies={[mainAccount.currency]} />
+      {account && (
```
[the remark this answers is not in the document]
chabroA: 63c770ea2583f91d8cc562a3e703e498c480f0de

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepMethod.tsx
```diff
@@ -0,0 +1,213 @@
...
+  t,
+  updateTransaction,
+  transitionTo,
+}: StepProps) => {
+  const canSpeedup = haveFundToSpeedup && isOldestEditableOperation;
+  const canCancel = haveFundToCancel;
+
+  const handleContinueClick = async () => {
+    invariant(account, "account required");
+    invariant(transaction, "transaction required");
```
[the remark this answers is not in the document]
chabroA: 63c770ea2583f91d8cc562a3e703e498c480f0de

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepMethod.tsx
```diff
@@ -0,0 +1,213 @@
...
+    });
+
+    updateTransaction(tx => bridge.updateTransaction(tx, patch));
+
+    transitionTo("summary");
+  };
+
+  if (!account || !transaction) {
+    return null;
+  }
```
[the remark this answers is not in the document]
chabroA: 63c770ea2583f91d8cc562a3e703e498c480f0de

apps/ledger-live-mobile/src/const/navigation.ts
```diff
@@ -75,6 +75,7 @@ export enum ScreenName {
   EditAccountName = "EditAccountName",
   EditAccountUnits = "EditAccountUnits",
   EditDeviceName = "EditDeviceName",
+  // TODO: rename to EvmEditTransactionMethodSelection
```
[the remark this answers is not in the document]
chabroA: 89a4fb811633594d169b96ab5b79df68c63396a2

apps/ledger-live-mobile/src/families/evm/EvmFeesStrategy.tsx
chabroA: 6359fba04b8137e73944511e540c2e1bb40ba39a

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionStatus.ts
chabroA: Moved `getEditTransactionStatus` as well as `getMinFees` under `coin-evm`

cf. 32c6670a2af79384d7cb2673d1bf6df28c1eb286

libs/ledger-live-common/src/families/evm/editTransaction/getMinFees.ts
hedi-edelbloute: I'm for the enum as it provides keywords for further research

libs/coin-evm/src/api/node/rpc.common.ts
```diff
@@ -106,23 +106,28 @@ export const getTransactionCount: NodeApi["getTransactionCount"] = (currency, ad
...
 
-    try {
-      const gasEtimation = await api.estimateGas({
-        from: account.freshAddress, // should be necessary for some estimations
-        to,
-        value,
-        data,
-      });
+      try {
+        const gasEtimation = await api.estimateGas({
```
hedi-edelbloute, line 115: typo

libs/coin-evm/src/api/transaction/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export const getTransactionByHash = async (
+  currency: CryptoCurrency,
+  transactionHash: string,
+): Promise<{ confirmations?: number }> => {
+  const { getTransactionByHash: apiGetTransactionByHash } = getExplorerApi(currency);
+  if (!apiGetTransactionByHash) {
+    return {};
```
hedi-edelbloute, line 11: Shouldn't this throw an error and then we use a catch where it's called with the corresponding error ?

libs/ledger-live-common/src/families/evm/editTransaction/getMinFees.ts
lvndry: Indeed I wasn't aware that `type 1` and `EIP-2930` could be used interchangeably and that it's a known term 👍🏿 Good for me then

apps/ledger-live-desktop/src/renderer/components/ErrorBanner.tsx
```diff
@@ -2,20 +2,26 @@ import React from "react";
 import { urls } from "~/config/urls";
 import TranslatedError from "./TranslatedError";
 import Alert from "./Alert";
+
 type Props = {
   error: Error;
   warning?: boolean;
+  fallback?: {
```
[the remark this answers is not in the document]
chabroA: This is needed to provide the `fallback` prop to the underlying `TranslatedError` component.
As we can see in `TranslatedError`, if the provided error does not have a key for the specified field (either no `title` or no `description`) AND does not have a `fallback` prop provided, the generic error is used.

cf.
https://github.com/LedgerHQ/ledger-live/blob/47ea0d6fc91048555eea0db1f8ddab08f1fcac8c/apps/ledger-live-desktop/src/renderer/components/TranslatedError/TranslatedError.tsx#L71-L72

And since some errors, like the `NotEnoughGas` one, don't have description, we use an empty fallback in order not to display the generic `Something went wrong. Please retry or contact Ledger Support.` error in this case

https://github.com/LedgerHQ/ledger-live/blob/d5dae1e738f76dcd030b45d7723e503ad89ba311/apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/components/TransactionErrorBanner.tsx#L5-L29

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -0,0 +1,336 @@
...
+  );
+
+  const [editType, setEditType] = useState<StepProps["editType"]>(
+    isOldestEditableOperation && haveFundToSpeedup
+      ? "speedup"
+      : haveFundToCancel
+      ? "cancel"
+      : undefined,
+  );
+  const handleSetEditType = useCallback(editType => setEditType(editType), []);
```
lvndry, line 253: Not 100% but if the array is empty then it's just the same as if there was no useCallback at all

```suggestion
  const handleSetEditType = setEditType(editType);
```
Both are the same

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepFees.tsx
```diff
@@ -0,0 +1,148 @@
...
+        <SendAmountFields
+          account={mainAccount}
+          parentAccount={parentAccount}
+          status={status}
+          transaction={transaction}
+          /**
+           * Looks like TS does not like the fact that the arg of the functions
+           * are going from specific (EvmTransaction) to generic (Transaction).
+           */
+          onChange={onChangeTransaction as SendAmountFieldsProps["onChange"]}
```
lvndry, line 60: Why using `StepProps` in line 27 if you force a cast few lines later ? If StepProps don't have correct type you can create a type inside this file and just do:

```typescript
type StepFeesProps = Omit<StepProps, "onChangeTransaction" | "updateTransaction"> & {
 onChangeTransaction: SendAmountFieldsProps["onChange"];
 updateTransaction: SendAmountFieldsProps["updateTransaction"]
}
```

That way we sure to always pass the correct types from the parent component

apps/ledger-live-mobile/src/const/navigation.ts
```diff
@@ -560,8 +561,8 @@ export enum NavigatorName {
   RequestAccount = "RequestAccount",
   RequestAccountsAddAccounts = "RequestAccountsAddAccounts",
 
-  // Edit transaction
-  EditTransaction = "EditTransaction",
+  // Evm
+  EvmEditTransaction = "EvmEditTransaction",
```
lvndry, line 565: Can be move with the rest close to line 333

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+  const { amount, totalSpent, errors, warnings } = status;
+
+  const firstError = errors[Object.keys(errors)[0]];
+
+  const mainAccount = getMainAccount(account, parentAccount);
+  const currencyOrToken = getAccountCurrency(account);
+
+  const hasNonEmptySubAccounts =
+    account &&
```
lvndry, line 128: Don't need this check as you have the invariant line 63 right ?

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
lvndry: Alright, it looks like a FIXME that never going to be handled to me ahah
I think that the check `!transaction` can be removed as it's handled by an invariant few lines above

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/MethodSelection.tsx
lvndry: Alright sounds good!

libs/coin-evm/src/api/transaction/index.ts
lvndry: Yeah seems like `getExplorerApi` can throw an error, it should be wrapped around a try/catch: https://github.com/LedgerHQ/ledger-live/blob/feat/LIVE-8957-migrate-speedup-cancel-to-evm/libs/coin-evm/src/api/explorer/index.ts#L23

libs/coin-evm/src/api/transaction/index.ts
lvndry: So should we delete this TODO or add a ticket in the backlog ?

chabroA: Turns out this whole `getTransactionByHash` logic is unnecessary and a duplication of the `getTransaction` function available through the node API as mentioned by @lambertkevin 

fixed here 9320e023af14319ab3ad4a6ef350426aa5f6d195

apps/ledger-live-mobile/src/const/navigation.ts
chabroA: Not really, it's not the same enum, `NavigatorName` vs `ScreenName`

apps/ledger-live-desktop/src/renderer/screens/nft/Send/Summary.tsx
```diff
@@ -29,6 +29,22 @@ const Summary = ({ transaction }: Props) => {
   const { status, metadata } = useNftMetadata(nft?.contract, nft?.tokenId, nft?.currencyId);
   const { nftName } = status === "loaded" ? metadata : ({} as Record<string, unknown>);
```
lvndry, line 30:
```suggestion
  const { nftName } = status === "loaded" ? metadata : ({} as Record<string, never>);
```

libs/coin-evm/src/api/node/rpc.common.ts
chabroA: Fixed here 76c5794277d7983ce75cb1eda25a9d31b136da9f

libs/coin-evm/src/api/transaction/index.ts
chabroA: This file has been rewritten, cf. https://github.com/LedgerHQ/ledger-live/pull/4886#discussion_r1352513038

cf. 9320e023af14319ab3ad4a6ef350426aa5f6d195

apps/ledger-live-desktop/src/renderer/components/OperationsList/DateCell.tsx
```diff
@@ -16,72 +17,62 @@ const Cell = styled(Box).attrs(() => ({
...
-              </Box>
+  return <InfiniteLoader size={12} style={{ verticalAlign: "middle" }} />;
+};
+
+const DateCell = ({ t, operation, compact, text, editable }: Props): JSX.Element => {
+  const ellipsis = {
+    display: "block",
+    textOverflow: "ellipsis",
+    overflow: "hidden",
+    whiteSpace: "nowrap" as const,
```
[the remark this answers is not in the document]
chabroA: It is, otherwise you get a type error (`Type 'string' is not assignable to type 'WhiteSpace | undefined'` ) when using the `ellipsis` object as a `style` prop.

The `WhiteSpace` type definition is as follow in [csstype lib](https://github.com/frenic/csstype) used by our `ui` lib
`type Property.WhiteSpace = "pre" | Globals | "normal" | "nowrap" | "-moz-pre-wrap" | "break-spaces" | "pre-line" | "pre-wrap"`

I let you dig locally if you are curious

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
chabroA: > Not 100% but if the array is empty then it's just the same as if there was no useCallback at all

Not exactly, `useCallback` is used to return a cached version of a function (the first param of `useCallback`). And the second param of `useCallback` is the dependency array, used to determine when to invalidate this cache.

Since in JavaScript, a function () {} or () => {} always creates a different function, without `useCallback`, the `handleSetEditType` function will always be a different reference, at each render, even though the function "did not change" (i.e: still does the same thing).

But then a question could be "why do we need to keep the same reference between renders for this function?"

As stated by the [`useCallback` doc](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere):
> Caching a function with useCallback  is only valuable in a few cases:
> - ...
> - The function you’re passing is later used as a dependency of some Hook. For example, another function wrapped in useCallback depends on it, or you depend on this function from [useEffect.](https://react.dev/reference/react/useEffect)

Here, we are in this case where the `handleSetEditType` function, passed as the `setEditType` prop of the `Stepper` component, is latter used as a dependency of a call to `useCallback` in the `StepMethod` component: 

https://github.com/LedgerHQ/ledger-live/blob/cc7761e190aa0717e7c55c269e2f035b65ba5c95/apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepMethod.tsx#L80-L90

Here we use `useCallback` in order to avoid unnecessary function re-creation, which could lead to unnecessary re-renders.

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/steps/StepFees.tsx
chabroA: Indeed good idea

9d1ad3bdb03e7a362b935432cb4e727e06db6c9d

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
chabroA: Indeed, some checks are unnecessary here

3ef6e05d4ec4ab99dd05bdde7db4597ea234eac3

apps/ledger-live-mobile/src/families/evm/EditTransactionFlow/EditTransactionSummary.tsx
chabroA: Indeed, some checks are unnecessary here

3ef6e05d4ec4ab99dd05bdde7db4597ea234eac3

apps/ledger-live-desktop/src/renderer/screens/nft/Send/Summary.tsx
chabroA: e87934ad4297326b821ca38132f7a7004515c397

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -151,6 +152,7 @@ export class OperationsList extends PureComponent<Props, State> {
                       return null;
                     }
                   }
+                  const mainAccount = getMainAccount(account, parentAccount);
```
hzheng-ledger, line 155: I think the change in this file (this mainAccount line) is not necessary because the "isEditableOperation" is already able to handle both token account and normal account.

chabroA: `isEditableOperation` takes an `Account` as param but `account` is an `AccountLike` so we need to get the associated main account.

cf.
https://github.com/LedgerHQ/ledger-live/blob/a12f49718a7451892610b442136c06583469dddc/libs/coin-framework/src/operation.ts#L243-L269

libs/ledger-live-common/src/families/evm/editTransaction/isStrategyDisabled.ts
```diff
@@ -0,0 +1,37 @@
...
+  const minFees = getMinFees({ transaction });
+
+  if (transaction.type === 2) {
+    if (
+      !feeData.maxFeePerGas ||
+      !feeData.maxPriorityFeePerGas ||
+      !minFees.maxFeePerGas ||
+      !minFees.maxPriorityFeePerGas
+    ) {
+      return false;
```
hzheng-ledger, line 23: it seems that we will never arrive there. maybe we can only just use invariant to assert feeData.maxFeePerGas, feeData.maxPriorityFeePerGas, minFees.maxFeePerGas and minFees.maxPriorityFeePerGas are not undefined

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionPatch.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  if (editType === "speedup") {
+    const patch: Partial<EvmTransactionLegacy> = {
+      ...(getCommonSpeedUpPatch({
+        transaction,
+        shouldUseFastStrategy,
+        gasOptions,
+      }) as Partial<EvmTransactionLegacy>),
+      maxFeePerGas: undefined,
+      maxPriorityFeePerGas: undefined,
```
hzheng-ledger, line 140: maybe the 

maxFeePerGas: undefined,
maxPriorityFeePerGas: undefined,

are not necessary? as you only need to return a Partial transaction.
same for other similar pattern in this file

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionPatch.ts
```diff
@@ -0,0 +1,264 @@
...
+}: {
+  transaction: EvmTransactionEIP1559;
+  gasOptions: GasOptions;
+  editType: EditType;
+  account: Account;
+}): Partial<EvmTransactionEIP1559> => {
+  const fastFeeData = gasOptions.fast;
+
+  if (!fastFeeData.maxFeePerGas) {
+    throw new Error("gasOptions.fast.maxFeePerGas is required");
```
hzheng-ledger, line 176: we will never arrive here. we can use invariant to assert it.

hzheng-ledger (approved):
Look good to me.
Very good job. Thanks a lot.
I just have some very minor comments.

apps/ledger-live-desktop/src/renderer/drawers/OperationDetails/index.tsx
```diff
@@ -237,45 +237,24 @@ const OperationD = (props: Props) => {
...
-    _transactionRaw: unknown,
-    _transactionHash: unknown,
-  ) => {};
+  const handleOpenEditModal = useCallback(() => {
+    setDrawer(undefined);
+
+    invariant(operation.transactionRaw, "operation.transactionRaw is required");
+
+    dispatch(
+      openModal("MODAL_EVM_EDIT_TRANSACTION", {
```
haammar-ledger, line 248: Same here, as expected, in a more indirect way :slightly_smiling_face:

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
```diff
@@ -57,13 +59,13 @@ function StepConfirmation({
...
-    // if (error.name === "LedgerAPI4xx" && error.message.includes("nonce too low")) {
-    //   const mainAccount = account ? getMainAccount(account, parentAccount) : null;
-    //   if (mainAccount?.currency?.family === "evm") {
-    //     error = new TransactionHasBeenValidatedError();
-    //   }
-    // }
+    // Edit ethereum transaction nonce error because transaction has been validated
+    if (error.name === "LedgerAPI4xx" && error.message.includes("nonce too low")) {
+      const mainAccount = account ? getMainAccount(account, parentAccount) : null;
+      if (mainAccount?.currency?.family === "evm") {
```
haammar-ledger, line 65: Same here :see_no_evil: 

For the record, the same mechanism exists on LLD and LLM of using `generated` files + a `sync-families-dispatch` script, if you want to look at some examples.

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -209,11 +209,11 @@
       },
       "speedUp": {
         "title": "Speed Up",
-        "description": "Re-submit the transaction with a higher network fee."
+        "description": "Re-submit the transaction with a higher network fee.\n\nThe network fee must be at least 30% higher than the fee of the original transaction."
```
haammar-ledger, line 212: Is it ok to hardcode 30%? This isn't something that we can get dynamically?

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -4958,6 +4958,10 @@
     }
   },
   "errors": {
+    "ReplacementTransactionUnderpriced": {
+      "title": "Replacement transaction underpriced",
+      "description": "The network fee must be at least 10% higher than the fee of the original transaction."
```
haammar-ledger, line 4963: Same question

haammar-ledger (changes requested):
Sorry for the late review :pray: 

Looks good, I'm mostly curious about the files in live-common, I'll approve right away when that question is sorted.

apps/ledger-live-desktop/static/i18n/en/app.json
chabroA: > Is it ok to hardcode 30%?

Not really.

> This isn't something that we can get dynamically?

As with most things, it's probably possible. But after some though, it might not be that interesting to do so at this stage. The upside of displaying this % increase is quite low (especially since a default sensible fee value is automatically applied and if the user wants to set the fees manually, invalid options are disabled) compared to the added cost and complexity of dynamically displaying this value, which is itself based on 2 different env variables (depending on the tx type) in both LLD and LLM.

What I would recommend at this stage is to keep it simple and stay with a generic description, such as:

```suggestion
        "description": "Re-submit the transaction with a higher network fee."
```

And iterate on this once this feature is in customers hands.
This simple message will probably be enough and we can always add a dynamic and more accurate % value latter on if really needed.

apps/ledger-live-desktop/static/i18n/en/app.json
chabroA: Same thing as above https://github.com/LedgerHQ/ledger-live/pull/4886#discussion_r1368380567

And event slightly more complicated, since this 10% value is actually not specific to us but rather a minimum defined by most node implementations.

cf. https://github.com/LedgerHQ/ledger-live/blob/51473ff1c859f71765667acd8aa5c8d6ce24cf76/libs/coin-evm/src/getTransactionStatus.ts#L298-L305

This 10% is a hard limit, but it does not mean it will necessarily work with this value (that's why is states "at least"), depending on network activity at the time and other parameters (including, but not limited to, [the values defined by LL](https://github.com/LedgerHQ/ledger-live/blob/f521f5ba0a7a5c85d32ae7f0025ddae2e0d75fe2/libs/env/src/env.ts#L743-L757))

Here, we have multiple options:
1. not change the message, which is not factually false (to me "at least" means that it's not guaranteed to work if I update by the proposed amount, but I might be wrong here)
2. if 1 is not clear enough (which I can agree on), it could be rephrased. The basic intent of the message is that the proposed fees are to low and must be updated (by 10% minimum but maybe more). Any proposal is welcomed.
3. get rid of the percent value if misleading or unclear, and go with a simpler message like "The network fees are too low and must be increased"

libs/coin-evm/src/signOperation.ts
```diff
@@ -60,7 +61,11 @@ export const buildSignOperation =
   }): Observable<SignOperationEvent> =>
     new Observable(o => {
       async function main(): Promise<void> {
-        const preparedTransaction = await prepareForSignOperation(account, transaction);
+        const preparedTransaction = await prepareForSignOperation({
+          account,
+          transaction,
+          isValidNonce: transaction.nonce !== DEFAULT_NONCE,
```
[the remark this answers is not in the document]
haammar-ledger: I believe you meant without the `isValidNonce` param of the function, right?... That is what I had in mind and what is cleaner and simpler code-wise, the only reason not to, would be if there's any use case we can think of where comparing to the value of `DEFAULT_NONCE` is _not_ the right logic.
But otherwise, I would put the comparison inside the function and remove the param indeed. Anyway not super important :)

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionPatch.ts
chabroA: Updated here b4865eae0e76c45ae4aadb76bf8806ad8fefbde3

libs/ledger-live-common/src/families/evm/editTransaction/isStrategyDisabled.ts
chabroA: We can arrive here in multiple scenarios, that are not necessarily errors (for example, maybe the server did not return a gasOption for the selected strategy, or maybe the gasOptions are not yet fully loaded). And if we use invariant it will throw an error, which we don't really want here.

libs/coin-evm/src/signOperation.ts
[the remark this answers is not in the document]
chabroA: Updated here 847a3a18f80781090fc485ecf3b18a9fce1479c0

libs/coin-framework/src/operation.ts
```diff
@@ -240,16 +240,67 @@ export const isAddressPoisoningOperation = (
...
-    isEvmFamily = account.currency.family === "evm";
-  } else if (account.type === "TokenAccount") {
-    isEvmFamily = account.token.parentCurrency.family === "evm";
+export const isEditableOperation = (account: Account, operation: Operation): boolean => {
+  const { currency } = account;
+
+  // the edit transaction feature is only available for evm family
+  if (currency.family !== "evm") {
+    return false;
   }
```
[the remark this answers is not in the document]
chabroA: This "generated" pattern does not exist in the coin framework, and I don't think we want to add it here since coin-framework should be blockchain agnostic.
With that being said, I agree that there should not be some evm specific code / logic in here.

I updated this here a376c27ba0b0c53b3b4c886fcaece41eac42bb59

libs/coin-framework/src/operation.ts
```diff
@@ -240,16 +240,67 @@ export const isAddressPoisoningOperation = (
...
+  });
+};
+
+/**
+ * pending operations that exceed the ETHEREUM_STUCK_TRANSACTION_TIMEOUT
+ * threshold are considered as stuck
+ */
+export const isStuckOperation = (operation: Operation): boolean => {
+  return (
+    new Date().getTime() - operation.date.getTime() > getEnv("ETHEREUM_STUCK_TRANSACTION_TIMEOUT")
```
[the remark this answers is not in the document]
chabroA: Same as above https://github.com/LedgerHQ/ledger-live/pull/4886#discussion_r1368570722

Updated here a376c27ba0b0c53b3b4c886fcaece41eac42bb59

libs/ledger-live-common/src/families/evm/editTransaction/getEditTransactionPatch.ts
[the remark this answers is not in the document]
chabroA: Not necessarily, updated here b8c0aa99f165d12fd102f1a0ed87249821cdec53

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
chabroA: Not as straightforward to fix.

Furthermore, this comes from the original PR (#2441) and did not seem to be challenged at the time (rightfully or not, I don't know)

cf. https://github.com/LedgerHQ/ledger-live/pull/2441/files#diff-805265fed9cfcb5a2cbfbc7788982568409a0ccf80a87d8706aa9be5cfccfb3cR62-R68

apps/ledger-live-desktop/src/renderer/drawers/OperationDetails/index.tsx
chabroA: Not as straightforward to fix.

Furthermore, this comes from the original PR (#2441) and did not seem to be challenged at the time (rightfully or not, I don't know)

cf. https://github.com/LedgerHQ/ledger-live/pull/2441/files#diff-42907b8096ef2aee623bba0ad42cb4f2a64bd09257203f8b8464e60a93331915R245-R277

apps/ledger-live-desktop/static/i18n/en/app.json
lvndry: Just my 2 cent here for next iterations but I feel like these kind of errors should guide the user on what actions needs to be made. So they would see something like `the fees must be at least xxx ETH` instead of `the fees must be at leat x% of the current network fees`. The latter asks more effort as we need to look for the current network fees somewhere and do a calculation.

haammar-ledger (approved):

haammar-ledger (approved):
:rocket:

PR #4931, opened 2023-10-04
WIP - collect coverage jest llm

KVNLS:
### 📝 Description

Collect coverage from LLM

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: `` 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach


.github/workflows/test-desktop.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           aws-access-key: ${{ secrets.AWS_S3_CACHE_ACCESS_KEY }}
           aws-secret-key: ${{ secrets.AWS_S3_CACHE_SECRET_KEY }}
       - name: Run unit tests
-        run: pnpm desktop test:jest
+        run: pnpm desktop test:jest:coverage
```
mohammed-ledger, line 125: Consider moving the unit tests to test.yml in order to collect coverage in one workflow.

KVNLS: `test.yml` only includes tests for our libraries, am I wrong? If it's the case I think that we might want to keep them separated.

KVNLS:
@gre , @valpinkman, I discovered that if you update `babel/core` in LLM and have a different version in `ui/packages/native` you will end up with two references of `react-native-reanimated` in `pnpm.lock` and it breaks the app. The symptom is that you can't have the Transfer Drawer and other Drawers in the app (detected with e2e tests).

apps/ledger-live-mobile/tsconfig.json
```diff
@@ -10,8 +10,11 @@
     "target": "esnext",
     "incremental": true,
     "baseUrl": ".",
+    "rootDir": ".",
     "paths": {
-      "@utils/*": ["./src/utils/*"]
+      "~/*": ["src/*"],
+      "@utils/*": ["./src/utils/*"],
+      "tests/*": ["src/__tests__/*"]
```
mcayuelas-ledger, line 17: Should we have a naming convention for this? :)

KVNLS: I agree, I will go with `@` prefix

gre:
@KVNLS it's a classical problem in our stack unfortunately. it's very easy to break the dedup of our libs. `pnpm dedupe` is the way to globally fix this, however we diverged **way too much** to make this realistic to directly jump to a pnpm deduped version, but it's a goal we need to keep in mind https://github.com/LedgerHQ/ledger-live/pull/5223

mcayuelas-ledger (approved):

PR #4983, opened 2023-10-09
Bugfix/live 6393 unlock ble drawer

aussedatlo:
### 📝 Description

display a drawer when stax is locked during pairing. This is working only if device has been already paired since this information is not available when not paired. If not, the previous generic error is displayed.

### ❓ Context

- **Impacted projects**: `LLM` `ledger-live-common`
- **Linked resource(s)**: [LIVE-6393]

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/35492518/196f157f-b84c-4b69-9927-75dc5e2184a8

### 🚀 Expectations to reach

[LIVE-6393]: https://ledgerhq.atlassian.net/browse/LIVE-6393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.ts
+++ b/libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.ts
@@ -1,7 +1,8 @@
 import { useState, useEffect } from "react";
-import { from } from "rxjs";
-import { first } from "rxjs/operators";
+import { from, throwError, timer } from "rxjs";
+import { first, retry } from "rxjs/operators";
 import type { FirmwareInfo } from "@ledgerhq/types-live";
+import { LockedDeviceError } from "@ledgerhq/errors";
 import { withDevice } from "../../hw/deviceAccess";
 import getVersion from "../../hw/getVersion";
 import { BleError } from "../types";
@@ -23,6 +24,9 @@ export type useBleDevicePairingResult = {
  * @returns An object containing:
  * - isPaired: a boolean set to true if the device has been paired, false otherwise
  * - pairingError: any PairingError that occurred, null otherwise
+ *
+ * When the device is locked, notify the pairing error and retry
+ * until the device is unlocked
  */
 export const useBleDevicePairing = ({
   deviceId,
@@ -31,8 +35,19 @@ export const useBleDevicePairing = ({
   const [pairingError, setPairingError] = useState<PairingError>(null);
 
   useEffect(() => {
-    const requestObservable = withDevice(deviceId)(t => from(getVersion(t))).pipe(first());
+    const requestObservable = withDevice(deviceId)(t => from(getVersion(t))).pipe(
+      first(),
+      retry({
+        delay: (err: BleError) => {
+          if (err instanceof LockedDeviceError) {
+            setPairingError(err);
+            return timer(1000);
+          }
 
+          return throwError(() => err);
+        },
+      }),
+    );
     const sub = requestObservable.subscribe({
       next: (_value: FirmwareInfo) => {
         setIsPaired(true);
--- a/.changeset/sharp-readers-occur.md
+++ b/.changeset/sharp-readers-occur.md
@@ -1,0 +1,6 @@
+---
+"live-mobile": patch
+"@ledgerhq/live-common": patch
+---
+
+add correct drawer when device is locked during pairing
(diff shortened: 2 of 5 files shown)


apps/ledger-live-mobile/src/components/BleDevicePairingFlow/index.tsx
```diff
@@ -91,22 +94,35 @@ const BleDevicePairingFlow: React.FC<BleDevicePairingFlowProps> = ({
...
-    (item: Device) => {
-      const deviceToPair = {
-        deviceId: item.deviceId,
-        deviceName: item.deviceName,
-        modelId: item.modelId,
-        wired: false,
-      };
+  const { pairingError } = useBleDevicePairing({
+    deviceId: deviceToPair?.deviceId,
+  });
```
alexandremgo, line 102: What scares me is that now we trigger the pairing flow with `useBleDevicePairing` in both  `apps/ledger-live-mobile/src/components/BleDevicePairingFlow/index.tsx` and `apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairing.tsx`

I think this is a bad idea :/ Why can't you put `UnlockDeviceDrawer` in `BleDevicePairing` ? And if it's a problem, maybe we could give to `BleDevicePairing` a setState prop like `setIsDeviceLocked` and have this `isDeviceLocked` state live here ;)

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/index.tsx
```diff
@@ -11,6 +13,7 @@ import type { BleDevicePairingProps } from "./BleDevicePairing";
 import { track } from "../../analytics";
 import { NavigationHeaderBackButton } from "../NavigationHeaderBackButton";
 import { NavigationHeaderCloseButton } from "../NavigationHeaderCloseButton";
+import UnlockDeviceDrawer from "../../screens/SyncOnboarding/UnlockDeviceDrawer";
```
alexandremgo, line 16: If this `UnlockDeviceDrawer` is re-used we should either copy it here in this flow (in the `components/BleDevicePairingFlow`) or move it at the root of our `components` folder

I would go with the moving it to `components/` so it can be re-used

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/index.tsx
aussedatlo: Yes this might not be the best solution indeed.

I found a way to keep the hook on `BleDevicePairing.tsx` and the drawer on `index.tsx`by calling a `onLock` callback from `index.tsx` to display the drawer.

The challenge was to not display the generic error from `BleDevicePairing.tsx` at the same time, so with the callback it redirect to previous screen, then popup the lock drawer without displaying any errors.

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/index.tsx
```diff
@@ -205,21 +208,40 @@ const BleDevicePairingFlow: React.FC<BleDevicePairingFlowProps> = ({
     requestToSetHeaderOptions,
   ]);
 
+  const onLocked = useCallback(
+    (device: Device) => {
+      onRetryPairingFlow();
```
alexandremgo, line 213: You do this to go back to the scanning step right ?

But so currently there is no automatic way to see that the device is not locked anymore (no retrying on the pairing step directly). The user has to click on cross button to exit. Maybe there is a way to update the `useBleDevicePairing` so it retries on this specific locked device error - and you don't need to restart the entire flow

aussedatlo: Done

libs/ledger-live-common/src/ble/hooks/useBleDevicePairing.ts
```diff
@@ -31,8 +32,19 @@ export const useBleDevicePairing = ({
   const [pairingError, setPairingError] = useState<PairingError>(null);
 
   useEffect(() => {
-    const requestObservable = withDevice(deviceId)(t => from(getVersion(t))).pipe(first());
+    const requestObservable = withDevice(deviceId)(t => from(getVersion(t))).pipe(
+      first(),
+      retry({
+        delay: (err: BleError) => {
+          if (err instanceof LockedDeviceError) {
```
alexandremgo, line 39: Can you just add a small line about the fact that on `LockedDeviceError` it automatically retries (but still propagated the error) in the doc comments above the function definition pls ?

alexandremgo (approved):

PR #4533, opened 2023-08-29
[LL] - [LIVE-9115] - Add typings to useFeature and improve its overall dev experience

sshmaxime:
### 📝 Description

This PR improves the developer experience of the `useFeature` hook by adding typings to it, it also slightly change few things here and there, all related to the `Feature` concept in our apps.

**Done:**

- Typed `useFeature`
- Renamed `feature` props to `featureId` in FeatureToggle component.
- Renamed `defaultFeatures` to `DEFAULT_FEATURES`.
- Created a norm naming for features: `Feature_X` (Open to changes, I didn't know what else to name it)
- Renamed `recoverFeatueFlag.ts` to `recoverFeatureFlag.ts`

**Notes:**

- I added few `@ts-expect-error` because some functions doesn't accept `undefined` as parameter and there is a need of thinking on what would be the correct to handle that. For example, what do we do if the link passed to `openLink` is undefined ? Because technically as of today it could be `undefined`. That being said I think this is to be done in another PR, this one is already pretty big.
- There is a lot of changed files but it mainly comes for the fact that `useFeature` is used everywhere.

**How to properly add a new Feature Flag:**

- Add its type in `libs/ledgerjs/packages/types-live/src/feature.ts`.
- Add its default value in `libs/ledger-live-common/src/featureFlags/defaultFeatures.ts`.
- Rebuild the lib if necessary: `pnpm build:libs`

### ❓ Context

- **Impacted projects**: `LL` 
- **Linked resource(s)**: [LIVE-9115] 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

[LIVE-9115]: https://ledgerhq.atlassian.net/browse/LIVE-9115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4999, opened 2023-10-10
Bugfix/lld manager apps not supported

ofreyssinet-ledger:
### 📝 Description

⚠️ please read this description entirely before reviewing the code

There is an issue in the manager where most apps are displaying "requires 3rd party wallet":
> Screenshot 2023-10-10 at 10 24 31

This issue was introduced in #4533 when the signature of `useFeature` was changed to never return `null`. If a feature flag value was not defined, it would default to a fallback "default value" of `{enabled: false}`.
We should not do this, if a feature flag value is not defined, `useFeature` should always return `null`.

The issue becomes apparent when reading the code that decides whether or not an app should be considered as "supported by live", in this case `flag` is always defined, and in the worst case (if the remote flag does not actually exist) its value will be 
`{enabled: false}`, which will make the app appear as "not supported by Live":

```ts
// apps/ledger-live-desktop/src/renderer/screens/manager/AppsList/Item.tsx

const flag = useFeature(camelCase(`currency_${currencyId}`) as FeatureId);
// when the flag doesn't exist it's equivalent to being enabled
const currencyFlagEnabled = !flag || flag.enabled;
const currencySupported = !!currency && isCurrencySupported(currency) && currencyFlagEnabled;
const isLiveSupported = currencySupported || ["swap", "plugin"].includes(type);
```

**The default values of feature flags should only be defined in `defaultFeatures` and the internal logic of the Firebase SDK will use that to fallback to those default values in case there is a network error for instance or any issue where the remote configuration cannot be obtained.**

Additionally, in this PR:
- `FeatureFlagsProvider` is now defined as `FeatureFlagsContext.Provider` so that the context value passed to it isn't destructured into a new object. This destructuration into a new object became obvious when writing a unit test for the provider. There were no important consequences to that but it's better to create this new object explicitly. Now the context value passed to it has to be defined directly in the component that renders it.
- I have added basic unit tests to `useFeature` and `useFeatureFlags`.

#### Additional remark on the typing of `useFeature`:
**In the current architecture, the return type of `useFeature` has to allow `null` as a returned value**.
The only alternative possible would be to throw an error if there is no feature flag found. But it's much harder to handle in React as it would require an error boundary etc. and in some cases we want to handle the `null` value directly.
Making it return a default value is a bad idea for all the reasons mentioned above.

The reason is that the current logic works with a dependency injection via a `React.Context` so the typing system has no way to ensure that the component will be rendered within a context that injects a remote config containing a value that is defined for the `featureId` passed to the hook as a parameter.

#### What remains to do later:
- The actual feature flag logic (with overrides by env or settings, with specific version & language rules override, with default features fallback behavior etc.) should probably be moved to `common` and then also unit tested. The part of the code that has been modified here is using dependency injection (via a context) so **the unit tests don't test the layer that implements the logic between the Firebase SDK, the LL env and settings and the `FeatureFlagsContext`**.

### ❓ Context

- **Impacted projects**: `LLC, LLM , LLD` 
- **Linked resource(s)**: [LIVE-9700] 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

The expected behavior for regular apps not controlled with a feature flag:
Screenshot 2023-10-10 at 16 55 06

Example with Astar feature flag enabled and then disabled:
    
- enabled:
Screenshot 2023-10-10 at 16 53 03
Screenshot 2023-10-10 at 16 53 10

- disabled:
Screenshot 2023-10-10 at 16 52 57
Screenshot 2023-10-10 at 16 52 46

### 🚀 Expectations to reach

[LIVE-9700]: https://ledgerhq.atlassian.net/browse/LIVE-9700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/featureFlags/FeatureFlagsContext.tsx
ofreyssinet-ledger, line 1: this file was originally `provider.ts`

libs/ledger-live-common/src/featureFlags/useFeature.ts
```diff
@@ -1,21 +1,20 @@
...
  *
  * @param featureId
- * @returns a feature.
+ * @returns a feature flag value or null if the feature flag is not found.
  */
-const useFeature = <T extends FeatureId>(featureId: T): Feature<FeatureParam<T>> => {
+const useFeature = <T extends FeatureId>(featureId: T): Feature<FeatureParam<T>> | null => {
   const featureFlags = useFeatureFlags();
   const value = useMemo(() => featureFlags.getFeature(featureId), [featureFlags, featureId]);
-  return value ?? DEFAULT_FEATURE;
```
ofreyssinet-ledger, line 18: this is the main bug being fixed in this PR

apps/ledger-live-mobile/src/components/FirebaseFeatureFlags.tsx
```diff
@@ -172,12 +172,14 @@ export const FirebaseFeatureFlagsProvider: React.FC<Props> = ({ children }) => {
...
-      resetFeatures={resetFeatures}
-      getAllFlags={getAllFlags}
+      value={{
+        isFeature,
+        getFeature: wrappedGetFeature,
+        overrideFeature,
+        resetFeature,
+        resetFeatures,
+        getAllFlags,
+      }}
```
ofreyssinet-ledger, line 182: `FeatureFlagsProvider` is now defined as `FeatureFlagsContext.Provider` so that the context value passed to it isn't destructured into a new object. (cf. definition of `FeatureFlagsProvider` in live-common).
This destructuration into a new object became obvious when writing a unit test for the provider. There were no important consequences to that but it's better to create this new object explicitly. Now the context value passed to it has to be defined directly in the component that renders it.

gre: should it be `useMemo` ied tho? like isn't it going to re-render all consumer of this each time the component itself render() ?

ofreyssinet-ledger: done @gre 👍

sshmaxime:
LGTM, well done @OlivierFreyssinet 🙏

sshmaxime (approved):

libs/ledger-live-common/src/featureFlags/useFeature.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { renderHook } from "@testing-library/react-hooks";
+import useFeature from "./useFeature";
+import { makeMockedFeatureFlagsProviderWrapper, makeMockedContextValue } from "./mock";
+
+describe("useFeature hook", () => {
+  it("should return null if a flag is not defined remotely", () => {
+    const mockedFeatures = {};
+    const { result } = renderHook(() => useFeature("mockFeature"), {
+      wrapper: makeMockedFeatureFlagsProviderWrapper(makeMockedContextValue(mockedFeatures)),
```
alexandremgo, line 9: Nice ! I did not know about the `wrapper` option

libs/ledger-live-common/src/featureFlags/FeatureFlagsContext.tsx
```diff
@@ -0,0 +1,69 @@
+import { createContext, useContext } from "react";
+import type { FeatureId, Feature } from "@ledgerhq/types-live";
+
+export type FeatureFlagsContextValue = {
+  /**
+   *
+   * @param featureId featureId identifying a potential feature flag
```
alexandremgo, line 7: `featureId` `featureId` :P

apps/ledger-live-desktop/src/renderer/screens/settings/sections/Developer/FeatureFlagsSettings/GroupedFeatures.tsx
```diff
@@ -1,5 +1,5 @@
 import { groupedFeatures } from "@ledgerhq/live-common/featureFlags/groupedFeatures";
-import { useFeatureFlags } from "@ledgerhq/live-common/featureFlags/provider";
+import { useFeatureFlags } from "@ledgerhq/live-common/featureFlags/FeatureFlagsContext";
```
alexandremgo, line 2: It does not matter, but everything in `featureFlags` that can be exported is exported from `index.ts` no ? You can directly import from `@ledgerhq/live-common/featureFlags/index`

ofreyssinet-ledger: ah yep indeed, here it was auto renaming in action

alexandremgo (approved):
LGTM !

PR #5019, opened 2023-10-11
Bugfix/live 9554 Fix design of bluetooth pairing error screen

aussedatlo:
### 📝 Description

This PR fix the design of new bluetooth pairing error screen using the FF flag `llmNewDeviceSelection `. This screen can appear on **My Ledger**, and during the new **onboarding.**

For the **previous onboarding** flow (Access your wallet -> Connect your ledger -> Pair my nano -> Add with bluetooth), the design is also improved but not totally correct. Since this older flow will be replaced we will not fix it right now.

### ❓ Context

- **Linked resource(s)**: [LIVE-9554] ; [Figma](https://www.figma.com/file/x9nwU0JtDtyUIhAyBB2El8/Onboarding?type=design&node-id=10078-34306&mode=design&t=RRo0LwguVpuw0RW8-0)

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/35492518/8239f729-43f9-4e5c-ae24-29907d773f14

### 🚀 Expectations to reach

[LIVE-9554]: https://ledgerhq.atlassian.net/browse/LIVE-9554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
+++ b/apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
@@ -58,6 +58,7 @@ type Props = {
   isChoiceDrawerDisplayedOnAddDevice?: boolean;
   withMyLedgerTracking?: boolean;
   filterByDeviceModelId?: DeviceModelId;
+  paddingBottom?: number;
 };
 
 export default function SelectDevice({
@@ -69,6 +70,7 @@ export default function SelectDevice({
   hasPostOnboardingEntryPointCard,
   withMyLedgerTracking,
   filterByDeviceModelId,
+  paddingBottom,
 }: Props) {
   const [USBDevice, setUSBDevice] = useState<Device | undefined>();
   const [ProxyDevice, setProxyDevice] = useState<Device | undefined>();
@@ -307,7 +309,7 @@ export default function SelectDevice({
   );
 
   return (
-    <>
+    <Flex flex={1} pb={paddingBottom && !isPairingDevices ? paddingBottom : 0}>
       {withMyLedgerTracking ? <TrackScreen {...trackScreenProps} /> : null}
       <RequiresBluetoothDrawer
         isOpenedOnIssue={isBleRequired}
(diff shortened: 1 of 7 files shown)


apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
```diff
@@ -326,7 +329,7 @@ export default function SelectDevice({
           filterByDeviceModelId={filterByDeviceModelId}
         />
       ) : (
-        <Flex flex={1}>
+        <Flex flex={1} pb={insets.bottom + TAB_BAR_SAFE_HEIGHT}>
```
ofreyssinet-ledger, line 332: Usually I would avoid using these safe area insets on components that can be integrated into screens, as the component doesn't have much info on the screen context, and maybe the parent screen is already adding this safe area itself, maybe it's already adding padding at the bottom, or maybe there is some content rendered below this component.
If you have checked all existing usages of the SelectDevice2 component and ensured it was ok then it's fine but otherwise please check, and as a general rule, setting "safe area" paddings like that should be the responsibility of the component that is rendered as a screen, so you ensure that only the "screen" component sets the safe area padding, and that its children don't also add a redundant padding.

ofreyssinet-ledger: The way to go in my opinion would be to add a `paddingBottom` prop to the `SelectDevice2` component

aussedatlo: Added a `paddingBottom` props to the `SelectDevice2` that is effective conditionally, depending on `isPairingDevices` to avoid unnecessary padding during the error screen.

ofreyssinet-ledger (approved):
lgtm 👍

PR #5030, opened 2023-10-12
chore: also run on macos-latest in case runners are busy

gre:
### 📝 Description

in context of not having enough runners for Mac. we reopen the possibility to consider a Github runner, even if slower, will relieve a bit the current queuing which takes more time that the low performance that github have.

### ❓ Context

- **Impacted projects**: `ci` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9699 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.github/workflows/build-desktop.yml
+++ b/.github/workflows/build-desktop.yml
@@ -95,7 +95,7 @@ jobs:
 
   build-desktop-app-macos:
     name: "Build Ledger Live Desktop (Mac OS X)"
-    runs-on: [m1, ARM64]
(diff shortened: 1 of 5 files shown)


sarneijim (approved):

gre:
we will not go that way for now. because adding an alternative runner can bring confusion in future, (since you can get x1 vs x3 perf ratio depending on if it's our own runner or github's slow ones) a long term plan is to have more runners

PR #5033, opened 2023-10-12
Add BIP-69 inputs / outputs indexing

chabroA:
### 📝 Description

Add Lexicographical Indexing of Transaction Inputs and Outputs following BIP-69 specification

Following an analysis by @bigspider on wallet fingerprinting (cf. [related blogpost](https://ishaana.com/blog/wallet_fingerprinting/)), it turns out that the Ledger Live wallet does not follow any standard for ordering transaction inputs and outputs, leading to Wallet Fingerprinting and privacy leak.

The [BIP-69 standard](https://en.bitcoin.it/wiki/BIP_0069) has been created in order to mitigate this issue and is implemented by other wallets handled Bitcoin as of today.

This PR adds lexicographical indexing of Bitcoin transaction inputs and outputs following this standard.

This ordering has been added in the `signAccountTx`, right before the transaction is sent to the signer for signature.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10818

### ✅ Checklist

---


libs/ledger-live-common/src/families/bitcoin/wallet-btc/wallet.ts
```diff
@@ -269,35 +283,35 @@ class BitcoinLikeWallet {
...
+      lockTime: params.lockTime,
+      sigHashType: params.sigHashType,
+      segwit: params.segwit,
       initialTimestamp,
-      ...(params.expiryHeight && { expiryHeight: params.expiryHeight }),
-      ...(txInfo.outputs[lastOutputIndex]?.isChange && {
-        changePath: `${fromAccount.params.path}/${fromAccount.params.index}'/${txInfo.changeAddress.account}/${txInfo.changeAddress.index}`,
-      }),
+      expiryHeight: params.expiryHeight,
+      changePath: orderedOutputs[lastOutputIndex]?.isChange
```
hzheng-ledger, line 297: as you've changed the output order, the "txInfo.outputs[lastOutputIndex]?.isChange" is not necessary the "orderedOutputs[lastOutputIndex]?.isChange".
we should use the "txInfo.outputs[lastOutputIndex]?.isChange" as before

libs/ledger-live-common/src/families/bitcoin/wallet-btc/wallet.ts
```diff
@@ -269,35 +283,35 @@ class BitcoinLikeWallet {
...
```
hzheng-ledger, line 312: as you've changed the output order, the "txInfo.outputs[lastOutputIndex]?.isChange" is not necessary the value as "orderedOutputs[lastOutputIndex]?.isChange".
we should use the "txInfo.outputs[lastOutputIndex]?.isChange" as before

libs/ledger-live-common/src/families/bitcoin/wallet-btc/wallet.ts
```diff
@@ -269,35 +283,35 @@ class BitcoinLikeWallet {
...
     });
 
     const tx = await btc.createPaymentTransaction({
       inputs,
       associatedKeysets,
       outputScriptHex,
-      ...(params.lockTime && { lockTime: params.lockTime }),
-      ...(params.sigHashType && { sigHashType: params.sigHashType }),
-      ...(params.segwit && { segwit: params.segwit }),
+      lockTime: params.lockTime,
```
hzheng-ledger, line 307: in theory, there should not be impact for this simplification. 
Launch the bot and confirm that there is no breaking change.
Thanks a lot.

hzheng-ledger (changes requested):
it looks good in general. Thanks a lot.
there is only a minor problem of "isChange" boolean I am afraid.

libs/ledger-live-common/src/families/bitcoin/wallet-btc/utils.ts
```diff
@@ -193,3 +193,80 @@ export function writeVarInt(buffer: Buffer, i: number, offset: number): number {
...
+  if (a === b) {
+    return 0;
+  }
+
+  if (a > b) {
+    return 1;
+  }
+
+  return -1;
+};
```
haammar-ledger, line 207: Any reason not to use `localeCompare` instead?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

libs/ledger-live-common/src/families/bitcoin/wallet-btc/wallet.ts
haammar-ledger: I agree with being cautious with that change :warning:
That weird pattern was here for a reason, and it was because providing these params as `undefined` caused some errors, and it was required to _not_ provide them instead...

haammar-ledger (commented):
Can be validated without QA but need to run the bot several times on the whole `bitcoin` family (which currently has a number of issues... so also need to compare with non-reg runs on #ledger-live-bot)

libs/ledger-live-common/src/families/bitcoin/wallet-btc/wallet.ts
haammar-ledger: From a quick look at the `createPaymentTransaction` function, its implem in _BtcNew.ts_ looks safe, but its implem in _BtcOld.ts_ doesn't. So I think it should stay like it is for now...

chabroA: Reverted and added a comment

libs/ledger-live-common/src/families/bitcoin/wallet-btc/wallet.ts
chabroA: Indeed, reverted that in e96dea30caec5f9092cb056b5e258a42d0655298

haammar-ledger:
Last discussions identified that breaking changes with risky impact are necessary on the `hw-app-btc` lib, and are considered not worth it at the moment. Closing this topic for now.

PR #5034, opened 2023-10-12
chore: remove default import stellarsdk

lvndry:
### 📝 Description

Screenshot 2023-10-13 at 10 47 03

StellarSDK default import is of type `any`. We can use named import to have correct types instead.

```typescript
import StellarSDK from "stellar-sdk"; // StellarSDK is of type any
```

### ❓ Context

- **Linked resource(s)**: []

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/package.json
+++ b/libs/ledger-live-common/package.json
@@ -226,7 +226,7 @@
     "semver": "^7.3.5",
     "sha.js": "^2.4.11",
     "simple-cbor": "^0.4.1",
-    "stellar-sdk": "^10.1.1",
+    "stellar-sdk": "^10.4.1",
     "superstruct": "0.14.2",
     "triple-beam": "^1.3.0",
     "tronweb": "^5.2.0",
--- a/libs/ledger-live-common/src/families/stellar/api/horizon.ts
+++ b/libs/ledger-live-common/src/families/stellar/api/horizon.ts
@@ -2,12 +2,23 @@ import { LedgerAPI4xx, LedgerAPI5xx, NetworkDown } from "@ledgerhq/errors";
 import { requestInterceptor, responseInterceptor } from "@ledgerhq/live-network/network";
 import type { Account, Operation } from "@ledgerhq/types-live";
 import { BigNumber } from "bignumber.js";
-import StellarSdk, {
+import {
   // @ts-expect-error stellar-sdk ts definition missing?
   AccountRecord,
   NetworkError,
   NotFoundError,
+  Server,
+  HorizonAxiosClient,
+  BASE_FEE,
+  Asset,
+  Operation as StellarOperation,
+  Account as StellarAccount,
+  Transaction as StellarTransaction,
+  TransactionBuilder,
+  Networks,
+  ServerApi,
 } from "stellar-sdk";
+
 import { getCryptoCurrencyById, parseCurrencyUnit } from "../../../currencies";
 import { getEnv } from "@ledgerhq/live-env";
 import {
@@ -15,7 +26,7 @@ import {
   getReservedBalance,
   rawOperationsToOperations,
 } from "../logic";
-import type { BalanceAsset, NetworkInfo } from "../types";
+import type { BalanceAsset, NetworkInfo, RawOperation } from "../types";
 import { NetworkCongestionLevel, Signer } from "../types";
 
 const LIMIT = getEnv("API_STELLAR_HORIZON_FETCH_LIMIT");
@@ -23,16 +34,16 @@ const FALLBACK_BASE_FEE = 100;
 const TRESHOLD_LOW = 0.5;
 const TRESHOLD_MEDIUM = 0.75;
 const currency = getCryptoCurrencyById("stellar");
-const server = new StellarSdk.Server(getEnv("API_STELLAR_HORIZON"));
+const server = new Server(getEnv("API_STELLAR_HORIZON"));
 
 // Constants
 export const BASE_RESERVE = 0.5;
 export const BASE_RESERVE_MIN_COUNT = 2;
 export const MIN_BALANCE = 1;
 
-StellarSdk.HorizonAxiosClient.interceptors.request.use(requestInterceptor);
+HorizonAxiosClient.interceptors.request.use(requestInterceptor);
 
-StellarSdk.HorizonAxiosClient.interceptors.response.use(response => {
+HorizonAxiosClient.interceptors.response.use(response => {
   responseInterceptor(response);
   // FIXME: workaround for the Stellar SDK not using the correct URL: the "next" URL
   // included in server responses points to the node itself instead of our reverse proxy...
@@ -66,7 +77,7 @@ export const fetchBaseFee = async (): Promise<{
     };
   }
 
-  const baseFee = StellarSdk.BASE_FEE || FALLBACK_BASE_FEE;
+  const baseFee = Number(BASE_FEE) || FALLBACK_BASE_FEE;
   let recommendedFee = baseFee;
   let networkCongestionLevel = NetworkCongestionLevel.MEDIUM;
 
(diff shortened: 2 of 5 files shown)


libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -107,24 +121,25 @@ export const fetchAccount = async (
   spendableBalance: BigNumber;
   assets: BalanceAsset[];
 }> => {
-  let account: typeof AccountRecord = {};
-  let balance: Record<string, any> = {};
+  let account: ServerApi.AccountRecord = {} as ServerApi.AccountRecord;
```
sprohaszka-ledger, line 124: I don't understand why you force cast `{}` as `account` is typed.

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -107,24 +121,25 @@ export const fetchAccount = async (
...
   let assets: BalanceAsset[] = [];
+  let balance = "0";
```
sprohaszka-ledger, line 126: 👍

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -66,7 +77,7 @@ export const fetchBaseFee = async (): Promise<{
     };
   }
 
-  const baseFee = StellarSdk.BASE_FEE || FALLBACK_BASE_FEE;
+  const baseFee = Number(BASE_FEE) || FALLBACK_BASE_FEE;
```
haammar-ledger, line 80: While `Number` may work, I advise just using `new BigNumber()` everywhere. It's the pattern LL uses for now, until we migrate away from using `BigNumber` and probably use `BigInt` instead.

Using Number's constructor is even discouraged: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/Number#return_value

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -107,24 +121,25 @@ export const fetchAccount = async (
...
   let assets: BalanceAsset[] = [];
+  let balance = "0";
 
   try {
     account = await server.accounts().accountId(addr).call();
-    balance = account.balances?.find(balance => {
-      return balance.asset_type === "native";
-    });
+    balance =
+      account.balances.find(balance => {
```
haammar-ledger, line 131: According to current code there's no guarantee that `account.balances` isn't undefined, maybe better to keep the safety? (even if types pretend it's non-optional, backend APIs don't always respect this, and types don't protect us against runtime data...)

(also below L135)

libs/ledger-live-common/src/families/stellar/logic.ts
```diff
@@ -315,10 +313,10 @@ export const recipientAccount = async (
 
   let accountAddress = address;
 
-  const isMuxedAccount = Boolean(StellarSdk.StrKey.isValidMed25519PublicKey(address));
+  const isMuxedAccount = StrKey.isValidMed25519PublicKey(address);
```
sprohaszka-ledger, line 316: Why don't you remove `Boolean()` call? Don't we lose the type?

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -276,41 +290,37 @@ export const buildPaymentOperation = ({
...
-    assetCode && assetIssuer
-      ? new StellarSdk.Asset(assetCode, assetIssuer)
-      : StellarSdk.Asset.native();
-  return StellarSdk.Operation.payment({
+  const asset = assetCode && assetIssuer ? new Asset(assetCode, assetIssuer) : Asset.native();
+  return StellarOperation.payment({
     destination: destination,
     amount: formattedAmount,
     asset,
-    withMuxing: true,
```
haammar-ledger, line 291: Why that change?

sprohaszka-ledger (commented):
I am not sure that we have to use `as` (force cast operator) so much 🤔

libs/ledger-live-common/package.json
```diff
@@ -248,7 +248,7 @@
     "sha.js": "^2.4.11",
     "simple-cbor": "^0.4.1",
     "source-map-support": "^0.5.21",
-    "stellar-sdk": "^10.1.1",
+    "stellar-sdk": "^10.4.1",
```
haammar-ledger, line 251: Let's at least run the bot just once to be safer with that lib update

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -2,37 +2,48 @@ import { LedgerAPI4xx, LedgerAPI5xx, NetworkDown } from "@ledgerhq/errors";
...
   AccountRecord,
   NetworkError,
   NotFoundError,
+  Server,
+  HorizonAxiosClient,
+  BASE_FEE,
+  Asset,
+  Operation as StellarOperation,
+  Account as StellarAccount,
+  Transaction as StellarTransaction,
```
haammar-ledger, line 16: Not super important, just to avoid confusion with the general pattern that's used in all families (and `StellarOperation` that's defined in `types.ts`)
```suggestion
  Operation as StellarSdkOperation,
  Account as StellarSdkAccount,
  Transaction as StellarSdkTransaction,
```

libs/ledger-live-common/src/families/stellar/api/horizon.ts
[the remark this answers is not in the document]
lvndry: Yes, it's not an optional value

libs/ledger-live-common/src/families/stellar/api/horizon.ts
lvndry: It's because we initialized the value just below, here we just making sure that the variable is not undefined

libs/ledger-live-common/src/families/stellar/logic.ts
lvndry: `isValidMed25519PublicKey` returns a Boolean

libs/ledger-live-common/src/families/stellar/api/horizon.ts
```diff
@@ -107,24 +121,25 @@ export const fetchAccount = async (
...
+    balance =
+      account.balances.find(balance => {
+        return balance.asset_type === "native";
+      })?.balance || "0";
     // Getting all non-native (XLM) assets on the account
-    assets = account.balances?.filter(balance => {
+    assets = account.balances.filter(balance => {
       return balance.asset_type !== "native";
-    });
+    }) as BalanceAsset[];
```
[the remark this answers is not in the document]
lvndry: Yeah, I think it's a legacy thing but the type of `account.balances` is `Horizon.BalanceLine` but the function expects assets to be of type `BalanceAsset` it demanded too much change so I decided to keep the `as` for now

libs/ledger-live-common/src/families/stellar/api/horizon.ts
lvndry: Yeah I instead tried to follow what was already used in the file but I can use Bignumber (even though it might be overkill and parseInt should be enough)

libs/ledger-live-common/src/families/stellar/api/horizon.ts
lvndry: Okay I think that this lib is not the best maintained indeed if the types can't be trusted

libs/ledger-live-common/src/families/stellar/api/horizon.ts
lvndry: `withMuxing` is not a valid argument: https://stellar.github.io/js-stellar-sdk/Operation.html#.payment

libs/ledger-live-common/src/families/stellar/api/horizon.ts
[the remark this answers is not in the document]
lvndry: I think the former developers weren't aware that the type of `assets` was available so they created custom types
It would be a lot more changes just to remove `BalanceAsset` type everywhere

sprohaszka-ledger (approved):

PR #5039, opened 2023-10-12
fix: fix walletAPI fee strategy integration

sarneijim:
### 📝 Description

Unable to change speed transaction in swap-web-app

Wallet-API fee strategy use upper case while LL is using lower case.

### ❓ Context

- **Impacted projects**: LLD, LLM (although I do not think no one is using those method yet)
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9727

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/sharp-avocados-turn.md
+++ b/.changeset/sharp-avocados-turn.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Fix WalletAPI fee strategy implementation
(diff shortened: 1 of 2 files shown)


ComradeAERGO:
@Justkant Is it normal that Wallet API and LL don't follow the same convention on this?

Justkant:
@ComradeAERGO We probably didn't see that when migrating from `live-app-sdk` https://github.com/LedgerHQ/live-app-sdk/blob/main/src/types.ts#L60-L67
We probably need to change that in the wallet-api repo too

PR #4207, opened 2023-08-03
Bot : currencies/family filter, disable & mooncake/silicium non reg removed

hedi-edelbloute:
### 📝 Description

Various update to the bot and workflows so :

- we can **filter** by a list of currencies/families
- we can **disable** a list of currencies/families
- we can see in the action how we run the bot : **daily / weekly basis**
- **disabled job starts on push** because we use workflow dispatch
- non reg on **mooncake, silicium disabled** as we already run those tests on other seeds

**Disabled on non reg** : Bitcoin, Ethereum as it costs too much and we test on testnets

**Weekly basis**: qtum,decred,cardano,axelar,cosmos,secret_network,avalanche_c_chain,bsc,filecoin,tron,cronos,fantom,boba,telos_evm,polygon_zk_evm,polkadot

**Daily basis** : the rest

I ignored coins that don't return addresses (evmos, flare...) as it is related to a problem

### ❓ Context

- **Impacted projects**: `LLC` 
- **Linked resource(s)**: `https://ledgerhq.atlassian.net/browse/LIVE-8283` 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #5040, opened 2023-10-12
[support] Fixes for CLI commands `getAddress` and `bot`

haammar-ledger:
### 📝 Description

- Fix `getAddress` command: remove a useless line causing `--derivationMode` param to be broken
- Fix `bot` command: adapt to recent changes in https://github.com/LedgerHQ/ledger-live/pull/4207 that broke the `--currency` param

### ❓ Context

- **Impacted projects**: `CLI` 
- **Linked resource(s)**: n/a 

### ✅ Checklist

- [ ] **Test coverage** :x: CLI commands aren't tested
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/apps/cli/src/commands/bot.ts
+++ b/apps/cli/src/commands/bot.ts
@@ -30,6 +30,12 @@ export default {
       throw new Error("Please define a SEED env variable to run this bot.");
     }
 
+    if (arg.currency) {
+      // Remapping to match arg format in libs/ledger-live-common/src/bot/cli.ts
+      arg.filter = { currencies: arg.currency };
+      delete arg.currency;
+    }
+
     return from(bot(arg));
   },
 };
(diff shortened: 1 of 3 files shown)


apps/cli/src/commands/bot.ts
```diff
@@ -30,6 +30,12 @@ export default {
       throw new Error("Please define a SEED env variable to run this bot.");
     }
 
+    if (arg.currency) {
+      // Remapping to match arg format in libs/ledger-live-common/src/bot/cli.ts
+      arg.filter = { currencies: arg.currency };
```
hedi-edelbloute, line 35: i think it should be currencies: [arg.currency] but i'm not sure

haammar-ledger: You're right :+1: thanks.

But it already works without it :open_mouth: which is why I didn't notice.
Looking at it, it's actually a very lucky case, because I'm seeing that only `.length` and `.includes` are used on `currencies`, which work for both string and array :smile:

hedi-edelbloute (approved):

apps/cli/src/commands/bot.ts
hedi-edelbloute: type doesn't cry but on code execution it's another story :smile:

PR #5041, opened 2023-10-12
test: fix LIVE-9694

gre:
### 📝 Description

Fixes LIVE-9694 by effectively stopping Lottie animations.

### ❓ Context

- **Impacted projects**: `LLD tests` 
- **Linked resource(s)**: [LIVE-9694] 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-9694]: https://ledgerhq.atlassian.net/browse/LIVE-9694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/animations/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/animations/index.tsx
@@ -22,8 +22,10 @@ const Animation = ({
   rendererSettings?: LottieProps["options"]["rendererSettings"];
   isPaused?: boolean;
   isStopped?: boolean;
-}) =>
-  animation ? (
+}) => {
+  // in case of playwright tests, we want to completely stop the animation
+  const isPlaywright = !!getEnv("PLAYWRIGHT_RUN");
+  return animation ? (
     <Flex
       style={{
         maxHeight: `200px`,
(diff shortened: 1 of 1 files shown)


apps/ledger-live-desktop/src/renderer/animations/index.tsx
```diff
@@ -22,8 +22,10 @@ const Animation = ({
   rendererSettings?: LottieProps["options"]["rendererSettings"];
   isPaused?: boolean;
   isStopped?: boolean;
-}) =>
-  animation ? (
+}) => {
+  // in case of playwright tests, we want to completely stop the animation
+  const isPlaywright = !!getEnv("PLAYWRIGHT_RUN");
+  return animation ? (
```
KVNLS, line 28: Here the component changes its behavior for tests, can we find another way to do it? I think that we will have better separation if components are not aware of the external world.
A solution I have in mind is to mock the Lottie component that returns the Lottie component itself with enforced autoplay: false
`LottieMock = props => <Lottie {...props, options: {...props.options, autoplay: false}} />`

That way we don't change our code for playwright + if someone uses the Lottie component elsewhere we don't break screenshots. wdyt?

gre: I agree with you and it's indeed a bad pattern we use not just here.
it will require more time to come up to a proper solution tho, we'll need to iterate on this.
are you suggesting that we then introduce another env (other than PLAYWRIGHT_RUN) to trigger the LottieMock mode?

KVNLS: Thanks for the explanation, I was not aware that we can't mock modules with playwright.
Then the solution you propose is the best one.
The other solution we have is to use the mask but it will require updating every test that uses a page with a Lottie file + we will lose the info of which Lottie file is used on the screenshot, so it's a poor solution.

KVNLS (approved):

beths-ledger (approved):
In-keeping with what we do elsewhere, and good to see the comments

PR #5042, opened 2023-10-12
[LIVE-9653][LIVE-9691] Bugfix - Fix incorrect token id encoding in account ids & incorrect BSC tokens type

lambertkevin:
### 📝 Description

This PR is fixing 2 problems:
-  the token ids encoding in `TokenAccounts` ids could contain characters not handled by the `
encodeURIComponent` (like `_` `-` or `+`), which would then break its "deserialization". 
    - https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-framework/src/account/accountId.ts#L36 
    - https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-framework/src/operation.ts#L70
    -  https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-framework/src/operation.ts#L93
- the BSC tokens were added to the CAL using the `convertERC20` method instead of the `convertBEP20` method, changing the ids from `bsc/bep20/XYZ` to `bsc/erc20/XYZ` which was breaking the mappings we have in some Ledger backends. ⚠️ Having 2 disctinct types for `erc20` and `bep20` is technically **useless**, as they're the same thing. This should be removed ASAP when the backends are capable of doing the migration of ids on their side.
 
### ❓ Context

- **Impacted projects**: `@ledgerhq/coin-framework` `@ledgerhq/coin-evm` 
- **Linked resource(s)**: 
- https://ledgerhq.atlassian.net/browse/LIVE-9653
- https://ledgerhq.atlassian.net/browse/LIVE-9691

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-framework/src/account/accountId.ts
+++ b/libs/coin-framework/src/account/accountId.ts
@@ -11,6 +11,26 @@ function ensureNoColon(value: string, ctx: string): string {
   return value;
 }
 
+export function safeEncodeTokenId(tokenId: string): string {
+  if (!tokenId) return "";
+
+  const URIEncoded = encodeURIComponent(tokenId);
+  const dashUnderscoreSafe = URIEncoded.replace(/-/g, "~!dash!~").replace(/_/g, "~!underscore!~");
+
+  return dashUnderscoreSafe;
+}
+
+export function safeDecodeTokenId(encodedTokenId: string): string {
+  if (!encodedTokenId) return "";
+
+  const dashUnderscoreUnsafe = encodedTokenId
+    .replace(/~!dash!~/g, "-")
+    .replace(/~!underscore!~/g, "_");
+  const decodedURIComponent = decodeURIComponent(dashUnderscoreUnsafe);
+
+  return decodedURIComponent;
+}
+
 export function encodeAccountId({
   type,
   version,
@@ -27,14 +47,14 @@ export function encodeAccountId({
   )}`;
 }
 export function encodeTokenAccountId(accountId: string, token: TokenCurrency): string {
-  return accountId + "+" + encodeURIComponent(token.id);
+  return accountId + "+" + safeEncodeTokenId(token.id);
 }
 export function decodeTokenAccountId(id: string): {
   accountId: string;
   token: TokenCurrency | null | undefined;
 } {
   const [accountId, tokenId] = id.split("+");
-  const decodedTokenId = decodeURIComponent(tokenId);
+  const decodedTokenId = safeDecodeTokenId(tokenId);
   let token = findTokenById(decodedTokenId);
   if (!token) {
     const { currencyId } = decodeAccountId(accountId);
--- a/.changeset/plenty-hairs-hear.md
+++ b/.changeset/plenty-hairs-hear.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-evm": minor
+---
+
+Ensure compatiblity with @ledgerhq/coin-framework change regarding `TokenAccounts` ids & update `preload` & `hydrate` method to use the `convertBEP20` method instead of the `convertERC20` method when dealing with the bsc chain. This last fix is technically useless as BEP20 = ERC20, it's only here in order to prevent breaking changes with the different backends at Ledger. As soon as those backends have the possibility of changing the "bsc/bep20/XYZ" token ids into "bsc/erc20/XYZ", this fix should be removed in order to avoid useless complexity.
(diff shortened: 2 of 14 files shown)


libs/coin-framework/src/account/accountId.ts
```diff
@@ -11,6 +11,43 @@ function ensureNoColon(value: string, ctx: string): string {
...
+  const URIEncoded = encodeURIComponent(tokenId);
+  const dashUnderscoreSafe = URIEncoded.replace(/-/g, "~!dash!~")
+    .replace(/_/g, "~!underscore!~")
+    .replace(/\+/g, "~!plus!~");
+
+  invariant(
+    !dashUnderscoreSafe.includes("-") &&
+      !dashUnderscoreSafe.includes("_") &&
+      !dashUnderscoreSafe.includes("+"),
+    `TokenId ${dashUnderscoreSafe} shouldn't contain - or _`,
```
gre, line 26: you can remove this invariant. the code before it makes this case impossible. that way we can optim a bit the perf (includes can be costly as i assume we call this function a lot)

libs/coin-framework/src/account/accountId.ts
```diff
@@ -11,6 +11,43 @@ function ensureNoColon(value: string, ctx: string): string {
...
+    .replace(/~!underscore!~/g, "_")
+    .replace(/~!plus!~/g, "+");
+  const decodedURIComponent = decodeURIComponent(dashUnderscoreUnsafe);
+
+  invariant(
+    !decodedURIComponent.includes("~!dash!~") &&
+      !decodedURIComponent.includes("~!underscore!~") &&
+      !decodedURIComponent.includes("~!plus!~"),
+    `TokenId ${decodedURIComponent} shouldn't contain ~!dash!~, ~!underscore!~ or ~!plus!~`,
+  );
```
gre, line 46: same

gre (commented):
ugly but necessary fix. I don't see another solution at the moment. 😢 

I think we shouldn't have the invariant in the code runtime because it doesn't help perf & also it's provable by the code that these are never possible (whatever the user's input param are), you have test that also will help on coverage

libs/coin-framework/src/account/accountId.ts
lambertkevin: Alright, done 👌

libs/coin-framework/src/account/accountId.ts
lambertkevin: Also done 👌

haammar-ledger (commented):
@lambertkevin Maybe `encodeCurrencyAsLedgerId` to cover too in `libs/ledger-live-common/src/countervalues/helpers.ts`?

lambertkevin:
> encodeCurrencyAsLedgerId

No idea what this thing is, I'll have a look 👌

lambertkevin:
> LGTM but regarding this:
> 
> > Ok just had a look, and this encodePairAsLedgerIdPair should not be impacted by the issue I'm fixing I guess because It's never "decoded" nor included in something that would then be splitted like the operation Ids or account Ids
> 
> Don't we want to be future-proof? Since it's the same situation that lead us here of the problem being invisible as long as it was never decoded. That said I didn't investigate at all what it's used for, and probably not an issue today indeed according to what you say.

@haammar-ledger Well the problem here was only happening because of this double serialization: a token id was "encoded", then added to a serialized `TokenAccount` id, which was also serialized into an `Operation` id. Going full reverse on this was problematic because you were deserializing something with a `string.split("X")`, and if that `X` was not safely removed during the encoding, then your whole deserialization was fucked, and you were splitting the string in the wrong places. 
As long as you don't have this "split" mecanism, using -_+ or anything in the token id is fine I guess.

So in the case of `encodePairAsLedgerIdPair`, it looks like it's a function used for counter values, and this time it's litteraly added to an URI:

```typescript
const latest = async (pairs: TrackingPair[], direct?: boolean) => {
  const all = await promiseAllBatched(4, chunk(pairs, LATEST_CHUNK), async partial => {
    const url = `${baseURL()}/v2/latest/${direct ? "direct" : "indirect"}?pairs=${partial
      .map(p => encodePairAsLedgerIdPair(p.from, p.to))
      .join(",")}`;
    const { data } = await network({ method: "GET", url });
    return data;
  });
  const data = all.reduce((acc, data) => acc.concat(data), []);
  return data;
};
```

So my guess is that changing anything in the encoding of `encodePairAsLedgerIdPair` would actually make the "pairs" break on the backend side, as they wouldn't know they need to replace the `~!dash!~` with a `-` etc.

PR #5138, opened 2023-10-19
Bugfix/8850 add battery check during ESC

aussedatlo:
### 📝 Description

Move the battery check drawer inside the `FirmwareUpdate` screen to check battery status during all updates including early security check.

### ❓ Context

- **Impacted projects**: `llm` `live-common` `live-env`
- **Linked resource(s)**: [LIVE-8850]

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/35492518/446400d1-aa5d-4ef6-a8c5-c53e0ec2fd82

### 🚀 Expectations to reach


apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -165,6 +173,19 @@ export const FirmwareUpdate = ({
...
+  const [showBatteryWarningDrawer, setShowBatteryWarningDrawer] = useState<boolean>(false);
+  const [showReleaseNotes, setShowReleaseNotes] = useState<boolean>(true);
+
+  const {
+    requestCompleted: batteryRequestCompleted,
+    batteryStatusesState,
+    triggerRequest: triggerBatteryCheck,
+    cancelRequest: cancelBatteryCheck,
+    isBatteryLow,
+  } = useBatteryStatuses({
```
alexandremgo, line 185: We'll need to see how we can handle this for the nano X when we enable the fw update via BLE

BUT actually: this screen/flow is used when updating a nano device via USB on Android ! Not sure it works with nano SP and nano X
(i'm almost sure the APDU does not exist on them, or at least on older fw versions)

You should double check, and if it does not work, either ignore the error or disable the hook for them

alexandremgo: And maybe it's better to handle this business logic: "is the battery check available on this current device" be handled on live-common in the action/hook (and not here in this component). It could return a state saying that the battery check is not available

apps/ledger-live-mobile/src/screens/FirmwareUpdate/BatteryWarningDrawer.tsx
```diff
@@ -0,0 +1,52 @@
...
+type Props = QueuedDrawerProps & {
+  state: GetBatteryStatusesActionState;
+  device: Device;
+  onRetry: () => void;
+  onQuit: () => void;
+};
+
+const BatteryWarningDrawer: React.FC<Props> = ({ device, state, onRetry, onQuit, ...props }) => {
+  const { t } = useTranslation();
+  const lowBatteryPercentage = useEnv("LOW_BATTERY_PERCENTAGE");
```
alexandremgo, line 18: Maybe inject this value into this component : you're already getting it from the hook with `isBatteryLow` if i'm not wrong ;)

alexandremgo (changes requested):
Great work ! Some changes are needed on the devices-"business logic" part

apps/ledger-live-mobile/src/screens/FirmwareUpdate/BatteryWarningDrawer.tsx
aussedatlo: `isBatteryLow` is a `boolean` that indicate whether the battery is low and `useEnv("LOW_BATTERY_PERCENTAGE")` represents the percentage at which the battery is considered low.

In this case, we need the percentage to update the i18n text, rather than the boolean.

So, you mean that I can make this number accessible through the hook, in addition to `isBatteryLow` and give it to this component with a props ?

alexandremgo: Hehe yes i went too fast: i would say what you suggest: returning `lowBatteryPercentage` from the hook, and give it to this component with a prop ;)

apps/ledger-live-mobile/src/screens/SyncOnboarding/EarlySecurityCheck.tsx
```diff
@@ -331,6 +331,11 @@ export const EarlySecurityCheck: React.FC<EarlySecurityCheckProps> = ({
         ) {
           setFirmwareUpdateCheckStatus("completed");
           currentDisplayedDrawer = "none";
+        } else if (isAlreadyGenuine) {
+          // When isAlreadyGenuine is true, it indicates that the user is already informed
+          // about the new firmware update, so there's no need to show the drawer again
+          currentDisplayedDrawer = "none";
+          firmwareUpdateUiStepStatus = "firmwareUpdateRefused";
```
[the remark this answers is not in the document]
alexandremgo: Perfect with the `isPreviousUpdateCancelled` ;)

apps/ledger-live-mobile/src/screens/FirmwareUpdate/BatteryWarningDrawer.tsx
aussedatlo: Done

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
aussedatlo: The new nano X respond correctly to the APDU with a firmware of `v2.2.2`. The hook handle the battery correctly and notify that the device is plugged in. In fact, there is no battery warning.

Therefore, on older firmware, such as `v2.0.0` this APDU does not exist. So, I need to ensure that the drawer doesn't block the flow.

aussedatlo: With new modifications, older firmware are now responding correctly

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
```diff
@@ -509,6 +538,17 @@ export const FirmwareUpdate = ({
       );
     }
 
+    if (batteryStatusesState.error !== null) {
```
alexandremgo, line 541: Just: in that cas are we not displaying an error message if we get a state with `error: { type: "UnknownApdu" }` ? I don't see where we stop doing the battery check if the apdu is unknown ?

apps/ledger-live-mobile/src/screens/SyncOnboarding/EarlySecurityCheck.tsx
```diff
@@ -71,6 +74,11 @@ export type EarlySecurityCheckProps = {
    */
   isAlreadyGenuine?: boolean;
 
+  /**
+   * To tell the ESC that there is no need to re display the FW update drawer (optionnal)
```
alexandremgo, line 78: `optional` ;)

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
aussedatlo: If there is an `UnknownApdu` event type, there is no errors since it's an event with no other attribute than `type`

apps/ledger-live-mobile/src/screens/SyncOnboarding/EarlySecurityCheck.tsx
aussedatlo: Done

libs/ledger-live-common/src/hw/getBatteryStatus.ts
```diff
@@ -35,13 +36,17 @@ const getBatteryStatus = async <StatusesType extends ReadonlyArray<BatteryStatus
   transport: Transport,
   statuses: StatusesType,
 ): Promise<BatteryStatusTuple<StatusesType>> => {
+  const tracer = new LocalTracer("hw", { statuses });
```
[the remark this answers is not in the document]
aussedatlo: Nice, thx for the tips

apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx
alexandremgo: But the `getBatteryStatusesAction` has an `error` prop in it's `state` (in `libs/ledger-live-common/src/deviceSDK/actions/getBatteryStatuses.ts`)

```TS
export type GetBatteryStatusesActionState = FullActionState<{
  batteryStatuses: (number | BatteryStatusFlags)[];
  error: { type: GetBatteryStatusesActionErrorType; message?: string; name?: string } | null;
}>;
```

And we update this state when receiving an event of type `taskError`:
```TS
case "taskError":
          return { ...initialState, error: { type: event.error } };
```

Which is the event (`taskError`) that we get from the command:
```TS
        catchError((err: Error) => {
          if (err instanceof TransportStatusError && err.statusCode === StatusCodes.UNKNOWN_APDU)
            return of<GetBatteryStatusesTaskEvent>({ type: "taskError", error: "UnknownApdu" });

          return throwError(() => err);
        }),
```

So at the end, we have a `batteryStatusesState` in `libs/ledger-live-common/src/deviceSDK/hooks/useBatteryStatuses.ts` that will be set with an error of type `UnknownApdu`

No ?

aussedatlo: If we have a `UnknownApdu` error, from `libs/ledger-live-common/src/deviceSDK/hooks/useBatteryStatuses.ts` we have an `event` and not an `error` in `getBatteryStatusesAction` `next` callback. So since we don't have any flags, we exit without setting `isBatteryLow`. And since it's an event, we don't have error either.

```
const sub = getBatteryStatusesAction({
        deviceId,
        statuses,
      }).subscribe({
        next: state => {
          setBatteryStatusesState(state);

          // no battery status flags available
          if (state.batteryStatuses.length <= 1) return;

          const [percentage, statusFlags] = state.batteryStatuses as [number, BatteryStatusFlags];

          if (percentage < lowBatteryPercentage && statusFlags.charging === 0) {
            setIsBatteryLow(true);
          }
        },
        complete: () => {
          setRequestCompleted(true);
        },
        error: err => log("error", "error while retrieving the battery statuses", err),
      });
```

So from `apps/ledger-live-mobile/src/screens/FirmwareUpdate/index.tsx`, `isBatteryLow=false` so no warning battery and since the error is handled as an event, no `batteryStatusesState.error` 😃

alexandremgo (approved):
Nice work 👌

PR #5171, opened 2023-10-23
Fix/LLM - BLE Transport: Not being able to communicate with device after seed generation

alexandremgo:
### 📝 Description

When the seed is being persisted on the device, all communication are stopped. But, at least via BLE (this might change for HID USB), the host/client can still connect to the device, just not send it any message.

Because of how/where the `getVersion` request was (not correctly) being aborted and retried (for the polling mechanism), requests were stacked and blocked, and no future requests could be sent to the device even once the communication with it was re-opened.

The fix contains:
- an (optional and configurable) abort timeout mechanism on APDU exchange that is at the lowest level possible of our JavaScript codebase: in `Transport.exchange`. `Promise` cannot be aborted, but `Observable` kind of can be: if `Promise` are used at one point in the inner function of the `Observable`, the  
- usage of `ble` cancel to actually cancel the pending operations when the abort timeout is reached (the message being written and waiting for )
- an (optional and configurable) abort timeout when "opening" a Transport to avoid hanging too long, especially with the BleTransport on which we could still "connect" to the device but not (exchange to) negotiate the MTU

TODO

### ❓ Context

- **Impacted projects**: `live-common`, `@ledgerhq/devices`, `@ledgerhq/errors`, `@ledgerhq/hw-transport`, `@ledgerhq/react-native-hw-transport-ble`, LLM
- **Linked resource(s)**: [LIVE-6275](https://ledgerhq.atlassian.net/browse/LIVE-6275)

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach


libs/ledgerjs/packages/errors/src/index.ts
```diff
@@ -44,6 +44,7 @@ export const DeviceExtractOnboardingStateError = createCustomErrorClass(
 export const DeviceOnboardingStatePollingError = createCustomErrorClass(
   "DeviceOnboardingStatePollingError",
 );
+export const ExchangeTimeoutError = createCustomErrorClass("ExchangeTimeoutError");
```
ofreyssinet-ledger, line 47: "Exchange" naming is a bit ambiguous (it could refer to the Exchange app for swapping crypto), could it be renamed to something more Transport related like MessageExchangeTimeoutError or something like that a bit more explicit?

libs/ledgerjs/packages/hw-transport/src/Transport.ts
```diff
@@ -294,8 +313,18 @@ export default class Transport {
...
   exchangeBusyPromise: Promise<void> | null | undefined;
-  exchangeAtomicImpl = async (f: () => Promise<Buffer | void>): Promise<Buffer | void> => {
+
+  /**
+   * Wrapper to make an exchange "atomic" (blocking any other exchange)
+   *
+   * It also handles "unresponsiveness" by emitting "unresponsive" and "responsive" events.
+   *
+   * @param f The exchange job, using the transport to run
+   * @returns a Promise with a Buffer containing the response, or void if not response
```
ofreyssinet-ledger, line 325: now that the type is generic maybe you should slightly update this comment ?

libs/ledgerjs/packages/errors/src/index.ts
alexandremgo: yeah you are right !

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
```diff
@@ -633,8 +771,13 @@ export default class BleTransport extends Transport {
         mtu = await firstValueFrom(
           merge(
             this.notifyObservable.pipe(
-              tap(maybeError => {
-                if (maybeError instanceof Error) throw maybeError;
+              map(maybeError => {
```
ofreyssinet-ledger, line 774: the point of replacing the `tap` by a `map` here is to guarantee that the resulting observable (created by `pipe`) errors ? if you just throw an error in `tap` it doesn't affect the overall observable created by pipe ?

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
```diff
@@ -685,19 +829,36 @@ export default class BleTransport extends Transport {
...
+    const transactionId = uuid();
+    this.currentTransactionIds.push(transactionId);
+
+    const tracer = this.tracer.withUpdatedContext({ transactionId });
+    tracer.trace("Writing to device", { willMessageBeAcked: !this.writeCmdCharacteristic });
+
     try {
       if (!this.writeCmdCharacteristic) {
-        await this.writeCharacteristic.writeWithResponse(buffer.toString("base64"), txid);
+        // The message will be acked by the device
```
ofreyssinet-ledger, line 850: is this availability of writing with response depending on the device ?
btw I'm wondering if we shouldn't use a wording closer to the react-native-ble-plx API, also not starting with a verb which to me means it's a function/method.
IMO it would improve greatly the readability for new eyes.
And those properties are only accessed in this file IMO so no risk of impacting other stuff.
Something like `characteristicWriteableWithResponse`/`characteristicWriteableWithoutResponse`.

libs/ledgerjs/packages/devices/src/ble/sendAPDU.ts
```diff
@@ -1,33 +1,66 @@
...
+/**
+ * Sends an APDU by encoding it into chunks and sending the chunks using the given `write` function
+ *
+ * @param write The function to send each chunk to the device
+ * @param apdu
+ * @param mtuSize The negotiated maximum size of the data to be sent in one chunk
+ * @param options Optional options containing:
+ *   - context An optional context object for log/tracing strategy
+ * @returns An observable that will only emit if an error occurred, otherwise it will complete
+ */
```
ofreyssinet-ledger, line 39: 🙌 great description thanks a lootttt

libs/ledgerjs/packages/devices/src/ble/receiveAPDU.test.ts
```diff
@@ -0,0 +1,116 @@
...
+            done();
+          } catch (error) {
+            done(error);
+            return;
+          }
+        },
+      });
+    });
+  });
+});
```
ofreyssinet-ledger, line 116: you're a hero Alex

libs/ledgerjs/packages/devices/src/ble/receiveAPDU.ts
```diff
@@ -65,11 +94,13 @@ export const receiveAPDU = (rawStream: Observable<Buffer>): Observable<Buffer> =
         if (notifiedData.length === notifiedDataLength) {
           o.next(notifiedData);
           o.complete();
-          sub.unsubscribe();
+          // Tries to unsubscribe from the raw stream as soon as we complete the parent observer
+          subscriptionCleaner.next();
```
ofreyssinet-ledger, line 98: what's the advantage of doing this over `sub.unsubscribe()`?

libs/ledgerjs/packages/devices/src/hid-framing.test.ts
```diff
@@ -0,0 +1,148 @@
+import createHIDframing, { ResponseAcc } from "./hid-framing";
+
+describe("createHIDframing.makeBlocks", () => {
+  const channel = Math.floor(Math.random() * 0xffff);
```
ofreyssinet-ledger, line 4: what's the point of adding some randomness here?
couldn't we just use one arbitrary value?

libs/ledger-live-common/src/hw/index.ts
```diff
@@ -107,7 +112,45 @@ export const open = (
...
+      return Promise.race([
+        p.then(transport => {
+          // Necessary to stop the ongoing timeout
+          if (timer) {
+            clearTimeout(timer);
+          }
+
+          return transport;
+        }),
+        new Promise((_resolve, rejects) => {
```
ofreyssinet-ledger, line 141: why the weird naming for `resolve` and `reject`?

libs/ledger-live-common/src/hw/index.ts
```diff
@@ -90,9 +90,14 @@ export const discoverDevices = (
 /**
  * Tries to call `open` on the 1st matching registered transport implementation
  *
+ * An optional timeout can be added. A timeout is applied directly in this function (racing between the matching Transport and the timeout).
+ * But also passed to each `open` Transport implementation. As there is no easy way to abort a Promise (returned by `open`), the Transport will
+ * continue to try connecting to the device even if this function timeout was reached. But certain Transport implementations can also use this
+ * timeout to try to stop the connection attempt.
```
ofreyssinet-ledger, line 96: maybe make it more explicit how it works: the `timeoutMs` parameter is passed to the `open` method of the transport module so each transport implementation can make use of that parameter and implement the timeout internally.

libs/ledger-live-common/src/hw/index.ts
```diff
@@ -107,7 +112,45 @@ export const open = (
...
+        return p;
+      }
+
+      let timer: ReturnType<typeof setTimeout> | null = null;
+
+      // Throws CantOpenDevice on timeout, otherwise returns the transport.
+      // Important: with Javascript Promise, when one promise finishes,
+      // the other will still continue, even if its return value will be discarded.
+      return Promise.race([
+        p.then(transport => {
```
ofreyssinet-ledger, line 133: have you tested the (possible ? idk) case where `p` throws an error because the timeout implemented in `open()` times out (and thus throws an error) before the `setTimeout` called here ?

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
alexandremgo: `tap` is supposed to be for "side effect" like changing an outside value or logging. But looking at ex [in the doc](https://rxjs.dev/api/operators/tap) they show an ex where they throw an error from `tap` ahah. So i guess it was fine and i was wrong to change it. I can put it back if you think it's clearer ? I just feel that `map` indicates that something is going to happen to the streamed value, while `tap` does not touch it

And to answer your 2nd question: 
>Any error that occurs -- synchronously -- in a handler provided to `tap` will be emitted as an error from the returned observable.

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
alexandremgo: I would say that it's decided during the "hand-shake" when the BLE communication is started with the device. We should search and add it to the documentation we wanted to do.
But my understanding is that we know the UUID of a "write with response/ack" characteristic and the UUID of a "write without response" characteristic for each device (defined in `libs/ledgerjs/packages/devices/src/index.ts`, there is the "notify characteristic" too), and when the device gives us the service and the characteristics we can use to communicate with it, we check which one we have access to.
And we always check that the "write with response" characteristic (`writeC`) exists (also the "notify characteristic") in the `open` function.

alexandremgo: and for the name i agree with you

alexandremgo: Just maybe:
- `writeableWithResponseCharacteristic`
- `writeableWithoutResponseCharacteristic`

libs/ledgerjs/packages/devices/src/ble/receiveAPDU.ts
```diff
@@ -1,61 +1,90 @@
...
+
+/**
+ * Parses a raw stream coming from a BLE communication into an APDU response
+ *
+ * @param rawStream An observable containing the raw stream as emitted buffers
+ * @param options Optional options containing:
+ *   - context An optional context object for log/tracing strategy
+ * @returns An observable containing the APDU response as one emitted buffer
+ */
+export const receiveAPDU = (
```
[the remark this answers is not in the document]
alexandremgo: I think `parseBLERawStreamIntoAPDU` would be better yeah - It's just going to need changes in the 3 BLE transport (node and web too), and if any project uses it directly, it will be a breaking change. Maybe we should do this in another PR ?

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
ofreyssinet-ledger: perfect 👍

libs/ledgerjs/packages/react-native-hw-transport-ble/src/BleTransport.ts
ofreyssinet-ledger: nah no need to change it back 👍

libs/ledgerjs/packages/devices/src/ble/receiveAPDU.ts
alexandremgo: Good question: we're inside the definition of `next`, and we could do:
```
const subscription = rawStream.subscribe({
	next: () => {
		...
		subscription.unsubscribe();
   }
})
```

But at the moment of the `next` definition, `subscription` does not yet exist (and i TS might tell you you are using a variable not yet defined).

libs/ledgerjs/packages/devices/src/hid-framing.test.ts
alexandremgo: Just to show that it should work with whatever value for the channel (as this `channel` value is also randomly generated when we instantiate a USB HID Transport (for ex: in [TransportNodeHidNoEvents](https://github.com/LedgerHQ/ledger-live/blob/aa27e0f1e8043d0814d0f548cf52fb6d872a48ae/libs/ledgerjs/packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts#L81) )

libs/ledgerjs/packages/devices/src/ble/receiveAPDU.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: yep 👍

libs/ledger-live-common/src/hw/index.ts
alexandremgo: - `resolve` is never used, so the `_` is for the linter
- `rejects` i don't know why i added an `s` ahah, removing it

libs/ledgerjs/packages/devices/src/ble/receiveAPDU.ts
ofreyssinet-ledger: it was working prior to this change though no ?

ofreyssinet-ledger: but yeah agreed it was weird

libs/ledger-live-common/src/hw/index.ts
alexandremgo: If the Promise `p` throws an error because of a timeout before this `setTimeout`, it is fine because the `Promise.race` will only reject at the 1st rejection/error (it's not going to/cannot reject twice because the 2 threw an error each)

I can look to test this case, but not sure

libs/ledger-live-common/src/hw/deviceAccess.ts
```diff
@@ -100,34 +101,98 @@ export const cancelDeviceAction = (transport: Transport): void => {
...
+      // When the new job is finished, this will unlock the associated device queue of jobs
+      let unlockQueuedJob;
+
+      const jobId = queuedJobManager.setLastQueuedJob(
+        deviceId,
+        new Promise(resolve => {
+          unlockQueuedJob = resolve;
+        }),
+      );
+
```
ofreyssinet-ledger, line 193: could we rename `unlockQueuedJob` into `resolveQueuedJob` ?

ofreyssinet-ledger (approved):
LGTM great job, 👏  for the docs and new tests !

libs/ledger-live-common/src/hw/index.ts
alexandremgo: fixed ;)

libs/ledger-live-common/src/hw/deviceAccess.ts
alexandremgo: done ;)

libs/ledger-live-common/src/hw/index.test.ts
```diff
@@ -0,0 +1,97 @@
+import { aTransportBuilder } from "@ledgerhq/hw-transport-mocker";
+import { registerTransportModule, open } from ".";
+import { CantOpenDevice, TransportError } from "@ledgerhq/errors";
+
+jest.useFakeTimers();
+
+describe("open", () => {
+  describe("When no transport working with the device has been registered", () => {
+    it("should return the associated opened Transport instance", async () => {
```
ofreyssinet-ledger, line 9: wrong description here, should be "should throw a CantOpenDevice error"

alexandremgo: oupsi

PR #5245, opened 2023-10-30
[BUGFIX] usage of explorer block endpoint in coin evm

chabroA:
### 📝 Description

For some legacy reasons, there is a difference in the response shape of `/blockchain/v4/eth/block/current` and `/blockchain/v4/eth/block/{height}`.

The first endpoint (`/current`) returns a block object in the shape
```
{
  "hash": "string",
  "height": 1073741824,
  "time": "string",
  "txs": [
    "string"
  ]
}
```
Whereas the second endpoint (`{height}`) returns an array of block objects (but with only one block inside)
```
[
  {
    "hash": "string",
    "height": 1073741824,
    "time": "string",
    "txs": [
      "string"
    ]
  }
]
```

cf. https://explorers.api.live.ledger.com/blockchain/v4/eth/docs/

The coin-evm implementation assumed both endpoints had the same response shape (the one returning a single object), leading to issues when the block heigh specific endpoint was called (values returned where null.

This PR also harmonise the `timestamp` value returned by `getBlockByHeight` to make it a milliseconds value.
Until then, the value returned by the different implementations (rpc and  ledger) where not consistent. The rpc implementation returned it in seconds while the ledger one returned it in milliseconds.
This leads to some issues when called inside the `getOperationStatus` (the only place it is called) since this function was multiplying the `timestamp` value returned by calls to `getBlockByHeight` by 1000 to convert it back to milliseconds (which lead to issues when the ledger implementation of `getBlockByHeight` was called).

We could have updated the ledger implementation to make it return the value in seconds instead of milliseconds but this has 2 downsides:
- loss of data when converting the timestamp from milliseconds to seconds in the ledger implementation
- unnecessary conversion since the value is the converted back to milliseconds to be used inside `getOperationStatus`

### ❓ Context

- **Impacted projects**: `coin-evm` 
- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9981
  - https://github.com/LedgerHQ/ledger-live/pull/4886

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-evm/src/api/node/ledger.ts
+++ b/libs/coin-evm/src/api/node/ledger.ts
@@ -283,7 +283,7 @@ export const getBlockByHeight: NodeApi["getBlockByHeight"] = async (
       url: `${getEnv("EXPLORER")}/blockchain/v4/${node.explorerId}/block/current`,
     });
 
-    return { hash, height, timestamp: Math.floor(new Date(time).getTime() / 1000) };
+    return { hash, height, timestamp: new Date(time).getTime() };
   }
 
   /**
@@ -307,7 +307,7 @@ export const getBlockByHeight: NodeApi["getBlockByHeight"] = async (
   return {
     hash,
     height,
-    timestamp: Math.floor(new Date(time).getTime() / 1000),
+    timestamp: new Date(time).getTime(),
   };
 };
 
--- a/libs/coin-evm/src/api/node/ledger.ts
+++ b/libs/coin-evm/src/api/node/ledger.ts
@@ -272,23 +272,42 @@ export const getBlockByHeight: NodeApi["getBlockByHeight"] = async (
     throw new LedgerNodeUsedIncorrectly();
   }
 
-  const { hash, height, time } = await fetchWithRetries<{
-    hash: string;
-    height: number;
-    time: string;
-    txs: string[];
-  }>({
+  if (blockHeight === "latest") {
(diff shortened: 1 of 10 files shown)


lvndry (approved):

hzheng-ledger (approved):
looks good to me

libs/coin-evm/src/api/node/ledger.ts
```diff
@@ -272,17 +272,36 @@ export const getBlockByHeight: NodeApi["getBlockByHeight"] = async (
     throw new LedgerNodeUsedIncorrectly();
   }
 
-  const { hash, height, time } = await fetchWithRetries<{
-    hash: string;
-    height: number;
-    time: string;
-    txs: string[];
-  }>({
+  if (blockHeight === "latest") {
```
haammar-ledger, line 275: @chabroA Has this been discussed with Backend Blockchain team? Unless there's a reason (which I doubt) it seems like it would better be fixed on backend side with a unique format.

haammar-ledger: After discussion, let's merge this then plan for later a proper backend fix and a rework of bitcoin & evm.

haammar-ledger (approved):

PR #5250, opened 2023-10-30
[SUPPORT] Remove necessary build workflow inputs

chabroA:
### 📝 Description

- remove `inputs.login` and use `github.actor` instead.
- remove `inputs.ref` in external build workflows since not used

### ❓ Context

- **Impacted projects**: `` 
- **Linked resource(s)**: [] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.github/workflows/build-desktop-external.yml
+++ b/.github/workflows/build-desktop-external.yml
@@ -1,15 +1,9 @@
 name: "@Desktop • Build App (external)"
-run-name: "@Desktop • Build App (external) triggered by ${{ inputs.login }}  ${{ format('on ref {0}', github.ref_name) }}"
+run-name: "@Desktop • Build App (external) triggered by ${{ github.actor }}  ${{ format('on ref {0}', github.ref_name) }}"
 
 on:
   workflow_dispatch:
     inputs:
-      ref:
-        description: the branch which triggered this workflow
-        required: false
-      login:
-        description: The GitHub username that triggered the workflow
-        required: true
       base_ref:
         description: The base branch to merge the head into when checking out the code
         required: false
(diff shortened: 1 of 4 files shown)


valpinkman (changes requested):
I'm sorry but you can't remove all of that as they are used by the github bot to trigger through api the workflows on the correct branch / ref.

You could use the github.actor in case inputs.login is emtpy, but it will always be `live github -bot`

chabroA:
> I'm sorry but you can't remove all of that as they are used by the github bot to trigger through api the workflows on the correct branch / ref (also manually)
> 
> You could use the github.actor in case inputs.login is emtpy, but it will always be `live github -bot`

Ok I see, thanks. Did some update to the PR and added some comments for future reference 📝 

From what I can see, `inputs.ref` does not seem to be used anywhere in the 2 "build external "workflows. Is it normal / expected? Or did I miss something in how it's used?
Regarding `inputs.login`, from what I understand the main (only) reason it exists is in the bot context, when called through API. Could we therefore update it as not required in the workflow file so that we easily and quickly manually run the workflows without having to fill this text field? Especially since in the case of manual run, the needed and correct value is already present in `github.actor`.

Screenshot 2023-10-31 at 10 09 51

valpinkman:
so, `ref` is also an input passed by the bot, basically, none of them are superfluous. However not making it required and updating the name of the workflows (and wherever `inputs.login` is used) by `inputs.login || github.actor` actually works fine and you are right, make it easier to run manually 👍

valpinkman (changes requested):
see previous comment

aboissiere-ledger:
@chabroA The orchestrator is handling this through the getInputs method: https://github.com/LedgerHQ/ledger-live/blob/develop/tools/github-bot/docs/orchestrator.md

chabroA:
Does not seem relevant/useful. Closing for now

PR #5252, opened 2023-10-30
Bugfix/recover onboarding device compat

ak-ledger:
### 📝 Description

Remove recover options for uncompatible devices (LLM/LLD)
Remove promotions after onboarding for uncompatible devices, based on wired device (LLD only)
### ❓ Context

- **Impacted projects**: `LLM, LLD` 
- **Linked resource(s)**: [LIVE-9980] 

### ✅ Checklist

- [ ] **Test coverage** 
- [ ] **Atomic delivery** 
- [ ] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/146743014/5b114f8f-03cc-402e-a3c0-48d44c59bfad

### 🚀 Expectations to reach

[LIVE-9980]: https://ledgerhq.atlassian.net/browse/LIVE-9980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/featureFlags/helper.tsx
```diff
@@ -0,0 +1,10 @@
+// Recover compatibility check function, based on feature and device
+export function checkRecoverCompatibility(feature, deviceModelId) {
```
KVNLS, line 2: I think that you can rename the function to remove the comment above with something like:
`checkRecoverAvailabality()` or `isRecoverDisplayed()`

`Compatibility` makes us think that it's only related to the device but it's also related to the feature flag.

libs/ledgerjs/packages/types-live/src/feature.ts
```diff
@@ -299,6 +299,11 @@ export type Feature_NewsfeedPage = Feature<{
 
 export type Feature_ProtectServicesMobile = Feature<{
   deeplink: string;
+  compatibleDevices: Array<{
+    available: boolean;
+    comingSoon: boolean;
+    name: string;
+  }>;
   onboardingRestore: {
```
KVNLS, line 307: Maybe we can use a type for `compatibleDevices` & `onboardingRestore` so we can reuse them, wdyt?

KVNLS (approved):
Works well, perfect!

libs/ledgerjs/packages/types-live/src/feature.ts
ak-ledger: Depends on the team sending us the infos. Is this a format they would like to keep and reuse for other purposes ?

KVNLS: I only suggest adding types so I'm sure to get your point.
``` 
  type CompatibleDevice = {
    available: boolean;
    comingSoon: boolean;
    name: string;
  };
  ...
  compatibleDevices: CompatibleDevice[];
  ```

PR #5260, opened 2023-10-31
fix: improve perf for sortByMarketCap

KVNLS:
### 📝 Description

Today we have some performance issues when we open the Swap page, because sortByMarkCap can be slow.
This PR improves the performance of the function.

### ❓ Context

- **Impacted projects**: `llm`, `lld` 
- **Linked resource(s)**: [] 
- https://ledgerhq.atlassian.net/browse/LIVE-9663
- https://ledgerhq.atlassian.net/browse/LIVE-9997

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Before
Screenshot 2023-10-30 at 17 14 04

After
Screenshot 2023-10-31 at 14 59 17

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/.unimportedrc.json
+++ b/libs/ledger-live-common/.unimportedrc.json
@@ -303,7 +303,8 @@
     "src/index.ts",
     "src/jest.d.ts",
     "src/postOnboarding/mock.ts",
-    "src/reactNativeSvg.d.ts"
+    "src/reactNativeSvg.d.ts",
+    "src/currencies/mock.ts"
   ],
   "ignoreUnused": [
     "@types/bchaddrjs",
--- a/libs/ledger-live-common/src/currencies/sortByMarketcap.ts
+++ b/libs/ledger-live-common/src/currencies/sortByMarketcap.ts
@@ -31,23 +31,33 @@ function lazyLoadTickerMap() {
   return m;
 }
 
+function getCurrencyId(currency: Currency): string {
+  if ("id" in currency && currency.id) {
+    return currency.id;
+  }
+  return currency.ticker;
+}
+
 export const sortByMarketcap = <C extends Currency>(currencies: C[], tickers: string[]): C[] => {
-  const m = lazyLoadTickerMap();
-  const list = currencies.slice(0);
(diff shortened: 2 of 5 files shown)


libs/ledger-live-common/src/currencies/sortByMarketcap.ts
```diff
@@ -31,23 +31,32 @@ function lazyLoadTickerMap() {
...
+    if (currency) {
+      sortedCurrenciesMap.set(getCurrencyId(currency), currency);
+    }
+  }
 
-      if (i !== -1) {
-        list.splice(i, 1);
-        prependList.push(item);
-      }
+  currencies.forEach(currency => {
```
gre, line 53: could you also use `for of` syntax here so we have similar code style as the previous for loop?

alternatively it seems the logic here is also fundamentally a
```
const remainingCurrencies = currencies.filter(c => !sortedCurrenciesMap.has(getCurrencyId(c)));
```
which may be similar in perf but simpler code ? 🤔

libs/ledger-live-common/.unimportedrc.json
```diff
@@ -303,7 +303,8 @@
     "src/index.ts",
     "src/jest.d.ts",
     "src/postOnboarding/mock.ts",
-    "src/reactNativeSvg.d.ts"
```
KVNLS, line 306: @valpinkman we need conventions on our file structure otherwise it will be hard to maintain. Something like __tests__/mock.ts in that case we will be able to exclude __tests__ pattern.

KVNLS:
Update with the latest version:
image
@gre I added a unit test covered by a snapshot. My first iteration did not cover it so I updated the algo.

libs/ledger-live-common/.unimportedrc.json
gre: we could also consider currencies/mock to be a valid entry point btw, because it is somehow (for the tests).
the problem rely more on the fact live-common is a multi entry point library.

PR #5326, opened 2023-11-06
Manage swap signature according to swap type

sprohaszka-ledger:
### 📝 Description

With the new Swap format, SwapNG expect a baseUrl signature and does not need to convert it into a DER. AppExchange is in charge of converting it.

Add a new test case for AppExchange with data coming from third party software.

### ❓ Context

- **Impacted projects**: `hw-app-exchange`
- **Linked resource(s)**: `N/A` 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [ ] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/ledger-live-common/src/exchange/platform/transfer/completeExchange.ts
+++ b/libs/ledger-live-common/src/exchange/platform/transfer/completeExchange.ts
@@ -156,13 +156,16 @@ const completeExchange = (
   });
 };
 
+/**
+ * For the Fund and Swap flow, the signature sent to the nano needs to
+ * be in DER format, which is not the case for Sell flow. Hence the
+ * ternary.
+ * cf. https://github.com/LedgerHQ/app-exchange/blob/e67848f136dc7227521791b91f608f7cd32e7da7/src/check_tx_signature.c#L14-L32
+ * @param {Buffer} bufferSignature
+ * @param {ExchangeTypes} exchangeType
+ * @return {Buffer} The correct format Buffer for AppExchange call.
+ */
 function convertSignature(bufferSignature: Buffer, exchangeType: ExchangeTypes): Buffer {
-  /**
-   * For the Fund and Swap flow, the signature sent to the nano needs to
-   * be in DER format, which is not the case for Sell flow. Hence the
-   * ternary.
-   * cf. https://github.com/LedgerHQ/app-exchange/blob/e67848f136dc7227521791b91f608f7cd32e7da7/src/check_tx_signature.c#L14-L32
-   */
   const goodSign =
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/exchange/platform/transfer/completeExchange.ts
```diff
@@ -169,4 +156,23 @@ const completeExchange = (
   });
 };
 
+function convertSignature(bufferSignature: Buffer, exchangeType: ExchangeTypes): Buffer {
+  /**
+   * For the Fund and Swap flow, the signature sent to the nano needs to
+   * be in DER format, which is not the case for Sell flow. Hence the
+   * ternary.
+   * cf. https://github.com/LedgerHQ/app-exchange/blob/e67848f136dc7227521791b91f608f7cd32e7da7/src/check_tx_signature.c#L14-L32
+   */
```
CremaFR, line 165:
```
/**
 * For the Fund and Swap flow, the signature sent to the nano needs to
 * be in DER format, which is not the case for Sell flow. Hence the
 * ternary.
 * cf. https://github.com/LedgerHQ/app-exchange/blob/e67848f136dc7227521791b91f608f7cd32e7da7/src/check_tx_signature.c#L14-L32
 */
function convertSignature(bufferSignature: Buffer, exchangeType: ExchangeTypes): Buffer {
```

could we put the comment before the function signature?

sarneijim (approved):
Changeset is missed

libs/ledger-live-common/src/exchange/swap/completeExchange.ts
```diff
@@ -229,4 +229,10 @@ const completeExchange = (
   });
 };
 
+function convertSignature(signature: string, exchangeType: ExchangeTypes): Buffer {
```
CremaFR, line 232: It seems to do the same as the other convertSignature function could we do: 
convertSignature(bufferSignature: **Buffer | string** , exchangeType: ExchangeTypes): Buffer ...
and manage the two different types in the same function? 
Or I think we could rename the function to convertSignatureFromBuffer and convertSignatureFromString

sprohaszka-ledger: Yes, it is.
I don't want to do a rework right now as we don't have automated test for those functions.
They are both local functions for those 2 `completeExchange` functions.

I'd prefer not do to much rework as it will imply testing also SELL and FUNDS.

CremaFR: understable :)

PR #5330, opened 2023-11-06
bugfix(lld): fix dark mode support of QR Code Camera Picker fallback

gre:
### 📝 Description

Previously to this PR, if you were in dark mode, you were not properly seeing the fallback error on the QR Camera picker:

Screenshot 2023-11-06 at 11 38 40

So I fixed it

Screenshot 2023-11-06 at 11 50 50

- I also took opportunity to fix the layout. It seems this mode isn't tested often (and it's not easy to test), also i didn't see a way to automate this (we can simulate a fake camera, but i don't see how to simulate a permission deny case)

- I also ensured that the error eventually appears (20s) in case of some OS that never trigger an error case (in MacOS case, I could see it instantly tho, but there is history of this problem https://stackoverflow.com/questions/42176505/getusermedia-what-if-user-makes-no-choice )

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-10043 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/.changeset/late-spiders-battle.md
+++ b/.changeset/late-spiders-battle.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Fix the QR Code Camera Picker fallback mode to properly display in dark mode context. Also add a 20s timeout in specific cases where OS would "never end" instead of failing the getUserMedia call.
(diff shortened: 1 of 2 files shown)


mcayuelas-ledger (approved):
LGTM

PR #3521, opened 2023-05-26
[support] Remove live-portfolio lib

haammar-ledger:
### 📝 Description

This is a proposal to remove the recently created `live-portfolio` lib, which contains very little code and is a bit overkill at the moment. It's scope isn't very clear either, I think it was intended for any other util that could be shared by both the portfolio section + other parts of the code, but there is no foreseeable need at the moment.

The compromise is to split the code that the lib currently contains into the different places where it's used, which causes only a little duplication. The necessary duplicate code is exactly this, which seemed acceptable to me:
```
const hourIncrement = 60 * 60 * 1000;
const dayIncrement = 24 * hourIncrement;
const weekIncrement = 7 * dayIncrement;

export function startOfHour(t: Date): Date {
  return new Date(t.getFullYear(), t.getMonth(), t.getDate(), t.getHours());
}
export function startOfDay(t: Date): Date {
  return new Date(t.getFullYear(), t.getMonth(), t.getDate());
}
export function startOfWeek(t: Date): Date {
  const d = startOfDay(t);
  return new Date(d.getTime() - d.getDay() * dayIncrement);
}
```
### ❓ Context

- **Impacted projects**: `coin-framework` `ledger-live-common` `live-portfolio` 
- **Linked resource(s)**: N/A 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

Nothing to demo.

### 🚀 Expectations to reach

Live still builds. 100% covered by CI.

PR #5331, opened 2023-11-06
[SUPPORT] update balanceHistoryCache test

chabroA:
### 📝 Description

~Update some tests that seem deprecated (for a while).
Was raised by this PR https://github.com/LedgerHQ/ledger-live/pull/4886 but since it's not directly related to this PR, the update is done independently, since it also fails on develop as of today.
Tested on the [last commit](https://github.com/LedgerHQ/ledger-live/commit/a605eb96d6781c7406784c59581cbb604896e95f) that changed these files (cf. [this PR](https://github.com/LedgerHQ/ledger-live/pull/3521)), these tests where already failing back then.~

No changeset added since only updates tests (i.e: dev content, no actual logic update)

EDIT: 
The issue was actually not coming from the tests themselves but from the way the tests are run, and more specifically how timezones are handled.

The CI uses the package.json `test` script to run tests, which manually set the timezone to `TZ=America/New_York` through `cross-env TZ=America/New_York`. (PS: why use the `America/New_York` instead of the default `UTC` timezone? Good question).

But of you run the tests with a simple `pnpm jest` (which is a valid command and should work, it's just a matter of personal preferences), the timezone is not set and thus the system timezone is used, leading to tests failing if your system timezone is not `America/New_York`.

Since the tests should be consistent, regarding of how they are run, the timezone setting has been moved as part of the jest config.

Also added a test to ensure that the timezone used is indeed the `America/New_York` once to avoid future misunderstanding.

### ❓ Context

- https://github.com/LedgerHQ/ledger-live/pull/4886

- **Impacted projects**: `` 
- **Linked resource(s)**: [] 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-framework/jest-global-setup.js
+++ b/libs/coin-framework/jest-global-setup.js
@@ -1,0 +1,9 @@
+module.exports = async () => {
+  /**
+   * Set the timezone to EST for all tests
+   * This allow us to run tests directly through jest without having to manually
+   * set the timezone as an env var beforehand, either manually or through a
+   * custom script in package.json
+   */
+  process.env.TZ = "America/New_York";
+};
--- a/libs/coin-framework/jest.config.js
+++ b/libs/coin-framework/jest.config.js
@@ -1,10 +1,7 @@
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
-  testPathIgnorePatterns: [
-    "lib/",
-    "lib-es/"
-  ]
+  preset: "ts-jest",
+  testEnvironment: "node",
+  testPathIgnorePatterns: ["lib/", "lib-es/"],
+  globalSetup: "<rootDir>/jest-global-setup.js",
 };
-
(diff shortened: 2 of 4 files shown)


haammar-ledger (approved):

gre:
do you know what broke this test before? are we still using nyc timezone for the test? i believe we're assessing somewhere around here that timezone behaves as expected.

chabroA:
> do you know what broke this test before? are we still using nyc timezone for the test? i believe we're assessing somewhere around here that timezone behaves as expected.

I updated the description and the commits. The issue was actually not coming from the tests themselves but from the way the tests are run, and more specifically how timezones are handled.

The CI uses the `package.json` `test` script to run tests, which manually set the timezone to TZ=America/New_York through cross-env TZ=America/New_York. (PS: why use the America/New_York instead of the default UTC timezone? Good question).

But if you run the tests with a simple `pnpm jest` (which is a valid command and should work, it's just a matter of personal preferences), the timezone is not set and thus the system timezone is used, leading to tests failing if your system timezone is not America/New_York.

Since the tests should be consistent, regarding of how they are run, the timezone setting has been moved as part of the jest config.

Also added a test to ensure that the timezone used is indeed the America/New_York once to avoid future misunderstanding.

gre:
> why use the America/New_York instead of the default UTC timezone? Good question

because when you use UTC, you may not reveal certain bug. for instance, the date to consider for a NYC time of 10PM is not the local YYYYMMDD date but the day after.

so if a logic is accidentally using the local date instead of the utc one, we can get into issue. this was specifically important for the caching of balance history where we had bad cache issue like this. for instance if you ask the backend what is the current rate for a future date (in the other case where you live in the future)

PR #5333, opened 2023-11-06
fix(common/listappsv2): don't use currencyId if no matching currency

ofreyssinet-ledger:
### 📝 Description

In the "polyfilling" logic of apps data in list apps v2, we were not checking that the `currencyId` coming from the backend was matching a currency locally. This was causing a crash at rendering:
image (6)

The fix is to only keep the `currencyId` coming from the backend if it actually matches a known currency.

### ❓ Context

- **Impacted projects**: `common` `llm` `lld` 
- **Linked resource(s)**: [LIVE-10067] 

### ✅ Checklist

- [] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

[LIVE-10067]: https://ledgerhq.atlassian.net/browse/LIVE-10067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/apps/polyfill.ts
+++ b/libs/ledger-live-common/src/apps/polyfill.ts
@@ -153,7 +153,7 @@ export const mapApplicationV2ToApp = ({
   indexOfMarketCap: -1, // We don't know at this point.
   type: name === "Exchange" ? AppType.swap : type,
   ...rest,
-  currencyId: currencyId || getCurrencyIdFromAppName(name),
+  currencyId: findCryptoCurrencyById(currencyId) ? currencyId : getCurrencyIdFromAppName(name),
   compatibleWallets: parseCompatibleWallets(compatibleWallets, name),
 });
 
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/src/apps/polyfill.ts
```diff
@@ -153,7 +153,7 @@ export const mapApplicationV2ToApp = ({
   indexOfMarketCap: -1, // We don't know at this point.
   type: name === "Exchange" ? AppType.swap : type,
   ...rest,
-  currencyId: currencyId || getCurrencyIdFromAppName(name),
+  currencyId: findCryptoCurrencyById(currencyId) ? currencyId : getCurrencyIdFromAppName(name),
```
aussedatlo, line 156: Nice, thx 👍

aussedatlo (approved):

PR #5350, opened 2023-11-07
[LIVE-8896] [LLM][ESC] Infinite loading when power off Stax during ESC pack language installation

jdabbech-ledger:
### 📝 Description

Fix UI issue on LLM if user shutdown device during language pack install
Whitelist errors that retry polling

### ❓ Context

- **Impacted projects**: `LLM, LL-COMMON` 
- **Linked resource(s)**: [LIVE-8896] 

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 📸 Demo

https://github.com/LedgerHQ/ledger-live/assets/145363160/a6a30497-4bf6-4b97-adf5-a5e9053955ca

https://github.com/LedgerHQ/ledger-live/assets/145363160/1283864f-5030-4c71-b53f-17a54bccd424

### 🚀 Expectations to reach


libs/ledger-live-common/src/hw/actions/implementations.ts
```diff
@@ -147,17 +148,16 @@ const pollingImplementation: Implementation = <SpecificType, GenericRequestType>
 
             debounce((event: EmittedEvent<SpecificType> | ImplementationEvent) => {
               if (event.type === "error" && "error" in event) {
+                const {
+                  retryErrorsWhitelist = [DisconnectedDevice, DisconnectedDeviceDuringOperation],
+                } = params;
```
alexandremgo, line 153: Maybe can you add a comment on top of `pollingImplementation` to explain what `retryErrorsWhiteList` does and that it has a default value pls ? (don't explain all the other fields of `params` no worries - just the one you added)

And i feel people try to avoid saying white/black list now. Maybe another name can be `retryableDisconnectedErrors` or `retryableWithDelayDisconnectedErrors` (because also those errors are being delayed with a `reconnectWaitTime` so they are kind of associated to disconnection errors)

gre: the code is quite hard to read, i initially thought it was a params deconstruction. 

I would suggest you create a `defaultErrors` somewhere.

also, generally speaking i think it would be more readable to do

```js
const errors = params.retryableWithDelayDisconnectedErrors || defaultErrors
```

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -429,15 +429,6 @@ const cmd = ({ deviceId, request }: Input): Observable<ConnectAppEvent> => {
               }
             }),
             catchError((e: unknown) => {
-              if (
-                e instanceof DisconnectedDeviceDuringOperation ||
-                e instanceof DisconnectedDevice
```
alexandremgo, line 434: You can remove the import of DisconnectedDeviceDuringOperation and DisconnectedDevice ;) 

Though by doing this it will have an impact on some other device actions (not just installLanguage) that might rely on this `disconnected` event. Are we sure we need to remove this check ?

libs/ledger-live-common/src/hw/actions/installLanguage.ts
```diff
@@ -12,6 +12,7 @@ import type {
 import type { Action, Device } from "./types";
 import { currentMode } from "./app";
 import { getImplementation } from "./implementations";
+import { DisconnectedDeviceDuringOperation } from "@ledgerhq/errors";
```
alexandremgo, line 15: This import is not used (is it normal) ?

libs/ledger-live-common/src/hw/connectApp.ts
jdabbech-ledger: It was done in order to fix a QA review on a different case that LIVE-8896 specified. Wouldn't it better to create another task for those commits @aussedatlo ?

aussedatlo: Indeed, it could be dangerous.

Let me remove this part, like @jdabbech-ledger suggest, it's better to fix this case in another ticket

alexandremgo (approved):
LGTM 👌

PR #5364, opened 2023-11-08
[LIVE-9998] Bugfix - Change `getSyncHash` hashing algorithm

lambertkevin:
### 📝 Description

Changing the hash algorithm used for creating the `syncHash`. We were using a cryptographic algorithm (sha256) provided by the `ethers` library, which is very efficient on a desktop computer, but is very poorly optimised for a React Native environment. 

This function will now use a [MurmurHash](https://en.wikipedia.org/wiki/MurmurHash) algorithm which is a non cryptographic hash resulting in a 32 bits unsigned integer, which clearly has a collision drawback, but this `getSyncHash` method isn't vital for the coin module to work correctly. 

If a collision was to exist after a modification of the CAL, it would only prevent a new full synchronization of the account, making it impossible to display the updated tokens if they were to be owned by the user. A clear cache would then be enough to make everything work correctly again.

For 7k tokens on a Galaxy S9:
- Generation of the long string of token properties: ~15ms
- Hashing of the long string using SHA256: ~10_000ms
- Hashing the long string using MurmurHashV3: ~150ms

Cache has also been added to act as a memoization in order to hash only when the long string changed, resulting in just ~15ms of response time for the `getSyncHash` method once the hash cached.

### ❓ Context

- **Linked resource(s)**: https://ledgerhq.atlassian.net/browse/LIVE-9998 

### ✅ Checklist

- [x] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

### DIFF
--- a/libs/coin-evm/package.json
+++ b/libs/coin-evm/package.json
@@ -63,12 +63,14 @@
     "eip55": "^2.1.1",
     "ethers": "5.7.2",
     "expect": "^27.4.6",
+    "imurmurhash": "^0.1.4",
     "invariant": "^2.2.2",
     "lodash": "^4.17.21",
     "react-native-get-random-values": "^1.8.0",
     "rxjs": "^7.8.1"
   },
   "devDependencies": {
+    "@types/imurmurhash": "^0.1.4",
     "@types/invariant": "^2.2.2",
     "@types/jest": "^29.2.4",
     "@types/lodash": "^4.14.191",
--- a/libs/coin-evm/src/__tests__/unit/logic.unit.test.ts
+++ b/libs/coin-evm/src/__tests__/unit/logic.unit.test.ts
@@ -5,6 +5,7 @@ import { CryptoCurrency, TokenCurrency, Unit } from "@ledgerhq/types-cryptoasset
 import BigNumber from "bignumber.js";
 import * as RPC_API from "../../api/node/rpc.common";
 import {
+  __testOnlyClearSyncHashMemoize,
   attachOperations,
   eip1559TransactionHasFees,
   getAdditionalLayer2Fees,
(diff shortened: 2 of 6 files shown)


libs/coin-evm/package.json
```diff
@@ -63,12 +63,14 @@
     "eip55": "^2.1.1",
     "ethers": "5.7.2",
     "expect": "^27.4.6",
+    "imurmurhash": "^0.1.4",
```
lvndry, line 66: The last commit on this lib was 10+ years ago and the repo has very few stars, don't you think this would be a "better" alternative: https://github.com/karanlyons/murmurHash3.js

lambertkevin: Nope, this is the most used version and more specifically the fastest implementation by far. I already benched the murmurhash libs, and this one should be the best option 👍

lambertkevin: (In this _almost_ cryptography world, implementations are pretty static once built & working. You have a spec, you follow it to get your implementation, then you probably won't touch it except for an important bugfix)

lambertkevin: For example here, all the libs you'll find are just rewritings of:
- https://github.com/garycourt/murmurhash-js
- https://github.com/kazuyukitanimura/murmurhash-js (which is a fork of the gaycourt repo)

And both are 12 years old repo, but fully functional.

gre:
btw even for LLD it's preferrable to defer any "sync" computation to a web worker like we did for heavy computation of bitcoin's segcp256k1 , maybe out of scope of this PR but if we identify more cases of heavy computaiton done in sync / on the renderer process, feel free to discuss with me / @haammar-ledger  / @hzheng-ledger  about the way we did the web worker, but basically we have a system where you can create web workers easily and then it's about introducing a hook.

libs/coin-evm/src/__tests__/unit/logic.unit.test.ts
```diff
@@ -398,8 +399,33 @@ describe("EVM Family", () => {
         setEnv("NFT_CURRENCIES", oldEnv);
       });
 
-      it("should provide a valid sha256 hash", () => {
-        expect(getSyncHash(currency)).toStrictEqual(expect.stringMatching(/^0x[A-Fa-f0-9]{64}$/));
+      beforeEach(() => {
+        // Clearing the cache of already calculated hashes before each new test
+        Object.keys(_simpleSyncHashMemoize).forEach(key => delete _simpleSyncHashMemoize[key]);
```
haammar-ledger, line 404: IIUC the need to export `_simpleSyncHashMemoize` is just to be able to clear it here, right?

Unimportant but I'd rather export a function to clear it, rather then the variable itself. So it doesn't need the `_` prefix that triggers me :stuck_out_tongue:

haammar-ledger (approved):

hedi-edelbloute (approved):

PR #5443, opened 2023-11-14
Ledger Live don't import tokens if parentAccount is not supported

lvndry:
### 📝 Description

We sometimes removes parent currencies in Ledger Live but not in CAL. This could raise an error in `tokens.ts` if the parentCurrency is present in a json file but not supported in Ledger Live. This PR handles this case.

### ❓ Context

- **JIRA or GitHub link**: `https://ledgerhq.atlassian.net/browse/LIVE-10198`

### ✅ Checklist

  - 

---

### DIFF
--- a/apps/ledger-live-mobile/src/screens/OperationDetails/Content.tsx
+++ b/apps/ledger-live-mobile/src/screens/OperationDetails/Content.tsx
@@ -144,7 +144,7 @@ export default function Content({
   const { EditOperationPanel: SpecificEditOperationPanel = undefined } =
     byFamiliesEditOperationPanel[
       mainAccount.currency.family as keyof typeof byFamiliesEditOperationPanel
-    ];
+    ] || {};
 
   const urlFeesInfo =
     specificOperationDetails &&


libs/ledgerjs/packages/cryptoassets/src/currencies.ts
```diff
@@ -4597,8 +4597,8 @@ export const hasCryptoCurrencyId = (id: string): boolean => id in cryptocurrenci
  *
  * @param {*} id
  */
-export function getCryptoCurrencyById(id: string | undefined): CryptoCurrency {
-  const currency = id && findCryptoCurrencyById(id);
```
lambertkevin, line 4601: I'm unsure of why we had this `undefined` option for `id`, and that change makes me worried about a nonsensical problem down the road somehow 😆

lambertkevin (approved):
LGTM with a small attention notice on the (quite logical) change made to `getCryptoCurrencyById` 👀 👍

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
lvndry: If ever were in the case where `id` is undefined then `currency` would be undefined and then it would throw an error which didn't happened AFAIK

libs/ledgerjs/packages/cryptoassets/src/tokens.ts
```diff
@@ -241,8 +247,13 @@ export function convertERC20([
   contractAddress,
   disableCountervalue,
   delisted,
-]: ERC20Token | PolygonERC20Token): TokenCurrency {
-  const parentCurrency = getCryptoCurrencyById(parentCurrencyId);
+]: ERC20Token | PolygonERC20Token): TokenCurrency | undefined {
+  const parentCurrency = findCryptoCurrencyById(parentCurrencyId);
```
haammar-ledger, line 251: Another way of doing this, a bit more concise and doesn't need to change the `convertXXXTokens` and `addTokens` functions:
```diff
-addTokens(erc20tokens.map(convertERC20));
+addTokens(erc20tokens.filter(t => !!findCryptoCurrencyById(t[0])).map(convertERC20));
```

haammar-ledger (commented):
Needs a changeset :slightly_smiling_face:

libs/ledgerjs/packages/cryptoassets/src/tokens.ts
lambertkevin: Like I was saying to @lvndry earlier, this would make 2 loops instead of 1 on a fairly big (7k items today) array, and it would modify our "source of truth" by not really dealing with whatever file we have locally (which doesn't really makes sense). 

I would advise against this

haammar-ledger (approved):

libs/ledgerjs/packages/cryptoassets/src/tokens.ts
haammar-ledger: I (briefly) though about perf, this would cause 1 more instruction for each processed token, so about ~30000 reads for the sake of code readability. I don't think that should be significant event on mobile?

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
haammar-ledger: This  code change actually has zero functional impact, since `findCryptoCurrencyById` will just return undefined if `id` is actually undefined, and the behaviour of `getCryptoCurrencyById` will be just the same. So should be all good :ok_hand:

libs/ledgerjs/packages/cryptoassets/src/tokens.ts
```diff
@@ -21,23 +21,38 @@ const tokensByTicker: Record<string, TokenCurrency> = {};
 const tokensByAddress: Record<string, TokenCurrency> = {};
 const tokensByCurrencyAddress: Record<string, TokenCurrency> = {};
 const tokenListHashes = new Set();
-addTokens(erc20tokens.map(convertERC20));
+
+addTokens(
+  erc20tokens
+    .filter(([parentCurrencyId]) => !!findCryptoCurrencyById(parentCurrencyId))
+    .map(convertERC20),
+);
```
chabroA, line 29: I think it was already raised by @lambertkevin [here](https://github.com/LedgerHQ/ledger-live/pull/5443#discussion_r1395590634), but traversing 2 times a penitentially large array could bring some performance concerns.

Here is a suggestion combining the `filter` and `map` operations into a single `reduce` operation to iterate over the array only once, thus optimizing the time complexity.

```suggestion
addTokens(
  erc20tokens.reduce((acc: TokenCurrency[], token: ERC20Token) => {
    const [parentCurrencyId] = token;
    const currency = findCryptoCurrencyById(parentCurrencyId);
    if (currency) {
      acc.push(convertERC20(token));
    }
    return acc;
  }, []),
);
```

chabroA (changes requested):

libs/ledgerjs/packages/cryptoassets/src/tokens.ts
chabroA: Alternatively, we could sanitize the token list right at the import step in `libs/ledgerjs/packages/cryptoassets/src/data/erc20.ts` and filter out tokens without a parent currency right there at the source.

If having a token without an associated parent currency available in the Live can cause issues, it could be safer to treat it as close as possible to the source.

lambertkevin: Indeed you can do that with only 1 loop if it was necessary, but like I was saying on Slack: My proposed change has a 0 cost/impact contrary to anything we'll do here. 
I initially proposed to simply change the `getCryptoCurrencyById` in `convertERC20` which would throw for a `findCryptoCurrencyById` with an an early return instead, which is already handled by `addTokens`.

Surprise suprise: `getCryptoCurrencyById` is litteraly calling `findCryptoCurrencyById` and throwing if the result is undefined. So exactly the same result, no unecessary changes here, and no changes of the `erc20.json` file in anyway.

PR #5496, opened 2023-11-17
[LIVE-9888] [LIVE-10041] set empty token account id to the parentid

Wozacosta:
### 📝 Description

`makeEmptyTokenAccount` (in libs/coin-framework/src/account/helpers.ts) is called when we need to create temporary accounts for them to be displayed and chosen from. 
For instance, here we are triggering the `makeEmptyTokenAccount` function to display two USDT accounts: 

https://github.com/LedgerHQ/ledger-live/assets/5050709/56a8171e-d9bb-4caf-ba2e-47038ba4da99

Those token have a parent account (USDT parent is ethereum). 
The problem is, that function creates an id that doesn't match any existing account, and subsequent wallet api calls to `account.receive` with that ID expectedly fail. 
Assigning to the empty token account id its parent ID value solves that problem.

### ❓ Context

- **JIRA or GitHub link**: [JIRA-9888](https://ledgerhq.atlassian.net/browse/LIVE-9888) [JIRA-10041](https://ledgerhq.atlassian.net/browse/LIVE-10041)

### ✅ Checklist

---

### DIFF
--- a/.changeset/brave-turtles-begin.md
+++ b/.changeset/brave-turtles-begin.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-framework": patch
+---
+
+empty token accounts ids set from their parent token account id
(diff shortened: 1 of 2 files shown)


gre (changes requested):
mmh, the previous code seems correct tho. a TokenAccount do not have the parent's Account's .id , however, what seems an issue is more the fact we don't even use the `encodeTokenAccountId` here which can lead to issues

https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-framework/src/account/accountId.ts#L50

PR #5524, opened 2023-11-20
[SUPPORT] - Adding logs for invalid dates

lambertkevin:
### 📝 Description

Adding logs to better understand the [RangeError](https://ledger.sentry.io/issues/4616952187/?project=6488660) some user get but impossible to reproduce.

### ❓ Context

- **JIRA or GitHub link**: 

### ✅ Checklist

- [ ] **Covered by automatic tests.** // nothing to test
  - no impact

---


apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -114,6 +115,21 @@ export class OperationsList extends PureComponent<Props, State> {
...
+        if (day instanceof Date && !isNaN(day as unknown as number)) {
+          log("Ethereum Date Error", "Date in operation is invalid", {
+            day,
+            accountId: account?.id,
+            groupOps: group.data,
+          });
+        }
+      }
+    }
+    // -- THIS CAN BE REMOVED ONCE THE DATE ERROR HAS BEEN FIGURED OUT
```
gre, line 131: i think it is preferable to do this in the logic side. (groupAccountsOperationsByDay function) so you don't have to do a loop again but just contain the temporary code inside the place it's generating the groups.

btw the date comes from `bestOp.date` to it means one of the coin impl produces Date that are NaN. 

I have seen this issue in the past (on sentry i think) and this is happening whenever `new Date(undefined)` happens, so indeed there is lot of place to this occurence, but sounds like a backend not returning a date where we expected it. it sounds a good idea to ignore the operation in that case because it can produces pretty ugly error on the end UI as the date is today assumed to always be settled.

gre (dismissed):
on top of my feedback, can you tell how much logs it will create, assuming you have N accounts and M operations?
if it's > N*M, i think it is too heavy to get these in since we have a logs size limit and it will be too much logs , so we may want some of these logs to only enable `if (condition)` , where condition is whatever logic feature detecting the bug (date NaN?)

gre:
on top of logging the issue, i think we must filter out invalid Operation as early as we can, maybe in mergeOperation or whatever generic util function we have. because an Operation with a date: Date that is invalid is going to nuke in many places

haammar-ledger (approved):

PR #5546, opened 2023-11-22
use safeEncode for etherscan adapter

lvndry:
### 📝 Description

`from` and `to` fields were not safely encoded in the etherscan adapter which caused an error to be thrown is some cases (e.g if the to field was empty).

This PR adds:

- Usage of `safeEncode` for `from` and `to` addresses in etherscan adapter
- Return the address if can't be safely encoded

### ❓ Context

- **JIRA or GitHub link**: `https://ledgerhq.atlassian.net/browse/LIVE-10027`

### ✅ Checklist

  - 

---


libs/coin-evm/src/logic.ts
```diff
@@ -22,6 +22,8 @@ import {
   EvmTransactionEIP1559,
   EvmTransactionLegacy,
 } from "./types";
+import eip55 from "eip55";
+import { log } from "@ledgerhq/logs";
```
lambertkevin, line 26: If possible, external imports should always be at the top compared to local imports

lambertkevin (approved):

libs/coin-evm/src/logic.ts
chabroA: Protip, in VS code you can organize imports with the **⇧⌥O** keyboard shortcut 👌 

cf. https://code.visualstudio.com/docs/languages/javascript#_organize-imports

libs/coin-evm/src/logic.ts
```diff
@@ -346,3 +348,37 @@ export const getMessageProperties = async (
...
+
+  try {
+    return eip55.encode(addr);
+  } catch (e) {
+    log("EVM Family - logic.ts", "Failed to eip55 encode address", {
+      address: addr,
+      error: e,
+    });
+
+    return addr;
```
chabroA, line 382: You changed the return value when the `eip55.encode` function failed.
It originally returned an empty string and is now replaced with the original (non-encoded) address.

>  we don't want the app to crash in these cases, we simply ignore the address and return an empty string.

Not sure we want to have badly encoded addresses flowing through the app (that's why we encode them on the edges) to avoid unexpected errors later down the road if / when these addresses are used in the codebase.

Also, could be worthwhile to add unit tests to this helper function to prevent (or at least more easily raise) these kind of changes in the future

libs/coin-evm/src/__tests__/unit/adapters/etherscan.unit.test.ts
```diff
@@ -19,6 +19,63 @@ describe("EVM Family", () => {
...
+            date: new Date("2022-06-08T00:02:50.000Z"),
+            transactionSequenceNumber: 7,
+            hasFailed: false,
+            nftOperations: [],
+            subOperations: [],
+            type: "FEES",
+            extra: {},
+          };
+
+          expect(etherscanOperationToOperations(accountId, etherscanOp)).toEqual([
```
chabroA, line 74: Not blocking but `toMatchInlineSnapshot` could be used here rather than manually creating and defining the whole `expectedOperation` object (yes I know we have this manually defined expected object pattern king of everywhere in the codebase, is it optimal? I don't know)
cf. https://jestjs.io/docs/snapshot-testing#inline-snapshots and https://jestjs.io/docs/expect#tomatchinlinesnapshotpropertymatchers-inlinesnapshot

libs/coin-evm/src/__tests__/unit/adapters/etherscan.unit.test.ts
```diff
@@ -19,6 +19,63 @@ describe("EVM Family", () => {
   describe("adapters", () => {
     describe("etherscan", () => {
       describe("etherscanOperationToOperations", () => {
+        it("should convert an etherscan-like smart contract creation operation (from their API) to a Ledger Live Operation", () => {
```
chabroA, line 22: Could be interesting to test as well the case where the address is badly formatted **but** does not fall in this condition: `if (!addr || addr === "0x" || addr === "0x0")`

chabroA (changes requested):

libs/coin-evm/src/logic.ts
lvndry: This should be a eslint rule then...

https://github.com/import-js/eslint-plugin-import

libs/coin-evm/src/logic.ts
lvndry: @lambertkevin do you valid ?

libs/coin-evm/src/__tests__/unit/adapters/etherscan.unit.test.ts
lvndry: Personally I'm not fan of snapshots, you end up updating them everytime you make a change
It adds a lot of line but yeah it's a personal preference

libs/coin-evm/src/logic.ts
lambertkevin: To me there is no reason the return "" either. As a matter of fact, in those specific places "to" and "from" should *never* reach the catch. Those addresses are coming from an explorer, so the addresses can't be wrong or malformed, they're all valid Ethereum addresses. So if we had to pick a behaviour that would be to actually indeed throw if that happened as something definitely extremly wrong somewhere.

Returning nothing when you had an input is an extremly weird behaviour to me, and I'm already a bit stressed by the implication of not setting anything but an empty string in a `recipient` value of an `Operation` due to a contract deployment, which could lead to errors later on in the UI by simply inferring that "recipient *has* to be an hexadecimal string of 20 bytes" 😕 

I would personally prefer still displaying the info we got as input rather that completely removing the data if ever we had this weird error case here.

libs/coin-evm/src/logic.ts
lambertkevin: > cf. https://code.visualstudio.com/docs/languages/javascript#_organize-imports

You're going to ruin my pyramids again 😭

libs/coin-evm/src/__tests__/unit/adapters/etherscan.unit.test.ts
lvndry: https://github.com/LedgerHQ/ledger-live/pull/5546/commits/15f50b21c5fb2b77a1fa692d4158ab5951f05b9b#diff-47974acb4fd1dfce54213ad9fc7cef06165b4e2463628370e022528b8223547d

libs/coin-evm/src/__tests__/unit/adapters/etherscan.unit.test.ts
lambertkevin: All solutions have pros & cons. 
I agree that snapshots are definitely not an amazing solution, it's faster but it's not in the test so you just never read it. 
Fixtures are little bit better as they're easier to access, and we have a lot of those already, but you still don't clearly see what you're testing right away as most of it is generated for you and generally contains a **lot** of fake data in it.

Clearly for a test for an adapter like this where you want to test that A — will become —> B, I think verbose is the best option. You want to see exactly what *should* be done directly, you're safe from any fake data or mock, and you're still as safe as if you were using fixtures since the typing is still added with TS.

libs/coin-evm/src/logic.ts
lvndry: https://github.com/LedgerHQ/ledger-live/pull/5546/commits/c7b9f6a3a8a5ed2a446eb7f18f8fdeb35c6006fc#diff-58968d2e02e9914b57f440daed3d8112e1b5becc68088e9458acee6eddc50674

libs/coin-evm/src/__tests__/unit/adapters/etherscan.unit.test.ts
chabroA: > I agree that snapshots are definitely not an amazing solution, it's faster but it's not in the test so you just never read it.

I agree with that and that is why I'm talking about **inline snapshots** (rather than the _normal_ snapshot we already use in some tests, which end up in a different file no one ever looks at).

I think inline snapshot provides the best of both worlds by writing the expected result inside the specific test case *and* being overall easier to write (generate) and maintain. Worth giving it a shot if you haven't played with it yet.

But at the end of the day, it's more or less the same and the result is the same (do we have a valid test that checks what we want). It's more a matter of personal preferences and ease of creating and maintenance.

libs/coin-evm/src/logic.ts
chabroA: > As a matter of fact, in those specific places "to" and "from" should never reach the catch. Those addresses are coming from an explorer, so the addresses can't be wrong or malformed, they're all valid Ethereum addresses. So if we had to pick a behaviour that would be to actually indeed throw if that happened as something definitely extremly wrong somewhere.

I agree with that, it's never supposed to happen in the _nominal_ case, but since it's data we can't control (coming from the outside of the app), it needs to be _sanityzed_ in some way to avoid issues in the app, like the original issue related to non EIP55 address returned by the backend.

But if the data is only used for display purposes in the app (possible but not sure), then we can let the original value through as is. But in this case, what's the point of encoding it to EIP 55 format in the first place then? 🤔

libs/coin-evm/src/__tests__/unit/logic.unit.test.ts
```diff
@@ -681,5 +683,37 @@ describe("EVM Family", () => {
         expect(padHexString("01")).toEqual("01");
       });
     });
+
+    describe("safeEncodeEIP55", () => {
+      it("Should return encoded address if valid address", () => {
+        const address = "0x9aa99c23f67c81701c772b106b4f83f6e858dd2e";
+        const encodedAddress = safeEncodeEIP55(address);
+        expect(encodedAddress).toBe(eip55.encode(address));
```
chabroA, line 691: By using `eip55.encode`, you are using and relying on implementation details of the function you are actually testing. I would rather write directly the expected output.
This makes the test more resilient (and actually agnostic) to implementation updates of the underlying function.

```suggestion
        expect(encodedAddress).toBe("0x9AA99C23F67c81701C772B106b4F83f6e858dd2E");
```

chabroA (commented):
Still wondering about this, but no strong opinion (as long as it does not break stuff in the app...) 🤷‍♂️ 

https://github.com/LedgerHQ/ledger-live/pull/5546#discussion_r1403413016

libs/coin-evm/src/logic.ts
lambertkevin: > it needs to be sanityzed
I don't think it's needed, our role is not to change whatever is providing by a backend that we can't verify or update with a real value. And even in the places where we are actively using the EIP-55 checksum (e.g. when you're inputting an address yourself), we only use is as a "warning" and let the user decide if they wanna continue.

> what's the point of encoding it to EIP 55 format in the first place then? 🤔
Specifically for visual help. Here when we're encoding those strings, it can help detect an address poisining attack for example where someone will try to find a vanity address looking like another one. The end result is that by doing this checksum, you can visually detect 2 address being different even with only the first & last few caracters. 
In general, you should always have an EIP-55 encoded address displayed in the live. And like I said, here the only case where this encoding doesn't work, is when you're simply giving something else than a 20 bytes hexadecimal string, and we're now covering those cases. The rest of the try/catch is technically useless.

lvndry: In the case where the address is `0x0` or `0x` we should still return an empty string or the address as is then ?

lambertkevin: `""`, `0x`, it's a no brainer, yes should return an empty string.

For `0x0` is a bit different as it's not representing an empty buffer here, it's representing "0". And it might reprensent the `0x0000000000000000000000000000000000000000` burn address as in fine it's the same value. 
`0x0` == `0x0000000000000000000000000000000000000000` so that's the only case I'm not really sure. 
I think I would not add the `0x0` condition as an early return and let the try/catch do its work, unless @chabroA can confirm that this case is indeed happening on the explorers for a specific transaction like a smart contract deployment 🤷‍♂️

libs/coin-evm/src/logic.ts
lambertkevin: @lvndry Please don't mixup the 2 subjects. This is a completely unrelated change to the problem you're actually fixing, and should be in a different PR where we can all discuss if the whole team wants that change or not 🙏

libs/coin-evm/src/logic.ts
chabroA: Yes, explorer returns `0x0` in some cases, cf. comment (that's one reason that led me to create this function in the first place).
>  For example the explorers may return "0x0" as an address (for example for some events or contract interactions, like a contract creation transaction)

In this case, we could add a specific case for `0x0` and have the function return `0x0000000000000000000000000000000000000000` (EIP-55 encoded equivalent). Because if I remember correctly, `eip55.encode("0x0")` throws (that's why I added it in the `if` clause in the first place).

But then what do we do in the catch? If we receive an invalid address that can't be EIP-55 encoded? Do we return an empty string? Do we return the raw string address argument as is (that might not actually be an address, we don't know what it is exactly since it comes from outside the app, like from an API call)?

And if we return the raw string address argument as is, then what is the point of this function in the first place? 🤔

lambertkevin:
Please see my comment: https://github.com/LedgerHQ/ledger-live/pull/5546#discussion_r1404176002

PR #5675, opened 2023-12-06
[FEATURE][LIVE-10220] make Qtum work with old and new nano app interfaces

chabroA:
### 📝 Description

- 9669119dbbac40e8c82532c9308da8babdb43fa8 makes qtum work with old and new nano app interfaces
- dd9dba41f598d1f8813f8d23d57a24b3f8087f82 adds some code improvements:
	- use switch instead of if else cascade
	- use [semver](https://www.npmjs.com/package/semver) for nano apps versions related tests

-----

All Bitcoin altcoins use the `BtcOld` implementation of the [`hw-app-btc` package](https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/hw-app-btc) today, including Qtum. But Qtum is being updated to use the [app-bitcoin-new](https://github.com/LedgerHQ/app-bitcoin-new) and should therefore now use the `BtcNew` implementation.

This will imply a version check in the codebase to make sure the correct version of the implementation is used depending on the app version.

The current version of the Qtum app is 2.1.0, and the new version currently on P4 is 2.1.6.
The new Qtum nano app migrating from the old implem to the new one (cf. using the [app-bitcoin-new repo](https://github.com/LedgerHQ/app-bitcoin-new)) is a breaking change.
To properly follow [semver](https://semver.org/) the major version of the app should be updated and the new Qtum app version should therefore be 3.X.X

But making the Qtum nano app use the `BtcNew` implementation of `hw-app-btc` results in a new error:
`non-standard path: 44'/88'/0'/0/0"`

From what I see, this error originates from the `getWalletPublicKey` here:
https://github.com/LedgerHQ/ledger-live/blob/01fdeb3b75cba3d9a4ce0a40c9cc749cfa67ce8a/libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts#L111-L113

It looks like as of today, the `BtcNew` implementation only allows a standard path. Probably because it was originally done only for Bitcoin.

However, after removing this normal path check, I’m able to properly add a Qtum account using their new nano app version currently on P4.

~The question is then to know what proper update is needed on the `BtcNew` implementation and the related (maybe unintended) consequences.
Should we just get rid of this `isPathNormal` [check](https://github.com/LedgerHQ/ledger-live/blob/01fdeb3b75cba3d9a4ce0a40c9cc749cfa67ce8a/libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts#L445-L497), or should the pathCheck specifications be extended to include non-standard paths as well (and if yes, how)?~
The `isPathNormal` [function](https://github.com/LedgerHQ/ledger-live/blob/01fdeb3b75cba3d9a4ce0a40c9cc749cfa67ce8a/libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts#L445-L497) has been updated to allow Qtum coin type (88)

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-10220

### ✅ Checklist

  - One can now add qtum accounts, and use the currency normally, with app version implemented using the legacy bitcoin app framework (version `2.1.0` of the app) and the version using the new bitcoin app framework (version currently available on P4, with version number ~either `2.1.6` or~ `3.x.x`)

---

### DIFF
--- a/libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
+++ b/libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
@@ -1,3 +1,4 @@
+import semver from "semver";
 import type Transport from "@ledgerhq/hw-transport";
 import BtcNew from "./BtcNew";
 import BtcOld from "./BtcOld";
@@ -49,9 +50,10 @@ export default class Btc {
       ],
       scrambleKey,
     );
-    // new APDU (nano app API) for bitcoin and old APDU for altcoin
-    if (currency === "bitcoin" || currency === "bitcoin_testnet") {
+    // new APDU (nano app API) for currencies using app-bitcoin-new implementation
+    if (currency === "bitcoin" || currency === "bitcoin_testnet" || currency === "qtum") {
       this._impl = new BtcNew(new AppClient(this._transport));
+      // old APDU (legacy API) for currencies using legacy bitcoin app implementation
     } else {
       this._impl = new BtcOld(this._transport);
     }
@@ -281,6 +283,9 @@ export default class Btc {
       // therefore, we use a workaround to distinguish legacy and new versions.
       // This can be removed once Ledger Live enforces minimum bitcoin version >= 2.1.0.
       isBtcLegacy = await checkIsBtcLegacy(this._transport);
+    } else if (appAndVersion.name === "Qtum") {
+      // we use the legacy protocol for versions below 2.1.6 of the Qtum app.
+      isBtcLegacy = semver.lt(appAndVersion.version, "2.1.6");
     }
 
     if (isBtcLegacy) {
--- a/libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
+++ b/libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
@@ -50,13 +50,19 @@ export default class Btc {
       ],
       scrambleKey,
     );
-    // new APDU (nano app API) for currencies using app-bitcoin-new implementation
-    if (currency === "bitcoin" || currency === "bitcoin_testnet" || currency === "qtum") {
-      this._impl = new BtcNew(new AppClient(this._transport));
-      // old APDU (legacy API) for currencies using legacy bitcoin app implementation
-    } else {
-      this._impl = new BtcOld(this._transport);
-    }
+
+    this._impl = (() => {
+      switch (currency) {
+        case "bitcoin":
+        case "bitcoin_testnet":
+        case "qtum":
+          // new APDU (nano app API) for currencies using app-bitcoin-new implementation
+          return new BtcNew(new AppClient(this._transport));
+        default:
+          // old APDU (legacy API) for currencies using legacy bitcoin app implementation
+          return new BtcOld(this._transport);
+      }
+    })();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
```diff
@@ -263,25 +271,31 @@ export default class Btc {
...
-      // therefore, we use a workaround to distinguish legacy and new versions.
-      // This can be removed once Ledger Live enforces minimum bitcoin version >= 2.1.0.
-      isBtcLegacy = await checkIsBtcLegacy(this._transport);
-    }
+    const isBtcLegacy = await (async () => {
+      switch (name) {
+        case "Bitcoin":
+        case "Bitcoin Test": {
+          // we use the legacy protocol for versions below 2.1.0 of the Bitcoin app.
+          return semver.lt(version, "2.1.0");
```
haammar-ledger, line 281: Are you 100% sure this does the same as the current implem?

Just being paranoid, because app versions can be different from a clean `major.minor.patch`, e.g they can have `-rc0` `-next0` or such additional stuff at the end, and it happened in the past that semver doesn't behave as we think it does.

libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts
```diff
@@ -474,7 +474,7 @@ function isPathNormal(path: string): boolean {
     pathElems.length >= 3 &&
     pathElems.length <= 5 &&
     [44 + h, 49 + h, 84 + h, 86 + h].some(v => v == pathElems[0]) &&
-    [0 + h, 1 + h].some(v => v == pathElems[1]) &&
+    [0 + h, 1 + h, 88 + h].some(v => v == pathElems[1]) &&
```
haammar-ledger, line 477: I suggest putting the supported coin IDs (0, 1, 88) in a dedicated static array with an explicit name, and use this here, to make it slightly easier to understand for future maintainers.

Not saying the rest of `hw-app-btc` is perfectly clear and maintainable... but let's do what we can. :slightly_smiling_face:

haammar-ledger (commented):
+ changeset :wink:

libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts
haammar-ledger: Additional suggestion: this array could already contain the [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) coin id of all the bitcoin family chains we support today, so that we don't have to change it again next time.

libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
hzheng-ledger: I think the code that he writes is good. There was a bug of the semver lib long time ago and it has been fixed. @haammar-ledger if you have a concrete scenario, we can test it

hzheng-ledger (approved):
looks good to me

libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts
haammar-ledger: Wow you're right :open_mouth: I didn't check Qtum in SLIP-44 but it's not 88 indeed...
Ok so forget about my 2nd suggestion, but let's create a const array with the proper values :+1:

libs/ledgerjs/packages/hw-app-btc/src/Btc.ts
chabroA: Seems fixed indeed. @haammar-ledger I added these tests just in case 3849cfc44119f04945f0aee5d51e6b25b35c59dd

They are testing the semver lib implementation so should not be done in our codebase, but can be a way of protecting ourselves against unexpected behaviour from semver on these cases (cf. comment)

Feel free to remove this commit if you feel it has nothing to do in our codebase

libs/ledgerjs/packages/hw-app-btc/src/BtcNew.ts
chabroA: Done here 90e808dc39fc7fbf583f6bb75d76233eddd86c0f
Plus added tests on this function here aebe24ca2d012ed05a3ffb84fbd6d6cd7f60f91f before updating it to prevent issues

libs/ledgerjs/packages/hw-app-btc/tests/newops/isPathNormal.test.ts
```diff
@@ -0,0 +1,30 @@
+import { isPathNormal } from "../../src/BtcNew";
+
+describe("isPathNormal", () => {
+  describe("should return true for normal and supported paths", () => {
```
haammar-ledger, line 4: You could add Qtum in the list

haammar-ledger (approved):
:+1:

PR #5724, opened 2023-12-12
filecoin - handle eth addresses, change default derivation scheme to native filecoin wallets

emmanuelm41:
### 📝 Description
- Filecoin runs EVM runtime now. It is possible to send FIL tokens to eth accounts that are actually valid in the filecoin space. This PR adds support to send tokens to these eth addresses. 

- Method num 0 is used to move native tokens from one address to another one. This is valid for all actors on Filecoin. However, EVM contracts should always use "InvokeContract" method (num 3844450837) in order to trigger possible smart contract business logic set in place whenever an address sends tokens to the evm address. In cases where the recipient is just an evm account, the method is still valid and does the same method 0 (Send) does. However, if the evm account is a smart contract, underlying logic could be executed by the smart contract.

- We have switched default derivation scheme to match native filecoin wallets. Now, new wallets scanned by LL will iterate over the address slot instead of the account slot. LL scheme was left as a "legacy" derivation scheme as backward compatible to already scanned wallets.

- In order to be able to scan possible addresses that has empty addresses in between, we increased the `mandatoryEmptyAccountSkip` value to 5.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10718 

### ✅ Checklist

---

- [x] **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- [x] **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- [x] **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- [x] **Any new dependencies** have been justified and documented.
- [x] **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)


haammar-ledger (commented):
@emmanuelm41 Changes look good, but please consider adding test coverage:
- updating the `filecoin/bridge.integration.test.ts` file with new supported address format
- jest unit tests where relevant, in their own `*.unit.test.ts` file(s)

A changeset is required too (`pnpm changeset`)

Thanks!

haammar-ledger (approved):
Changes look OK, ready for QA when CI is green

emmanuelm41:
> Changes look OK, ready for QA when CI is green

@haammar-ledger I think the issue comes from develop branch. Can you check it please?

libs/ledger-live-common/src/families/filecoin/bridge/utils/utils.ts
haammar-ledger, line 1: What are the changes in this file doing? Please add a clear explanation in the PR's description

libs/coin-framework/src/derivation.ts
```diff
@@ -168,14 +168,20 @@ const modes = Object.freeze({
...
   },
   glif: {
-    overridesDerivation: "44'/461'/0'/0/<account>",
+    overridesDerivation: "44'/<coin_type>'/<address>'/<node>/<account>",
+    mandatoryEmptyAccountSkip: 5,
+  },
+  filecoinBIP44: {
+    overridesDerivation: "44'/<coin_type>'/<account>'/<node>/<address>",
     startsAt: 1,
     tag: "third-party",
```
haammar-ledger, line 183: Just wondering if this `"third-party"` tag still makes sense? Is it clear for Live users?

libs/coin-framework/src/derivation.ts
```diff
@@ -168,14 +168,20 @@ const modes = Object.freeze({
...
     tag: "legacy",
+    mandatoryEmptyAccountSkip: 5,
   },
   glif: {
-    overridesDerivation: "44'/461'/0'/0/<account>",
+    overridesDerivation: "44'/<coin_type>'/<address>'/<node>/<account>",
+    mandatoryEmptyAccountSkip: 5,
+  },
+  filecoinBIP44: {
+    overridesDerivation: "44'/<coin_type>'/<account>'/<node>/<address>",
```
haammar-ledger, line 181: I'm not sure what you tried with `glif` and `filecoinBIP44` works? Just inverting `<address>` and `<account>`.
To be honest that derivation code is a lot of tech debt and even I am not sure of how it works in advanced cases as such.

Bottomline is, **make sure to test a lot and validate it works as you think it does, in all possible cases.**

haammar-ledger (commented):
@emmanuelm41 Multiple changes in a same PR is usually discouraged, our QA needs to be able to test every change, so please make it as clear as possible in the description how to test everything in this PR. Thanks

libs/ledger-live-common/src/families/filecoin/bridge/utils/utils.ts
emmanuelm41: Done!

libs/coin-framework/src/derivation.ts
emmanuelm41: I would change this one to another tag, but if we remove it, it will break current LL which are using this tag. Do you think there is a better approach?

emmanuelm41: I have applied some changes on tags. What about now?

libs/coin-framework/src/derivation.ts
emmanuelm41: I have changed this in order to make it clearer. What do you think now? Now it is the same we had previously... it will now use `glif` as default derivation scheme for new accounts though.

libs/coin-framework/src/derivation.ts
```diff
@@ -168,14 +168,20 @@ const modes = Object.freeze({
   polkadotbip44: {
     overridesDerivation: "44'/354'/<account>'/0'/<address>'",
   },
-  gliflegacy: {
+  glifLegacy: {
     overridesDerivation: "44'/1'/0'/0/<account>",
     tag: "legacy",
+    mandatoryEmptyAccountSkip: 5,
```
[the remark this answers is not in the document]
emmanuelm41: Done!

emmanuelm41:
@haammar-ledger just committed some changes! please, feel free to resolve comments whenever you consider them ready!

haammar-ledger (approved):

haammar-ledger:
Replaced with https://github.com/LedgerHQ/ledger-live/pull/5796

PR #5731, opened 2023-12-13
Feat/LIVE-9513 config lib dev

hzheng-ledger:
### 📝 Description

1. Add GetValueByKey helper in live-config lib.
2. Default values for GetValueByKey
3. Unit tests for GetValueByKey.
4. live-config lib refactoring
5. Multiple providers supported in live-config lib.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-9513

### ✅ Checklist

- [X] **Impact of the changes:** No impact in this PR. Only non reg test are required.
  - 

---


apps/ledger-live-desktop/src/config/configInit.ts
```diff
@@ -0,0 +1,10 @@
+import { LiveConfig } from "@ledgerhq/live-config/LiveConfig";
+import liveConfigSchema from "@ledgerhq/live-common/config/sharedConfig";
+
+LiveConfig.init({
+  appVersion: __APP_VERSION__,
+  platform: "desktop",
+  environment: process.env.NODE_ENV || "developement",
```
hedi-edelbloute, line 7: shouldn't this be development ? (I didn't check node_env value)

libs/ledger-live-common/src/config/sharedConfig.ts
```diff
@@ -0,0 +1,78 @@
+import { ConfigSchema } from "@ledgerhq/live-config/LiveConfig";
+
+const liveConfig: ConfigSchema = {
```
hedi-edelbloute, line 3: We could divide this config in multiple types, for instance this schema could be CosmosConfig and we extend it here ? just an idea

So cosmos logic stays in cosmos folder

libs/live-config/src/providers/firebaseRemoteConfig/index.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export class FirebaseRemoteConfigProvider implements Provider {
+  public getValue: (key: string) => Value;
+
+  constructor(config: { getValue: (key: string) => Value }) {
+    this.getValue = config.getValue;
+  }
+
+  getValueBykey<K>(key: K, info: ConfigInfo) {
```
hedi-edelbloute, line 19: Can we unit test main methods ? It serves as comment aswell

hedi-edelbloute (commented):
Weren't we supposed to have a generic LiveConfig<T> ?

apps/ledger-live-desktop/src/config/configInit.ts
hzheng-ledger: I am not following what you meant ?
process.env.NODE_ENV can be "test", "development" or "production".

libs/ledger-live-common/src/config/sharedConfig.ts
hzheng-ledger: as this config variable "liveConfig" here is a global variable for all the config in ledger live. (It contains not only the cosmos family config, but other configs in the future)
So I type it as ConfigSchema. 

to inject this, we do this when the app starts:
LiveConfig.setConfig(liveConfig);

if we want to get a strong typed config as CosmosConfig.
we can do that on client side like this:
const cosmosConfig =  LiveConfig.getValueBykey("cosmos") as CosmosConfig;

if we move this config to the cosmos folder, the common part (where we did "LiveConfig.setConfig(liveConfig);") will refer to the cosmos part(inside the families folder) which is an anti pattern.

libs/live-config/src/providers/firebaseRemoteConfig/index.ts
hzheng-ledger: we have unit test for the jsonFileReader provider getValueBykey method in "/libs/live-config/src/__tests__/LiveConfig.test.ts". 
The idea is that the LiveConfig can plug any provider as long as it implements the "getValueBykey" method.
It is not possible to write unit test for firebase remote config provider because the it queries the firebase server.

apps/ledger-live-desktop/src/config/configInit.ts
hedi-edelbloute: yes, you wrote "developement" here :p

hzheng-ledger: very good catch :P 
fixed now

libs/ledger-live-common/src/config/sharedConfig.ts
hedi-edelbloute: import ConfigCosmos from 'families/cosmos/config.ts"
ConfigLLC: {
cosmos: ConfigCosmos
}

libs/live-config/src/providers/firebaseRemoteConfig/index.ts
hedi-edelbloute: it("should return default value if provider is undefined")
it("should return default value if parser returns null")

hzheng-ledger: fixed

libs/ledger-live-common/src/config/sharedConfig.ts
hzheng-ledger: fixed in my another PR:
https://github.com/LedgerHQ/ledger-live/pull/5763/files#diff-32a5530819ada406eb6bf26e05e32b0e3cd028eec5d59615726233eeebccb3b8R6

hedi-edelbloute: Ok

libs/ledger-live-common/src/config/sharedConfig.ts
```diff
@@ -0,0 +1,78 @@
...
+        lcd: "https://cosmoshub4.coin.ledger.com",
+        ledgerValidator: "cosmosvaloper10wljxpl03053h9690apmyeakly3ylhejrucvtm",
+        minGasPrice: 0.025,
+      },
+      desmos: {
+        lcd: "https://desmos-api.ibs.team",
+        minGasPrice: 0.0025,
+      },
+      dydx: {
+        lcd: "https://api.dydx.nodestake.top:443",
```
hedi-edelbloute, line 22: is it me or there are extra spaces before some keys ? maybe run prettier

libs/ledger-live-common/src/config/sharedConfig.ts
```diff
@@ -0,0 +1,78 @@
...
+      },
+      umee: {
+        lcd: "https://umee-api.polkachu.com",
+        minGasPrice: 0.1,
+      },
+      coreum: {
+        lcd: "https://full-node.mainnet-1.coreum.dev:1317",
+        minGasPrice: 0.1,
+      },
+      injective: {
```
hedi-edelbloute, line 70: We miss a ledgerValidator here /!\

https://ledger-live-production-default-rtdb.europe-west1.firebasedatabase.app/config.json

Please check all of them

libs/ledger-live-common/.unimportedrc.json
```diff
@@ -290,6 +290,7 @@
     "**/datasets/**",
     "**/mock/fixtures/**",
     "src/deviceSDK/**",
+    "src/config/sharedConfig.ts",
```
hedi-edelbloute, line 293: Is it actually intended ? Maybe it's because we import this on LLD and LLM only

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
```diff
@@ -0,0 +1,27 @@
+import { Value } from "../firebaseRemoteConfig";
+import { ConfigInfo } from "../../LiveConfig";
+
+const objectParser = (value: unknown): object | undefined => {
+  const stringValue = (value as Value).asString();
+  return stringValue ? JSON.parse(stringValue) : undefined;
+};
+
+export function parser(value: unknown, type: ConfigInfo["type"]) {
```
hedi-edelbloute, line 9: Why do we have to access it with ["type"] ?

libs/live-config/src/providers/jsonFileReader/index.ts
```diff
@@ -0,0 +1,33 @@
...
+      const fileContent = fs.readFileSync(this.filePath);
+      const parsedValue = JSON.parse(fileContent.toString());
+      return parsedValue[key] ?? info.default;
+    } catch (err) {
+      console.error(
+        `The config file doesn't have a key ${key}. Returning default value: ${info.default}`,
+        err,
+      );
+
+      return info.default;
```
hedi-edelbloute, line 30: Shouldn't we manage the case where info is undefined/null ? Just a question

Accessing it directly typescript will warn us, but if we do something dynamic, for example accessing a family or currency dynamically, i'm not sure it will be fine

Maybe we should throw if info is null/undefined ?

libs/live-config/src/providers/jsonFileReader/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  constructor(config: { filePath: string }) {
+    if (!config.filePath.endsWith(".json")) {
+      throw new Error("Only .json files can be read by the provider");
+    }
+    this.filePath = config.filePath;
+  }
+
+  getValueBykey<K>(key: K, info: ConfigInfo) {
+    if (!fs.existsSync(this.filePath)) {
```
hedi-edelbloute, line 16: Shouldn't we use an async method ?

Using a sync method here could cause performance issues

Also I wonder if we really want to use fs, isn't it wrong if we port the app in the web ? maybe just a http request for fetching could work

libs/live-config/src/providers/jsonFileReader/index.ts
```diff
@@ -0,0 +1,33 @@
...
+      throw new Error("Invalid file path, could not read config");
+    }
+
+    try {
+      const fileContent = fs.readFileSync(this.filePath);
+      const parsedValue = JSON.parse(fileContent.toString());
+      return parsedValue[key] ?? info.default;
+    } catch (err) {
+      console.error(
+        `The config file doesn't have a key ${key}. Returning default value: ${info.default}`,
```
hedi-edelbloute, line 26: We could also log this if JSON.parse fails

Maybe separate two cases : key doesn't exist, parsing error

Minor

hedi-edelbloute (commented):
IMO not shippable as is because of the sync method, rest is fine

libs/ledger-live-common/.unimportedrc.json
hzheng-ledger: yes, as we import this on LLD and LLM.
I found that if I don't add this file here. There would be an CI error.

libs/live-config/src/providers/jsonFileReader/index.ts
hzheng-ledger: This "jsonFileReader" will not be used in production. This is just an example to show how we add a new tailored provider in this lib.

And the firebase provider will be used in production. (In ledger-live/libs/live-config/src/providers/firebaseRemoteConfig/index.ts) There is no performance issues as the getValue won't fetch config from remote server. It just fetch config from local cache. In our current codebase, the "getValue" method is not async either.

For reference, here is the codebase of firebase SDK. As you can see, the design of "getValue" method is not supposed to be async.
https://github.com/firebase/firebase-js-sdk/blob/537d39982f90aff50519b8c1ad6d58048fb8f244/packages/remote-config/src/api.ts#L209

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
hzheng-ledger: As we have 
```
type ConfigInfoShape<Type extends keyof ValidConfigTypes> = {
  type: Type;
  default: ValidConfigTypes[Type];
};

export type ConfigInfo =
  | ConfigInfoShape<"string">
  | ConfigInfoShape<"boolean">
  | ConfigInfoShape<"number">
  | ConfigInfoShape<"object">
  | ConfigInfoShape<"array">;

```

the ConfigInfo["type"] accepts "string", "boolean", "number", "array" and "object". Do you have any suggestion?

libs/ledger-live-common/src/config/sharedConfig.ts
hzheng-ledger: good catch. In fact, I miss this config because the "SharedConfig" is a migration of the current codebase from: https://github.com/LedgerHQ/ledger-live/blob/24ebd63bf5c6b9534e4d6f103736f2bfc2d47df5/libs/ledger-live-common/src/config/defaultConfig.ts
I now add the missing injective config in my another PR: 
https://github.com/LedgerHQ/ledger-live/pull/5763/commits/3e4d8cdb7644737b0b5d4079e99e3320d9265561

libs/ledger-live-common/src/config/sharedConfig.ts
hzheng-ledger: it is weird. I don't find the "extra spaces" that you mentioned. 
by the way, this file is not longer used after my another PR https://github.com/LedgerHQ/ledger-live/pull/5763/commits/

I run the ```pnpm prettier ``` on my side and it seems that everything is OK

libs/live-config/src/providers/jsonFileReader/index.ts
hzheng-ledger: I think it doesn't matter a lot as the jsonFileReader is not used in production. It is just an example to show how to use this lib. We use only firebaseRemoteConfig as provider in production.

libs/live-config/src/providers/jsonFileReader/index.ts
hzheng-ledger: the jsonFileReader is not used in production. This is just an example to show how we add a new tailored provider in this lib.

We can handle the case for firebase provider when info is null/undefined 
please refer to
 /libs/live-config/src/providers/firebaseRemoteConfig/index.ts
and 
libs/live-config/src/providers/firebaseRemoteConfig/parser.ts

in the case of firebase provider that we use in production, when info is null/undefined, it means that when we use setConfig from our client. The injected config doesn't include that key. this is the case of featureflag. In my implementation, the setConfig only inject the cosmos family related config. So, when info is null/undefined, I don't throw an exception, instead, I handle it as an object and parse it with JSON.parse (see  /libs/live-config/src/providers/firebaseRemoteConfig/index.ts)

libs/live-config/src/featureFlags/firebaseFeatureFlags.ts
```diff
@@ -3,13 +3,13 @@ import { snakeCase } from "lodash";
...
 import { getEnv } from "@ledgerhq/live-env";
-import { LiveConfig } from "./LiveConfig";
+import { LiveConfig } from "../LiveConfig";
 
 export type FirebaseFeatureFlagsProviderProps = PropsWithChildren<unknown>;
 export const formatToFirebaseFeatureId = (id: string) => `feature_${snakeCase(id)}`;
 
 export const checkFeatureFlagVersion = (feature: Feature) => {
-  const platform = LiveConfig.getInstance().platform;
+  const platform = LiveConfig.configInstance!.platform;
```
haammar-ledger, line 12: I don't think the ! is needed on `configInstance!` you should remove it if possible (also on other places in this file)

libs/live-config/src/LiveConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export type ConfigSchema = Record<string, ConfigInfo>;
+
+export class LiveConfig {
+  public provider?: Provider;
+  public config: ConfigSchema = {};
+  public appVersion?: string;
+  public platform?: string;
+  public environment?: string;
+  public static configInstance: LiveConfig = new LiveConfig();
```
haammar-ledger, line 31:
```suggestion
  public static instance: LiveConfig = new LiveConfig();
```

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
```diff
@@ -0,0 +1,32 @@
+import { Value } from "../firebaseRemoteConfig";
+import { ConfigInfo } from "../../LiveConfig";
+
+const objectParser = (value: unknown): object | undefined => {
```
haammar-ledger, line 4:
```suggestion
const objectParser = (value: Value): object | undefined => {
```

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
```diff
@@ -0,0 +1,32 @@
...
+  const stringValue = (value as Value).asString();
+  try {
+    return stringValue ? JSON.parse(stringValue) : undefined;
+  } catch (error) {
+    console.error(`JSON.parse error: ${error}. Original string: ${stringValue}`);
+    return undefined;
+  }
+};
+
+export function parser(value: unknown, type: ConfigInfo["type"]) {
```
haammar-ledger, line 14: Since you only use value as Value, you can make the assumption in the param type directly:
```suggestion
export function parser(value: Value, type: ConfigInfo["type"]) {
```

libs/live-config/src/providers/index.ts
```diff
@@ -0,0 +1,7 @@
+import { ConfigInfo } from "../LiveConfig";
+
+export interface Provider {
+  getValueBykey<K>(key: K, info: ConfigInfo);
```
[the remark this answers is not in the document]
hzheng-ledger: good point. Fixed.

libs/live-config/src/featureFlags/firebaseFeatureFlags.ts
hzheng-ledger: good point. Fixed.

libs/live-config/src/LiveConfig.ts
hzheng-ledger: fixed

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
hzheng-ledger: good point. Fixed.

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
hzheng-ledger: fixed

libs/live-config/src/providers/jsonFileReader/index.ts
hedi-edelbloute: If this provider isn't used why don't we just add how to add a provider in the documentation ? I think we should not keep unused code, WDYT ?

hzheng-ledger: yes, we can remove it of course. but it is used in the unit tests in live-config lib. This is why we keep it here.
we can't use the firebase provider in unit tests directly as it depends on remove server. so we use this simple provider instead.

libs/live-config/src/providers/firebaseRemoteConfig/parser.ts
hedi-edelbloute: Oops indeed it's fine

libs/live-config/src/providers/jsonFileReader/index.ts
hedi-edelbloute: Only used as a unit test

libs/live-config/src/providers/jsonFileReader/index.ts
hedi-edelbloute: Will only be used in unit tests

libs/live-config/src/providers/jsonFileReader/index.ts
hedi-edelbloute: Will only be used in unit tests

hedi-edelbloute (dismissed):
lgtm when we will move jsonFileReader

hzheng-ledger:
> lgtm when we will move jsonFileReader

jsonFileReader is removed. Thanks

haammar-ledger (dismissed):
Since this is not awaited for a specific release, let's try to merge _just after_ the start of next pre-release, so we have at least a few days to fix any unexpected impact.

haammar-ledger (approved):

gre:
catching up after this merge. it seems that some code is moving back to live-common? it could have been better to make a new library for the feature flag hooks part, as we have a general effort to split things out of live-common. anyway, we'll re-move this later.
please wait before merging when code affects code owners from other team 🙏

PR #5743, opened 2023-12-13
[SUPPORT][LIVE-10636] Improve code cov for coin-evm editEvmTransaction codebase

chabroA:
### 📝 Description

Add missing tests to improve code cov for coin-evm editEvmTransaction codebase:
- `operation.ts`
- `getTransactionStatus.ts`
- `hasMinimumFunds.ts`
- `isTransactionConfirmed.ts`
- `isStrategyDisabled.ts`
- `getTransactionStatus.ts`

Removed an unreachable statement in the `getStuckAccountAndOperation` function in `libs/coin-evm/src/operation.ts` (see related comment)

|before1|after1|

|before2|after2|

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-10636

### ✅ Checklist

- [x] `npx changeset`  was attached
  - N/A, only adding some tests

---

### DIFF
--- a/libs/coin-evm/src/editTransaction/index.ts
+++ b/libs/coin-evm/src/editTransaction/index.ts
@@ -1,3 +1,5 @@
+/* istanbul ignore file: pure exports */
+
 import { getEditTransactionStatus } from "../getTransactionStatus";
 import { getEditTransactionPatch } from "./getEditTransactionPatch";
 import { getFormattedFeeFields } from "./getFormattedFeeFields";
--- a/libs/coin-evm/src/getTransactionStatus.ts
+++ b/libs/coin-evm/src/getTransactionStatus.ts
@@ -303,10 +303,10 @@ type EditTransactionValidationIssues = Partial<
  * the transaction will be rejected by the network with a
  * "replacement transaction underpriced" error
  */
-const validateEditTransaction = ({
-  editType,
+export const validateEditTransaction = ({
   transaction,
   transactionToUpdate,
+  editType,
 }: {
   transaction: EvmTransaction;
   transactionToUpdate: EvmTransaction;
--- a/libs/coin-evm/src/operation.ts
+++ b/libs/coin-evm/src/operation.ts
@@ -77,7 +77,6 @@ export const getStuckAccountAndOperation = (
   }
 
   const oldestStuckOperation = stuckOperations.reduce((oldestOp, currentOp) => {
-    if (!oldestOp) return currentOp;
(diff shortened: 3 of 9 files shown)


libs/coin-evm/src/operation.ts
```diff
@@ -77,7 +77,6 @@ export const getStuckAccountAndOperation = (
   }
 
   const oldestStuckOperation = stuckOperations.reduce((oldestOp, currentOp) => {
-    if (!oldestOp) return currentOp;
```
chabroA, line 80: In the provided code, `oldestOp` could be `undefined` only on the first iteration of the reduce function.
This is because the reduce function in JavaScript starts with the first element of the array as the initial value of the accumulator (`oldestOp` in this case) if no initial value is provided.
However, after the first iteration, `oldestOp` will always be an Operation object because that's what we are returning at the end of each iteration.
So, in this specific case, as long as the `stuckOperations` array is not empty (which we are checking with `stuckOperations.length === 0` just above), `oldestOp` will never be `null` or `undefined` during the execution of the reduce function.

libs/coin-evm/src/getTransactionStatus.ts
```diff
@@ -303,10 +303,10 @@ type EditTransactionValidationIssues = Partial<
  * the transaction will be rejected by the network with a
  * "replacement transaction underpriced" error
  */
-const validateEditTransaction = ({
-  editType,
+export const validateEditTransaction = ({
   transaction,
   transactionToUpdate,
+  editType,
```
chabroA, line 309: I just moved the `editType` in the param list to match the type ordering below (since `editType` is optional).
It does not change anything since the function param is an object, order does not matter

libs/coin-evm/src/editTransaction/index.ts
```diff
@@ -1,3 +1,5 @@
+/* istanbul ignore file: pure exports */
```
lvndry, line 1: Why is it needed ?

chabroA: To explicitly skip the coverage requirement for this file, since this is an empty shell only used to export functions that are already unit-tested in their modules.

Similar to what is done in other pure exports files, like https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-evm/src/api/node/rpc.native.ts

cf. https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

lambertkevin (approved):
Thanks for the coverage <3 🚀

PR #5785, opened 2023-12-20
Remove usage of deprecated navigator.product

lvndry:
### 📝 Description
`navigator.product` is now deprecated. It was used to inject an https agent in all platforms besides React Native. We now use an other way to detect the platform running the code. 

* Remove usage of `navigator.product` and replace it with the already existing `LEDGER_CLIENT_VERSION`

On an other note I would probably look for an other solution would be using [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch#keepalive) that supports keep alive with no additional third party.

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-10727

### ✅ Checklist

  - 

---


libs/live-network/src/network.ts
```diff
@@ -99,15 +99,6 @@ export const errorInterceptor = (error: InterceptedError): InterceptedError => {
 axios.interceptors.request.use(requestInterceptor);
 axios.interceptors.response.use(responseInterceptor, errorInterceptor);
 
-// not react native
-if (!(typeof navigator !== "undefined" && navigator.product === "ReactNative")) {
-  // the keepAlive is necessary when we make a lot of request in in parallel, especially for bitcoin sync. Otherwise, it may raise "connect ETIMEDOUT" error
-  // refer to https://stackoverflow.com/questions/63064393/getting-axios-error-connect-etimedout-when-making-high-volume-of-calls
-  // eslint-disable-next-line global-require,@typescript-eslint/no-var-requires
-  const https = require("https");
-  axios.defaults.httpsAgent = new https.Agent({ keepAlive: true });
```
hzheng-ledger, line 108: I think you shouldn't remove this code. You just need to replace the condition 
```
if (!(typeof navigator !== "undefined" && navigator.product === "ReactNative"))
```

we should run this piece of code for LLD

hzheng-ledger: if you remove this piece of code, it raises an "connect ETIMEDOUT" error in windows if I remember well

lvndry: Not sure to understand why removing this code would cause a timeout on windows ? The condition is only executed of we're on ReactNative

hzheng-ledger (approved):
looks good now. thanks a lot

hedi-edelbloute (dismissed):
I think this whole thing needs to be a bit reworked... but it will do for now, according to comments and what i see it should be equivalent to what we had before

hedi-edelbloute (approved):

PR #5798, opened 2023-12-22
[BUGFIX] : Favorites currencies in Market do not load properly

mcayuelas-ledger:
### 📝 Description

Favorites currencies in Market do not load properly

### ❓ Context

- **JIRA or GitHub link**: [LIVE-10674]

### ✅ Checklist

  - 

---

[LIVE-10674]: https://ledgerhq.atlassian.net/browse/LIVE-10674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/screens/Market/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/Market/index.tsx
@@ -85,14 +85,8 @@ const BottomSection = ({ navigation }: { navigation: NavigationProps["navigation
   const { range, orderBy, order, top100 } = requestParams;
   const starredMarketCoins: string[] = useSelector(starredMarketCoinsSelector);
   const filterByStarredAccount: boolean = useSelector(marketFilterByStarredAccountsSelector);
-  const firstMount = useRef(true); // To known if this is the first mount of the page
 
   useEffect(() => {
-    if (firstMount.current) {
-      // We don't want to refresh the market data directly on mount, the data is already refreshed with wanted parameters from MarketDataProviderWrapper
-      firstMount.current = false;
-      return;
-    }
     if (filterByStarredAccount) {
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/screens/Market/index.tsx
```diff
@@ -85,14 +85,8 @@ const BottomSection = ({ navigation }: { navigation: NavigationProps["navigation
...
   const starredMarketCoins: string[] = useSelector(starredMarketCoinsSelector);
   const filterByStarredAccount: boolean = useSelector(marketFilterByStarredAccountsSelector);
-  const firstMount = useRef(true); // To known if this is the first mount of the page
 
   useEffect(() => {
-    if (firstMount.current) {
-      // We don't want to refresh the market data directly on mount, the data is already refreshed with wanted parameters from MarketDataProviderWrapper
-      firstMount.current = false;
-      return;
-    }
```
gre, line 95: following down the rabbit hole =>

i imagine this was made to prevent the re-rendering "cascading" problem. there may be a wider problem to solve here: the fact we do an action depending on parameters -> it sounds like these parameters should instead be fed into the 'useMarketData()'  function (which don't take param) because it's dependency of them, there is probably an antipattern to dig later on this topic.

a typical example of something that shouldn't be an effect https://react.dev/learn/you-might-not-need-an-effect

mcayuelas-ledger: The problem is that the data is not always up to date when it is first rendered. What's more, there's no intenpestive rerender when you remove the condition

PR #5876, opened 2024-01-09
[Polkadot] Update clear nomination rule

lvndry:
### 📝 Description

Users can now clear nomination even if they don't have +250 DOT bounded.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10806

### ✅ Checklist

  - 

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
@@ -9,7 +9,6 @@ import {
   canBond,
   canUnbond,
   hasPendingOperationType,
-  isStashAccount,
 } from "@ledgerhq/live-common/families/polkadot/logic";
 import { openModal } from "~/renderer/actions/modals";
 import Box from "~/renderer/components/Box";
@@ -131,7 +130,7 @@ const ManageModal = ({ account, source, ...rest }: Props) => {
   const hasUnlockedBalance = unlockedBalance && unlockedBalance.gt(0);
   const hasPendingWithdrawUnbondedOperation = hasPendingOperationType(account, "WITHDRAW_UNBONDED");
   const nominationEnabled = !electionOpen && accountCanNominate;
-  const chillEnabled = !electionOpen && isStashAccount(account) && nominations?.length;
+  const chillEnabled = !electionOpen && nominations?.length;
   const bondingEnabled = !electionOpen && canBond(account);
   const unbondingEnabled = !electionOpen && canUnbond(account);
   const withdrawEnabled =
--- a/apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
@@ -9,6 +9,7 @@ import {
   canBond,
   canUnbond,
   hasPendingOperationType,
+  isStashAccount,
 } from "@ledgerhq/live-common/families/polkadot/logic";
 import { openModal } from "~/renderer/actions/modals";
 import Box from "~/renderer/components/Box";
@@ -126,10 +127,11 @@ const ManageModal = ({ account, source, ...rest }: Props) => {
   );
 
   const electionOpen = staking?.electionClosed !== undefined ? !staking?.electionClosed : false;
+  const accountCanNominate = canNominate(account);
   const hasUnlockedBalance = unlockedBalance && unlockedBalance.gt(0);
   const hasPendingWithdrawUnbondedOperation = hasPendingOperationType(account, "WITHDRAW_UNBONDED");
-  const nominationEnabled = !electionOpen && canNominate(account);
-  const chillEnabled = !electionOpen && canNominate(account) && nominations?.length;
+  const nominationEnabled = !electionOpen && accountCanNominate;
+  const chillEnabled = !electionOpen && isStashAccount(account) && nominations?.length;
   const bondingEnabled = !electionOpen && canBond(account);
   const unbondingEnabled = !electionOpen && canUnbond(account);
   const withdrawEnabled =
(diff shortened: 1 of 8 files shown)


apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
```diff
@@ -126,10 +127,11 @@ const ManageModal = ({ account, source, ...rest }: Props) => {
   );
 
   const electionOpen = staking?.electionClosed !== undefined ? !staking?.electionClosed : false;
+  const accountCanNominate = canNominate(account);
   const hasUnlockedBalance = unlockedBalance && unlockedBalance.gt(0);
   const hasPendingWithdrawUnbondedOperation = hasPendingOperationType(account, "WITHDRAW_UNBONDED");
-  const nominationEnabled = !electionOpen && canNominate(account);
-  const chillEnabled = !electionOpen && canNominate(account) && nominations?.length;
+  const nominationEnabled = !electionOpen && accountCanNominate;
+  const chillEnabled = !electionOpen && isStashAccount(account) && nominations?.length;
```
hedi-edelbloute, line 134: If i'm not mistaken, this is the only real update in terms of behavior, right ? (same for LLM)

I think we should not change isController to isStashAccount, IMO it is misleading for the dev, we should use hasStash name (as it is actually what the method does : check if stash is defined)

History : Controller is an account that can control a stash account, acting as a staking proxy so we don't sign those tx on the stash account directly

This mechanism has been discontinued but this doesn't mean a controller account becomes a stash (if you see what i mean :sweat_smile: )

libs/coin-polkadot/src/logic.ts
```diff
@@ -24,11 +24,9 @@ export const isStash = (a: PolkadotAccount): boolean => {
 /**
  * Returns true if account is a controller.
  * when account is a controller, we have the information of which stash it controls
- *
- * @param {PolkadotAccount} a
  */
-export const isController = (a: PolkadotAccount): boolean => {
-  return !!a.polkadotResources?.stash;
+export const isStashAccount = (account: PolkadotAccount): boolean => {
+  return !!account.polkadotResources?.stash;
```
hedi-edelbloute, line 29: resources && resources.stash != null ? I think it's equivalent and more readable

Btw can we add unit test on this ? Allowing us to check easily if both syntax are the same, test each branch eg:

- polkadotResources stash undefined/null
- polkadotResources stash defined
- polkadotResources hasn't a stash key defined

apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
lvndry: Okay I might I have misunderstood the role of the controller when doing this change then 👍🏿 I'll revert it

hedi-edelbloute: image
Are you sure we need to edit the method name ?

It looks like stash is actually the linked stash to the controller
And controller is the controller account linked to this stash

hedi-edelbloute: Then it means the fix is just removing the minimumBondBalance condition on chillEnabled (which you did, but also edited the method name)

libs/coin-polkadot/src/logic.ts
hedi-edelbloute: If you don't update the method, i'll approve without unit tests there as we only cover what we edit
image

lvndry: Typechecking is not great here because stash is of type `string | null | undefined`
I can do the check `typeof account.polkadotResources?.stash !== undefined` which does what we want.

[source](https://stackoverflow.com/a/19891952/7280833)

apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
lvndry: So the `isStash` method checks the `controller` property and the `isController` method checks the `stash` property.
I reverted the name to how it was I don't mind but this is confusing to me

hedi-edelbloute: Yes it's kind of a link between them i guess

If you have a controller it means you are a stash, if you have a stash property it means you are a controller of that stash

libs/coin-polkadot/src/logic.ts
hedi-edelbloute: != null includes both null and undefined, and the missing key

hedi-edelbloute: I find `typeof account.polkadotResources?.stash !== undefined` even more weird than the base expression :sweat_smile:

lvndry: If a variable is not null doesn't mean it's not undefined right 🤔

apps/ledger-live-desktop/src/renderer/families/polkadot/ManageModal.tsx
hedi-edelbloute: It's wrote in comments :
```
 * Returns true if account is a stash.
 * When the account is a stash, we have the information of which account is the controller
```

libs/coin-polkadot/src/logic.ts
hedi-edelbloute: yes it means it's not undefined

!= null includes null, undefined, and key not present

lvndry: Ah yeah you mean using `!=` and not `!==` in that sense I don't think it's more clear than `!!` then

hedi-edelbloute: Ok let's use the base condition then

My thoughts where :

We want resources to be defined (`resources`) and (`&&`) resource stash to be defined (`resources.stash != null `)

A bit more human

hedi-edelbloute (approved):

haammar-ledger (approved):

PR #5913, opened 2024-01-15
[bugfix] Solana: extend the retry mechanism to mitigate error when fetching fees

haammar-ledger:
### 📝 Description

For an unknown reason, it seems that the Solana backend call made by `getFeeForMessage` often fails to return a valid fee value. 
The current mitigation was to call a `retryWithNewBlockhash` function, but only once.

This is apparently not enough anymore, so this PR extends the retry mechanism to an arbitrary 10 times.
Manual testing shows that this makes the user facing error to happen a lot less.

In case of a failure even after all the retries, this PR also changes the user-facing message for a cleaner NetworkError, suggesting to the user to retry, which in most cases has a good chance to work the second time.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10945

### ✅ Checklist

  - Solana sync of each account can take up to ~50s longer, in the worst scenario

---

- [x] **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

### DIFF
--- a/.changeset/lemon-spies-leave.md
+++ b/.changeset/lemon-spies-leave.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Mitigate Solana error when fetching fees for an address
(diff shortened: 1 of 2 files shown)


hedi-edelbloute (approved):
LGTM, hopefully we find a better fix but IMO nothing can be done if backend result is inconsistent

PR #5923, opened 2024-01-16
[bugfix] Solana: extend the retry mechanism to mitigate error when fetching fees - RELEASE

haammar-ledger:
### 📝 Description

**(Backport of https://github.com/LedgerHQ/ledger-live/pull/5913 for release branch)**

For an unknown reason, it seems that the Solana backend call made by `getFeeForMessage` often fails to return a valid fee value. 
The current mitigation was to call a `retryWithNewBlockhash` function, but only once.

This is apparently not enough anymore, so this PR extends the retry mechanism to an arbitrary 10 times.
Manual testing shows that this makes the user facing error to happen a lot less.

In case of a failure even after all the retries, this PR also changes the user-facing message for a cleaner NetworkError, suggesting to the user to retry, which in most cases has a good chance to work the second time.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10945

### ✅ Checklist

  - 

---

### DIFF
--- a/libs/ledger-live-common/src/families/solana/tx-fees.ts
+++ b/libs/ledger-live-common/src/families/solana/tx-fees.ts
@@ -5,6 +5,9 @@ import { Transaction, TransactionModel } from "./types";
 import { assertUnreachable } from "./utils";
 import { VersionedTransaction as OnChainTransaction } from "@solana/web3.js";
 import { log } from "@ledgerhq/logs";
+import { NetworkError } from "@ledgerhq/errors";
+
+const MAX_RETRIES = 10; // TODO Candidate to put in coin config
 
 export async function estimateTxFee(
   api: ChainAPI,
@@ -14,12 +17,18 @@ export async function estimateTxFee(
   const tx = createDummyTx(address, kind);
   const [onChainTx] = await buildTransactionWithAPI(address, tx, api);
 
-  const fee = await api.getFeeForMessage(onChainTx.message);
+  let fee = await api.getFeeForMessage(onChainTx.message);
+  let retries = 0;
+
+  while (typeof fee !== "number" && retries++ < MAX_RETRIES) {
+    // FIXME HACK: sometimes fee is not a number, retrying with a next blockhash
+    log("debug", `api.getFeeForMessage returned invalid fee: <${fee}>`);
+    fee = await retryWithNewBlockhash(api, onChainTx);
+  }
 
   if (typeof fee !== "number") {
-    log("error", `api.getFeeForMessage returned invalid fee: <${fee}>`);
-    // TEMP HACK: sometimes fee is not a number, retrying with a next blockhash
-    return retryWithNewBlockhash(api, onChainTx);
+    log("error", `unexpected fee: <${fee}>, after retry with a new blockhash`);
+    throw new NetworkError();
   }
   return fee;
 }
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/src/families/solana/tx-fees.ts
```diff
@@ -5,6 +5,9 @@ import { Transaction, TransactionModel } from "./types";
 import { assertUnreachable } from "./utils";
 import { VersionedTransaction as OnChainTransaction } from "@solana/web3.js";
 import { log } from "@ledgerhq/logs";
+import { NetworkError } from "@ledgerhq/errors";
+
+const MAX_RETRIES = 10; // TODO Candidate to put in coin config
```
gre, line 10: before putting this env, i think i would consider this for backend to fix the stability, don't you think?

libs/ledger-live-common/src/families/solana/tx-fees.ts
```diff
@@ -14,12 +17,18 @@ export async function estimateTxFee(
...
   const [onChainTx] = await buildTransactionWithAPI(address, tx, api);
 
-  const fee = await api.getFeeForMessage(onChainTx.message);
+  let fee = await api.getFeeForMessage(onChainTx.message);
+  let retries = 0;
+
+  while (typeof fee !== "number" && retries++ < MAX_RETRIES) {
+    // FIXME HACK: sometimes fee is not a number, retrying with a next blockhash
+    log("debug", `api.getFeeForMessage returned invalid fee: <${fee}>`);
+    fee = await retryWithNewBlockhash(api, onChainTx);
```
gre, line 26: i would advice to add a small delay after this, it's a general good practice to add a delay (even one that becomes higher and higher) to avoid high pressure from client spamming an api

libs/ledger-live-common/src/families/solana/tx-fees.ts
haammar-ledger: It's not certain that it's a stability issue, Solana is a bit complex and it may just be "normal" blockchain behaviour.

But I agree, discussion is already ongoing with Solana devs and I've asked that exact question, it's on them to tell whether it's normal backend behaviour or not.

In the meantime this is only a workaround.

libs/ledger-live-common/src/families/solana/tx-fees.ts
haammar-ledger: You're right but it's already done in the internal implem of `retryWithNewBlockhash`
https://github.com/LedgerHQ/ledger-live/blob/26f589097ccdeb5853aba2e2ce963533287bb681/libs/ledger-live-common/src/families/solana/tx-fees.ts#L186

hedi-edelbloute (approved):

libs/ledger-live-common/src/families/solana/tx-fees.ts
haammar-ledger: @gre Got the anwser just now: it's indeed not backend instability, it's a strong requirement to have and use latest block hash when calling the `getFeeForMessage` function. And block creation time on Solana is < 1s, so it easily happens that the block hash is already outdated when calling `getFeeForMessage`...

I'm discussing the best approach, we'll probably go for less retries and a hardcoded fees fallback if it fails too much.

PR #5975, opened 2024-01-22
LLM - New architecture for market page

KVNLS:
### 📝 Description

The goal of this PR is to iterate on a new architecture for LLM. Here are the steps required to conclude on a new architecture:

 - [x] Move Market pages under a new architecture folder, with controllers decoupled from views. A basic Container / Presentational pattern is used. The goal is to have small components that we can move easily and restructure.
- [x] Add integration tests tooling to cover the feature.
- [ ] Cover all use cases with Integration tests.
- [ ] POC a better DX by moving from Container / Presentational pattern to Hooks. Can we inject them? To avoid coupling our views to hooks too much. This must be validated by integration tests written before.
- [ ] Rework how we fetch the data (LLC part)

Behind the following feature flag: `llmMarketNewArch`

### ❓ Context

- **JIRA or GitHub link**: 

### ✅ Checklist

  - 

---


apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketDetail/MarketDetailCont.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const resetState = () => {
+      // selectCurrency();
+    };
+    const sub = navigation.addListener("blur", resetState);
+    return () => {
+      sub();
+    };
+  }, [selectCurrency, resetSearchOnUmount, navigation]);
+
+  const allAccounts = useSelector(flattenAccountsByCryptoCurrencyScreenSelector(internalCurrency));
```
gre, line 60: do you know if allAccounts ref will change on each render() or if it's cached and only recalc if accounts store changes? i suspect it may cascade to perf issue if it's not the case

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketDetail/components/MarketStats/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  counterCurrency: string;
+}) {
+  const { t } = useTranslation();
+  const { locale } = useLocale();
+
+  let loc = locale;
+  // TEMPORARY : quick win to transform arabic to english
+  if (locale === "ar") {
+    loc = "en";
+  }
```
gre, line 39: same

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketList/components/BottomSection/BottomSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+        </StyledBadge>
+      </TouchableOpacity>
+      <SortBadge
+        label={t("market.filters.sort")}
+        valueLabel={t(
+          top100 ? `market.filters.order.topGainers` : `market.filters.order.${orderBy}`,
+        )}
+        Icon={getIcon(top100, order)}
+        value={top100 ? "top100" : `${orderBy}_${order}`}
+        options={[
```
gre, line 123: it seems like the creation of this options array is static and should be moved out to a global var?

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketList/components/BottomSection/BottomSectionCont.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { requestParams, counterCurrency, refresh } = useMarketData();
+  const { range, orderBy, order, top100 } = requestParams;
+  const starredMarketCoins: string[] = useSelector(starredMarketCoinsSelector);
+  const filterByStarredAccount: boolean = useSelector(marketFilterByStarredAccountsSelector);
+  const firstMount = useRef(true); // To known if this is the first mount of the page
+
+  useEffect(() => {
+    if (firstMount.current) {
+      // We don't want to refresh the market data directly on mount, the data is already refreshed with wanted parameters from MarketDataProviderWrapper
+      firstMount.current = false;
```
gre, line 25: it's an antipattern of react effect hooks, there may be better ways to treat this in the future and bailout from the 'first' case

apps/ledger-live-mobile/src/newArch/features/Market/utils/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const I = t(`numberCompactNotation.${i}`);
+
+    const formattedNumber = number.replace(/([0-9,. ]+)/, `${sign}$1 ${I} `);
+
+    return formattedNumber;
+  }
+
+  // FIXME: HOW DID THIS WORK WHEN CURRENCY IS EMTPY
+  // PLEASE FIX
```
gre, line 96: 😱

gre (commented):
there is a lot of code complexity in current codebase of the Market components that would deserve another deeper look, but i guess it's not the purpose of this PR that copied them from the past, correct? i give a few feedbacks on some points i have seen

hedi-edelbloute (dismissed):
LGTM overall :+1: 

Those integration tests on components look promising ! :smile: 

I just did not understand the need for a feature flag, shouldn't QA phase be enough ? At some point we will need to update the code anyway :thinking: 

If it is to be able to rollback if a problem occurs, maybe we can dev a proper rollout/feature arch ? With the app split by features aswell :thinking: I don't know how lazy loading is supported by our stack but we even could not include it in the bundle if we migrate to web on LLD ! (I know lazy loading of modules is a feature on angular so prob react handles this aswell)

apps/ledger-live-mobile/src/newArch/features/Market/__integrations__/searchACoin.integration.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const searchInput = await screen.findByTestId("search-box");
+    await user.type(searchInput, "BTC");
+
+    await waitForElementToBeRemoved(() => screen.queryByText("Ethereum (ETH)"));
+
+    expect(await screen.findByText("Bitcoin (BTC)")).toBeOnTheScreen();
+
+    await user.press(screen.getByText("Bitcoin (BTC)"));
+
+    expect(await screen.findByText("Price Statistics")).toBeOnTheScreen();
```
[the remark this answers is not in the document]
KVNLS: You can read more about queries here: https://callstack.github.io/react-native-testing-library/docs/how-should-i-query#idiomatic-query-predicates
The best query is *ByRole to enforce the accessibility of our application. It's easier on the web with aria.

> [The more your tests resemble the way your software is used, the more confidence they can give you.](https://twitter.com/kentcdodds/status/977018512689455106)

*ByTestId is the last query you'd like to use since it's the furthest from our final user.

apps/ledger-live-mobile/__tests__/handlers/index.ts
```diff
@@ -0,0 +1,9 @@
+import marketHandlers from "./market";
+
+export const IGNORED_REQUESTS = [
```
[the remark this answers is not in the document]
KVNLS: That's exactly it. `ALLOWED_REQUESTS` make me think that we authorize these requests to be intercepted by the proxy. I agree with you that `IGNORED_REQUESTS` is not explicit enough.
Wdyt of `ALLOWED_UNHANDLED_REQUESTS`?

KVNLS:
> LGTM overall 👍
> 
> Those integration tests on components look promising ! 😄
> 
> I just did not understand the need for a feature flag, shouldn't QA phase be enough ? At some point we will need to update the code anyway 🤔
> 
> If it is to be able to rollback if a problem occurs, maybe we can dev a proper rollout/feature arch ? With the app split by features aswell 🤔 I don't know how lazy loading is supported by our stack but we even could not include it in the bundle if we migrate to web on LLD ! (I know lazy loading of modules is a feature on angular so prob react handles this aswell)

We want to make sure that we can rollout the feature as you explained. I agree that it's not ideal, but with what we have today I do not see any other solution. Can be a subject for a dedicated PR.
Regarding lazy loading, we already tried it with this task https://ledgerhq.atlassian.net/browse/LIVE-9814 (without success, we have another task to understand why we had no impact, is it already handled by react-native?)

KVNLS:
> there is a lot of code complexity in current codebase of the Market components that would deserve another deeper look, but i guess it's not the purpose of this PR that copied them from the past, correct? i give a few feedbacks on some points i have seen

Yes, I copy pasted the code and split components when needed. I agree that there's too much complexity on this feature and I will fix the ones you've seen.

apps/ledger-live-mobile/__tests__/handlers/index.ts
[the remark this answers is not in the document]
hedi-edelbloute: I'm fine with it :+1: or `UNHANDLED_REQUEST_WHITELIST`, both are ok !

apps/ledger-live-mobile/src/newArch/features/Market/__integrations__/searchACoin.integration.test.tsx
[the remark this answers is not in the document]
hedi-edelbloute: I understand, so let's focus on roles/text etc, thanks

apps/ledger-live-mobile/src/newArch/features/Market/components/MarketRowItem/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        {priceChangePercentage !== null && !isNaN(priceChangePercentage) ? (
+          <DeltaVariation percent value={priceChangePercentage} />
+        ) : (
+          <Text variant="body" color="neutral.c70">
+            {" "}
+            -
+          </Text>
+        )}
+      </Flex>
+    </Flex>
```
mbertin-ledger, line 110: [ASK] I know, you didn't refactor components but here some logic should be directly inside the component (countervaluesformatter, isNaN, etc.), you didn't not change it right ?

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketCurrencySelect/MarketCurrencySelectCont.tsx
```diff
@@ -0,0 +1,41 @@
...
+    },
+    [dispatch, navigation, setCounterCurrency],
+  );
+
+  return (
+    <MarketCurrencySelect
+      items={items}
+      onSelectCurrency={onSelectCurrency}
+      counterCurrency={counterCurrency}
+    />
```
mbertin-ledger, line 37: [COMMENT] Thx for extracting view logic outside component and creating container ⭐

mbertin-ledger (dismissed):
THX !

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketDetail/MarketDetail.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const [hoveredItem, setHoverItem] = useState<Item | null | undefined>(null);
+  const { t } = useTranslation();
+  const { colors } = useTheme();
+  const { locale } = useLocale();
+
+  let loc = locale;
+  // TEMPORARY : quick win to transform arabic to english
+  if (locale === "ar") {
+    loc = "en";
+  }
```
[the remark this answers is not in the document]
KVNLS: As it's behind FF, I will remove this code for new market

apps/ledger-live-mobile/src/newArch/features/Market/Navigator.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
```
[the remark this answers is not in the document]
KVNLS: The idea I had in mind for the migration was the following:
- Make a clear distinction between what is under the new arch and what is not by defining a dedicated folder => `newArch`.
- Use a dedicated alias to point to this `newArch` folder => LLM
```
      "~/*": ["src/*"],
      "LLM/*": ["src/newArch/*"],
```
- New features are developed under the newArch, are tested, and follow the guidelines. We progressively migrate existing features when we work on them.

Once we are ready to switch completely to the new arch we remove the `newArch` folder and we have to update the alias to make it work.

apps/ledger-live-mobile/src/components/RootNavigator/BaseNavigator.tsx
```diff
@@ -391,16 +394,20 @@ export default function BaseNavigator() {
...
-          component={MarketCurrencySelect}
-          options={{
-            title: t("market.filters.currency"),
-            headerLeft: () => null,
-            // FIXME: ONLY ON BOTTOM TABS AND DRAWER NAVIGATION
-            // unmountOnBlur: true,
-          }}
-        />
+        {marketNewArch?.enabled ? (
+          MarketNavigator({ Stack })
```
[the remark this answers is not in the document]
KVNLS: What I want to do is to compose our Navigators. Initially, I tried with your recommendation:
`<MarketNavigator Stack={Stack}/>`, but if I do so I have the following error:
image

This is well described here: https://github.com/react-navigation/react-navigation/issues/9578

I want to compose our Navigators for two reasons:
- Keep our Navigators close to our features. Each feature should be able to register the screens without having to do it in a common file as we have today.
- Ease Integration tests, as you will test User Stories, and most of the time that means navigation between multiple screens of a feature. If you have a Navigator for your feature you can render it directly without rebuilding it for each test. (This was not so easy for the searchACoin Integration test as we have a specific use case with the feature injected in a TabNavigator)

The trick I found is to call the function directly to avoid the error, it works and I have not found any impact (Maybe I have missed something?)
In conclusion, `MarketNavigator` is a React function component so I suggest that we stick with PascalCase.

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketDetail/MarketDetailCont.tsx
KVNLS: We do have an issue here: we recompute it 11x when we navigate to the details page. I will investigate

KVNLS: The issue we have is that we create a new object each time we call `flattenAccountsByCryptoCurrencyScreenSelector` => `{currencies: [currency.id]}`
So it breaks the createSelector cache mechanism. I fixed it by only using currency.id as a parameter.

```
export const flattenAccountsByCryptoCurrencyScreenSelector =
  (currency?: CryptoCurrency | TokenCurrency) => (state: State) => {
    if (!currency) return emptyArray;
    return flattenAccountsByCryptoCurrencySelector(state, {
      currencies: [currency.id],
    });
  };
  ```
  
I think that we have this kind of issue in a lot of other places.

Comment (KVNLS) [reply apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketDetail/MarketDetailCont.tsx:60]: Ok, so I fixed this issue by fixing our reselect usage.
Today we recreate a new object, or a new array for parameters we send to createSelector.
It's documented here on how to deal with multiple arguments: https://reselect.js.org/faq/#how-do-i-create-a-selector-that-takes-an-argument

I was not able to fix all selectors as I have this TS error https://github.com/reduxjs/reselect/issues/550.
Can be the subject of a dedicated improvement.

File: apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketList/components/BottomSection/BottomSection.tsx (hunk 3 above)
Comment (KVNLS) [reply apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketList/components/BottomSection/BottomSection.tsx:123]: Not really static as we depend on the translation:
`label: t("market.filters.order.topGainers")`
Static part has been extracted.

File: apps/ledger-live-mobile/src/newArch/features/Market/components/DeltaVariation/index.tsx [hunk 13]
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  const absDelta = Math.abs(delta);
+
+  const [color, ArrowIcon, sign] =
+    delta !== 0
+      ? delta > 0
+        ? ["success.c50", IconsLegacy.ArrowEvolutionUpMedium, "+"]
+        : ["error.c50", IconsLegacy.ArrowEvolutionDownMedium, "-"]
+      : ["neutral.c100", null, ""];
```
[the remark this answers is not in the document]
KVNLS: As you mentioned it's a readability issue, that we should enforce with eslint: https://eslint.org/docs/latest/rules/no-nested-ternary
This has no impact on performance

apps/ledger-live-mobile/src/newArch/features/Market/components/MarketRowItem/index.tsx
KVNLS: Yes I did not change it and imo we can keep this logic in the view as it's related to the display (we are between two worlds), it depends on the local + the function format the number for what we want to display here. We could do it in a container too, but as we don't have one here, looks like overkill to create one just for this.

apps/ledger-live-mobile/src/newArch/features/Market/utils/index.ts
KVNLS: I created this JIRA ticket to keep track of this issue: https://ledgerhq.atlassian.net/browse/LIVE-11233

mbertin-ledger (approved):
🚀

apps/ledger-live-mobile/src/newArch/features/Market/Navigator.tsx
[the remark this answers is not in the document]
gre: not really fan of aliasing anything, that how we create tech debt and more config (and lead to our present problem of having too much custom config)

but it's ok to have temp folders and just replace the imports later, no worries 👍

apps/ledger-live-mobile/src/newArch/features/Market/screens/MarketList/components/BottomSection/BottomSection.tsx
gre: (if it's not static, it may not be a bad idea to useMemo it, but it's probably a minor problem vs all other thing we can find as improvements)

gre (approved):

PR #5976, opened 2024-01-22
[LLD] - [LIVE-10489] - Add new carousel

sshmaxime:
### 📝 Description

- Add new carousel in our Design System 
- Added component on StoryBook
- Remove outdated and unused Carousel from DS

ps: Yes, we are depending on a RC(rc20 here). We could use version [7.1.0](https://www.npmjs.com/package/embla-carousel-react/v/7.1.0) but I was having some issue with it. The `rc` has been worked on for about a year and has become pretty stable so far. I'll follow this regularly and will be responsible for it.

https://github.com/LedgerHQ/ledger-live/assets/25589782/86d2e5c7-187a-4691-baa5-5fd94038f388

### ❓ Context

- **JIRA or GitHub link**: [LIVE-10489]

### ✅ Checklist

  - 

---

[LIVE-10489]: https://ledgerhq.atlassian.net/browse/LIVE-10489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/components/NewCarousel/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+const EmblaContainer = styled.div`
+  display: flex;
+`;
+
+const EmblaSlide = styled.div`
+  flex: 0 0 100%;
+  min-width: 0;
+`;
+
+const Carousel = ({ slides }: Props) => {
```
KVNLS, line 31: This component might belong to our Design System.
Why do we introduce a new component to do the same as https://react-ui-storybook.vercel.app/?path=/docs/cards-portfolio-carousel--docs?
If the design evolves we should update the previous component.

apps/ledger-live-desktop/src/renderer/components/NewCarousel/index.tsx
```diff
@@ -0,0 +1,67 @@
+import { Icons } from "@ledgerhq/react-ui";
+import useEmblaCarousel, { UseEmblaCarouselType } from "embla-carousel-react";
```
KVNLS, line 2: Why do we use `embla-carousel-react`? Is there any reason not to stick with `react-transition-group`?

sshmaxime: `react-transition-group` is an old and unmaintained library whereas `embla-carousel` is a maintained fully fledge customizable carousel. Btw, afaik for this specific thing we weren't using `react-transition-group` but `react-spring`. That being said, I think we shouldn't re-invent the wheel and stick with the most common/used package for basic things such as a carousel. Less thing to maintain on our side.

apps/ledger-live-desktop/src/renderer/components/NewCarousel/index.tsx
sshmaxime: Agreed. I was thinking of doing that in another PR if that's alright.

apps/ledger-live-desktop/src/renderer/components/NewCarousel/index.tsx
KVNLS: We will introduce a new way to manage the carousel without taking into consideration existing components. We have the Carousel component in our Design System that relies on `react-transition-group` for the slide effect: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ui/packages/react/src/components/transitions/TransitionSlide.tsx.
Is this component still used? Can we replace it with the new one you created? Otherwise, we will have developers that use the new one and some the other from the DS.
+ Regarding the choice of `embla-carousel` it needs to be discussed because if we all add libraries for specific needs the bundle will explode in size. We need to have a global strategy for how we deal with animations. As we don't only use slides we may benefit from one library to manage our animations such as `react-spring` if possible.
+ We depend on an RC here: `rc19`

apps/ledger-live-desktop/src/renderer/components/NewCarousel/index.tsx
sshmaxime: I added it to the DS.

libs/ui/packages/react/src/components/carousel/footer/pagination/bullets.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const bulletStyles = useBulletStyles();
+
+  const props = useSpring({
+    width: bulletStyles[type].width,
+    height: bulletStyles[type].height,
+    backgroundColor: bulletStyles[type].backgroundColor,
+    borderRadius: bulletStyles[type].borderRadius,
+    opacity: bulletStyles[type].opacity,
+    marginRight: bulletStyles[type].marginRight,
+  });
```
KVNLS, line 66:
```suggestion
  const props = useSpring(bulletStyles[type]);
```

libs/ui/packages/react/src/components/carousel/footer/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+export const getItemStatus = (itemIndex: number, activeIndex: number) => {
+  const itemDistanceFromActiveIndex = Math.abs(itemIndex - activeIndex);
+
+  if (itemDistanceFromActiveIndex === 0) {
+    return ItemStatus.active;
+  } else if (itemDistanceFromActiveIndex === 1) {
+    return ItemStatus.nearby;
+  } else if (itemDistanceFromActiveIndex === 2) {
+    return ItemStatus.far;
+  }
```
KVNLS, line 15:
```suggestion
  }
  if (itemDistanceFromActiveIndex === 1) {
    return ItemStatus.nearby;
  }
  if (itemDistanceFromActiveIndex === 2) {
    return ItemStatus.far;
  }
```
You don't need the else as you early return

KVNLS (changes requested):
Can we follow the same conventions we have in other components for this project, please?
Example:
Carousel
- index.tsx
- Carousel.stories.tsx
---Footer
------index.tsx
etc

As we moved the Carousel under Layout it must be reflected in the position of the source code.

Next time we should discuss prior to an implementation when we add a library because using a rc is a no go for me.

PR #6045, opened 2024-01-30
feat: first attempt to move bitcoin family in its own package

sprohaszka-ledger:
### 📝 Description

Attempt to move in its own package, bitcoin family.

__Side effects__
A `crypto-utils` package has also been created to share **bip32** functions across all other packages.

`DerivationMode` as been deduplicated.

Due to issue with LLD webworker, import directly bitcoin module in LLD. If we use the indirect import from LLC (`families/bitcoin/logic`), the [webworker](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/webworkers/workers/publicKeyTweakAdd.ts) never receive the `postMessage` from `publicTweakAddKey` [call](https://github.com/LedgerHQ/ledger-live/blob/daf0a0bc8e40bbceb0207dabf4ad0cfed1e85ba5/apps/ledger-live-desktop/src/renderer/families/bitcoin/live-common-setup.ts#L24).
Maybe it is due to the instanciation [type](https://developer.mozilla.org/fr/docs/Web/API/Worker/Worker#options) of the Worker (`module` instead of `classic`).

TO CHECK:
* Is it ok to use HashMap instead of object for some cases (ex: `speeds` in `getFeeItems.ts`)? ✅ 
* Some typing enforce the use of `Record` with a `Promise` as a value (ex: `crypto/base.ts`) ; is it ok? ✅
* With the type declaration of `coininfo` lib, the import has been modified (ex: `crypto/bitcoincash.ts`) ; is it ok? ✅

TODO:
- [x] Uncomment test with `fromAccountRaw` call

TO QA:
- [ ] Test serialization through LLD.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-8165]

### ✅ Checklist

  - ...

---

[LIVE-8165]: https://ledgerhq.atlassian.net/browse/LIVE-8165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-bitcoin/src/api/index.ts
```diff
@@ -55,10 +55,10 @@ export const getFeeItems = async (currency: CryptoCurrency): Promise<FeeItems> =
     const blockCount = parseInt(key, 10);
     if (blockCount === defaultBlockCount) defaultFeePerByte = feePerByte;
 
-    if (!Number.isNaN(blockCount) && !feePerByte.isNaN() && blockCount in speeds) {
+    if (!Number.isNaN(blockCount) && !feePerByte.isNaN() && speeds.has(blockCount)) {
```
hedi-edelbloute, line 58: Makes sense to use Map to me if key is supposed to be a number, we can do this update. Else it's a "sidegrade"

Just beware as speeds.has(blockCount) when blockCount is a number will return true but won't with a string compared to blockCount in speeds

image

libs/coin-bitcoin/src/wallet-btc/crypto/base.ts
```diff
@@ -29,8 +29,8 @@ export function fallbackValidateAddress(address: string): boolean {
 
 class Base implements ICrypto {
   network: any;
-  protected static bip32Cache = {}; // xpub + account + index to publicKey
-  public static addressCache = {}; // derivationMode + xpub + account + index to address
+  protected static bip32Cache: Record<string, Promise<BIP32>> = {}; // xpub + account + index to publicKey
```
hedi-edelbloute, line 32: We can discuss it on slack, where comes the need to change BIP32 => Promise<BIP32> ? :thinking: is it because of the sctricter typing environment you were talking about ?

libs/coin-bitcoin/src/wallet-btc/crypto/bitcoincash.ts
```diff
@@ -1,5 +1,5 @@
 import bchaddr from "bchaddrjs";
-import { toOutputScript } from "bitcoinjs-lib/src/address";
+import { address as btcAddress } from "bitcoinjs-lib";
```
hedi-edelbloute, line 2: we can see [here](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/index.js) that index acts as a [barrel](https://flaming.codes/en/posts/barrel-files-in-javascript/) so it should be fine editing the import. 

```
const address = require('./address');
exports.address = address;
```

libs/coin-bitcoin/src/wallet-btc/crypto/base.ts
sprohaszka-ledger: The need is related to the consistency of the code (ex: line 71).
Sometimes we stored Promises, sometimes directly BIP32.
We can switch to BIP32, I don't have a strong opinion on that.

libs/coin-bitcoin/src/api/index.ts
sprohaszka-ledger: > Just beware as speeds.has(blockCount) when blockCount is a number will return true but won't with a string compared to blockCount in speeds

It is expected to returns false, as `"1"` and `1` is not the same thing. On line 55, `blockCount` is declared as an int.

Keeping an object instead of a map in this case, lead to a type error on line 61. To resolve it, it will imply more code to change

libs/coin-bitcoin/src/wallet-btc/crypto/bitcoincash.ts
sprohaszka-ledger: I don't get it. Do you want to use `require` instead? `import "bicoinjs-lib/address"` seems not to work (I think the lib miss some declaration in it package.json to work properly).

libs/coin-bitcoin/src/api/index.ts
hedi-edelbloute: ok, using a map as you did is fine with me :+1:

libs/coin-bitcoin/src/wallet-btc/crypto/bitcoincash.ts
hedi-edelbloute: No :smile: the snippet is from bitcoinjs-lib, `index` exports `src/address`

libs/coin-bitcoin/src/wallet-btc/crypto/base.ts
hedi-edelbloute: `accountLevelP = Base.bip32Cache[keyAccount]; // it's stored as promise`

Is it what you are talking about ?

libs/coin-bitcoin/src/api/index.ts
hzheng-ledger, line 58: not directly related to the PR but it is a tech debt
this getFeeItems is not correct.
fees item can be an integer other than 1,3,6 (e.g. https://explorers.api.live.ledger.com/blockchain/v4/kmd/fees)
the "getFeeItems" is now only used in mock and duplicated with the logic in "bitcoin/getAccountNetworkInfo.ts"
we can removed completely this method and use same logic of "js-prepareTransaction.ts"

libs/coin-bitcoin/src/cli-transaction.ts
```diff
@@ -38,15 +39,22 @@ function inferTransactions(
   const feePerByte = new BigNumber(opts.feePerByte === undefined ? 1 : opts.feePerByte);
   return transactions.map(({ transaction }) => {
     invariant(transaction.family === "bitcoin", "bitcoin family");
+
+    const bitcoinStragegy = opts["bitcoin-pick-strategy"];
+    if (typeof bitcoinStragegy !== "string" && !(bitcoinStragegy in bitcoinPickingStrategy)) {
```
hzheng-ledger, line 44: I guess it should be "||" other than "&&"?
and maybe you can indicate the error message such as throw new Error("bitcoin picking strategy not found");

libs/coin-bitcoin/src/explorer.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getEnv } from "@ledgerhq/live-env";
+
+type LedgerExplorer = {
+  version: string;
+  id: string;
+  endpoint: string;
+};
+
+const findCurrencyExplorer = (currency: CryptoCurrency): LedgerExplorer | null | undefined => {
+  if (getEnv("SATSTACK") && currency.id === "bitcoin") {
```
grey220022, line 12: not directly related to the PR. it is a tech debt.
@haammar-ledger as we can't support SATSTACK since v4. 
Should we remove the code related to SATSTACK then?

libs/coin-bitcoin/src/js-getFeesForTransaction.ts
```diff
@@ -27,7 +27,7 @@ const getFeesForTransaction = async ({
       address: o.address,
       isChange: o.isChange,
       value: new BigNumber(o.value),
-      hash: undefined,
+      hash: "", //TODO: check that convert `undefined` to empty string is ok.
```
grey220022, line 30: I am not very sure about the impact. Can we just keep undefined for safety ?

libs/coin-bitcoin/src/js-getTransactionStatus.ts
```diff
@@ -72,12 +72,10 @@ const getTransactionStatus = async (a: Account, t: Transaction): Promise<Transac
     );
   }
 
-  const sumOfInputs = (txInputs || []).reduce(
-    (sum, input) => sum.plus(input.value),
-    new BigNumber(0),
-  );
+  //TODO: check if it is ok in case of `undefined` value, a zero is ok
```
hzheng-ledger, line 75: I think it is OK. 
"undefined" is not supposed to occur in input.value unless there's a bug. We can consider “undefined” as 0

libs/coin-bitcoin/src/js-getFeesForTransaction.ts
hzheng-ledger: I just checked. It is very safe. the return object is only used  to calculate the fees and the fees are not related to the hash. this hash field is useless in this case.

libs/coin-bitcoin/src/js-synchronisation.ts
```diff
@@ -57,14 +55,14 @@ const fromWalletUtxo = (utxo: WalletOutput, changeAddresses: Set<string>): Bitco
 // wallet-btc limitation: returns all transactions twice (for each side of the tx)
 // so we need to deduplicate them...
 const deduplicateOperations = (operations: (Operation | undefined)[]): Operation[] => {
-  const seen = {};
+  const seen = new Map();
```
hzheng-ledger, line 58: very minor issue:
you can use
```
const seen = new Set();
```
and
for the check:
```
seen.has(operation.id)
```

libs/coin-bitcoin/src/mockBtc.ts
```diff
@@ -1,18 +1,19 @@
...
+  CreateTransaction,
+  SignerTransaction,
+} from "./signer";
 
-class MockBtc extends Btc {
-  constructor() {
-    super({ transport: new Transport() });
-  }
-
+class MockBtc implements BitcoinSigner {
```
hzheng-ledger, line 11: maybe name it as "MockBtcSigner"

libs/coin-bitcoin/src/mockBtc.ts
```diff
@@ -1,18 +1,19 @@
...
   // eslint-disable-next-line class-methods-use-this
   async getWalletPublicKey(
     path: string,
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    opts: { verify?: boolean; format?: AddressFormat },
+    _opts: { verify?: boolean; format?: AddressFormat },
```
hzheng-ledger, line 16: as we've hard coded the return "chaincode" and "publickey" in the mock method. this "_opts" parameter can be deleted and it is not used.

libs/coin-bitcoin/src/mockBtc.ts
```diff
@@ -36,10 +37,30 @@ class MockBtc extends Btc {
     }
   }
 
-  // eslint-disable-next-line class-methods-use-this,@typescript-eslint/no-unused-vars
-  async createPaymentTransaction(arg: CreateTransactionArg) {
+  async createPaymentTransaction(arg: CreateTransaction) {
     return hash(arg);
   }
+
```
hzheng-ledger, line 43: This mock class has "getWalletXpub", "SignMessage", "SplitTransaction" because you inherit the "BitcoinSigner" interface. But these methods are useless.
I suggest you can set these method as "optional" in "BitcoinSigner" interface. So you don't have to write such methods in mockBtc.ts
About optional method in typescript. I find https://compiletab.com/optional-properties-in-interfaces-typescript/

libs/ledgerjs/packages/crypto-utils/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@ledgerhq/crypto-utils",
```
gre, line 2: Is there a reason we make this a dedicated library, why not keeping this bip32 in hw-app-btc?  is it useful for another thing than bitcoin? looking at the code it is specific to bitcoin and xpub concepts.

If we still want to make this a dedicated library, i think we shouldn't recreate libs that have "common", "core" or "utils" in their name but directly tells what is the current scope of that lib.  at the moment the scope is a bit weird, what do you think?

libs/ledgerjs/packages/types-live/src/index.ts
```diff
@@ -15,5 +15,5 @@ export * from "./storyly";
 export * from "./chainwatch";
 export * from "./messages";
 export * from "./cexDeposit";
-export * from "./storyly";
```
gre, line 18: what happened here?

gre (changes requested):
i have reviewed the part related to live-hub ownership, as general feedback: there are many parts touched that don't strictly relates to moving bitcoin to a coin-bitcoin family lib (for instance the typing of DerivationMode, no idea why it's in this PR really, i would recommend to split things into more PRs in the future)

i have mainly one feedback on the need to make a new library that have the word "utils" in it, we need to avoid it as much as possible, that names means we don't really know what is the scope of the lib, it's better to scope our library names with a bit more meaning
(i know naming is hard, i wish live-common was never a thing 😆 )

libs/coin-bitcoin/src/signer.ts
```diff
@@ -0,0 +1,82 @@
...
+  bitcoinAddress: string;
+  chainCode: string;
+};
+export type BitcoinSignature = {
+  v: number;
+  r: string;
+  s: string;
+};
+export type AddressFormat = "legacy" | "p2sh" | "bech32" | "bech32m" | "cashaddr";
+
```
hzheng-ledger, line 15: First,
I guess that the methods in BitcoinSigner come from ledgerjs/packages/hw-app-btc/src/Btc.ts

So, do you want the "Btc" class in "Btc.ts" inherits also this interface?

Second,
we are thinking about update the Btc.ts with a new interface in the future and the new interface is here:

https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/bitcoin_client_js/src/lib/appClient.ts

because the APIs are more powerful and it supports the feature "wallet policy"

but we need to keep the old interface for altcoin because of the compatibility of nano app.
the methods of new interface are no compatible with the old one.
What do you think about this change in the future

libs/coin-bitcoin/types/coininfo.d.ts
```diff
@@ -0,0 +1,100 @@
...
+  | "nmc"
+  | "peercoin"
+  | "qtum"
+  | "rvn"
+  | "rdd"
+  | "viacoin"
+  | "vertcoin"
+  | "x42"
+  | "zcash";
+declare const coininfo: Record<CoinName, Network> & { bitcoin: BitcoinNetwork };
```
hzheng-ledger, line 99: very good initiative to do that.
just a question, why do you separate "bitcoin" and BitcoinNetwork aside?
I think we can add "bitcoin" into the CoinName list

libs/ledgerjs/packages/types-live/src/index.ts
sprohaszka-ledger: Just look at line 14

libs/ledgerjs/packages/crypto-utils/package.json
sprohaszka-ledger: I am open to any feedback about the name of this package.

The purpose is to remove the hard dependency of "coin-bitcoin" to "hw-app-btc". For the time being I simply put the minimal code required to be shared, but we can add more.

I think we need to stop melting "blockhain specific logic", "hardware logic" and "cryptographic algorithm".
This last category is a tool for the others. But the first two have to be independent.

sprohaszka-ledger:
> i have reviewed the part related to live-hub ownership, as general feedback: there are many parts touched that don't strictly relates to moving bitcoin to a coin-bitcoin family lib and it's maybe why this PR have pinged other teams than strictly blockchain team (for instance the typing of DerivationMode, no idea why it's in this PR really, i would recommend to split things into more PRs in the future)
> 
> i have mainly one feedback on the need to make a new library that have the word "utils" in it, we need to avoid it as much as possible, that names means we don't really know what is the scope of the lib, it's better to scope our library names with a bit more meaning (i know naming is hard, i wish live-common was never a thing 😆 )

The modification I made are unfortunately necessary to create a real independent "bitcoin-module". DerivationMode was defined twice, and used for the same purpose at different part of the code. I wish I could have done something lighter.

For the name of the package, I knew I would have comments about it so I wanted to resolve it after the debates ;)
We can call it `crypto-algo` or `bip-algo` if we want to specialiaze it. WDYT @gre ?

libs/coin-bitcoin/src/js-getFeesForTransaction.ts
sprohaszka-ledger: Thanks

libs/coin-bitcoin/src/mockBtc.ts
sprohaszka-ledger: As seen together, we need to define those methods to conform to the interface definition (a.k.a. `BitcoinSigner`)

libs/coin-bitcoin/src/signer.ts
sprohaszka-ledger: For the time being, I prefer to keep a duplicate definition of the Signer, as it is for the other coins.

I propose that we will see later together how we want to manage the separation between BtcOld and BtcNew.

libs/coin-bitcoin/types/coininfo.d.ts
sprohaszka-ledger: I separate it to ease the migration of the coin module.
In the current code, we suppose that we have a testnet available for bitcoin, whereas for the other coin it is an optional property.

libs/coin-bitcoin/src/api/index.ts
sprohaszka-ledger: Sorry, but it breaks some test logic ("libcore" account Need Sync error). I prefer not refacto the test and rely on them for this PR.

hedi-edelbloute (dismissed):

hzheng-ledger (approved):
looks good to me

hedi-edelbloute (approved):

gre (approved):
ok for live-hub own scope

PR #6053, opened 2024-01-31
[coin-config] Make getValueBykey backward compatible

lvndry:
### 📝 Description

The schema for a config is shipped for a specific version of the live, however the values in remote (firebase or else) are not dependent on a specific version. So we make sure that the configuration is backward compatible so that if a field is renamed or deleted in firebase, Ledger Live can fallback on the default value.

### ❓ Context

- **JIRA or GitHub link**: 

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/live-config/package.json
+++ b/libs/live-config/package.json
@@ -27,10 +27,11 @@
     "test": "jest"
   },
   "dependencies": {
-    "mergician": "^2.0.0"
+    "lodash": "^4.17.21"
   },
   "devDependencies": {
     "@types/jest": "^29.5.4",
+    "@types/lodash": "^4.14.202",
     "@types/node": "^20.2.5",
     "jest": "^28.1.1",
     "jest-environment-jsdom": "28",
--- a/libs/live-config/package.json
+++ b/libs/live-config/package.json
@@ -27,6 +27,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "mergician": "^2.0.0"
   },
   "devDependencies": {
     "@types/jest": "^29.5.4",
--- a/libs/live-config/src/LiveConfig.ts
+++ b/libs/live-config/src/LiveConfig.ts
@@ -66,7 +66,7 @@ export class LiveConfig {
       // this is for backward compatiblity, a value could be renamed or deleted in a remote provider
       // but the default value will not change for a given version of Ledger Live
       // so we make sure that there's always a fallback in a case the value changed remotely
-      return merge(LiveConfig.instance.config[key].default as object, providerValue as object);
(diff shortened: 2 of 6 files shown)


libs/live-config/src/LiveConfig.ts
```diff
@@ -54,7 +55,21 @@ export class LiveConfig {
...
+
+    const providerValue = LiveConfig.instance.provider.getValueByKey(
+      key,
+      LiveConfig.instance.config[key],
+    );
+
+    if (LiveConfig.instance.config[key].type === "object") {
+      // we spread the default values first and then the values from the provider
+      // this is for backward compatiblity, a value could be renamed or deleted in a remote provider
+      // but the devault value will not change for a given version of Ledger Live
```
hedi-edelbloute, line 67: default

libs/live-config/package.json
```diff
@@ -27,6 +27,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "mergician": "^2.0.0"
```
hedi-edelbloute, line 30: WDYT about using lodash instead ? It is already in the codebase and is 20 000x more downloaded per week

https://www.npmjs.com/package/lodash
https://www.npmjs.com/package/mergician

hzheng-ledger:
Looks good to me in generally.
but
1. add changeset plz.
2. I think you should add return type for the "getValueByKey" method in /libs/live-config/src/providers/firebaseRemoteConfig/index.ts . otherwise, there is a build error there.

hzheng-ledger (changes requested):
have a look at my comments plz. Thanks

hedi-edelbloute (commented):
As huiqi said missing changeset, rest lgtm :+1:

hedi-edelbloute (approved):

hzheng-ledger (dismissed):
looks good to me

libs/live-config/src/LiveConfig.ts
```diff
@@ -1,4 +1,5 @@
 import { Provider } from "./providers";
+import _ from "lodash";
```
gre, line 2: please do not import the whole lodash.
i think we never did it in the whole project (if we do it's a mistake) and then instead we always used the functional import.

for instance

```
import merge from "lodash/merge"
```

gre (changes requested):

libs/live-config/src/LiveConfig.ts
gre: https://ledgerhq.atlassian.net/browse/LIVE-11142
(unfortunately it seems we have lost track of this good practice over the last months, i created another task as a side topic on top of the fact it would be great if you could also solve this one)

PR #6112, opened 2024-02-07
bugfix(LIVE-11143): fix useless renders and render issues swap live app

lpaquet-ledger:
### 📝 Description

Remove useless renders from LL
verify if swap live state hook props before creating a new cacheKey or it would rerender infinitely 

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-11143

### ✅ Checklist

  - ...

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/hooks/swap-migrations/useSwapLiveAppHook.tsx
+++ b/apps/ledger-live-desktop/src/renderer/hooks/swap-migrations/useSwapLiveAppHook.tsx
@@ -1,10 +1,15 @@
 import { v4 } from "uuid";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
+import { isEqual } from "lodash";
 import { useSelector } from "react-redux";
 import { accountToWalletAPIAccount } from "@ledgerhq/live-common/wallet-api/converters";
 import { getProviderName } from "@ledgerhq/live-common/exchange/swap/utils/index";
 import { SwapTransactionType } from "@ledgerhq/live-common/exchange/swap/types";
-import { SwapProps, SwapWebManifestIDs } from "~/renderer/screens/exchange/Swap2/Form/SwapWebView";
+import {
+  SwapProps,
+  SwapWebManifestIDs,
+  SwapWebProps,
+} from "~/renderer/screens/exchange/Swap2/Form/SwapWebView";
 import { rateSelector } from "~/renderer/actions/swap";
 import { getEnv } from "@ledgerhq/live-env";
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/renderer/hooks/swap-migrations/useSwapLiveAppHook.tsx
```diff
@@ -51,16 +59,20 @@ export const useSwapLiveAppHook = (props: UseSwapLiveAppHookProps) => {
...
-        cacheKey: v4(),
         error: !!swapError,
         loading,
         providerRedirectURL,
         swapApiBase: SWAP_API_BASE,
-      });
+      };
+
+      if (!isEqual(newSwapWebProps, swapWebProps)) {
+        setSwapWebProps(newSwapWebProps);
```
bethylogism, line 73: setting state inside a useEffect that is watching that state could create an infinite loop of re-rendering.

The new `swapWebProps` state in this hook seems to mirror whatever the state is that is being updated by `updateSwapWebProps`. So do we not have access to that SwapWebProps state?

I wonder if we could store SwapWebProps in a ref, if we must do a comparison? But it strikes me that we should only need to watch for the `cacheKey` changing. Or update in a more selective reducer, piece by piece, instead of lumping all SwapWebProps into a single object?

apps/ledger-live-desktop/src/renderer/hooks/swap-migrations/useSwapLiveAppHook.tsx
```diff
@@ -73,5 +85,6 @@ export const useSwapLiveAppHook = (props: UseSwapLiveAppHookProps) => {
     swapTransaction.bridgePending,
     exchangeRatesState.status,
     updateSwapWebProps,
+    swapWebProps,
```
bethylogism, line 88: https://stackoverflow.com/questions/53715465/can-i-set-state-inside-a-useeffect-hook

apps/ledger-live-desktop/src/renderer/hooks/swap-migrations/useSwapLiveAppHook.tsx
lpaquet-ledger: The cacheKey updates on every render that's what war creating the problem since `updateSwapWebProps` is also a setter for a parent component
I can put `swapWebProps` in a ref and compare them in the same way

gre (approved):

CremaFR (approved):

PR #6124, opened 2024-02-08
Smartling translations are updated for PR#5986 from feat/LIVE-10541-add-translated-ts-discover

ldg-smartling-sa:
Please review this translation file update. Changes in the source file have been applied. The relevant Smartling translation job may be empty - https://dashboard.smartling.com/app/accounts/d32c14dd/account-jobs/be90da906:a5duchyxixmy


Justkant:
Not sure we actually need this as we don't have i18n modifications anymore in the source branch

PR #6133, opened 2024-02-08
[Bugfix/Stacks] Fix Account sync error for imported accounts 

lawRathod:
### 📝 Description

Fix for account sync error on stacks caused by import of accounts on ledger live using `accountId` string. The address is now derived from pubkey which is always stored in the accountId string and acts as seedId. 

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10786

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
+++ b/libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
@@ -5,6 +5,7 @@ import {
   makeUnsignedSTXTokenTransfer,
   UnsignedTokenTransferOptions,
   createMessageSignature,
+  getAddressFromPublicKey,
 } from "@stacks/transactions";
 
 import { GetAccountShape, AccountShapeInfo } from "../../../../bridge/jsHelpers";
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
```diff
@@ -131,18 +133,22 @@ export const mapTxToOps =
...
   const accountId = encodeAccountId({
     type: "js",
     version: "2",
     currencyId: currency.id,
-    xpubOrAddress: publicKey,
+    xpubOrAddress: pubKey,
     derivationMode,
   });
 
+  const address = getAddressFromPublicKey(pubKey);
```
hedi-edelbloute, line 150: So this is the real diff here, address comes from the pubkey instead of AccountShapeInfo

hedi-edelbloute (approved):

haammar-ledger (approved):

libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
lawRathod: Sorry for the late reply @hedi-edelbloute!! You understand correct, the above is the core change. We in past always assumed that the xpub/Address will always be available but it seems in cases where accounts are restored using accountId only one can be found. This eliminates the need for both to be present and will now get the address from pubkey which is stored in the accountId string.

PR #6138, opened 2024-02-08
accountScreenSelector handling parentAccount equaling account

Wozacosta:
## What was wrong:

In  **libs/ledger-live-common/src/wallet-api/logic.ts**, `getParentAccount` changed :

- [Before](https://github.com/LedgerHQ/ledger-live/commit/479c6e84ed97721b1d6176414a8dc6d23b267146), it was a function defined in the file as such

```js
function getParentAccount(account: AccountLike, fromAccounts: AccountLike[]): Account | undefined {
  return isTokenAccount(account)
    ? (fromAccounts.find(a => a.id === account.parentId) as Account)
    : undefined;
}
```

- Now, we import it from **libs/coin-framework/src/account/helpers.ts** where it's defined as such:

```js
export function getParentAccount(account: AccountLike, accounts: AccountLike[]): Account {
  switch (account.type) {
    case "Account":
      return account;
    case "TokenAccount":
    case "ChildAccount": {
      const parentAccount = accounts.find(a => a.id == account.parentId);
      if (!parentAccount) {
        throw new Error("No 'parentAccount' account provided for token account");
      }

      return parentAccount as Account;
    }
  }
}
```

Now, if we pass an account of type `Account` to `getParentAccount`, we get back the same Account.

In case of initial swap transaction (MATIC to 1inch), we have account = parentAccount = polygon main account, before we would have had parentAccount as `undefined`, now we `parentAccount == account`

Downstream from there, the difference means that in **apps/ledger-live-mobile/src/reducers/accounts.ts** In the **accountScreenSelector** function: 

having a `parentAccount` that is not `undefined` means that **the account search is done only on the parentAccount subaccounts.**

It means that now (before the bugfix introduced by this PR), we would search the account amongst its subaccounts, and getting undefined. Thus then failing the invariant check.

### 📝 Description

_For bugfixes, you can explain the previous behavior and how it was fixed._

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/jira/software/c/projects/LIVE/boards/623?selectedIssue=LIVE-11191

### ✅ Checklist

  - ...

---

### DIFF
--- a/.changeset/lazy-spiders-confess.md
+++ b/.changeset/lazy-spiders-confess.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+accountScreenSelector handling parentAccount being the same as the account arg
--- a/apps/ledger-live-mobile/src/reducers/accounts.ts
+++ b/apps/ledger-live-mobile/src/reducers/accounts.ts
@@ -260,6 +260,10 @@ export const accountScreenSelector =
       : null;
     let account = route?.params?.account;
 
+    if (accountId === parentId && parentAccount) {
+      account = parentAccount;
+    }
+
     if (!account) {
       if (parentAccount) {
         const { subAccounts } = parentAccount;


gre:
this is weird. we shouldn't have to do this, we are supposed to be in a strict convention where it's never possible to have `parentAccount == account`

when an account `a` is an Account (and not a TokenAccount) it is supposed to always be `parentAccount=null , account=a`

at some point we broke this general convention.

quoting the documentation of the types 'Account':

```
// CONVENTION:
// a SubAccount is living inside an Account but is not an entity on its own,
// therefore, there is no .parentAccount in it, which will means you will need to always have a tuple of (parentAccount, account)
// we will use the naming (parentAccount, account) everywhere because a sub account is not enough and you need the full context with this tuple.
// These are two valid examples:
// I'm inside a ZRX token account of Ethereum 1: { parentAccount: Ethereum 1, account: ZRX }
// I'm just inside the Ethereum 1: { account: Ethereum 1, parentAccount: undefined }
// "account" is the primary account that you use/select/view. It is a `AccountLike`.
// "parentAccount", if available, is the contextual account. It is a `?Account`.
```

Wozacosta:
Thanks for taking a look @gre, this clears things up.
So the `getParentAccount` function https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-framework/src/account/helpers.ts#L374-L387 would be better named `getParentAccountOrAccount`. 
And a proper `getParentAccount` function would look like this then: 

```diff
export function getParentAccount(account: AccountLike, accounts: AccountLike[]): Account {
  switch (account.type) {
    case "Account":
-      return account;
+     return null;
    case "TokenAccount":
    case "ChildAccount": {
      const parentAccount = accounts.find(a => a.id == account.parentId);
      if (!parentAccount) {
        throw new Error("No 'parentAccount' account provided for token account");
      }

      return parentAccount as Account;
    }
  }
}
```

Justkant (commented):
LGTM just the question for the triple equal

Justkant (approved):

cgrellard-ledger (approved):

PR #6149, opened 2024-02-09
[Bugfix/LIVE-11195]: Tracking issue for pair scalability

kallen-ledger:
### 📝 Description

Tracking issue for pair scalability
- use globals for ff analytics identity flags
- use updateIdentify to set with rest of context

### ❓ Context

[- **JIRA or GitHub link**: ](https://ledgerhq.atlassian.net/browse/LIVE-11195)

### ✅ Checklist

  - ... No QA, just MixPanel monitoring

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/analytics/segment.ts
+++ b/apps/ledger-live-desktop/src/renderer/analytics/segment.ts
@@ -55,33 +55,33 @@ export function setAnalyticsFeatureFlagMethod(method: typeof analyticsFeatureFla
   analyticsFeatureFlagMethod = method;
 }
 
-let isBatch1Enabled: boolean,
-  isBatch2Enabled: boolean,
-  isBatch3Enabled: boolean,
-  ptxEarnEnabled: boolean,
-  stakingProvidersEnabled: number | string;
-
-const getFeatureFlagProperties = () => {
+const getPtxAttributes = () => {
   if (!analyticsFeatureFlagMethod) return {};
   const ptxEarnFeatureFlag = analyticsFeatureFlagMethod("ptxEarn");
   const fetchAdditionalCoins = analyticsFeatureFlagMethod("fetchAdditionalCoins");
   const stakingProviders = analyticsFeatureFlagMethod("ethStakingProviders");
 
-  isBatch1Enabled = !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 1;
-  isBatch2Enabled = !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 2;
-  isBatch3Enabled = !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 3;
-  ptxEarnEnabled = !!ptxEarnFeatureFlag?.enabled;
-  stakingProvidersEnabled =
+  const ptxEarnEnabled: boolean = !!ptxEarnFeatureFlag?.enabled;
+  const isBatch1Enabled: boolean =
+    !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 1;
+  const isBatch2Enabled: boolean =
+    !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 2;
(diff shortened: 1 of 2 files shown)


beths-ledger (commented):
are you adding an analytics trait to update the user, or do you want it to appear on every event? If the latter then adding it to the `extraProperties` in the `segment.ts` file will work. Let's quickly discuss if possible as I have an overlapping PR that uses some of this same logics

apps/ledger-live-desktop/src/renderer/analytics/segment.ts
```diff
@@ -55,34 +55,22 @@ export function setAnalyticsFeatureFlagMethod(method: typeof analyticsFeatureFla
   analyticsFeatureFlagMethod = method;
 }
 
+let isBatch1Enabled: boolean,
+  isBatch2Enabled: boolean,
+  isBatch3Enabled: boolean,
+  ptxEarnEnabled: boolean;
```
gre, line 61: it's weird to have a "global" property on ALL events where this only concern one part of our app?
have we considered moving this from a "global" to a "local" approach? (like only adding the metadata on event specific to the feature related to that feature flag 🤔 )

beths-ledger: this seems like a sensible approach to me in general. However whilst implementing my own AB test analytics, Amine needs more data on the test groups general behaviour in order to get historical data to make sure the ab test groups are unbiased, so requested (at least in my case) that the data be on every event. I think it is AB test specific, and can be removed or targeted once the flag is 100%

beths-ledger (dismissed):
let's monitor this one together once it's in

apps/ledger-live-desktop/src/renderer/analytics/segment.ts
```diff
@@ -55,38 +55,30 @@ export function setAnalyticsFeatureFlagMethod(method: typeof analyticsFeatureFla
...
+  isBatch1Enabled = !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 1;
+  isBatch2Enabled = !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 2;
+  isBatch3Enabled = !!fetchAdditionalCoins?.enabled && fetchAdditionalCoins?.params?.batch === 3;
+  ptxEarnEnabled = !!ptxEarnFeatureFlag?.enabled;
+  stakingProvidersEnabled =
+    !!stakingProviders?.enabled &&
+    stakingProviders?.params &&
+    stakingProviders?.params?.listProvider?.length > 0
+      ? stakingProviders?.params?.listProvider.length
+      : "flag not loaded";
```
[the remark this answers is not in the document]
beths-ledger: love it great idea, much preferable to global vars

apps/ledger-live-desktop/src/renderer/analytics/segment.ts
```diff
@@ -236,6 +236,8 @@ export const updateIdentify = async () => {
     context: getContext(),
   });
 };
+/** Ensure PTX flag attributes are set as soon as feature flags load */
+runOnceWhen(() => !!analyticsFeatureFlagMethod && !!getAnalytics(), updateIdentify);
```
beths-ledger, line 240: I'm not sure this is even necessary since it no longer has the extra properties being added to the user traits, and the extraProperties re-run all the time. But there's not much harm to leave it in since it only runs once.

beths-ledger (approved):

apps/ledger-live-desktop/src/renderer/analytics/segment.ts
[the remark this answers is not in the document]
beths-ledger: in the process of updating this pr i realised that the `getFeatureFlagProperties` fn no longer adds traits to the user, so it wasn't actually doing anything with them except calling `updateIdentify`. So I've keep the `callOnce` of updateIdentify after the flags are definitely loaded, since that seems harmless, but added. the traits back to the extra properties. The `getPtxAttributes` fn may be useful again if some of them do need to be added to the user traits.

KVNLS (approved):
LGTM!

PR #6154, opened 2024-02-12
Add SELL_NG

sprohaszka-ledger:
### 📝 Description

* Update "Platform" CompleteSell function to Swap last evolution.
* Put together exchange "Providers" in a dedicated directory.
* Set `hw-app-exchange` error converter function in a share file between Swap and Sell&Fund
* Update startExchange in customHandler in order to require `providerId` in Sell
* Remove "*_NG" exchange type, as it will be handled entirely by Ledger Live code base

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11570] [LIVE-11679]

### ✅ Checklist

  - ...

---

[LIVE-9022]: https://ledgerhq.atlassian.net/browse/LIVE-9022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-11570]: https://ledgerhq.atlassian.net/browse/LIVE-11570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-11679]: https://ledgerhq.atlassian.net/browse/LIVE-11679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/exchange/providers/swap.ts
+++ b/libs/ledger-live-common/src/exchange/providers/swap.ts
@@ -8,7 +8,7 @@ export type SwapProviderConfig = {
 
 type CEXProviderConfig = ExchangeProviderNameAndSignature & SwapProviderConfig & { type: "CEX" };
 type DEXProviderConfig = SwapProviderConfig & { type: "DEX" };
-type ProviderConfig = CEXProviderConfig | DEXProviderConfig;
+export type ProviderConfig = CEXProviderConfig | DEXProviderConfig;
 
 const swapProviders: Record<string, ProviderConfig> = {
   changelly: {
--- a/libs/ledger-live-common/src/exchange/providers/swap.ts
+++ b/libs/ledger-live-common/src/exchange/providers/swap.ts
@@ -1,0 +1,93 @@
+import { ExchangeProviderNameAndSignature } from ".";
+import { isIntegrationTestEnv } from "../swap/utils/isIntegrationTestEnv";
+
+export type SwapProviderConfig = {
+  needsKYC: boolean;
+  needsBearerToken: boolean;
+};
+
+type CEXProviderConfig = ExchangeProviderNameAndSignature & SwapProviderConfig & { type: "CEX" };
+type DEXProviderConfig = SwapProviderConfig & { type: "DEX" };
+type ProviderConfig = CEXProviderConfig | DEXProviderConfig;
+
+const swapProviders: Record<string, ProviderConfig> = {
+  changelly: {
+    name: "Changelly",
+    publicKey: {
+      curve: "secp256k1",
+      data: Buffer.from(
+        "0480d7c0d3a9183597395f58dda05999328da6f18fabd5cda0aff8e8e3fc633436a2dbf48ecb23d40df7c3c7d3e774b77b4b5df0e9f7e08cf1cdf2dba788eb085b",
+        "hex",
+      ),
+    },
+    signature: Buffer.from(
+      "3045022100e73339e5071b5d232e8cacecbd7c118c919122a43f8abb8b2062d4bfcd58274e022050b11605d8b7e199f791266146227c43fd11d7645b1d881f705a2f8841d21de5",
+      "hex",
+    ),
+    needsKYC: false,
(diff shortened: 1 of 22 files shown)


libs/ledger-live-common/src/exchange/providers/swap.ts
```diff
@@ -0,0 +1,93 @@
...
+  },
+  oneinch: {
+    type: "DEX",
+    needsKYC: false,
+    needsBearerToken: false,
+  },
+  paraswap: {
+    type: "DEX",
+    needsKYC: false,
+    needsBearerToken: false,
```
sarneijim, line 74: I believe these two attributes are deprecated. Do we employ the same signature for selling and swapping? Does it make more sense to differentiate between CEX and DEX rather than distinguishing between swapping and selling?

```
getDexProvider

**Review by sarneijim — APPROVED** [review]

File: libs/ledger-live-common/src/exchange/providers/swap.ts (hunk 1 above)
Comment (sprohaszka-ledger) [reply libs/ledger-live-common/src/exchange/providers/swap.ts:74]: I didn't touch the logic in swap. I think the legacy code still using it.
Maybe for a next small refacto to revamp this part.

File: libs/ledger-live-common/src/wallet-api/Exchange/server.ts [hunk 2]
```diff
@@ -97,78 +104,34 @@ export const handlers = ({
...
-        const fromAccount = accounts.find(acc => acc.id === realFromAccountId);
+      let exchangeParams: ExchangeStartParamsUiRequest = {
+        exchangeType: params.exchangeType,
+      };
 
-        if (!fromAccount) {
-          throw new ServerError(createAccountNotFound(params.fromAccountId));
+      switch (params.exchangeType) {
+        case "SWAP": {
+          exchangeParams = extractSwapStartParam(params as ExchangeStartSwapParams, accounts);
```
Justkant, line 114: I think we could remove the `as` keyword here by using a guard function in the case check that takes `params` as a parameter to use the return type `_params is ExchangeStartSwapParams` (the underscore is needed if we don't use params in the guard function)

Now that I think about it, another and maybe simpler solution might just be to use a `if` instead of the `switch` so typescript can type narrow correctly

libs/ledger-live-common/src/wallet-api/Exchange/server.ts
```diff
@@ -53,13 +60,13 @@ export type CompleteExchangeUiRequest = {
   amountExpectedTo?: number;
 };
 
-export type ExchangeStartParamsUiRequest = {
-  exchangeType: "FUND" | "SELL" | "SWAP" | "FUND_NG" | "SELL_NG" | "SWAP_NG";
+type ExchangeStartParamsUiRequest = {
+  exchangeType: "FUND" | "SELL" | "SWAP";
   provider?: string;
   exchange?: Exchange;
 };
```
Justkant, line 67: I'm not sure if it makes sense to change this type like that but it looks like we only want the provider and exchange param when `exchangeType !== "FUND"`
```suggestion
type ExchangeStartParamsUiRequest = {
  exchangeType: "FUND";
} & {
  exchangeType: "SELL" | "SWAP";
  provider: string;
  exchange: Exchange;
};
```

Justkant (commented):
LGTM just some small details around types and some comments that might be cleaned up

libs/ledger-live-common/src/wallet-api/Exchange/server.ts
sprohaszka-ledger: That's a shame Typescript is not able to type narrow on `switch` :(

Justkant (approved):

Justkant (approved):

PR #6166, opened 2024-02-12
evolve live-countervalues-react to include Market Performance requirements

gre:
### 📝 Description

Rework live-countervalues stack to include Market Performance requirements.

***WORK IN PROGRESS – will be documented***

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-11259

### ✅ Checklist

  - ...

---


libs/live-countervalues-react/src/index.tsx
```diff
@@ -90,6 +151,15 @@ export function useTrackingPairForAccounts(
   return useMemo(() => c.pairs, [c.hash]);
 }
 
+export function CountervaluesMarketcap({ children }: { children: React.ReactNode }): ReactElement {
+  const marketcapIds = useMarketcap();
```
mcayuelas-ledger, line 155: As it is a hook, why just not use it when it's needed instead of doing a Context that will do a lot of rerender when ids are refetched?

I'm maybe wrong

gre: @mcayuelas-ledger I tried to cover this part in

```
<CountervaluesMarketcap> context is introduced to be able to share the same cache of fetching the available currencies of the API and sorted by marketcap. It is not required to depends on it, but it is strongly recommended to migrate to it. the older way to do currenciesByMarketcap is still functional but deprecated in favor of this new paradigm that we will progressively migrate to.
```

Here is what I understand of how to use React Context so far:

it's better to re-render a context once, than having N parts of the app that independently load again the data.

in theory, what you explained should not happen if we properly memoize components. here is an example: let's say we have this tree of component:

A -> B -> C.

A renders B which renders C (from parent to children).

when we make this 

```
<CountervaluesMarketcap>
  <A />
  </CountervaluesMarketcap>
```

as soon as A is properly memo ized, it will not re-render.

and when C will need to `useContextMarketcap()` (which is the context version of the hook, hook that share the same state of marketcap) it will indeed re-render, like if it was a direct usage of the useMarketcap() one. but A and B will not re-render in that case.

normally our app is using react router which itself will memoize the pages. we probably will want to double confirm that it's the case, because it will benefit all our existing contexts usage.

mcayuelas-ledger: Indeed :) 
Sounds good !

mcayuelas-ledger (approved):

gre:
the codeowners have assigned @LedgerHQ/live-blockchain-support  but actually it's not really impacted, it's only due to fact we have set codeowners of "cli" to be blockchain team, i think we should rethink this in future, maybe actually remove codeowners completely on the CLI? wdyt @aboissiere-ledger ?

aboissiere-ledger:
@gre We actually discuss the CLI ownership with Hakim if I remember well. Blockchain should not own the integrality of the CLI, we should be more granular, command by command maybe.

gre:
@aboissiere-ledger ok thanks. i'll update codeowners in another PR. merging this one for now because team blockchain is not concerned.

PR #6205, opened 2024-02-15
feat: update UI sign transaction [LIVE-10708]

RamyEB:
### 📝 Description

Update sign transaction UI 

Here's a visual recap of what's the confirm screen : 

SCR-20240228-lybv

SCR-20240228-lylw

SCR-20240228-mbxo

SCR-20240228-mexj

SCR-20240228-mfrx

SCR-20240228-mgrk

### ❓ Context

- **JIRA or GitHub link**: [LIVE-10708]

### ✅ Checklist

  - [x] Wallet-API sign transaction flow

---

[LIVE-10708]: https://ledgerhq.atlassian.net/browse/LIVE-10708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -192,79 +193,88 @@ const TransactionConfirm = ({
...
+              <Trans
+                i18nKey="approve.termsAndConditions"
+                values={{ appName: manifestName }}
+                components={[
+                  <Link
+                    key={manifestId}
+                    onClick={() => openURL(termsOfUse[manifestId as keyof typeof termsOfUse])}
+                    to={"https://www.ledger.com"}
+                  >
+                    {manifestId}`s terms of use.
```
gre, line 267: to put in wording files?

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -192,79 +193,88 @@ const TransactionConfirm = ({
...
+            <HorizontalSeparator />
+            <Text marginTop={30}>
+              <Trans
+                i18nKey="approve.termsAndConditions"
+                values={{ appName: manifestName }}
+                components={[
+                  <Link
+                    key={manifestId}
+                    onClick={() => openURL(termsOfUse[manifestId as keyof typeof termsOfUse])}
+                    to={"https://www.ledger.com"}
```
gre, line 265: it could be wise to make it a constant in `urls.ts` and maybe it's already an url there? are we sure we go to the / of ledger.com ?

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -192,79 +193,88 @@ const TransactionConfirm = ({
...
-          {t("approve.description")}
-        </Text>
-        <Alert type="primary" mb={26}>
-          <Trans
-            i18nKey={
-              amountTransaction === "Unlimited MATIC" ? "approve.unlimited" : "approve.limited"
-            }
-            values={{
-              recipientWording,
+  console.log("typeTransaction", typeTransaction);
```
gre, line 196: console.log left in the code

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -192,79 +193,88 @@ const TransactionConfirm = ({
...
-            const MaybeComponent = fieldComponents[field.type];
-            if (!MaybeComponent) {
-              console.log(
-                `TransactionConfirm field ${field.type} is not implemented! add a generic implementation in components/TransactionConfirm.js or inside families/*/TransactionConfirmFields.js`,
+            mb={20}
+          >
+            {fields.map((field, i) => {
+              const MaybeComponent = fieldComponents[field.type];
+              if (!MaybeComponent) {
+                console.log(
```
gre, line 230: - Could you make this a console.warn?
- is it a case that should NOT happen? only in development phase? i would just not expect a production app to display these, if it's a normal case, then we don't want it to log

gre (changes requested):
I would suggest to put back the PR to Draft mode, a PR that is opened not in draft should have the description properly filled 🙏 

some feedback. thanks

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -129,14 +189,91 @@ const TransactionConfirm = ({ t, device, account, parentAccount, transaction, st
...
-    transaction,
-    status,
-  });
+
   const key = ("mode" in transaction && transaction.mode) || "send";
   const recipientWording = t(`TransactionConfirm.recipientWording.${key}`);
+
+  console.log("typeTransaction", typeTransaction);
+
+  if (typeTransaction === "Approve" || typeTransaction === "Present")
```
Wozacosta, line 198: I don't think this will work as "Present" isn't a type of transaction afail.
It is set here https://github.com/LedgerHQ/ledger-live/blob/8fdc17621d6a5b238ae8517e490844af330a3f41/libs/coin-evm/src/deviceTransactionConfig.ts#L396 and the field is not `Type` but rather it is `Data`

Wozacosta: My bad just noticed you filtered correctly on `Data` fields also

.changeset/polite-ducks-wait.md
```diff
@@ -0,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"@ledgerhq/live-common": patch
```
Justkant, line 3: We only modify LLD here

```suggestion
```

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmAlert.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+type Props = {
+  t: TFunction;
+  transaction: Transaction;
+  typeTransaction: string | undefined;
+  fields: DeviceTransactionField[];
+};
+
+const ConfirmAlert = ({ t, transaction, typeTransaction, fields }: Props) => {
+  const key = ("mode" in transaction && transaction.mode) || "send";
```
Justkant, line 16: The ternary looks a bit easier to read
```suggestion
  const key = "mode" in transaction ? transaction.mode : "send";
```

Wozacosta:
Will do a rebase and fix some conflicts due to feat/LIVE-10709-change-ui-approval-token being merged into develop

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmAlert.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const recipientWording = t(`TransactionConfirm.recipientWording.${key}`);
+
+  const amountTransaction: string = useMemo(
+    () =>
+      (
+        fields.find(
+          (field: { label: string }) => field.label && field.label === "Amount",
+        ) as DeviceTransactionField & { value: string }
```
Justkant, line 24: Do we really need the `as` here ?
I suppose it's copied from the other component so we might want to look at that

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
Wozacosta: this `to` was unnecessary there

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
Wozacosta: 🧹

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmTitle.tsx
```diff
@@ -0,0 +1,62 @@
...
+type Props = {
+  t: TFunction;
+  Title:
+    | React.ComponentType<{
+        account: Account | SubAccount;
+        parentAccount: Account | null | undefined;
+        transaction: TransactionCommon;
+        status: TransactionStatusCommon;
+      }>
+    | undefined;
```
Justkant, line 23: It's a bit strange to pass the ComponentType instead we should expect a `ReactNode` and we can avoid tying the props of title to this component

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmTitle.tsx
```diff
@@ -0,0 +1,62 @@
...
+  if (Title) {
+    return (
+      <Title
+        account={account}
+        parentAccount={parentAccount}
+        transaction={transaction}
+        status={status}
+      />
+    );
+  }
```
Justkant, line 54: By expecting a `ReactNode` instead we can simplify this part and do it like we would with children (we could also use children here) and we can also remove the other props we don't need now
```suggestion
  if (title) {
    return title;
  }
```

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -94,12 +101,12 @@ const commonFieldComponents: Record<string, FieldComponent> = {
...
   text: TextField,
 };
-const Container = styled(Box).attrs(() => ({
+export const Container = styled(Box).attrs(() => ({
   alignItems: "center",
   fontSize: 4,
   pb: 4,
 }))``;
-const Info = styled(Box).attrs(() => ({
+export const Info = styled(Box).attrs(() => ({
```
Justkant: Do we need to export those components ? if not we should probably remove the export

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
```diff
@@ -142,13 +144,13 @@ const TransactionConfirm = ({
     status,
   });
 
-  const typeTransaction: string = useMemo(
+  const typeTransaction: string | undefined = useMemo(
     () =>
       (
         fields.find(
           (field: { label: string }) => field.label && field.label === "Type",
         ) as DeviceTransactionField & { value: string }
```
Justkant, line 152: Same comment as above for the `as` keyword

apps/ledger-live-desktop/src/renderer/modals/SignTransaction/Body.tsx
```diff
@@ -49,16 +49,13 @@ type Props = {
...
+function useSteps(canEditFees = false): St[] {
   const { t } = useTranslation();
   return useMemo(() => {
     const steps: St[] = [
       {
         id: "summary",
-        label:
-          typeTransaction === "Approve"
-            ? t("send.steps.reviewSummary.title")
-            : t("send.steps.summary.title"),
```
Justkant, line 61: Is this removed because we only needed it for the wallet-api integration ?
We might want to make sure we don't break the platform-sdk part tho, I don't know what component it uses

Justkant (commented):
Some comments, looks good but I'll test and check again after the changes from the review

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmAlert.tsx
Wozacosta: We don't 👍 , cleaned up the memoized function also to make it a bit more readable

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmTitle.tsx
Wozacosta: Indeed that's now cleaned up following your recommendation here and the one below, same change was made for the footer

https://github.com/LedgerHQ/ledger-live/pull/6205/commits/ebc8b5f97a83710b48e6fd925a2fd8a149860c3f

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmAlert.tsx
Wozacosta: 🧹 https://github.com/LedgerHQ/ledger-live/pull/6205/commits/7082400185baebf65ec655faed2bba8bef4f0f1e

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
Wozacosta: 🧹 https://github.com/LedgerHQ/ledger-live/pull/6205/commits/7082400185baebf65ec655faed2bba8bef4f0f1e

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
Wozacosta: removed the export for `Container` and `Info` wasn't even used anymore so removed it altogether: 3172830c4a6a8069a2f75cb5771a3f8d88370112

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
Wozacosta: I'd guess it's useful in development phase and later on to catch potential oversights. 
In any case for now I've turned it into a console.warn instead.

apps/ledger-live-desktop/src/renderer/modals/SignTransaction/Body.tsx
Wozacosta: It's not removed per se but instead put back to a non-ternary form: 

https://github.com/LedgerHQ/ledger-live/pull/6205/files/5359b50f963b3efd669960afbcaffb8692187ddc#diff-39ce21d822cea075fe02e6bd016907e5a1fc4c0da01380cc36efd52c3115bd1eL72-L78 

In all cases of a transaction signature the text is now the same for the device step (_Connect Ledger_)

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/ConfirmFooter.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type Props = {
+  footer: React.ReactNode | undefined;
+  manifestId?: string | null;
+  manifestName?: string | null;
+};
+
+const ConfirmFooter = ({ footer, manifestId, manifestName }: Props) => {
+  if (manifestId) {
```
Justkant, line 30: This might be a bit cleaner to instead do an early return when there is no `manifestId` and same for `termsOfUseUrl` too

Justkant (commented):
Small details

Justkant (dismissed):

apps/ledger-live-desktop/src/renderer/components/TransactionConfirm/index.tsx
Wozacosta: it was moved into wording files 👍

gre (approved):

PR #6212, opened 2024-02-15
[coin-config] Under maintenance status

lvndry:
### 📝 Description

We implemented coin-config to be able to remotely change the state of the app without depending on the release cylcle.
The first use of coin-config will be to indicate to user the current status of a coin. For now we support "active" and "under_maintenance" which are implemented in this PR.

### ❓ Context

- **JIRA or GitHub link**: 

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/ledger-live-common/src/config/sharedConfig.ts
+++ b/libs/ledger-live-common/src/config/sharedConfig.ts
@@ -7,6 +7,8 @@ import { cardanoConfig } from "../families/cardano/config";
 import { casperConfig } from "../families/casper/config";
 import { celoConfig } from "../families/celo/config";
 import { cryptoOrgConfig } from "../families/crypto_org/config";
+import { elrondConfig } from "../families/elrond/config";
+import { fileCoinConfig } from "../families/filecoin/config";
 
 const liveCommonConfig: ConfigSchema = {};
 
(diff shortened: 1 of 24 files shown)


libs/ledger-live-common/src/config/sharedConfig.ts
```diff
@@ -1,6 +1,13 @@
 import { ConfigSchema } from "@ledgerhq/live-config/LiveConfig";
-import cosmosConfig from "../families/cosmos/config";
+import { cosmosConfig } from "../families/cosmos/config";
+import { bitcoinConfig } from "../families/bitcoin/config";
+import { evmConfig } from "../families/evm/config";
 
 const liveCommonConfig: ConfigSchema = {};
 
-export default { ...liveCommonConfig, ...cosmosConfig } as ConfigSchema;
```
gre, line 6: it sounds like this should be a "generated" file. (it's good when we can have all the "multiplexer" entrypoints in generated/ because it's a pattern we will eventually want to opt out when we truly modularize live-common)

lvndry: Can be discussed further be overall we try move away from this generated file pattern

hzheng-ledger: It's true that it would be better to use the 'generated' file pattern. In this case, we should modify 'ledger-live-common/scripts/sync-families-dispatch.mjs'. You can sync with Gaetan regarding this improvement. However, I consider that to be an enhancement for a separate task. I am okay with this PR as it is.

hzheng-ledger (dismissed):
looks good to me

PR #6307, opened 2024-02-24
[Bugfix/Casper] Missing operations in lld for certain accounts

lawRathod:
### 📝 Description

Fix missing operations for Casper accounts. Some operations didn't make it to the final list because of unhandled cl_type. The `cl_type` can be either casper address or the public key. This is now handled and should result in all operations for an account to appear in the final list on lld/sync.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-11175

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/ledger-live-common/src/families/casper/bridge/bridgeHelpers/txn.ts
+++ b/libs/ledger-live-common/src/families/casper/bridge/bridgeHelpers/txn.ts
@@ -10,6 +10,7 @@ import { casperAccountHashFromPublicKey, casperGetCLPublicKey, isAddressValid } 
 import { ITxnHistoryData } from "../../api/types";
 import { getEstimatedFees } from "./fee";
 import { CasperOperation } from "../../types";
+import invariant from "invariant";
 
 export const getUnit = (): Unit => getCryptoCurrencyById("casper").units[0];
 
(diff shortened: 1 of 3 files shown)


libs/ledger-live-common/src/families/casper/bridge/bridgeHelpers/txn.ts
```diff
@@ -22,7 +23,15 @@ export function mapTxToOps(
...
-    const toAccount = txArgs.target.parsed;
+    let toAccount;
+
+    if (txArgs.target.cl_type === "PublicKey") {
+      toAccount = casperAccountHashFromPublicKey(txArgs.target.parsed);
+    } else {
+      toAccount = txArgs.target.parsed;
+    }
+    invariant(toAccount, "toAccount is required");
+    invariant(fromAccount, "fromAccount is required");
```
haammar-ledger, line 34: Isn't this a bit harsh? It will cause a crash if any data is missing, I think we would prefer a silent failure, even if it means skipping some operations, the impact would be better for the user. WDYT?

lawRathod: I see what you mean, it does feel like a harsh expectation for the addresses to be present. I will fix it in favor of skipping rather than just giving up altogether.

lawRathod: I moved the logic of composing an operation inside a try block so if there's any exception raised we can still proceed with other operations. I realized this should have been done sooner 😓

haammar-ledger (approved):
Look

hedi-edelbloute (approved):

PR #6310, opened 2024-02-26
[LIVE-11369][LLM] Delete feature flag "llmNewFirmwareUpdateUx", replace by business logic

ofreyssinet-ledger:
### 📝 Description

Right now, the `llmNewFirmwareUpdateUx` feature flag enables the new firmware update UX for ALL devices at once (including Nanos over USB, not for Nano X BLE). It is inconvenient as this flow needs adaptations in order to work properly on Nano devices. **The feature flag therefore serves no purpose**.

- Functional changes:
  - Delete the flag `llmNewFirmwareUpdateUx`
  - Refactor the decisional tree so that the new update flow is only available for Stax & Europa (BLE+USB) (for now) and that for other (wired) devices it uses the old fully functional flow, as it is today in production.
  - Fix end of firmware update: use `setLastSeenDeviceInfo` (instead of broken `setLastSeenDevice`) so that LLM knows the version of the last seen device. The `setLastSeenDevice` action creator had no associated reducer so it was not doing anything at all. Because of that, after a first successful update, the firmware update banner was not disappearing (or getting updated info in case of another update).
- Refactoring:
  - The entry point for the firmware update is the component `FirmwareUpdateBanner` and the logic it contains
    - Moved this component to `newArch`.
    - Split the logic in smaller logical blocks with unit tests

**new tests:**
Screenshot 2024-03-01 at 13 28 21
Screenshot 2024-03-04 at 16 51 18

**QA: what should be tested ?**
- For the last connected device, if a firmware update is NOT available, the firmware update banner should not be displayed on the portfolio page or in My Ledger.
- For the last connected device, if a firmware update is available, the firmware update banner should be displayed on the portfolio page and in My Ledger.
- Tapping on the firmware update banner:
  - For Ledger Stax, it should trigger a navigation to the new firmware update flow. This is the flow were the UI shows the several steps of the update.
  - For Ledger Nano models:
    - iOS, any device model: it should display a banner prompting the user to update their device using LLD.
    - Android, Nano X that was last connected via bluetooth: it should prompt the user to update their device using LLD.
    - Android, Nano (S/SP/X) that was last connected via USB: if the version does not satisfy this criteria, it should display a banner prompting the user to update their device using LLD:
      - Nano S: version ≥ 1.6.1
      - Nano X: version ≥ 1.3.0
      - Nano SP: version ≥ 1.0.0
    - Android, Nano (S/SP/X) last connected via USB: if the version satisfies the criteria above, it should navigate to the (old) firmware update flow.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11369]

### ✅ Checklist

  - LLM: firmware update banner and redirection to firmware update flows depending on the device, its firmware version and the mobile OS.

---

[LIVE-11369]: https://ledgerhq.atlassian.net/browse/LIVE-11369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/index.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { memo } from "react";
+import { Pressable } from "react-native";
+import { useTranslation } from "react-i18next";
+import { Text, Flex, Icons } from "@ledgerhq/native-ui";
+import { DownloadMedium, UsbMedium } from "@ledgerhq/native-ui/assets/icons";
+import { DeviceModelId, getDeviceModel } from "@ledgerhq/devices";
+
+import Button from "../../../../../components/Button";
+import QueuedDrawer from "../../../../../components/QueuedDrawer";
```
KVNLS, line 9: Please use `~` alias

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/useUpdateBannerViewModel.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+jest.mock("../../utils/navigateToFirmwareUpdateFlow", () => ({
+  ...jest.requireActual("../../utils/navigateToFirmwareUpdateFlow"),
+  navigateToFirmwareUpdateFlow: jest.fn(),
+}));
+
+jest.requireMock("../../utils/navigateToFirmwareUpdateFlow");
+const navigateToFwUpdateFlowSpy = jest.spyOn(
+  jest.requireMock("../../utils/navigateToFirmwareUpdateFlow"),
+  "navigateToFirmwareUpdateFlow",
+);
```
KVNLS, line 60: I think that you should cover this hook through an integration test. You mock too much stuff here. We can have a look together if you want.

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/ViewProps.tsx
```diff
@@ -0,0 +1,11 @@
+import { lastConnectedDeviceSelector } from "~/reducers/settings";
+
+export interface ViewProps {
```
KVNLS, line 3: Open question, do we want to extract types into a dedicated file? This is something I kept next to the component in Market.

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/utils/isFirmwareUpdateSupported.ts
```diff
@@ -0,0 +1,40 @@
...
+      return false;
+  }
+}
+
+export function isOldFirmwareUpdateUxSupported({
+  lastSeenDeviceModelInfo,
+  lastConnectedDevice,
+}: {
+  lastSeenDeviceModelInfo: DeviceModelInfo | null | undefined;
+  lastConnectedDevice: ReturnType<typeof lastConnectedDeviceSelector> | null | undefined;
```
KVNLS, line 22: Do we want to explicitly pass `undefined`? Can't we make it optional with `?:`?

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/utils/isFirmwareUpdateSupported.ts
```diff
@@ -0,0 +1,40 @@
+import { Platform } from "react-native";
+import { DeviceModelInfo } from "@ledgerhq/types-live";
+import { DeviceModelId } from "@ledgerhq/devices";
+import isFirmwareUpdateVersionSupported from "@ledgerhq/live-common/hw/isFirmwareUpdateVersionSupported";
+import { lastConnectedDeviceSelector } from "~/reducers/settings";
+
+export function isNewFirmwareUpdateUxSupported(deviceModelId?: DeviceModelId) {
+  switch (deviceModelId) {
```
KVNLS, line 8: As it's only true or false, an alternative could be to use includes:
```js
const NEW_UX_SUPPORTED_DEVICES = [DeviceModelId.stax, DeviceModelId.europa];
...
const isNewFirmwareUpdateUxSupported(deviceModelId?: DeviceModelId) => NEW_UX_SUPPORTED_DEVICES.includes(deviceModelId)
```

KVNLS (changes requested):
Great to see the newArch used!

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/ViewProps.tsx
ofreyssinet-ledger: I think it's useful so you can at the same time use it as return type from the `useViewModel` and props type for the view.
It makes it much easier then to do simple refactoring operations like renaming one of the props without breaking everything.
The alternative would be to define it in the same file as the `useViewModel` (because the component already imports from this file).

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/utils/isFirmwareUpdateSupported.ts
ofreyssinet-ledger: The `undefined` type comes from the definition of the settings state (in the redux store), I'm going to check because it might actually be unnecessary (`null` being the default value).

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/ViewProps.tsx
```diff
@@ -0,0 +1,11 @@
...
+  bannerVisible: boolean;
+  lastConnectedDevice: ReturnType<typeof lastConnectedDeviceSelector>;
+  version: string;
+  onClickUpdate: () => void;
```
jiyuzhuang, line 7:
```suggestion
  onClickUpdate(): void;
```
You can define a function type like this, more readable and simpler.

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { UpdateStep } from "~/screens/FirmwareUpdate";
+import { useUpdateBannerViewModel } from "./useUpdateBannerViewModel";
+import { ViewProps } from "./ViewProps";
```
jiyuzhuang, line 12:
```suggestion
import type { ViewProps } from "./ViewProps";
```

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export type FirmwareUpdateBannerProps = {
+  onBackFromUpdate: (updateState: UpdateStep) => void;
+};
+
+const UpdateBanner = ({
```
jiyuzhuang, line 18: I'm not sure why the component and the export are defined separately. The props passed to the component is the returns of a hook, you can put it in the component with a destruction of the return value.
```TypeScript
export default memo(function(outerProps: FirmwareUpdateBannerProps ) {
    const { ... }: ViewProps = useUpdateBannerViewModel(outerProps);
    return <Component />;
});
```

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/utils/navigateToFirmwareUpdateFlow.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export function navigateToFirmwareUpdateFlow({
+  lastConnectedDevice,
+  lastSeenDeviceModelInfo: lastSeenDeviceModelInfo,
+  latestFirmware,
+  route,
+  navigation,
+  onBackFromUpdate,
+}: {
+  lastConnectedDevice: ReturnType<typeof lastConnectedDeviceSelector>;
```
jiyuzhuang, line 18: For the type as 'big' as this one, maybe give it name is a good choice and make it easier to read. 🥹

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/index.tsx
ofreyssinet-ledger: This is [the new convention](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/4508058220/Code+Guidelines+-+React+Native#useComponentViewModel) that was proposed by @KVNLS last week.
It allows to decouple the rendering logic and the business model. But tell me maybe I don't understand exactly your point.

jiyuzhuang: Ok, now that it's in the convention, following it is good.

valpinkman (approved):

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/UpdateBanner.test.tsx
```diff
@@ -0,0 +1,399 @@
...
+
+const oldUpdateFlowSupportedDataSet: Array<{
+  deviceModelId: DeviceModelId;
+  version: string;
+  productName: string;
+}> = [
+  { deviceModelId: DeviceModelId.nanoS, version: "1.6.1", productName: "Nano S" },
+  { deviceModelId: DeviceModelId.nanoX, version: "1.3.0", productName: "Nano X" },
+  { deviceModelId: DeviceModelId.nanoSP, version: "1.0.0", productName: "Nano S Plus" },
+];
```
KVNLS, line 53: I think that you can reduce copy paste here, with shared TS type + default object per device that you spread and override with only the value you want to change

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/UpdateBanner.test.tsx
```diff
@@ -0,0 +1,399 @@
...
+            hasCompletedOnboarding: true,
+            wired: true,
+          }),
+        },
+        appstate: {
+          ...state.appstate,
+          hasConnectedDevice: true,
+        },
+      }),
+    });
```
KVNLS, line 94: Same here you can create a function that you reuse everywhere with only what you need to update as parameter

apps/ledger-live-mobile/src/newArch/features/FirmwareUpdate/components/UpdateBanner/UpdateBanner.test.tsx
```diff
@@ -0,0 +1,399 @@
...
+      expect(navigateToOldUpdateFlow).not.toHaveBeenCalled();
+      expect(navigateToNewUpdateFlow).not.toHaveBeenCalled();
+
+      // Check that the unsupported drawer is displayed
+      expect(await screen.findByText("Firmware Update")).toBeOnTheScreen();
+      expect(
+        await screen.findByText(
+          "Update your Ledger Nano firmware by connecting it to the Ledger Live application on desktop",
+        ),
+      ).toBeOnTheScreen();
```
KVNLS, line 310: You should always assert the change in the UI before anything else. Because if you don't assert that the change is done you can assert wrongly that something has not been called by being too quick in execution.
For example, here we could have:
`navigateToOldUpdateFlow` or `navigateToNewUpdateFlow` has been called after the change in the UI but as you assert it just after the action without checking a change in the UI we are not sure if it works thanks to execution time or because it really works.

KVNLS (changes requested):
Great ! Few details regarding the test

lduchemin-ledger (approved):

KVNLS (approved):

PR #6311, opened 2024-02-26
bugfix/LIVE 11372 swap moonpay transaction broadcasted screen is not seen

cng-ledger:
### 📝 Description

Platform exchange flow is missing transaction success content, supporting success flow with this PR

Code is mostly ported from Swap exchange drawer, including the page analytics

https://github.com/LedgerHQ/ledger-live/assets/138497251/5373c22e-1326-4e80-b99e-af24b32dabb9

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11372](https://ledgerhq.atlassian.net/browse/LIVE-11372)

### ✅ Checklist

  - ...

---

[LIVE-11372]: https://ledgerhq.atlassian.net/browse/LIVE-11372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/cyan-peas-draw.md
+++ b/.changeset/cyan-peas-draw.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+fix(LIVE-11372): show transaction success for platform exchange flow
--- a/apps/ledger-live-desktop/src/renderer/modals/Platform/Exchange/CompleteExchange/Body.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/Platform/Exchange/CompleteExchange/Body.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import { useDispatch } from "react-redux";
 import { Operation, SignedOperation } from "@ledgerhq/types-live";
 import { Exchange } from "@ledgerhq/live-common/exchange/platform/types";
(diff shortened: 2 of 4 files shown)


sarneijim (changes requested):
Up to the team, but in my perspective, the responsibility of 'completeExchange' is solely to execute the transaction and nothing more. Following the transaction successful response from LL,  the provider has the flexibility to decide whether to display their own success message or utilise our custom swap methods.
Specific methods for these scenarios have been defined in the demo 0 at `apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/SwapWebView.tsx`.
If the rest of the team prefer to proceed with this approach, we need to remove those methods and update demo 0.

Faster speed on the provider implementation vs personalise dappexperience

kallen-ledger:
@sarneijim I tend to agree with your point of view on this. However may be best to get products view on this tomorrow?

lpaquet-ledger:
> Specific methods for these scenarios have been defined in the demo 0 at `apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/SwapWebView.tsx`. If the rest of the team prefer to proceed with this approach, we need to remove those methods and update demo 0.

So instead of all having them in the same place they would be scattered in the files where we need them ?

cng-ledger:
@sarneijim @kallen-ledger 

Thanks for the comment. Here is my digestion of this, let me know if this is correct. 

We will implement:
- Separate Transaction success drawer / modal
- A new handler to display successful swap message. using the new drawer / modal

Then it will depend on the Provider to call this method to show success drawer, it the future any other providers can just call this new handler?

sarneijim:
> @sarneijim @kallen-ledger
> 
> Thanks for the comment. Here is my digestion of this, let me know if this is correct.
> 
> We will implement:
> 
> * Separate Transaction success drawer / modal
> * A new handler to display successful swap message. using the new drawer / modal
> 
> Then it will depend on the Provider to call this method to show success drawer, it the future any other providers can just call this new handler?

Yes, I assume when you talk about drawer / modal is always drawer for lld. Is that correct?
We do not need a new "drawer" is the same one but defined from a different place.
And the providers can trigger them using custom methods
- Following my understood we have:
    - Save transaction and success message
    - Redirect to the history
Both are defined as a custom methods in: `apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/SwapWebView.tsx`
The only thing that we need to do is move them from this file (demo 0) to the main swap and add them as accesible methods in Moonpay manifest, then they can call them.

kallen-ledger (approved):

cng-ledger:
Outcome is transaction success can stay within the platform exchange drawer for now. The saving swap to history logic should probably be centralised, however these will be a separate ticket

sarneijim (approved):

PR #6327, opened 2024-02-28
Safer solana hydrate

hedi-edelbloute:
### 📝 Description

Solana preloaded data somehow ends up with a data without version, this PR adds a safer check before hydrate

I cannot reproduce so it looks like it is not consistent, anyway hydrate data field should be treated as unsafe, if data is corrupted we have to return; directly.

```
// reinject the preloaded data (typically if it was cached)
// method need to treat the data object as unsafe and validate all fields / be backward compatible.
hydrate(data: unknown, currency: CryptoCurrency): void;
```

If it's unsafe we drop it and rely on fresh data. It looks like in the sentry issue case the user somehow ended up with preload data without version which should not happen as you can see on linked screenshot

### ❓ Context

- **JIRA or GitHub link**: [https://ledgerhq.atlassian.net/browse/LIVE-11568](https://ledgerhq.atlassian.net/browse/LIVE-11568)

### ✅ Checklist

- [x] **Impact of the changes:** Solana hydrate
  - ...

---

### DIFF
--- a/libs/ledger-live-common/src/families/solana/js-preload.test.ts
+++ b/libs/ledger-live-common/src/families/solana/js-preload.test.ts
@@ -1,0 +1,14 @@
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { hydrate } from "./js-preload";
+import { SolanaPreloadData } from "./types";
+
+describe("hydrate", () => {
+  it.each([undefined, null, {}, { version: null }, { version: undefined }, {}])(
+    "should return undefined if data is corrupted (%p value)",
+    value => {
+      expect(hydrate(value as unknown as SolanaPreloadData, {} as CryptoCurrency)).toEqual(
+        undefined,
+      );
+    },
+  );
+});
--- a/.changeset/unlucky-bottles-dress.md
+++ b/.changeset/unlucky-bottles-dress.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Safer solana hydrate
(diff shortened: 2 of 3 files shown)


libs/ledger-live-common/src/families/solana/js-preload.test.ts
```diff
@@ -0,0 +1,14 @@
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { hydrate } from "./js-preload";
+import { SolanaPreloadData } from "./types";
+
+describe("hydrate", () => {
+  it.each([undefined, null, {}, { version: null }, { version: undefined }, {}])(
+    "should return undefined if data is corrupted (%p value)",
+    value => {
+      expect(hydrate(value as unknown as SolanaPreloadData, {} as CryptoCurrency)).toEqual(
```
sprohaszka-ledger, line 9: Why casting it as `unknown`? You should be able to call "hydrate" with the accepted parameter type (i.e. `null` should never be able to call this function, as well as a non string `version` value)

sprohaszka-ledger (commented):
We should rely on the type system and avoid testing force cast call.

libs/ledger-live-common/src/families/solana/js-preload.test.ts
hedi-edelbloute: Please check description and Sentry, this value is not intended and should not happen. It can't be reproduced manually

hedi-edelbloute: Do you think I should add more cases ?

hedi-edelbloute: Sure, can do ! :) 
I think it's supposed to be the base behavior of hydrate and it should have been this way since the beginning

PR #6333, opened 2024-02-28
fix: keep a margin of 0,005eth/bsc/matic for swaps using evm

CremaFR:
### 📝 Description

fix attempt 

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-11085

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
+++ b/libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
@@ -32,16 +32,21 @@ export const useUpdateMaxAmount = ({
   const [isMaxLoading, setIsMaxLoading] = useState(false);
 
   const toggleMax: SwapTransactionType["toggleMax"] = useCallback(
-    () =>
+    account =>
       setIsMaxEnabled(previous => {
         const next = !previous;
         if (previous) {
           setFromAmount(ZERO);
           setIsMaxLoading(false);
         }
+        let additionalFees;
+        if (account?.type === "Account" && account?.unit?.code === "ETH") {
+          additionalFees = new BigNumber(10000000000000000); // 0,01 ETHs
+        }
         bridge.updateTransaction(tx => ({
           ...tx,
           useAllAmount: next,
+          additionalFees,
         }));
         return next;
       }),
--- a/libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
+++ b/libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
@@ -32,22 +32,24 @@ export const useUpdateMaxAmount = ({
   const [isMaxLoading, setIsMaxLoading] = useState(false);
 
   const toggleMax: SwapTransactionType["toggleMax"] = useCallback(
-    account =>
+    () =>
       setIsMaxEnabled(previous => {
         const next = !previous;
         if (previous) {
           setFromAmount(ZERO);
           setIsMaxLoading(false);
         }
-        let additionalFees;
-        if (account?.type === "Account" && account?.unit?.code === "ETH") {
-          additionalFees = new BigNumber(10000000000000000); // 0,01 ETHs
-        }
-        bridge.updateTransaction(tx => ({
-          ...tx,
-          useAllAmount: next,
-          additionalFees,
-        }));
+        bridge.updateTransaction(tx => {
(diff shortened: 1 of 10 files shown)


libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
```diff
@@ -32,16 +32,21 @@ export const useUpdateMaxAmount = ({
...
         if (previous) {
           setFromAmount(ZERO);
           setIsMaxLoading(false);
         }
+        let additionalFees;
+        if (account?.type === "Account" && account?.unit?.code === "ETH") {
+          additionalFees = new BigNumber(10000000000000000); // 0,01 ETHs
+        }
         bridge.updateTransaction(tx => ({
           ...tx,
```
sarneijim, line 47: Does 'tx' have 'fromAccount' attribute? I do not think you need to pass any parameter to toggleMax

libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
```diff
@@ -32,16 +32,21 @@ export const useUpdateMaxAmount = ({
...
       setIsMaxEnabled(previous => {
         const next = !previous;
         if (previous) {
           setFromAmount(ZERO);
           setIsMaxLoading(false);
         }
+        let additionalFees;
+        if (account?.type === "Account" && account?.unit?.code === "ETH") {
+          additionalFees = new BigNumber(10000000000000000); // 0,01 ETHs
+        }
```
sarneijim, line 45: Could we use more descriptive with the variable name (although I am not sure my name is the best one)?
I think it can be better create something more generic...tomorrow I assume we will add other families here 

```suggestion
        const familySecurityFee = {
                 ETH: 10000000000000000 // 0,01 ETHs
        }
         bridge.updateTransaction(tx => ({
          ...tx,
          useAllAmount: next,
          securityFee: account?.type === "Account" && familySecurityFee[account?.unit?.code],
        }));
```

libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
CremaFR: Transaction is undefined in this context. I had to pass it

libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
CremaFR: Idea was to reuse an existing field for ETH as a quick fix, so I'm not in favor or renaming it. Especially since the modification made is generic, it is just more fees deducted, not specific to a use case. More info on [slack](https://ledger.slack.com/archives/C02LACPH3U3/p1709033172133369)

Also I don't think this fix will be extended upon more coins. 
Max button logic will be removed with swap-live-app to use a better approach of treating max button as user input to not have two different logics

libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
sarneijim: > Transaction is undefined in this context. I had to pass it

If 'tx' if undefined in this context, can we remove it? (I can not understand how a bridge which update transaction can have transaction undefined)

libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
sarneijim: Follow my understood this is the solution that we want to apply in the web app and migrate LLM will take months...
What is the the difference between "evm" and other chains? If the fix works, I am sure we will e applying it for other coins. 
On the other hand can we change "additionalFees" for "marginFee" or marginEthFee"?

haammar-ledger (commented):
Need review from @lambertkevin here

lambertkevin (dismissed):

libs/ledger-live-common/src/exchange/swap/hooks/useUpdateMaxAmount.ts
CremaFR: You are right updating it to include all evm chains
but don't want to rename `additionalFees` it is generic like this

sarneijim (approved):

PR #6342, opened 2024-03-01
Add test in Polkadot module

sprohaszka-ledger:
### 📝 Description

Add some test to prepare rework for Lama compatibility.
Those tests are also here for example for more tests (for other coin modules).

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11678]

### ✅ Checklist

  - ...

---

[LIVE-11678]: https://ledgerhq.atlassian.net/browse/LIVE-11678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-polkadot/src/bridge/prepareTransaction.test.ts
```diff
@@ -0,0 +1,48 @@
...
+}));
+
+describe("prepareTransaction", () => {
+  afterEach(() => {
+    mockGetEstimatedFees.mockClear();
+  });
+
+  it("returns a new Transaction with new fees", async () => {
+    // Given
+    const fees = new BigNumber(faker.number.int(50));
```
hedi-edelbloute, line 18: I don't think using random values in unit tests is the right way... but it's my opinion

In your test specific case I don't think it can hurt but it's a new pattern in our codebase, and has a benefit/ratio cost to discuss

Thoughts:
- Each run is different, we don't provide static values but instead ranges. It goes against a deterministic test and overall against F.I.R.S.T. principle of testing
- We introduce "maybe" more accurate real data (dev has to know and define the real range of values, which changes in real life anyway if we are dealing with a backend => not unit tests role) at the cost of "possible" flakiness. Failing cases are supposed to be tested anyway
- Reduced performance for unit tests runs (random values computing + added library)
- Random values should be used for "exploratory testing" only IMO
- Sometimes there are relations between attributes that faker usage needs to take into account, and it's extra work (imagine relation between spendableBalance, balance, lockedBalance)
- Unit tests are sort of guidelines to the future developer on what to care about, if we really want to add different cases we can use `it.each`

libs/coin-polkadot/src/bridge/prepareTransaction.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  getEstimatedFees: () => mockGetEstimatedFees(),
+}));
+
+describe("prepareTransaction", () => {
+  afterEach(() => {
+    mockGetEstimatedFees.mockClear();
+  });
+
+  it("returns a new Transaction with new fees", async () => {
+    // Given
```
hedi-edelbloute, line 17: We could use this Given When Then pattern (Gerkhin) as describe and beforeEach clauses, this is what they are here for ! In my experience though, it gets heavy quickly and the only important thing is a high branch coverage of your tested method

libs/coin-polkadot/src/logic/getFeesForTransaction.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const result = await getEstimatedFees({
+      a: createFixtureAccount(),
+      t: createFixtureTransaction(),
+    });
+
+    // Then
+    expect(mockPaymentInfo).toHaveBeenCalledTimes(1);
+    // Receive hex signature computed by Polkadot lib
+    expect(mockPaymentInfo.mock.lastCall).not.toBeNull();
+    expect(result).toEqual(new BigNumber(155099814));
```
hedi-edelbloute, line 50: We could put this in a variable and use it in the mock aswell

libs/coin-polkadot/src/types/model.fixture.ts
```diff
@@ -0,0 +1,109 @@
+import BigNumber from "bignumber.js";
+import { faker } from "@faker-js/faker";
+import { listCryptoCurrencies } from "@ledgerhq/cryptoassets/currencies";
+import { PolkadotAccount, PolkadotOperationExtra, PolkadotResources, Transaction } from "../types";
+import { getAbandonSeedAddress } from "../../../ledgerjs/packages/cryptoassets/lib";
+import { Operation } from "@ledgerhq/types-live";
+
+export function createFixtureAccount(account?: Partial<PolkadotAccount>): PolkadotAccount {
```
hedi-edelbloute, line 8: I'm more of the "specific mock for specific method team", using global fixtures can end up with you editing the fixture output in your test anyway, as you are using "more than what's needed" and need to target specific attributes usually.

If you use global fixture it can also hide uncovered execution branches that you should be aware of

Example : 
- Imagine a method checking if creationDate of your account is recent we can either use a global createMockAccount method and edit its creationDate attribute or provide a { creationDate: value } mock version as input
- If there is another attribute usage that you aren't aware of (because the method is complex) you won't notice it as you are using a global fixture (example is bad in that case as it's pretty straightforward)

libs/coin-polkadot/src/types/model.fixture.ts
```diff
@@ -0,0 +1,109 @@
...
+  const currency = listCryptoCurrencies(true).find(c => c.id === "polkadot")!;
+
+  const polkadotResources: PolkadotResources = account?.polkadotResources || {
+    controller: undefined,
+    stash: account?.polkadotResources?.stash || undefined,
+    nonce: faker.number.int(100_000),
+    lockedBalance: new BigNumber(faker.string.numeric()),
```
hedi-edelbloute, line 15: Here there might be a real life logic between lockedBalance,unlockedBalance, etc... that faker does not take into account

libs/coin-polkadot/src/bridge/prepareTransaction.test.ts
sprohaszka-ledger: My opinion is:
* We don't need or want to always use faker
* Using it prove that our code is indenpendent from this parameter OR that our code is able to manage different value (like you said range)

If you need consistency or have a strong need in relationship between values, you can set your own value instead of faker's one.

I disagree with the performance argument as we are only creating test data, we are losing much more time in E2E test.

libs/coin-polkadot/src/bridge/prepareTransaction.test.ts
sprohaszka-ledger: Those are only comments to clearly separate data needed from what you test and expect.

However I agree they can tend to be heavier, but it highlights the heavy dependency of what you want to test.

hedi-edelbloute: We discussed this point in our huddle, makes sense with multiple contexts to add describe clauses

libs/coin-polkadot/src/logic/utils.test.ts
```diff
@@ -68,11 +43,11 @@ describe("canUnbond", () => {
...
         ],
       },
-    };
-    expect(canUnbond(account as PolkadotAccount)).toBeTruthy();
+    });
+    expect(canUnbond(account)).toBeTruthy();
   });
   test("can't unbond because unlockings is too much", () => {
-    const account: Partial<PolkadotAccount> = {
+    const account = createFixtureAccount({
```
hedi-edelbloute, line 50: We already discussed this, unit tests have by definition no dependencies, we are a adding a small one

To me this is the role of integration tests to provide stubs/global fixtures and test features

I do not agree with this but we already spent too much energy for no reason as the team did not decide yet on a consensus and we are only two people arguing :zipper_mouth_face: 

https://softwareengineering.stackexchange.com/questions/287613/javascript-unit-testing-mocks-or-fixtures
https://softwareengineering.stackexchange.com/a/227985
https://softwareengineering.stackexchange.com/a/198456
https://softwareengineering.stackexchange.com/a/154255
https://paytonrules.com/post/2016-04-18-dont-use-fixtures/
https://medium.com/worldsensing-techblog/tips-and-tricks-for-unit-tests-b35af5ba79b1

hedi-edelbloute (approved):
No consus from the team yet, let's move on and thanks for those additional tests !

hedi-edelbloute (approved):

lvndry (approved):

PR #6370, opened 2024-03-05
[evm] Get explorer, node and gasTracker information from coin config

lvndry:
### 📝 Description

Use coin config to get explorerViews information

Default value:

Screenshot 2024-03-08 at 17 27 18

After changing the node uri in Firebase

Screenshot 2024-03-08 at 17 32 50

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-11477

### ✅ Checklist

  - ...

---


libs/coin-evm/src/api/explorer/types.ts
```diff
@@ -16,7 +17,7 @@ export type ExplorerApi = {
   }>;
 };
 
-type ExplorerConfig = EthereumLikeInfo["explorer"];
+type ExplorerConfig = EvmConfig["explorer"];
```
sprohaszka-ledger, line 20: Not sure you need to change this. What we want is to be sure that the `ExplorerConfig` is the kind describe in `EthereumLikeInfo`

libs/coin-evm/src/api/gasTracker/types.ts
```diff
@@ -13,7 +14,7 @@ export type GasTrackerApi = {
   }) => Promise<GasOptions>;
 };
 
-type GasTrackerConfig = EthereumLikeInfo["gasTracker"];
+type GasTrackerConfig = EvmConfigInfo["gasTracker"];
```
sprohaszka-ledger, line 17: Same comment about where should be the source of truth about the type definition (maybe more in `EthereumLikeInfo`)

libs/coin-evm/src/api/node/ledger.ts
```diff
@@ -223,7 +231,6 @@ export const getFeeData: NodeApi["getFeeData"] = async (currency, transaction) =
       ...currency,
       ethereumLikeInfo: {
         ...currency.ethereumLikeInfo!,
-        gasTracker: { type: "ledger", explorerId: node.explorerId },
```
sprohaszka-ledger, line 226: Why removing it?

libs/coin-evm/src/api/node/types.ts
```diff
@@ -34,7 +35,7 @@ export type NodeApi = {
   ) => Promise<BigNumber>;
 };
 
-type NodeConfig = EthereumLikeInfo["node"];
+type NodeConfig = EvmConfigInfo["node"];
```
sprohaszka-ledger, line 38: (same comment about type source of truth)

libs/coin-evm/src/logic.ts
```diff
@@ -199,7 +200,9 @@ export const getSyncHash = (
     .map(token => token.id + token.contractAddress + token.name + token.ticker)
     .join("");
   const isNftSupported = isNFTActive(currency);
-  const { node = {}, explorer = {} } = currency.ethereumLikeInfo || {};
+
+  const config = getCoinConfig(currency).info;
+  const { node = {}, explorer = {} } = config;
```
sprohaszka-ledger, line 205: Do you think we still need this default value? Otherwise we can clean it at the same time.

libs/ledger-live-common/src/config/types.ts
```diff
@@ -1 +1 @@
-export type CurrencyConfigCommon = Record<string, unknown>;
+export type CurrencyConfigCommon<Config = Record<string, unknown>> = Config;
```
sprohaszka-ledger, line 1: I am not fan of the use of generics for composition :/

libs/coin-framework/src/config.ts
```diff
@@ -0,0 +1,7 @@
+export type CommonCurrencyConfig = {
+  status: {
+    type: "active" | "under_maintenance";
+  };
+};
+
+export type CurrencyConfig<Config = Record<string, unknown>> = CommonCurrencyConfig & Config;
```
sprohaszka-ledger, line 7: Same comment about using generics for composition

libs/ledger-live-common/src/explorers.ts
```diff
@@ -1,29 +1,37 @@
...
+};
+
 export const getAddressExplorer = (
-  explorerView: ExplorerView | null | undefined,
+  explorerView: ExplorerView | undefined,
   address: string,
-): string | null | undefined =>
-  explorerView && explorerView.address && explorerView.address.replace("$address", address);
+): string | undefined => {
+  return explorerView?.address?.replace("$address", address);
```
sprohaszka-ledger, line 19: 👍

libs/ledger-live-common/src/families/evm/setup.ts
```diff
@@ -15,12 +15,22 @@ import {
 import { Resolver } from "../../hw/getAddress/types";
 import Transport from "@ledgerhq/hw-transport";
 import { Bridge } from "@ledgerhq/types-live";
+import { getCurrencyConfiguration } from "../../config";
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { EvmConfigInfo } from "@ledgerhq/coin-evm/lib/config";
```
sprohaszka-ledger, line 20: I think the import should work without the `lib` path.

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
```diff
@@ -265,9 +265,6 @@ export const cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency> = {
     ],
     ethereumLikeInfo: {
       chainId: 43114,
-      node: { type: "ledger", explorerId: "avax" },
```
sprohaszka-ledger, line 268: Just to be sure about the behaviour: we don't want to keep them as a default value if CoinConfig is not set or loaded?

sprohaszka-ledger (commented):
I wonder where should be the source of truth about the definition EthereumLikeInfo.

libs/coin-evm/src/api/explorer/types.ts
lvndry: Actually the EthereumLikeInfo object only has `{ chainId: number}` so `explorer` and `node` are only information that can be found in the config

libs/coin-evm/src/api/node/ledger.ts
lvndry: We get the gasTracker information directly in `getGasOptions`

libs/coin-evm/src/logic.ts
lvndry: Yes the default values are still needed because node and explorer are not in the config for every evm currency. However we use the objects line 219 to create the hash

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
lvndry: There's always default values set here:
https://github.com/LedgerHQ/ledger-live/blob/feat/live-1147-evm-coin-config-explorer/libs/ledger-live-common/src/families/evm/config.ts

libs/coin-evm/src/__tests__/fixtures/synchronization.fixtures.ts
```diff
@@ -20,17 +21,24 @@ export const currency: CryptoCurrency = Object.freeze({
...
-      uri: "https://my-rpc.com",
-    },
-    explorer: {
-      type: "etherscan" as const,
-      uri: "https://api.com",
-    },
   },
 });
 
+setCoinConfig((): any => {
```
hedi-edelbloute, line 27: I'm scared this edits the coinConfig globally, Ideally we want to mock the returned value of getConfig call we can do this together :)

another dirty solution is to store in beforeAll and reapply in afterAll but not clean

To apply on every tests

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
```diff
@@ -3716,11 +3689,6 @@ export const cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency> = {
     units: ethereumUnits("ETH", "ETH"),
     ethereumLikeInfo: {
       chainId: 42161,
```
hedi-edelbloute, line 3691: This is the latest element in currencies.ts ethereumLikeInfo :thinking: We could add it in coinconfig but this won't be updated and i'm not sure this makes sense... WDYT ?

I think long term a lot more will be pulled remotely soooo

hedi-edelbloute:
Missing a changeset btw !

libs/ledgerjs/packages/cryptoassets/src/currencies.ts
lvndry: I don't think it's relevant to have the chainId as a "config", it's an attribute of a currency

libs/coin-evm/src/__tests__/fixtures/synchronization.fixtures.ts
lvndry: In the tests we usually have a `beforeEach` that overrides this one. but it's still needed before in the file we call `getSyncHash` that requires a coinConfig set up

sprohaszka-ledger: I tend to agree with Hedi here.
If you don't want to use a mock, call a `afterAll` to erase the CoinConfig setted this way. It will avoid side effects.

hedi-edelbloute (approved):

libs/coin-evm/src/createTransaction.ts
lambertkevin, line 1: Unnecessary change

libs/coin-evm/src/api/node/ledger.ts
lambertkevin: No actually @sprohaszka-ledger has a point, this removal is kinda problematic. 
This overriding mechanism is here to help with the separation of concerns of the 3 "backends" we can rely on with an EVM: A node, and explorer and a gas tracker.

Here, the goal was to say that Ledger "as a node" should be able to provide the exact same methods as a classic RPC node. But Ledger's backend isn't providing all features for now. In order to fake the behaviour of an eth_gasPrice/eth_feeHistory RPC request, we go get the result of the Ledger `gasTracker` (no matter what the configuration of the currency is) and return it as if it was an eth_gasPrice/eth_feeHistory RPC response.

Here if the configuration of the coin is set to use let's say Blocknative gas tracker (not implemented today, I know), using Ledger "as a node" would result in using Blocknative when requesting fees, and not just Ledger anymore.

So we need a way to enforce this ledger config when using `getGasOptions`.

lambertkevin (changes requested):

gre (dismissed):
ok for hub scope

lambertkevin (approved):

PR #6406, opened 2024-03-08
[LIVE-11378] Stax mcu upgrade failure

jiyuzhuang:
### 📝 Description

_For bugfixes, you can explain the previous behavior and how it was fixed._

### ❓ Context

- **JIRA or GitHub link**: 

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/ledger-live-common/src/__tests__/hw/getDeviceInfo.ts
+++ b/libs/ledger-live-common/src/__tests__/hw/getDeviceInfo.ts
@@ -36,7 +36,7 @@ test("1.3.1", async () => {
   expect(res).toMatchObject({
     version: "1.3.1",
     mcuVersion: "1.1",
-    majMin: "1.3",
+    majMin: "1.3.1",
     providerName: null,
     targetId: 823132162,
     isOSU: false,
@@ -63,7 +63,7 @@ test("1.3.1 BL", async () => {
     mcuVersion: "",
     pinValidated: false,
     providerName: null,
-    majMin: "0.0",
+    majMin: "0.0.0",
     targetId: 16777217,
   });
 });
@@ -85,7 +85,7 @@ test("1.5.5", async () => {
     mcuVersion: "1.7",
     pinValidated: false,
     providerName: null,
-    majMin: "1.5",
+    majMin: "1.5.5",
     targetId: 823132164,
   });
 });
@@ -107,7 +107,7 @@ test("1.5.5 manager allowed", async () => {
     mcuVersion: "1.7",
     pinValidated: false,
     providerName: null,
-    majMin: "1.5",
+    majMin: "1.5.5",
     targetId: 823132164,
   });
 });
@@ -129,7 +129,7 @@ test("1.4.2", async () => {
     mcuVersion: "1.6",
     pinValidated: true,
     providerName: null,
-    majMin: "1.4",
+    majMin: "1.4.2",
     targetId: 823132163,
   });
 });
@@ -151,7 +151,7 @@ test("1.4.2 manager allowed", async () => {
     mcuVersion: "1.6",
     pinValidated: true,
     providerName: null,
-    majMin: "1.4",
+    majMin: "1.4.2",
     targetId: 823132163,
   });
 });
@@ -262,7 +262,7 @@ test("0.0 bootloader", async () => {
   expect(res).toMatchObject({
     isBootloader: true,
     isOSU: false,
-    majMin: "0.0",
+    majMin: "0.0.0",
     version: "0.0.0",
     mcuVersion: "",
   });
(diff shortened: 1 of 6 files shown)


libs/ledger-live-common/src/__tests__/hw/getDeviceInfo.ts
```diff
@@ -36,7 +36,7 @@ test("1.3.1", async () => {
   expect(res).toMatchObject({
     version: "1.3.1",
     mcuVersion: "1.1",
-    majMin: "1.3",
```
gre, line 39: we have made these tests in the past to make sure we don't accidentally break old device logic.
and the changes revealed by this test indicate possible breaking for past versions.

if you search for it, you will notice in some part of the firmware update process, we do

```
(deviceInfo.majMin === "0.6" || deviceInfo.majMin === "0.7")
```

so we have a strong expectation in our codebase that these fields are in form of `x.y` . ugly legacy :'( but i think we don't want to break firmware update for old device too much.

gre (changes requested):

libs/ledger-live-common/src/__tests__/hw/getDeviceInfo.ts
jiyuzhuang: If the old firmware version is in format of `x.y` w/o `.z`, I think my fix will always work, e.g. there are some test cases for this and I don't change them. If not, can you please give out a criteria with which I can identify it's an old version firmware ?

gre: it would be preferable to introduce a new field than rewrite this one tbh.
the name of the field is literally "majMin" as in "maj.min.patch" version of the full value (that full value is in version)

jiyuzhuang: I think this is the origin of a bug when updating firmware. Actually, we have changed the version from `x.y` to `x.y.z`. After installing new firmware, some mechanism will check whether the version downloaded (`x.y.z`) is the same to the version installed (was `x.y` before, returned by `getDeviceInfo()` in field `majMin`). The field `majMin` is an extrait of the complete version, if the old version is `1.2 - dev` for example, the field will be `1.2`, if the new version is `1.2.3 - rc2`, the field will be `1.2.3`. This will not impact the old version.

jiyuzhuang: @gre Actually the `version` is not always equal to `majMin` in my proposal. The `version` may be `x.y.z - dev` for example.

gre: i understand there is a bug on this that we need to fix for Stax 👍

However, i think the current changes are breaking: if you look in `libs/ledger-live-common/src/hw/firmwareUpdate-repair.ts` you will notice we have hardcoded logic that expects this field to be a `x.y`, that part is needed for old Ledger Nano S firmware to "repair" their firmware update. This is an old mechanism but as we are still manage the Nano S devices, **if** we want to change this part, we will have to be sure to be able to test it and the scope is wider than "fixing stax mcu update" i think. Typically we will need a bunch of old devices on old firmwares (and these are harder to find these days).


Capture d’écran 2024-03-18 à 10 22 15

another location where this will break is very near the code you have modified:

```
const blVersionAliases = {
  "0.0": "0.6",
};
...
of(blVersionAliases[deviceInfo.majMin])
```

won't work anymore. this code is again a very old logic for very old firmware (if i recall, the very first firm of nano S had a specific case on this, i think some doc exists on confluence)

---

So i'm not exactly sure what is the best solution forward from here, but we could make the change less impacting by introducing a new field "majMinPatch" (can't figure out a better name, i was trying to figure out what is the semantic name in https://semver.org/ but they don't have specific names for the `x.y.z` part)

and then use this new field when it makes sense but then still have a "fallback" to the old logic when it doesn't find the data related to majMinPatch.

If we however, consider reworking this completely, we will have to test on many past firmwares on nominal and error flows (and that's challenging)
example of situation (2020) where this broke before https://ledgerhq.atlassian.net/wiki/spaces/HUB/pages/1269989380/Postmortem+LLD+1.8.0+broke+Firmware+Update+for+LNS+1.3.1

gre (changes requested):
(see comment above, but afaik this code have breaking changes on nano S upgrade paths that relies on fact majMin is a x.y format, so i just explicit a :x: on this PR so we don't introduce regression)

gre (approved):
ok, i missed the fact majMin was still going to be a x.y on these old firmware. So it might work then!

i think we will however have to be sure we don't have cases where there is a breaking transition from x.y to x.y.z , eg when we compare majMin with the data from backend in code such as `deviceInfo.majMin === m.from_bootloader_version`, it's possible we have no issue, but i think we need to make sure the QA plan includes updating from some older devices 🤔 

cc @mbertin-ledger @jiyuzhuang

PR #6421, opened 2024-03-12
LIVE-11476: remote config for app minversion

hzheng-ledger:
### 📝 Description

We add app min version configuration with the LiveConfig lib. So that we can modify min version configuration remotely via firebase in real time.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-11476

### ✅ Checklist

- [X] **Impact of the changes:** No impact
  - ...

---


libs/ledger-live-common/src/apps/config.ts
```diff
@@ -0,0 +1,72 @@
+import { ConfigInfo } from "@ledgerhq/live-config/LiveConfig";
+
+const appConfig: Record<string, ConfigInfo> = {
```
lvndry, line 3: I think it'd be nice to add a small comment to know what this file is about

lvndry (dismissed):

hzheng-ledger:
> lgtm, I wonder, should we specify nanoAppMinVersion in coinconfig instead of minVersion ?

I think "minVersion" is OK because the "minVersion" is under "config_app_cosmos", "config_app_algorand" ...
it shouldn't be confusing.
I rename "config_nanoapp_cosmos", "config_nanoapp_algorand" to "config_nanoapp_cosmos", "config_nanoapp_algorand", so it would be clear

hedi-edelbloute (approved):
Just noticed that we could have used a string for config, but like this we can use different attribute names (or specify by model in the future) so i guess it's ok

PR #6457, opened 2024-03-18
fix: rename currency id "osmo" to wallet api "osmosis"

beths-ledger:
### 📝 Description

Context: 
- Convo with wallet API https://ledger.slack.com/archives/C02LW5K2WCR/p1709288941086839
- Latest from Clement / PTX:  https://ledger.slack.com/archives/C06CZFD0Y77/p1709634306951389

### ❓ Context

- **JIRA or GitHub link**: 

### ✅ Checklist

  - ...

---

### DIFF
--- a/libs/ledgerjs/packages/cryptoassets/src/currencies.ts
+++ b/libs/ledgerjs/packages/cryptoassets/src/currencies.ts
@@ -21,7 +21,7 @@
  * if bitcoin family, supportsSegwit defines if it supports segwit.
  */
 
-import { CryptoCurrency, CoinType, Unit, CryptoCurrencyId } from "@ledgerhq/types-cryptoassets";
+import { CoinType, CryptoCurrency, CryptoCurrencyId, Unit } from "@ledgerhq/types-cryptoassets";
 
 const makeTestnetUnit = u => ({ ...u, code: `𝚝${u.code}` });
 
@@ -2745,7 +2745,7 @@ export const cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency> = {
   },
   osmosis: {
     type: "CryptoCurrency",
-    id: "osmo",
+    id: "osmosis", // Does it need to be "osmo"? It's not a ticker. But updating this could cause breaking changes where this package is used.
     coinType: CoinType.ATOM,
     name: "Osmosis",
     managerAppName: "Cosmos",


libs/ledgerjs/packages/cryptoassets/src/currencies.ts
```diff
@@ -2745,7 +2745,7 @@ export const cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency> = {
   },
   osmosis: {
     type: "CryptoCurrency",
-    id: "osmo",
+    id: "osmosis", // Does it need to be "osmo"? It's not a ticker. But updating this could cause breaking changes where this package is used.
```
beths-ledger, line 2748: In order for us to use `osmosis` in earn, which is the correct ticker in wallet-api, it looks like we would need to update the ID which we get from here, in ledger live. 

An alternative, ideally, would be to use Wallet API itself, where the ID is already updated, and not be dependent on this outdated currency list.  This alternative requires some more investigation as I am not sure how the two interact yet. For example, the Wallet API tool in ledger live Discover also has the same problem (list accounts by "osmosis" and you will not receive them, but request by "osmo" and you will). 

We use the method `getCryptoCurrencyById` (libs/coin-framework/src/currencies/support.ts) in Earn, which may be part of the problem.

gre: 👏  wow, good catch, this is super weird that `cryptocurrenciesById: Record<CryptoCurrencyId, CryptoCurrency>` have a case where the .id if the currency is not the one used in the map, it seems like we absolutely need a test that ensure it's always the case, i imagine it was an accident? ( this dates from 2022! https://github.com/LedgerHQ/ledger-live/commit/ac5f6028dd15e582a3ebc5b3ee8e47ed3f04237b )

🤔  unfortunately historically we have built a lot of things along the promise that IDs must never be changed, which is why it's the single (as documented at beginning of file, the only field we can't afford to change as documented by https://github.com/LedgerHQ/ledger-live/blob/bef2f6a32cccbcc0b6bc65d319756c4bd841ba7e/libs/ledgerjs/packages/cryptoassets/src/currencies.ts#L4 ) – For instance, when Ripple renamed to XRP, we started on "ripple" because it's the historical id we used.

❓  **That said**, there is something super weird i don't really understand, it seems like we are actually using "osmosis" today to identify the coin, does it mean that the map key is > the .id field? damn we really need a test to guarantee we have this equality. (to have key of the map always be the .id). **so the good news is it's maybe non breaking to rename the .id on this specific case?**

🧪  So I think, since it's a bugfix, please provide a test that would have detected the issue. that test simply need to iterate on all Currency listed by listCryptoCurrencies and verify that the currency is strictly === to the `cryptocurrenciesById[currency.id]` 🙏 

cc @LedgerHQ/live-blockchain-support  : please continue on the assessment here too, as it's more your scope than mine 🙏  in term of QA, to me, we just need to confirm there is no regression by coming with an existing app.json as well as coming on existing data on LLM (because not exactly same way and flow we use there)

PR #6478, opened 2024-03-19
chore: fix wording of UserRefusedFirmwareUpdate to mention the deviceName

gre:
### 📝 Description

Expected behaviour:

> I should see the message “OS update install cancelled on Ledger Stax”

Actual behaviour:

> The message is “OS update install cancelled on device”

---

the was I have fixed it is not the best: injecting at the UI time the need to have a .deviceName value on the UserRefusedFirmwareUpdate error

ℹ️  a better way to have fixed this would have been to make the error emitter sending the .deviceName, **the problem** is that it's not in the context at all on the deviceSDK side. i'll need insights from @LedgerHQ/live-devices to know if it's something we can consider (and how we can solve this?)
ℹ️  risk of the current injecting solution is that if the error is emitted in other UI places, the UI will display "OS update install cancelled on" without the deviceName.

❌  unfortunately I can't figure out the location to do the same for LanguageInstallRefusedOnDevice and RestoreLanguageInstallRefusedOnDevice so we're blocked by a solution to have the name in the context of the error from the deviceSDK thrown error.

---

Capture d’écran 2024-03-19 à 14 49 44

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-10653 https://ledgerhq.atlassian.net/browse/LIVE-10654

### ✅ Checklist

  - ...

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/errors/DeviceError.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/errors/DeviceError.tsx
@@ -11,6 +11,7 @@ import {
   ImageLoadRefusedOnDevice,
   LanguageInstallRefusedOnDevice,
 } from "@ledgerhq/live-common/errors";
+import { DeviceModel } from "@ledgerhq/types-devices";
 
 type Props = {
   error: Error;
@@ -18,6 +19,7 @@ type Props = {
   onDrawerClose: () => void;
   onRetry: (isRetry?: boolean) => void;
   onSkip: () => void;
+  deviceModel: DeviceModel;
 };
 
 const DeviceCancel = ({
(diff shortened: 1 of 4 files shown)


apps/ledger-live-desktop/src/renderer/modals/UpdateFirmwareModal/errors/DeviceError.tsx
```diff
@@ -57,6 +60,10 @@ const DeviceCancel = ({
     ? t("common.retry")
     : t("manager.firmware.restartUpdate");
 
+  if (typeof error === "object" && error) {
+    (error as Error & { deviceName?: string }).deviceName = deviceModel.productName;
```
gre, line 64: (this is too much an hack. we don't want to do this i think)

PR #6523, opened 2024-03-22
chore: remove unused Account#freshAddresses field

gre:
### ✅ Checklist

- [ ] `npx changeset` was attached. (no need)
- [x] **Covered by automatic tests.** if it builds, it's good. the field is not used.
- [x] **Impact of the changes:** no changes

### 📝 Description

`freshAddresses` field on Account is not used today nor planned by product anytime soon, so we can completely drop the feature for now. this work is part of Account model simplification.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-11772

---


KVNLS (commented):
@gre do you know if integration tests are failing following your changes?

libs/ledger-live-common/src/families/near/__snapshots__/bridge.integration.test.ts.snap
```diff
@@ -25,7 +19,7 @@ exports[`near currency bridge scanAccounts near seed 1 1`] = `
         {
           "available": "10000000000000000001",
           "pending": "0",
-          "rewards": "1",
+          "rewards": "10000000000000000001",
```
sprohaszka-ledger, line 22: ?! Strange...

libs/ledger-live-common/src/families/cosmos/datasets/__snapshots__/coreum.integration.test.ts.snap
```diff
@@ -85,108 +67,8 @@ exports[`coreum currency bridge scanAccounts coreum seed 1 1`] = `
...
-        "core10l6h3qw05u7qduqgafj8wlrx3fjhr852z7r5l2",
-      ],
-      "standard": undefined,
-      "tokenId": undefined,
-      "transactionSequenceNumber": 0,
-      "type": "IN",
-      "value": "1000000",
-    },
-  ],
+  [],
```
sprohaszka-ledger, line 70: ?! What happened here?

sprohaszka-ledger (commented):
There are some strange modification in the snapshots file.

gre:
in this PR, i have huge changes due to re-running the tests to have up to date snapshots.
the problem is that, many of these snapshots are already outdated on develop.
so maybe, i will change a bit the strategy on this and first send a PR that just update the snapshot against current's develop?

that might highlights the PR feebacks from you both @sprohaszka-ledger @KVNLS  because some of these changes actually are not related to the field removal.

gre:
^FYI this PR is in standby until we solve this

gre:
This PR is ready again, develop is up to date with integration tests 🤞

KVNLS (commented):
LGTM 🚀 Nice cleanup!

libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
```diff
@@ -1,6 +1,6 @@
 import { BigNumber } from "bignumber.js";
 import flatMap from "lodash/flatMap";
-import { Account, Address, Operation, OperationType } from "@ledgerhq/types-live";
```
sprohaszka-ledger, line 3: Why remove it and change the signature of the `getAddress` method? It is more synthetic

sprohaszka-ledger (commented):
One quick last comment

hedi-edelbloute (approved):

valpinkman (approved):

libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
gre: this type { freshAddress, freshAddressPath } and getAddress function is now strictly internal to the coin, i didn't want to change it too much, but it's possible that coin could simplify even further this: there is no need to return Address anymore there, they could just afford to be a getAddress(path)=>address actually, since we don't have to push them into a freshAddresses array.
Anyway as it is not used externally, on types-live perspective there are no need to have this concept of 'Address' pair, so indeed took a direction to remove what's not useful from the type side.
we could always reintroduce the concept if we want in future, like if we want to reorganize the field together, but today in the Account type, they are just nested as part of Account.
also to be clear: there is no type regression on this, we still have the same type safety, it allows coins to do the type they want.

PR #6524, opened 2024-03-22
chore: sunset unused ChildAccount type

gre:
### ✅ Checklist

- [ ] `npx changeset` was attached. (not needed)
- [x] **Covered by automatic tests.** if it builds, it should be good.
- [x] **Impact of the changes:** none (unused)

### 📝 Description

We introduced the concept of ChildAccount back in 2020 on the first implementation of Tezos which had this concept of KT Accounts. These not being strictly a token, but a "sub account of tezos", it made sense at the time to diverge from the Token Account.
Since the JS rework of tezos (2021?) as well as Tezos' sunset of kt accounts (delegation now directly happens on the main account), we no longer needed to support KT Accounts. However that type was still forgotten in our type model and became tech debt.

We can now remove it, this makes possible simplifications in the future, you can notably observe that the type `SubAccount` is now strictly equivalent to `TokenAccount`. 

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-11884

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/SelectAccount.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/SelectAccount.tsx
@@ -106,11 +106,8 @@ export const AccountOption = React.memo<AccountOptionProps>(function AccountOpti
   const currency = getAccountCurrency(account);
   const unit = getAccountUnit(account);
   const name = getAccountName(account);
-  const nested = ["TokenAccount", "ChildAccount"].includes(account.type);
-  const balance =
-    account.type !== "ChildAccount" && account.spendableBalance
-      ? account.spendableBalance
-      : account.balance;
+  const nested = "TokenAccount" === account.type;
+  const balance = account.spendableBalance || account.balance;
   const textContents = singleLineLayout ? (
     <>
       <Box flex="1" horizontal alignItems="center">
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/TargetAccountDrawer/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/TargetAccountDrawer/index.tsx
@@ -71,10 +71,7 @@ const TargetAccount = memo(function TargetAccount({
   const parentAccount =
     account?.type !== "Account" ? allAccounts?.find(a => a.id === account?.parentId) : null;
   const parentName = parentAccount ? getAccountName(parentAccount) : undefined;
-  const balance =
-    account.type !== "ChildAccount" && account.spendableBalance
-      ? account.spendableBalance
-      : account.balance;
+  const balance = account.spendableBalance || account.balance;
 
   const onClick = useCallback(() => {
     track("button_clicked2", {
--- a/apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
+++ b/apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
@@ -5,7 +5,7 @@ import {
   getAccountName,
   getAccountUnit,
 } from "@ledgerhq/live-common/account/index";
-import { TokenAccount, AccountLike, ChildAccount, DerivationMode } from "@ledgerhq/types-live";
+import { TokenAccount, AccountLike, DerivationMode } from "@ledgerhq/types-live";
 import { getTagDerivationMode } from "@ledgerhq/coin-framework/derivation";
 import { useSelector } from "react-redux";
 import { GestureResponderEvent } from "react-native";
@@ -45,9 +45,7 @@ const AccountRow = ({
   // makes it refresh if this changes
   useEnv("HIDE_EMPTY_TOKEN_ACCOUNTS");
   const currency = getAccountCurrency(account);
-  const parentAccount = useSelector((state: State) =>
-    parentAccountSelector(state, { account: account as ChildAccount }),
-  );
+  const parentAccount = useSelector((state: State) => parentAccountSelector(state, { account }));
 
   const name = getAccountName(account);
   const unit = getAccountUnit(account);
--- a/libs/coin-framework/src/account/helpers.test.ts
+++ b/libs/coin-framework/src/account/helpers.test.ts
@@ -1,5 +1,5 @@
 import { CryptoCurrency, TokenCurrency, Unit } from "@ledgerhq/types-cryptoassets";
-import { Account, ChildAccount, Operation, SubAccount, TokenAccount } from "@ledgerhq/types-live";
+import { Account, Operation, SubAccount, TokenAccount } from "@ledgerhq/types-live";
 import BigNumber from "bignumber.js";
 import {
   areAllOperationsLoaded,
@@ -11,14 +11,9 @@ import {
   getFeesCurrency,
 } from ".";
 import { isAccountEmpty, isAccountBalanceSignificant, clearAccount } from "./helpers";
+import { getTokenById } from "../currencies";
 
 const mockAccount = {} as Account;
-const childAccount = {
-  ...mockAccount,
-  type: "ChildAccount",
-  parentId: "sampleId",
-  address: "sampleAddress",
-} as ChildAccount;
 const tokenAccount = {
   ...mockAccount,
   type: "TokenAccount",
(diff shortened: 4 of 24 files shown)


libs/coin-framework/src/account/helpers.test.ts
```diff
@@ -241,40 +185,13 @@ describe(getAccountSpendableBalance.name, () => {
...
-
   describe("given a TokenAccount", () => {
     it("should return the token account spendable balance", () => {
       const sampleAccountBalance = new BigNumber(10);
       tokenAccount.spendableBalance = sampleAccountBalance;
       expect(getAccountSpendableBalance(tokenAccount)).toEqual(sampleAccountBalance);
     });
   });
-
-  describe("given an unknown type Account", () => {
```
sprohaszka-ledger, line 260: I don't understand why you remove those tests. They are limited to `ChildAccount` (it is not obvious)?

libs/coin-framework/src/account/helpers.test.ts
```diff
@@ -38,40 +33,13 @@ describe(getAccountCurrency.name, () => {
...
-
   describe("given a TokenAccount", () => {
     it("should return the token currency", () => {
       const sampleToken = { id: "tokenId" } as TokenCurrency;
       tokenAccount.token = sampleToken;
       expect(getAccountCurrency(tokenAccount)).toEqual(sampleToken);
     });
   });
-
-  describe("given an unknown type Account", () => {
```
sprohaszka-ledger, line 57: I don't understand why you remove those tests. They are limited to ChildAccount (it is not obvious)?

apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
```diff
@@ -46,7 +46,7 @@ const AccountRow = ({
   useEnv("HIDE_EMPTY_TOKEN_ACCOUNTS");
   const currency = getAccountCurrency(account);
   const parentAccount = useSelector((state: State) =>
-    parentAccountSelector(state, { account: account as ChildAccount }),
+    parentAccountSelector(state, { account: account as TokenAccount }),
```
sprohaszka-ledger, line 49: Why force cast with a new type? The removed one should imply that the `account` cannot be cast, no?

sprohaszka-ledger (approved):
Thanks for lighten the model :D 
I am not fully sure about the WalletAPI part. I'd let @Justkant give you his opinion.

hedi-edelbloute (commented):
I'm scared we someday need to use child account for family accounts with currency subAccounts, if it works i'd keep it

libs/ledgerjs/packages/types-live/src/account.ts
```diff
@@ -183,16 +160,18 @@ export type Account = {
   nfts?: ProtoNFT[];
 };
 
-/** super type that is either a token or a child account */
-export type SubAccount = TokenAccount | ChildAccount;
+/**
+ * deprecated use TokenAccount instead
```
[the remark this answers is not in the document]
gre: @sprohaszka-ledger unfortunately can't do this due to a bug with `documentation` that will completely break the README. i lost half a day trying to make this possible. I think @hedi-edelbloute is aware of this problem as it arise in the past too. very unfortunate

libs/coin-framework/src/account/helpers.test.ts
gre: because there is no longer an unknown type. it only can be `TokenAccount`. This test was covering the case where user data was having a wrong data, which shouldn't happen. in the new implementation we will ignore it from the import.

libs/coin-framework/src/account/helpers.test.ts
gre: (answered above)

apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
gre: i'm not familiar with this specific code. 
I will create a ticket on Jira to investigate why we need this casting / if it can be fixed, but I think out of scope of the type simplification of this PR I think.  
The flaw is in the existing code of develop, it's just that here the new ChildAccount is now TokenAccount, it seems like we loosely expect account is a token account here and we need this cast to not break the typing.

gre:
@hedi-edelbloute was we discussed in DM too: ChildAccount are sunset since 2021, unfortunately we just forgot this tech debt in the past, but i'm almost certain that our UI have not survived the successive rework and will not support out of the box this idea from the past. It would be easier to just reintroduce the concept whenever you re-need it in the future, but i think we have to sunset what is unused first.

apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
sprohaszka-ledger: I mean also casting from `ChildAccount` (which is a specific type) to `TokenAccount` is quite weird as it is not the same type.

Justkant (dismissed):

apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
gre: in this PR, ChildAccount === TokenAccount

Justkant (dismissed):

apps/ledger-live-mobile/src/screens/Accounts/AccountRow.tsx
gre: ooh I see what you mean. i think it was a mistake in the past indeed, instead of ChildAccount we meant to do SubAccount.
because the parentAccountSelector takes a SubAccount out there. and the ChildAccount was type matching to be a SubAccount, the problem is that, indeed, casting hide the type issue in this place.

It isn't typed well at this selector side, it could accept an AccountLike so we get rid of this casting

```js
export const parentAccountSelector = createSelector(
  accountsSelector,
  (_: State, { account }: { account?: SubAccount }) => (account ? account.parentId : null),
  (accounts, accountId) => accounts.find(a => a.id === accountId),
);```

**i'm pushing a fix.**

*(side note: the usecase for such a selector is very weird, we shouldn't have to infer back the parent, it should be provided in the first place... but i'm not going to look at this now)*

**Review by hedi-edelbloute — DISMISSED** [review]

File: libs/ledger-live-common/src/csvExport.ts [hunk 5]
```diff
@@ -50,7 +50,7 @@ const fields: Field[] = [
   {
     title: "Operation Fees",
     cell: (account, parentAccount, op) =>
-      ["TokenAccount", "ChildAccount"].includes(account.type)
+      ["TokenAccount"].includes(account.type)
```
mcayuelas-ledger, line 53: could be 
```suggestion
      account.type === "TokenAccount"
```

apps/ledger-live-desktop/src/renderer/components/SelectAccount.tsx
```diff
@@ -106,11 +106,8 @@ export const AccountOption = React.memo<AccountOptionProps>(function AccountOpti
   const currency = getAccountCurrency(account);
   const unit = getAccountUnit(account);
   const name = getAccountName(account);
-  const nested = ["TokenAccount", "ChildAccount"].includes(account.type);
-  const balance =
-    account.type !== "ChildAccount" && account.spendableBalance
-      ? account.spendableBalance
-      : account.balance;
+  const nested = ["TokenAccount"].includes(account.type);
```
sprohaszka-ledger, line 109: This is simpler no?
```js
const nested = account.type === "TokenAccount"
```

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/TargetAccountDrawer/index.tsx
```diff
@@ -71,10 +71,7 @@ const TargetAccount = memo(function TargetAccount({
   const parentAccount =
     account?.type !== "Account" ? allAccounts?.find(a => a.id === account?.parentId) : null;
   const parentName = parentAccount ? getAccountName(parentAccount) : undefined;
-  const balance =
-    account.type !== "ChildAccount" && account.spendableBalance
-      ? account.spendableBalance
-      : account.balance;
+  const balance = account.spendableBalance ? account.spendableBalance : account.balance;
```
sprohaszka-ledger, line 74: Simpler, no ?
```js
const balance = account.spendableBalance || account.balance
```

libs/ledger-live-common/src/csvExport.ts
sprohaszka-ledger, line 53: Why don't use `===` ?

sprohaszka-ledger (dismissed):
Some nitpicking ;)

hedi-edelbloute (approved):

Justkant (approved):

sprohaszka-ledger (approved):
👍

CremaFR (approved):

mcayuelas-ledger (approved):

PR #6528, opened 2024-03-23
[LLD] Dynamically set the minimum of global summary chart

dkaplan65:
### ✅ Checklist

  - Sets the `suggestedMin` of the chart in the global summary to be the minimum of the `balanceHistory` that we are currently looking at

### 📝 Description

It is difficult to view the small changes in the balance if the values are much bigger than 0. Ideally, we are able to view the chart values in the relative range that we are currently viewing, making the viewing experience better.

## Old view:
Screenshot 2024-03-23 at 4 35 04 PM

## New view:
Screenshot 2024-03-23 at 4 34 44 PM

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/dashboard/GlobalSummary.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/dashboard/GlobalSummary.tsx
@@ -24,6 +24,11 @@ export default function PortfolioBalanceSummary({ range, chartColor, counterValu
     (val: number | string) => formatShort(counterValue.units[0], BigNumber(val)),
     [counterValue],
   );
+
+  const suggestedMin = Math.max(
+    portfolio.balanceHistory.reduce((a, b) => (b.value < a ? b.value : a), Infinity),
+    0,
+  );
   const dayFormatter = useDateFormatter(dayFormat);
   const hourFormatter = useDateFormatter(hourFormat);
   const renderTooltip = useCallback(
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/screens/dashboard/GlobalSummary.tsx
```diff
@@ -24,6 +24,7 @@ export default function PortfolioBalanceSummary({ range, chartColor, counterValu
     (val: number | string) => formatShort(counterValue.units[0], BigNumber(val)),
     [counterValue],
   );
+  const suggestedMin = Math.max(Math.min(...portfolio.balanceHistory.map(b => b.value)), 0);
```
gre, line 27: could you rework this with a reduce instead? It is generally considered unsafe to spread arguments on min / max because it can fails on certain array size.
( ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max )

Capture d’écran 2024-03-25 à 10 39 22

gre (changes requested):

gre:
**However, before considering this feature, I will have to discuss with our Product, we actually had this in the past and it was decided to rollback the idea. I think there are pros and cons and i'll suggest to product if we couldn't make this a feature by having a User settings that users can configure their preferred way.**

dkaplan65:
> Thanks for the contribution.
> 
> before considering this feature, I will inform our Product team, because we actually had this in the past and it was decided to rollback the idea. I think there are pros and cons (on non zero based axis) and i'll suggest to product if we couldn't make this a feature by having a User settings that users can configure their preferred way.

I reworked the `Math.min` into a reduce.

I would greatly appreciate it! Please let me know what Product says. Thanks

PR #6566, opened 2024-03-27
support: prevent unnecessary http call retries when 422

cgrellard-ledger:
### ✅ Checklist

  - `promises` and `network` libraries

### 📝 Description

Added a `retryCondition` optional parameter to the `retry` function from the `promises` library to be able to prevent retries based on a condition. It is specifically useful to prevent retries of http calls for requests that returns a 422.

|Screenshot 2024-03-27 at 17 28 59 |Screenshot 2024-03-27 at 18 34 00|

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11527]

---

[LIVE-11527]: https://ledgerhq.atlassian.net/browse/LIVE-11527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/live-network/src/network.ts
+++ b/libs/live-network/src/network.ts
@@ -169,6 +169,12 @@ const implementation = <T = any>(arg: AxiosRequestConfig): AxiosPromise<T> => {
 
     promise = retry(() => axios(arg), {
       maxRetry: getEnv("GET_CALLS_RETRY"),
+      retryCondition: error => {
+        if (error && error.status) {
+          return error.status !== 422;
+        }
+        return true;
+      },
     });
   } else {
     promise = axios(arg);
--- a/libs/live-network/src/network.ts
+++ b/libs/live-network/src/network.ts
@@ -171,6 +171,8 @@ const implementation = <T = any>(arg: AxiosRequestConfig): AxiosPromise<T> => {
       maxRetry: getEnv("GET_CALLS_RETRY"),
       retryCondition: error => {
         if (error && error.status) {
+          // A 422 shouldn't be retried without change as explained in this documentation
+          // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
           return error.status !== 422;
         }
         return true;
--- a/libs/promise/src/promise.test.ts
+++ b/libs/promise/src/promise.test.ts
@@ -1,20 +1,70 @@
-import { promiseAllBatched } from "./promise";
-test("promiseAllBatched", async () => {
-  const promisifyIdPlusOne = (a: number) => Promise.resolve(a + 1);
+import { promiseAllBatched, retry } from "./promise";
 
-  const p = promiseAllBatched(5, [], promisifyIdPlusOne);
-  expect(typeof p.then).toBe("function");
-  expect(await p).toEqual([]);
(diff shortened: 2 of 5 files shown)


LucasWerey (dismissed):

libs/promise/src/promise.test.ts
```diff
@@ -1,20 +1,70 @@
...
+          retryCondition: e => {
+            expect(e).toBe(0);
+            return false;
+          },
+        });
+      } catch (e) {
+        expect(e).toBe(0);
+      }
+      expect(numberOfExecutions).toBe(1);
+    });
```
lambertkevin, line 68: Thanks for adding tests 👌 
My only concern here is that you're just testing the `retryCondition` gets executed, which is already a good thing, but there is no test creating the specifics of what you added in `retryCondition`. 
Could you please add a test mocking a network call returning a 422 status code just to make sure we the day someone changes that `retryCondition` logic, they'll notice that it was an expected behaviour ?

libs/live-network/src/network.ts
```diff
@@ -168,6 +168,12 @@ const implementation = <T = any>(arg: AxiosRequestConfig): AxiosPromise<T> => {
 
     promise = retry(() => axios(arg), {
       maxRetry: getEnv("GET_CALLS_RETRY"),
+      retryCondition: error => {
+        if (error && error.status) {
+          return error.status !== 422;
+        }
+        return true;
+      },
```
lambertkevin, line 176: As a general comment, I feel like this a very specific to a single endpoint behaviour ? I'm not super familiar with 422 errors, but it feels to me like there is no specific reason to not retry on this more than on 404 for example. 
Couldn't we add that `retryCondition` as an optional argument of `network` so that this can be adapted to the different responses possible for a request ?

cgrellard-ledger: As per [this documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422), a 422 Unprocessable content shouldn't be retried without modification so it makes sense to me to make it general and not optional. What do you think ?

libs/promise/src/promise.test.ts
cgrellard-ledger: Yep, I'll add a test for this

libs/live-network/src/network.ts
lambertkevin: Ah, interesting ! I'm more blind everyday and I didn't notice the big red warning 😆
By reading it I see that there is still an option to retry it based on a response header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After (which I guess would make the retryCondition way harder to use, so just mentioning it as I think it's weird to document "you shouldn't retry this as is. Except after this date. 🤔)

I'm still trying to get my head around a 422 is less "retriable" than a 404 but I guess that's another question. 

Can you please just add a comment explaining that 422 are documented this way and link the doc ? Just to make sure someone else will understand this without context and not brain fart like I did 👍

lambertkevin (approved):
LGTM 👌

mcayuelas-ledger (approved):

sshmaxime (approved):

lambertkevin (approved):

PR #6604, opened 2024-04-03
[SUPPORT]: Remove .unit usage from account.unit

mcayuelas-ledger:
### ✅ Checklist

  - Remove usage of account.unit or getAccountUnit and replaced it with a more centralised way in Settings for each currency

### 📝 Description

Move "unit" field out of Account and move the related settings from Account settings back to the Currency settings (LLD only currently, LLM will be next)

Screenshot 2024-04-03 at 14 45 04

https://github.com/LedgerHQ/ledger-live/assets/112866305/bd0d409d-4287-4b19-9e8d-66913405f6da

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11770](https://ledgerhq.atlassian.net/browse/LIVE-11770) 

---

[LIVE-11770]: https://ledgerhq.atlassian.net/browse/LIVE-11770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
@@ -18,8 +18,7 @@ import {
   addNewDeviceModel,
 } from "~/renderer/actions/settings";
 import {
-  CurrencySettings,
-  currencySettingsLocaleSelector,
+  storeSelector as settingsSelector,
   preferredDeviceModelSelector,
 } from "~/renderer/reducers/settings";
 import { DeviceModelId } from "@ledgerhq/devices";
@@ -76,10 +75,9 @@ import { Transaction, TransactionStatus } from "@ledgerhq/live-common/generated/
 import { AppAndVersion } from "@ledgerhq/live-common/hw/connectApp";
 import { Device } from "@ledgerhq/types-devices";
 import { LedgerErrorConstructor } from "@ledgerhq/errors/lib/helpers";
-import { Currency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
+import { TokenCurrency } from "@ledgerhq/types-cryptoassets";
 import { isDeviceNotOnboardedError } from "./utils";
 import { useKeepScreenAwake } from "~/renderer/hooks/useKeepScreenAwake";
-import { State } from "~/renderer/reducers";
 
 type LedgerError = InstanceType<LedgerErrorConstructor<{ [key: string]: unknown }>>;
 
@@ -232,14 +230,7 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
   const dispatch = useDispatch();
   const preferredDeviceModel = useSelector(preferredDeviceModelSelector);
   const swapDefaultTrack = useGetSwapTrackingProperties();
-
-  const useCurrencySettingsSelector = (currency: Currency): CurrencySettings => {
-    const currencySettings = useSelector((state: State) =>
-      currencySettingsLocaleSelector(state.settings, currency),
-    );
-
-    return currencySettings;
-  };
+  const stateSettings = useSelector(settingsSelector);
 
   const type = useTheme().colors.palette.type;
 
@@ -376,7 +367,7 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
           swapDefaultTrack,
           amountExpectedTo: amountExpectedTo.toString() ?? undefined,
           estimatedFees: estimatedFees?.toString() ?? undefined,
-          getCurrencySettings: useCurrencySettingsSelector,
+          stateSettings,
         });
       }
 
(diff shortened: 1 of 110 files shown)


apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -362,6 +375,7 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
           swapDefaultTrack,
           amountExpectedTo: amountExpectedTo.toString() ?? undefined,
           estimatedFees: estimatedFees?.toString() ?? undefined,
+          getCurrencySettings: useCurrencySettingsSelector,
```
gre, line 378: we probably should rename the `getCurrencySettings` function with a `useCurrencySettings` naming? (in order for linter to correctly apply the use* rules)

apps/ledger-live-desktop/src/renderer/components/OperationsList/Operation.tsx
```diff
@@ -49,69 +44,58 @@ type OwnProps = {
...
+      />
+      {withAccount && <AccountCell accountName={getAccountName(account)} currency={currency} />}
+      {withAddress ? <AddressCell operation={operation} /> : <Box flex="1" />}
+      <AmountCell operation={operation} currency={currency} unit={unit} isConfirmed={isConfirmed} />
+    </OperationRow>
+  );
 }
-const ConnectedOperationComponent: React.ComponentType<OwnProps> =
-  connect(mapStateToProps)(OperationComponent);
+const ConnectedOperationComponent: React.ComponentType<OwnProps> = OperationComponent;
```
gre, line 100: to make it iso, can you React.memo the component? since it was before a PureComponent

apps/ledger-live-desktop/src/renderer/components/TokenRow.tsx
```diff
@@ -38,29 +39,26 @@ const NestedRow = styled(Box)`
...
+      <Countervalue account={account} currency={currency} range={range} />
+      <Delta account={account} range={range} />
+      <Star
+        accountId={account.id}
+        parentId={account.type !== "Account" ? account.parentId : undefined}
+      />
+    </Row>
+  );
 }
 export default TokenRow;
```
gre, line 64: React.memo it

apps/ledger-live-desktop/src/renderer/drawers/DataSelector/AccountList.tsx
```diff
@@ -78,6 +85,8 @@ export function AccountList({ currency, onAccountSelect, accounts$ }: Props) {
       {accountTuples.map((accountTuple, index) => {
         const { account, subAccount } = accountTuple;
         const accountCurrency = getAccountCurrency(subAccount || account);
+        const unit = currencySettings.unit;
```
gre, line 88: code is no longer equivalent.
unit is supposed to be inferred from the `accountCurrency` but here you use a global currencySettings one that is fixed for one currency? do we have guarantee it's the same currency?

apps/ledger-live-desktop/src/renderer/families/bitcoin/FullNode/steps/StepAccounts.tsx
```diff
@@ -36,7 +37,11 @@ const Accounts = ({
   // FIXME Not using the AccountList component because styles differ quite a bit, we should unify.
   const accounts = useSelector(accountsSelector);
   const currency = getCryptoCurrencyById("bitcoin");
+
   const bitcoinAccounts = accounts.filter(a => getAccountCurrency(a) === currency);
+
+  const unit = useAccountUnit(bitcoinAccounts[0]);
```
gre, line 43: `[0]` can be falsy so this code seems unsafe? *(btw FullNode is likely untestable because support was dropped but code is still there..)*

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
mcayuelas-ledger: Can't do it because it throw a lint error about usage Hook

React Hook "useCurrencySettings" is called in function "renderSwapDeviceConfirmation" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".eslint[react-hooks/rules-of-hooks](https://reactjs.org/docs/hooks-rules.html)

gre: ok, then i think it's a problem, renderSwapDeviceConfirmation must not use hooks, since they are indeed not a component.

we probably need to pass in the whole Settings store, and lense from it the necessary information.

This is what i had to do for the name too: https://github.com/LedgerHQ/ledger-live/blob/398e3e09e55db9369bbd861eaf8b384e77220704/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx#L367

gre:
=> https://github.com/LedgerHQ/ledger-live/pull/6796

PR #6606, opened 2024-04-03
[LIVE-9595] Rework QueuedDrawer + Fix navigation when app update is needed

ofreyssinet-ledger:
### ✅ Checklist

  - Navigation from a device action drawer to My Ledger
  - QueuedDrawers: all drawers in LLM

### 📝 Description

Initially, this PR is a fix for navigating to the `MyLedgerChooseDevice` screen which itself should navigate to `MyLedgerDevice` and open the "app update" drawer.
The only thing for this fix is to add the `device` parameter that was missing in `navigation.navigate` parameters.

**However**, I noticed that **navigating from a screen with an open drawer** (in this case a `DeviceActionModal`, in any screen) **to another screen that opens another drawer right away** (again here a `DeviceActionModal`, in `MyLedgerChooseDevice` screen, that establishes the secure channel and then redirects to `MyLedgerDevice`) was causing a big issue:
- The drawer being opened while the first one is still closing (even though we switched screens) will not appear visually, even though its children are mounted. And while this is the case, the whole space of the screen below the navigation header will be unresponsive to touch events.

So in order to avoid a hacky solution involving an arbitrary timeout, **I reimplemented QueuedDrawer from scratch**, as the existing implementation was convoluted and a bit hard to understand, using custom `useFocusNavigation` and lots of bits of internal state.
Hopefully the new implementation is hopefully cleaner and easier to understand.

**QueuedDrawer new logic:**
- **Main principle:** two drawers must never be opened at the same time.
- Therefore, there is a unique queue of drawers, handled by an orchestrator (a `QueuedDrawerContext.Provider`).
- The first drawer in the queue opens as soon as it reaches that "first drawer" position.
- When that drawer closes (and only when it effectively is closed), it tells the orchestrator that it is done, so the orchestrator can remove it from the queue, and signal the following drawer to open.
- Any drawer can ask to be "forced opened".
  - When this happens, the entire queue is cleared.
  - The currently open drawer gets closed, and removes itself from the queue onces it's closed.
  - The "forced" drawer is finally opened.
- Navigation:
  - all drawers in a given screen, opened or requesting or forcing to be opened, get closed and removed from the queue as that screen loses focus.
  - drawers that are not embedded in a navigation screen are not affected by navigation events, but the queuing and forcing logic still fully applies to them.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-9595] 

---

[LIVE-9595]: https://ledgerhq.atlassian.net/browse/LIVE-9595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre (dismissed):

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/QueuedDrawersContextProvider.tsx
```diff
@@ -0,0 +1,108 @@
...
+} from "./QueuedDrawersContext";
+
+type QueueItem = {
+  id: string;
+  setDrawerOpenedCallback: SetDrawerOpenedCallback;
+  markedForClose?: boolean;
+};
+export type DrawersQueue = Array<QueueItem>;
+
+export function logDrawer(
```
aussedatlo, line 18: [COULD] move `logDrawer` from the provider to a folder `utils`

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/QueuedDrawersContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+export type DrawerInQueue = {
+  removeDrawerFromQueue: () => void;
+  getPositionInQueue: () => number;
+};
+
+type QueuedDrawersContextType = {
+  /**
+   * Add a drawer to the queue.
+   * @param setDrawerOpenedCallback The callback to call when the drawer should be opened or closed.
```
aussedatlo, line 21: [ASK] Is `setDrawerOpenedCallback` not misleading because this callback is called on a open and a close ?

aussedatlo (approved):
Nice to see the new arch in place and a real context for drawers 🚀

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/QueuedDrawersContext.tsx
aussedatlo: `OnDrawerStateChange` ?

ofreyssinet-ledger: Good one thanks 👍

aussedatlo (approved):

cgrellard-ledger (approved):

PR #6616, opened 2024-04-05
no strict code ownerships on the end 2 end tests of playwright/detox

gre:
### ✅ Checklist

- [x] **Impact of the changes:** none

### 📝 Description

We have noticed in recent PRs that a bunch of PR auto assign live-hub team just because they add tests in their own scope in the generic end 2 end tests of LLD and LLM. Similarly to what we did for i18n wordings, errors list, navigations,... there are no strict codeowners of these parts. the End 2 end tests on the end product are actually concerned by all teams (unlike the UI tests on the component system for instance, that indeed belong to live-hub), therefore  we should not have any strict ownership on these paths: actually when devs will write tests in context of adding new feature or modifying some codebase in `src/`, the ownership will be automatically assigned through these src/* changes – and they generally relates to the same tests modified – some tests being "generic" (e.g snapshotting the dashboard/account screen can be impacted by a rework of a coin (recent example of https://github.com/LedgerHQ/ledger-live/pull/6609), but only the blockchain team is actually relevant to determine if the screenshot change is legit in context of that PR change).

### ❓ Context

- **JIRA or GitHub link**: na

---

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -122,11 +122,6 @@ apps/ledger-live-mobile/src/components/ServicesWidget/Protect/          @ledgerh
 apps/ledger-live-mobile/src/components/WebRecoverPlayer/                @ledgerhq/recover-software
 apps/ledger-live-mobile/src/screens/Protect/                            @ledgerhq/recover-software
 libs/ledger-live-common/src/hooks/recoverFeatureFlag.ts                 @ledgerhq/recover-software
-
-# UI tests
-apps/ledger-live-desktop/tests/                          @ledgerhq/live-hub @ledgerhq/live-qa
-apps/ledger-live-mobile/e2e/                             @ledgerhq/live-hub @ledgerhq/live-qa
-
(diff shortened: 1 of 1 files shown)


aboissiere-ledger (changes requested):
I agree to remove the live-qa team but not to leave these tests with no owners as the dev teams are responsible for them. Please refer to https://docs.google.com/spreadsheets/d/1okt8STPc0j_ubRRq0GEwz4UBPZKSqJx_M9C9d6uaLhs/edit#gid=243842674

gre:
as shared in description: the modification of these tests are done usually in context of modification of features, and that create a hub amount of PR ownership ping due to this, whereas, like we did for i18n files, the actual ownership is determined by the src/* files that are changed.
if we don't agree on a solution here to remove hub ownership, i close the PR for now, i see no benefit on just removing live-qa and not live-hub as the main problem we face is to be pinged by a lot of PRs

PR #6704, opened 2024-04-17
[Support/Filecoin] Add ERC20 tokens support for Filecoin

lawRathod:
### ✅ Checklist

  - ...

### 📝 Description

Add ERC20 token support for filecoin network. Adds new tokens specific to filecoin and enable send/receive. 

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/ledger-live-common/src/families/filecoin/bridge/utils/addresses.ts
```diff
@@ -32,3 +57,50 @@ export const validateAddress = (input: string): ValidateAddressResult => {
...
+  }
+
+  const valid = validateAddress(addr);
+  if (!valid.isValid) {
+    return false;
+  }
+
+  const accountType = addr.substring(0, 2);
+  if (["f0", "0x", "f4"].includes(accountType)) {
+    return true;
```
hzheng-ledger, line 73: if I understand, an address begins with "f0" or "f4", we should convert them into eth address and check whether it is a valid one. right?
we can't consider all the addresses begin with "f0" or "f4" are valid. right?

libs/ledger-live-common/package.json
```diff
@@ -240,6 +242,7 @@
     "tsx": "^4.7.1",
     "usehooks-ts": "^2.13.0",
     "utility-types": "^3.10.0",
+    "web3": "^4.7.0",
```
hzheng-ledger, line 245: we use 
the "ethers" lib instead of "web3" in our codebase.

i.e.
https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-evm/package.json#L75

could you plz change it to "ethers" lib to be coherent with the rest of our codebase? Thanks a lot

libs/ledger-live-common/package.json
```diff
@@ -131,6 +131,7 @@
     "@elrondnetwork/erdjs": "11.0.0",
     "@elrondnetwork/erdjs-network-providers": "^1.1.2",
     "@hashgraph/sdk": "2.14.2",
+    "@ipld/dag-cbor": "^9.2.0",
```
hzheng-ledger, line 134: I don't find it is used in the current code. Isn't it?

libs/ledger-live-common/src/families/filecoin/bridge/utils/erc20/tokenAccounts.ts
```diff
@@ -0,0 +1,179 @@
...
+
+export const encodeTxnParams = (abiEncodedParams: string) => {
+  if (!abiEncodedParams) {
+    throw new Error("Cannot encode empty abi encoded params");
+  }
+
+  const buffer = Buffer.from(abiEncodedParams.slice(2), "hex"); // buffer/byte array
+  const dataEncoded = cbor.encode(buffer);
+
+  return base64.stringify(dataEncoded);
```
hzheng-ledger, line 160: I think we can use
Buffer.from
and
buffer.toString('base64');

instead, so that we don't have to use "rfc4648" lib

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -5495,6 +5495,9 @@
       "title": "Impossible to calculate amount and fees",
       "description": "Impossible to calculate amount and fees"
     },
+    "InvalidRecipientForTokenTransfer": {
```
hzheng-ledger, line 5498: should we do the same translation for Ledger Mobile?

libs/ledger-live-common/src/families/filecoin/bridge/utils/erc20/tokenAccounts.ts
```diff
@@ -0,0 +1,179 @@
...
+      if (!token) {
+        log("error", `filecoin token not found, addr: ${cAddr}`);
+        continue;
+      }
+
+      const balance = await fetchERC20TokenBalance(nativeEthAddr, cAddr);
+      const bnBalance = new BigNumber(balance.toString());
+      const tokenAccountId = encodeTokenAccountId(parentAccountId, token);
+
+      const operations = txns
```
hzheng-ledger, line 114: Just a question: For the subaccount operations, are all the addresses retrieved from the explorer endpoint expressed in the Ethereum format instead of the Filecoin format?

If a user clicks the "receive" button, does he get an Ethereum format address or a Filecoin format address?

Do you think it is better to convert back to the Filecoin format in the UI? If a user sends ERC20 tokens using his Filecoin account in the UI, I think he expects to see his Filecoin address as the sender in the operation list.

hzheng-ledger (changes requested):
Please have a look at my comments. Thanks a lot

libs/ledger-live-common/src/families/filecoin/bridge/utils/addresses.ts
lawRathod: you understand correct, for token transfer we need the address to be eth style (0x) not `f[0,1,4]`. In filecoin ecosystem there's ways to convert all native addresses to eth style but f1 is exception which requires network call (insecure) to get equivalent f0 and then f0 -> 0x (secure). 

For addresses beginning with f1 we do not want to support them for token transfer because network call can be insecure in practice and lead to possible known attacks. For when invalid we therefore have this [error](https://github.com/LedgerHQ/ledger-live/blob/a4344ecb8d7e980c48c6b450b7021119681ffa8e/apps/ledger-live-desktop/static/i18n/en/app.json#L5498).

libs/ledger-live-common/package.json
lawRathod: Definitely, now using ethers instead of web3.

libs/ledger-live-common/package.json
lawRathod: It's definitely not, now removed from package.json... it was an accident 😅

libs/ledger-live-common/src/families/filecoin/bridge/utils/erc20/tokenAccounts.ts
lawRathod: Great tip, I didn't know it could be done this way. Now using this way and the `rfc4648` is removed from the package.json. 😄

apps/ledger-live-desktop/static/i18n/en/app.json
lawRathod: Yes definitely, I am sorry it wasn't there but now I have added it.

libs/ledger-live-common/src/families/filecoin/bridge/utils/erc20/tokenAccounts.ts
lawRathod: > For the subaccount operations, are all the addresses retrieved from the explorer endpoint expressed in the Ethereum format instead of the Filecoin format?

Yes they are, the erc20 endpoints on indexing only stores the address data from the evm world. The translation isn't yet there. 

> If a user clicks the "receive" button, does he get an Ethereum format address or a Filecoin format address?

The user will see a filecoin address format with `f1` protocol. Converting to `f1` -> 0x involves a lookup from node through network call which we think might cause trouble hence the translation isn't supported. More info [here](https://docs.filecoin.io/smart-contracts/filecoin-evm-runtime/address-types#converting-to-a-0x-style-address)

> Do you think it is better to convert back to the Filecoin format in the UI?

It's definitely better user experience IMO. It will definitely cost more network calls because going from 0x to f1 might be tricky, the other way is much easier. We already do convert f1 -> 0x to check if the account is sending or receiving, I think it would be okay to override the sender/reciever address at least when creating operations. What do you think?

lawRathod:
> Please have a look at my comments. Thanks a lot

Thanks for the review 🙌🏻, I tried to work on the comments. Please have another look when possible.

lawRathod:
Seems like there's been some refac related to the files, I can resolve conflicts once we resolve the comments @hzheng-ledger

libs/ledgerjs/packages/cryptoassets/src/data/filecoin-erc20.json
```diff
@@ -0,0 +1,98 @@
+[
```
hzheng-ledger, line 1: We are now using "importers" now to dynamically fetch token info from our CDN from backend, instead of hard code from in the codebase.

Please refer to:
/ledger-live/libs/ledgerjs/packages/cryptoassets/src/crypto-assets-importer/importers

We are currently working with our backend team to host filecoin erc20 token. We will let you know when the CDN is ready, so that you can do that in dynamic/ importer way.

hzheng-ledger (changes requested):
Sorry for the delay.
Please have a look at my comments related to token "importer".

hzheng-ledger:
@lawRathod our CDN is available now: https://cdn.live.ledger.com/cryptoassets/evm/314/erc20.json
you can use it to retrieve filecoin evm tokens from there instead of hard code tokens in cryptoassets folder.

libs/ledgerjs/packages/cryptoassets/src/data/filecoin-erc20.json
lawRathod: Great thanks @hzheng-ledger! Now using importer for fetching tokens.

PR #6735, opened 2024-04-24
Fix: LIVE-12068 [cosmos]Adapt LL code to avoid using deprecated endpoints

hzheng-ledger:
### ✅ Checklist

- [X] **Impact of the changes: no impact
  - ...

### 📝 Description

1. Since sdk 0.50, the GetTxsEvent will use "query" param instead of "event", see https://docs.cosmos.network/api#tag/Service/operation/GetTxsEvent for more details.
So, we should firstly check the cosmos sdk version and use query/event according to different cosmos sdk version.

2. Since sdk 0.50, the GetTxsEvent endpoint returns empty "logs"(a field in the GetTxsEvent api response). so we can't use logs to retrieve transactions any more. We should use "tx.body.messages" from now on.

Fixed regression:
For example, dydx and dydx testnet upgrade the cosmos SDK to 0.50 recently. So, the current codebase doesn't work with them.
Here is an example: 

https://testnet-dydx-api.lavenderfive.com/cosmos/tx/v1beta1/txs?query=message.sender=%27dydx1pax8j79vm78qqvsnd5maklujqxahqm6j6r2mkl%27&pagination.limit=100&pagination.offset=0&pagination.reverse=true

as we can see that the "logs" array is always empty in the response.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-12068

---


libs/ledger-live-common/src/families/cosmos/api/Cosmos.ts
```diff
@@ -85,6 +93,12 @@ export class CosmosAPI {
     }
   };
 
+  private getCosmosSDKVersion = async (): Promise<string> => {
+    const { application_version } = await this.getNodeInfo();
```
hedi-edelbloute, line 97: We do twice this call, right ? :thinking: maybe there is a way to store it so we can do only one call to nodeInfo, it is constant

hzheng-ledger: why do you think we do this call twice? 

I've save the "_cosmosSDKVersion" in a variable here
```
  private _cosmosSDKVersion: Promise<string> | null = null;
  private get cosmosSDKVersion(): Promise<string> {
    if (!this._cosmosSDKVersion) {
      this._cosmosSDKVersion = this.getCosmosSDKVersion();
    }
    return this._cosmosSDKVersion;
  }
```
so we call only once the getNodeInfo each time we sync an account.

it is true that we can save the node_info somewhere so we don't need to do that for each account that has the same currency.
but as the node_info api is very light and we may switch to the mintscan API in the near future to fetch transactions. I think it is a minor issue at this moment

hedi-edelbloute: Yes I was talking about the node_info part :smile: , ok understood

libs/ledger-live-common/src/families/cosmos/datasets/__snapshots__/cosmos.integration.test.ts.snap
hedi-edelbloute, line 1: Changes in this file mean history prior to this update looks different than current

Some OUT transactions become UNDELEGATE
Some OUT transactions become REWARD
Some OUT transactions become REDELEGATE

Did we have an issue before ?

hzheng-ledger: yes, I did it on purpose.
The current "transaction - > operation" interpretation is wrong because the message logs are incomplete for failed transactions. (There are some missing logs for failed transactions) So I've fixed such issue in the current PR by not using the message logs.
Here are some examples in the snapshot that I've fixed

04CA428FE20A51E9C0F0D18818AC24DF5BD2ED5056B7485071620E6CC1D6C13B
is a "undelegate" transaction. (not a OUT transaction)

3A2EFD293BADC0E880F2C6466C4762F89D370919123C3560F51F49C94877B4FC is a get rewards transaction. (not a OUT transaction)

4309BEE6A8F72579012AA2800696078B3C3F488F2346E4738D57C7E1932D653C is a redelegate transaction. (not a OUT transaction)

hedi-edelbloute: Understood :)

hedi-edelbloute (approved):

libs/ledger-live-common/src/families/cosmos/helpers.ts
```diff
@@ -1,12 +1,17 @@
...
-    "delegate",
-    "withdraw_rewards",
-    "transfer",
+    "MsgUndelegate",
+    "MsgBeginRedelegate",
+    "MsgDelegate",
+    "MsgWithdrawDelegatorReward",
+    "MsgTransfer",
+    "MsgRecvPacket",
+    "MsgSend",
```
lambertkevin, line 14: 👏 👍

libs/ledger-live-common/src/families/cosmos/api/Cosmos.unit.test.ts
lambertkevin, line 1: Maybe I missed it, but I don't think we have a test checking that with a node on v0.50 we use a different URI ?

lambertkevin (approved):

PR #6771, opened 2024-04-29
CVS live-countervalues performance evolutions

gre:
### ✅ Checklist

- [ ] **Covered by automatic tests.** for most of them 
  - countervalues

### 📝 Description

Implements https://ledgerhq.atlassian.net/browse/LIVE-11861 tasks
notably:

- [LIVE-11865](https://ledgerhq.atlassian.net/browse/LIVE-11865): Add live-config fields 
  - they are provided as new fields in `CountervaluesSettings` so the live-countervalues lib don't directly depend on live-config but on LLD/LLM we will inject them from `countervalues_*`
- switch refreshRate to 1min default (remote config)
  - that way, the live price will be way more often up to date
- [LIVE-11864](https://ledgerhq.atlassian.net/browse/LIVE-11864) introduce hybrid batch fetching mechanism 
  - that will improve the performance of frontend<>backend by reducing the amount of Cloudflare cache efficiency
- switch from `/v3/currencies/supported` to `/v3/supported/crypto`
- closes https://github.com/LedgerHQ/ledger-live/pull/6768

### ❓ Context

- **JIRA or GitHub link**: LIVE-11865 LIVE-11864 LIVE-11862

---

[LIVE-11865]: https://ledgerhq.atlassian.net/browse/LIVE-11865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-11864]: https://ledgerhq.atlassian.net/browse/LIVE-11864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-framework/src/currencies/support.ts
```diff
@@ -72,7 +71,7 @@ async function initializeUserSupportedFiats() {
 
 export async function fetchSupportedFiatsTokens(): Promise<string[]> {
   try {
-    const response = await fetch(`${getEnv("LEDGER_COUNTERVALUES_API")}/v2/supported-to`, {
+    const response = await fetch(`${getEnv("LEDGER_COUNTERVALUES_API")}/v3/supported/fiat`, {
```
jnicoulaud-ledger, line 74: I thought we can't do this move yet because the market page does not support all the extra fiats until moved to /v3/markets API.

But when I tried on a build of this PR I still see the the "short" list in the currency selection menu, so maybe I am missing something ?

gre: @jnicoulaud-ledger 
- AFAIK market don't uses this function
- the data of this API is currently crossed with hardcoded data: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/cryptoassets/src/fiats.ts#L23 so it's possible that not all fiat appears yet. I can probably complete the list with the missing fiat data (but also indeed in future we would need a way to have this as an API if we don't want to maintain it statically)

jnicoulaud-ledger: Ha OK, that explains. But if my understanding is correct, market inherits the globally set currency so we are limited by what market supports.

For the harcoded data, I guess we could expose it in an API, but it's ISO-4217 standard, there are libraries that provide this metadata: https://www.npmjs.com/search?q=keywords:iso4217

gre: @jnicoulaud-ledger https://github.com/LedgerHQ/ledger-live/pull/6798

gre: i've merged the additions of fiats so we can test more. i seems to have a subset of fiat that all works on market. do you have a specific fiat that works on the new api but not on the old?

jnicoulaud-ledger: Yes, it's the diff between https://countervalues.live.ledger.com/v2/supported-to and https://countervalues.live.ledger.com/v3/supported/fiat, so for instance: `AFN`, `ALL`, `AMD`...

gre: ok, after testing, i can confirm that market uses a totally different list , so we are fine with these changes. however, I had to do extra fix to actually make Live instance support the new list.

note for ourself: coin-framework also hardcode a `locallySupportedFiats` list, which i suspect market uses, so i'm not touching this, but in future we will need to move this out of coin-framework. ( cc @sprohaszka-ledger fyi )

jnicoulaud-ledger: I just tested again:
 - On prod build global currency and market are always synced
 - On this PR, market currency is always USD even if I set a global currency that is supported (EUR)

I would at least expect market currency to follow the global one when possible.

But also I remember last time this was discussed, it was considered not acceptable to have a different set of supported currencies globally and for market, so we might want to double check this ?

gre: > I would at least expect market currency to follow the global one when possible.

it doesn't seem it's the case today, and this PR didn't have the scope to address this. therefore if things are not connected i'm not sure if it's going to be a problem to have different list on both sides. indeed it's not ideal but i'm not seeing it a blocker to do this first step.
I'll ask product on monday.

cgrellard-ledger (approved):

libs/ledger-live-common/src/config/sharedConfig.ts
```diff
@@ -22,11 +22,23 @@ import { tronConfig } from "../families/tron/config";
...
+    default: 20,
+  },
+};
+
 const liveCommonConfig: ConfigSchema = {
   ...appConfig,
 };
 
 export const liveConfig: ConfigSchema = {
+  ...countervaluesConfig,
```
hedi-edelbloute, line 41: I think this can be a part of liveCommonConfig above, minor

hedi-edelbloute (approved):

libs/ledger-live-common/src/config/sharedConfig.ts
gre: the reasoning behind this is that "countervalues" is a dedicated library and not part of "live-common"

(ultimately the existence of this file that glue things together manually is problematic imo but it's another topic)

libs/coin-framework/src/currencies/support.ts
gre: seen with Anthony, we're good to have this evolution even if market is disconnected for the time being

PR #6793, opened 2024-05-01
[Fix/stacks] Optimistic operation disappears for long running transactions

lawRathod:
### ✅ Checklist

  - ...

### 📝 Description

The optimistics transaction disappears when transactions on stacks network takes long time. The pending operations are now added to accounts object during sync to fix the problem. 

### ❓ Context

- **JIRA or GitHub link**: [https://ledgerhq.atlassian.net/browse/LIVE-10510](https://ledgerhq.atlassian.net/browse/LIVE-10510)

---

### DIFF
--- a/libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
+++ b/libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
@@ -10,7 +10,12 @@ import {
 
 import { decodeAccountId } from "../../../../account";
 import { fetchFullMempoolTxs, fetchNonce } from "../../bridge/utils/api";
-import { DecodedSendManyFunctionArgsCV, StacksNetwork, TransactionResponse } from "./api.types";
+import {
+  DecodedSendManyFunctionArgsCV,
+  MempoolTransaction,
+  StacksNetwork,
+  TransactionResponse,
+} from "./api.types";
 import { getCryptoCurrencyById } from "../../../../currencies";
 import { encodeOperationId, encodeSubOperationId } from "../../../../operation";
 import { StacksOperation } from "../../types";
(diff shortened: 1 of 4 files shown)


hzheng-ledger (changes requested):
it looks good to me.
I've some questions, 
I don't know whether the balance return from /extended/v1/address/${addr}/stx includes the pending transactions(mempool transactions)

I see this piece of code
```
  for (const tx of mempoolTxs) {
    spendableBalance = spendableBalance
      .minus(new BigNumber(tx.fee_rate))
      .minus(new BigNumber(tx.token_transfer.amount));
  }
```

Does the mempoolTxs includes both incoming and outgoing (send and receive) transactions?
because the "fee_rate" is not paid by the current account for receive transactions.
should we give the "tx.token_transfer.amount" a positive or negative value according to send/receive transaction?

lawRathod:
@hzheng-ledger Thanks for the review. 

> whether the balance return from /extended/v1/address/${addr}/stx includes the pending transactions(mempool transactions)

Only sent transactions are part of the mempool txns object https://github.com/Zondax/ledger-live/blob/develop/libs/ledger-live-common/src/families/stacks/bridge/utils/api.ts#L142-L147 so we substract the fee and amount from the balance.

hzheng-ledger (approved):
it looks good to me. thanks

libs/ledger-live-common/src/families/stacks/bridge/utils/misc.ts
```diff
@@ -206,6 +263,7 @@ export const getAccountShape: GetAccountShape = async info => {
     balance,
     spendableBalance,
     operations: flatMap(rawTxs, mapTxToOps(accountId)),
+    pendingOperations: flatMap(mempoolTxs, mapPendingTxToOps(accountId, address)),
```
hzheng-ledger, line 266: we shouldn't add pendingOperations here.
This pendingOperations is used for new broadcasted transaction from our UI. (not even necessarily in mempool, but we want to display them to users in our Ledger live)
what we should do is to merge  ```flatMap(rawTxs, mapTxToOps(accountId))``` and ```flatMap(mempoolTxs, mapPendingTxToOps(accountId, address))```
in the above "operations" field. for confirmed transaction, we set "blockHeight" and "blockHash" field correctly and for pending transaction in mempool, we set "blockHeight" and "blockHash" field as null. So that we can see these pending transactions in our Ledger live
So that

hzheng-ledger (approved):
looks good to me

hzheng-ledger:
Thanks a lot. I've approved the PR

hedi-edelbloute (approved):

PR #6799, opened 2024-05-02
Feat/LIVE-12193 track swap cancel accept

andreicovaciu:
### ✅ Checklist

  - ...

### 📝 Description

Updates `custom.exchange.start` custom handler to return the device too.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-12193

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
+++ b/apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
@@ -52,10 +52,6 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"]) {
         tracking,
         manifest,
         uiHooks: {
-          "custom.device.get": () => ({
-            deviceId: device?.deviceId,
-            modelId: device?.modelId,
-          }),
           "custom.exchange.start": ({ exchangeParams, onSuccess, onCancel }) => {
             dispatch(
               openExchangeDrawer({
@@ -63,10 +59,10 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"]) {
                 ...exchangeParams,
                 exchangeType: ExchangeType[exchangeParams.exchangeType],
                 onResult: (nonce: string) => {
-                  onSuccess(nonce);
+                  onSuccess(nonce, device);
                 },
                 onCancel: (error: Error) => {
-                  onCancel(error);
+                  onCancel(error, device);
                 },
               }),
             );
--- a/apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
+++ b/apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
@@ -14,11 +14,13 @@ import { closePlatformAppDrawer, openExchangeDrawer } from "~/renderer/actions/U
 import { WebviewProps } from "../Web3AppWebview/types";
 import { context } from "~/renderer/drawers/Provider";
 import WebviewErrorDrawer from "~/renderer/screens/exchange/Swap2/Form/WebviewErrorDrawer";
+import { getCurrentDevice } from "~/renderer/reducers/devices";
 
 export function usePTXCustomHandlers(manifest: WebviewProps["manifest"]) {
   const dispatch = useDispatch();
   const accounts = useSelector(flattenAccountsSelector);
   const { setDrawer } = React.useContext(context);
+  const device = useSelector(getCurrentDevice);
 
   const tracking = useMemo(
     () =>
@@ -50,6 +52,10 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"]) {
         tracking,
         manifest,
         uiHooks: {
+          "custom.device.get": () => ({
+            deviceId: device?.deviceId,
+            modelId: device?.modelId,
+          }),
           "custom.exchange.start": ({ exchangeParams, onSuccess, onCancel }) => {
             dispatch(
               openExchangeDrawer({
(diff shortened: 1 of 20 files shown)


kallen-ledger (dismissed):
Nice, looks good to me. Let's wait for the wallet api team to take a look before merging 💪

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -50,6 +52,10 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"]) {
         tracking,
         manifest,
         uiHooks: {
+          "custom.device.get": () => ({
+            deviceId: device?.deviceId,
+            modelId: device?.modelId,
+          }),
```
Justkant, line 58: If there is now ui you don't really need a uiHook for `custom.device.get`

libs/ledger-live-common/src/wallet-api/Exchange/server.ts
```diff
@@ -309,6 +317,7 @@ export const handlers = ({
         }),
       );
     }),
+    "custom.device.get": customWrapper<void, DeviceGetRequest>(() => deviceGet()),
```
Justkant, line 320: You could implement `deviceGet` in here directly IMO, if you still want to get the device from LLD and LLM you can receive it as a param like `accounts` or `manifest` instead of using a uiHook

Justkant (commented):
I think we can move the logic from LLD and LLM to LLC and remove the uiHook for device as it doesn't really need ui in your context

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -57,10 +59,10 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"]) {
                 ...exchangeParams,
                 exchangeType: ExchangeType[exchangeParams.exchangeType],
                 onResult: (nonce: string) => {
-                  onSuccess(nonce);
+                  onSuccess(nonce, device);
```
Justkant, line 62: Not sure how much it keeps the device in sync here because it will be hoisted when the `custom.exchange.start` is called it could be the old value

It would probably be better if `onResult` and `onCancel` could return the device as it's done for LLM
I checked the code and it seems pretty straightforward to add some data:
- Inputs: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/exchange/platform/startExchange.ts#L16
- Success: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/exchange/platform/startExchange.ts#L49-L52
- Error: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/exchange/platform/startExchange.ts#L62-L65
- StartExchange inputs: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/hw/actions/startExchange.ts#L150-L155
- startExchange mapResult: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/hw/actions/startExchange.ts#L37-L49

Another solution could be to simply use `const device = useSelector(getCurrentDevice);` in `apps/ledger-live-desktop/src/renderer/components/LiveAppDrawer.tsx` because if we do the changes above we might want to update LLM so result.device comes from the action and not a state.

You can also cleanup [this types](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/components/LiveAppDrawer.tsx#L153-L161) that are not needed anymore
Screenshot 2024-05-06 at 16 08 00

Justkant (commented):
I think we have the main logic good here just on desktop I'd like to modify it a bit to make sure we have a device in the end

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
Justkant: Hey not sure we have a lot of documentation on the device actions and exchange part to be honest
I'm pretty familiar with this because I worked on some issues around it

My main concern here is to avoid having to much differences between LLD and LLM and sharing as much as possible, especially because you have a lot of ways to pass the data in the end correctly

kallen-ledger (approved):

gre (approved):

ofreyssinet-ledger (approved):
LGTM on the device experience team's scope 👍

apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
```diff
@@ -336,19 +336,19 @@ export const PlatformAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
                 provider,
               },
               onResult: (result: {
-                startExchangeResult?: string;
-                startExchangeError?: Error;
+                startExchangeResult?: { nonce: string; device: Device };
+                startExchangeError?: { error: Error; device?: Device };
                 device?: Device;
               }) => {
```
Justkant, line 342: Here we should use the types already present instead of redeclaring them, I tested that locally too
```ts
              onResult: result => {
```

apps/ledger-live-mobile/src/components/Web3AppWebview/helpers.ts
```diff
@@ -434,17 +434,17 @@ function useUiHook(): UiHook {
               exchangeType: ExchangeType[exchangeType],
             },
             onResult: (result: {
-              startExchangeResult?: string;
-              startExchangeError?: Error;
+              startExchangeResult?: { nonce: string; device: Device };
+              startExchangeError?: { error: Error; device?: Device };
               device?: Device;
             }) => {
```
Justkant, line 440: Same as above we can avoid typing result here as it's already done
```ts
            onResult: result => {
```

Justkant (approved):

PR #6819, opened 2024-05-07
[Coin Fmk] Batcher moving to Network

sprohaszka-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Move Batcher from `coin-framework` to `live-network`.
Might need to consider to expands this concept.

Also propose a new `network` function to get rid of *Axios type* (this is a good practice to partition types of external libraries). This method is compatible with the old one.

**OLD implementation**
```ts
import network from `@ledgerhq/live-network/network`
```

**NEW implementation**
```ts
import network from `@ledgerhq/live-network`
```

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/live-network/src/network.ts
```diff
@@ -158,8 +162,52 @@ const extractErrorMessage = (raw: string): string | undefined => {
...
+    request.method = "GET";
+  }
+
+  if (request.method === "GET") {
+    if (!("timeout" in request)) {
+      request.timeout = getEnv("GET_CALLS_TIMEOUT");
+    }
+
+    response = await retry(() => axios(request), {
+      maxRetry: getEnv("GET_CALLS_RETRY"),
```
lvndry, line 192: Do you think it'd make sense to also have `maxRetry` this in `LiveNetworkRequest` ?

lvndry (dismissed):
Nice work 👏🏿

libs/live-network/src/network.ts
sprohaszka-ledger: What bothering me, is that this is not part of the `live-network` package.
I think we should do, but maybe with a small refacto.

lambertkevin (approved):

lvndry (approved):

PR #6822, opened 2024-05-07
[LIVE-8174] Feature - coin-xrp module

lambertkevin:
### ✅ Checklist

  - should have no impact on features
  - XRP send flow can be impacted in case of bugs

### 📝 Description

Creation of the `coin-xrp` module, with support for better spendable balance (monitoring base reserve & trustlines), better typing, signer injection & removal of unmaintained libs like `ripple-bs58check`

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-8174

---


libs/coin-modules/coin-xrp/src/prepareTransaction.ts
```diff
@@ -0,0 +1,45 @@
...
+  let networkInfo: NetworkInfo | null | undefined = t.networkInfo;
+
+  if (!networkInfo) {
+    try {
+      const info = await getServerInfos();
+      const serverFee = parseAPIValue(info.info.validated_ledger.base_fee_xrp.toString());
+      networkInfo = {
+        family: "xrp",
+        serverFee,
+        baseReserve: new BigNumber(0), // NOT USED. will refactor later.
```
lvndry, line 28: should we delete this property ?

libs/ledger-live-common/src/families/xrp/walletApiAdapter.test.ts
```diff
@@ -16,6 +16,7 @@ describe("getWalletAPITransactionSignFlowInfos", () => {
 
       const expectedLiveTx: Partial<Transaction> = {
         ...xrpPlatformTx,
+        family: "xrp",
```
lvndry, line 19: Maybe we should have somewhere in Live common an object

```ts
export const defaultXRPTransaction = {
 family: "xrp"
}
```

And then do 

```ts
const expectedLiveTx: Partial<Transaction> = {
        ...xrpPlatformTx,
        ...defaultXRPTransaction,
}
```

That way we can easily extends the default values everywhere its needed

libs/ledger-live-common/src/hw/signTransaction/index.ts
```diff
@@ -1,6 +1,5 @@
 import type { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
 import Transport from "@ledgerhq/hw-transport";
-import ripple from "./ripple";
 import tron from "./tron";
 // TODO deprecate this approach
```
lvndry, line 4: Just out of curiosity do we have a better approach in mind already ?

libs/ledger-live-common/src/families/xrp/bridge.integration.test.ts
```diff
@@ -9,8 +9,8 @@ import {
   // NotEnoughBalanceBecauseDestinationNotCreated,
   InvalidAddressBecauseDestinationIsAlsoSource,
 } from "@ledgerhq/errors";
-import { fromTransactionRaw } from "./transaction";
-import type { Transaction } from "./types";
+import { fromTransactionRaw } from "@ledgerhq/coin-xrp/lib/transaction";
+import type { Transaction } from "@ledgerhq/coin-xrp/lib/types";
 import { addNotCreatedRippleMockAddress } from "./bridge/mock";
 // import { formatCurrencyUnit } from "../../currencies";
 const newAddress1 = "rZvBc5e2YR1A9otS3r9DyGh3NDP8XLLp4";
```
sprohaszka-ledger, line 16: The `dataset` object should be define inside `coin-xrp` package.

libs/coin-modules/coin-xrp/src/cli-transaction.ts
```diff
@@ -0,0 +1,11 @@
+/* istanbul ignore file: untested because empty */
```
sprohaszka-ledger, line 1: Why didn't you keep the content of the old `cli-transaction` file?

libs/ledger-live-common/src/families/xrp/setup.ts
```diff
@@ -0,0 +1,31 @@
...
+import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets/currencies";
+import { CreateSigner, createResolver, executeWithSigner } from "../../bridge/setup";
+import { getCurrencyConfiguration } from "../../config";
+import { Resolver } from "../../hw/getAddress/types";
+
+const createSigner: CreateSigner<Xrp> = (transport: Transport) => {
+  return new Xrp(transport);
+};
+
+const getCoinConfig = () => getCurrencyConfiguration<XrpConfig>(getCryptoCurrencyById("ripple"));
```
sprohaszka-ledger, line 20: Shouldn't be "xrp" instead?

sprohaszka-ledger (commented):
Thanks for splitting the code into different files :)

Some questions, but no hard blocker.

libs/coin-modules/coin-xrp/src/signer.ts
```diff
@@ -0,0 +1,17 @@
+export type XrpAddress = {
+  publicKey: string;
+  address: string;
+  chainCode?: string;
+};
+
+export type XrpSignature = string; // `0x${string}`
```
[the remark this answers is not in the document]
lambertkevin: What do you mean by enforce the type ? Using `0x${string}` instead of `string` ?
If so, unfortunately, I really like this templated type, but it requires to do a lot of assertion everywhere if your base layer is not implemented straight away with this in mind. Doing an ``` "0xabc" as `0x${string}` ``` makes it kinda useless for now. 
But yeah, if possible it would be nice to use it in the whole code base and implement it at every level of LL

libs/ledger-live-common/src/families/xrp/setup.ts
lambertkevin: Unfortunately I can't change the currency name without removing every XRP account from all Ledger Live instances... That wouldn't problematic to add them again, but I guess that's a big impact on CS for a small change like this 🤷‍♂️

libs/ledger-live-common/src/families/xrp/bridge.integration.test.ts
lambertkevin: Done 👍

libs/coin-modules/coin-xrp/src/types.ts
```diff
@@ -8,28 +9,47 @@ import type {
...
   networkInfo: NetworkInfoRaw | null | undefined;
   tag: number | null | undefined;
   feeCustomUnit: Unit | null | undefined;
 };
+
 export type TransactionStatus = TransactionStatusCommon;
 export type TransactionStatusRaw = TransactionStatusCommonRaw;
+
+const { TRANSACTION_TYPES } = XrplDefinitions;
+export type XrplTransaction = {
```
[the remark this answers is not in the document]
lambertkevin: Sure ! No strong opinion about that 🤷‍♂️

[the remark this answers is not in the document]
lambertkevin: Done 👍

libs/coin-modules/coin-xrp/src/cli-transaction.ts
lambertkevin: It's just me doing a bad copy/paste of the EVM coin-module... (And I'm wondering now why this one is empty...) 🤦‍♂️

lambertkevin: Added the file back with some modifications for typing and removal of the lodash dep 👍

libs/coin-modules/coin-xrp/src/signer.ts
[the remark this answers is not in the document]
sprohaszka-ledger: Yeap, we could later use it, but from the bottom up, to be sure that there is no mistake.

sprohaszka-ledger (dismissed):
👌

gre (approved):
ok for live-hub impact (which is just the env lib)

ofreyssinet-ledger (approved):

libs/ledger-live-common/src/families/xrp/platformAdapter.ts
```diff
@@ -1,14 +1,14 @@
...
-import { Transaction } from "./types";
+import { Transaction } from "@ledgerhq/coin-xrp/types";
 
 const CAN_EDIT_FEES = true;
 
 const areFeesProvided = (tx: PlatformTransaction): boolean => !!tx.fee;
 
 const convertToLiveTransaction = (tx: PlatformTransaction): Partial<Transaction> => {
-  return tx;
+  return { ...tx, family: "xrp" };
```
Justkant, line 11: Not sure why we need to add `family` here as I would expect it to already be in the `tx`
I suppose it's for making sure you always have the correct one

Justkant (dismissed):
LGTM for my part I just wonder if this change is really necessary

libs/ledger-live-common/src/families/xrp/platformAdapter.ts
Justkant: Ok got it makes sense, thx for the details

Justkant (approved):

PR #6857, opened 2024-05-14
:rotating_light: Release merge conflicts

ofreyssinet-ledger:
# The release had some merge conflicts :rotating_light:

This pull request was created in order to resolve the merge conflicts between the release and develop branches.

```bash
git checkout develop && git pull

git checkout support/release-merge-conflicts

git merge develop
```

## Checklist

To be ready to be merged, we must comply with all theses checks:

- [ ] all CI checks have passed
- [ ] reviews of the code owners impacted by the conflict

## ⚠️ Warning ⚠️

> **Warning**
> Only merge the PR with the option `Create a merge commit`

PR #6871, opened 2024-05-16
fix(lld): fix analytics useEffect in default

LucasWerey:
### ✅ Checklist

  - ...

### 📝 Description

Fix analytics useEffect previously introduced that was overriding app.json 

### ❓ Context

- **JIRA or GitHub link**: https://github.com/LedgerHQ/ledger-live/pull/6857/commits/9ce66ffc48695fd0add96d48091a7d533b2ff091 & [LIVE-12571](https://ledgerhq.atlassian.net/browse/LIVE-12571)

---

[LIVE-12571]: https://ledgerhq.atlassian.net/browse/LIVE-12571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/Default.tsx
+++ b/apps/ledger-live-desktop/src/renderer/Default.tsx
@@ -52,6 +52,7 @@ import {
   hasCompletedOnboardingSelector,
   hasSeenAnalyticsOptInPromptSelector,
 } from "~/renderer/reducers/settings";
+import { isLocked as isLockedSelector } from "~/renderer/reducers/application";
 import { setShareAnalytics, setSharePersonalizedRecommendations } from "./actions/settings";
 
 const PlatformCatalog = lazy(() => import("~/renderer/screens/platform"));
@@ -198,13 +199,16 @@ export default function Default() {
   const listAppsV2 = useFeature("listAppsV2minor1");
   const analyticsFF = useFeature("lldAnalyticsOptInPrompt");
   const hasSeenAnalyticsOptInPrompt = useSelector(hasSeenAnalyticsOptInPromptSelector);
+  const isLocked = useSelector(isLockedSelector);
   const dispatch = useDispatch();
 
   useEffect(() => {
-    if (!analyticsFF?.enabled || hasSeenAnalyticsOptInPrompt) return;
-    dispatch(setShareAnalytics(false));
-    dispatch(setSharePersonalizedRecommendations(false));
-  });
+    if (!isLocked && analyticsFF?.enabled && !hasSeenAnalyticsOptInPrompt) {
+      dispatch(setShareAnalytics(false));
+      dispatch(setSharePersonalizedRecommendations(false));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isLocked]);
 
   useEffect(() => {
     if (!listAppsV2) return;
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/Default.tsx
```diff
@@ -198,13 +199,16 @@ export default function Default() {
   const listAppsV2 = useFeature("listAppsV2minor1");
   const analyticsFF = useFeature("lldAnalyticsOptInPrompt");
   const hasSeenAnalyticsOptInPrompt = useSelector(hasSeenAnalyticsOptInPromptSelector);
+  const isLocked = useSelector(isLockedSelector);
   const dispatch = useDispatch();
 
   useEffect(() => {
+    if (isLocked) return;
```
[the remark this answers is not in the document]
LucasWerey: Done !

apps/ledger-live-desktop/src/renderer/Default.tsx
```diff
@@ -198,13 +199,17 @@ export default function Default() {
...
+  const isLocked = useSelector(isLockedSelector);
   const dispatch = useDispatch();
 
   useEffect(() => {
-    if (!analyticsFF?.enabled || hasSeenAnalyticsOptInPrompt) return;
-    dispatch(setShareAnalytics(false));
-    dispatch(setSharePersonalizedRecommendations(false));
-  });
+    if (!isLocked) {
+      if (!analyticsFF?.enabled || hasSeenAnalyticsOptInPrompt) return;
```
cksanders, line 207:
```suggestion
    if (!isLocked && (analyticsFF?.enabled || !hasSeenAnalyticsOptInPrompt)) {
```

cgrellard-ledger (approved):

apps/ledger-live-desktop/src/renderer/Default.tsx
```diff
@@ -198,13 +199,16 @@ export default function Default() {
...
   const hasSeenAnalyticsOptInPrompt = useSelector(hasSeenAnalyticsOptInPromptSelector);
+  const isLocked = useSelector(isLockedSelector);
   const dispatch = useDispatch();
 
   useEffect(() => {
-    if (!analyticsFF?.enabled || hasSeenAnalyticsOptInPrompt) return;
-    dispatch(setShareAnalytics(false));
-    dispatch(setSharePersonalizedRecommendations(false));
-  });
+    if (!isLocked && analyticsFF?.enabled && !hasSeenAnalyticsOptInPrompt) {
```
cksanders, line 206: why are we changing the logic? previously it was an or case?

cgrellard-ledger: It does the same as before because now we execute the logic instead of returning when the if is true

LucasWerey: Yes it was an or case but we want to trigger the theses dispatches only if those 3 conditions are met: 
- App unlocked
- feature flag is enabled
- the prompt has not been seen by the user

The OR case was a return if the user didn't see the prompt or if the FF is disabled 
So we have the same things here as before

cksanders: yikes

cksanders (approved):

PR #6874, opened 2024-05-16
[Coin-Polkadot] Use node calls instead of sidecar forks

sprohaszka-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Replace call to Ledger's Sidecar fork by direct call to Ledger's Polkadot node.
The code is mostly a copy paste from [Sidecar fork](https://github.com/LedgerHQ/substrate-api-sidecar/).

To fix issue with validators list, the latest version of Polkadot SDK is used.

[Integration tests](libs/coin-modules/coin-polkadot/src/network/sidecar.integ.test.ts) have been added in order to check the consistency between sidecar calls and node calls.
Those tests **must not** been integrated into the CI, as they have different output with time (at least with this first implementation)

Add CoinConfig setup for Polkadot too.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-12399], [LIVE-12400], [LIVE-12401], [LIVE-12402]

---

[LIVE-12399]: https://ledgerhq.atlassian.net/browse/LIVE-12399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-framework/src/config.ts
```diff
@@ -18,3 +18,5 @@ export type CurrencyConfig = {
   status: ConfigStatus;
   [key: string]: unknown;
 };
+
+export type CoinConfig<T> = () => T;
```
lvndry, line 22:
```suggestion
export type CoinConfig<T extends CurrencyConfig> = () => T;
```

libs/coin-modules/coin-polkadot/src/network/sidecar.types.ts
```diff
@@ -13,23 +13,19 @@ interface IPallet {
   pallet: string;
   palletIndex: number;
 }
-interface IUnlocking {
+export interface IUnlocking {
```
lvndry, line 16:
```suggestion
export interface Unlocking {
```
I think we moved away from this convention in the code base

sprohaszka-ledger: Yeah, it's a little bit too much "Java old style". But I prefer to not touch everything in this PR and stay focus on the main purpose (use node calls)

lvndry (approved):

libs/ledger-live-common/src/families/polkadot/config.ts
```diff
@@ -7,6 +8,16 @@ export const polkadotConfig: Record<string, ConfigInfo> = {
       status: {
         type: "active",
       },
+      sidecar: {
+        url: getEnv("API_POLKADOT_SIDECAR"),
+        credentials: getEnv("API_POLKADOT_SIDECAR_CREDENTIALS"),
```
hzheng-ledger, line 13: I've a just question about "API_POLKADOT_SIDECAR_CREDENTIALS". 
It seems that it is an empty string by default in env.ts.
I am wondering whether it is useful ? Where is this "credentials" used?

sprohaszka-ledger: It is used in staging. The staging information requires a token / basic auth header. In production, it is not used (yet).

libs/ledger-live-common/src/families/polkadot/config.ts
```diff
@@ -7,6 +8,16 @@ export const polkadotConfig: Record<string, ConfigInfo> = {
...
+      },
+      node: {
+        url: "https://polkadot-rpc.publicnode.com",
+      },
+      staking: {
+        electionStatusThreshold: getEnv("POLKADOT_ELECTION_STATUS_THRESHOLD"),
```
hzheng-ledger, line 19: it seems that this "electionStatusThreshold" is not used either if I don't make mistake? Should we keep it here?

hzheng-ledger: I suggest that we don't use environmental variables in config file. we can just write the default value here since with the firebase config framework, we can already change the config in real time.

hzheng-ledger (approved):
looks good to me in general

PR #6912, opened 2024-05-23
[LIVE-12648] Crash because of invalid DeviceModelId

ofreyssinet-ledger:
### ✅ Checklist

  - USB connection on Android

### 📝 Description

The store has some bad data in it because we are doing bad type assertions of some objects into the `Device` type.
It's currently causing crashes at different levels of LLM, on Android.

The solution proposed here is:
- Remove the `as Device` type assertions. If an object is incorrect (here, because it does not contain device model id), it is actually unusable so it should be ignored. We could not reproduce the issue but it appears that in some rare cases, the "device discovery" of the Android USB transport sends event with such invalid objects.
- Fix the selectors so that they filter out this invalid data. In cases where it previously would have crashed, it will just behave as if the last connected device was not saved, so the app stays fonctional. The invalid data will be replaced at some point in the store, the next time the user connects a device.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-12648], [Sentry issue #1](https://ledger.sentry.io/issues/5361767582), [Sentry issue #2](https://ledger.sentry.io/issues/5370096252)

---

[LIVE-12648]: https://ledgerhq.atlassian.net/browse/LIVE-12648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/reducers/settings.test.ts
+++ b/apps/ledger-live-desktop/src/renderer/reducers/settings.test.ts
@@ -1,0 +1,51 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { describe, it, expect } from "@jest/globals";
+import { DeviceModelId } from "@ledgerhq/types-devices";
+import { lastSeenDeviceSelector, INITIAL_STATE as SETTINGS_INITIAL_STATE } from "./settings";
+import { State } from ".";
+import { aDeviceInfoBuilder } from "@ledgerhq/live-common/mock/fixtures/aDeviceInfo";
+
+const invalidDeviceModelIds = ["nanoFTS", undefined, "whatever"];
+const validDeviceModelIds: DeviceModelId[] = Object.values(DeviceModelId);
+
+describe("lastSeenDeviceSelector", () => {
+  it("should return the last seen device if the deviceModelId is valid", () => {
+    validDeviceModelIds.forEach(deviceModelId => {
+      const lastSeenDevice: State["settings"]["lastSeenDevice"] = {
+        modelId: deviceModelId,
+        deviceInfo: aDeviceInfoBuilder(),
+        apps: [],
+      };
+      const state = {
+        ...({} as State),
+        settings: {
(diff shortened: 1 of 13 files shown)


apps/ledger-live-desktop/src/renderer/reducers/settings.test.ts
valpinkman, line 1: 😍

apps/ledger-live-mobile/src/components/SelectDevice2/DeviceList.tsx
```diff
@@ -18,7 +18,7 @@ export function DeviceList({ deviceList, handleOnSelect }: Props) {
   return (
     <Flex>
       {deviceList.slice(0, showAll ? deviceList.length : MAX_LEDGERS_DISPLAYED).map(device => (
-        <Item key={device.deviceId} device={device as Device} onPress={handleOnSelect} />
+        <Item key={device.deviceId} device={device} onPress={handleOnSelect} />
```
valpinkman, line 21: 👌

apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
```diff
@@ -173,15 +173,16 @@ export default function SelectDevice({
...
 
+        if (!deviceModel) return;
+
         setDevice((maybeDevice: Device | undefined) => {
           return (
-            maybeDevice ||
-            ({
+            maybeDevice || {
               deviceName: name,
               modelId: deviceModel?.id,
```
valpinkman, line 182: [ASK] can this still be undefined now ?

valpinkman (approved):
Very good. From what we saw together it definitely make things more robust. 
And shipped with tests 👌

valpinkman (approved):

sshmaxime (approved):

PR #6853, opened 2024-05-14
feat(LIVE-9638): change provider info from static object to API call

lpaquet-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [ ] **Covered by automatic tests.** 
- [ ] **Impact of the changes:** 
  - ...

### 📝 Description

Change static provider info to an API call

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-9638

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #6968, opened 2024-05-29
fix(live-common/exchange): fix CAL service integration

jnicoulaud-ledger:
### ✅ Checklist

- [ ] not covered by automatic tests =>  isofonctional change
  - swap pane

### 📝 Description

Followup of #6853:
- fix internal URL used, bypassing Cloudflare cache
- fix unused field requested
- limit request to swap partners
- add `X-Ledger-Client-Version` header for monitoring

image

I did not fix in this PR:
- partners data is requested again at every refresh. It is loaded from disk cache but still seems unnecessary and will cause extra backend load if cache does not work
- I am not sure if using `fetch()` is the preferred way in Ledger Live. Other parts of Ledger Live use `network` shared lib that sets the `X-Ledger-Client-Version` header we need for monitoring: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/live-network/src/network.ts#L251

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/ledger-live-common/src/exchange/providers/swap.ts
+++ b/libs/ledger-live-common/src/exchange/providers/swap.ts
@@ -1,6 +1,6 @@
 import { ExchangeProviderNameAndSignature } from ".";
 import { isIntegrationTestEnv } from "../swap/utils/isIntegrationTestEnv";
-import PACKAGE from "../../../package.json";
+import network from "@ledgerhq/live-network";
 
 export type SwapProviderConfig = {
   needsKYC: boolean;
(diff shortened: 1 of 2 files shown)


lpaquet-ledger (dismissed):

libs/ledger-live-common/src/exchange/providers/swap.ts
```diff
@@ -141,7 +142,12 @@ export const getProvidersData = async () => {
   try {
     const providersData = await (
       await fetch(
-        "https://crypto-assets-service.api.aws.prd.ldg-tech.com/v1/partners?output=name,payload_signature_computed_format,signature,public_key,public_key_curve",
+        "https://crypto-assets-service.api.ledger.com/v1/partners?output=name,signature,public_key,public_key_curve&service_name=swap",
+        {
+          headers: {
+            "X-Ledger-Client-Version": `live-exchange/${PACKAGE.version}`,
```
gre, line 148: using `network` instead of `fetch` where network is from `live-network` will benefit the auto set of the Ledger-Client-Version header.

(moreover but comment that is beyond this PR: we should have defined a domain env in live-env, like all other services we use, maybe someone from live-hub can takes this over)

gre (changes requested):
minimal change suggested: just replace the header by

```ts
"X-Ledger-Client-Version": getEnv("LEDGER_CLIENT_VERSION")
```

libs/ledger-live-common/src/exchange/providers/swap.ts
jnicoulaud-ledger: Thanks, I switched it to network lib. Please double check @lpaquet-ledger, I am not familiar with this codebase

CremaFR (approved):
Thanks I wasn't aware of this too
closing #6979

gre (approved):

jnicoulaud-ledger:
@CremaFR @lpaquet-ledger (I don't have merge rights)

PR #6977, opened 2024-05-30
[LIVE-12806] Feature - Add Scroll & Scroll Sepolia currencies

lambertkevin:
### ✅ Checklist

  - no impact

### 📝 Description

Adding Scroll & Scroll Sepolia in the Live.
~~Waiting for tokens in CAL.~~

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-12806

---


libs/coin-modules/coin-evm/src/__tests__/unit/logic.unit.test.ts
```diff
@@ -1,5 +1,6 @@
 import BigNumber from "bignumber.js";
 import { getEnv, setEnv } from "@ledgerhq/live-env";
+import * as EVM_TOOLS from "@ledgerhq/evm-tools/lib/message/EIP712/index";
```
lvndry, line 3:
```suggestion
import * as EVM_TOOLS from "@ledgerhq/evm-tools/message/EIP712/index";
```

libs/coin-modules/coin-evm/src/api/node/ledger.ts
```diff
@@ -395,6 +394,72 @@ export const getOptimismAdditionalFees: NodeApi["getOptimismAdditionalFees"] = a
...
+  transaction,
+) => {
+  const config = getCoinConfig(currency).info;
+  const { node } = config || /* istanbul ignore next */ {};
+  if (!isLedgerNodeConfig(node)) {
+    throw new LedgerNodeUsedIncorrectly();
+  }
+
+  if (!["scroll", "scroll_sepolia"].includes(currency.id)) {
+    return new BigNumber(0);
```
lvndry, line 417: Shouldn't it throw an error if we can call this function from an unsupported currency ?

libs/coin-modules/coin-evm/src/synchronization.ts
```diff
@@ -67,7 +67,7 @@ export const getAccountShape: GetAccountShape<Account> = async (infos, { blackli
           Math.max(latestSyncedHeight - SAFE_REORG_THRESHOLD, 0),
           blockHeight,
         );
-      } catch (e) {
+      } catch (e) /* istanbul ignore next: just logs */ {
```
lvndry, line 70: why istanbul comments everywhere ?

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
```diff
@@ -288,7 +289,7 @@ const Body = ({
     editType,
     transaction,
     transactionToUpdate,
-    status,
+    status: status as TransactionStatus,
```
lvndry, line 292: `as` don't seem necessary ?

libs/coin-modules/coin-evm/src/synchronization.ts
lambertkevin: Because that's parts of the code that we don't want/need to test, so it's unnecessary in the code coverage to have them as uncovered and potentially showing files as yellow or red as is if it was problematic.
TLDR: Nicer statistics

apps/ledger-live-desktop/src/renderer/families/evm/EditTransaction/Body.tsx
lambertkevin: It is because the `TransactionStatus` from the props is generic while this is the `TransactionStatus` of coin-evm only. 
I tried to change it at the bridge level to have better typing with generics & without casting, but it's, again, a PR of hundreds of files... So it's for another time :p

libs/coin-modules/coin-evm/src/api/node/ledger.ts
lambertkevin: We could, I just prefer a graceful fallback when possible to avoid having huge errors during development for a function called during the `prepareTransaction` phase.

PR #7023, opened 2024-06-06
⚡️ [FEAT] : WalletSync - Synchronize with LLD 🔄

mcayuelas-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Synchronize Flow 
- QRCode Screen ✅
- PinCode Screen ✅

https://github.com/LedgerHQ/ledger-live/assets/112866305/150ed5cd-f90c-4d83-99d1-0c95ec9f337a

### ❓ Context

- **JIRA or GitHub link**: [LIVE-12299] [LIVE-12302] 

---

[LIVE-12299]: https://ledgerhq.atlassian.net/browse/LIVE-12299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-12302]: https://ledgerhq.atlassian.net/browse/LIVE-12302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/newArch/WalletSync/Flows/Synchronize/01-SynchModeStep.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useTranslation } from "react-i18next";
+import { Card } from "../../components/Card";
+import styled, { useTheme } from "styled-components";
+
+type Props = {
+  goToQRCode: () => void;
+  goToSynchWithDevice: () => void;
+};
+
+export default function SynchronizeModeStep({ goToQRCode, goToSynchWithDevice }: Props) {
```
KVNLS, line 12: We usually shortcut Synchronize with `Sync` instead of `Synch`, could you update accordingly, please?

apps/ledger-live-desktop/src/newArch/WalletSync/Flows/Synchronize/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const SynchronizeWallet = () => {
+  const dispatch = useDispatch();
+
+  const { currentStep, goToNextScene } = useFlows({
+    flow: Flow.Synchronize,
+  });
+
+  const goToActivation = () => {
+    dispatch(setFlow(Flow.Activation));
```
KVNLS, line 19: I think that when we change the flow we should also be able to change the step too. Because a step belongs to a flow.
``` 
dispatch(setFlow(Flow.ManageBackups));
dispatch(setStep(Step.ManageBackupStep));
```
Maybe we could have something like this instead:
`dispatch(setFlow({flow: Flow.ManageBackups, step:Step.ManageBackupStep} ));`

KVNLS (changes requested):
Could we add a debug cta to start a specific flow in developer settings please? Because here we can't test the activation flow without changing the code.

LucasWerey (approved):
LGTM !

apps/ledger-live-desktop/src/newArch/WalletSync/Flows/Synchronize/02-QRCodeStep.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const { colors } = useTheme();
+
+  const steps = [
+    { element: t("walletSync.synchronize.qrCode.steps.step1") },
+    { element: t("walletSync.synchronize.qrCode.steps.step2") },
+    { element: t("walletSync.synchronize.qrCode.steps.step3") },
+  ];
+
+  // TO CHANGE WHEN INTRAGRATION WITH TRUSTCHAIN
```
gre, line 20: notes to anticipate a bit the integration:

this code may just disappear because we need to be able to "control" the steps, it's not the step that will do a call so some api and do something based on this because we will have a global function to call at the beginning and we will just hook to the various events.

so i think we don't need to have the "displayPinCode" injected as a prop neither, it's going to be a switch of step from the parent directly.

see the usage here for instance: https://github.com/LedgerHQ/ledger-live/blob/live-12627/apps/web-tools/trustchain/components/App.tsx#L468-L479 (generally speaking, this App.tsx file will act as general/canonical examples for all trustchain integrations)

```ts
createQRCodeHostInstance({
      onDisplayQRCode: url => {
        setUrl(url);
      },
      onDisplayDigits: digits => {
        setDigits(digits);
      },
      addMember: async member => {
        // this sdk will be somehow available as a react context when we do the trustchain integration
        const jwt = await sdk.liveAuthenticate(trustchain, liveCredentials);
        await sdk.addMember(jwt, trustchain, liveCredentials, member);
      },
    })
      .catch(e => {
        if (e instanceof InvalidDigitsError) {
          return; // in this case, we could display an error of invalid digital error (it can also be handled by the generic component)
        }
        setError(e); // in this case, we render the generic error (we have a generic component for handling of errors) , typically it can be Networking error, but there could be also protocol errors
      })
      .then(() => {
        // everything is finished. the other member was added ✅ 
      });
```

gre (commented):
some notes to anticipate the integration, i let you decide if we fix this in a future PR, but the integration will have to modify a bunch of code (and not just the block of // INTEGRATION that you have defined) specifically for the qr code flow that needs to be one function call for the whole thing (since it's managing a websocket channel)

apps/ledger-live-desktop/src/newArch/WalletSync/Flows/Synchronize/02-QRCodeStep.tsx
mcayuelas-ledger: Yeah nice, I just putted that to simulate a behavior to develop UI but I'll make another PR right now to start the job on this part :)

mcayuelas-ledger: We are currently able to control which step we want to display :)

PR #7068, opened 2024-06-11
[LIVE-12968][LLM] QueuedDrawer: if onClose changes, drawer shouldn't auto close

ofreyssinet-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** Fully covered by automatic tests, I adapted the existing tests so that they would break without this fix in the same kind of situation as the one detected, and they pass.
  - all drawers in LLM

### 📝 Description

The property `onClose` was indirectly a dependency of the `useEffect` that triggers auto closing in some cleanup logic of the `QueuedDrawer` component.
The solution is to keep `onClose` in a ref in the component so that it can still be used in that `useEffect` without triggering the useEffect when its value changes.

https://github.com/LedgerHQ/ledger-live/assets/91890529/952ab900-1ba4-4d0e-909a-a416a4e5c0af

https://github.com/LedgerHQ/ledger-live/assets/91890529/54ae643a-d1fe-4448-a19b-fedbf3ed1600

### ❓ Context

- **JIRA or GitHub link**: [LIVE-12968] 

---

[LIVE-12968]: https://ledgerhq.atlassian.net/browse/LIVE-12968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/TestScreens.tsx
+++ b/apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/TestScreens.tsx
@@ -181,7 +181,7 @@ export const MainTestScreen = () => {
   const [drawer3RequestingToBeOpened, setDrawer3RequestingToBeOpened] = useState(false);
   const [drawer4ForcingToBeOpened, setDrawer4ForcingToBeOpened] = useState(false);
 
-  const handleDrawer1Close = useCallback(() => setDrawer1RequestingToBeOpened(false), []);
+  const handleDrawer1Close = () => setDrawer1RequestingToBeOpened(false); // purposely not using useCallback to check if drawer behaves well when onClose prop changes
   const handleDrawer2Close = useCallback(() => setDrawer2RequestingToBeOpened(false), []);
   const handleDrawer3Close = useCallback(() => setDrawer3RequestingToBeOpened(false), []);
   const handleDrawer4Close = useCallback(() => setDrawer4ForcingToBeOpened(false), []);
--- a/apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/index.tsx
+++ b/apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/index.tsx
@@ -81,14 +81,17 @@ const QueuedDrawer = ({
     setIsDisplayed(true);
   }, []);
 
+  const onCloseRef = useRef(onClose);
+  onCloseRef.current = onClose;
+
   const triggerClose = useCallback(() => {
     setIsDisplayed(false);
     if (drawerInQueueRef.current?.getPositionInQueue() !== 0) {
       drawerInQueueRef.current?.removeDrawerFromQueue();
       drawerInQueueRef.current = undefined;
     }
-    onClose && onClose();
-  }, [onClose]);
+    onCloseRef.current && onCloseRef.current(); // hack to avoid triggering the useEffect below if the parent changes the onClose prop
+  }, []);
 
   useEffect(() => {
     if (!isFocused && (isRequestingToBeOpened || isForcingToBeOpened)) {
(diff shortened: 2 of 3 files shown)


KVNLS (approved):
LGTM! I tested the app and it works well! Thanks!

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/TestScreens.tsx
```diff
@@ -181,11 +181,18 @@ export const MainTestScreen = () => {
...
   const handleDrawer2Close = useCallback(() => setDrawer2RequestingToBeOpened(false), []);
   const handleDrawer3Close = useCallback(() => setDrawer3RequestingToBeOpened(false), []);
   const handleDrawer4Close = useCallback(() => setDrawer4ForcingToBeOpened(false), []);
 
+  // adding this to force a re-render every 200ms and ensure the drawers behave well when rerendering
+  const [_, setNonce] = useState(0);
+  useEffect(() => {
+    const interval = setInterval(() => setNonce(nonce => nonce + 1), 200);
+    return () => clearInterval(interval);
+  }, []);
```
Justkant, line 194: This feels really hacky and if we need this I'd say we have a bigger problem somewhere else

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/index.tsx
```diff
@@ -81,14 +81,17 @@ const QueuedDrawer = ({
...
   const triggerClose = useCallback(() => {
     setIsDisplayed(false);
     if (drawerInQueueRef.current?.getPositionInQueue() !== 0) {
       drawerInQueueRef.current?.removeDrawerFromQueue();
       drawerInQueueRef.current = undefined;
     }
-    onClose && onClose();
-  }, [onClose]);
+    onCloseRef.current && onCloseRef.current(); // hack to avoid triggering the useEffect below if the parent changes the onClose prop
+  }, []);
```
Justkant, line 94: No problem with the hack here, just pointing out that they recently talked about `useEvent` for this kind of scenario
https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md
https://github.com/facebook/react/issues/14099

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/TestScreens.tsx
ofreyssinet-ledger: I'll rewrite the comment.

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/index.tsx
ofreyssinet-ledger: aaah interesting !

apps/ledger-live-mobile/src/newArch/components/QueuedDrawer/TestScreens.tsx
ofreyssinet-ledger: Would this be clearer @Justkant ?
`adding this to force a re-render every 200ms and ensure that we test that the drawers behave well when rerendering`

ofreyssinet-ledger: Actually I don't need this artificial way of re-rendering to test that scenario since the state of TestScreens already changes when requesting to open a new drawer, and it triggers the bug if I revert my fix.
I'll remove this part.

Justkant: Oh my bad, I didn't see it was only for tests purposes

PR #7124, opened 2024-06-17
[DSDK-332][LL][Passkey] Implement command and types related for the backup config.

jiyuzhuang:
### ✅ Checklist

  - ...

### 📝 Description
Implementation of App data backup commands:
#### Backup
- `getAppStorageInfo` to get App data information of format in `libs/device-core/src/commands/entities/AppStorageInfo.ts`.
    - LC: `APP_NAME_LEN`;
    - Data: `APP_NAME`.
- `backupAppStorage` to get a chunk of the App data to backup
    - No need to specify offset;
    - No need to format the data in response;
    - Only parse the response to string.
#### Restore
- `restoreAppStorageInit` to initialise restoration state on the device;
    - LC: `APP_NAME_LEN` + `BACKUP_LEN_LEN` whose length is always 4.
    - Data: `BACKUP_LEN` + `APP_NAME` where the length of the backup always take 4 places.
- `restoreAppStorage` to send backup data in chunk to the device, no need to pass offset in the command;
- `restoreAppStorageCommit` to commit the data sent in previous step on the device.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-332]

---

[DSDK-332]: https://ledgerhq.atlassian.net/browse/DSDK-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,82 @@
+import Transport, { StatusCodes, TransportStatusError } from "@ledgerhq/hw-transport";
+import { LocalTracer } from "@ledgerhq/logs";
+import type { AppStorageInfo } from "../../entities/AppStorageInfo";
+import type { APDU } from "../../entities/APDU";
+
+/**
+ * Name in documentation: INS_APP_STORAGE_GET_INFO
```
ofreyssinet-ledger, line 7: typo: `IN_APP_STORAGE_GET_INFO`

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * 0x9000: Success.
+ * 0x5123: Application not found.
+ * 0x662F: If the device is in recovery mode.
+ * 0x670A: Invalid application name length, two chars minimum.
+ */
+const RESPONSE_STATUS_SET: number[] = [
+  StatusCodes.OK,
+  StatusCodes.APP_NOT_FOUND_OR_INVALID_CONTEXT,
+  StatusCodes.CUSTOM_IMAGE_BOOTLOADER,
```
ofreyssinet-ledger, line 25: [SHOULD] It's not a good idea to reuse `CUSTOM_IMAGE_BOOTLOADER`, rather add a new entry with the same value.
It doesn't have the same meaning: in the context of Custom Lock Screen it's when the device is in bootloader mode while here it's when the device is in recovery mode.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,82 @@
...
+    transport: transport.getTraceContext(),
+    function: "getAppStorageInfo",
+  });
+  tracer.trace("Start");
+
+  const params: Buffer = Buffer.concat([
+    Buffer.from([appName.length]),
+    Buffer.from(appName, "ascii"),
+  ]);
+  GET_APP_STORAGE_INFO[4] = params;
```
ofreyssinet-ledger, line 51: [SHOULD] create a byte array like that instead of modifying the global ref, which should be a const:

`apdu = [... GET_APP_STORAGE_INFO.slice(4), params]`

libs/device-core/src/commands/entities/AppStorageInfo.ts
```diff
@@ -0,0 +1,7 @@
+export type AppStorageInfo = {
+  size: number;
+  version: string;
```
ofreyssinet-ledger, line 3: I would call this `dataVersion` to avoid any confusion with the app's version, if I recall correctly it's a different thing.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
aussedatlo: `INS` for `Instruction` ?

.changeset/spicy-coins-bow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ledgerhq/errors": minor
+"@ledgerhq/device-core": minor
+---
+
+Implemente App data backup relative commands
```
ofreyssinet-ledger, line 6: typo: `Implement`

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
jiyuzhuang: This is due to the `APDU` definition's incompatibility. In order to make it a `APDU` object, the 5th value must be defined (`Buffer` or `undefined`) which is not good. The `Lc` and `data` should be parameters and sometimes not be able to determine at the APDU definition level.

libs/device-core/src/commands/use-cases/app-backup/restoreAppStorageCommit.ts
```diff
@@ -0,0 +1,65 @@
...
+  const response = await transport.send(...RESTORE_APP_STORAGE_COMMIT, RESPONSE_STATUS_SET);
+  parseResponse(response);
+}
+
+export function parseResponse(data: Buffer): void {
+  const tracer = new LocalTracer("hw", {
+    function: "parseResponse@restoreAppStorageCommit",
+  });
+  const status = data.readUInt16BE(data.length - 2);
+  if (tracer) {
```
aussedatlo, line 56: [ASK] Is the `if` condition necessary ? You d'ont use it all the time, for example [here](https://github.com/LedgerHQ/ledger-live/pull/7124/files/936e0c40805ad8eaba6702f9eae67a3532dfa1cf#diff-b8e640d39117378a4ccc6f8db3c68bdb5800204a8cfdd4898245dd9f54cda24bR45)

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (tracer) {
+    tracer.trace("Result status from 0xe06a0000", { status });
+  }
+
+  if (status === StatusCodes.OK) {
+    const size = data.readUInt32BE(0); // Len = 4
+    const version = data.subarray(4, 8).toString(); // Len = 4
+    const hasSettings = data.readUIntBE(8, 1) === 1; // Len = 1
+    const hasData = data.readUIntBE(9, 1) === 1; // Len = 1
+    const hash = data.subarray(10, 30).toString(); // Len = 20
```
ofreyssinet-ledger, line 77: Wrong length here: from the documentation length is `0x20`, which is 32

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
ofreyssinet-ledger: Makes sense 👍

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
ofreyssinet-ledger: Ok but still you shouldn't modify a constant

ofreyssinet-ledger:
Overall remark: we should always try to add individual commands to the CLI or the REPL to make them easily testable.
With this implementation I have no way to test myself the commands until the full feature is developed. And even then I won't be able to test them independently from each other.

libs/device-core/src/commands/use-cases/app-backup/backupAppStorage.ts
```diff
@@ -0,0 +1,74 @@
...
+ * 0x6642: Invalid backup state, backup already performed.
+ */
+const RESPONSE_STATUS_SET: number[] = [
+  StatusCodes.OK,
+  StatusCodes.APP_NOT_FOUND_OR_INVALID_CONTEXT,
+  StatusCodes.FAILED_GEN_AES_KEY,
+  StatusCodes.INTERNAL_CRYPTO_OPERATION_FAILED,
+  StatusCodes.INTERNAL_COMPUTE_AES_CMAC_FAILED,
+  StatusCodes.APP_STORAGE_ENCRYPT_FAILED,
+  StatusCodes.CUSTOM_IMAGE_BOOTLOADER,
```
ofreyssinet-ledger, line 32: same remark as on `getAppStorageInfo`

jiyuzhuang:
> Overall remark: we should always try to add individual commands to the CLI or the REPL to make them easily testable. With this implementation I have no way to test myself the commands until the full feature is developed. And even then I won't be able to test them independently from each other.

The commands here are not independent. 
1. The only commands that can be tested individually is `getAppStorageInfo` and `restoreAppStorageInit`;
2. The command `backupAppStorage` must be sent after `getAppStorageInfo`;
3. `restoreAppStorage` and `restoreAppStorageCommit` must be sent after `restoreAppStorageInit`.

To make them testable via CLI or on the REPL, we must combine some of them which will be far from the command conception.

libs/device-core/src/commands/use-cases/app-backup/restoreAppStorageInit.ts
```diff
@@ -0,0 +1,85 @@
...
+ * 0x662F: Invalid device state, recovery mode.
+ * 0x5501: Invalid consent, user rejected.
+ * 0x5502: Invalid consent, pin is not set.
+ * 0x670A: Invalid application name length, two chars minimum.
+ * 0x6733: Invalid BACKUP_LEN value.
+ */
+const RESPONSE_STATUS_SET: number[] = [
+  StatusCodes.OK,
+  StatusCodes.APP_NOT_FOUND_OR_INVALID_CONTEXT,
+  StatusCodes.CUSTOM_IMAGE_BOOTLOADER,
```
ofreyssinet-ledger, line 33: [SHOULD] same remark as for other commands, define another StatusCode even if the value is the same

libs/device-core/src/commands/use-cases/app-backup/restoreAppStorageInit.ts
```diff
@@ -0,0 +1,85 @@
...
+    function: "restoreAppStorageInit",
+  });
+  tracer.trace("Start");
+
+  const params: Buffer = Buffer.concat([
+    Buffer.from([appName.length + 4]), // LC
+    Buffer.from(backupSize.toString(16).padStart(8, "0"), "hex"), // BACKUP_LEN
+    Buffer.from(appName, "ascii"), // APP_NAME
+  ]);
+  RESTORE_APP_STORAGE_INIT[4] = params;
```
ofreyssinet-ledger, line 65: [SHOULD] not modify a constant

libs/device-core/src/commands/use-cases/app-backup/restoreAppStorageCommit.ts
jiyuzhuang: It seems like that the `tracer` here always exists, I remove the condition.

libs/device-core/src/commands/use-cases/app-backup/backupAppStorage.ts
```diff
@@ -0,0 +1,74 @@
...
+  StatusCodes.INVALID_BACKUP_STATE,
+];
+
+/**
+ * Retrieves the app storage information from the device and returns it as a string.
```
ofreyssinet-ledger, line 37: It retrieves a chunk of that information, not the whole information

libs/device-core/src/commands/use-cases/app-backup/backupAppStorage.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param transport - The transport object used to communicate with the device.
+ * @returns A promise that resolves to the app storage information as a string.
```
ofreyssinet-ledger, line 40: It's rather resolving to _a chunk_ of the application storage _data_ as a string.

libs/device-core/src/commands/use-cases/app-backup/backupAppStorage.ts
```diff
@@ -0,0 +1,74 @@
...
+  const status = data.readUInt16BE(data.length - 2);
+  if (tracer) {
+    tracer.trace("Result status from 0xe06b0000", { status });
+  }
+
+  if (status !== StatusCodes.OK) {
+    throw new TransportStatusError(status);
+  }
+
+  return data.subarray(0, data.length - 2).toString("ascii");
```
ofreyssinet-ledger, line 73: [COULD] use `toString("base64")`, it's raw bytes of encrypted data so no need for ASCII.
If you make the change don't forget to update `restoreAppStorage` as well.

libs/device-core/src/commands/use-cases/app-backup/restoreAppStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Restores the application storage.
+ *
+ * @param transport - The transport object used for communication with the device.
+ * @param chunk - The chunk of data to restore.
+ * @returns A promise that resolves to a string representing the parsed response.
+ * @throws {TransportStatusError} If the response status is invalid.
+ */
+export async function restoreAppStorage(transport: Transport, chunk: string): Promise<void> {
```
ofreyssinet-ledger, line 44: I would suggest to switch to `base64` encoding for the chunk data. In which case it could be wise to rename that `chunk` parameter to `base64Chunk` to avoid any confusion.
Another approach would be to simply use a Buffer directly instead of a string.

libs/device-core/src/commands/use-cases/app-backup/restoreAppStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+  const tracer = new LocalTracer("hw", {
+    transport: transport.getTraceContext(),
+    function: "restoreAppStorage",
+  });
+  tracer.trace("Start");
+
+  const params = Buffer.concat([Buffer.from([chunk.length]), Buffer.from(chunk, "ascii")]);
```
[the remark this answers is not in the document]
ofreyssinet-ledger: It should be the length in bytes of the data.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
jiyuzhuang: It's not a constant, it's a container which should be configured at runtime.
The `const` keyword cannot be used to define an object(Array) constant, it means the object's reference is 'constant', but adding, updating and deleting properties in the object are permitted and always happen.
The possible solution here is to redefine the `APDU` object with only the first four values mandatory and constant by adding `as const` or `readonly` keyword. Then combine with the rest at runtime.

ofreyssinet-ledger:
@jiyuzhuang 
> The commands here are not independent.
> 
> 1. The only commands that can be tested individually is `getAppStorageInfo` and `restoreAppStorageInit`;
> 2. The command `backupAppStorage` must be sent after `getAppStorageInfo`;
> 3. `restoreAppStorage` and `restoreAppStorageCommit` must be sent after `restoreAppStorageInit`.
> 
> To make them testable via CLI or on the REPL, we must combine some of them which will be far from the command conception.

I know that they have to be sent in a specific order, the person using the REPL or CLI can do that themselves if they have access to each individual command.
The restoring one will be trickier to test but at least they can be tested.
It just seems a bit crazy for me to validate the implementation of those commands if it has never been tested for real.
Maybe we don't do it in this PR, but to me we should make a task that is exactly about that: making those commands testable outside of Ledger Live.
The same was made for the custom lock screen.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
jiyuzhuang: Adding new entry with the same value has nothing to do to clarify the meaning. The error will always [find the first match value](https://github.com/LedgerHQ/ledger-live/blob/5d17fab23f59b888190cd9e122165be6d0eb8c55/libs/ledgerjs/packages/errors/src/index.ts#L320) then return its key. So either differentiating the value at embedded team or using a different error system (with context) can be the solution.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
ofreyssinet-ledger: Yes by constant I meant that we should probably add `"as const"` and also if it's in `FULL_CAPS` it usually means that it's a constant.

ofreyssinet-ledger (changes requested):
Good job 👍 
A few things overall to change:
- `StatusCodes.CUSTOM_IMAGE_BOOTLOADER` should not be reused in my opinion even if the status code for "device is in recovery mode" has the same value. It's a bit confusing, and bootloader mode / recovery mode are not the same thing.
- this thing with changing the value of variables that are written in full caps.
- naming of `AppStorageInfo.version` -> `dataVersion` seems slightly better.
- the length of the hash is wrong in `getAppStorageInfo` (20, should be 32)
- some TS doc to be made more accurate in `backupAppStorage`
- format of chunks string: ASCII or base64 ? I'd say base64.
- input of `backupAppStorage`: maybe a Buffer instead of a string

And I think it would be a great addition to make all those commands accessible in the CLI, even if they are not "independent" in the sense that they have to be called in a specific sequence. It's also the case for the commands related to fetching/loading the custom lock screen and they are still available. It allows anyone to test "independently", meaning, without the bias/side-effects/potential bugs of a future LL implementation.

jiyuzhuang:
About the encoding, I think it's better to send / receive the buffer directly and encode / decode them at higher level. If not, calculating the base64 encoded string will make it more complicated.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,80 @@
...
+  const status = data.readUInt16BE(data.length - 2);
+  tracer.trace("Result status from 0xe06a0000", { status });
+
+  if (status === StatusCodes.OK) {
+    const size = data.readUInt32BE(0); // Len = 4
+    const dataVersion = data.subarray(4, 8).toString(); // Len = 4
+    const hasSettings = data.readUIntBE(8, 1) === 1; // Len = 1
+    const hasData = data.readUIntBE(9, 1) === 1; // Len = 1
+    const hash = data.subarray(10, 42).toString(); // Len = 20
+    return { size, dataVersion, hasSettings, hasData, hash };
```
ofreyssinet-ledger, line 76: we should use the `ApduParser` for this kind of stuff, it will make it much easier to read and avoid bugs

ofreyssinet-ledger, line 76: It just occurred to me that we could extract the `ApduParser` from the Device SDK to do this kind of stuff 🤔 It would be super helpful.

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,80 @@
...
+  });
+  const status = data.readUInt16BE(data.length - 2);
+  tracer.trace("Result status from 0xe06a0000", { status });
+
+  if (status === StatusCodes.OK) {
+    const size = data.readUInt32BE(0); // Len = 4
+    const dataVersion = data.subarray(4, 8).toString(); // Len = 4
+    const hasSettings = data.readUIntBE(8, 1) === 1; // Len = 1
+    const hasData = data.readUIntBE(9, 1) === 1; // Len = 1
+    const hash = data.subarray(10, 42).toString(); // Len = 20
```
ofreyssinet-ledger, line 75: the comment here is outdated
`Len = 32`

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
ofreyssinet-ledger: ok 👍 indeed our system isn't perfect

libs/ledgerjs/packages/errors/src/index.ts
```diff
@@ -243,7 +243,7 @@ export const StatusCodes = {
   CONDITIONS_OF_USE_NOT_SATISFIED: 0x6985,
   CONTRADICTION_INVALIDATION: 0x9810,
   CONTRADICTION_SECRET_CODE_STATUS: 0x9808,
-  CUSTOM_IMAGE_BOOTLOADER: 0x662f,
+  DEVICE_IN_RECOVERY_MODE: 0x662f,
```
ofreyssinet-ledger, line 246: can you then rename to `DEVICE_IN_BOOTLOADER_OR_RECOVERY_MODE` ? it's a bit long but since there are 2 use cases 🤷‍♂️

libs/device-core/src/commands/use-cases/app-backup/getAppStorageInfo.ts
```diff
@@ -0,0 +1,102 @@
...
+      let offset = 0;
+      const size = data.readUInt32BE(offset); // Len = 4
+      offset += 4;
+      const dataVersion = data.subarray(offset, offset + 4).toString(); // Len = 4
+      offset += 4;
+      const hasSettings = data.readUIntBE(offset, 1) === 1; // Len = 1
+      offset += 1;
+      const hasData = data.readUIntBE(offset, 1) === 1; // Len = 1
+      offset += 1;
+      const hash = data.subarray(offset, offset + 32).toString(); // Len = 32
```
ofreyssinet-ledger, line 89: 👍

libs/device-core/src/commands/use-cases/app-backup/backupAppStorage.ts
```diff
@@ -0,0 +1,96 @@
...
+  StatusCodes.GEN_AES_KEY_FAILED,
+  StatusCodes.INTERNAL_CRYPTO_OPERATION_FAILED,
+  StatusCodes.INTERNAL_COMPUTE_AES_CMAC_FAILED,
+  StatusCodes.ENCRYPT_APP_STORAGE_FAILED,
+  StatusCodes.DEVICE_IN_RECOVERY_MODE,
+  StatusCodes.INVALID_BACKUP_STATE,
+];
+
+/**
+ * Retrieves the app storage information from the device and returns it as a buffer.
```
ofreyssinet-ledger, line 44: This comment is inaccurate: it retrieves a chunk of that app's data.

ofreyssinet-ledger (approved):
lgtm 👍

mbertin-ledger (approved):
👍

PR #7126, opened 2024-06-17
chore(live-wallet): implement Wallet Sync <> Accounts management

gre:
**Recommendation for reviews: Read the files changed from the bottom! the important changes are at the end, most of the +3k changes are in test and web tools parts and not the important ones**. Please also ack the PR description to understand the mecanism at stake.

### ✅ Checklist

  - a lot of unit tests were added in live-wallet. including specific module tests and generic tests (with data generators)
  - live-wallet library

#### Generalisation of `@ledgerhq/live-wallet/cloudsync` module

the cloudsync module is now independent of the data and API location. In our case, Ledger Live will use the `{accounts,accountNames}` data type (evolutive) with the `"live"` API slug.

#### Introduces `WalletSyncDataManager` in `@ledgerhq/live-wallet/walletsync`

WalletSyncDataManager is a module that is responsible for managing the reconciliation of incremental data updates. We distinguish local data from distant data: local data is client-side data whereas distant data is the subset of data that we push to cloud sync (essentially the identifier parts of the data that local data can be restored from).

**Key Functions**

1. **Determine Changes**: We need to determine if there are changes between the local state and the distant state that we need to push. This is done by `diffLocalToDistant(localState, latestDist)`.

2. **Resolve Incoming State**: When receiving a new distant state from cloud sync, the module must calculate the update to apply to the local state. `resolveIncomingDistantState` calculates the update and `applyUpdate` applies it.

**State Transition**

State transition is done incrementally with a diff, moving from A to B is essentially:

- `resolveIncomingDistantState`: Solving the transition (`distA -> distB`) from `stateA`
- `applyUpdate`: Applying that transition update to get to `stateB`

So in other words:

    stateB = stateA + (distB - distA)

**Glossary:**

- **LocalState**: All the data that the client has locally that the module needs as an input for the reconciliation.
- **Update**: An action payload to express the update mutation that you need to do on the local state after determining an update to do with distant state changes.
- **Schema**: The Schema is a Zod type that allows validating the data that this module will store in cloud sync data. The `typeof schema`. **IMPORTANT**: the schema must not change over time.
- **DistantState**: The type that corresponds to the Schema. Basically, the exact data that we store in cloud sync for that module. (NB: it is automatically inferred from the Schema)

```typescript
export interface WalletSyncDataManager<
  LocalState,
  Update,
  Schema extends ZodType,
  DistantState = z.infer<Schema>,
> {
  schema: Schema;

  diffLocalToDistant: (
    localData: LocalState,
    latestState: DistantState | null,
  ) => DistantDiff<DistantState>;

  resolveIncomingDistantState: (
    ctx: WalletSyncDataManagerResolutionContext,
    localData: LocalState,
    latestState: DistantState | null,
    incomingState: DistantState | null,
  ) => Promise<UpdateDiff<Update>>;

  applyUpdate: (localData: LocalState, update: Update) => LocalState;
}
```

#### a modular architecture for WalletSync

The `{accounts,accountNames}` data type being something evolutive, we propose here a generic system that will allows us to build up new modules for future new fields in this data.

The types are now entirely inferred from the zod schema, and aggregated from all available modules, as shown in this screenshot:

Capture d’écran 2024-07-02 à 16 07 25

Each module focus only on its field. For instance, accountNames will only deal with managing accountNames reconciliation.

#### Integration with CloudSync and TrustchainSdk

A full feature working implementation example in done in web tools.

Web Tools implement full account synchronisation and reconciliation, with the ability to edit account names and add more accounts with a hardware wallet.

you can already see an example implementation of Wallet Sync integration, the important part is the main loop that implement wallet sync:

We create a **CloudSyncSDK**:

```ts
  const walletSyncSdk = useMemo(
    () =>
      new CloudSyncSDK({
        slug: "live",
        schema: walletsync.schema,
        trustchainSdk,
        getCurrentVersion,
        saveNewUpdate,
      }),
    [trustchainSdk, getCurrentVersion, saveNewUpdate],
  );
```

it needs an implementation of `saveNewUpdate` which implements the logic to dispatch a state update. it is heavily dependents on the way each implementation (LLD/LLM/..) manage the store which is why it's not in the cloud sync sdk / walletsync side. However, you will have `walletsync.resolveIncomingDistantState` and `walletsync.applyUpdate` helpers to manage most of the actual diffing and reconciliation logic.

```typescript
  // saveNewUpdate implements the state update logic
  const saveNewUpdate = useCallback(
    async (event: UpdateEvent<DistantState>) => {
      switch (event.type) {
        case "new-data": {
          const state = stateRef.current;
          const version = event.version;
          const data = event.data;
          const wsState = state.walletState.wsState;
          const localState = localStateSelector(state);
          const ctx = { getAccountBridge, bridgeCache, blacklistedTokenIds: [] };

          const resolved = await walletsync.resolveIncomingDistantState(ctx, localState, wsState.data, data,);

          if (resolved.hasChanges) {
            // the important part is to keep one atomic update for all the states
            setState(s => {
              const localState = localStateSelector(s);
              const newLocalState = walletsync.applyUpdate(localState, resolved.update);

              // we now need to "reverse" the localStateSelector back into our own internal state
              let walletState = s.walletState;
              walletState = walletH.BULK_SET_ACCOUNT_NAMES(walletState,setAccountNames(newLocalState.accountNames));
              walletState = walletH.WALLET_SYNC_UPDATE(walletState, walletSyncUpdate(data, version));
              return {
                accounts: newLocalState.accounts.list, // save new accounts
                walletState,
              };
            });
          }
          break;
        }
        case "pushed-data": {
          // when we push the state, we also need to implement the update on the local state
          setState(s => ({
            ...s,
            walletState: walletH.WALLET_SYNC_UPDATE(s.walletState, walletSyncUpdate(event.data, event.version)),
          }));
          break;
        }
        case "deleted-data": {
          console.log("deleted data");
          setState(s => ({
            ...s,
            walletState: walletH.WALLET_SYNC_UPDATE(s.walletState, walletSyncUpdate(null, 0)),
          }));
          break;
        }
      }
    },
    [bridgeCache, setState],
  );
```

**Finally we can start the watch loop that will constantly observe local changes, push them if necessary and manage incoming updates of wallet sync**

```ts
  const [visualPending, setVisualPending] = useState(true);

  // pull and push wallet sync loop
  useEffect(() => {
    const { unsubscribe } = walletSyncWatchLoop({
      walletSyncSdk,
      trustchain,
      memberCredentials,
      setVisualPending,
      getState: () => stateRef.current,
      localStateSelector,
      latestDistantStateSelector,
    });
    return unsubscribe;
  }, [trustchainSdk, walletSyncSdk, trustchain, memberCredentials]);
  ```

It's worth nothing that each time trustchain changes, you will indeed need to restart the watch loop. we will also be able to catch error cases if necessary (many other parameters can be passed)

Capture d’écran 2024-07-02 à 16 15 08

#### Future notes for integration in LLD and LLM 

In the real LLD/LLM implementation, we use redux, so the implementation will differ, here are some important parts:
- that we don't inject the redux state as a retrigger for effect/callbacks, instead we will have to use redux's getState to access the state imperatively. 
- we will have to define a selector of "localState", which is the input of the Wallet Sync Aggregator, in the case of the web tools, we have used simple state:
```
 const localStateSelector = (state: State) => ({
  accounts: {
    list: state.accounts,
  },
  accountNames: state.walletState.accountNames,
});
```
- we will have to implement the "reverse" of that selector, which is to update the state with the data, for this it will be important to do it with only one action, things rely on the fact it is an atomic sync update.
 

### ❓ Context

- **JIRA or GitHub link**: LIVE-12639 LIVE-13122 LIVE-13204 LIVE-13120

### Limitations and remaining topics to solve

- LIVE 12973 when a removeMember is done, we need in wallet sync to push a new version to reapply the new encryption. 
- LIVE 13312 `wsStateNonImportedAccountIds: string[]` management must be done on accounts that could possibly fail the synchronisation (during the resolve). At the moment they become completely ignored and will never appear on other members. (fail sync = the user had network issues / the backend had issues / the Ledger Live instance don't yet support the coin)
- LIVE 13313 we will need to have a websocket implementation of the watch loop too.
- LIVE 13218 discussion on some key topics

---

### CODE REVIEW

File: libs/live-wallet/src/store.ts [hunk 1]
```diff
@@ -172,3 +210,83 @@ export const accountRawToAccountUserData = (raw: AccountRaw): AccountUserData =>
...
+      hasChanges = true;
+    }
+  }
+
+  // let's figure out the locally deleted accounts that we will need to take into account
+  const removed: string[] = [];
+  const localAccountIds = new Set(localAccounts.map(a => a.id));
+  for (const id of distantServerAccountIds) {
+    if (!localAccountIds.has(id)) {
+      removed.push(id);
```
gre, line 261: TBD: how to manage the wsStateNonImportedAccountIds. maybe we shouldn't be considering the account ids of wsStateNonImportedAccountIds in this?

libs/live-wallet/src/store.ts
```diff
@@ -7,18 +7,31 @@ import { Account, AccountLike, AccountRaw, AccountUserData } from "@ledgerhq/typ
...
   starredAccountIds: Set<string>;
+
+  /////
+  // FIXME this part of the state is not correctly saved/restored at the moment, we need to improve the INIT_ACCOUNTS to include these data!
+
+  // local copy of the wallet sync data last synchronized with the backend of wallet sync, in order to be able to diff what we need to do when we apply an incremental update
+  wsState: WalletSyncState;
+
+  // list of account ids that are not imported in the wallet sync data, because we need to be able to retry them later. possibly on startup.
+  wsStateNonImportedAccountIds: string[];
```
gre, line 27: TODO: we need a way on the web tools to fake a non supported coin and be able to test this.

gre (commented):
self notes

apps/web-tools/trustchain/components/AppAccountsSync.tsx
```diff
@@ -0,0 +1,340 @@
...
+}: {
+  bridgeCache: BridgeCacheSystem;
+  setAccounts: (_: (_: Account[]) => Account[]) => void;
+}) {
+  // merge accounts with the existing ones
+  const addAccounts = useCallback(
+    (accounts: Account[]) => {
+      setAccounts(state => {
+        const existingSet = new Set(state.map(a => a.id));
+        return state.concat(accounts.filter(a => !existingSet.has(a.id)));
```
gre, line 194: TBD: something broken in this logic? experiencing dups

apps/web-tools/trustchain/components/AppAccountsSync.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const saveNewUpdate = useCallback(
+    async (event: UpdateEvent) => {
+      // in this current version, we just display the data as is, but in real app we would first reconciliate the account data and manage the sync
+      switch (event.type) {
+        case "new-data": {
+          const version = event.version;
+          const data = event.data;
+          const wsState = stateRef.current.walletState.wsState;
+
+          // TODO there remain a case to solve: when an account failed to be resolved in the past, we would need to eventually retry it. we could do this basically by adding in the resolved the failed
```
gre, line 76: TODO

apps/web-tools/trustchain/components/App.tsx
```diff
@@ -76,6 +77,8 @@ const App = () => {
         ? "PROD"
         : "MIXED";
 
+  // TODO when Accounts Sync is on, we need to share the WS SDK instance and the Wallet Sync SDK section becomes read only. the version is also shared top level (so are accounts?!) in order for us to display Version: X next to the title
```
gre, line 80: TODO

libs/live-wallet/src/store.ts
```diff
@@ -7,18 +7,31 @@ import { Account, AccountLike, AccountRaw, AccountUserData } from "@ledgerhq/typ
...
 
 export type WalletState = {
   // user's customized name for each account id
   accountNames: Map<string, string>;
 
   // a set of all the account ids that are starred (NB: token accounts can also be starred)
   starredAccountIds: Set<string>;
+
+  /////
+  // FIXME this part of the state is not correctly saved/restored at the moment, we need to improve the INIT_ACCOUNTS to include these data!
```
gre, line 21: TODO

libs/live-wallet/src/store.ts
```diff
@@ -56,7 +71,12 @@ export const handlers: WalletHandlers = {
         starredAccountIds.add(starredId);
       }
     });
-    return { accountNames, starredAccountIds };
+    return {
+      accountNames,
+      starredAccountIds,
+      wsState: { data: null, version: 0 },
```
gre, line 77: TODO

libs/live-wallet/src/store.ts
```diff
@@ -91,6 +115,24 @@ export const handlers: WalletHandlers = {
...
+        failures,
+        newState: { data, version },
+      },
+    },
+  ) => {
+    const wsStateNonImportedAccountIds = Object.keys(failures);
+    return {
+      ...state,
+      accountNames: newNamesState,
+      wsStateNonImportedAccountIds,
```
gre, line 132: problem 1: we override it over. we need to fusion it and remove those who were treated.
problem 2: in a generalisation, how would we address this state. do we really need to store this state ultimately, can we do without?

KVNLS (approved):

apps/web-tools/trustchain/components/AppCloudSync.tsx
```diff
@@ -88,7 +86,13 @@ export function AppWalletSync({
   );
 
   const walletSyncSdk = useMemo(() => {
-    return new CloudSyncSDK({ trustchainSdk, getCurrentVersion, saveNewUpdate });
+    return new CloudSyncSDK({
+      slug: "live",
```
mcayuelas-ledger, line 90: You created a const for that no?

gre: in another PR.

libs/live-wallet/src/cloudsync/sdk.ts
```diff
@@ -1,34 +1,50 @@
...
 import api, { JWT } from "./api";
 import Base64 from "base64-js";
 import { compress, decompress } from "fflate";
 import { Observable } from "rxjs";
+import { z, ZodType } from "zod";
 
-export type UpdateEvent =
+export type UpdateEvent<Data> =
   | {
       type: "new-data";
```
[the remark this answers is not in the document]
gre: i suggest we do this somewhere else, not the scope of the PR change.

libs/live-wallet/src/cloudsync/__tests__/sdk.test.ts
```diff
@@ -0,0 +1,342 @@
...
+describe("CloudSyncSDK basics", () => {
+  const port = 54034;
+  const base = "http://localhost:" + port;
+
+  setEnv("CLOUD_SYNC_API", base);
+
+  let storedData: string | null = null;
+  let storedVersion = 0;
+
+  const handlers = [
```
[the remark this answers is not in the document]
mcayuelas-ledger: btw it's on different folder... forget it 

Just if we can move it outside just for readablity

libs/live-wallet/src/cloudsync/sdk.ts
```diff
@@ -1,34 +1,50 @@
...
+    slug,
+    schema,
     trustchainSdk,
     getCurrentVersion,
     saveNewUpdate,
   }: {
+    /**
+     * slug used with cloud sync API ((example "live")
+     */
+    slug: string;
```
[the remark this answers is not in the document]
gre: no, this is a string. the CloudSync implementation is generic, it needs a slug:string,
then on the WalletSync side, we hardcode it to `"live"`https://github.com/LedgerHQ/ledger-live/pull/7298/files#diff-0b2b5452ff58b687fd73d274d635cf0f0f9ee2f7fa39413d600fe392d395f78eL37

libs/live-wallet/src/cloudsync/__tests__/sdk.test.ts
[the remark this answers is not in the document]
gre: we could have helper, but i think it's better to collocate this, this deeply work with the tests we have in context here: we have contextual `storedData` and `storedVersion`, etc..

PR #6534, opened 2024-03-25
introduce wallet store, decoupling of account user's data (name, starred,..) from the account's blockchain data

gre:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  - all projects (LLD, LLM, CLI)
  - Add Accounts flow
  - State management of accounts (load / save of accounts in the app.json / llm's storage)
  - ordering of accounts
  - starring an Account feature (LLD)
  - Account name edition
  - Wallet API and its ability to send the user's account names
  - export to mobile feature (LiveQR)
  - account search by name

#### Goals

initial specification: https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/4574642221/move+Account+user+data+states+out+of+Account+model

**The main goal** is to make the type `Account` exclusively having "blockchain data" and no longer any "user data" fields. (example of user data field includes: starred, name, unit,... those are editable by the end user and typically can't be restored from the blockchain). By splitting these user data fields out, we simplify the logic of the Account[] lifecycle as well as allowing ourself a future where we can work on different ways to "export/import" user data (aka "wallet sync") more easily and no longer with data "nested" among the blockchain public/immutable data.

The fields that are dropped by this PR are:
- **Account#name**
- **Account#starred** and **TokenAccount#starred**

**The secondary goal** includes the modularisation of live-common by exporting all logic related to account name management as well as lifecycle of some account data that need the `account.name` to work (the UI logic for addAccounts, ordering of accounts, importAccounts). We have therefore created a library `live-wallet` to allow this modularization.

#### Backward compatible!

✅  There are no migration involved!
- We have preserved the same app.json data. Which means the AccountRaw has now been touched. Instead, we have made that AccountRaw will be split into two parts: Account + AccountUserData

this is why you can see in the code that we have moved from

```
DataModel<AccountRaw, Account>
```

to
```
DataModel<AccountRaw, [Account, AccountUserData]>
```

To make this work, the underlying function `toAccountRaw` has been reworked to include a userData as a second parameter, as seen in the implementation:

```ts
decode: (raw: AccountRaw) => [fromAccountRaw(raw), accountRawToAccountUserData(raw)],
encode: ([account, userData]: [Account, AccountUserData]): AccountRaw => toAccountRaw(account, userData),
```

#### `live-wallet` breakdown

The `live-wallet` library currently includes:

[...]

PR #7148, opened 2024-06-20
fix(llm): improve typing of ReceiveFunds/02-SelectAccount and fix crash case

gre:
### ✅ Checklist

  - no tests yet but covered by type safety that wasn't correctly made before my change, on the whole SelectAccount file. 
  - Receive funds flow when selecting a token

### 📝 Description

**Root cause:** the rework of https://github.com/LedgerHQ/ledger-live/pull/6534

Capture d’écran 2024-06-21 à 13 38 13

as seen here, we did a mistake on LLM ReceiveFunds/02-SelectAccount file where we passed a `Currency` instead of an `Account` on the getDefaultAccountName function.
unfortunately the bad typing was hiding the issue: bad casting using `item.parentAccount as Account` which hides the problem. because actually .parentAccount is a `Account | undefined`. since we didn't have the | undefined part, TypeScript will infer the code as a `Account` and not as an `Account | Currency` that is actually the type of the previously bad code here.

the fix:

Capture d’écran 2024-06-21 à 13 36 08

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-13092

---

### DIFF
--- a/apps/ledger-live-mobile/src/screens/ReceiveFunds/02-SelectAccount.tsx
+++ b/apps/ledger-live-mobile/src/screens/ReceiveFunds/02-SelectAccount.tsx
@@ -4,7 +4,7 @@ import { useSelector } from "react-redux";
 
 import { Button, Flex, Text } from "@ledgerhq/native-ui";
 import { useTranslation } from "react-i18next";
-import { Account, AccountLike, SubAccount, TokenAccount } from "@ledgerhq/types-live";
+import { Account, SubAccount, TokenAccount } from "@ledgerhq/types-live";
 import { makeEmptyTokenAccount } from "@ledgerhq/live-common/account/index";
 import { flattenAccountsByCryptoCurrencyScreenSelector } from "~/reducers/accounts";
 import { NavigatorName, ScreenName } from "~/const";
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/screens/ReceiveFunds/02-SelectAccount.tsx
```diff
@@ -82,16 +85,16 @@ function ReceiveSelectAccount({
...
         track("account_clicked", {
           asset: currency.name,
           page: "Select account to deposit to",
         });
         navigation.navigate(ScreenName.ReceiveConfirmation, {
           ...route.params,
-          accountId: (account as SubAccountEnhanced)?.parentId || account.id,
-          createTokenAccount: (account as SubAccountEnhanced)?.triggerCreateAccount,
+          accountId: (account.type !== "Account" && account?.parentId) || account.id,
+          createTokenAccount: "triggerCreateAccount" in account && account?.triggerCreateAccount,
```
KVNLS, line 97: I don't think that we need to test it twice, am I missing something?
```suggestion
          createTokenAccount: "triggerCreateAccount" in account,
```

KVNLS (changes requested):

KVNLS (changes requested):
When I look at the Sentry error, the error happens in `live-wallet/lib/accountName.js`
Could make the function resilient, please?

```
const getDefaultAccountName = (account) => {
    if (account.type === "Account") {
        return (0, exports.getDefaultAccountNameForCurrencyIndex)(account);
    }
    return account.token.name;
};
```
Here we are not sure that we can access account.token.name.
Could you add the unit test to simulate the issue we had, please?

gre:
@KVNLS the error is not in live-wallet, it's a misusage of the `getDefaultAccountName` function. Instead of giving an Account we have a CryptoCurrency.

KVNLS (approved):

PR #7226, opened 2024-06-28
TrustchainSDK to become stateful on the JWT management to ease LLD/LLM integration error management

gre:
### ✅ Checklist

  - ...

### 📝 Description

**BEFORE THIS PR**, the SDK was like this

    const jwt = await sdk.auth(creds);
    await sdk.doSomething(jwt, ...params);

which means it’s your responsability to manage the jwt state, including all the expiration strategy management as well as auth failure problems.

typically, here are some (non exhaustive) typical things to do

after 5mn, the token expires but we can just refreshToken to obtain a new one with just a http call

after 1h, the token fully expires and we need to reauth completely (two http calls: challenge+signature)

any http calls can fail due to changes done by another member (e.g. another member removed you, another member have destroyed the trustchain,…)

at the moment, the sdk don’t handle at all these error cases and it’s hard for userland side (lld/llm) to manage them in unified way, ultimately this unification can be solved by the SDK.

**WITH THIS PR**: like a browser who automatically manage cookie sessions

    await sdk.doSomething(...params);

jwt becomes implicit in the sdk.

---

**refer to the PR diff of types for the breaking changes impact**

### ❓ Context

- **JIRA or GitHub link**: 

---


apps/web-tools/trustchain/components/App.tsx
```diff
@@ -83,9 +84,20 @@ const App = () => {
         ? "PROD"
         : "MIXED";
 
+  const [deviceInteractionVisible, setDeviceInteractionVisible] = useState(false);
+  const callbacks = useMemo(
+    () => ({
+      onStartRequestUserInteraction: () => setDeviceInteractionVisible(true),
+      onEndRequestUserInteraction: () => setDeviceInteractionVisible(false),
```
mcayuelas-ledger, line 91: It will not work for all case, in this case it's works because it's very atomical and driven by click but on the LLD UI sometimes  you don't have the same amount of interactions due to the current flow you are. It means that sometimes callbacks are not called

libs/trustchain/src/__tests__/sdk/create2trustchainInARow.json
mcayuelas-ledger, line 1: It should be better to split in two folders : 

- Mock/response
- Tests files

btw could you rename tests files like we do in LLM/LLD to keep consistency?
 like `create2trustchainInARow.test.ts` insteand of  `create2trustchainInARow.ts`

libs/trustchain/src/mockSdk.ts
```diff
@@ -60,27 +57,27 @@ export class MockSDK implements TrustchainSDK {
...
-  async authWithDevice(transport: Transport): Promise<JWT> {
-    void transport;
-    return Promise.resolve(mockedSeedIdAccessToken);
-  }
-
-  async auth(trustchain: Trustchain, memberCredentials: MemberCredentials): Promise<JWT> {
+  withAuth<T>(
+    trustchain: Trustchain,
+    memberCredentials: MemberCredentials,
+    f: (jwt: JWT) => Promise<T>,
```
mcayuelas-ledger, line 63: could you avoid using signe char to name function/const ? 🫣 I know it's for test purpose but it more easy to read !

libs/trustchain/src/sdk.ts
```diff
@@ -24,91 +25,84 @@ import Transport from "@ledgerhq/hw-transport";
...
     this.context = context;
   }
 
-  async initMemberCredentials(): Promise<MemberCredentials> {
-    const kp = await crypto.randomKeypair();
-    return convertKeyPairToLiveCredentials(kp);
+  withAuth<T>(
+    trustchain: Trustchain,
+    memberCredentials: MemberCredentials,
+    f: (jwt: JWT) => Promise<T>,
```
mcayuelas-ledger, line 49: same as above :)

libs/trustchain/src/sdk.ts
```diff
@@ -309,6 +295,157 @@ export class SDK implements TrustchainSDK {
...
+): Promise<T> {
+  function refresh(jwt: JWT) {
+    return api.refreshAuth(jwt).catch(e => {
+      log("trustchain", "JWT refresh failed, reauthenticating", e);
+      if (networkCheckJwtExpiration(e).hasExpired) {
+        return auth();
+      }
+      throw e;
+    });
+  }
```
mcayuelas-ledger, line 330: You can move it outside the function no?

libs/trustchain/src/test-scenarios/getOrCreateTransactionCases.ts
mcayuelas-ledger, line 1: Generally speaking, for the test part, I don't know what lib is present in trustchain but it would have been nice to have something uniform like with Jest with the "describe" and "test". It makes it easier to understand what's going on and what we're testing.

mcayuelas-ledger (changes requested):

libs/trustchain/src/sdk.ts
gre: It's only a function to factorize the code of the two calls that contains logic specific to the general recovering of errors, not meant to be used externally also auth() is contextual here

libs/trustchain/src/__tests__/sdk/create2trustchainInARow.json
gre: Good idea but I think we have a wider topic to figure out first: @KVNLS wanted these tests to be outside of src folder, because today they are included in the build, I didn't manage to solve this yet because when you put files on top level , the dist output folder will still include them as soon as they are imported and will have a src sub level.

These are good point but I suggest we solve this in another PR because this PR just added some more test on top of existing systems

apps/web-tools/trustchain/components/App.tsx
gre: I'm not sure I understand, let's discuss today.

libs/trustchain/src/test-scenarios/getOrCreateTransactionCases.ts
gre: These are end to end scenario script that indeed currently don't depends on jest. We'll also have unit tests for more unitary and classical-jest tests.

The problem is that they all need to run in the same spot.
But I was recently thinking of a possible improvement in that we could have "step" that explicit what we do on each step.

Like

```ts

const trustchain = await step("member1 creates a trustchain", async () => {
  return await sdk1.getOrCreateTrustchain(...
})

...etc
```

This would help on scenario visibility + better log when it crashes (if we decorate the error that is thrown on a step) + we could make the .json grouped by steps which could allow better readability and also having ability to edit the json more easily if we isolates the APDU/network of each step etc..

Now it's some more possible evolution idea and it all depends what we want to continue building in priority ( cc @KVNLS )

mcayuelas-ledger (approved):

PR #7262, opened 2024-07-04
chore(trustchain): simplify test folders

gre:
### ✅ Checklist

  - tests only

### 📝 Description

**the lib/ output of the library no longer is polluted by test scenarios**

Capture d’écran 2024-07-04 à 14 29 33

**test scenarios and test helpers are in `/tests` and scenario records are in `/mocks`** outside of the src folder

(NB: it can't be a `__tests__` because Jest default's config would still consider these as runnable tests)

**a `sdkForName` helper is given to scenario to simplify the boilerplate you need to do to create a sdk**

Capture d’écran 2024-07-04 à 16 05 45

**no longer need to generate individual jest test for each scenario**

We simplify the necessary boilerplate to run the tests by having one test file that run them all:

```ts
fs.readdirSync(mockFolder).forEach(file => {
  if (!file.endsWith(".json")) return;
  const slug = file.slice(0, -5);
  const json = require(path.join(mockFolder, file));
  const mod = require(path.join(scenarioFolder, slug + ".ts"));
  test(slug, async () => {
    const scenario = mod.scenario;
    await replayTrustchainSdkTests(json, scenario);
  });
});

```

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/trustchain/tests/scenarios/_template.ts
```diff
@@ -1,15 +1,14 @@
 import Transport from "@ledgerhq/hw-transport";
```
KVNLS, line 1: Can you keep the convention we have in most projects for `tests` and `mocks` folders, please? By using `__tests__` and `__mocks__`

libs/trustchain/src/__tests__/integration/mock.sdk.test.ts
```diff
@@ -13,7 +15,7 @@ const nonMockableScenarios = [
   "userRefusesAuth", // can't simulate device interaction at the moment
 ];
 
-const scenarioFolder = path.join(__dirname, "./test-scenarios");
+const scenarioFolder = path.join(__dirname, "../../../tests/scenarios");
```
KVNLS, line 18: Maybe we can add aliases for mocks and tests folder, wdyt?

KVNLS (changes requested):

libs/trustchain/src/__tests__/integration/mock.sdk.test.ts
gre: not possible with node.js path resolvers. here we will fs.read the files, it's not a require

libs/trustchain/tests/scenarios/_template.ts
gre: ok but the problem is it doesn't work for the default jest config that wants to runs the files inside these even if they are outside of src. shouldn't we stay relatively jest standard?
(on top of these, they are not even jest tests)

gre:
@mcayuelas-ledger which test files?
i'm ok with the idea for all files that need to be run by Jest, which is not the case of scenario files that are NOT test files.

PR #7280, opened 2024-07-08
chore: generalize encryption to encrypt trustchain store

gre:
### ✅ Checklist

  - LLD password encryption

### 📝 Description

- app.accounts is the main field that we use to detect if data is encrypted
- but other fields can be introduced as encrypted too and actions done to set the encryption key acts on all
- also rename trustchainStore to trustchain in the app.json

to check if the field is encrypted:

```
cat app.json | jq '.data.trustchain'
```

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-13307

---


gre:
@mcayuelas-ledger it appears that there is a bug that is found by tests 🤔  , if you have time to dig tomorrow

apps/ledger-live-desktop/src/renderer/storage.ts
```diff
@@ -57,7 +57,7 @@ type DatabaseValues = {
   countervalues: Countervalues;
   postOnboarding: PostOnboarding;
   settings: Settings;
-  trustchainStore: TrustchainStore;
+  trustchain: TrustchainStore;
```
mcayuelas-ledger, line 60:
```suggestion
  trustchain: Trustchain;
```

apps/ledger-live-desktop/src/renderer/middlewares/db.ts
```diff
@@ -52,12 +52,11 @@ const DBMiddleware: Middleware<{}, State> = store => next => action => {
   } else if (DB_MIDDLEWARE_ENABLED && action.type.startsWith(trustchainStoreActionTypePrefix)) {
     next(action);
     const state = store.getState();
-    setKey("app", "trustchainStore", trustchainStoreSelector(state));
+    setKey("app", "trustchain", trustchainStoreSelector(state));
   } else if (DB_MIDDLEWARE_ENABLED && action.type.startsWith("MARKET")) {
     next(action);
     const state = store.getState();
     setKey("app", "market", marketStoreSelector(state));
-  } else {
```
mcayuelas-ledger, line 60: You should not remove this `else`

apps/ledger-live-desktop/src/main/db/index.ts
```diff
@@ -86,41 +86,70 @@ async function reload() {
...
+async function removeEncryptionKey() {
+  const nsToSave = new Set<string>();
+  for (const [ns, keyPath] of encryptedDataPaths) {
+    nsToSave.add(ns);
+    set(encryptionKeys, `${ns}.${keyPath}`, undefined);
+  }
+
+  for (const ns of nsToSave) {
+    await save(ns);
+  }
```
mcayuelas-ledger, line 152: Why using 2 loop?

apps/ledger-live-desktop/src/main/db/index.ts
```diff
@@ -86,41 +86,70 @@ async function reload() {
...
+          decrypted = decrypted.data;
+          break;
+        }
+      }
+
+      // only set decrypted data in memory
+      set(memoryNamespaces[ns]!, keyPath, decrypted);
+    } catch (err) {
+      log("db", "setEncryptionKey failure: " + String(err));
+      console.error(err);
```
mcayuelas-ledger, line 134: 🚮

mcayuelas-ledger (changes requested):

apps/ledger-live-desktop/src/main/db/index.ts
gre: because if we do it in the first loop, we will `save()` multiple time on the same `ns` which would be performance regression in this context.

apps/ledger-live-desktop/src/main/db/index.ts
gre: i think `console.error` are legit for unexpected cases, they log more details than the log. (nb: i don't think i introduced it in this PR)

apps/ledger-live-desktop/src/renderer/storage.ts
gre: not sure, it will conflict with our `Trustchain` type

apps/ledger-live-desktop/src/renderer/middlewares/db.ts
gre: bien vu 😱

mcayuelas-ledger (approved):

gre:
seems like test are failing

PR #7298, opened 2024-07-09
chore(walletsync): introduce hooks to be able to restoreTrustchain

gre:
### ✅ Checklist

- [x] **Covered by automatic tests.** unit tests only, manual test on web tools 
  - trustchain and live-wallet (wallet sync)

### 📝 Description

DEMO

https://github.com/LedgerHQ/ledger-live/assets/211411/73a60c95-1ff4-4f8e-8718-f1eec8f99daa

This implement the automatic management of Trustchain update and Cloud Sync data updates when a `removeMember` occurs which rotates the trustchain (= change the encryption keys).

- **trustchain**: introduce `TrustchainLifecycle` that can be given to TrustchainSDK in order to hook to specific event.
  - `onTrustchainRotation: (TrustchainSDK, Trustchain, MemberCredentials) => (Trustchain) => void` is the first hook we need in order to properly be able to remove cloud sync data with the jwt token taken before the rotation and push the new encryption with the new jwt token after the rotation. (i have simplified the signature in this description, but it's actually all async. the first function level is called before the rotation, the second is called after)
- **live-wallet/walletsync**: `trustchainLifecycle` implements the lifecycle in context of wallet sync.
- **live-wallet/cloudsync**
  - the encryption/compression implementation of cloudsync sdk is moved into `cipher.ts` in order to be reused by the implementation of `trustchainLifecycle`.
  - `onTrustchainRefreshNeeded` is introduced to have opportunity for client to call sdk.restoreTrustchain
- [x] web tools is adapted to integrate the new evolutions.

### ❓ Context

- **JIRA or GitHub link**: live-12973

---


libs/live-wallet/src/walletsync/__tests__/trustchainLifecyle.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  let mswServer: ReturnType<typeof setupServer>;
+  beforeAll(() => {
+    mswServer = setupServer(...handlers);
+    mswServer.listen();
+  });
+
+  afterAll(() => {
+    mswServer.close();
+  });
```
mcayuelas-ledger, line 48: You can move all of that outside in a specific file, maybe in the future we will have to do same thing on another files? 

wdyt?

libs/live-wallet/src/walletsync/trustchainLifecyle.ts
```diff
@@ -0,0 +1,41 @@
...
+        );
+        // we then need to push back data to a new CloudSync id with the new encryption key
+        const { version, data } = getCurrentWSState();
+        if (!data) return;
+        const cipher = makeCipher(trustchainSdk);
+        const payload = await cipher.encrypt(newTrustchain, data);
+        await api.uploadData(newJwt, liveSlug, version, payload);
+      };
+    },
+  };
```
mcayuelas-ledger, line 40: IMHO it will be more readeble to do something like that, if we add more functions 

```suggestion
 async function onTrustchainRotation(trustchainSdk, oldTrustchain, memberCredentials) {
    const oldJwt = await trustchainSdk.withAuth(
      oldTrustchain,
      memberCredentials,
      jwt => Promise.resolve(jwt),
      "refresh",
    );
    return async newTrustchain => {
      // when trustchain rotates, we need to delete old data to inform members still on the old id
      await api.deleteData(oldJwt, liveSlug);
      const newJwt = await trustchainSdk.withAuth(
        newTrustchain,
        memberCredentials,
        jwt => Promise.resolve(jwt),
        "refresh",
      );
      // we then need to push back data to a new CloudSync id with the new encryption key
      const { version, data } = getCurrentWSState();
      if (!data) return;
      const cipher = makeCipher(trustchainSdk);
      const payload = await cipher.encrypt(newTrustchain, data);
      await api.uploadData(newJwt, liveSlug, version, payload);
    };
  }

  return {
    onTrustchainRotation,
  };
```

libs/live-wallet/src/cloudsync/cipher.ts
```diff
@@ -0,0 +1,38 @@
...
+export type Cipher<Data> = {
+  decrypt: (trustchain: Trustchain, base64payload: string) => Promise<Data>;
+  encrypt: (trustchain: Trustchain, data: Data) => Promise<string>;
+};
+
+/**
+ * Create a cipher that can encrypt/decrypt data using a trustchain for cloudsync
+ */
+export function makeCipher<Data>(trustchainSdk: TrustchainSDK): Cipher<Data> {
+  return {
```
mcayuelas-ledger, line 14: same comment for readability :)

libs/trustchain/src/sdk.ts
```diff
@@ -203,6 +206,9 @@ export class SDK implements TrustchainSDK {
       memberCredentials.pubkey !== member.id,
       "removeMember must not be used to remove the current member.",
     );
+
+    const hook = await this.lifecycle?.onTrustchainRotation(this, trustchain, memberCredentials);
```
mcayuelas-ledger, line 210: can we find a better naming? 
 `rotation` will be more suitable I think

I mean because you use a `if` and by definition Hooks can be used like that

libs/live-wallet/src/walletsync/__tests__/trustchainLifecyle.test.ts
gre: mmh good point but i think here the code is way too test specific for a generalisation at the moment.
things like

`if (authHeader !== "Bearer new-jwt") {`

are extremely specific for the trustchain rotation test.

libs/live-wallet/src/walletsync/trustchainLifecyle.ts
gre: this was my original code, but it doesn't pass typescript.

TypeScript can't infer the signature of function onTrustchainRotation in this code version, i think it's better not to have to explicit the type therefore i moved to the former code style

libs/trustchain/src/sdk.ts
gre: ok i will probably rename to `afterRotation` then 👍

mcayuelas-ledger (approved):

PR #7300, opened 2024-07-09
LLD integration of Wallet Sync

gre:
### ✅ Checklist

  - LLD when trustchain is active

### 📝 Description

**This integrates Wallet Sync feature on LLD.**

- `walletSyncWatchLoop` is integrated in LLD.
- error automatic handling (removing someone in the members, it will detects that it needs to restoreTrustchain, if member is yourself it will also automatically ejects)
- UI: error and pending state are displayed on the global **✅  Synchronized** top activity banner.

### ❓ Context

- **JIRA or GitHub link**: live-12636

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/init.tsx
+++ b/apps/ledger-live-desktop/src/renderer/init.tsx
@@ -47,8 +47,9 @@ import { addDevice, removeDevice, resetDevices } from "~/renderer/actions/device
 import { Device } from "@ledgerhq/live-common/hw/actions/types";
 import { listCachedCurrencyIds } from "./bridge/cache";
 import { LogEntry } from "winston";
-import { importTrustchainStoreState } from "@ledgerhq/trustchain/store";
 import { importMarketState } from "./actions/market";
+import { fetchWallet } from "./actions/wallet";
+import { fetchTrustchain } from "./actions/trustchain";
 
 const rootNode = document.getElementById("react-root");
 const TAB_KEY = 9;
@@ -180,10 +181,9 @@ async function init() {
       }),
     );
   }
-  const trustchainStoreState = await getKey("app", "trustchain");
-  if (trustchainStoreState) {
-    store.dispatch(importTrustchainStoreState(trustchainStoreState));
-  }
+
+  await fetchTrustchain()(store.dispatch);
+  await fetchWallet()(store.dispatch);
 
   const marketState = await getKey("app", "market");
   if (marketState) {
--- a/apps/ledger-live-desktop/src/main/db/index.ts
+++ b/apps/ledger-live-desktop/src/main/db/index.ts
@@ -92,6 +92,7 @@ async function reload() {
 const encryptedDataPaths = [
   ["app", "accounts"],
   ["app", "trustchain"],
+  ["app", "wallet"],
 ];
 
 /**
--- a/apps/ledger-live-desktop/src/newArch/features/WalletSync/components/WalletSyncContext.tsx
+++ b/apps/ledger-live-desktop/src/newArch/features/WalletSync/components/WalletSyncContext.tsx
@@ -1,0 +1,17 @@
+import React from "react";
+import { useWatchWalletSync, WalletSyncUserState } from "../hooks/useWatchWalletSync";
+
+export const WalletSyncContext = React.createContext<WalletSyncUserState>({
+  visualPending: false,
+  walletSyncError: null,
+  onUserRefresh: () => {},
+});
+
+export const useWalletSyncUserState = () => React.useContext(WalletSyncContext);
+
+export function WalletSyncProvider({ children }: { children: React.ReactNode }) {
+  const walletSyncState = useWatchWalletSync();
+  return (
+    <WalletSyncContext.Provider value={walletSyncState}>{children}</WalletSyncContext.Provider>
+  );
+}
--- a/apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useDestroyTrustchain.ts
+++ b/apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useDestroyTrustchain.ts
@@ -6,24 +6,32 @@ import {
   memberCredentialsSelector,
 } from "@ledgerhq/trustchain/store";
 import { useMutation } from "@tanstack/react-query";
-import { MemberCredentials, Trustchain } from "@ledgerhq/trustchain/types";
 import { setFlow } from "~/renderer/actions/walletSync";
 import { Flow, Step } from "~/renderer/reducers/walletSync";
 import { QueryKey } from "./type.hooks";
+import { useCloudSyncSDK } from "./useWatchWalletSync";
+import { walletSyncUpdate } from "@ledgerhq/live-wallet/store";
 
 export function useDestroyTrustchain() {
   const dispatch = useDispatch();
   const sdk = useTrustchainSdk();
+  const cloudSyncSDK = useCloudSyncSDK();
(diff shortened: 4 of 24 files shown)


apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
```diff
@@ -0,0 +1,191 @@
...
+  const getCurrentVersion = useCallback(() => {
+    const state = walletSelector(store.getState());
+    return state.wsState.version;
+  }, [store]);
+
+  const saveNewUpdate = useCallback(
+    async (event: UpdateEvent<DistantState>) => {
+      log("walletsync", "saveNewUpdate", { event });
+      switch (event.type) {
+        case "new-data": {
```
mcayuelas-ledger, line 63: can we type `event.type` ?

.changeset/nasty-windows-suffer.md
mcayuelas-ledger, line 1: Not the purpose of this PR I think ^^

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
```diff
@@ -0,0 +1,191 @@
...
+          const state = store.getState();
+          const version = event.version;
+          const data = event.data;
+          const wsState = walletSelector(state).wsState;
```
mcayuelas-ledger, line 67: `const wsState = walletSelector(state).wsState;` we can create a specific selector for that instead of getting all the store?

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
gre: mmh it is typed (as a union of static strings), if you do `case "foo": `  it will produce an error.

Capture d’écran 2024-07-15 à 16 22 55

.changeset/nasty-windows-suffer.md
gre: will disappear after we merge the other PR and rebase

gre: ( if we even merge it. otherwise we'll undo, i wanted to test the PR with it for now )

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
cgrellard-ledger, line 1: Shouldn't we split this in multiple files ?

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useTrustchainSdk.ts
```diff
@@ -5,6 +5,10 @@ import { getEnv } from "@ledgerhq/live-env";
 import { getSdk } from "@ledgerhq/trustchain/index";
 import { withDevice } from "@ledgerhq/live-common/hw/deviceAccess";
 import Transport from "@ledgerhq/hw-transport";
+import { trustchainLifecycle } from "@ledgerhq/live-wallet/walletsync/index";
+import { useStore } from "react-redux";
+import { walletSelector } from "~/renderer/reducers/wallet";
+import { wsStateSelector } from "@ledgerhq/live-wallet/store";
```
cgrellard-ledger, line 11: I feel like we should avoid abbreviations as much as possible for clarity reasons especially when it's in a package that isn't specific to Wallet Sync

apps/ledger-live-desktop/src/renderer/init.tsx
```diff
@@ -47,8 +47,9 @@ import { addDevice, removeDevice, resetDevices } from "~/renderer/actions/device
 import { Device } from "@ledgerhq/live-common/hw/actions/types";
 import { listCachedCurrencyIds } from "./bridge/cache";
 import { LogEntry } from "winston";
-import { importTrustchainStoreState } from "@ledgerhq/trustchain/store";
 import { importMarketState } from "./actions/market";
+import { fetchWallet } from "./actions/wallet";
+import { fetchTrustchain } from "./reducers/trustchain";
```
cgrellard-ledger, line 52: Is there a reason for the `fetchWallet` to be in an `actions` file and the `fetchTrustchain` to be in a `reducers` one ?

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useTrustchainSdk.ts
gre: `wsStateSelector` because it gets the `wsState: WSState` field of the store. we could globally rename it. what would you suggest then? basically it's a object with `{ data : the latest distance state , version : the version associated to it }`

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
gre: what is important is that the "read" is a mirror of the "write" parts, and typically when we will add more features like a new module on the reconciliation, you would have to both edit the read and the write, which is why i found it interesting to keep things a bit collocated.
but in any case, we are still secured by the types (at least for the read parts, for the write you just have to not forget to save the data with a dispatch)

apps/ledger-live-desktop/src/renderer/init.tsx
gre: fetchTrustchain should probably move to a `actions/trustchain`, this file just don't exist yet.

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useTrustchainSdk.ts
cgrellard-ledger: We could at least rename those to `walletSyncState` and `walletSyncStateSelector` to understand immediately that they are used for this feature

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
cgrellard-ledger: I would have done one file per hook that is exported but I don't know if it's overkill or not. In any case I find it weird that there are some hooks related to cloud sync and others related to trustchain in the same file

gre: it's all related to wallet sync, but it does indeed the glue with the trustchain. only two hooks are exported (and needs to be) all other are internal usage of the wallet sync loop.
i could only see we can split `useCloudSyncSDK` out, that means we move the "write" parts away from the "read" and that we need to be careful when adding a new module that we will need to modify both parts

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useTrustchainSdk.ts
gre: or maybe it's actually `cloudSyncState` 👀  except it's the one specific to `live` schema.

apps/ledger-live-desktop/src/newArch/WalletSync/hooks/useWatchWalletSync.ts
cgrellard-ledger: Ok thanks for the explanations, if it makes sense let's keep it that way then

cgrellard-ledger (approved):

mcayuelas-ledger (approved):

PR #7328, opened 2024-07-12
[LIVE-13128][LLD][Passkey] Create use case for saving app data inside local storage

jiyuzhuang:
### ✅ Checklist

  - Add backup App data use case that can be used to backup the App data from Ledger device.

### 📝 Description

- Add use case for backing up App data;
- `backupAppData` used to organise all APDU sending issues;
- `backupAppDataUseCase` used to data persistence operations;
- `backupAppDataUseCaseDI` used for DI implementation;
- Errors instantiated as `BackupAppDataError`;
- Desktop storage provider implemented using `global.localStorage`.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-13128]

---

[LIVE-13128]: https://ledgerhq.atlassian.net/browse/LIVE-13128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/device-core/src/commands/entities/AppStorageInfo.ts
```diff
@@ -5,3 +5,20 @@ export type AppStorageInfo = {
...
+    "dataVersion" in data &&
+    typeof data.dataVersion === "string" &&
+    "hasSettings" in data &&
+    typeof data.hasSettings === "boolean" &&
+    "hasData" in data &&
+    typeof data.hasData === "boolean" &&
+    "hash" in data &&
+    typeof data.hash === "string"
+  );
+}
```
ofreyssinet-ledger, line 24: nice

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface StorageProvider<VType> {
+  getItem(key: string): Promise<VType | null>;
+  setItem(key: string, value: VType): Promise<void>;
+  removeItem(key: string): Promise<void>;
+}
+
+export type AppName = string;
+export type AppStorageKey =
+  `${Exclude<DeviceModelId, DeviceModelId.blue | DeviceModelId.nanoS>}-${AppName}`;
```
ofreyssinet-ledger, line 13: [Ask] is it really necessary to exclude the blue & nanoS here?

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
```diff
@@ -0,0 +1,59 @@
...
+  AppDataInfoFetched = "appDataInfoFetched",
+  AppDataBackedUp = "appDataBackedUp",
+  AppDataAlreadyBackedUp = "appDataAlreadyBackedUp",
+  NoAppDataToBackup = "noAppDataToBackup",
+}
+
+export type BackupAppDataEvent =
+  | {
+      type: BackupAppDataEventType.Progress;
+      data: number; // Progress percentage
```
ofreyssinet-ledger, line 42: [SHOULD] Use TSDoc for those comments.

libs/ledger-live-common/src/device/use-cases/appDataBackup/DesktopStorageProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+   * @param key - The key to retrieve the value for, type of AppStorageKey.
+   * @returns A promise that resolves to the value associated with the key, or null if the key does not exist.
+   * @throws Error if the retrieved data is not of type AppStorageType.
+   */
+  getItem(key: AppStorageKey): Promise<AppStorageType | null> {
+    const data: string | null = global.localStorage.getItem(key);
+    if (!data) {
+      return Promise.resolve(null);
+    }
+    const parsedData = JSON.parse(data);
```
ofreyssinet-ledger, line 26: [COULD] try/catch and throw a `BackupAppDataError`

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
```diff
@@ -0,0 +1,59 @@
...
+export function appStorageTypeGuard(data: AppStorageType | unknown): data is AppStorageType {
+  return (
+    typeof data === "object" &&
+    data !== null &&
+    "appDataInfo" in data &&
+    appStorageInfoGuard(data.appDataInfo) &&
+    "appData" in data &&
+    typeof data.appData === "string"
+  );
+}
```
ofreyssinet-ledger, line 29: [SHOULD] unit test all the type guards

libs/ledger-live-common/src/device/use-cases/appDataBackup/DesktopStorageProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+  AppStorageKey,
+  appStorageTypeGuard,
+  BackupAppDataError,
+} from "./types";
+
+/**
+ * The storage provider for LLD.
+ * Implements the StorageProvider interface.
+ */
+export class DesktopStorageProvider implements StorageProvider<AppStorageType> {
```
jdabbech-ledger, line 13: [ask] why do you introduce this class with the usage of `global.localStorage` in `ledger-live-common` instead of `ledger-live-desktop`? This value isn't defined in `ledger-live-mobile`

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
ofreyssinet-ledger: Also I would rather name this key `progress`, and the one below for `AppDataInfoFetched`, `appStorageInfo`.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppData.ts
```diff
@@ -0,0 +1,69 @@
...
+          subscriber.complete();
+        }),
+      )
+      .subscribe(subscriber);
+
+    return () => {
+      sub.unsubscribe();
+    };
+  });
+  return obs as Observable<BackupAppDataEvent>;
```
ofreyssinet-ledger, line 68: [SHOULD] not use `as`, you can just do `const obs = new Observable<BackupAppDataEvent>(...)`

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
jiyuzhuang: I've forgotten why they are excluded, but if they are support to backup data, I think I need to add them back. I will check.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppData.ts
jiyuzhuang: It does not work, after some tries, this is the single solution in this case.

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
ofreyssinet-ledger: yes but even if they are not supported I'm not sure it's necessary to narrow the type of the storage key 🤷‍♂️

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppData.ts
ofreyssinet-ledger: I'm pretty sure there is a way to avoid using `as`, there must be a mistake somewhere if `new Observable<BackupAppDataEvent>(...)`.

I think the `subscribe(subscriber)` line 61 is what is incorrect because that observable you're subscribing to just returns `void` so your Observable is in fact an `Observable<BackupAppDataEvent | void>`.

You can just replace line 62 by `.subscribe()` and it works:

Screenshot 2024-07-18 at 15 23 18

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
jiyuzhuang: This aimes to have the same format for event `{ type: Enum, data: DataType }` and it's easy to maintain and treat an event without knowing the specific property name of each event type. When someone wants to treat an event of type progress, he/she must know the event type is `BackupAppDataEventType.Progress`, so the data cannot be something else. In terms of security, the event type defines strictly the data type.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppData.ts
ofreyssinet-ledger: By the way by forcing with `as` and not removing the subscribe(subscriber) there was a bug: you were emitting `undefined` in your observable, which might cause a crash in the client as they're assuming that the type is correct.

ofreyssinet-ledger: That's why `as` is really a trigger warning and we should always be very careful when using it. Sometimes there is no choice (but it's rare), and often when we use it, it just hides an underlying bug or at least some bad typing.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppDataUseCase.ts
```diff
@@ -0,0 +1,63 @@
...
+          return event;
+        }
+        case BackupAppDataEventType.AppDataBackedUp:
+          // Store the app data
+          await storageProvider.setItem(`${deviceModelId}-${appName}`, {
+            appDataInfo,
+            appData: event.data,
+          });
+          // Erase the app data, then return the event
+          return { type: BackupAppDataEventType.AppDataBackedUp, data: "" };
```
ofreyssinet-ledger, line 53: [ASK] what's the point of returning `data: ""` ?

jiyuzhuang: Erase data, exposing the encrypted data to end user is not needed.

libs/ledger-live-common/src/device/use-cases/appDataBackup/DesktopStorageProvider.ts
jiyuzhuang: I'll move it to LLD folder. In a mobile context, the use case (`backupAppDataUseCaseDI`) should be initialised by `MobileStorageProvider` with mobile features.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppData.ts
jiyuzhuang: Yes, I've found the solution. 👍

libs/ledger-live-common/src/device/use-cases/appDataBackup/types.ts
ofreyssinet-ledger: Yep yep it's just for readability but what you're saying makes sense too, as you want 👍

libs/device-core/src/commands/entities/AppStorageInfo.ts
```diff
@@ -5,3 +5,20 @@ export type AppStorageInfo = {
   hasData: boolean;
   hash: string;
 };
+
+export function appStorageInfoGuard(data: AppStorageInfo | unknown): data is AppStorageInfo {
```
aussedatlo, line 9: [ASK] IDK about the team sentiment but I would prefer `isAppStorageInfo` to explicitly notite that the function return a boolean.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppDataUseCase.test.ts
```diff
@@ -0,0 +1,158 @@
...
+describe("backupAppDataUseCase", () => {
+  const storageProviderMock = {
+    getItem: jest.fn(),
+    setItem: jest.fn(),
+    removeItem: jest.fn(),
+  };
+  const appName = "MyApp";
+  const deviceModelId = DeviceModelId.stax;
+
+  beforeEach(() => {});
```
aussedatlo, line 21: [COULD] remove this two lines

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppData.ts
```diff
@@ -0,0 +1,69 @@
...
+          }
+
+          // Backup app data by chunks
+          let appData = Buffer.from([]);
+          let offset = 0;
+          while (offset < dataSize) {
+            const chunk: Buffer = await backupAppStorage(transport);
+            // Make sure that the process advances
+            if (chunk.length === 0) {
+              subscriber.error(new BackupAppDataError("Chunk data is empty."));
```
aussedatlo, line 38: [COULD] all other errors on LL doesn't have a dot at the end

aussedatlo (commented):
LGTM overall, just 3 comments about the form, open to discuss 👍

libs/device-core/src/commands/entities/AppStorageInfo.ts
valpinkman: agreed

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppDataUseCase.test.ts
jiyuzhuang: 🫡

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppDataUseCase.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export function backupAppDataUseCase(
+  appName: AppName,
+  deviceModelId: DeviceModelId,
+  storageProvider: StorageProvider<AppStorageType>,
+  backupAppDataFn: () => Observable<BackupAppDataEvent>,
+): Observable<BackupAppDataEvent> {
+  let appDataInfo: AppStorageInfo;
+  const obs: Observable<BackupAppDataEvent> = backupAppDataFn().pipe(
+    switchMap(async (event: BackupAppDataEvent): Promise<BackupAppDataEvent> => {
```
valpinkman, line 31: [ASK] Why is switchMap returning a `Promise<BackupDataEvent>` ? I believe you should return an `Observable<BackupDataEvent>` using `of` or something like that, and probably `throwError` as well instead of just throwing as you are in an observable

libs/device-core/src/commands/entities/AppStorageInfo.ts
jiyuzhuang: Updated.

libs/ledger-live-common/src/device/use-cases/appDataBackup/backupAppDataUseCase.ts
valpinkman: It seems my previous comment was wrong, there is interoperability with promises / async function (although not very well documented). You can disregard the previous comment but i still believe that staying on one paradygm (either Observable  or Promises) is better.

valpinkman (dismissed):

valpinkman (approved):

ofreyssinet-ledger (approved):

KVNLS (approved):

PR #7334, opened 2024-07-15
chore(live-network): only retry network for status code that makes sense

gre:
### ✅ Checklist

  - live-network

### 📝 Description

This is a generalisation of existing code to explicit the list of statusCodes that are allowed to be retried.

this avoid our stack to do 3 failing http code in a row for http code that will always fails anyway, for instance a Unauthorized case (where you fail authentication). 
**in live-hub context, this is a problem for Wallet Sync feature:**
Capture d’écran 2024-07-15 à 14 50 22

the committed statusCodes list is taken from https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#retry

### ❓ Context

- **JIRA or GitHub link**: na

---

### DIFF
--- a/libs/live-network/src/network.ts
+++ b/libs/live-network/src/network.ts
@@ -175,6 +175,10 @@ export type LiveNetworkResponse<T> = {
   data: T;
   status: number;
 };
+
+// inspired by https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#retry
+const retryableHttpStatusCodes = [408, 413, 429, 500, 502, 503, 504, 521, 522, 524];
+
 /**
  * Network call
  * @param request
@@ -198,9 +202,8 @@ export const newImplementation = async <T = unknown, U = unknown>(
       maxRetry: getEnv("GET_CALLS_RETRY"),
       retryCondition: error => {
         if (error && error.status) {
-          // A 422 shouldn't be retried without change as explained in this documentation
-          // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
-          return error.status !== 422;
+          // not all status codes are retryable
+          return retryableHttpStatusCodes.includes(error.status);
         }
         return true;
       },
--- a/.changeset/nasty-windows-suffer.md
+++ b/.changeset/nasty-windows-suffer.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-network": minor
+---
+
+live-network no longer retry network calls on HTTP status code that shouldn't be retried like 401 & some others.


valpinkman (dismissed):

mcayuelas-ledger (dismissed):

Justkant (approved):

libs/live-network/src/network.ts
```diff
@@ -175,6 +175,10 @@ export type LiveNetworkResponse<T> = {
   data: T;
   status: number;
 };
+
+// inspired by https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#retry
+const retryableHttpStatusCodes = [408, 413, 429, 500, 502, 503, 504, 521, 522, 524];
```
sprohaszka-ledger, line 180: I find weird that we retry on server errors. Maybe "internal server" error, but "gateway" issue seems difficult to overcome.
Same for 413. If the payload is too huge, I am not sure it is a good idea to retry.

libs/live-network/src/network.ts
```diff
@@ -198,9 +202,8 @@ export const newImplementation = async <T = unknown, U = unknown>(
       maxRetry: getEnv("GET_CALLS_RETRY"),
       retryCondition: error => {
         if (error && error.status) {
-          // A 422 shouldn't be retried without change as explained in this documentation
-          // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
-          return error.status !== 422;
+          // not all status codes are retryable
+          return retryableHttpStatusCodes.includes(error.status);
```
sprohaszka-ledger, line 206: Be careful, everyone has not migrate to this call.
You might need to duplicate the code with `implementation` function.

sprohaszka-ledger (commented):
Not sure if we can rely on all those return codes.

libs/live-network/src/network.ts
gre: i just have removed a lot of statusCodes, today's behavior is that they are ALL retried, so it's already a progress
but i agree with you tho, an ultimately target should to not even have this retry behavior on the network stack side imo

libs/live-network/src/network.ts
gre: in my context i only need it fixed for `live-wallet`'s wallet sync code so at least it fixes this.
the benefits being that it limits the impact of the PR changes. only the new implementation users will benefit the evolution and i didn't break the behavior of the deprecated way.
when all users of live-network migrate to newImplementation they will then have the opportunity to verify this change is working 🤔

PR #7346, opened 2024-07-16
fix(llm): remove global.crypto in LLM because it's broken

gre:
### ✅ Checklist

  - LLM `global.crypto` value is removed to prevent feature detection to think it's a correct Crypto implementation.

### 📝 Description

In React Native, the global `crypto` object is only partially implemented. It includes `crypto.getRandomBytes` but lacks other essential functionalities like `crypto.subtle.digest`. This causes issues with libraries such as `@node/secp256k1`, which assume a complete implementation of `crypto` if `global.crypto` is defined.

To prevent `@node/secp256k1` from incorrectly assuming that a complete `crypto` implementation is available, we should delete `global.crypto` in the React Native environment when only `crypto.getRandomBytes` is present. This will ensure that the library does not attempt to use the incomplete `crypto` object.

---

a normal Web implementation of crypto:
Capture d’écran 2024-07-16 à 15 55 39

React Native one's:

```
{"getRandomValues": [Function getRandomValues]}
```

### ❓ Context

- **JIRA or GitHub link**: in context of LLM integration of Wallet Sync

---

### DIFF
--- a/apps/ledger-live-mobile/src/polyfill.ts
+++ b/apps/ledger-live-mobile/src/polyfill.ts
@@ -76,3 +76,9 @@ if (__DEV__ && process.env.NODE_ENV !== "test") {
     }
   }, 100);
 }
+
+// the current global.crypto implementation isn't complete and feature detecting it create bad bugs,
+// notably in the library `@node/secp256k1` that will try to use `global.crypto.subtle.digest` when it finds global.crypto to be available.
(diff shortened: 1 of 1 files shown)


gre:
unfortunately way more complex than this https://ledgerhq.atlassian.net/browse/LIVE-13417

PR #7366, opened 2024-07-18
Fixed data tracking for Braze cards

cgrellard-ledger:
### ✅ Checklist

  - ...

### 📝 Description

The action cards on LLD were not registering the clicks and the flexible content cards on LLM always had the unique recipients at 0.

For LLD, the url property of the card had to be set for Braze to register the click (we set it to the card's id just so the value is not null).
For LLM, we had to log the impressions of the flexible cards so Braze could calculate the unique recipients.

Screenshot 2024-07-18 at 18 33 27
Screenshot 2024-07-18 at 18 33 40

### ❓ Context

- **JIRA or GitHub link**: [LIVE-12815] [LIVE-13258]

---

[LIVE-12815]: https://ledgerhq.atlassian.net/browse/LIVE-12815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/hooks/useActionCards.tsx
+++ b/apps/ledger-live-desktop/src/renderer/hooks/useActionCards.tsx
@@ -54,7 +54,10 @@ const useActionCards = () => {
     const actionCard = findActionCard(cardId);
 
     if (currentCard) {
-      isTrackedUser && braze.logContentCardClick(currentCard);
+      // For some reason braze won't log the click event if the card url is empty
+      // Setting it as the card id just to have a dummy non empty value
+      isTrackedUser &&
+        braze.logContentCardClick({ ...currentCard, url: currentCard.id } as braze.ClassicCard);
       link && openURL(link);
     }
     if (actionCard) {
--- a/apps/ledger-live-mobile/src/dynamicContent/ContentCardsCategory/Layout.tsx
+++ b/apps/ledger-live-mobile/src/dynamicContent/ContentCardsCategory/Layout.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { Linking } from "react-native";
 import HorizontalCard from "../../contentCards/cards/horizontal";
 import {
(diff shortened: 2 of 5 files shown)


apps/ledger-live-mobile/src/dynamicContent/ContentCardsCategory/Layout.tsx
```diff
@@ -67,8 +67,16 @@ type LayoutProps = {
...
-  const { logClickCard, dismissCard, trackContentCardEvent } = useDynamicContent();
+  const { logClickCard, dismissCard, trackContentCardEvent, logImpressionCard } =
+    useDynamicContent();
+
+  useEffect(() => {
+    logImpressionCard(category.id);
+    for (const card of cards) {
+      logImpressionCard(card.id);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
KVNLS, line 78: Please do not "lie" to React: https://react.dev/learn/removing-effect-dependencies#why-is-suppressing-the-dependency-linter-so-dangerous

apps/ledger-live-desktop/src/renderer/hooks/useActionCards.tsx
```diff
@@ -54,6 +54,11 @@ const useActionCards = () => {
     const actionCard = findActionCard(cardId);
 
     if (currentCard) {
+      // For some reason braze won't log the click event if the card url is empty
+      // Setting it as the card id just to have a dummy non empty value
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      currentCard.url = currentCard.id;
```
KVNLS, line 61: Is it not dangerous to mutate the `currentCard` URL value? Don't we want to do it only for `logContentCardClick` function?

KVNLS (changes requested):

apps/ledger-live-mobile/src/dynamicContent/ContentCardsCategory/Layout.tsx
cgrellard-ledger: Looking at the doc, while useEffectEvent isn't stable yet, I don't see how I can run a function at the component mount without having to mute the linter 🤔

apps/ledger-live-desktop/src/renderer/hooks/useActionCards.tsx
cgrellard-ledger: Technically this property is only used for "classic" content cards but as we only leverage the custom design ones this property is never set but ideally yes we shouldn't mutate an object like this. I pushed a modification :)

KVNLS (approved):

PR #7376, opened 2024-07-19
chore(walletsync): introduce nonImportedAccountInfos management

gre:
### ✅ Checklist

  - wallet sync management of account sync failures

### 📝 Description

This improve the Wallet Sync reconciliation in order to **keep a memory of non imported account ids to be able to eventually restore them.**

this can be tested by limiting the number of supported currencies on the web tools: (but you can also import tezos/cosmos which currently also fails on the web)

Capture d’écran 2024-07-22 à 17 25 05

**The previous behaviour:**

non imported accounts management must be done on accounts that could possibly fail the synchronisation (during the resolve). At the moment they become completely ignored and will never appear again on other members. (fail sync = the user had network issues / the backend had issues / the Ledger Live instance don't yet support the coin). worse: in current implementation, account disappear when another member don’t handle such account. this is reproductible today on web tools that don’t support Cosmos nor Tezos (due to CORS issues) and you can easily see that LLD’s account disappear after web tools fetch the incoming update and then determines there is a diff to push since the unsupported accounts are not imported. and this could happen in case of tmp backend issue too, so in short: account can randomly disappear.

**The next behaviour:**

Accounts should be preserved on the instances that don’t support them yet and eventually be imported as soon as they do. let’s consider member A that supports everything and member B that don’t support all accounts.

- when member A push its accounts, member B needs to keep a memory of them, possibly ignore them but make sure to push them back as part of the cloud sync data in order for member A do not delete them.
- when member B eventually support accounts, they will appear after a new resolve
- when member A remove accounts, their deletion need to be propagated in the memory that member B, so when member B do a new edition, member A should not see again the account they just removed.

#### Technical notes

- `resolveIncomingDistantState` is renamed to `resolveIncrementalUpdate` because it serves a bigger purpose: it is no longer just about resolving incoming distant state, but we will also have to regularly call `resolveIncrementalUpdate` in order to search for possible updates.
- the retry strategy that was implemented is to slow down the retries over time to not spam the backends if there is a sync issue that will last for long. it will only retry every X minutes. this is configurable.

#### To schedule in future (JIRA)

- consider a different lifecycle on the wallet sync watch loop to separate the different loops (pull VS local resolve VS check for pushable). only the "pull polling" is something that induce costs on backend, the other are local loops that do nothing if no changes.
- add more unit tests that cover the `nonImportedAccountInfos` management

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-13312

---

### DIFF
--- a/apps/web-tools/trustchain/components/AppAccountsSync.tsx
+++ b/apps/web-tools/trustchain/components/AppAccountsSync.tsx
@@ -1,11 +1,11 @@
 import React, { FormEvent, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Observable, concat, find, from, ignoreElements, mergeMap, tap } from "rxjs";
-import { JWT, MemberCredentials, Trustchain } from "@ledgerhq/trustchain/types";
+import { MemberCredentials, Trustchain } from "@ledgerhq/trustchain/types";
 import { useTrustchainSDK } from "../context";
 import { CloudSyncSDK } from "@ledgerhq/live-wallet/cloudsync/index";
 import {
   WalletState,
-  handlers as walletH,
+  handlers as walletHandlers,
   accountNameWithDefaultSelector,
   setAccountName as setAccountNameAction,
   WSState,
@@ -32,19 +32,13 @@ import { listSupportedCurrencies } from "@ledgerhq/coin-framework/lib-es/currenc
 import { getCurrencyColor } from "@ledgerhq/live-common/currencies/color";
 import { Loading } from "./Loading";
 import { TrustchainEjected } from "@ledgerhq/trustchain/lib-es/errors";
-import { NonImportedAccountInfo } from "@ledgerhq/live-wallet/lib-es/walletsync/modules/accounts";
 import { Tick } from "./Tick";
+import { State } from "./types";
 
 /*
 import * as icons from "@ledgerhq/crypto-icons-ui/react";
 import { inferCryptoCurrencyIcon } from "@ledgerhq/live-common/currencies/cryptoIcons";
 */
-
-type State = {
-  accounts: Account[];
-  nonImportedAccounts: NonImportedAccountInfo[];
-  walletState: WalletState;
-};
 
 const latestWalletStateSelector = (s: State): WSState => walletSyncStateSelector(s.walletState);
 
@@ -114,12 +108,15 @@ export default function AppAccountsSync({
         // we now need to "reverse" the localStateSelector back into our own internal state
         let walletState = s.walletState;
         if (newLocalState) {
-          walletState = walletH.BULK_SET_ACCOUNT_NAMES(
+          walletState = walletHandlers.BULK_SET_ACCOUNT_NAMES(
             walletState,
             setAccountNames(newLocalState.accountNames),
           );
         }
-        walletState = walletH.WALLET_SYNC_UPDATE(walletState, walletSyncUpdate(data, version));
+        walletState = walletHandlers.WALLET_SYNC_UPDATE(
+          walletState,
+          walletSyncUpdate(data, version),
+        );
         if (newLocalState) {
           return {
             accounts: newLocalState.accounts.list, // save new accounts
@@ -230,7 +227,7 @@ export default function AppAccountsSync({
     (id: string, name: string) => {
       setState(s => ({
         ...s,
-        walletState: walletH.SET_ACCOUNT_NAME(s.walletState, setAccountNameAction(id, name)),
+        walletState: walletHandlers.SET_ACCOUNT_NAME(s.walletState, setAccountNameAction(id, name)),
       }));
     },
     [setState],
--- a/apps/web-tools/trustchain/components/AppAccountsSync.tsx
+++ b/apps/web-tools/trustchain/components/AppAccountsSync.tsx
@@ -2,7 +2,7 @@ import React, { FormEvent, useCallback, useEffect, useMemo, useRef, useState } f
 import { Observable, concat, find, from, ignoreElements, mergeMap, tap } from "rxjs";
 import { JWT, MemberCredentials, Trustchain } from "@ledgerhq/trustchain/types";
 import { useTrustchainSDK } from "../context";
-import { CloudSyncSDK, UpdateEvent } from "@ledgerhq/live-wallet/cloudsync/index";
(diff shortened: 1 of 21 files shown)


libs/live-wallet/src/walletsync/incrementalUpdates.ts
```diff
@@ -0,0 +1,89 @@
...
+}): () => Promise<void> {
+  return async () => {
+    // we resolve possible local incremental update
+    const state = getState();
+    const { data, version } = latestWalletStateSelector(state);
+    const local = localStateSelector(state);
+    const resolved = await walletsync.resolveIncrementalUpdate(ctx, local, data, data);
+
+    if (resolved.hasChanges) {
+      const localState = localStateSelector(getState()); // fetch again latest state because it might have changed
```
KVNLS, line 83: Is it not risky to rely on new values of the state? Because here we could have `resolved` that would be computed on a previous version of the state and applied in applyUpdate with a different version of the state.

apps/web-tools/trustchain/components/AppAccountsSync.tsx
```diff
@@ -226,13 +236,29 @@ export default function AppAccountsSync({
...
     <div>
+      {error ? (
+        <div style={{ padding: 10, color: "red" }}>{error.message}</div>
+      ) : timestamp ? (
+        <div
+          style={{
+            textAlign: "center",
+          }}
+        >
+          Synced <Tick timestamp={timestamp} />.
```
KVNLS, line 247: Could you provide translations, please?

apps/web-tools/trustchain/components/Tick.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useState } from "react";
+
+export function Tick({ timestamp }: { timestamp: number }) {
```
KVNLS, line 3: We should handle translations for this component

libs/live-wallet/src/walletsync/incrementalUpdates.ts
gre: @KVNLS on the generic part (of that code you point out) everything is fine, as long at the module have an "incremental" implementation of an "update". so it all depends on how the module is implemented:
- for `accounts`, i think we are safe because we express an "update" as a really incremental thing, instead of saying "here is the new accounts" in a resolve, we say "here is the added, here is the deleted", and we apply on top of latest state these "actions".
- for `accountNames`, **in the current naive implementation** there is a risk of overriding the account names: if you modify the account names on both sides at the same time, one will win over the other, because the module is currently very naive: it replaces all the names it receives without taking care of the one modified. => i'll write a ticket on this one for future improvement. IMO it's not really a big issue because you rarely modify account names at the same time on both sides.

apps/web-tools/trustchain/components/AppAccountsSync.tsx
gre: we don't have translations for web-tools.

apps/web-tools/trustchain/components/Tick.tsx
gre: not for web tools.

apps/web-tools/trustchain/components/AppAccountsSync.tsx
KVNLS: Sorry, I thought we were on the app

libs/live-wallet/src/walletsync/incrementalUpdates.ts
KVNLS: Thanks for the explanation! Do we have unit tests for both use cases?

gre: *also, the answer i just gave is a generic comment on the current implementation. the update of this PR have not introduced or regress on this and it not actually related to the code added here: because the code here is exclusively going to be effective on `accounts` for the case of "pending non imported accounts" to handle, as for the accountNames, it is a noop because the module bail out the case where "latestState === incomingState" in resolveIncrementalUpdate

gre: @KVNLS yes, we have a lot of unit tests around modules. both "generic" (generatively test them on tons of data) and "specific" ones (cover specific usecases and expectations).

apps/web-tools/trustchain/components/AppAccountsSync.tsx
```diff
@@ -36,12 +42,16 @@ import { inferCryptoCurrencyIcon } from "@ledgerhq/live-common/currencies/crypto
 
 type State = {
   accounts: Account[];
+  nonImportedAccounts: NonImportedAccountInfo[];
   walletState: WalletState;
 };
```
mcayuelas-ledger, line 47: You use this type multiple times, you can store it in a file and reuse it everywhere it's needed? :)

apps/web-tools/trustchain/components/AppAccountsSync.tsx
```diff
@@ -92,75 +103,49 @@ export default function AppAccountsSync({
...
-            localState,
-            walletSyncState.data,
-            data,
+  const saveUpdate = useCallback(
+    async (data: DistantState | null, version: number, newLocalState: LocalState | null) => {
+      setState(s => {
+        // we now need to "reverse" the localStateSelector back into our own internal state
+        let walletState = s.walletState;
+        if (newLocalState) {
+          walletState = walletH.BULK_SET_ACCOUNT_NAMES(
```
mcayuelas-ledger, line 117: can you rename walletH ? to be more comprehensive ? :)

libs/live-wallet/src/store.test.ts
```diff
@@ -193,20 +194,24 @@ describe("Wallet store", () => {
   it("can export the wallet state", () => {
     const exportedState = {
       walletSyncState: { data: {}, version: 42 },
+      nonImportedAccountInfos: [],
```
mcayuelas-ledger, line 197: maybe move it outside and then you can do 

expect(exportWalletState(result)).toEqual(exportedState);

apps/web-tools/trustchain/components/AppAccountsSync.tsx
gre: not sure to follow, where is it used? it's the web-tools here, right?

mcayuelas-ledger (approved):

PR #7388, opened 2024-07-22
🧹 [SUPPORT] Factor device interactions out of the Trustchain SDK

thesan:
### ✅ Checklist

  - This refactor doesn't impact the usage of the SDK as provided by `getSdk` function.

### 📝 Description

Factor all device interactions out of the Trustchain SDK into a new `HWDeviceProvider`. This is done in order to help unit testing of the Trustchain SDK later on and to better separate concerns.

In a followup PR: for even better separation of concerns the device transport should not be passed to each SDK method which potentially results in a device interaction. Instead the transport will be passed to the `HWDeviceProvider` constructor. However this requires some breaking changes in the Trustchain SDK so it will be done as a followup.

### ❓ Context

- [LIVE-13211]

---

[LIVE-13211]: https://ledgerhq.atlassian.net/browse/LIVE-13211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/trustchain/tests/scenarios/tokenExpires.ts
```diff
@@ -1,17 +1,18 @@
...
+import { SDK } from "../../src/sdk";
 
-export async function scenario(
-  transport: Transport,
-  { pauseRecorder, sdkForName }: ScenarioOptions,
-) {
-  const sdk = sdkForName("Foo");
+export async function scenario(transport: Transport, { pauseRecorder }: ScenarioOptions) {
+  const hwDeviceProvider = new HWDeviceProvider();
+  const applicationId = 16;
```
KVNLS, line 8: Where does it come from? Can we use a const shared if it's reused somewhere?

libs/trustchain/src/HWDeviceProvider.ts
```diff
@@ -0,0 +1,90 @@
...
+import { AuthCachePolicy, JWT, TrustchainDeviceCallbacks } from "./types";
+
+export class HWDeviceProvider {
+  private policy?: AuthCachePolicy;
+  private jwt?: JWT;
+  private hw?: ApduDevice;
+
+  constructor(policy?: AuthCachePolicy) {
+    this.policy = policy;
+  }
```
KVNLS, line 16: As policy is optional, don't we want to have a default value for it?

libs/trustchain/tests/scenarios/tokenExpires.ts
thesan: It comes from this file when I started working on this refactor: https://github.com/LedgerHQ/ledger-live/blob/bf28296e92939103a6dd94fb17e09fe16ff6ba8b/libs/trustchain/src/test-scenarios/tokenExpires.ts#L6
When I rebased I changed it back because the new `sdkForName` helper doesn't allow setting a custom `HWDeviceProvider`. As it's just a mock value I thought it was fine to leave it as is.

libs/trustchain/src/HWDeviceProvider.ts
gre: also, I think `AuthCachePolicy` is not something we should set at the creation time, it's something that depends on the usecase usually.

for instance, we use a specific AuthCachePolicy on the `CloudSyncSDK#notifications` method to force a refresh (because there, we need to be sure to have a fresh jwt for the listenNotifications):

```ts
  listenNotifications(
    trustchain: Trustchain,
    memberCredentials: MemberCredentials,
  ): Observable<number> {
    const getFreshJwt = (): Promise<JWT> =>
      this.trustchainSdk.withAuth(
        trustchain,
        memberCredentials,
        jwt => Promise.resolve(jwt),
        "refresh",
      );
    return api.listenNotifications(getFreshJwt, this.slug);
  }
  ```

gre (commented):
> for even better separation of concerns the device transport should not be passed to each SDK method which potentially results in a device interaction. Instead the transport will be passed to the HWDeviceProvider constructor. However this requires some breaking changes in the Trustchain SDK so it will be done as a followup.

mmh i don't really see how this can work, a `Transport` is something that is obtained at runtime during the "device actions flow", it's not something that is global but it is instanciated / destroyed each time we do something with a device.

so i think we still have to pass in the Transport in params.

thesan:
> so i think we still have to pass in the Transport in params on the specific method we need it.

Yeah I think I'll leave it how it is right now.

FYI, as I think Mathieu told me that the transport didn't have to be open and closed on every interaction, I wanted to pass an `Observable<Transport>` to the `HWDeviceProvider` this way the device switch would have been seemless and the sdk could still have been created with an empty observable. But then `withDevice` doesn't seem to work for me when the job returns an observable which stays open (I haven't made sure of this but it looks like it waits for the job observable to complete to send it's first value). Plus even if that worked `withDevice` waits for the observable to complete to run the next job so without a major refactor of that function, keeping the transport open won't be possible.

mcayuelas-ledger (commented):
More easy to debug :)

KVNLS (dismissed):
Thanks for this rework!

gre:
Could we have a unit test that shows how this new architecture help writing one? 

Also if we could wait this to gets in first :pray: https://github.com/LedgerHQ/ledger-live/pull/7404

PR #7404, opened 2024-07-23
fix(walletsync): detect more cases of TrustchainOutdated

gre:
### ✅ Checklist

  - hw-trustchain, trustchain, live-wallet

### 📝 Description
- **fixes a bug in `useOnTrustchainRefreshNeeded` implementation**
- **cloudsync API on get / post / delete now need `path` in query param** in order to secure extra checks and prevent a Live still on outdated trustchain to push on new path with wrong encryption key.
  - move from `0h/16h/0h` syntax to `m/0'/16'/0'` syntax to match what server side uses. 
- **trustchain API now returns `permissions` as part of the JWT object.**
  - all `withAuth` are going to check this `permissions` and throw a `TrustchainOutdated` error if it's not matching the `Trustchain` passed in context of the function.
  - that means that methods like `getMembers` can now returns `TrustchainOutdated` and we remap this to `onTrustchainRefreshNeeded` in our integrations.

Capture d’écran 2024-07-23 à 18 40 53

### ❓ Context

- **JIRA or GitHub link**: live-13381

---


mcayuelas-ledger (commented):
Need to be handled in Front also `TrustchainOutdated?`

gre:
@mcayuelas-ledger it's not necessarily a blocker for this PR but i think yes, we need to call `onTrustchainRefreshNeeded` (which can be obtained by `const onTrustchainRefreshNeeded = useOnTrustchainRefreshNeeded(trustchainSdk, memberCredentials);` ) when there is a `TrustchainOutdated` error. 

ideally, it's transparent for users and we can retry again the sdk calls, for instance of it's a getMembers.

we can discuss it tomorrow to see if it's easy to add, but we could do it in another PR i think.

mcayuelas-ledger (approved):

PR #7406, opened 2024-07-24
Fix: LIVE-13301 Set min tezos gaslimit

hzheng-ledger:
### ✅ Checklist

- [X] **Impact of the changes:** add firebase config for tezos coin. 
  - ...

### 📝 Description

Set a min value for gaslitmit for tezos transaction. If the gas simulation fails, this value will be used.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-13301

---

### DIFF
--- a/libs/coin-modules/coin-tezos/src/bridge/prepareTransaction.ts
+++ b/libs/coin-modules/coin-tezos/src/bridge/prepareTransaction.ts
@@ -3,9 +3,14 @@ import { AccountBridge } from "@ledgerhq/types-live";
 import { TezosAccount, Transaction } from "../types";
 import { validateRecipient } from "../logic";
 import getEstimatedFees from "./getFeesForTransaction";
+import coinConfig from "../config";
 
 function bnEq(a: BigNumber | null | undefined, b: BigNumber | null | undefined): boolean {
   return !a && !b ? true : !a || !b ? false : a.eq(b);
+}
+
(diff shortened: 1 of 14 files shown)


libs/coin-modules/coin-tezos/src/bridge/prepareTransaction.ts
```diff
@@ -29,7 +34,18 @@ export const prepareTransaction: AccountBridge<
   }
 
   const estimation = await getEstimatedFees({ account, transaction });
```
hedi-edelbloute, line 36: Shouldn't we replace all values if we see a taquito error here instead ?

hzheng-ledger: In fact, I've had a look at our code.
if the taquito returns an error. (e.g. 500 internal error)

we catches the exception here
https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-tezos/src/logic/estimateFees.ts#L134

and returns the initial value here
https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-tezos/src/logic/estimateFees.ts#L147

the initial value is 0 gaslitmit and 0 estimatedFees

With the "maxBigNumber", we can set the "gasLimit", "estimatedFees"... to the value from coin config.

hedi-edelbloute (approved):

PR #7440, opened 2024-07-26
🏗️ [Support] Request device access within `HWDeviceProvider` 

thesan:
### ✅ Checklist

  - ...

### 📝 Description

Factor all device interactions out of the Trustchain SDK into a new HWDeviceProvider. This is done in order to help unit testing of the Trustchain SDK later on and to better separate concerns.

Initialise the `HWDeviceProvider` with the `withDevice` function from `@ledgerhq/live-common/hw/deviceAccess`. The device id is then passed to the method of the Trustchain SDK which interact with the device.

### ❓ Context

- [LIVE-13211]

---

[LIVE-13211]: https://ledgerhq.atlassian.net/browse/LIVE-13211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre (changes requested):
**it's a great iteration but i would like to suggest a simplification on the SDK interface.**

if i understand correctly, we're abstracting out the need to pass `transport` in methods of the SDK, by instanciating the SDK with a way to create the transport instead (giving an implementation of withTransport)

on top of this we also need to provide a `deviceId` observable, this is the part where i'm a bit more loss both technically and conceptually: in term of lifecycle, some methods need the device, some don't, and they only need it in context of that method, so i think it would prefer to have the `deviceId` being passed as explicit parameter to these methods to express the fact that "hey i will need a device id because i will need the device". This is a pattern actually quite similar to what was done on the Coin Framework AccountBridge : we pass a deviceId (and no longer a transport), so we can still keep what you did on the `withDevice` implementation abstraction, but i would prefer to simplify out the need to pass in a global "device id observable" but instead just pass in the device.id (instead of the transport) where it's necessary (which also makes the diff a bit less impacting with current's sdk signature)

so each time we had "transport", it becomes "deviceId", for instance, I mean this change:

```diff
- await runWithDevice(device.id, async transport => {
+ await (
  sdk.getOrCreateTrustchain(
-   transport,
+   device.id,
    memberCreds,
    callbacks
  )
)
```

I think it may be especially important for LLM because the observable pattern will not work on LLM: we manage multiple devices on LLM unlike LLD where we always have a singleton device. so this deviceId observable lifecycle don't really exists. it also doesn't really exist on the web tools.

thesan:
> so each time we had "transport", it becomes "deviceId"

Got it! Thank you for having a look. I'll make these changes and apply the rework to LLM and web tools too.

libs/trustchain/src/HWDeviceProvider.ts
```diff
@@ -1,17 +1,25 @@
...
 
 export class HWDeviceProvider {
+  // NOTE withDevice should be imported statically but ATM making @ledgerhq/live-common a dependency of @ledgerhq/hw-trustchain creates a circular dependency
+  private withDevice: WithDevice;
+
+  private deviceId$: Observable<string>;
   private jwt?: JWT;
-  private hw?: ApduDevice;
+
+  constructor(withDevice: WithDevice, deviceId$: Observable<string>) {
```
KVNLS, line 17: Do we really need to rely on rxjs here? Can't we do it without it?

mcayuelas-ledger (approved):

apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useRemoveMember.ts
```diff
@@ -51,20 +51,21 @@ export function useRemoveMember({ device, member }: Props) {
...
-              onEndRequestUserInteraction: () => setUserDeviceInteraction(false),
-            },
-          );
+        if (!deviceRef.current) {
+          throw new Error("Device not found");
+        }
+        const newTrustchain = await sdkRef.current.removeMember(
+          deviceRef.current?.deviceId,
+          trustchainRef.current as Trustchain,
+          memberCredentialsRef.current as MemberCredentials,
```
gre, line 60: we shouldn't have to cast these, instead we may want to have the same trick you did with the `if(!deviceRef.current) throw`
you can have a catch all `if (!a || !b || !c) throw new Error("unexpected missing args")` for instance

also you can do `device.current.deviceId` as you just have proven to TS that it's settled. if TS can't manage it, then it"s preferable to do:

```
const device = deviceRef.current
if (!device) ...

removeMember(device.deviceId, 
```

> Side question but does it mean removeMember accepts a falsy deviceId? it probably shouldn't 🤔

apps/ledger-live-mobile/src/newArch/features/WalletSync/hooks/useAddMember.ts
```diff
@@ -40,19 +40,17 @@ export function useAddMember({ device }: { device: Device | null }) {
...
-              onEndRequestUserInteraction: () => setUserDeviceInteraction(false),
-            },
-          );
-          if (trustchainResult) {
-            transitionToNextScreen(trustchainResult);
-          }
-        });
+        const trustchainResult = await sdk.getOrCreateTrustchain(
+          device?.deviceId ?? "",
+          memberCredentialsRef.current as MemberCredentials,
```
gre, line 45: same casting we could get rid of (but what about the `?.deviceId` one ?)

apps/ledger-live-mobile/src/newArch/features/WalletSync/hooks/useRemoveMember.ts
```diff
@@ -59,20 +59,18 @@ export function useRemoveMember({ device, member }: Props) {
...
-            {
-              onStartRequestUserInteraction: () => setUserDeviceInteraction(true),
-              onEndRequestUserInteraction: () => setUserDeviceInteraction(false),
-            },
-          );
+        const newTrustchain = await sdk.removeMember(
+          device.deviceId,
+          trustchain,
+          memberCredentials,
+          member,
```
gre, line 66: ( 👍  on mobile we had cleaner code I hope we can also reach on LLD, see here, no casting )

libs/trustchain/tests/scenarios/removedMemberEjectedOnDeletedTrustchain.ts
```diff
@@ -13,7 +13,7 @@ export async function scenario(transport: Transport, { sdkForName }: ScenarioOpt
   const member2 = { name: name2, id: member2creds.pubkey, permissions: 0xffffffff };
 
   // auth with the device and init the first trustchain
-  const { trustchain } = await sdk1.getOrCreateTrustchain(transport, member1creds);
+  const { trustchain } = await sdk1.getOrCreateTrustchain("foo", member1creds);
```
gre, line 16: could we just use `""` instead of foo? or maybe we can also make a deviceId constant somewhere to explicit what this is.

gre (changes requested):

libs/trustchain/tests/scenarios/removedMemberEjectedOnDeletedTrustchain.ts
thesan: I ended up passing the device id to the test scenario instead of the transport

libs/trustchain/src/HWDeviceProvider.ts
```diff
@@ -1,17 +1,33 @@
...
+  /**
+   * TODO withDevice should be imported statically from @ledgerhq/live-common/hw/deviceAccess
+   *
+   * but ATM making @ledgerhq/live-common a dependency of @ledgerhq/trustchain causes:
+   * > Turbo error: Invalid package dependency graph: cyclic dependency detected:
+   * >   @ledgerhq/trustchain,@ledgerhq/live-wallet,@ledgerhq/live-common
+   *
+   * Maybe hw/deviceAccess.ts and hw/index.ts could be moved to @ledgerhq/devices
+   * This would break the cyclic dependency as @ledgerhq/live-common would depend on @ledgerhq/devices
+   * but not the other way around.
```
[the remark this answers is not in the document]
KVNLS: I prefer to keep it as is, otherwise it involves much work to be done. ( + I prefer that you inject the withDevice as a parameter rather than importing it in the trustchain sdk)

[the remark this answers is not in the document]
gre: the idea to move hw/* and notably the deviceAccess / index files of it to a library is a good idea, but i think it's the scope of device team tho. also the impact is quite important in that many parts depends on this. but this could be discussed with that team.

apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useRemoveMember.ts
gre: using React ref usually is done when we want to prevent an effect to be retriggered over and over, and when we instead just need to access the "current" state and not to "trigger each time it changes"

gre (approved):

PR #7513, opened 2024-08-05
ton jettons integration

ayelenmurano:
### ✅ Checklist

  - ...

### 📝 Description

This PR adds support for TON jetton transfers into ledger live.

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/api.ts
```diff
@@ -82,6 +86,37 @@ export async function fetchAccountInfo(addr: string): Promise<TonAccountInfo> {
...
+  opts?: {
+    jettonMaster?: string;
+    startLt?: string;
+    endLt?: string;
+  },
+): Promise<TonJettonTransfer[]> {
+  const address = Address.parse(addr);
+  const urlAddr = address.toString({ bounceable: false, urlSafe: true });
+  let url = `/jetton/transfers?address=${urlAddr}&limit=256`;
+  if (opts?.jettonMaster != null) url += `&jetton_master=${opts.jettonMaster}`;
```
hzheng-ledger, line 100: If I understand well, the jettonMaster should never be null. Isn't it?

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/api.ts
```diff
@@ -82,6 +86,37 @@ export async function fetchAccountInfo(addr: string): Promise<TonAccountInfo> {
...
+  return (await fetch<TonResponseJettonTransfer>(url)).jetton_transfers;
+}
+
+export async function fetchJettonWallets(opts?: {
+  address?: string;
+  jettonMaster?: string;
+}): Promise<TonJettonWallet[]> {
+  let url = `/jetton/wallets?limit=256`;
+  if (opts?.jettonMaster != null) url += `&jetton_address=${opts.jettonMaster}`;
+  if (opts?.address != null) {
```
hzheng-ledger, line 112: If I understand well, the address and the jettonMaster should never be null in Ledger Live. Isn't it?

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
```diff
@@ -33,6 +40,32 @@ export async function getTransactions(
...
+export async function getJettonTransfers(
+  addr: string,
+  startLt?: string,
+): Promise<TonJettonTransfer[]> {
+  const txs = await fetchJettonTransactions(addr, { startLt });
+  if (txs.length === 0) return txs;
+  let tmpTxs: TonJettonTransfer[];
+  let isUncompletedResult = true;
+
+  while (isUncompletedResult) {
```
hzheng-ledger, line 52: I think you can check whether the txs.length<256, if so, you can return txs directly.
because in most case, the number of transactions is between 1 and 256. so you can avoid calling "fetchJettonTransactions" twice.
if you check whether txs.length<256 in the loop, you can simplify your code instead of using isUncompletedResult.

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
```diff
@@ -164,3 +197,109 @@ export function mapTxToOps(
...
+    const hash = tx.transaction_hash;
+
+    if (isReceiving) {
+      ops.push({
+        id: encodeOperationId(tokenAccountId, hash, "IN"),
+        hash,
+        type: "IN",
+        value: BigNumber(tx.amount),
+        fee: BigNumber(0),
+        blockHeight: 1, // we don't have block info
```
hzheng-ledger, line 242: if blockHeight is always 1, all the transactions are considered as confirmed transactions?

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
```diff
@@ -164,3 +197,109 @@ export function mapTxToOps(
...
+          explorerHash: hash,
+          comment: {
+            isEncrypted: false,
+            text: tx.forward_payload ? decodeForwardPayload(tx.forward_payload) : "",
+          },
+        },
+      });
+    }
+
+    if (isSending) {
```
hzheng-ledger, line 260: I know it is not possible to send a tx to the itself in Ledger live. but if the account is exported from other wallets, a transaction is interpreted as two operations if the recipient and the sender is the same?

libs/coin-modules/coin-ton/src/utils.ts
```diff
@@ -118,3 +195,108 @@ export const getLedgerTonPath = (path: string): number[] => {
   }
   return numPath;
 };
+
+/**
+ * Checks if the given payload is a jetton transfer.
+ */
+export const isJettonTransfer = (payload: TonPayloadFormat): boolean =>
```
hzheng-ledger, line 202: the "isJettonTransfer" method is not used, right?

libs/coin-modules/coin-ton/src/synchronisation.ts
```diff
@@ -24,43 +47,134 @@ export const getAccountShape: GetAccountShape<Account> = async info => {
...
+  return {
+    type: "TokenAccount",
+    id: tokenAccountId,
+    parentId,
+    token: { ...token, contractAddress: jettonWalletAddress }, // the contract address is replaced for the jetton wallet address, it will be use for the token transfer
+    balance: new BigNumber(balance),
+    spendableBalance: new BigNumber(balance),
+    operations,
+    operationsCount: operations.length,
+    pendingOperations: maybeExistingSubAccount ? maybeExistingSubAccount.pendingOperations : [],
```
hzheng-ledger, line 140: I think we don't use pendingOperations in getAccountShape. 
operations with blockheight = null are considered as pending operations

you can just 

```
pendingOperations: [],
```
I think

hzheng-ledger (changes requested):
see my comments. Thanks

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
ayelenmurano: But what happen if the address has more than 256? We are requesting all of the token transfers (related with the different tokens)

libs/coin-modules/coin-ton/src/utils.ts
ayelenmurano: No, the function is used in a test file, I'll move it into that test file. Thank you

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/api.ts
ayelenmurano: Hi! When synchronizing the account, we don't search by token, we search all of the transfers and then we separate them by token with the function [getSubAccounts](https://github.com/LedgerHQ/ledger-live/blob/f063790970e66efb7db488bfef53b8df5b99ea7f/libs/coin-modules/coin-ton/src/synchronisation.ts#L147)

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/api.ts
ayelenmurano: Yes, this information will always be in Ledger Live. I'll delete the conditional

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
ayelenmurano: Yes, in this case we would receive one only transaction from the endpoint (for the IN and OUT) and we would display both (eceived and sent transactio would be displayed).

libs/coin-modules/coin-ton/src/synchronisation.ts
ayelenmurano: I tested and it didn't work as expected if I change the code. When the account is synchronized again (and the transaction wasn't received yet), the pending operation is no longer displayed.
We receive the pendingOperations in getAccountShape and we use them in postSync too.

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
ayelenmurano: Yes, TON transactions are irreversible after just one confirmation

hzheng-ledger (dismissed):
It looks good to me.

PR #7530, opened 2024-08-06
fix: export language and theme from desktop settings

KVNLS:
### ✅ Checklist

  - ...

### 📝 Description

Import language and theme from Desktop settings with QRCode.

https://github.com/user-attachments/assets/5783c533-839b-4c2c-bc39-f491073348c1

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-13609

---

### DIFF
--- a/.changeset/lucky-pugs-own.md
+++ b/.changeset/lucky-pugs-own.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": minor
+"@ledgerhq/live-wallet": minor
+---
+
+Export theme and language from desktop settings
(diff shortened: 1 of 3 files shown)


gre (changes requested):
I forgot the fact we will literally spread the result of the "ResultSettings" object directly into the `settings` store of LLM.

this is kinda unsafe. if we want to introduce this feature, I think we need to start to actively validate field by field.

Because, typically here, we have no guarantee that the same language and theme from LLD will exist on LLM.  **typically, a few years ago, we had more than Light and Dark themes in the list of choices of the themes** so there is a possibility that importing from such old LLD would break LLM here 🤔 so it can be dangerous in that we have no guarantee that the string value that comes from LLD for both the lang and the theme is going to exists and be recognise on LLM side. some theme or lang are not necessarily on both sides, at least we don’t guarantee in the code, so we need protection, and fallback to default theme if not recognised.

On the ~~fix~~ feature itself, I think we need to wait product decision, both lang and theme on LLM uses the system settings. shouldn't we keep this and not override the one defined by LLD? `theme: "system"` is the default, maybe user prefer LLM to use the system wide settings 🤔

PR #7559, opened 2024-08-08
feat: add header and bottom bar animations on webview scroll [LIVE-13180]

Justkant:
### ✅ Checklist

- [x] **Covered by automatic tests.** The animation itself is not covered by tests as we disable/mock them in tests, but it's tested a lot manually
  - Animations on web3hub pages that are still behind a disabled FF

### 📝 Description

Add header and bottom bar animations on Webview scroll
Also sync scroll on other pages with paddingTop under the absolute position for the header

Refactor:
- move header directly in the screen component
- Allows to remove shared header context
- Simplify a bit by removing the need for insets on some children components
- Share most of the animation logic (exception with the BottomBar currently as it's not even using the transform)
- Fix some edge cases with the scrollHandler logic to not diff if no begin drag was done and to not start on a negative value
- Going with headerShown false requires us to use the SafeAreaView on our screens
- All tests still work after big refactor with no changes (no implementation detail in the tests makes this possible)
- Could extract the opacity animation from the AnimatedBar to simplify a bit the component and allow the parent to decide how to display the children

https://github.com/user-attachments/assets/64684828-c6c5-4618-bc83-31be4652d6ef

### ❓ Context

- **JIRA or GitHub link**: [LIVE-13180]

---

[LIVE-13188]: https://ledgerhq.atlassian.net/browse/LIVE-13188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-13180]: https://ledgerhq.atlassian.net/browse/LIVE-13180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/Disclaimer/index.tsx
```diff
@@ -35,7 +35,7 @@ export default function Disclaimer({
 
       {description ? (
         <Flex mt={6}>
-          <Text fontSize={14} lineHeight={"22px"} color="smoke">
+          <Text numberOfLines={12} fontSize={14} lineHeight={"22px"} color="smoke">
```
RamyEB, line 38: Does it impact Disclaimer or just a small modification ?

apps/ledger-live-mobile/src/locales/en/common.json
```diff
@@ -6844,6 +6844,9 @@
       "description": "Discover the best of web3 curated by Ledger"
     },
     "main": {
+      "manifestsList": {
+        "title": "Discover"
```
RamyEB, line 6848: Is it possible to put it in `web3hub.manifestsList` ? 
If not, could be a good idea to clarify a bit the difference btw  `web3hub.manifestsList ` and `web3hub.main.manifestsList `

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/ManifestsList/index.tsx
```diff
@@ -38,9 +39,8 @@ const renderItem = ({
   return <ManifestItem manifest={item} onPress={extraData} />;
 };
 
-export default function ManifestsList({ navigation, onScroll, pb = 0 }: Props) {
+export default function ManifestsList({ navigation, onScroll, title, pt = 0, pb = 0 }: Props) {
```
RamyEB, line 42: Not a big fan of passing style this way, can we use spreading ?

RamyEB (changes requested):

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/Disclaimer/index.tsx
Justkant: While navigating on the mocks I saw a big issue with descriptions too long, so just included the fix here

apps/ledger-live-mobile/src/locales/en/common.json
Justkant: The difference is clear we have a manifestList component in web3hub that defaults to `web3hub.manifestsList` and we also use the manifestList in the main screen with a specific title so we also need `web3hub.main.manifestsList`

> Is it possible to put it in web3hub.manifestsList ?
That would be the wrong place to put it as it's only used in the main screen using the manifestList component and not in the component itself

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/ManifestsList/index.tsx
Justkant: You loose performances by spreading and changing refs all the time
I also don't want people to pass more props/styles than what is needed, if I could do this cleaner no props for style would be needed at all but in this case you still need some control to share the logic between different screens
Spreading hides what you need and expect

RamyEB: Alright

apps/ledger-live-mobile/src/locales/en/common.json
Justkant: Both are components one is in main the other outside

If you want to be consistant with the actual folder hierarchy you would go for `web3hub.components.manifestList` and `web3hub.screens.main.components.manifestList` (you could probably do `web3hub.screens.main.manifestList` but then why not `web3hub.main.manifestList`)
But all this feels a bit convoluted just for strings that you can search

RamyEB (approved):

PR #7565, opened 2024-08-08
Feat(CN-472): add category filter with discover category deeplink param

beths-ledger:
### ✅ Checklist

  - DESKTOP ONLY: ledgerlive://discover?restaking should load discover section with "Restaking" tagged platforms pre-filtered.
  - Other categories should filter by those categories
  - invalid category does not return any results 

### 📝 Description

Add discover section deep link initial category filter to enable links like ledgerlive://discover?category=restaking to load the discover section with an initial category pre-selected.

This is required for the new "restaking" banner on earn dashboard.

Routes with a discover platform app id will not be impacted.

ledgerlive://discover?category=restaking
image

### ❓ Context

- **JIRA or GitHub link**: [CN-472](https://ledgerhq.atlassian.net/browse/CN-472) https://ledgerhq.atlassian.net/browse/CN-472

---

[CN-472]: https://ledgerhq.atlassian.net/browse/CN-472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/Default.tsx
+++ b/apps/ledger-live-desktop/src/renderer/Default.tsx
@@ -333,6 +333,11 @@ export default function Default() {
                                 render={withSuspense(PlatformCatalog)}
                                 exact
                               />
+                              <Route
+                                path="/platform?category=:category"
+                                render={withSuspense(PlatformCatalog)}
+                                exact
+                              />
                               <Route path="/platform/:appId?" component={LiveApp} />
                               <Route path="/earn" render={withSuspense(Earn)} />
                               <Route
--- a/apps/ledger-live-desktop/src/renderer/Default.tsx
+++ b/apps/ledger-live-desktop/src/renderer/Default.tsx
@@ -333,11 +333,6 @@ export default function Default() {
                                 render={withSuspense(PlatformCatalog)}
                                 exact
                               />
-                              <Route
-                                path="/platform?category=:category"
-                                render={withSuspense(PlatformCatalog)}
-                                exact
-                              />
                               <Route path="/platform/:appId?" component={LiveApp} />
                               <Route path="/earn" render={withSuspense(Earn)} />
                               <Route
--- a/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
+++ b/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
@@ -310,11 +310,7 @@ export function useDeepLinkHandler() {
           if (path?.startsWith("protect")) {
             navigate(`/recover/${path}`, undefined, search);
           } else {
-            // load discover section wtih an initial category filter parameter, e.g. /discover?category=staking
-            if (!path && query.category) {
-              navigate(`/platform`, query);
-            }
-            navigate(`/platform/${path ?? ""}`, query);
+            navigate(`/platform/${path ?? ""}`, query, search);
           }
           break;
         case "wc": {
--- a/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
+++ b/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
@@ -310,6 +310,10 @@ export function useDeepLinkHandler() {
           if (path?.startsWith("protect")) {
             navigate(`/recover/${path}`, undefined, search);
           } else {
+            // load discover section wtih an initial category filter parameter, e.g. /discover?category=staking
+            if (!path && query.category) {
+              navigate(`/platform`, query);
+            }
             navigate(`/platform/${path ?? ""}`, query);
           }
           break;
--- a/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
+++ b/apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
@@ -310,7 +310,7 @@ export function useDeepLinkHandler() {
           if (path?.startsWith("protect")) {
             navigate(`/recover/${path}`, undefined, search);
           } else {
-            navigate(`/platform/${path ?? ""}`, query, search);
+            navigate(`/platform/${path ?? ""}`, query);
           }
           break;
         case "wc": {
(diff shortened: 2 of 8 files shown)


libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -797,8 +797,11 @@ export interface Categories {
   reset: () => void;
 }
 
-export function useCategories(manifests): Categories {
-  const [selected, setSelected] = useState(DISCOVER_INITIAL_CATEGORY);
+/** e.g. "all", "restaking", "services", etc */
+export type CategoryId = Categories["selected"];
+
+export function useCategories(manifests, initialCategory?: CategoryId): Categories {
```
beths-ledger, line 803: @marcotoniut-ledger this will be needed for the LLM route

apps/ledger-live-desktop/src/renderer/Default.tsx
```diff
@@ -333,6 +333,11 @@ export default function Default() {
                                 render={withSuspense(PlatformCatalog)}
                                 exact
                               />
+                              <Route
+                                path="/platform?category=:category"
+                                render={withSuspense(PlatformCatalog)}
+                                exact
+                              />
```
Justkant, line 340: I don't think you want the exact here
Also and more importantly I don't think you need a new route for a param that should already be accessible in the route above

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
```diff
@@ -310,6 +310,10 @@ export function useDeepLinkHandler() {
           if (path?.startsWith("protect")) {
             navigate(`/recover/${path}`, undefined, search);
           } else {
+            // load discover section wtih an initial category filter parameter, e.g. /discover?category=staking
+            if (!path && query.category) {
+              navigate(`/platform`, query);
+            }
```
Justkant, line 316: Same here, I'm not confident that we actually need this as query is already passed below even if we don't have an `appId` (named `path` here)

apps/ledger-live-desktop/src/renderer/screens/platform/v2/Catalog/Browse/Search.tsx
```diff
@@ -22,6 +22,8 @@ export function Search({ categories, search }: Props) {
     [categories.categories, t],
   );
 
+  const selected = options.find(opt => opt.value === categories.selected);
```
Justkant, line 25: This should be done in a memo to avoid recomputing on every render, also it should be better to also move the fallback `options[0]` in the memo directly
We can also rename the const to `selectedOption` or `defaultOption`

apps/ledger-live-desktop/src/renderer/screens/platform/v2/Catalog/index.tsx
```diff
@@ -6,13 +6,26 @@ import { Browse } from "./Browse";
...
 
 export function Catalog() {
+  const { t } = useTranslation();
   const recentlyUsedDB = useRecentlyUsedDB();
+  const { state } = useLocation() as ReturnType<typeof useLocation> & {
+    state?: {
+      category?: Categories["selected"];
+    };
+  };
+  const deeplinkInitialCategory = state?.category;
```
Justkant, line 20: Location state with our usage of the hash router can be unreliable, as it might be reset on navigation
You should already have the information you want in the search field instead

Justkant (changes requested):
I think we might want to see if we can simplify the changes as I don't think most of them are necessary for the goal

apps/ledger-live-desktop/src/renderer/screens/platform/v2/Catalog/index.tsx
beths-ledger: Nice one, I'll just use the `search` instead

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking.ts
beths-ledger: Ah yes the `/` is still a match with the original Route even if there's only a search param after it.
Do you see any issues with passing the `search` (if there) to all platform navigations including those to specific app ids?
i.e. 
```
            navigate(`/platform/${path ?? ""}`, query, search);
```

apps/ledger-live-desktop/src/renderer/screens/platform/v2/Catalog/index.tsx
```diff
@@ -6,13 +6,28 @@ import { Browse } from "./Browse";
...
 
   const { t } = useTranslation();
-  const { categories, recentlyUsed, disclaimer, search, localLiveApps } =
-    useCatalog(recentlyUsedDB);
+
+  const { state } = useLocation() as ReturnType<typeof useLocation> & {
+    state?: {
+      category?: Categories["selected"];
+    };
+  };
```
Justkant, line 21: Small detail but I think you should be able to use the generics here
```suggestion
  const { state } = useLocation<{
    category?: Categories["selected"];
  }>();
```

Justkant (commented):
Latest small detail

apps/ledger-live-desktop/src/renderer/screens/platform/v2/Catalog/index.tsx
```diff
@@ -6,13 +6,27 @@ import { Browse } from "./Browse";
...
+
+  const deeplinkInitialCategory = state?.category;
+
+  const {
+    categories,
+    recentlyUsed,
+    disclaimer,
+    search: searchInput,
+    localLiveApps,
+  } = useCatalog(recentlyUsedDB, deeplinkInitialCategory ?? undefined);
```
Justkant, line 29: Do we need the fallback here ?
I expect category to be `Categories["selected"] | undefined`
```suggestion
  } = useCatalog(recentlyUsedDB, deeplinkInitialCategory);
```

Justkant (approved):

gre (approved):

PR #7598, opened 2024-08-14
Fixes typescript and build issues

bretep:
### ✅ Checklist

No new features. Only fixes project setup and build when following the README

### 📝 Description
I ran into project setup issues in preparation for adding PulseChain support to ledger-live. These should be non-breaking changes.

- Fixed issues related to TypeScript and the build process in multiple files across the project.
- Enforce pnpm version otherwise, the lock file will change.
- The "target" field in tsconfig.json files is changed from "es5" to "ES2020", and changes have been made in other configuration files.
- The versions of some dependencies like "@solana/spl-token" and "@solana/web3.js" in package.json files are updated.

### Tests performed

1. Cloned branch to new location
2. Followed the README.md
Ran the following:
```
pnpm install
pnpm build:libs
pnpm build:tests
pnpm build:dummy-apps
pnpm build:dummy-wallet-app
pnpm build:dummy-live-app
pnpm build:actions
pnpm build:cli
pnpm build:coin
pnpm build:domain
pnpm build:network
pnpm build:config
pnpm build:llc
pnpm build:lld:deps
pnpm build:lld
pnpm build:llm:deps
pnpm build:ljs
pnpm build:web-tools
```

All were built without errors, which was not the case before this commit.
---


bretep:
I'm not able to see the failures. Please let me know if there is something I need to resolve.

package.json
```diff
@@ -177,6 +177,10 @@
     "unimported": "1.30.0",
     "zx": "^7.2.2"
   },
+  "engines": {
+    "pnpm": "9.7.0"
```
Justkant, line 181: If you check the `.prototools` file, you'll see that we use `9.4.0` of `pnpm`
```suggestion
    "pnpm": "9.4.0"
```

In the past I advocated for the use of `packageManager` field which is better than engines IMO because engines is more intended for consumers of a lib not developers AFAIK.
Probably better to use `packageManager` instead of `engines` here

libs/ledger-live-common/package.json
```diff
@@ -258,7 +258,7 @@
     "@ledgerhq/types-cryptoassets": "workspace:^",
     "@ledgerhq/types-devices": "workspace:^",
     "@ledgerhq/types-live": "workspace:^",
-    "@solana/web3.js": "1.77.3",
+    "@solana/web3.js": "1.95.2",
```
Justkant, line 261: Not sure we want to update unrelated libs in the same PR that just touches typescript and build stuff

apps/ledger-live-mobile/ios/Podfile
```diff
@@ -8,6 +8,7 @@ require Pod::Executable.execute_command('node', ['-p',
 
 
 platform :ios, '13.4'
+use_modular_headers!
```
Justkant, line 11: We already setup `use_frameworks! :linkage => linkage.to_sym` below that will be static or dynamic depending on env vars
So I don't think you need to add anything here, as seen here too: https://stackoverflow.com/a/53265888

apps/ledger-live-mobile/android/app/build.gradle
```diff
@@ -128,7 +128,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 36176158
-        versionName "3.40.1"
+        versionName "3.47.2"
```
Justkant, line 131: Not sure you need to touch this

apps/ledger-live-mobile/ios/ledgerlivemobile/Info.plist
```diff
@@ -35,7 +35,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.40.1</string>
+	<string>3.47.2</string>
```
Justkant, line 38: Not sure you need to touch this either because it's handle by scripts during the deploy on main/prod

libs/coin-modules/coin-solana/package.json
```diff
@@ -53,8 +53,8 @@
     "@ledgerhq/logs": "workspace:^",
     "@ledgerhq/types-cryptoassets": "workspace:^",
     "@ledgerhq/types-live": "workspace:^",
-    "@solana/spl-token": "^0.3.7",
-    "@solana/web3.js": "1.77.3",
+    "@solana/spl-token": "^0.4.8",
+    "@solana/web3.js": "1.95.2",
```
Justkant, line 57: Same as the other not sure about bumpimg solana deps, is this related to some types changes ?

libs/coin-modules/coin-solana/src/api/chain/web3.ts
```diff
@@ -334,25 +334,29 @@ export function buildStakeWithdrawInstructions({
...
+  // Create the split instruction with the rentExemptReserve
+  const splitIx = StakeProgram.split(
+    {
+      authorizedPubkey: new PublicKey(authorizedAccAddr),
+      lamports: amount,
+      stakePubkey: new PublicKey(stakeAccAddr),
+      splitStakePubkey: new PublicKey(splitStakeAccAddr),
+    },
+    rentExemptReserve,
+  ).instructions[1];
```
Justkant, line 353: Not sure why you change logic in a PR that should only touch types and build issues

libs/coin-modules/coin-solana/src/buildTransaction.ts
```diff
@@ -46,18 +46,24 @@ export const buildTransactionWithAPI = async (
...
   }
-  return buildInstructionsForCommand(commandDescriptor.command);
+  return await buildInstructionsForCommand(commandDescriptor.command, api);
 }
 
-function buildInstructionsForCommand(command: Command): TransactionInstruction[] {
+async function buildInstructionsForCommand(
+  command: Command,
+  api: ChainAPI,
+): Promise<TransactionInstruction[]> {
```
Justkant, line 66: And that change might even be breaking now that you return a promise

libs/live-countervalues/package.json
```diff
@@ -21,20 +21,22 @@
...
-    "@ledgerhq/logs": "workspace:*",
+    "@ledgerhq/cryptoassets": "workspace:*",
+    "@ledgerhq/live-env": "workspace:*",
     "@ledgerhq/live-network": "workspace:*",
     "@ledgerhq/live-promise": "workspace:*",
-    "@ledgerhq/live-env": "workspace:*"
+    "@ledgerhq/logs": "workspace:*",
+    "@ledgerhq/types-cryptoassets": "workspace:*",
+    "@ledgerhq/types-live": "workspace:*",
+    "url": "^0.11.4"
```
Justkant, line 32: Do we need to add url here ?

tools/actions/turbo-affected/build/main.js
Justkant, line 1: We probably want to revert this change if it was only formatting changes, unless you did a rebuild of this script for a reason

apps/cli/src/commands/device/managerListApps.ts
```diff
@@ -4,6 +4,11 @@ import { withDevice } from "@ledgerhq/live-common/hw/deviceAccess";
 import getDeviceInfo from "@ledgerhq/live-common/hw/getDeviceInfo";
 import { listAppsUseCase } from "@ledgerhq/live-common/device/use-cases/listAppsUseCase";
 import { deviceOpt } from "../../scan";
+import { ListAppsEvent, ListAppsResult } from "@ledgerhq/live-common/apps/types";
+
+function isResultEvent(event: ListAppsEvent): event is { type: "result"; result: ListAppsResult } {
+  return event.type === "result";
+}
```
Justkant, line 11: Nice, another small improvement we could do would be to create a type for `{ type: "result"; result: ListAppsResult }` and use it both here and in `ListAppsEvent`

Justkant (commented):
Hello and thank you for the contribution

We'll need to have reviews from people maintaining our mono-repo and build system too

Could you also list/explain the issues you fixed with the build process in multiple files across the project ? I don't really see a lot of changes on the build side (unless it's the main script rebuild)

Could you also explain why you need to target `ES2020` in the tsconfig ? Also I don't think you modified the correct config as we currently share a single config for all libs in the monorepo and cli or web-tools are not the ones

About the pnpm version we already enforce a version with Proto and it's documented in the README

I also don't see a big reason to update dependencies of solana (unless big issues with types) and modifying the logic while we should only be fixing typescript/build issues

bretep:
I will review and document each error and the reason for the package upgrades. 

By chance, have you checked out the repo on a clean computer, without global packages installed, and followed the readme for setup?

bretep:
I took a step back because of your comment about `proto` to check and see why I would have received all these issues. It turns out that `proto` was not being used when working in my IDE, and I was using newer node versions.

When running outside of my IDE, I could clone and run through the README with no issues. Many of the errors that were stopping the build are now `warnings`.

Thank you for reviewing the PR and I apologize for the wasted time.

PR #7618, opened 2024-08-19
[WIP] Feat/jetton integration

hzheng-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Duplication of the PR: https://github.com/LedgerHQ/ledger-live/pull/7513. I create this PR to check the CI status.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-13530

---


libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/api.ts
```diff
@@ -82,6 +86,37 @@ export async function fetchAccountInfo(addr: string): Promise<TonAccountInfo> {
...
+  opts?: {
+    jettonMaster?: string;
+    startLt?: string;
+    endLt?: string;
+  },
+): Promise<TonJettonTransfer[]> {
+  const address = Address.parse(addr);
+  const urlAddr = address.toString({ bounceable: false, urlSafe: true });
+  let url = `/jetton/transfers?address=${urlAddr}&limit=256`;
+  if (opts?.jettonMaster != null) url += `&jetton_master=${opts.jettonMaster}`;
```
hzheng-ledger, line 100: If I understand well, the jettonMaster should never be null. Isn't it?

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
```diff
@@ -33,6 +40,32 @@ export async function getTransactions(
...
+export async function getJettonTransfers(
+  addr: string,
+  startLt?: string,
+): Promise<TonJettonTransfer[]> {
+  const txs = await fetchJettonTransactions(addr, { startLt });
+  if (txs.length === 0) return txs;
+  let tmpTxs: TonJettonTransfer[];
+  let isUncompletedResult = true;
+
+  while (isUncompletedResult) {
```
hzheng-ledger, line 52: I think you can check whether the txs.length<256, if so, you can return txs directly. 
because in most case, the number of transactions is between. you can avoid calling "fetchJettonTransactions" twice.
if you check txs.length<256 in the loop, I think you can simplify your code.

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
```diff
@@ -164,3 +197,109 @@ export function mapTxToOps(
...
+          explorerHash: hash,
+          comment: {
+            isEncrypted: false,
+            text: tx.forward_payload ? decodeForwardPayload(tx.forward_payload) : "",
+          },
+        },
+      });
+    }
+
+    if (isSending) {
```
hzheng-ledger, line 260: I know it is not possible to send a tx to the itself in Ledger live. but if the account is exported from other wallets, a transaction is interpreted as two operations if the recipient and the sender is the same?

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
```diff
@@ -164,3 +197,109 @@ export function mapTxToOps(
...
+    const hash = tx.transaction_hash;
+
+    if (isReceiving) {
+      ops.push({
+        id: encodeOperationId(tokenAccountId, hash, "IN"),
+        hash,
+        type: "IN",
+        value: BigNumber(tx.amount),
+        fee: BigNumber(0),
+        blockHeight: 1, // we don't have block info
```
hzheng-ledger, line 242: if blockHeight is always 1, all the transactions are considered as confirmed transactions?

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/api.ts
ayelenmurano: Hi! When synchronizing the account, we don't search by token, we search all of the transfers and then we separate them by token with the function [getSubAccounts](https://github.com/LedgerHQ/ledger-live/blob/f063790970e66efb7db488bfef53b8df5b99ea7f/libs/coin-modules/coin-ton/src/synchronisation.ts#L147)

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
ayelenmurano: But what happen if the address has more than 256? We are requesting all of the token transfers (related with the different tokens)

libs/coin-modules/coin-ton/src/bridge/bridgeHelpers/txn.ts
ayelenmurano: Yes, in this case we would receive one only transaction from the endpoint (for the IN and OUT) and we would display both (eceived and sent transactio would be displayed).

PR #7642, opened 2024-08-21
[LIVE-13694][LLD] Fix app install hanging on Windows due to performance monitoring related stuff

ofreyssinet-ledger:
### ✅ Checklist

- [ ] **Covered by automatic tests.** we're removing dead code
  - removal of dead code that was previously used to monitor performance using Sentry.

#### 1. WHAT'S THE ISSUE ?

When installing apps, or firmware updates, on Windows, the installation process completely halts after a few percents of completion, never resuming, never showing any error message.

#### 2. WHAT'S THE REASON FOR THE ISSUE ?
 - a) at each send of an APDU, we are logging something like that `["apdu", "=> e012345678"]`. So far, nothing crazy.
    > Screenshot 2024-08-22 at 10 28 27

- b) those logs were captured by `connectSentryToLogs`, with the purpose of getting nice perf monitoring like that:
  > 178246419-e84c4a87-32ad-4996-9e10-0c44d6b4884f

  > Screenshot 2024-08-22 at 10 31 25

- c) `@sentry/node` was recently updated, we suspect that there is a bad memory management on the windows side.

- d) this usage of Sentry spans anyway was not great because it means we were _always_ recording "apdu spans" even when we were _not_ trying to measure performance, so it was bound to actually negatively impact performance.

**The trigger:** when we install apps, we receive a bunch of APDUs (several hundreds) from the HSM (backend), to sequentially send to the device. Then the system above goes 💥 except it doesn't actually explode, it just blocks completely the internal thread, with no exception being thrown... Silent death 💀 

#### 3. WHY FULLY DELETING THIS PERF MONITORING CODE IS OK ?
**TL;DR: it's "dead" code** ever since the coin logic was definitely moved to the `renderer` process of LLD in this PR https://github.com/LedgerHQ/ledger-live/pull/2721.
It's used only in one place in the codebase: 
> Screenshot 2024-08-21 at 18 45 36

The thing is that if you look at how `startSpan` works, it will only work if in the same process, `connectSentryToLogs` has been previously called.
And as we can see quickly with a search, `connectSentryToLogs` is only called in the `internal` process.
When this was setup, it was probably working, because all the coin logic also lived in the `internal` process... It's not longer the case, now it's on `renderer` (cf. #2721).

Everything related to `startSpan` has thus been removed, **including in bitcoin-related code** as it can be considered "dead", or at least completely useless.

200w-1

### ❓ Context

- **JIRA or GitHub link**: [LIVE-13694]

---

[LIVE-13694]: https://ledgerhq.atlassian.net/browse/LIVE-13694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Wozacosta (approved):

mcayuelas-ledger (approved):

valpinkman (approved):
🚀

PR #7666, opened 2024-08-23
Setup analytics for Web3Hub

RamyEB:
### ✅ Checklist

  - ...

### 📝 Description

Setup analytics for Web3Hub

- [x] App clicked from main page
- [x] App clicked from horizontal list
- [x] App clicked from search page
- [x] Input focus from main page
- [x] categories clicked
- [x] disclaimer toggle

### ❓ Context

- **JIRA or GitHub link**:  [Ticket](https://ledgerhq.atlassian.net/browse/LIVE-12752)

---

### DIFF
--- a/.changeset/pink-walls-tickle.md
+++ b/.changeset/pink-walls-tickle.md
@@ -2,4 +2,4 @@
 "live-mobile": patch
 ---
 
-Analytics improved on Web3Hub
+Setup analytics on Web3Hub
--- a/.changeset/pink-walls-tickle.md
+++ b/.changeset/pink-walls-tickle.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Analytics improved on Web3Hub
--- a/apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
@@ -148,9 +148,6 @@ export default function MainNavigator() {
           listeners={({ navigation }) => ({
             tabPress: e => {
               e.preventDefault();
-              track("web3hub_section_clicked", {
-                page: "Home",
-              });
               managerLockAwareCallback(() => {
                 navigation.navigate(NavigatorName.Web3HubTab);
               });
--- a/apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
@@ -20,7 +20,6 @@ import customTabBar from "../TabBar/CustomTabBar";
 import { MainNavigatorParamList } from "./types/MainNavigator";
 import { isMainNavigatorVisibleSelector } from "~/reducers/appstate";
 import EarnLiveAppNavigator from "./EarnLiveAppNavigator";
-import { track } from "~/analytics";
 
 const Tab = createBottomTabNavigator<MainNavigatorParamList>();
 
--- a/apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
@@ -20,6 +20,7 @@ import customTabBar from "../TabBar/CustomTabBar";
 import { MainNavigatorParamList } from "./types/MainNavigator";
 import { isMainNavigatorVisibleSelector } from "~/reducers/appstate";
 import EarnLiveAppNavigator from "./EarnLiveAppNavigator";
+import { track } from "~/analytics";
 
 const Tab = createBottomTabNavigator<MainNavigatorParamList>();
 
@@ -147,6 +148,9 @@ export default function MainNavigator() {
           listeners={({ navigation }) => ({
             tabPress: e => {
               e.preventDefault();
+              track("web3hub_section_clicked", {
+                page: "Home",
+              });
               managerLockAwareCallback(() => {
                 navigation.navigate(NavigatorName.Web3HubTab);
               });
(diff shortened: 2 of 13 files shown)


apps/ledger-live-mobile/src/newArch/features/Web3Hub/screens/Web3HubSearch/components/Header/index.tsx
```diff
@@ -54,6 +55,11 @@ export default function Web3HubSearchHeader({ navigation, onSearch, layoutY }: P
               returnKeyType="done"
               value={search}
               onChangeText={setSearch}
+              onFocus={() => {
+                track("web3hub_search_focus", {
+                  component: `SearchInput`,
+                });
+              }}
```
Justkant, line 62: Not sure tracking this is really needed

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/Disclaimer/useDisclaimerViewModel.ts
```diff
@@ -71,6 +76,9 @@ export default function useDisclaimerViewModel(goToApp: (manifestId: string) =>
   }, [disclaimerManifest, goToApp, isChecked, setWeb3HubDB]);
 
   const onClose = useCallback(() => {
+    track("web3hub_disclaimer_closed", {
+      component: `Disclaimer`,
+    });
```
Justkant, line 81: That might send too much informations because close is usually called multiple times but also bad informations because close is called regardless of it being close onConfirm or when really closing

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/Disclaimer/useDisclaimerViewModel.ts
```diff
@@ -51,8 +52,12 @@ export default function useDisclaimerViewModel(goToApp: (manifestId: string) =>
   );
 
   const toggleCheck = useCallback(() => {
+    track("web3hub_disclaimer_toggle", {
+      component: `Disclaimer`,
+      action: `turn checkbox to ${isChecked ? "off" : "on"}`,
+    });
```
Justkant, line 58: Not sure we actually need to know this, even more as we don't even look if we succeed going to the app
Using the `TrackScreen` should be enough for us to know if the user is able to navigate to different screens

apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
```diff
@@ -147,6 +148,9 @@ export default function MainNavigator() {
           listeners={({ navigation }) => ({
             tabPress: e => {
               e.preventDefault();
+              track("web3hub_section_clicked", {
+                page: "Home",
+              });
```
Justkant, line 153: Again not sure we want to track this as we can just track the screen instead, we never did track this kind of click for other screens and it doesn't tell you if you end up in the actual screen

.changeset/pink-walls-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Analytics improved on Web3Hub
```
Justkant, line 5: Probably more correct
```suggestion
Setup analytics on Web3Hub
```

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/ManifestsList/CategoriesList/index.tsx
```diff
@@ -21,7 +22,13 @@ const renderItem = ({
 }) => {
   return (
     <Badge
-      onPress={() => extraData?.selectCategory(item.id)}
+      onPress={() => {
+        track("web3hub_category_clicked", {
+          component: `CategoriesList`,
+          category: item.id,
+        });
```
Justkant, line 29: Not sure about this one too as we should be able to track the category selected with a `TrackScreen` and the proper props
It would give us the guarantee that we actually updated the state when the user clicks instead of just tracking a click

apps/ledger-live-mobile/src/newArch/features/Web3Hub/components/ManifestsList/index.tsx
```diff
@@ -54,6 +55,11 @@ export default function ManifestsList({
 
   const goToApp = useCallback(
     (manifestId: string) => {
+      track("web3hub_app_clicked", {
+        component: `ManifestsList`,
+        page: "Web3HubMain",
+        app: manifestId,
+      });
```
Justkant, line 62: What I said on the other track events that could be replaced by a simple `TrackScreen` could also be applied for this as it would gives us the full flow from each screen with minimal events on the dashboard

Justkant (commented):
You should try to use `TrackScreen` instead of manually adding `track` calls on every interaction
Using `TrackScreen` with the arguments like category or manifestId gives us the full user journey through the app already and we can reduce the number of events and `track` calls that we need to add in the code, also `TrackScreen` is used every where

It might still be good to keep some of the `track` calls for some interactions if we really want to track this but in general we should be able to do most of it with the `TrackScreen` as we did in the `screens/Platform/v2`

apps/ledger-live-mobile/src/components/RootNavigator/MainNavigator.tsx
RamyEB: Having TrackScreen on the main page would be called every time he switch from search to main or quiting an app ? If we just wanna track how many timesa user clicl on the Discover then might be better to implement this here

Justkant (approved):

Justkant (approved):

PR #7690, opened 2024-08-27
🐦‍🔥 [FIX]: Refresh ledger sync QR code on expiration

thesan:
### ✅ Checklist

  - ...

### 📝 Description

It retries the QR code processing flow when the WS closes unexpectedly. Since AFAIK there is no way to make sure the WS closed because of the QR code expiration: if the connection closes in less than 30 seconds the flow is not retried and the old error is shown. Here's a demo the behaviour with a min time 3 seconds:

https://github.com/user-attachments/assets/14c923dd-8da3-4e9c-a1f9-e3733958dbbe

I ended up using react query because I was getting some weird behaviour with the old logic. However I didn't manage to use the build in `retry` option because the `delay` option didn't work for me it kept on retrying after +30 seconds instead of immediatley 🤷.

### ❓ Context

- [LIVE-13766]

---

[LIVE-13766]: https://ledgerhq.atlassian.net/browse/LIVE-13766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useQRCode.ts
+++ b/apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useQRCode.ts
@@ -1,6 +1,11 @@
 import { useCallback, useState } from "react";
 import { createQRCodeHostInstance } from "@ledgerhq/trustchain/qrcode/index";
-import { InvalidDigitsError, NoTrustchainInitialized } from "@ledgerhq/trustchain/errors";
+import {
+  InvalidDigitsError,
+  NoTrustchainInitialized,
+  QRCodeWSClosed,
+} from "@ledgerhq/trustchain/errors";
+import { MemberCredentials } from "@ledgerhq/trustchain/types";
 import { useDispatch, useSelector } from "react-redux";
 import { setFlow, setQrCodePinCode } from "~/renderer/actions/walletSync";
 import { Flow, Step } from "~/renderer/reducers/walletSync";
@@ -12,9 +17,11 @@ import {
 import { useTrustchainSdk } from "./useTrustchainSdk";
 import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
 import getWalletSyncEnvironmentParams from "@ledgerhq/live-common/walletSync/getEnvironmentParams";
-import { useQueryClient } from "@tanstack/react-query";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
 import { QueryKey } from "./type.hooks";
 import { useInstanceName } from "./useInstanceName";
+
+const MIN_TIME_TO_REFRESH = 30_000;
 
 export function useQRCode() {
   const queryClient = useQueryClient();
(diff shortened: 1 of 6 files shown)


mcayuelas-ledger:
At the end we have an error message?

thesan:
> At the end we have an error message?

@mcayuelas-ledger the error message won't show even if you stayed several days on the QR code page. Every time the QR code expire a new WS connection will created and the QR code will be refreshed. But if the WS connection closes in less than 30 seconds something else is wrong so the "WS closed prematurely" error would be shown. I wanted to avoid to constantly refresh the WS connection when something actually is wrong with it (and also potentially IP ban users).

(In the screen recording I lowered the 30 seconds to 3).

mcayuelas-ledger (dismissed):

apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useQRCode.ts
```diff
@@ -26,73 +33,76 @@ export function useQRCode() {
...
+      if (e instanceof QRCodeWSClosed) {
+        const { time } = e as unknown as { time: number };
+        if (time >= MIN_TIME_TO_REFRESH) startQRCodeProcessing();
+      }
+      if (e instanceof InvalidDigitsError) {
+        dispatch(setFlow({ flow: Flow.Synchronize, step: Step.PinCodeError }));
+      }
+      if (e instanceof NoTrustchainInitialized) {
+        dispatch(setFlow({ flow: Flow.Synchronize, step: Step.UnbackedError }));
+      }
```
mcayuelas-ledger, line 77: we need to add 

`  setError(e);
        throw e;` also no? Like before

apps/ledger-live-mobile/src/newArch/features/WalletSync/hooks/useQRCodeHost.ts
```diff
@@ -36,77 +43,68 @@ export function useQRCodeHost({ setCurrentStep, currentStep, currentOption }: Pr
...
+        const { time } = e as unknown as { time: number };
+        if (time >= MIN_TIME_TO_REFRESH) startQRCodeProcessing();
+      }
+      if (e instanceof InvalidDigitsError) {
+        setCurrentStep(Steps.SyncError);
+      }
+      if (e instanceof NoTrustchainInitialized) {
+        setCurrentStep(Steps.UnbackedError);
+      }
+    },
```
mcayuelas-ledger, line 102: same here?

apps/ledger-live-desktop/src/newArch/features/WalletSync/hooks/useQRCode.ts
thesan: The `setError` is not needed because react-query manages the error (that's why I removed the `useState`).

For the `throw` I don't think it's needed either. From my understanding it was here to avoid the `then` that follows. Or am I missing something ?

mcayuelas-ledger (approved):
LGTM !

PR #7453, opened 2024-07-29
[LIVE-11435]: Remove list apps v2 feature flag

ofreyssinet-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  - list apps v2 feature flag doesn't exist anymore, v2 is now the default
  - list apps v1 code is gone

### 📝 Description

- Removing list apps v2 feature flag
- Removing list apps v1 related code
- List apps v2 is now the only implementation

### ❓ Context

- **JIRA or GitHub link**: [LIVE-11435] 

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

[LIVE-11435]: https://ledgerhq.atlassian.net/browse/LIVE-11435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7693, opened 2024-08-28
[LIVE-13826] Fix access to My Ledger when listApps APDU not supported

ofreyssinet-ledger:
### ✅ Checklist

  - LLM+LLD: Access to My Ledger

### 📝 Description

On old firmwares, the `listApps` APDU is not supported (for instance on Nano X 1.2.4).
Prior to #7453, we were not using that APDU, but always using the HSM to request the list of apps to the device. Sending the `listApps` APDU directly is an optimisation, as it avoids network calls, and it's doable in case the user has already allowed a secure connection on their device. But it fails for old firmwares.

So the solution chosen here, to not have a logic that depends on versions, is to first try to send the APDU, and if it fails, we proceed with the HSM method.

#### Before:

https://github.com/user-attachments/assets/d4925a32-5acd-4c97-904d-0fba0022ac5f

#### After:

https://github.com/user-attachments/assets/ffaa7afc-72b2-4341-93c7-0cb30b501180

### ❓ Context

- **JIRA or GitHub link**: [LIVE-13826] 

---

[LIVE-13826]: https://ledgerhq.atlassian.net/browse/LIVE-13826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/thick-fans-give.md
+++ b/.changeset/thick-fans-give.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Fix listApps: fallback to HSM script runner if listApps APDU is not available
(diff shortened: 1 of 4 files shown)


valpinkman (approved):
👌

PR #7723, opened 2024-09-02
feat: dynamic CAL fetch using API calls

CremaFR:
### ✅ Checklist

  - ...

### 📝 Description

CAL API call to fetch exchange currency information. 
Check if we can clean up more stuff

### ❓ Context

[- **JIRA or GitHub link**: ](https://ledgerhq.atlassian.net/browse/LIVE-13875)

---


lambertkevin (dismissed):
Just a small suggestion for readability, but other than that LGTM 👍

libs/ledger-live-common/src/exchange/providers/swap.ts
```diff
@@ -213,6 +225,34 @@ export const getProvidersData = async () => {
...
+  currencyId: string,
+  type: "tokens" | "coins",
+): Promise<CurrencyData> => {
+  const currencyData = (await network({
+    url:
+      `https://crypto-assets-service.api.ledger.com/v1/${type}` +
+      "?output=id,exchange_app_config_serialized,exchange_app_signature" +
+      `&id=${currencyId}`,
+  })) as { data: CurrencyDataResponse };
+  // throw errors/warning if many currencies fetched with the same ID, can it happen?
```
beths-ledger, line 245: I notice that this comment doesn't do what it says it will do and throw a warning if length > 1. Do we want to? 

And out of interest, do we know why the response is an array if only one object is possible? (E.g. can this endpoint can take `currencyId: string[]`, too? Or should it really _not_ be an array in response?)

CremaFR: Yeah this comment is more a TODO: xxx I'll update it. 
And this point is used to returned all currencies and we can filter them based on chains, ids ... so array is required. It is just by filtering with IDs we assume only one result but I remember that on some edge cases (renaming and other) it could happen ids are not unique

libs/ledger-live-common/src/exchange/providers/swap.ts
```diff
@@ -213,6 +225,34 @@ export const getProvidersData = async () => {
...
+export const findExchangeCurrencyData = async (
+  currencyId: string,
+  type: "tokens" | "coins",
+): Promise<CurrencyData> => {
+  const currencyData = (await network({
+    url:
+      `https://crypto-assets-service.api.ledger.com/v1/${type}` +
+      "?output=id,exchange_app_config_serialized,exchange_app_signature" +
+      `&id=${currencyId}`,
+  })) as { data: CurrencyDataResponse };
```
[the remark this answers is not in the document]
CremaFR: Done thanks for the recommandation

libs/ledger-live-common/src/exchange/providers/swap.ts
```diff
@@ -203,14 +231,47 @@ function transformData(providersData) {
...
+    method: "GET",
+    url: `https://crypto-assets-service.api.ledger.com/v1/${type}`,
+    params: {
+      output: "id,exchange_app_config_serialized,exchange_app_signature",
+      id: currencyId,
+    },
+  });
+  // TODO: maybe throw errors/warning if many currencies fetched with the same ID.
+  if (!currencyData.length) {
+    throw new Error(`Currency data not found for id ${currencyId}`);
```
cpellizzari-ledger, line 268: Old message is "Exchange, missing configuration for ..."
https://github.com/LedgerHQ/ledger-live/blob/f0721f22c6f1bb662806299415c3a0bcb333db7c/libs/ledger-live-common/src/exchange/index.ts#L60C22-L60C57
-> please add proper i18n support for this error
-> use old message for english

Also, yes, let's have another error is there are 2 currencies with same id. Should not happen because on CAL backend, ids are supposed to be unique. But in case it happens, it's better to throw an error than "just use the first one".

CremaFR: Added check to throw error for many ids founds. But no i18n this is LLC

libs/ledger-live-common/src/exchange/providers/swap.ts
```diff
@@ -203,14 +231,50 @@ function transformData(providersData) {
...
+};
+
+/**
+ * Retrieves the currency data for a given ID
+ * coins and tokens are litteraly the same on our case but we need to call the correct endpoint.
+ * @param currencyId The unique identifier for the currency.
+ * @param type Specifies the type of currency, either "tokens" or "coins".
+ * @returns A promise that resolves to the currency data including ID, serialized config, and signature.
+ */
+export const findExchangeCurrencyData = async (
```
jnicoulaud-ledger, line 254: There is a `/currencies` that unifies both: https://crypto-assets-service.api.ledger.com/v1/currencies?id=bitcoin&output=id,exchange_app_config_serialized,exchange_app_signature

cpellizzari-ledger (approved):

PR #7778, opened 2024-09-09
fix: 🔁 device selection after aborted flow

thesan:
### ✅ Checklist

  - ...

### 📝 Description

**I'm opened to alternative solutions.**

TLDR: I'm not completely satisfied with this solution but I don't see another way which doesn't involve a major refactor.

The issue comes from the current structure of `DeviceActionModal` which relies on the `device` prop and `result` internal state to trigger the drawer. Once `result` is set the drawer is not supposed to re-opened and since `device` is not necessary reset something like this doesn't always work:
```js
useEffect(() => {
  if (!device) return () => setResult(null);
}, [device]);
```
(also it might break some other flow unrelated to ledger sync).

The way I see it selecting a device should imperatively trigger the device interaction. But this component is used this way everywhere else there is a device interaction on LLM.

https://github.com/user-attachments/assets/30f1e63d-fbae-4d7c-899a-d4767ce1c0d9

### ❓ Context

- [LIVE-13937]

---

[LIVE-13937]: https://ledgerhq.atlassian.net/browse/LIVE-13937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/DeviceSelection/index.tsx
+++ b/apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/DeviceSelection/index.tsx
@@ -52,6 +52,8 @@ const WalletSyncActivationDeviceSelection: React.FC<ChooseDeviceProps> = ({
 
   const navigation = useNavigation<NavigationProps["navigation"]>();
 
+  const onClose = () => selectDevice(null);
+
   const onResult = useCallback(
     (payload: AppResult) => {
       goToFollowInstructions(payload.device);
@@ -116,7 +118,7 @@ const WalletSyncActivationDeviceSelection: React.FC<ChooseDeviceProps> = ({
         />
       </Flex>
       <DeviceActionModal
-        onClose={() => selectDevice(null)}
+        onClose={onClose}
         device={device}
         onResult={onResult}
         action={action}
--- a/apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/DeviceSelection/index.tsx
+++ b/apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/DeviceSelection/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState, memo } from "react";
+import React, { useCallback, useState, memo, useRef } from "react";
 import { useIsFocused, useNavigation } from "@react-navigation/native";
 import { Trans } from "react-i18next";
 import { Device } from "@ledgerhq/live-common/hw/actions/types";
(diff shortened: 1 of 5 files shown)


apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/DeviceSelection/index.tsx
```diff
@@ -116,18 +110,19 @@ const WalletSyncActivationDeviceSelection: React.FC<ChooseDeviceProps> = ({
...
         <SelectDevice2
-          onSelect={onSelectDevice}
+          onSelect={selectDevice}
           stopBleScanning={!!device || !isFocused}
           requestToSetHeaderOptions={requestToSetHeaderOptions}
         />
       </Flex>
       <DeviceActionModal
-        onClose={onClose}
+        onClose={() => selectDevice(null)}
```
themooneer, line 119: can you replace this with a handler before the render

apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/DeviceSelection/index.tsx
```diff
@@ -71,7 +65,7 @@ const WalletSyncActivationDeviceSelection: React.FC<ChooseDeviceProps> = ({
     // and avoids a duplicated error drawers/messages.
     // The only drawback: the user has to select again their device once the bluetooth requirements are respected.
     if (error instanceof BluetoothRequired) {
-      setDevice(undefined);
+      selectDevice(undefined);
```
themooneer, line 68: why not null instead of undefined ?

thesan: I'm not sure TBH... I don't fully understand the comment above:
```
// Both the old (SelectDevice) and new (SelectDevice2) device selection components handle the bluetooth requirements with a hook + bottom drawer.
// By setting back the device to `undefined` it gives back the responsibilities to those select components to check for the bluetooth requirements
// and avoids a duplicated error drawers/messages.
// The only drawback: the user has to select again their device once the bluetooth requirements are respected.
```
I didn't take the time to dive into the `SelectDevice2` component logic but since the comment seems to insist on `undefined` I kept it.

themooneer: okay

themooneer (approved):

PR #7836, opened 2024-09-18
:bug:(llm): avoid blink on unsynch LLM

LucasWerey:
### ✅ Checklist

  - LS

### 📝 Description

To avoid blinking when an LLM instance has been removed a loader will now ensure that the user is or not inside a trustchain. 

It works well only in build version on real device. In dev the real device is too slow to update and still blinks
To do this I've used what I received from tanStack

Normal flow : 
https://github.com/user-attachments/assets/bc0cc7a2-111d-4875-8669-a2273862907f

User unsynch : 
https://github.com/user-attachments/assets/447b233c-d3ff-4c1f-9f30-0c938252f350

### ❓ Context

- **JIRA or GitHub link**: [LIVE-14085](https://ledgerhq.atlassian.net/browse/LIVE-14085) & [LIVE-14086](https://ledgerhq.atlassian.net/browse/LIVE-14086) & [LIVE-14087](https://ledgerhq.atlassian.net/browse/LIVE-14087) 

---


apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/Manage/index.tsx
```diff
@@ -23,13 +23,22 @@ import { TrustchainNotFound } from "@ledgerhq/trustchain/errors";
 
 const WalletSyncManage = () => {
   const { t } = useTranslation();
+  const [forcedTimeLoader, setForcedTimeLoader] = useState(true);
```
cgrellard-ledger, line 26: Can we extract the logic regarding this timer in a separate hook ?

thesan (commented):
I wrote 3 comments but it's more of one general suggestion about the mocking approach:
When I added the trustchain unit tests I felt like just parsing loads of hashes from real life transactions made it hard to understand what this data meant and as a result to understand what was tested. Instead I generated trustchain data in the code with virtual device and wrote easy to understand dummy strings whenever possible. It might not be as critical when testing the actual app UI but what do you think of this approach ?

apps/ledger-live-mobile/__mocks__/api/LedgerSync/authenticate.json
```diff
@@ -0,0 +1,8 @@
+{
+  "access_token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NiJ9.eyJpc3MiOiJ0cnVzdGNoYWluLWJhY2tlbmQuYXBpLmF3cy5zdGcubGRnLXRlY2guY29tIiwic3ViIjoiMDMyNjk2ZWJkN2MyNzZjYjIwODM4MmNjYTk2Mzg4YWJjNWYxNTBkNDE1YjU5OWY3OTgzZmYzNTc1YmNlNDQxNDAxIiwiYXVkIjpbIkNsb3VkU3luYyIsInRydXN0Y2hhaW4iXSwiZXhwIjoxNzI2NjY1MjUwLCJpYXQiOjE3MjY2NjQ5NDksImp0aSI6IjhkM2JiYTVmLTg1NGYtNDkyNy1iZDFjLWUwMzE1NjE4Y2VmMCIsInBlcm1pc3Npb25zIjp7fSwiZGV2aWNlIjp0cnVlLCJyZWZyZXNoRXhwaXJhdGlvbiI6IjIwMjQtMDktMThUMTQ6MDk6MTBaIn0.wUyEXD488WSbKbdOM7Wk33SjKMzm7VGP3nCGRfTYhgNkI6vNdPrEQnrwKo6uloWsVvL5BrNpBCkd_8GPf2eyZw",
+  "permissions": {
+    "000c9ec1a1ab774f7eaeff2b0d4ad695f1fa07ea28d33f5d34126cb1152d6d83f6": {
```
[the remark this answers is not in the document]
LucasWerey: I do agree that it would be easier to read but I don't know because we might want to test with real datas to ensure that works. Since these datas have been generated from the  web tools they are safe.

apps/ledger-live-mobile/src/newArch/features/WalletSync/screens/Manage/index.tsx
LucasWerey: Done ✅

apps/ledger-live-mobile/src/newArch/features/WalletSync/hooks/useCustomTimeOut.ts
mcayuelas-ledger, line 1: Don't we have already something like that in LLD or live-common?

LucasWerey: I've not find something similar so I guess no

apps/ledger-live-mobile/__mocks__/api/LedgerSync/authenticate.json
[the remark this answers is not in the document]
thesan: Right these are more integration tests so it's fine to go about it this way. I just wanted to point this out because I don't know whether there is some project convention on how to approach these mocks. But this approach is fine too 👍

cgrellard-ledger (approved):

PR #7908, opened 2024-09-25
✨(lld): add details drawer for inscriptions

LucasWerey:
### ✅ Checklist

  - Ordis btc account

### 📝 Description

Add the detail drawer for inscriptions
Add utxos icons inside the rows

Since I don't have the utxos rarity directly inside inscription. I needed to group the utxo object and the corresponding inscription object using location and nft_id. (nft-react) 
This way we avoid to do a lot of API calls

Hide and Actions are not yet implemented 

The demo has been done with mocked address to be sure to have many ordinals. There is some visual glitch due to the record. Also I've tested it with an account with a lot of ordinals (about one thousand) and all looks good 

The mock has been updated because I did overwrite it before but it doesn't fit now

https://github.com/user-attachments/assets/9cc79bea-7567-48a3-a4ca-d74e500e1eb8

### ❓ Context

- **JIRA or GitHub link**: [LIVE-12483](https://ledgerhq.atlassian.net/browse/LIVE-12483) 

---

[LIVE-12483]: https://ledgerhq.atlassian.net/browse/LIVE-12483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/newArch/features/Collectibles/Ordinals/components/Inscriptions/DetailsDrawer/index.tsx
```diff
@@ -1,18 +1,66 @@
...
+    contentType={data.contentType}
+    collectionName={data.collectionName}
+    details={data.details}
+    previewUri={data.previewUri}
+    originalUri={data.originalUri}
+    isPanAndZoomOpen={data.isPanAndZoomOpen}
+    mediaType={data.mediaType}
+    tags={data.tags}
+    useFallback={data.useFallback}
+    tokenId={data.tokenId}
```
mcayuelas-ledger, line 34: {...data} will be cleaner :)

apps/ledger-live-desktop/src/newArch/features/Collectibles/utils/createInscriptionDetailsArrays.ts
mcayuelas-ledger, line 1: You can imagine maybe something more readable

like 

```

const createDetail = ({
    key,
    title,
    value,
    isCopyable = false,
    isHash = false,
  }: {
    key: string;
    title: string;
    value?: string | number;
    isCopyable?: boolean;
    isHash?: boolean;
  }) => {
    if (value !== undefined && value !== null) {
      return {
        key,
        title,
        value: String(value),
        isCopyable,
        isHash,
      };
    }
    return null;
  };

  return [
    createDetail({
      key: "Description",
      title: "ordinals.inscriptions.detailsDrawer.description",
      value: inscription.collection.description,
    }),
    createDetail({
      key: "Collection Name",
      title: "ordinals.inscriptions.detailsDrawer.collectionName",
      value: inscription.collection.name,
    }),
    ...]
```

apps/ledger-live-desktop/src/newArch/features/Collectibles/Ordinals/components/Inscriptions/DetailsDrawer/index.tsx
themooneer: I've a small remark about this review:  @mcayuelas-ledger  / @LucasWerey 
spreading data assumes that all the object properties are expected from `DetailDrawer` component's api (or contract), if you do so  then you should strictly type the data props and why not type-check (with zod)  so that you guarantee that everything spread satisfies the component props and to avoid that any changes (for example accidently injecting `dangerouslySetInnerHTML` prop) on the passed `data`props will be handled. 
In the meanwhile, there is a rule called commonly used in react code guidelines `react/jsx-props-no-spreading`, this rule don't recommend making spread inside the jsx , instead you can deconstruct the data object before the return operation. With this approach 
- it's more readable (i don't need to access to the component and do reverse engineering to understand what's passed by data and what's not)
- You make sure what to extract from data. (exactly like you ve done, but with less props chaining access operation via data `data.prop`.

Both ways leads to the same result no doubt.

apps/ledger-live-desktop/src/newArch/features/Collectibles/Ordinals/components/Inscriptions/DetailsDrawer/index.tsx
```diff
@@ -1,18 +1,66 @@
...
+    openCollectiblesPanAndZoom={data.openCollectiblesPanAndZoom}
+    setUseFallback={data.setUseFallback}
+  >
+    <DetailDrawer.Actions>
+      <Flex>
+        <Button
+          outlineGrey
+          style={{ flex: 1, justifyContent: "center" }}
+          my={4}
+          onClick={() => console.log("hide " + data.collectibleName + " clicked")}
```
themooneer, line 47: if there is no significant handler here, delete this console.log and make a TODO comment

ypolishchuk-ledger (approved):

apps/ledger-live-desktop/src/newArch/features/Collectibles/Ordinals/components/Inscriptions/DetailsDrawer/index.tsx
LucasWerey: I've reverted to previous version without the spread operator

apps/ledger-live-desktop/src/newArch/features/Collectibles/Ordinals/components/Inscriptions/DetailsDrawer/Actions.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { t } from "i18next";
```
cgrellard-ledger, line 2: Get it from your component like this instead : `const { t } = useTranslation();`

LucasWerey: ✅

cgrellard-ledger (approved):

PR #7933, opened 2024-09-27
🧑‍💻(chore):migrate unimported to knip for LLD

LucasWerey:
### ✅ Checklist

- [ ] `npx changeset` was attached. NOT NEEDED
  - Unimported -> Knip on LLD
  - Proto

### 📝 Description

> [!IMPORTANT]  
> MIGRATION FROM UNIMPORTED TO KNIP
> this is LLD only for the moment
> the idea is to migrate every projects one by one to knip to have a single config for the whole project. We can still have custom config for each lib if needed

config schema https://unpkg.com/knip@5.30.6/schema.json
config doc : https://knip.dev/reference/configuration

run `pnpm knip-check --filter=./apps/ledger-live-desktop`

knip recommends @types/node but we use deprecated function so we can't use it. Knip still works without 

many rules have turned off because idk which one could be useful in our case

LLD CI check : 
https://github.com/LedgerHQ/ledger-live/actions/runs/11213445955/job/31166600892?pr=7933

KNIP Config has been customized to ensure that it remains useful but not overly burdensome :
image

| image|image (the unused file is local not pushed)|

### ❓ Context

- **JIRA or GitHub link**: 

---


themooneer (approved):
Very nice one 👍

.github/workflows/test-desktop-external-reusable.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           name: lint-desktop-external
           path: ${{ github.workspace }}/apps/ledger-live-desktop/lint-desktop-external.json
       - name: check for dead code
-        run: pnpm desktop unimported
+        run: pnpm knip-check --filter=./apps/ledger-live-desktop
```
valpinkman, line 58:
```suggestion
        run: pnpm desktop knip-check
```

apps/ledger-live-desktop/package.json
```diff
@@ -48,8 +48,8 @@
     "test:playwright:clean": "git clean -fdX tests/artifacts",
     "typecheck": "node scripts/typecheck.js",
     "check": "pnpm lint; pnpm typecheck",
-    "unimported": "unimported",
-    "assets:replace": "zx ./scripts/replace-assets.mjs"
+    "assets:replace": "zx ./scripts/replace-assets.mjs",
+    "knip-check": "pnpm knip --directory ../.. --workspace apps/ledger-live-desktop"
```
valpinkman, line 52: [ASK] is it necessary to use workspace tracking here since you are already in the correct workspace ?

apps/ledger-live-mobile/package.json
```diff
@@ -68,8 +68,9 @@
     "download-hermes-profile": "zx ./scripts/download-hermes-profile.mjs",
     "e2e:loadConfig": "ts-node ./e2e/bridge/start-server.ts",
     "check": "pnpm lint; pnpm typecheck",
+    "assets:replace": "zx ./scripts/replace-assets.mjs",
     "unimported": "unimported",
-    "assets:replace": "zx ./scripts/replace-assets.mjs"
+    "knip-check": "pnpm knip --directory ../.. --workspace apps/ledger-live-mobile"
```
valpinkman, line 73: [ASK] Same as desktop, do you really need to specify the workspace since you are already in it

valpinkman (changes requested):
Very nice PR, I can see `unkip` looks much more efficient than `unimported`

I do have a few suggestion I've left as comment, but otherwise good job

apps/ledger-live-desktop/package.json
LucasWerey: Yep need that otherwise knip doesn't understand in which workspace we are and will trigger the test for every lib. In the docs they mentioned that the wanted workspace must be in the cli argument to trigger it correctly

.github/workflows/test-desktop-external-reusable.yml
LucasWerey: Done ✅

apps/ledger-live-desktop/package.json
LucasWerey: The --directory allows us to use the common file directory by running the command from this directory. For example, if we want to keep a customized knip configuration in a specific library, we won't need this option. And --workspace specifies which entry the test should take in the configuration

Wozacosta (dismissed):
🔥 🚀

apps/ledger-live-desktop/package.json
valpinkman: Gotchu 👍

LucasWerey: Hmm not sure about that because it's using pnpm-workspace.yaml (I've tried it doesn't work).IMO the only way would be to have a config for every lib that can extend from a baseConfig (so we don't centralise everything at root using knip workspace as it's done here). But if we want to keep the workspace part we need to use these arguments. I can at least reduce them with -W and -D.
It only depends on how we want to implement it actually we can keep the logic to have a config file for each lib that extends a knip.config.js with only the rules but no entry point etc or we keep the workspace part. A cool thing is that we can mix both

valpinkman: Alright, makes sense. 
As a note, all the other tools in the monorepo work the way I described (a based config at root + extend in each project).
This is non blocking however and good to go this way, I just wanted to understand the options we had.

Thanks for the informations 🙏

valpinkman (approved):

liviuciulinaru (approved):

themooneer (dismissed):

PR #7949, opened 2024-09-30
Send nightly iOS builds to new "Ledger Live (nightly)" app in Testflight

angusbayley:
### ✅ Checklist

  - nightly iOS builds will go to a separate app "Ledger Live (nightly/staging) in Testflight"

### 📝 Description

Creates a new configuration in Xcode for nightly builds, with a new associated bundle id. The iOS `build` lane is updated so CI nightly builds use this config when calling `gym()`. New app icons (same as production except red instead of black) are added and linked to this configuration.

Nightly builds build appear in Testflight under this new app, called "Ledger Live (nightly/staging)":

Screenshot 2024-09-30 at 17 39 32

The icon set has been updated since the previous screenshot. Latest is:

image

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-13485

---


apps/ledger-live-mobile/fastlane/Fastfile
```diff
@@ -151,7 +163,7 @@ platform :ios do
 
       build_number = latest_testflight_build_number(
         version: trim_version_number(package["version"]),
-        app_identifier: "com.ledger.live"
+        app_identifier: ENV["APP_IDENTIFIER"]
```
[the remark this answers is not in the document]
angusbayley: `APP_IDENTIFIER` is already defined in every .env file, so all good 👍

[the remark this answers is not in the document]
themooneer: in your local `.env` yes, but on the ci one i am not sure about it.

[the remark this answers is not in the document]
angusbayley: I just looked into this to check my understanding. Those `.env`s are checked into git and are present during the CI runs. Fastlane loads these files directly during runs using the `--env` command, and every pnpm script we have for the `:build` lane specifies an `--env` file. So I don't think there's a need for a default.

What I am concerned about though is that if this is called with `--env staging`, we load the staging env file which has `"com.ledger.live.dev"` as the `APP_IDENTIFIER`. Something seems wrong from the outset that we were overriding this value before this change, but nevertheless I'm going to look into the implications of disabling this override/loading the app identifier from the .env file.

themooneer (dismissed):

apps/ledger-live-mobile/fastlane/Fastfile
```diff
@@ -130,6 +130,18 @@ platform :ios do
       keychain_password: ENV["CI_KEYCHAIN_PASSWORD"],
       git_basic_authorization: Base64.strict_encode64("#{ENV["GIT_REPO_USER"]}:#{ENV["GH_TOKEN"]}"),
     )
+    match(
+      type: "appstore",
+      app_identifier: ["com.ledger.live.nightly"],
```
themooneer, line 135: is it possible to make it like this 
`app_identifier: [ENV["APP_IDENTIFIER"]]`?

apps/ledger-live-mobile/scripts/mobile-env-md.mjs
```diff
@@ -14,7 +14,7 @@ function main() {
 }
 
 function gen() {
-  let md = `<!-- this file is generated by mobile-env-md.test.ts -->\n`;
+  let md = `<!-- this file is generated by mobile-env-md.mjs -->\n`;
```
angusbayley, line 17: unrelated fix - this reference was wrong, which meant extra time tracking down why this was causing build failures

apps/ledger-live-mobile/fastlane/Fastfile
angusbayley: this is a separate lane for syncing the profiles. We do want to hardcode this for every profile we want to sync, so we get the full range of profiles synced. (If we set it as `ENV["APP_IDENTIFIER"]` this would only retrieve 1 profile, the one for the current env at runtime of this code)

themooneer: ah ok, understood, so while it's a single app_identifier we can simply remove the array and do it like this `app_identifier: "com.ledger.live.nightly",` can we ?

apps/ledger-live-mobile/fastlane/Fastfile
```diff
@@ -151,7 +163,10 @@ platform :ios do
 
       build_number = latest_testflight_build_number(
         version: trim_version_number(package["version"]),
-        app_identifier: "com.ledger.live"
+        # we are moving to using ENV["APP_IDENTIFIER"] but staging + prerelease still override their own
+        # app identifier to publish to com.ledger.live . This will be removed once we are are fully publishing
+        # to separate apps
+        app_identifier: options[:nightly] ? ENV["APP_IDENTIFIER"] : "com.ledger.live"
```
themooneer, line 169: can we just put `ENV["APP_IDENTIFIER"]` here , and rely on their values in 
image

apps/ledger-live-mobile/fastlane/Fastfile
```diff
@@ -243,6 +258,9 @@ platform :ios do
         gym(
           scheme: PROJECT_NAME,
           workspace: XCODE_WORKSPACE,
+          # This should come from env files - as we move each release type to its own app we will
+          # move them away from "Release" and to their own .env-file defined configuration
+          configuration: options[:nightly] ? ENV["APP_CONFIGURATION"] : "Release",
```
themooneer, line 263: same question here, i think the best practice is to rely on a single source of truth which is .env files

apps/ledger-live-mobile/fastlane/Fastfile
```diff
@@ -279,9 +297,8 @@ platform :ios do
       begin
         pilot(
           skip_submission: false,
-          app_identifier: "com.ledger.live",
+          app_identifier: ENV["APP_IDENTIFIER"],
```
themooneer, line 300: i like this one 👍

apps/ledger-live-mobile/fastlane/Fastfile
angusbayley: Yes, this is definitely what I prefer and what I want to get to eventually. The problem is staging still:
- relies on a staging-specific provisioning profile, and it fetches the app identifier from the env file to locate this
- expects to push to the main app under com.ledger.live

So we have to have a hardcoded overwrite somewhere (this is what existed before this PR and I am simply reinstating here, but much more clearly) OR create a new env variable for `APP_IDENTIFIER_FOR_PROVISIONING_PROFILE_LOOKUP_ONLY`. The latter is also not a good long term solution IMO, and we are probably about to send staging builds to one of the new apps (e.g. Prerelease) in an upcoming PR so I don't view the code above as the final solution.

apps/ledger-live-mobile/fastlane/Fastfile
angusbayley: 👍 - see above

apps/ledger-live-mobile/fastlane/Fastfile
angusbayley: Yep, good point! I'll update

apps/ledger-live-mobile/fastlane/Fastfile
angusbayley: Actually I overlooked this one - this needs to adopt the pattern above

apps/ledger-live-mobile/fastlane/Fastfile
themooneer: ok ! we keep in mind these point for futur improvements.

themooneer (approved):

apps/ledger-live-mobile/fastlane/Fastfile
angusbayley: Gah - sorry ignore the comment above - this is inside the `if (options[:nightly])`, so we don't need to change it. The else clause below contains the hardcoded app identifier for all other release types.

IMO this Fastfile is quite confusing and could be cleaned up significantly, but I just want to be deliberate about what we are taking on in this PR.

apps/ledger-live-mobile/ios/ledgerlivemobile/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,182 +1,116 @@
...
+      "platform" : "ios",
+      "scale" : "3x",
+      "size" : "40x40"
     },
     {
-      "size" : "50x50",
-      "idiom" : "ipad",
-      "filename" : "Icon-Small-50x50@1x.png",
-      "scale" : "1x"
+      "filename" : "Icon-App-40x40@3x 1.png",
```
aboissiere-ledger, line 60: Weird filename. Is it an error? Also why are we modifying icons of the Prod app?

angusbayley: I'm not sure how this occurred, but I know Xcode makes lots of automatic updates in this directory whenever you adjust any icon settings. I did confirm that the icon set wasn't corrupted in this change, but as a precaution I have reverted all AppIcon changes for the main app - see latest commit

themooneer (approved):

PR #8059, opened 2024-10-09
Fix bug where flag stakePrograms was not checked into account screen,…

adammino-ledger:
### 📝 Description

This PR fixes a bug where the `stakePrograms` flag was not checked into the Account tab screen therefore even if a currency was disabled on it, it would show the `stake` entrypoint. 

There is also an additional check for bitcoin action on the stakePrograms to be sure it's not displayed anywhere if disabled.


apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
```diff
@@ -44,16 +49,20 @@ const AccountHeaderActions = ({ account, parentAccount }: Props) => {
...
-      key: "Stake",
-      icon: IconCoins,
-      label: t("accounts.contextMenu.yield"),
-      event: "button_clicked2",
-      eventProperties: {
-        button: "stake",
-      },
-      onClick: stakeOnClick,
-    },
+    ...(availableOnStake
```
beths-ledger, line 52: I guess technically we shouldn't need to check the flag here, too, if we check in the main screens/account file below?

beths-ledger (approved):
Looks like duplicated check, but that is better than no check

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
```diff
@@ -7,13 +7,18 @@ import { stakeDefaultTrack } from "~/renderer/screens/stake/constants";
...
 
 type Props = {
   account: BitcoinAccount | TokenAccount;
   parentAccount: BitcoinAccount | undefined | null;
 };
 
 const AccountHeaderActions = ({ account, parentAccount }: Props) => {
+  const stakeProgramsFeatureFlag = useFeature("stakePrograms");
+  const listFlag = stakeProgramsFeatureFlag?.params?.list ?? [];
+  const stakeProgramsEnabled = stakeProgramsFeatureFlag?.enabled ?? false;
```
LucasWerey, line 20:
```suggestion
const stakeProgramsEnabled = stakeProgramsFeatureFlag?.enabled
```

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
themooneer, line 52: it's better to move the condition on the return level

adammino-ledger: if i move the condition to the return level and there are more actions to be passed they will be skipped based on the `stakePrograms` flag, in this way you can add another action to the array and it would be independent.

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
```diff
@@ -7,13 +7,18 @@ import { stakeDefaultTrack } from "~/renderer/screens/stake/constants";
...
 import IconCoins from "~/renderer/icons/Coins";
+import useFeature from "@ledgerhq/live-common/featureFlags/useFeature";
 
 type Props = {
   account: BitcoinAccount | TokenAccount;
   parentAccount: BitcoinAccount | undefined | null;
 };
 
 const AccountHeaderActions = ({ account, parentAccount }: Props) => {
+  const stakeProgramsFeatureFlag = useFeature("stakePrograms");
```
LucasWerey, line 18: You can do something like that too to directly have enabled and params
```  
const { params, enabled: stakeProgramsEnabled } = useFeature("stakePrograms") || {};
```

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
adammino-ledger: i know i agree however there are lots of nested screens that display the stake/yield button, adding the check here too is to make sure no place has been missed on the codebase. (like this one that was missed for all the currencies!) , and considering the release timing i thought it would be good to cover all the possibilities (also screens that i don't know about) . Let me know if it doesn't sound fair though!

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
adammino-ledger: again i took this from the rest of the codebase, but let me know if you want me to change it

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
themooneer: ok, thanks for the clarification

themooneer (approved):

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountHeaderManageActions.ts
beths-ledger: imho it makes sense to force a boolean here instead of possible undefined. But open to correction.

hedi-edelbloute (approved):

ypolishchuk-ledger (approved):

PR #8069, opened 2024-10-10
feat(web3Hub): fix state management with jotai

Canestin:
### ✅ Checklist

#### Problem

Currently, the way we update the app storage causes issues because the data state isn't shared between instances.
Specifically, when we modify a variable with instance X, the state of instance Y isn't updated (as the state isn't shared). As a result, if two different instances update the app storage at the same time, only the last modification is applied, leading to potential data loss.

#### Solution

Ensure that the database state is shared across all instances. This way, a modification in instance Z will automatically reflect in instance Y, by using `Jotai` npm lib, a lightweight library that simplifies state management across an application.

https://github.com/user-attachments/assets/5b840552-b12b-46c2-8067-a9e91f866022

- **JIRA** : https://ledgerhq.atlassian.net/browse/LIVE-14442

---

### DIFF
--- a/apps/ledger-live-mobile/package.json
+++ b/apps/ledger-live-mobile/package.json
@@ -154,7 +154,7 @@
     "hoist-non-react-statics": "3.3.2",
     "i18next": "20.6.1",
     "invariant": "2.2.4",
-    "jotai": "^2.7.0",
+    "jotai": "^2.10.1",
     "json-rpc-2.0": "^0.2.19",
     "lodash": "4.17.21",
     "lottie-react-native": "^6.7.0",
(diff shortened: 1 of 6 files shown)


apps/ledger-live-mobile/package.json
```diff
@@ -154,6 +154,7 @@
     "hoist-non-react-statics": "3.3.2",
     "i18next": "20.6.1",
     "invariant": "2.2.4",
+    "jotai": "^2.7.0",
```
Justkant, line 157: Small detail but we could update jotai to the latest version, it includes some fixes

Justkant (approved):
LGTM

PR #8075, opened 2024-10-11
:goal_net:(lld): rework ui of node erros on send flow

LucasWerey:
### ✅ Checklist

  - Send flow

### 📝 Description

Rework of the UI of the node errors (after signing on device) to be more user friendly. Support url is custom depending of the error (if there is a dedicated page or not for the error)

|image|image|

TO test we can cut the network while signing on device (video not on last design but it's to illustrate how it works)

https://github.com/user-attachments/assets/d976a179-523f-42ff-acf0-31f9cca5df04

### ❓ Context

- **JIRA or GitHub link**: [LIVE-14067](https://ledgerhq.atlassian.net/browse/LIVE-14067)

---

[LIVE-14067]: https://ledgerhq.atlassian.net/browse/LIVE-14067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  column-gap: 8px;
+`;
+
+const NodeError = ({ currencyName, networkName, error }: Props) => {
+  const theme = useTheme();
+  const { t } = useTranslation();
+  const { handleExportLogs } = useExportLogs();
+  const { copyToClipboard } = useCopyToClipboard();
+
+  const customError = createSendFlowError(error);
```
thesan, line 34: Hey I'm still struggling with the UI on LLM but I saw that the LLM codebase relies on the error names to show the error messages straight from the translation file. It's a bit verbose but I think it's clever as it makes the messages easy to update. I'm not sure if it's the same in LLD but just in case I pushed a draft: [#8081](https://github.com/LedgerHQ/ledger-live/pull/8081/files#diff-86e7954ab3d823c424cf12e3320ccabf3be653f81a8d4ee89c3163d6ac9ddc78).

Basically I added a `UnknownSendFlowError` returned by default by [sendFlow.ts](https://github.com/LedgerHQ/ledger-live/pull/8081/files#diff-44f82b6515fd35cd9afdfb74b6a3139974c9c9ed27524e0b3e71528829a3af6c) then by passing this error to `GenericErrorView` and adding the translations it just show the correct error.

LucasWerey: There is the same on LLD but I'm not sure that it would be worth to add since we don't have specific message to display for each error. But we can discuss of it with product eventually if they want to add custom message in the future it will definitely be worth

apps/ledger-live-desktop/src/newArch/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,9 @@
+import { useCallback } from "react";
+
+export function useCopyToClipboard() {
```
thesan, line 3: I still haven't tested yet (I'll do it tomorrow morning). But in the mean time I noticed in my PR that tapping on the copy buttons but not getting any feedback is not great it feels like nothing actually happened. This is why the legacy buttons show a tick mark after clicking. Although it does not look very good it makes for a much better UX IMO.

As a result I implemented the LLM hook like this:
```ts
export function useCopyToClipboard(callback?: (text: string) => void) {
  const onCopyRef = useRef<() => void>();
  return (text: string) => {
    if (!onCopyRef.current) {
      onCopyRef.current = () => {
        Clipboard.setString(text);
        callback?.(text);
      };
    }
    return onCopyRef.current;
  };
}
```
This way a callback can be passed to the hook to trigger an animation or any kind of feedback.

The other difference is that it keeps the same reference for the returned function instead for the factory (the returned function is passed down the components but the factory is not) if the returned function reference changes because it's re-declared: memoize components will still re-render (so the `useCallback` on the factory is not useful).

LucasWerey: Yep definitely a good idea

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+          top={
+            <CircleWrapper size={72} color={color}>
+              <Icons.DeleteCircleFill size="L" color="error.c50" />
+            </CircleWrapper>
+          }
+        />
+      </Flex>
+      <Flex flexDirection="column" rowGap={16} width="100%" alignItems="flex-start">
+        <Flex alignItems="flex-start">
+          <Text> {t("errors.TransactionBroadcastError.needHelp")}</Text>
```
thesan, line 80: The "Need help ?" is visibly bigger than the rest of the page:
image
According to Figma it should be the same size.

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/HelpSection.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const InteractFlex = styled(Flex)`
+  &:hover {
+    cursor: pointer;
+  }
+  padding: 8px;
+  background-color: ${({ theme }) => theme.colors.palette.opacityDefault.c05};
+  border-radius: 8px;
+  column-gap: 8px;
+`;
```
thesan, line 18: The buttons are also visibly too big.
image

On Figma the font size should be 13px (so basically it should matches the other text font size):
image

As a result the size of the buttons and icons should be reduced too.

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+          <InteractIcon onClick={onShowMore}>
+            {isShowMore ? (
+              <Icons.ChevronDown color="neutral.c100" />
+            ) : (
+              <Icons.ChevronRight color="neutral.c100" />
+            )}
+          </InteractIcon>
+        </Flex>
+        {isShowMore && (
+          <Flex columnGap={1}>
```
thesan, line 90: For longer error messages the text is cut by the container overflow hidden:
image
I think a 2 column grid layout would solve this problem.

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/TechnicalErrorSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <Flex
+      flex={0.5}
+      alignItems="flex-start"
+      flexDirection="column"
+      bg="opacityDefault.c05"
+      borderRadius={8}
+      justifyContent="space-between"
+      p={2}
+      overflow="hidden"
```
thesan, line 33: In which case this wouldn't be needed anymore.

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+            <CircleWrapper size={72} color={color}>
+              <Icons.DeleteCircleFill size="L" color="error.c50" />
+            </CircleWrapper>
+          }
+        />
+      </Flex>
+      <Flex flexDirection="column" rowGap={16} width="100%" alignItems="flex-start">
+        <Flex alignItems="flex-start">
+          <Text> {t("errors.TransactionBroadcastError.needHelp")}</Text>
+          <InteractIcon onClick={onShowMore}>
```
thesan, line 81: Having to click on the chevron icon specifically is not ideal IMO. Maybe could the click event be on the whole `Flex alignItems="flex-start"` element ?

thesan (commented):
I finally tested it! Sorry for the delay.
It looks good overall but I think a few styling improvement would be nice.

libs/ledger-live-common/.unimportedrc.json
```diff
@@ -277,7 +277,7 @@
     "src/wallet-api/useDappLogic.ts",
     "src/wallet-api/types.ts",
     "src/wallet-api/version.ts",
-    "src/errors/sendFlow.ts"
+    "src/errors/transactionBroadcastErrors.ts"
```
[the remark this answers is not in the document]
LucasWerey: Needed because we don't import it inside common but outside so unimported failed

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
LucasWerey: Done ✅

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/HelpSection.tsx
LucasWerey: Done ✅

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
LucasWerey: Done ✅

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/TechnicalErrorSection.tsx
LucasWerey: removed

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/index.tsx
LucasWerey: Done ✅

apps/ledger-live-desktop/src/newArch/hooks/useCopyToClipboard.ts
LucasWerey: A check icon has been added and some hover and active bg also

apps/ledger-live-desktop/src/renderer/modals/Send/steps/Confirmation/NodeError/TechnicalErrorSection.tsx
```diff
@@ -0,0 +1,77 @@
...
+  column-gap: 8px;
+  align-items: center;
+
+  &:active {
+    background-color: ${({ theme }) => theme.colors.palette.opacityDefault.c20};
+  }
+`;
+
+const StyledText = styled(Text)`
+  word-break: break-all;
```
thesan, line 29: Why the word-break ?
image
That's just my opinion but I think the line break between network and error would look better.

LucasWerey: Could look weird with strange error message I think like long words this would go directly to next line

apps/ledger-live-desktop/src/newArch/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  return (text: string) => {
+    if (!onCopyRef.current) {
+      onCopyRef.current = () => {
+        navigator.clipboard.writeText(text).then(() => {
+          callback?.(text);
+        });
+      };
+    }
+    return onCopyRef.current;
```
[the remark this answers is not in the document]
LucasWerey: NP just changed

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
```diff
@@ -75,11 +77,14 @@ function StepConfirmation({
           currencyName={currencyName}
         />
         {signed ? (
-          <BroadcastErrorDisclaimer
-            title={<Trans i18nKey="send.steps.confirmation.broadcastError" />}
+          <NodeError
+            error={error}
+            currencyName={String(currencyName)}
+            networkName={String(mainAccount?.currency.name)}
```
thesan, line 83: In my understanding here `currencyName` and `mainAccount?.currency.name` are the same thing.
image

In LLM I used `currency.familly` for the network (eg. evm) and `currency.ticker` for the coin:
E.g:
image

WDYT ?

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -6374,6 +6374,16 @@
     },
     "PinNotSet": {
       "title": "PIN not set"
+    },
+    "TransactionBroadcastError": {
+      "title": "Transaction Broadcast Unsuccesful",
+      "description": "Your transaction failed to broadcast on the {{networkName}} network. Your {{currencyName}} have not be transferred and remain in your account.",
```
thesan, line 6380: I realized writting the prev comment that there's a typo in Figma !
```suggestion
      "description": "Your transaction failed to broadcast on the {{networkName}} network. Your {{currencyName}} have not been transferred and remain in your account.",
```

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
LucasWerey: hmmm not sure for the curency.familly evm is not a network it's based on Eth network. So for example we need to display on the Base/Scroll/... network

I've added the ticker too 
image

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
```diff
@@ -75,11 +87,14 @@ function StepConfirmation({
           currencyName={currencyName}
         />
         {signed ? (
-          <BroadcastErrorDisclaimer
-            title={<Trans i18nKey="send.steps.confirmation.broadcastError" />}
+          <NodeError
+            error={error}
+            ticker={ticker}
+            networkName={String(mainAccount?.currency.name)}
```
thesan, line 93: Got it I'll use the same data in llm.

I just have one last suggestion: as I wrote here https://github.com/LedgerHQ/ledger-live/pull/8081#issuecomment-2419967707 I'm passing the network and coin to `TransactionBroadcastError` to keep on re-using GenericErrorView. I made these field optionnal but maybe it would be more consistent to pass them in the here too:
```ts
error={createTransactionBroadcastError(err, {
  coin: ticker,
  network: String(mainAccount?.currency.name),
}
```
But only if you think it makes sense and it doesn't complicate things.

LucasWerey: Done ✅

apps/ledger-live-desktop/src/newArch/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,19 @@
...
+    navigator.clipboard.writeText(text).then(() => {
+      if (callback) {
+        callback(text);
+      }
+    });
+  }, [callback]);
+
+  return (text: string) => {
+    textRef.current = text;
+    copy();
```
thesan, line 17: That's kind of a details but here the `useCallback` on copy is still not used because the reference is not returned. It only exist in the returned closure where it's called. This is why I was suggesting to have copy be returned:
```suggestion
    return copy;
```
So it can be passed directly to the component:
```tsx
const handleCopyError = useCopyToClipboard(_text => ...);
...
<InteractFlex onClick={handleCopyError(error.message)}>{icon}</InteractFlex>
```
But again it's a detail and I don't think it should be a blocker. I'll retest and approve the PR if all goes well.

thesan (approved):

PR #8117, opened 2024-10-16
Support/pin pr workflow deps

angusbayley:
### 📝 Description

Pins dependencies for all PR workflows. That means all workflows triggered `on: pull_request` or `on: push`.

### ❓ Context

Ticket: https://ledgerhq.atlassian.net/browse/LIVE-14091

### DIFF
--- a/.github/workflows/build-desktop-external-reusable.yml
+++ b/.github/workflows/build-desktop-external-reusable.yml
@@ -19,7 +19,7 @@ jobs:
   build-desktop-app:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-22.04, windows-2022, macos-14]
     name: "Build Ledger Live Desktop"
     runs-on: ${{ matrix.os }}
     outputs:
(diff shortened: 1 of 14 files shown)


.github/workflows/build-desktop-external-reusable.yml
```diff
@@ -19,13 +19,18 @@ jobs:
   build-desktop-app:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-    name: "Build Ledger Live Desktop"
-    runs-on: ${{ matrix.os }}
```
angusbayley, line 24: this is a result of the workflows in `build-desktop-external-reusable.yml` and `build-desktop-reusable.yml` drifting out of sync, as they are partial copy-pasted duplicates of each other. Would love to remove as much duplication as possible (ideally delete the unused workflow)

KVNLS: I'm not sure. We don't want to use our internal runners for external contributions. So I think that ubuntu-latest is used for Github runner. Can you double check, please?

KVNLS (changes requested):

.github/workflows/build-desktop-external-reusable.yml
angusbayley: Good point - fixed

KVNLS (approved):
LGTM!

PR #8158, opened 2024-10-21
Pin LLM + LLD + root dependencies

angusbayley:
## 📝 Description

Pin dependency versions in all package.json files. This was done by:

- running `pnpm i` on latest `develop`
- running a script that, for every dependency listed in every package.json:
  - `pnpm list`s the current installed version in that particular directory
  - pins that exact version in that package.json

See the first commit of the PR for the script (it's removed in a later commit).

### Caveats:
- doesn't cover `/libs/*` (out of scope)
- [doesn't pin `@playwright/tests` in LLD](https://github.com/LedgerHQ/ledger-live/pull/8158/files#diff-e6a4af44a62c7b5a59d95834370b8bf55242f494cb31004eefa5931ac0f575f4L169) - pinning this caused tricky testing issues that I haven't yet found a solution for

## ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-14476


angusbayley:
Build https://github.com/LedgerHQ/ledger-live/actions/runs/11456967860/job/31876347210?pr=8158 was for just the root package.json, which failed only on test-libraries. @KVNLS has said this is an unrelated problem with libraries that will be fixed today. Expanding this test to include ledger live desktop and mobile

themooneer (approved):
LGTM

PR #8192, opened 2024-10-24
Pin dependencies in `apps/cli*` and `tools/*`

angusbayley:
### 📝 Description

pins dependencies for cli and tools

### ❓ Context

part of https://ledgerhq.atlassian.net/browse/LIVE-14476

---


docs/package.json
```diff
@@ -26,12 +26,12 @@
...
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react": "18.2.0",
+    "react-dom": "18.2.0"
   },
   "devDependencies": {
-    "@types/node": "^20.8.10",
+    "@types/node": "20.12.12",
     "eslint-config-next": "13.5.6",
-    "typescript": "^5.2.2"
```
themooneer, line 35: it's a version bump here

tests/dummy-live-app/package.json
```diff
@@ -25,11 +25,11 @@
     ]
   },
   "devDependencies": {
-    "@types/jest": "^27.0.1",
```
themooneer, line 28: i'm seeing a version bump here instead of pinning

docs/package.json
angusbayley: This is because when you run `pnpm install` at the root of the repo with the old rule in the package json (^5.2.2) version `5.4.3` is what you end up having installed. If you want to check, on @develop do

```
pnpm i 
cd ./docs/
pnpm list typescript # => 5.4.3
```

Pinning to the real installed version is less disruptive than bumping down a version IMO

tests/dummy-live-app/package.json
angusbayley: see comment above

themooneer: ok thank you

themooneer (approved):

PR #8211, opened 2024-10-25
fix(swap): install Ethereum app if at least one evm family in the swap and not already present in the deps [LIVE-14666]

Justkant:
### ✅ Checklist

- [ ] **Covered by automatic tests.** Tested manually the installation of the app
  - Swap eth (no regression expected as we only add the Eth app in the deps)

### 📝 Description

Install Ethereum app if at least one evm family in the swap and the app is not already present in the deps
Avalanche needs the Ethereum app for the swap but we reference the Avalanche app in the cryptoassets lib, meaning if we don't have the Ethereum app already installed we end up with an error

### ❓ Context

- **JIRA or GitHub link**: [LIVE-14666]

---

[LIVE-14666]: https://ledgerhq.atlassian.net/browse/LIVE-14666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/tiny-cobras-refuse.md
+++ b/.changeset/tiny-cobras-refuse.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+fix(swap): install Ethereum app if at least one evm family in the swap and not already present in the deps
--- a/libs/ledger-live-common/src/hw/actions/startExchange.ts
+++ b/libs/ledger-live-common/src/hw/actions/startExchange.ts
@@ -123,7 +123,7 @@ export const createAction = (
         ? getMainAccount(exchange.toAccount, exchange.toParentAccount)
         : null;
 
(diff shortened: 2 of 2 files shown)


sprohaszka-ledger (approved):
LGTM

ofreyssinet-ledger (approved):
It works, but it's a bit strange that we have to add family/chain specific logic there...

valpinkman (approved):

Justkant:
> It works, but it's a bit strange that we have to add family/chain specific logic there...

Yes we don't want this in the long term but it's a more generic solution than only doing this for `Avalanche`
The issue is that we cannot specify a different app for swap vs send/receive at the moment

PR #8244, opened 2024-10-29
[DNM] add changesets-enforce workflow

angusbayley:
Adds workflow to enforce changesets version bumps at the `major` or `minor` level only (`patch` will cause build failure).

[Ticket here](https://ledgerhq.atlassian.net/browse/LIVE-14704)

This works on an entire-repo-snapshot basis, so if there are `patch`es in @develop and you've branched off @develop, and your commit does't create an offending changeset, your PR will still fail even though it wasn't your fault (this is currently the case for all tests and code quality checks too - I'm just being explicit). What this does mean is when we merge this all builds will go red until the changeset history is cleared on next release.

Note that the workflow fails with a useful error message: 

image

The workflow is async, and typically takes 15-20s to run:

Screenshot 2024-10-31 at 11 47 18

The checkout step was optimised to only pull in what was needed for this workflow ([more details](https://ledger.slack.com/archives/C07KP3A8CNA/p1730223516050169)). This is a practice we should roll out to the rest of the workflows.

My only concern about this is it's adding yet *another* workflow at a time where there are too many workflows.

### DIFF
--- a/.github/workflows/changeset-enforce.yml
+++ b/.github/workflows/changeset-enforce.yml
@@ -1,0 +1,43 @@
+name: No patch level changesets
+on:
+  pull_request
+
+jobs:
+  check_changesets:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repository (feature branch)
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          fetch-depth: 200
+        
+      - name: Fetch develop branch
+        run: |
+          git fetch origin develop:develop --depth=1
+
+      - name: Verify develop branch
+        run: git branch --all
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 9.12.3
+
+      - name: Install changesets
+        run: pnpm add -g @changesets/cli@2.27.7
+
+      - name: Generate changeset status
+        run: pnpm changeset status --output=changeset-status.json
+
+      - name: Check for patch level changesets
+        run: |
+          PATCH_CHANGESET_IDS=$(jq -r '.changesets[] | select(.releases[].type == "patch") | .id' changeset-status.json)
+          if [ -n "$PATCH_CHANGESET_IDS" ]; then
+            echo "Patch changesets found:"
+            echo "$PATCH_CHANGESET_IDS"
+            echo "❌ Patch level changesets are reserved for hotfixes, use major or minor only"
+            exit 1
+          else
+            echo "✅ No patch level changesets found."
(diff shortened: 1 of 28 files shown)


.github/workflows/changeset-enforce.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Checkout repository (develop branch)
+        uses: actions/checkout@v4
+        with:
+          ref: develop
+          fetch-depth: 1
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 9.11.0
```
themooneer, line 24: the actual used version of pnpm in the code base is `9.12.3`

.github/workflows/changeset-enforce.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Install changesets
+        # run: pnpm install -w @changesets/cli --no-frozen-lockfile
+        run: pnpm add -g @changesets/cli
```
themooneer, line 28: i would recommend to pin this to avoid issues with latest versions of cli 
`pnpm add -g @changesets/cli@2.27.7`

.github/workflows/changeset-enforce.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          ref: develop
+          fetch-depth: 1
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 9.11.0
+
+      - name: Install changesets
+        # run: pnpm install -w @changesets/cli --no-frozen-lockfile
```
themooneer, line 27: to remove

.github/workflows/changeset-enforce.yml
```diff
@@ -0,0 +1,43 @@
...
+          ref: ${{ github.head_ref }}
+          fetch-depth: 1
+
+      - name: Checkout repository (develop branch)
+        uses: actions/checkout@v4
+        with:
+          ref: develop
+          fetch-depth: 1
+
+      - uses: pnpm/action-setup@v4
```
themooneer, line 22: can't we use the cached pnpm store instead of installing it on each and every workflow run ?

.github/workflows/changeset-enforce.yml
angusbayley: done

.github/workflows/changeset-enforce.yml
angusbayley: done

themooneer (approved):
LGTM

PR #8278, opened 2024-11-01
Optimise checkout action implementations

angusbayley:
[Ticket](https://ledgerhq.atlassian.net/browse/LIVE-14752)

Streamlines our use of the [checkout](https://github.com/actions/checkout) action so it runs as fast as possible in our workflows. Typically this is because we were checking out the entire repo with all branches and complete commit history when we only need the latest snapshot of one branch.

### PR build and test

(⭐ = on the critical path)

✅ ⭐ `turbo-affected-reusable.yml`
**before**: 50s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11632744890/job/32396487027?pr=8278))
**after**: 10s checkout (9s checkout feature branch, 1s fetch develop) ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11634052124/job/32400562112?pr=8278))
(**NOTE**: https://github.com/LedgerHQ/ledger-live/pull/8279 shaves a few extra seconds off on top of this. Not strictly part of the checkout process so different PR)

✅ `test-design-system-reusable.yml`
**before**: 52s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11666662762/job/32482186983?pr=8278#step:2:1))
**after**: 7s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11666851871/job/32482852973?pr=8278#step:2:1))

✅ `test-libs-reusable.yml`
**1. Test Libraries**
**before**: 48s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11666662762/job/32482188221?pr=8278#step:2:1))
**after**: 8s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11666851871/job/32482855073?pr=8278#step:2:1))

**2. Codecheck Libraries**
**before**: 48s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11666662762/job/32482188221?pr=8278#step:2:1))
**after**: 8s checkout ([workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11666851871/job/32482855073?pr=8278#step:2:1))

### Elsewhere

These aren't in PR build and test, so less important, but I've similarly shaved ~45s off the checkout from each of these

✅ `sonar.yml`
✅ `test-integration.yml`

## Doesn't include...
There are opportunities to take off a few more seconds from most of these by using `sparse-checkout` to only download the directories of interest. This is likely to save a further ~20-40% of the remaining checkout time. Given the checkout time is now typically ~10s this means 2-4s further saved, so it isn't a high priority and we should back to this once the other massive time sinks are addressed elsewhere in PR build and test.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -24,8 +24,6 @@ jobs:
     runs-on: [ledger-live-4xlarge]
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
       - name: Setup git user
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-git-user@develop
       - name: Setup the caches
--- a/.github/workflows/test-libs-reusable.yml
+++ b/.github/workflows/test-libs-reusable.yml
@@ -32,7 +32,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || github.sha }}
-          fetch-depth: 1
       - name: Setup the caches
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-caches@develop
         id: cache
@@ -79,7 +78,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || github.sha }}
-          fetch-depth: 0
       - name: Setup the toolchain
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-toolchain@develop
         id: toolchain
--- a/.github/workflows/test-libs-reusable.yml
+++ b/.github/workflows/test-libs-reusable.yml
@@ -78,6 +78,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || github.sha }}
+          fetch-depth: 2
       - name: Setup the toolchain
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-toolchain@develop
         id: toolchain
--- a/.github/workflows/test-libs-reusable.yml
+++ b/.github/workflows/test-libs-reusable.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || github.sha }}
-          fetch-depth: 0
+          fetch-depth: 1
       - name: Setup the caches
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-caches@develop
         id: cache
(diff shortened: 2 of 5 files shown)


.github/workflows/test-libs-reusable.yml
```diff
@@ -79,7 +78,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || github.sha }}
-          fetch-depth: 0
+          fetch-depth: 2
```
themooneer, line 81: What it the purpose of increasing the `--depth` from 0 to 2 commits from each remote branch ?

.github/workflows/sonar.yml
```diff
@@ -24,8 +24,6 @@ jobs:
     runs-on: [ledger-live-4xlarge]
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
```
themooneer, line 28: Question : by deleting the fetch depth here and in the below Workflows, you will pass to a complete shallow fetch of the remote no ? 
Maybe it should be passed to 1 , so that you limit it to 1 commit from the remote.

.github/workflows/test-libs-reusable.yml
angusbayley: 0 is a special value that means fetch the entire history (for every branch if `ref` not specified)
1 would mean just fetch the state at the latest commit, which is usually what we really want
In this case we want 2 (fetch state at latest commit, and the commit before that too) because [some of the steps in test libs](https://github.com/LedgerHQ/ledger-live/blob/develop/.github/workflows/test-libs-reusable.yml#L96) references the previous commit when doing the checks.

Full docs for `fetch-depth` is [here](https://github.com/actions/checkout?tab=readme-ov-file#usage).

.github/workflows/sonar.yml
angusbayley: `fetch-depth` defaults to 1, so that's essentially what I'm doing

themooneer: ok didn't know that

themooneer: thanks

.github/workflows/test-libs-reusable.yml
themooneer: clear, thanks

themooneer (approved):

PR #8294, opened 2024-11-04
LLD - force live apps cache busting

Wozacosta:
### Associated PRs

manifest-api: https://github.com/LedgerHQ/manifest-api/pull/586

### ✅ Checklist

  - ...

### Why? 

We need a simple way to force live apps to clean up (bust) their cache. This issue came into light following the [supply chain attack affecting 1inch](https://coinmarketcap.com/community/articles/6725e9641571bc0543b77da0/).

### Other solutions explored

- We could try to pass `cache=false` in the `<webview>`'s preference, not documented in `<webview>`, Chrome's `webview` accepts it as an option, but toggling it doesn't seem to change anything during my tests. 

- Pass extra headers when calling `loadURL`: https://github.com/electron/electron/blob/main/docs/api/web-contents.md#contentsloadurlurl-options and https://github.com/electron/electron/issues/20680, problem is, intially live-apps url are passed as url props, we could do something hacking and force `loadURL` once the liveapp has loaded, doesn't seem like an elegant solution. [More documentation](https://www.electronjs.org/docs/latest/api/webview-tag#webviewloadurlurl-options), [caching headers good article](https://simonhearne.com/2022/caching-header-best-practices/)

This last option didn't seem to work on a few live apps, [this is expected as the cache headers cannot be relied on](https://stackoverflow.com/a/76769421) (the ball would then in the court of the liveapp developers). 
Also, some cached data such as those handled by service workers follow their own rules and get reused. It adds a layer of complexity for us to be 100% sure an app is free of compromised file.

### 📝 Description LLD

What if we could force a live app to start from a clean state, much like going incognito in your browser or reinstalling Ledger Live. 
This is what this solution proposes

Electron has a concept of sessions, an isolated browser-like environment that manages cookies, cache, and other storage separately from the main application and/or other webviews.
https://www.electronjs.org/docs/latest/api/session

By default, all our webview's used the default session storage (a persisting storage).

You can set the session that a webview uses by passing the `partition` params ([doc](https://www.electronjs.org/docs/latest/api/webview-tag#partition)), changing the partition param effectively cleans up everything that could have been cached during the hack, even files handled by the service worker.

Doc: [To distinguish paritition and session](https://www.electronjs.org/docs/latest/api/browser-window): 

image

#### Why we want to persist

A session not starting with `persist:` won't be persisted across launches of Ledger Live. This is troublesome as many of our liveapps have settings in app, cookies banner that have to be accepted, etc.

image

#### What this PR introduces: 

If a manifest has a `cacheBustingId` set, the liveapp's webview will not use the default session and instead set it like this: 
`partition = 'persist:${idSlug}-${manifestCacheBustingParam}' `

### ❓ Context

- **JIRA or GitHub link**: LIVE-14770

### Tests: 

Validate 1inch resetting properly its cache.

image

Cache busting: 

image

Useful tool I've used also, https://www.lagado.com/tools/cache-test

- live app launched
- when relaunching app, cache not reset
- when changing cacheBustingId, cache is reset
- when relaunching the app, cache not reset

To launch the debugger as soon as the liveapp launches, add this statement: 

image

---

### DIFF
--- a/.changeset/popular-jobs-occur.md
+++ b/.changeset/popular-jobs-occur.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"@ledgerhq/live-common": patch
+---
+
+feat: cache busting liveapp webviews
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/PlatformAPIWebview.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/PlatformAPIWebview.tsx
@@ -41,7 +41,10 @@ import { walletSelector } from "~/renderer/reducers/wallet";
 
 export const PlatformAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
   ({ manifest, inputs = {}, onStateChange }, ref) => {
-    const { webviewState, webviewRef, webviewProps } = useWebviewState({ manifest, inputs }, ref);
+    const { webviewState, webviewRef, webviewProps, webviewPartition } = useWebviewState(
+      { manifest, inputs },
+      ref,
+    );
 
     const tracking = useMemo(
       () =>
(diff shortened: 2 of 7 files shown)


ComradeAERGO (approved):
LGTM

PR #8301, opened 2024-11-05
chore: align sonar code coverage file resolution/ignore with jest for…

themooneer:
… lld & llm

### ✅ Checklist

  - ...

### 📝 Description

While comparing Sonar Code Coverage and Jest LCOV report is not relevant, in this PR i fixed : 
- LCOV Path resolution 
- refine the sonar code coverage by ignoring almost same jest.config files from being scanned for unit-test report
- add merge lcov file as a step in the workflow and get rid of useless directory such as desktop-coverage and mobile-coverage generated previously while */coverage directory is already ignored by SCM. 

As result 
- LLD CC increased from 23,4% to 26,6% as the nearest pct comparing to LCOV line/line coverage. (5 to 10% gap because of additional sonar metrics) 
- LLM CC previously 15,1 to 15.7 

##### Reminder 

1-to-1 comparison between SonarQube and coverage report is not relevant
We sometimes have users reporting that the code coverage is different between SonarQube and the tool used to gather it.
The reason for this is most often because people are not comparing the same metrics.
SonarQube gets the covered lines from the coverage report given to the analyzer. Then it calculates all its coverage metrics from there and the executable lines also called lines to cover.
What is often compared is the Line Coverage, most often displayed by the external tool used to gather the covered lines, and what we define as Code Coverage, which is computed from the numbers extracted from the coverage report passed to the analyzer.

### ❓ Context

- **JIRA or GitHub link**: 
https://ledgerhq.atlassian.net/browse/LIVE-14731

---


.github/workflows/sonar.yml
```diff
@@ -52,8 +50,16 @@ jobs:
 
       - name: Generate Unit test coverage for LLD & LLM
         run: |
-          pnpm desktop test:jest:coverage --coverageDirectory=desktop-coverage
-          pnpm mobile test:jest:coverage --coverageDirectory=mobile-coverage
+          pnpm desktop test:jest:coverage --collectCoverage=true
+          pnpm mobile test:jest:coverage --collectCoverage=true
```
KVNLS, line 54: Do we need to use `--collectCoverage=true` as for both commands we use `--coverage` which is an alias to `collectCoverage=true`?

apps/ledger-live-desktop/jest.config.js
```diff
@@ -72,10 +72,14 @@ module.exports = {
     "!src/**/__integration__/**",
     "!src/**/__tests__/**",
   ],
-  coverageReporters: ["json", "lcov", "json-summary"],
+  collectCoverage: true,
```
KVNLS, line 75: I think that we might not want to collect coverage as default. When we develop a feature we just want to replay the test quickly. And we can control coverage collection through the cli. Is there a reason to force it in the config?

apps/ledger-live-mobile/jest.config.js
```diff
@@ -47,17 +47,21 @@ module.exports = {
   ],
   testPathIgnorePatterns: ["<rootDir>/node_modules/"],
   moduleDirectories: ["node_modules"],
+  collectCoverage: true,
```
KVNLS, line 50: Same for coverage as default, is it needed?

apps/ledger-live-desktop/jest.config.js
themooneer: yes ! just forgotten these lines

apps/ledger-live-mobile/jest.config.js
themooneer: yes sure

.github/workflows/sonar.yml
themooneer: you right, just added it to make sure unit test field in sonar take coverage per file .. just removed thanks

apps/ledger-live-mobile/jest.config.js
themooneer: deleted

KVNLS (approved):
LGTM! Thanks!

PR #8339, opened 2024-11-08
chore: fix redirecting to GITHUB_ENV in fastfile for ios build

themooneer:
### ✅ Checklist

  - ...

### 📝 Description
- fix build number setting to GITHUB_ENV
- The build number should be transited to slack , like android one. 
- To avoid ambiguous redirect and directory not found we need to use `tee` with a temporary variable to set the build number
image

#### Successfull build with a correct build number 
image

### ❓ Context

- **JIRA or GitHub link**: 

https://ledgerhq.atlassian.net/browse/LIVE-14346
---

### DIFF
--- a/apps/ledger-live-mobile/fastlane/Fastfile
+++ b/apps/ledger-live-mobile/fastlane/Fastfile
@@ -187,10 +187,11 @@ platform :ios do
 
       if (options[:ci])
         sh(
-          <<-SHELL
-          echo "IOS_BUILD_NUMBER=#{build_number+1}" >> $env:GITHUB_ENV
-          SHELL
-        )
+            <<-SHELL
+              temp_ios_build_number_assignment="IOS_BUILD_NUMBER=#{build_number+1}"
+              echo $temp_ios_build_number_assignment | tee -a $GITHUB_ENV
+            SHELL
+          )
       end
     end
 
--- a/.changeset/good-boxes-boil.md
+++ b/.changeset/good-boxes-boil.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+fix redirecting to GITHUB_ENV in fastfile for ios build and setting the build number correctly for slack notification


apps/ledger-live-mobile/fastlane/Fastfile
```diff
@@ -187,10 +187,11 @@ platform :ios do
 
       if (options[:ci])
         sh(
-          <<-SHELL
-           echo "IOS_BUILD_NUMBER=#{build_number+1}" >> "$GITHUB_ENV"
-          SHELL
-        )
+            <<-SHELL
+              new_build_number="IOS_BUILD_NUMBER=#{build_number+1}"
```
angusbayley, line 191: nit: the variable contains the _build number assignment statement_, not just the build number. I'd choose a different variable name for that

angusbayley, line 191: I also don't understand why this fixes the issue - are you able to explain?

angusbayley (commented):
2 comments in the code

apps/ledger-live-mobile/fastlane/Fastfile
themooneer: Yes sure ! so , the main problem is that the redirection to `>> env:$GITHUB_ENV` is  not injecting the build number in the right environment variable that will be used in the next steps for slack message notification. 
Previously the redirection was directly to `>> $GITHUB_ENV` , the problem with this is at fastlane runtime, it encounters an `Ambiguous Redirect error` which is the case of this [ticket](https://ledgerhq.atlassian.net/browse/LIVE-14473) . 
An official solution was to use the $GITHUB_ENV between `"$GITHUB_ENV"` like in the documentation of setting environment variables. The problem with that, is on fastlane runtime, it occures a `No such directory ...` issue.
So to workaround it, the solution (and the only one that i found related to this specific issue ) it to pass by a temporary variable, set the new build number and use the `tee` to append it at the end of the $GITHUB_ENV path the new build number which is a workable solution. Same issue in this [discussion](https://github.com/orgs/community/discussions/36746)

themooneer: answered here https://github.com/LedgerHQ/ledger-live/pull/8339#discussion_r1837708153

angusbayley: Hmm, I think one of us has misunderstood.

I'm saying `new_build_number` is not a build number, it's a build number assignment statement string

I suggest `new_build_number` ==> `new_build_number_assignment_cmd` or similar

angusbayley: Okay thanks!

themooneer: ah ok got you ! i've replaced it with `temp_ios_build_number_assignment`

angusbayley (approved):

PR #8264, opened 2024-10-31
[DNM] Post-"ghost"-release hotfix workflow

angusbayley:
Ticket [here](https://ledgerhq.atlassian.net/browse/LIVE-12917)

This PR implements the new hotfix process for targeting arbitrary past releases to issue a hotfix for.

# Demo

A demo of the entire new process is [here](https://ledger.slack.com/archives/C07KP3A8CNA/p1730724401574429).

# Change 1: Create hotfix branch from any release

This modifies the *existing* create hotfix workflow to accept parameters for `version` and `application`. Specifying these values allows you to create a hotfix branch at the point of any previous release

<kbd>
  Screenshot 2024-10-31 at 15 19 15
</kbd>

- `version` is optional, and if unspecified defaults to "latest" - a special keyword that bypasses the tag finding process (see below), instead creating the hotfix branch from `main` just as the workflow did before this PR
- `application` is mandatory and is implemented as a `choice` dropdown with LLM and LLD as the only two options

If `version` is specified it is reformatted into a tag using the specified `application`, e.g. version `2.91.0` + `LLD` will be reformatted into a the tag for the 2.91.0 release `@ledgerhq/live-desktop@2.91.0`. This is then used to checkout the repo prior to the remaining hotfix steps. This is all implemented in a new composite action in `./tools/actions/composites/generate-tag`.

Bonus: the version + application being targeted is appended to the hotfix PR title:

<kbd>
  Screenshot 2024-11-01 at 11 19 57
</kbd>

### Edge case behaviour:
- when someone deletes the default value in `version` and submits it as empty: it falls back to a ref of `main` ([test run](https://github.com/LedgerHQ/ledger-live/actions/runs/11616218105/job/32348639402#step:3:22))
- when `version` is a non-existent version: `git fetch` fails during the checkout action ([test run](https://github.com/LedgerHQ/ledger-live/actions/runs/11616331415/job/32349026190#step:4:63)).

# Change 2: Implement new hotfixing process in `release-prepare-hotfix`

The PR also:
- **enforces patch changesets only on the hotfix branch** during the `release-prepare-hotfix` workflow
  - [here's a test run](https://github.com/LedgerHQ/ledger-live/actions/runs/11629980897/job/32387985117#step:10:22) of this working as expected
  - to run this workflow you must now re-specify the `application` and `tag` to target. I don't love this, but haven't thought about a slick and low complexity way to handle this automatically yet
- **removes workflow steps for merging the hotfix branch back into `main` and `develop`**

[...]

PR #8342, opened 2024-11-11
release: make it so release-final can target the hotfix branch

angusbayley:
Ticket [here](https://ledgerhq.atlassian.net/browse/LIVE-14836)

This allows us to perform the release-final workflow targeting an arbitrary branch. It defaults to `main` as before, but if we specify `hotfix` instead this will cause us to run the build and release processes from the hotfix branch instead, as part of the [new hotflix process](https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/4710989838/LL+Incident+Recovery+-+Hotfix+in+all+cases).

This workflow is in `ledger-live` but kicks off the `release-desktop` + `release-mobile` workflows in `ledger-live-build`, which already take a ref/branch as an argument. This argument is used to specify the hotfix branch, and as such there are no changes to be made in the `ledger-live-build` repo.

This PR is based off https://github.com/LedgerHQ/ledger-live/pull/8264

### DIFF
--- a/.github/workflows/release-final.yml
+++ b/.github/workflows/release-final.yml
@@ -12,6 +12,10 @@ on:
           - LLD
           - LLM
           - ALL
+      ref:
+        description: "the ref (branch) to release from"
+        required: false
+        default: main
 
   workflow_run:
     workflows:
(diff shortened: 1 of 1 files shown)


themooneer (approved):
LGTM

PR #8347, opened 2024-11-12
fix: use correct refs for turbo-affected checkout + fetch

angusbayley:
No ticket - this fixes an issue [encountered](https://ledger.slack.com/archives/C03MHGMAMRQ/p1731403867731049?thread_ts=1731066157.692919&cid=C03MHGMAMRQ) in the latest release. `turbo-affected` errored because the head and base branches weren't correctly handled as variables - they were hardcoded as the PR head and `develop` during the `checkout` and `fetch` stages of `turbo-affected`. When it comes to release, the relevant branches are not the head branch and `develop`, they are the head branch and `main`. Because the `fetch` step hardcoded `develop` as the branch to fetch, `main` was not available for comparison when the turbo-affected script ran, causing an error.

This PR correctly handles the head and base branch names as variables, fixing the problem

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -15,7 +15,7 @@ jobs:
   determine-affected:
     name: "Turbo Affected"
     if: ${{!github.event.pull_request.head.repo.fork }}
-    uses: LedgerHQ/ledger-live/.github/workflows/turbo-affected-reusable.yml@develop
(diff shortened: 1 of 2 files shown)


beths-ledger (approved):

angusbayley:
This has had a 👍 , dismissed only by a commit that runs the feature branch turbo-affected as a temporary check. That commit's been reverted so I consider the review still good. Bypassing branch protections for the merge

PR #8406, opened 2024-11-19
force change changeset versions during release process

angusbayley:
- Automatically upgrade `patch` changesets to `minor` in releases
- Automatically downgrade `minor` and `major` changesets to `patch` in hotfixes

The `changeset` cli itself doesn't provide any ability to modify changeset levels, so this is done via a new action that I've written.

### DIFF
--- a/.github/workflows/release-create.yml
+++ b/.github/workflows/release-create.yml
@@ -44,6 +44,13 @@ jobs:
         run: |
           git add .
           git commit -m 'update sortByMarketcap snapshot' || echo "No changes in snapshot of sortByMarketcap.test.ts"
+      - name: Move patch updates to minor
+        # For more info about why we do this, see this doc:
+        # https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/4710989838/LL+Incident+Recovery+-+Hotfix+in+all+cases
+        uses: LedgerHQ/ledger-live/tools/actions/composites/adjust-changeset-level@support/automatic-ci-changeset-control
+        with:
+          from_level: patch
+          to_level: minor
       - name: enter prerelease mode
         run: pnpm changeset pre enter next
       - name: commit
--- a/tools/actions/composites/adjust-changeset-level/action.yml
+++ b/tools/actions/composites/adjust-changeset-level/action.yml
@@ -1,0 +1,37 @@
+name: Adjust Changeset Level
+description: This action adjusts the changeset level from one value to another.
+inputs:
+  from_level:
+    description: 'The changeset level to target'
+    required: true
+  to_level:
+    description: 'The desired changeset level for the target'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Validate input levels
+      shell: bash
+      run: |
+        valid_levels=("patch" "minor" "major")
+        if [[ ! " ${valid_levels[@]} " =~ " ${{ inputs.from_level }} " ]]; then
+          echo "Invalid from_level: ${inputs.from_level}. Must be one of ${valid_levels[@]}"
+          exit 1
+        fi
+        if [[ ! " ${valid_levels[@]} " =~ " ${{ inputs.to_level }} " ]]; then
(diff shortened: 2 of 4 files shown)


.github/workflows/release-create.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       NODE_OPTIONS: "--max-old-space-size=7168"
-      RELEASE_BRANCH: "release"
+      RELEASE_BRANCH: "release-angusbayley-TEST"
```
angusbayley, line 12: to revert

.github/workflows/release-create.yml
```diff
@@ -57,6 +66,6 @@ jobs:
       - name: push
         run: |
           git push origin $RELEASE_BRANCH
-          gh pr create --title ":rocket: Release ${{ steps.date.outputs.date }}" -F .github/templates/release.md --base main --head $RELEASE_BRANCH
+          gh pr create --title ":rocket: [NOT A REAL RELEASE - TEST ONLY, by @angusbayley] Release ${{ steps.date.outputs.date }}" -F .github/templates/release.md --base main --head $RELEASE_BRANCH
```
angusbayley, line 69: to revert before merge

tools/actions/composites/adjust-changeset-level/action.yml
```diff
@@ -0,0 +1,34 @@
...
+        if [[ ! " ${valid_levels[@]} " =~ " ${inputs.from_level} " ]]; then
+          echo "Invalid from_level: ${inputs.from_level}. Must be one of ${valid_levels[@]}"
+          exit 1
+        fi
+        if [[ ! " ${valid_levels[@]} " =~ " ${inputs.to_level} " ]]; then
+          echo "Invalid to_level: ${inputs.to_level}. Must be one of ${valid_levels[@]}"
+          exit 1
+        fi
+
+    - name: Adjust changeset level
```
angusbayley, line 26: Inviting comments on this. As you can see this is done by direct manipulation of the changeset files via `sed` in a shell script. `sed` targets the YAML front matter specifically and not the rest of the file, but this approach still makes the assumption that the format of the YAML front matter in the changeset file will not change to use the `level` in some different context. I think this is a very unlikely scenario and that this approach is fine for a solution that is not intended to last many years.

An alternative to is to do this in node, convert the YAML front matter into JS objects via a standard yaml parser and writer - this is the way the changesets library itself manipulates its changeset files and is more future-proof, but is slightly more code.

tools/actions/composites/adjust-changeset-level/action.yml
```diff
@@ -0,0 +1,34 @@
...
+      run: echo "Adjusting changeset level from ${{ inputs.from_level }} to ${{ inputs.to_level }}"
+        for file in .changeset/*.md; do
```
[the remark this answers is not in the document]
angusbayley: yep, I will do this and link to the workflow when done 👍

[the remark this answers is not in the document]
angusbayley: [Here is a new PR](https://github.com/LedgerHQ/ledger-live/pull/8449/files) based off this one, which adds a patch changeset that should be updated by the changes in this PR when the release workflow runs.

[Here is the release create branch workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/11976088855/job/33390956845#step:16:119) for that PR. A the line I have linked, you can see that the mentions of "patch" in the description are unchanged, while the level of the changeset has been upgraded from patch to minor.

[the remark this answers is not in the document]
themooneer: thanks ! sounds good for me

themooneer (approved):
LGTM

.github/workflows/release-create.yml
angusbayley: done

.github/workflows/release-create.yml
angusbayley: done

PR #8372, opened 2024-11-15
No more setup toolchain (Github runners) batch 1

angusbayley:
[Ticket](https://ledgerhq.atlassian.net/browse/LIVE-14750)

Migrates the first batch of workflows to `setup-caches` from `setup-toolchain`.
- test-design-system
- test-libs-reusable
- test-web-tools-resusable

To do this, we stay on the GH runners but force the proto download during `setup-caches`. This costs about 10s but overall doesn't impact run time that much as GH runners outperform ours in several I/O limited steps (e.g. checkout).

### workflow testing
https://github.com/LedgerHQ/ledger-live/pull/8474 shows this PR with all this branch's workflows forced to run on the CI (I decided to do this in a separate PR this time to prevent a repeat of the issue we had earlier this afternoon). That PR's base is this branch, so the diffs are useful

### performance
This is a win runtime-wise. The `setup-caches` step in Test Docs is just 10s in [this PR's run](https://github.com/LedgerHQ/ledger-live/actions/runs/12015389589/job/33493400780?pr=8372), vs 1m 51s in [another workflow run](https://github.com/LedgerHQ/ledger-live/actions/runs/12014200638/job/33489446653) without these changes applied. The switch does add ~20s to the install step that follows, giving a total saving of 1m 21s. This difference seems to be stable - randomly inspecting other jobs shows similar before/after differences for setup-caches (I've looked at about 5 before and 5 after so far).

PR #8483, opened 2024-11-26
No more setup toolchain (Github runners) batch 2

angusbayley:
Follows https://github.com/LedgerHQ/ledger-live/pull/8372

See description in there


.github/workflows/test-desktop-reusable.yml
```diff
@@ -82,10 +82,11 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || github.sha }}
-      - name: Setup the toolchain
-        id: toolchain
-        uses: LedgerHQ/ledger-live/tools/actions/composites/setup-toolchain@develop
+      - name: Setup the caches
```
themooneer, line 85: question: do we really need another setup cache here as it's already done in line 43 ? am i wrong ?

angusbayley: this is a different job - you can see if you expand the collapsed code between the 2 invocations

themooneer (approved):

PR #8526, opened 2024-11-28
feat(llm): 🗒️ log content card impression when 50% of the card is shown

thesan:
### ✅ Checklist

  - ...

### 📝 Description

Depending on the type of content card, use a mix of:
- React Native: `FlatList.onViewableItemsChanged`
- and a custom hook (which compare the element position to the screen position).

#### Demo:

1. Analytics log:

https://github.com/user-attachments/assets/4716ff2c-7995-41e5-9da6-b7a6138b28ec

2. Make things clearer with a console log and the category id and the cards id sliced from the 6 to the 12 char:

https://github.com/user-attachments/assets/6bfa4dfe-d99a-4075-b6a7-f0cfbe5a73e0

### ❓ Context

- [LIVE-14465]

---

[LIVE-14465]: https://ledgerhq.atlassian.net/browse/LIVE-14465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/newArch/features/DynamicContent/components/LogContentCardWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+  id: string;
+  children: React.ReactNode;
+};
+
+export default function LogContentCardWrapper({ id, children }: Props) {
+  const ref = useRef<View>(null);
+  const { logImpressionCard } = useDynamicContent();
+
+  useInViewContext(ref, ({ isInView }) => {
+    if (isInView) logImpressionCard(id);
```
LucasWerey, line 17: We may also need to add normal analytics too here

apps/ledger-live-mobile/src/screens/WalletCentricAsset/AssetDynamicContent.tsx
```diff
@@ -1,21 +1,17 @@
 import { Flex, SideImageCard } from "@ledgerhq/native-ui";
 import { CryptoOrTokenCurrency } from "@ledgerhq/types-cryptoassets";
-import React, { useCallback, useEffect } from "react";
+import React, { useCallback } from "react";
 import { Linking } from "react-native";
 import useDynamicContent from "~/dynamicContent/useDynamicContent";
+import LogContentCardWrapper from "~/newArch/features/DynamicContent/components/LogContentCardWrapper";
```
LucasWerey, line 6:
```suggestion
import LogContentCardWrapper from "LLM/features/DynamicContent/components/LogContentCardWrapper";
```

apps/ledger-live-mobile/src/newArch/features/DynamicContent/components/LogContentCardWrapper.tsx
thesan: Done!

apps/ledger-live-mobile/src/newArch/contexts/InViewContext/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const InViewContext = createContext<InViewContext>({});
+
+export function useInViewContext(
+  target: RefObject<View>,
+  onInViewUpdate: (entry: InViewEntry) => void,
+  deps: DependencyList = [],
+) {
+  const { addWatchedItem, removeWatchedItem } = useContext(InViewContext);
+  const onInViewUpdateCb = useCallback(onInViewUpdate, deps); // eslint-disable-line react-hooks/exhaustive-deps
```
LucasWerey, line 25: is the bypass really necessary ? We should avoid that

thesan: I might be missing another posibility but here's how I see it:

The way I see not to bypass the `react-hooks/exhaustive-deps` would be to go:
From version 1
```ts
// Declaration
function useInViewContext(
  target: RefObject<View>,
  onInViewUpdate: (entry: InViewEntry) => void,
  deps: DependencyList = [],
)
...
// Usage
useInViewContext(target, ({ isInView }) => {
  if (isInView) logImpressionCard(id);
}, [id]);
```
To version 2
```ts
// Declaration
function useInViewContext(
  target: RefObject<View>,
  onInViewUpdate: (entry: InViewEntry) => void
)
...
// Usage
useInViewContext(target, useCallback(({ isInView }) => {
  if (isInView) logImpressionCard(id);
}, [id]));
```
The problem I have with version 2 is that the typing is that the type signature doesn't make it clear that `onInViewUpdate` is a dependency and that as a result writing this is problematic:
```ts
useInViewContext(target, ({ isInView }) => {
  if (isInView) logImpressionCard(id);
});
```
(because the internal `useEffect` will rerender on every re-render).

In react we are used to pass functions to hooks directly (without wrapping them in `useCallback`. That's why I think that here forwarding the dependencies to the hook is better to avoid mistakes. However it means that the variable used to forward the dependencies can't be statically analized by `react-hooks/exhaustive-deps` which makes the bypass necessary.

That said I could add `useInViewContext` to the config `rules."react-hooks/exhaustive-deps".additionalHooks`. So I'll have to add `logImpressionCard` to all the deps.

thesan: Done: https://github.com/LedgerHQ/ledger-live/pull/8526/commits/d444580c06da41f1debd703a2ebc98e7eb97a226

apps/ledger-live-mobile/src/components/Carousel/index.tsx
```diff
@@ -48,13 +49,15 @@ const Carousel = () => {
...
-        <CarouselCard
-          key={cardProps.id + index}
-          id={cardProps.id}
-          cardProps={cardProps}
-          index={index}
-          width={cardsWidth}
-        />
+        <LogContentCardWrapper key={cardProps.id + index} id={cardProps.id}>
+          <CarouselCard
+            key={cardProps.id + index}
```
LucasWerey, line 54: I think you can remove this key as it's already used on the parent

LucasWerey (approved):

PR #8510, opened 2024-11-27
ci: remove setup-toolchain action completely

angusbayley:
Last part of [this ticket](https://ledgerhq.atlassian.net/browse/LIVE-14750)

Following the recent PRs migrating `setup-toolchain` invocations to `setup-caches`, there are no more references to `setup-toolchain` in either `ledger-live` or `ledger-live-build`

This PR removes it for good! 🥳

PR #8535, opened 2024-11-28
Revert "ci: remove setup-toolchain action completely"

angusbayley:
Reverts LedgerHQ/ledger-live#8510

Other PRs and branches still rely on setup-toolchain as they were created before the migration to setup caches. Revery now with the intention of re-merging the original PR at the end of the release cycle


themooneer (approved):
thanks

PR #8564, opened 2024-12-02
FIX/LIVE-15117: exchange.swap does not have up-to-date accounts list when attempting to swap with newly created account.

kallen-ledger:
### ✅ Checklist

  - ...

### 📝 Description
- When creating a new account on the swap live app interface.
- The user will attempt to swap "to" this new account.
- The user will then be unable to.

This was because the `exchangeHandlers` accounts param did not get updated when accounts were changed. Instead this PR proposes passing a getter function that will instead grab the users currents accounts when the function is run. 

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-15117

---

### DIFF
--- a/.changeset/famous-mirrors-carry.md
+++ b/.changeset/famous-mirrors-carry.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Add customPTXHandlers to list of deps for swap customHandlers
--- a/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/SwapWebViewDemo3.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/SwapWebViewDemo3.tsx
@@ -389,7 +389,7 @@ const SwapWebView = ({ manifest, liveAppUnavailable }: SwapWebProps) => {
       },
     }),
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
+    [customPTXHandlers],
   );
 
   const hashString = useMemo(


CremaFR (dismissed):
LGTM 🔥

liviuciulinaru (dismissed):

apps/ledger-live-desktop/src/renderer/init.tsx
Justkant, line 1: Is this change necessary ?
It doesn't seem correct to init the rest and not this in the function

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -1,21 +1,27 @@
...
 import { context } from "~/renderer/drawers/Provider";
 import WebviewErrorDrawer from "~/renderer/screens/exchange/Swap2/Form/WebviewErrorDrawer";
 import { platformAppDrawerStateSelector } from "~/renderer/reducers/UI";
+import { flattenAccountsSelector } from "~/renderer/reducers/accounts";
+import { store } from "~/renderer/init";
 
-export function usePTXCustomHandlers(manifest: WebviewProps["manifest"], accounts: AccountLike[]) {
+function getAccounts() {
+  return flattenAccountsSelector(store.getState());
+}
```
Justkant, line 22: It's quite strange to do that directly instead of using a the selector with `useSelector`
It's seems we're not addressing the root cause of the issue, a change to the accounts should be reflected to the handlers

Justkant (commented):
I feel like this solution is bypassing the issue instead of solving it

apps/ledger-live-desktop/src/renderer/init.tsx
```diff
@@ -55,6 +55,10 @@ import { registerTransportModules } from "~/renderer/live-common-setup";
 const rootNode = document.getElementById("react-root");
 const TAB_KEY = 9;
 
+export const store = createStore({
```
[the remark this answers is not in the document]
kallen-ledger: As far as I can tell this does not create any side effects and just moves the creation of the store outside of the init function to allow access to the store directly.

kallen-ledger:
@Justkant @KVNLS agree that it is not solving the core issue, will take a re-look at this and try and figure out why the accounts are not getting updated on call. Thank you for the feedback 🙏

CremaFR (approved):

Justkant (approved):

PR #8566, opened 2024-12-02
feat: Use listenToKnownDevices from DMK

valpinkman:
### ✅ Checklist

  - live-dmk package updated to trigger add and remove device event on listen
  - LLD useListenToHidDevices updated

### 📝 Description

Update `useListenToHidDevices` to be able to rely on the DMK `listenToKnownDevices()` when the feature flag is on.
Also update the way we register the transports so everything is available at any time (but the `open` on the **LiveDMKTransport** only works when FF is on)
Changed the default verbosity on the DMK Logger.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-508]

---

[DSDK-508]: https://ledgerhq.atlassian.net/browse/DSDK-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/hooks/useListenToHidDevices.ts
```diff
@@ -1,19 +1,29 @@
...
 import { addDevice, removeDevice, resetDevices } from "~/renderer/actions/devices";
-import { IPCTransport } from "../IPCTransport";
+import { overriddenFeatureFlagsSelector } from "~/renderer/reducers/settings";
+import { State } from "~/renderer/reducers";
 
 export const useListenToHidDevices = () => {
   const dispatch = useDispatch();
+  const store = useStore<State>();
+  const localOverrides = overriddenFeatureFlagsSelector(store.getState());
+  const ldmkFeatureFlag = getFeature({ key: "ldmkTransport", localOverrides });
```
ofreyssinet-ledger, line 16: [SHOULD] Here you can just do `useFeature("ldmkTransport")` since you're in a react context, with the FF provider & the store at the root of the app, the overrides will already be computed.
This logic was needed in `live-common-setup.ts` because it's not in a react context.

apps/ledger-live-desktop/src/renderer/hooks/useListenToHidDevices.ts
```diff
@@ -1,19 +1,29 @@
...
+
+  const deviceManagementKit = useDeviceManagementKit();
+
   useEffect(() => {
     let sub: Subscription;
-    function syncDevices() {
-      const devices: { [key: string]: boolean } = {};
+    const devices: { [key: string]: boolean } = {};
```
ofreyssinet-ledger, line 22: [COULD] As we saw this variable is unused, it can be removed.

ofreyssinet-ledger (dismissed):

PR #8575, opened 2024-12-03
turbo affected fix scope experiment - with `ref` back in

angusbayley:
This is https://github.com/LedgerHQ/ledger-live/pull/8573 except it reinstates the `ref` parameter that's currently in develop

Hypothesis is that this will mean the merge commit is NOT included, and this branch will have a different commit diff to https://github.com/LedgerHQ/ledger-live/pull/8573

### DIFF
--- a/.github/workflows/turbo-affected-reusable.yml
+++ b/.github/workflows/turbo-affected-reusable.yml
@@ -35,19 +35,9 @@ jobs:
       packages: ${{ steps.affected.outputs.packages }}
       paths: ${{ steps.affected.outputs.paths }}
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/turbo-affected-reusable.yml
```diff
@@ -37,6 +37,8 @@ jobs:
     steps:
       - name: Checkout head branch
         uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.head_branch }}
```
angusbayley, line 41: *excludes* the merge commit

therefore turbo affected will compare the head of this branch to the base branch

PR #8592, opened 2024-12-03
fix(CN-513): all docs=true suffixed links after zendesk migration

beths-ledger:
### ✅ Checklist

  - Navigate to the updated links. The links are no longer broken

### 📝 Description

See https://ledger.slack.com/archives/C02PQA777EU/p1732181750558089

Staking links such as https://support.ledger.com/article/360014339340-Earn-Cosmos-ATOM-staking-rewards-in-Ledger-Live?docs=true-zd?redirect=false
were broken in the zen desk migration. 

- Links are fixed, and any that suffered the same problem because of the query parameters `?docs=true` are also now directing to what appears to be the correct articles again.
- Also added --runInBand and maxWorkers 50% to mobile tests to see if we could improve reliability of tests. See https://jestjs.io/docs/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server
- This should mean that the CPUs do not max out, which is a common cause of random "timeout" fails on CI.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/CN-513

---

### DIFF
--- a/apps/ledger-live-mobile/e2e/jest.config.js
+++ b/apps/ledger-live-mobile/e2e/jest.config.js
@@ -21,7 +21,7 @@ const jestAllure2ReporterOptions = {
 
 module.exports = async () => ({
   rootDir: "..",
-  maxWorkers: "50%",
+  maxWorkers: process.env.CI ? 2 : 1,
   preset: "ts-jest",
   transform: {
     "^.+\\.(js|jsx)?$": "babel-jest",
--- a/apps/ledger-live-mobile/e2e/jest.config.js
+++ b/apps/ledger-live-mobile/e2e/jest.config.js
@@ -21,7 +21,7 @@ const jestAllure2ReporterOptions = {
 
 module.exports = async () => ({
   rootDir: "..",
-  maxWorkers: process.env.CI ? 2 : 1,
+  maxWorkers: "50%",
   preset: "ts-jest",
   transform: {
     "^.+\\.(js|jsx)?$": "babel-jest",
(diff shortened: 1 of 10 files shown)


LucasWerey (approved):

apps/ledger-live-mobile/e2e/jest.config.js
```diff
@@ -21,7 +21,7 @@ const jestAllure2ReporterOptions = {
 
 module.exports = async () => ({
   rootDir: "..",
-  maxWorkers: process.env.CI ? 2 : 1,
+  maxWorkers: "50%",
```
abdurrahman-ledger, line 24: This reduces emulator/simulator number to one on the CI. Are we sure we want this ?
Could also be the reason why test are more stable (less ressources used with only one device)

apps/ledger-live-mobile/scripts/e2e-ci.mjs
```diff
@@ -61,6 +62,7 @@ const test_android = async () => {
     --forceExit \\
     --headless \\
     --retries 1 \\
+    --runInBand \\
```
abdurrahman-ledger, line 65: Why applying this on Android also? As far as I know we do not have performance issue on Linux runners.

apps/ledger-live-mobile/e2e/jest.config.js
beths-ledger: Ah, interesting, I wonder why. So would you prefer `process.env.CI ? "50%" : 1,` ?

apps/ledger-live-mobile/scripts/e2e-ci.mjs
beths-ledger: interestingly the Android run failed for the first time if I only added it to the ios tests. It also shouldn't cause any issues to apply unilaterally, as far as I understand

ComradeAERGO (approved):

LucasWerey (approved):

PR #8621, opened 2024-12-06
fix: revalidate recipient field on account change in the send flow

themooneer:
### ✅ Checklist

  - ...

### 📝 Description
- When a user fill the recipient field and change account, the revalidation wasn't triggered. This PR will fix this issue.

See the following video 

<b> Before </b>

https://github.com/user-attachments/assets/1535fba6-ec9f-4a60-8ddc-9c889911de5c

Note: in the end of the video, i was forced to delete and re-paste the casper address so that it can be revalidated via the input onChange event.

<b> After </b>

https://github.com/user-attachments/assets/3a2219af-f61c-4e38-9265-4b1aea763e84

Note: here the revalidation is controlled by both input onChange event and watching the account.id change from the selector above.

### ❓ Context

- **JIRA or GitHub link**: 

https://ledgerhq.atlassian.net/browse/LIVE-14904
---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.tsx
@@ -48,6 +48,21 @@ const RecipientField = <T extends Transaction, TS extends TransactionStatus>({
     }
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+  useEffect(() => {
+    // if the account changed, we should revalidate the recipient field
+    if (!initValue && value !== "" && value !== transaction.recipient) {
+      onChangeTransaction(bridge.updateTransaction(transaction, { recipient: value }));
+    }
+  }, [
+    account.id,
+    value,
+    transaction.recipient,
+    bridge,
+    onChangeTransaction,
+    transaction,
+    initValue,
+  ]);
+
   const onChange = useCallback(
     async (recipient: string, maybeExtra?: OnChangeExtra | null) => {
       const { currency } = maybeExtra || {};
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.tsx
```diff
@@ -48,6 +48,21 @@ const RecipientField = <T extends Transaction, TS extends TransactionStatus>({
     }
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+  useEffect(() => {
```
mcayuelas-ledger, line 51: wdyt about combine both useEffect ? With something like that?

```ts
useEffect(() => {
    if (value !== "" && value !== transaction.recipient) {
      onChangeTransaction(bridge.updateTransaction(transaction, { recipient: value }));
      if (resetInitValue && initValue) {
        resetInitValue();
      }
    }
  }, [
    value,
    transaction.recipient,
    bridge,
    onChangeTransaction,
    transaction,
    initValue,
    resetInitValue,
    account.id,
  ]);
```

themooneer: i did before, it didn't work.  `if (resetInitValue && initValue)` is not a valid condition to reset the initValue. Also the first effect will run one single time at the component mount. What i need is a second layer that manages the change of account.id basically. I would rather ignore exhaustive deps, but the functional works

themooneer: what you recommend as a refactor is a better abstraction, but with an uncontrolled resetInitiallValue, it gaves me a different behaviour.

mcayuelas-ledger (approved):

PR #8623, opened 2024-12-06
fix: ock app on inactive/backgournd state of app (except deeplinking)

themooneer:
### ✅ Checklist

  - ...

### 📝 Description

https://github.com/user-attachments/assets/7e2cdb4c-f091-40ac-b800-55a0c19bae31

Note: the blurr behaviour when swipe up the app on ios is already in another error ticket. 

### ❓ Context

- **JIRA or GitHub link**: 

https://ledgerhq.atlassian.net/browse/LIVE-15065
---

### DIFF
--- a/apps/ledger-live-mobile/src/context/AuthPass/index.tsx
+++ b/apps/ledger-live-mobile/src/context/AuthPass/index.tsx
@@ -1,5 +1,5 @@
 import React, { PureComponent } from "react";
-import { StyleSheet, View, AppState, Platform } from "react-native";
+import { StyleSheet, View, AppState } from "react-native";
 import * as Keychain from "react-native-keychain";
 import type { TFunction } from "i18next";
 import { connect } from "react-redux";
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/context/AuthPass/index.tsx
```diff
@@ -110,10 +110,7 @@ class AuthPass extends PureComponent<Props, State> {
 
   // The state lifecycle differs between iOS and Android. This is to prevent FaceId from triggering an inactive state and looping.
   isBackgrounded = (appState: string) => {
-    const isAppInBackground =
-      Platform.OS === "ios" ? appState === "background" : appState.match(/inactive|background/);
-
-    return isAppInBackground;
+    return appState.match(/inactive|background/);
```
KVNLS, line 113: Are we sure we don't create other issues here? When I look at the comment it seems that it was done intentionally

themooneer: i've already seen that the development done before, is related to bypass the lock on deeplink redirection context which i've tested it (in the video) so, i don't think that it will cause another issue.

KVNLS: Thanks a lot, can you double-check when we open the camera please? I have seen issues related to it in the past. Such as scan a QR Code for Ledger Sync or scan a QR Code for the send flow

themooneer: hummmm nice recommendation ! i'll do it

themooneer: when i tested it the first time, it prompt me a camera permission, then when i accept it , the phone is locked. After that the access to the camera wasn't blocking , no issues, but i think it's problematic for BLE pairing for the first time (ios prompt make the app state inactif)
I should reconsider this implementation and look for another status. 
Thank you very much for this instructive recommendation @KVNLS

PR #8670, opened 2024-12-11
feat: migrate all entry screen, ctas to new add account v2

themooneer:
### ✅ Checklist

- [x] **Covered by integration test of AssetSelection in this scope.** 
  - ...

### 📝 Description
When llmNetworkBasedAddAccount feature flag is enabled, all the ctas that open an add account process will be redirected to the new flow and support currency and related route params.
Entry point list: 
- Swap
- Buy
- Earn/Stake
- Asset
- Wallet 
- Deeplinks
- Receive Flow

The new add account flow start either by a new Crypto selection  or via Network provider (AssetSelection feature). If those informations are provided in the route params then the next step is to Connect the device and Scan & Select found account (will be adjusted in the next ticket). 

<b style="color:red;"> Important </b>

- The scope of the ticket is the entry point, SelectCrypto or SelectNetwork when `currency` is provided in the route params.

---
##### Changes in ledger-live-common 
Here is my analysis on the subject!

It is the SelectCrypto component, during its mounting, that intercepts the currency from the params and uses this info in a useEffect to simulate an explicit redirection:

- To SelectNetwork: if the currency has at least one network provider (I will come back to this)
Otherwise,
- To SelectAccount: if a scan has already been done previously on this currency
Otherwise,
- To the SelectDevice process and scan & co

What happens during the mounting is that it intercepts ethereum as currency, calls a custom hook useGroupedCurrenciesByProvider, and triggers a useEffect to redirect if needed.

The problem is that the custom hook is slow since it depends on an asynchronous operation to return a 'result' containing 2 arrays:

- `sortedCryptoCurrencies`
- `currenciesByProvider`
The first execution returns 2 empty arrays.

From the second execution, the arrays are filled.

Hence, the `useEffect` is triggered twice on the SelectCrypto side,

The first without providers redirects to `SelectAccount`,
Then the second arrives a few seconds later, following the new return of the custom hook which changes the ref of the onPress function that triggers the `useEffect `a second time
Consequently, a redirection to SelectNetwork occurs. This results in this strange and ambiguous behavior from a UX point of view.
The root cause of all this is that the `useEffect `managing potential navigation depends on an onPress callback, which in turn depends on the result of the custom hook useGroupedCurrenciesByProvider, while we do not have the information that it is loading the data. It executes with the result of its initial state and then with the filled arrays.

On the add-account, I made sure to better control this custom hook with a loading system with 4 states: `idle`, `pending`, `success`, and `error`, so that we extract this information at the instantiation of the hook. 

Consequently, the useEffect only triggers if the `asynchronous operation` is finished (success OR error state).

- Example of issue 

https://github.com/user-attachments/assets/280ce5d3-d8be-4f4e-af43-0c4957a9e06f

- When fixed 

https://github.com/user-attachments/assets/4be46ec0-6320-4f5b-9dcc-7f4addc86666

#### Test  case revision is done 

- here ➡️ https://ledgerhq.atlassian.net/browse/LIVE-15374?focusedCommentId=504723
---
#### Full Video preview 

https://github.com/user-attachments/assets/5f5f5edc-27cd-4b99-b6c5-c0b6a51137f8

##### Post onboarding workflow test 

https://github.com/user-attachments/assets/273b4ba0-80b0-4777-ac4c-41098f4a43b5

### ❓ Context

- **JIRA or GitHub link**: 

https://ledgerhq.atlassian.net/browse/LIVE-15374
---


KVNLS:
We need to fix the issue with the Select Asset screen blink, please.
We can't keep it as is for the end user + it breaks navigation by adding this step.
Can you have a look on how to fix it, please?

CremaFR (dismissed):

libs/ledger-live-common/src/deposit/useGroupedCurrenciesByProvider.hook.ts
```diff
@@ -12,18 +12,32 @@ const initialResult: GroupedCurrencies = {
...
   );
 
   // Get mapped assets filtered by supported & sorted currencies, grouped by provider id
   useEffect(() => {
-    loadCurrenciesByProvider(coinsAndTokensSupported).then(setResult);
-  }, [coinsAndTokensSupported]);
-
-  return result;
+    if (withLoading) {
+      setLoadingStatus("pending");
```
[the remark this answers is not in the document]
themooneer: it's a string literal type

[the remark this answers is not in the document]
themooneer: i'll add an enum instead

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccount/components/SelectAddAccountMethod/useSelectAddAccountMethodViewModel.ts
```diff
@@ -29,10 +29,15 @@ const useSelectAddAccountMethodViewModel = ({
...
-  }, [hasCurrency, currency]);
+        ? {
+            token: currency,
+            ...(llmNetworkBasedAddAccountFlow?.enabled && { context: "addAccounts" }),
+          }
+        : { currency, ...(llmNetworkBasedAddAccountFlow?.enabled && { context: "addAccounts" }) }
+      : llmNetworkBasedAddAccountFlow?.enabled
+        ? { context: "addAccounts" }
+        : {};
+  }, [hasCurrency, currency, llmNetworkBasedAddAccountFlow?.enabled]);
```
mcayuelas-ledger, line 40: Could we please improve readability? 

Maybe use `if else` instead of ternary ?

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccount/types.ts
```diff
@@ -2,12 +2,16 @@ import { CryptoCurrency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
 import { ScreenName } from "~/const";
 import { Device } from "@ledgerhq/types-devices";
 
+type CommonParams = {
+  context?: "addAccounts" | "receiveFunds";
```
mcayuelas-ledger, line 6: Enum?

apps/ledger-live-mobile/src/newArch/features/DeviceSelection/types.ts
```diff
@@ -1,24 +1,39 @@
 import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
 import { AccountLike } from "@ledgerhq/types-live";
-import { ScreenName } from "~/const";
+import { NavigatorScreenParams } from "@react-navigation/core";
+import { NavigatorName, ScreenName } from "~/const";
+import { NetworkBasedAddAccountNavigator } from "../Accounts/screens/AddAccount/types";
+import { StackNavigatorProps } from "~/components/RootNavigator/types/helpers";
+
+type CommonParams = {
+  context?: "addAccounts" | "receiveFunds";
```
mcayuelas-ledger, line 9: same as above

mcayuelas-ledger: it's also duplicated `type CommonParams`

apps/ledger-live-mobile/src/screens/ReceiveFunds/01-SelectCrypto.tsx
```diff
@@ -50,7 +51,8 @@ export default function AddAccountsSelectCrypto({ navigation, route }: Props) {
   const { t } = useTranslation();
   const accounts = useSelector(flattenAccountsSelector);
 
-  const { currenciesByProvider, sortedCryptoCurrencies } = useGroupedCurrenciesByProvider();
+  const { currenciesByProvider, sortedCryptoCurrencies } =
+    useGroupedCurrenciesByProvider() as GroupedCurrencies;
```
mcayuelas-ledger, line 55: why do you need to force casting?

libs/ledger-live-common/src/deposit/type.ts
```diff
@@ -31,3 +31,8 @@ export type GroupedCurrencies = {
   currenciesByProvider: CurrenciesByProviderId[];
   sortedCryptoCurrencies: CryptoOrTokenCurrency[];
 };
+
+export type LoadingBasedGroupedCurrencies = {
+  result: GroupedCurrencies;
+  loadingStatus: "idle" | "pending" | "success" | "error";
```
mcayuelas-ledger, line 37: enum could be better ? :)

mcayuelas-ledger (changes requested):

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccount/types.ts
themooneer: string literal typing are more useful in this case for autocompletion. 
If you use an enum, typscript won't be able to infer the value of the context. 
Just used the enum in live-common as it's more adapted

apps/ledger-live-mobile/src/newArch/features/DeviceSelection/types.ts
themooneer: it seems duplicate but for the next tickets DeviceSelection feature screens will have more common props. 
they are common props related to DeviceSelection feature.

libs/ledger-live-common/src/deposit/type.ts
themooneer: done here only

themooneer:
@KVNLS fixed

apps/ledger-live-mobile/src/screens/ReceiveFunds/01-SelectCrypto.tsx
themooneer: **With the current implementation** simply because this "hook" has 2 return type with different object shape, so typescript won't be able to infer which one you are using. 
I'm not willing to add more refactoring to this `useGroupedCurrenciesByProvider` with generics as we will clean it up with `GroupedCurrencies`

libs/ledger-live-common/src/deposit/useGroupedCurrenciesByProvider.hook.ts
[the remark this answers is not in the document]
themooneer: done

CremaFR (approved):

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectCrypto/useSelectCryptoViewModel.ts
```diff
@@ -0,0 +1,137 @@
...
+import { LoadingBasedGroupedCurrencies } from "@ledgerhq/live-common/deposit/type";
+
+export default function useSelectCryptoViewModel({
+  context,
+  filterCurrencyIds,
+  paramsCurrency,
+}: {
+  context?: "addAccounts" | "receiveFunds";
+  filterCurrencyIds: string[] | undefined;
+  paramsCurrency: string | undefined;
```
mcayuelas-ledger, line 26:
```suggestion
  filterCurrencyIds?: string[];
  paramsCurrency?: string;
```

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectNetwork/index.tsx
```diff
@@ -182,29 +53,37 @@ export default function SelectNetwork({
...
-        <FlatList
-          testID="receive-header-step2-networks"
-          contentContainerStyle={styles.list}
-          data={sortedCryptoCurrenciesWithAccounts}
-          renderItem={renderItem}
-          keyExtractor={keyExtractor}
-          showsVerticalScrollIndicator={false}
-          keyboardDismissMode="on-drag"
-        />
+        {["success", "error"].includes(providersLoadingStatus) ? (
```
mcayuelas-ledger, line 72: why not use a type or enum here?

libs/ledger-live-common/src/deposit/useGroupedCurrenciesByProvider.hook.ts
```diff
@@ -12,18 +12,30 @@ const initialResult: GroupedCurrencies = {
...
+          setResult(data);
+          setLoadingStatus(LoadingStatus.Success);
+        })
+        .catch(() => setLoadingStatus(LoadingStatus.Error));
+    } else {
+      loadCurrenciesByProvider(coinsAndTokensSupported).then(setResult);
+    }
+  }, [coinsAndTokensSupported, withLoading]);
+  return withLoading ? { result, loadingStatus } : result;
 };
```
mcayuelas-ledger, line 41: You can use TanStack Query to do that ?

themooneer: within the scope of this PR i juste used the existing promise, sure we can do it with tanstack query, react query mutations or any other async operation library. but it's not the scope of this ticket.

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectNetwork/index.tsx
themooneer: `providersLoadingStatus` is already a literal string type with 4 variants

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectCrypto/useSelectCryptoViewModel.ts
themooneer:
```typescript
type SelectCryptoViewModelProps = Pick<CommonParams, "context"> & {
  filterCurrencyIds?: string[];
  paramsCurrency?: string;
};
```
Done like this

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectNetwork/index.tsx
themooneer: replaced the string by the corresponding enums.

themooneer:
image
Unit test for LLC are working fine even after purging turbo cache, probably it's a flaky test on ubuntu

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectNetwork/index.tsx
```diff
@@ -182,29 +54,37 @@ export default function SelectNetwork({
...
-        <FlatList
-          testID="receive-header-step2-networks"
-          contentContainerStyle={styles.list}
-          data={sortedCryptoCurrenciesWithAccounts}
-          renderItem={renderItem}
-          keyExtractor={keyExtractor}
-          showsVerticalScrollIndicator={false}
-          keyboardDismissMode="on-drag"
-        />
+        {[LoadingStatus.Success, LoadingStatus.Error].includes(providersLoadingStatus) ? (
```
mcayuelas-ledger, line 73: If it's "Error" what happens?

themooneer: it will show an empty list as ReceiveFlow Network component see `apps/ledger-live-mobile/src/screens/ReceiveFunds/02-SelectNetwork.tsx` 
The list intercept the array whether it's empty or not. I think it can be improved but outside of this ticket scope.
image

mcayuelas-ledger (approved):

PR #8714, opened 2024-12-17
B2CQA-2110 : Swapping with Send Max button

stual-ledger:
### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: 

---


apps/ledger-live-desktop/tests/specs/speculos/swap.spec.ts
```diff
@@ -262,6 +262,67 @@ for (const { swap, xrayTicket } of tooLowAmountForQuoteSwaps) {
   });
 }
 
+const swapWithSendMax = [
+  {
+    swap: new Swap(Account.ETH_1, Account.BTC_NATIVE_SEGWIT_1, "0", Fee.MEDIUM),
+    xrayTicket: "B2CQA-2110, QAA-292",
+  },
+  {
+    swap: new Swap(Account.ETH_1, Account.SOL_1, "0", Fee.MEDIUM),
```
VicAlbr, line 271: same

apps/ledger-live-desktop/tests/specs/speculos/swap.spec.ts
```diff
@@ -262,6 +262,67 @@ for (const { swap, xrayTicket } of tooLowAmountForQuoteSwaps) {
...
+    xrayTicket: "B2CQA-2110, QAA-292",
+  }
+];
+
+for (const { swap, xrayTicket } of swapWithSendMax) {
```
VicAlbr, line 276: Idk if we need to create a new test to check the send max maybe we can just modify existing e2e test and modify the function `fillInOriginCurrencyAmount` do detect when we want to do a sendmax.
You can do something like that:
```
  @step("Fill in amount: $1")
  async fillInOriginCurrencyAmount(electronApp: ElectronApplication, amount: string) {
    const [, webview] = electronApp.windows();
    if(amount == "sendMax"){
    toggle send max button
    } else {
    await webview.getByTestId(this.fromAccountAmoutInput).fill(amount);
    }
    //wait for potential origin amount error to be loaded
    await this.page.waitForTimeout(500);
  }
```

stual-ledger: I'm hesitant about doing this because this code is meant to be maintained and extended by QAs with less experience in automation, I'm not sure whether redundancy is not preferable in this case for the sake of readability

alalmi-ledger: +1 for readbility over making the most elegant code, seeing as this is meant to be maintained by functional QA's who are ramping up on automation skills

ypolishchuk-ledger: ok, to have a separate test instead of adding another entry to an already huge swaps array. It will indeed improve understanding when looking at the Allure report. However, duplicating almost entirely performSwapUntilQuoteSelectionStep doesn't help in reading and understanding code (making the swap file bigger), so maybe a middle-ground option here would be to remove last line from performSwapUntilQuoteSelectionStep so we can easily re-use it in both cases - with max send and without and then as a last step to call `await app.swap.sendMax(electronApp)` directly in your test

PR #8723, opened 2024-12-17
[Alpaca] Add block hash and timestamp

Salim-belkhir:
### ✅ Checklist

  - an extra field named `block` on the `Operation` object returned by the coin framework for `stellar`, `tezos` and `xrp`

### 📝 Description

Update `listOperations` for the coins module `stellar`, `tezos` and `xrp`. This change is needed because we need extra information about the block for each operation (the time and the hash of the block)

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/coin-framework/src/api/types.ts
+++ b/libs/coin-framework/src/api/types.ts
@@ -1,7 +1,7 @@
 export type BlockInfo = {
   height: number;
-  hash: string;
-  time: Date;
+  hash?: string;
+  time?: Date;
 };
 
 export type Operation = {
@@ -10,8 +10,7 @@ export type Operation = {
   type: string;
   value: bigint;
   fee: bigint;
-  blockHeight: number;
-  block?: BlockInfo;
+  block: BlockInfo;
   senders: string[];
   recipients: string[];
   date: Date;
--- a/libs/coin-framework/src/api/types.ts
+++ b/libs/coin-framework/src/api/types.ts
@@ -11,6 +11,7 @@ export type Operation = {
   value: bigint;
   fee: bigint;
   blockHeight: number;
+  block?: BlockInfo;
   senders: string[];
   recipients: string[];
   date: Date;
--- a/libs/coin-modules/coin-stellar/src/bridge/buildOptimisticOperation.ts
+++ b/libs/coin-modules/coin-stellar/src/bridge/buildOptimisticOperation.ts
@@ -28,6 +28,7 @@ export async function buildOptimisticOperation(
     transactionSequenceNumber: transactionSequenceNumber?.plus(1).toNumber(),
     extra: {
       ledgerOpType: type,
+      blockTime: new Date(),
     },
   };
 
--- a/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
@@ -13,11 +13,10 @@ export type Operation = {
   type: string;
   value: bigint;
   fee: bigint;
-  blockHeight: number;
-  block?: {
-    hash: string;
+  block: {
+    hash?: string;
     height: number;
-    time: Date;
+    time?: Date;
   };
   senders: string[];
   recipients: string[];
@@ -58,7 +57,6 @@ function convertOperation(
     value: BigInt(amount),
     // storageFee for transaction is always present
     fee: BigInt(storageFee ?? 0),
-    blockHeight: block.level,
     block: {
       hash: block.hash,
       height: block.level,
--- a/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
@@ -1,5 +1,6 @@
 import { tzkt } from "../network";
 import {
+  APIBlock,
   type APIDelegationType,
   type APITransactionType,
   isAPIDelegationType,
@@ -13,6 +14,11 @@ export type Operation = {
   value: bigint;
   fee: bigint;
   blockHeight: number;
+  block?: {
+    hash: string;
+    height: number;
+    time: Date;
+  };
   senders: string[];
   recipients: string[];
   date: Date;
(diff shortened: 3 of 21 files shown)


libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -24,17 +30,21 @@ export async function listOperations(
...
   const operations = await tzkt.getAccountOperations(address, { lastId, limit });
-  return [
+  const operationswithBlock = await Promise.all(
     operations
       .filter(op => isAPITransactionType(op) || isAPIDelegationType(op))
-      .reduce((acc, op) => acc.concat(convertOperation(address, op)), [] as Operation[]),
-    operations.slice(-1)[0].id,
-  ];
+      .map(async op => {
+        const block = await tzkt.getBlockByHash(op.block);
```
sprohaszka-ledger, line 37: This will trigger hundreds of request each time we want the `listOperations`.
There is no other way to retrieve the info?

libs/coin-framework/src/api/types.ts
```diff
@@ -11,6 +11,7 @@ export type Operation = {
   value: bigint;
   fee: bigint;
   blockHeight: number;
+  block?: BlockInfo;
```
sprohaszka-ledger, line 14: I wonder if it is not better to remove `blockHeight` and add: 
```js
block: BlockInfo
```
With an update of `BlockInfo`:
```js
export type BlockInfo = {
  height: number;
  hash?: string;
  time?: Date;
}
```

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -48,7 +58,12 @@ function convertOperation(
     value: BigInt(amount),
     // storageFee for transaction is always present
     fee: BigInt(storageFee ?? 0),
-    blockHeight: 0, // operation.block is a string
+    blockHeight: block.level,
+    block: {
+      hash: block.hash,
```
sprohaszka-ledger, line 63: By looking at [tztk doc](https://api.tzkt.io/#operation/Accounts_GetOperations), it seems you can have those `block` info, directly from the account request.
You may don't need to add extra calls. The Datamodel on our side is maybe incorrect.

libs/coin-modules/coin-xrp/src/network/index.ts
```diff
@@ -72,8 +72,8 @@ export const getTransactions = async (
 ): Promise<XrplOperation[]> => {
   const result = await rpcCall<AccountTxResponse>("account_tx", {
     account: address,
-    ledger_index: "validated",
```
sprohaszka-ledger, line 75: Why do you remove this parameter?

sprohaszka-ledger (dismissed):
I am not confident with the Tezos update. It will create hundreds of requests.
Also, we can remove extra `blockNumber` field and create a mandatory `block` field.

libs/coin-modules/coin-stellar/src/bridge/buildOptimisticOperation.ts
```diff
@@ -28,6 +28,7 @@ export async function buildOptimisticOperation(
     transactionSequenceNumber: transactionSequenceNumber?.plus(1).toNumber(),
     extra: {
       ledgerOpType: type,
+      blockTime: new Date(),
```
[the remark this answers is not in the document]
Salim-belkhir: Because of the changes that I have done on the type StellarOperation defined in the file [types/bridge.ts](https://github.com/LedgerHQ/ledger-live/blob/12e9475fabb55f65c4823e63f1c6d0a7b3ac4be0/libs/coin-modules/coin-stellar/src/types/bridge.ts#L105), I needed to update this file or an error of compilation appears. This type is used on the network too. Maybe I need to create a new type only for the network ?

libs/coin-framework/src/api/types.ts
Salim-belkhir: I thought about that too, but I don't know who use the coin-framework, so I decided to add an optional type to avoid changements for the actual users. And for the moment we need this changes only for `xrp`, `tezos` and `stellar`bin the lama adapter. If it is not more optional, I will need to update polkadot too. If it is necessary, I can make the changes

libs/coin-modules/coin-xrp/src/network/index.ts
Salim-belkhir: Thanks, I forget to add the good field. It was an error on the naming of this query param. ledger_index is an integer (the height) and not a string. To get only the ledger who are validated, we need to use validated param who is a boolean (see the docs [here](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_tx). I tried to call the node to get transaction with the ledger_index param but I get no transactions. I wonder who it worked before 🤔

libs/coin-framework/src/api/types.ts
sprohaszka-ledger: The `coin-framework/api` module is only used by Alpaca and the coin-module adapt for it. So you can safely update this :)

libs/coin-modules/coin-xrp/src/network/index.ts
Salim-belkhir: I am sorry, I made a mistake. The `validated` parameter is not needed anymore. See the docs :

> The account_tx method retrieves a list of validated transactions that involve a given account.

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -48,7 +58,12 @@ function convertOperation(
...
+      height: block.level,
+      time: new Date(block.timestamp),
+    },
```
Wozacosta, line 64:
```suggestion
    block: {
      hash: operation.block,
      height: operation.level,
      time: new Date(operation.timestamp),
    },
``` 


Shottr 2024-12-23 16 14 57

As per the doc and @sprohaszka-ledger, we can indeed rely only on the data found in the operation object

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
Wozacosta: As per the suggestion below, I'd also suggest removing this, unless we need other block-specific information?

For reference here's what the block info call returns

```
{
    "cycle": 802,
    "level": 7176273,
    "hash": "BMKnKkbSALea5sDqb3YdyRwGiEKT22Qm1Gw5ZKvKXBZLTpaEbjw",
    "timestamp": "2024-11-21T17:04:20Z",
    "proto": 20,
    "payloadRound": 0,
    "blockRound": 0,
    "validations": 6989,
    "deposit": 0,
    "rewardDelegated": 3038694,
    "rewardStakedOwn": 1402472,
    "rewardStakedEdge": 1055150,
    "rewardStakedShared": 0,
    "bonusDelegated": 3021927,
    "bonusStakedOwn": 1394734,
    "bonusStakedEdge": 1049327,
    "bonusStakedShared": 0,
    "fees": 179996,
    "nonceRevealed": false,
    "proposer": {
        "address": "tz3UC5hM9uCUiqmAw1pHdxjAdGueQMd8u3wf"
    },
    "producer": {
        "address": "tz3UC5hM9uCUiqmAw1pHdxjAdGueQMd8u3wf"
    },
    "software": {
        "date": "2024-11-14T23:12:25Z"
    },
    "lbToggleEma": 446596880,
    "aiToggleEma": 181960867,
    "rewardLiquid": 3038694,
    "bonusLiquid": 3021927,
    "reward": 5496316,
    "bonus": 5465988,
    "priority": 0,
    "baker": {
        "address": "tz3UC5hM9uCUiqmAw1pHdxjAdGueQMd8u3wf"
    },
    "lbEscapeVote": false,
    "lbEscapeEma": 446596880
}
```

Salim-belkhir: For the moment we need only the time, the hash and the height of the block, nothing else. So I removed this call like you suggested

Wozacosta (approved):
👍

PR #8766, opened 2024-12-24
Feat/live 15542 aptos send receive fix send funds simulation failed when amount to send is slightly above max spendable amount

may01:
### ✅ Checklist

  - ...

### 📝 Description
During simulation of transaction, if the tx amount is lower then account balance, but we don't have enough founds to cover fees, then we get the "Abort" error from the network. 
To avoid additional message during this errors the amount check is added for the unprocessed errors. 

### ❓ Context

- **JIRA or GitHub link**: [](https://ledgerhq.atlassian.net/browse/LIVE-15542)

---


libs/ledger-live-common/src/families/aptos/js-getFeesForTransaction.ts
```diff
@@ -69,7 +69,14 @@ export const getFee = async (
             break;
           }
           default: {
-            throw Error(`Simulation failed with following error: ${completedTx.vm_status}`);
+            // HOTFIX: we need to check if there is enough balance for gas
+            // idealy it should be covered with EINSUFFICIENT_BALANCE from network
+            const expectedGas = BigNumber(gasLimit * gasPrice);
+            if (transaction.amount.plus(expectedGas).isGreaterThan(account.spendableBalance)) {
+              break;
```
joaoccmartins, line 76: Is it possible to move this check to an early exit before line 54 `const simulation = await aptosClient.simulateTransaction(publicKeyEd, tx);`?
It seems like it to me, and that would result in tidier code (not a fan of test inside default cases)

joaoccmartins (changes requested):

joaoccmartins (approved):

jccguimaraes (approved):

semeano (approved):

PR #8858, opened 2025-01-09
FIX: LIVE-15628: Convert account address to wallet api id which is the value expected by the dapp browser V2

adammino-ledger:
### ✅ Checklist

  - The Wallet API account is passed now as a parameter to the eth staking modal rather than the LL Account

### 📝 Description

This PR converts the account address selected on LL to the wallet api id which is the value expected by the dapp browser V2

### ❓ Context

- **JIRA or GitHub link**: [JIRA-15628](https://ledgerhq.atlassian.net/browse/LIVE-15628)

---


apps/ledger-live-desktop/src/renderer/families/evm/AccountHeaderManageActions.ts
```diff
@@ -42,9 +43,11 @@ const AccountHeaderActions = ({ account, parentAccount }: Props) => {
         }),
       );
     } else if (account.type === "Account") {
+      const walletApiAccountId = getWalletApiIdFromAccountId(account.id);
+
       dispatch(
         openModal("MODAL_EVM_STAKE", {
-          account,
+          account: { ...account, id: walletApiAccountId },
```
Justkant, line 50: What if we instead sent a walletApiAccount here using `accountToWalletAPIAccount` directly ?
We need to look at `MODAL_EVM_STAKE` but I would prefer to update the type and set the correct expectations instead of changing the id while keep the other type

adammino-ledger: Great idea, i've pushed the change, thanks 🙏🏻

apps/ledger-live-desktop/src/renderer/families/evm/StakeFlowModal_deprecated/EthStakingModalBody.tsx
```diff
@@ -87,13 +87,13 @@ export function EthStakingModalBody({
 
   const checkBoxOnChange = useCallback(() => {
     const value = !doNotShowAgain;
-    global.localStorage.setItem(`${LOCAL_STORAGE_KEY_PREFIX}${account?.currency?.id}`, `${value}`);
```
Justkant, line 90: I think it's important to use the currency id here and not account id as we want the checkBox to work on all accounts for a given currency
I think you should be able to use `account?.currency` for this

libs/ledger-live-common/src/wallet-api/converters.ts
Justkant, line 1: I think we can revert the changes (or most of them) in this file too now that we only use `accountToWalletAPIAccount`

adammino-ledger: I thought the function wouldn't harm. I used it on the same file, and it could be useful for that conversion on the outside. Do you still want me to revert it ?

Justkant: I suppose we could keep it

apps/ledger-live-desktop/src/renderer/families/evm/StakeFlowModal_deprecated/EthStakingModalBody.tsx
Justkant: Not sure why we still have it then but I would prefer to keep it correct for the logic

adammino-ledger: yeah but then "However I changed to account id as it's a key to store something on the local storage so it should in theory be unique, account?.currency wouldn't be unique as more accounts can be having the same currency. But let me know if you still want me to do the change, i don't mind"  ? if i change it to currency is it not gonna be unique i guess ? it's just a key for the local storage

Justkant: You want it to be for a given currency as you choose your preferred provider for a stacking on a given currency not an account

adammino-ledger: cool i changed it to currency 👍🏻

Justkant (approved):

beths-ledger (dismissed):
[Aside]: Do you know if we need to update LLM types, too?

adammino-ledger:
> [Aside]: Do you know if we need to update LLM types, too?

yes same update needs to be done for LLM, i was gonna fix the desktop bug quickly and keep that for another PR but now that the it's taking some time for the e2e tests, i will push the changes here 👍🏻

ypolishchuk-ledger (approved):

Justkant (dismissed):

Justkant (approved):

apps/ledger-live-desktop/src/renderer/families/evm/StakeFlowModal_deprecated/EthStakingModalBody.tsx
beths-ledger: sorry pals - we will remove this deprecated evm modal soon! https://ledgerhq.atlassian.net/browse/CN-1053

beths-ledger (approved):

cgrellard-ledger (approved):

PR #8863, opened 2025-01-10
Run pr-build-and-test on merge queue

angusbayley:
Part of https://ledgerhq.atlassian.net/browse/LIVE-14748

Updates our workflows so that ~`build-and-test-pr` workflows are run on Github runners if the triggering event is a merge queue group.~ the `build-and-test-pr` job runs on merge group, clearing the merge queue. 

This anticipates the activation of a merge queue on `develop`, as has been done in the device-sdk-ts repo.

There is no easy way to test this - I recommend we do a live switchover late in the evening at the end of a workday, announcing to #live-engineering in advance, or at a weekend. Therefore, merge strategy is to merge at the end of a workday, in advance of an end-of-workday session where we turn on the merge queue and validate that it works.

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -1,5 +1,7 @@
 name: "[PR] Build and Test"
 on:
+  merge_group:
+    types: [checks_requested]
   pull_request:
 
 concurrency:


KVNLS (commented):
I do not understand why we leverage Github runners when we are in a merge group, can you elaborate, please? Does the workflow work well on GH runners even with all the setup done in our self-hosted runners?
For the test part, I think that we can do what you proposed; We can communicate on live-engineering and test it and if something is not working well we can revert.

angusbayley:
To answer the question above - the reason for GH runners for the merge queue checks is to ensure the extra load created by merge queue checks does not create a longer job pickup time for Ledger runners. Github runners of course are slower running but this is not an issue seen as the merge queue is expected to rarely fail. When a PR gets put into the merge queue it is fairly safe to assume it'll be merged.

lewisd5 (approved):

PR #8918, opened 2025-01-16
fix :bug:(lld) arrow tooltip wrong color

LucasWerey:
### ✅ Checklist

  - sidebar

### 📝 Description

There is an issue with the tooltip arrow that's not dynamically changing color. I've deactivated it since it wasn't use before the changes I did on the sidebar. I think the issue is more based on tippy.js that is not updating the arrow (it's a svg so it might be the reason)

## Dark mode
|image|image|

## Light mode
|image|image|

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/SideBar/SideBarListItem.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/SideBar/SideBarListItem.tsx
@@ -52,7 +52,7 @@ class SideBarListItem extends PureComponent<Props> {
         </span>
       );
     return (
-      <Tooltip content={renderedLabel} enabled={!!collapsed} placement="right">
+      <Tooltip arrow={false} content={renderedLabel} enabled={!!collapsed} placement="right">
         <Container
           data-testid={`drawer-${id}-button`}
           isActive={!disabled && isActive}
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/renderer/components/SideBar/SideBarListItem.tsx
```diff
@@ -52,7 +52,7 @@ class SideBarListItem extends PureComponent<Props> {
         </span>
       );
     return (
-      <Tooltip content={renderedLabel} enabled={!!collapsed} placement="right">
+      <Tooltip arrow={false} content={renderedLabel} enabled={!!collapsed} placement="right">
```
themooneer, line 55: by default it's false in the Tooltip component implementation 
 `arrow={arrow ? roundArrow : false}` 
 Why you need to explicitly force it ? otherwise the fix should be on the component level, because normally when the boolean props is not passed on the top level , it should be considered falsy.
 See `apps/ledger-live-desktop/src/renderer/components/Tooltip.tsx`

themooneer: forget about that, it will require a tramendous change on every component that calls the Tooltip -> i will proceed to a test asap.

themooneer (approved):
lgtm

PR #8959, opened 2025-01-20
feat: close add account v2 multiple navigator via a confirmation drawer

themooneer:
### ✅ Checklist

  - ...

### 📝 Description

Simply when a user is in an Add account process and try to quit there are 2 possibilities: 
- A confirmation dialog will appear in the following screens:  Crypto and Network and Device Selection, ScanDeviceAccoun screen & Hedera no associated account screen.
- Direct exit from warning/success screen 

The close process will get back to the main wallet page. 

image

https://github.com/user-attachments/assets/db7016fb-7546-4d99-972a-2fa5779e2abf

### ❓ Context

- **JIRA or GitHub link**: 

https://ledgerhq.atlassian.net/browse/LIVE-15606

---

### DIFF
--- a/apps/ledger-live-mobile/src/newArch/components/CloseWithConfirmation/index.tsx
+++ b/apps/ledger-live-mobile/src/newArch/components/CloseWithConfirmation/index.tsx
@@ -1,0 +1,50 @@
+import React from "react";
+import { Trans } from "react-i18next";
+import { useRoute } from "@react-navigation/native";
+import { NavigationHeaderCloseButtonAdvanced } from "~/components/NavigationHeaderCloseButton";
+import { ScreenName } from "~/const";
+
+const routesWithConfirmation: string[] = [
+  ScreenName.SelectDevice,
+  ScreenName.SelectNetwork,
+  ScreenName.AddAccountsSelectCrypto,
+  ScreenName.ScanDeviceAccounts,
+  ScreenName.NoAssociatedAccounts,
+];
+export default function CloseWithConfirmation({
+  onClose,
+  showButton,
+  buttonText,
+}: {
+  onClose?: () => void;
+  showButton?: boolean;
+  buttonText?: string;
+}) {
+  const route = useRoute();
+  return (
+    <NavigationHeaderCloseButtonAdvanced
+      withConfirmation={routesWithConfirmation.includes(route.name)}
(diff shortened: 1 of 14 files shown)


apps/ledger-live-mobile/src/newArch/components/CloseWithConfirmation/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const route = useRoute();
+  return (
+    <NavigationHeaderCloseButtonAdvanced
+      withConfirmation={routesWithConfirmation.includes(route.name)}
+      confirmationTitle={<Trans i18nKey="addAccounts.quitConfirmation.v2.title" />}
+      confirmationDesc={<Trans i18nKey="addAccounts.quitConfirmation.v2.desc" />}
+      {...(onClose && { onClose })}
+      {...(showButton && { showButton })}
+      {...(buttonText && { buttonText })}
+      customDrawerStyle={{
```
cgrellard-ledger, line 32: Should we take this opportunity to refactor the `NavigationHeaderCloseButtonAdvanced` component to avoid having to pass all these hardcoded values ? Or at least use the values from the design system ?
Wdyt ?

themooneer: it's absolutely not a hardcoded props, it's an extension of the `NavigationHeaderCloseButtonAdvanced` components contract which give it more flexibility in term of customizing the drawer without modifying it's behaviour for other related feature. It's near the Open-Closed principle of SOLID (Open for extension , closed for modification) which is in my point of view a good practice. 

On the other hand, refactoring this component in particular is not a good decision, because there are a lot of transversal feature that depends on it. It can be done out of this scope.

cgrellard-ledger: Understood, I was just thinking about using values from the DS for the fontSize, fontWeight, ... instead of putting numbers like 24 and 600 but I just looked into it and it seems like they are not exposed by the native-ui library.
Nevermind for this PR but it's something we could do when we revamp the DS

cgrellard-ledger (approved):

apps/ledger-live-mobile/src/newArch/components/CloseWithConfirmation/index.tsx
themooneer: Totally agree ! nice catch

PR #8875, opened 2025-01-10
fix tezos pagination

jprudent:

PR #8987, opened 2025-01-23
feat(coin:xrp): apply api change on operations

jprudent:
Sorry, this is a big PR.

I changed the API of "operations" so, it impacted the code on all coins that implements the coin framework. Specifically I changed the semantic of "operations" so that it returns all operations, without pagination. In the future we may paginate using a proper token described in this document https://ledgerhq.atlassian.net/wiki/x/XIGeRAE

For stellar/tezos/xrp coins: I also changed signature of "logic" layer so that the token used for iteration is string not a number. The value of the token is a stringified json representation of the token returned by the underlying "network" layer. It's returned by the listOperations so that it could be reused as input in future implementation.

For stellar and tezos, the "api" is adapted for compilation but not yet ready to use, we'll have 2 respective separate PR for that:
- https://github.com/LedgerHQ/ledger-live/pull/8898
- https://github.com/LedgerHQ/ledger-live/pull/8875

For XRP specifically, 
- the "api" loop over the "logic" using the dedicated token provided by the explorer instead of the block height (that was buggy). There is a hard coded limit over the number of iteration to avoid infinite loop (if there is a bug or something) that could result in freezing users. There is a log when that limit is reached.
- introduced a `order` parameter as a workaroud for LIVE-16705

integrations tests: https://github.com/LedgerHQ/ledger-live/actions/runs/13056449704

### DIFF
--- a/libs/coin-modules/coin-polkadot/src/api/index.integ.test.ts
+++ b/libs/coin-modules/coin-polkadot/src/api/index.integ.test.ts
@@ -42,7 +42,7 @@ describe("Polkadot Api", () => {
   describe("listOperations", () => {
     it("returns a list regarding address parameter", async () => {
       // When
-      const [tx, _] = await module.listOperations(address, { limit: 100 });
+      const [tx, _] = await module.listOperations(address, { minHeight: 0 });
 
       // Then
       expect(tx.length).toBeGreaterThanOrEqual(1);
@@ -56,7 +56,7 @@ describe("Polkadot Api", () => {
 
     it("returns all operations", async () => {
       // When
-      const [tx, _] = await module.listOperations(address, {});
+      const [tx, _] = await module.listOperations(address, { minHeight: 0 });
 
       // Then
       const checkSet = new Set(tx.map(elt => elt.hash));
--- a/libs/coin-modules/coin-polkadot/src/api/index.integ.test.ts
+++ b/libs/coin-modules/coin-polkadot/src/api/index.integ.test.ts
@@ -54,11 +54,9 @@ describe("Polkadot Api", () => {
       });
     }, 20000);
 
-    it("returns paginated operations", async () => {
+    it("returns all operations", async () => {
       // When
-      const [tx, idx] = await module.listOperations(address, { limit: 100 });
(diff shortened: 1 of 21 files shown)


libs/coin-modules/coin-polkadot/src/api/index.integ.test.ts
```diff
@@ -54,11 +54,9 @@ describe("Polkadot Api", () => {
       });
     }, 20000);
 
-    it("returns paginated operations", async () => {
+    it("returns all operations", async () => {
       // When
-      const [tx, idx] = await module.listOperations(address, { limit: 100 });
-      const [tx2, _] = await module.listOperations(address, { limit: 100, start: idx });
-      tx.push(...tx2);
+      const [tx, _] = await module.listOperations(address, {});
```
hedi-edelbloute, line 59: I'm wondering if it makes sense to have Pagination object as optional

libs/coin-modules/coin-xrp/src/api/index.integ.test.ts
```diff
@@ -42,15 +43,17 @@ describe("Xrp Api", () => {
...
-
+      const [tx, _] = await module.listOperations(bigAddress, { start: 0 });
       // Then
       const checkSet = new Set(tx.map(elt => elt.hash));
       expect(checkSet.size).toEqual(tx.length);
+      // the first transaction is returned
+      expect(tx[0].block.height).toEqual(73126713);
+      // 200 is the default XRP explorer hard limit,
+      // so here we are checking that this limit is bypassed
+      expect(tx.length).toBeGreaterThan(200);
```
hedi-edelbloute, line 56: We could check that the first tx on this account is present via tx hash

hedi-edelbloute (approved):
Thank you ! minor improvements possible but not mandatory

jprudent:
@hedi-edelbloute I fixed the comments
There are tests that are failing in https://github.com/LedgerHQ/ledger-live/actions/runs/13032147026/job/36353665865#step:7:4009
looks like there is more tx than before, so is it ok to merge ?

hedi-edelbloute (approved):

hedi-edelbloute (approved):

PR #8989, opened 2025-01-23
ci: Bump node version to LTS

lewisd5:
Bumping the version of node to the current LTS version.

Notes: 
- Test run can be seen downloading the package in the cache step (via proto)
- this (optionally) requires all caches to be updated on the various build images

Test run: https://github.com/LedgerHQ/ledger-live/actions/runs/13013224395

https://ledgerhq.atlassian.net/browse/LIVE-14805


ypolishchuk-ledger (approved):

KVNLS (dismissed):

Justkant (dismissed):

jiyuzhuang (approved):

ypolishchuk-ledger (approved):

apps/cli/bin/index.js
```diff
@@ -1,2 +1,2 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --disable-warning=DEP0040
```
jccguimaraes, line 1: I don't agree with hiding the warning. As official documentation states

> The version of the punycode module bundled in Node.js is being deprecated. In a future major version of Node.js this module will be removed. Users currently depending on the punycode module should switch to using the userland-provided [Punycode.js](https://github.com/bestiejs/punycode.js) module instead. For punycode-based URL encoding, see [url.domainToASCII](https://nodejs.org/api/url.html#urldomaintoasciidomain) or, more generally, the [WHATWG URL API](https://nodejs.org/api/url.html#the-whatwg-url-api).

Which means things will be broken at some point if dependencies don't update to the alternative

jccguimaraes: but even the `punycode.js` is not maintained since 2023.

jiyuzhuang (dismissed):

Justkant (dismissed):

ypolishchuk-ledger (dismissed):

valpinkman (dismissed):

PR #8999, opened 2025-01-24
fix: add account v2 redirection from receive when no account found / add account v1 is OK

themooneer:
### ✅ Checklist

  - ...

### 📝 Description

Issue analysis:

After reviewing the scenario in detail, it appears the account selection screen displayed in this particular use case is the "add account v1" version (designed for receiving), even the screen preceding the scan is the "add account v1" (intended for receiving). Therefore, when you select an account from the list, it will be added.

Essentially, on the production environment, this isn't a bug, because it's a logical sequence:

I initiate a "receive" operation.

I select an asset for which I don't have an existing account.

The scan that launches is the "add account v1" version.

The displayed list is the "add account v1" list for receiving (since it's a single selection).

Normally, if the "add account v2" feature flag is activated, we should go through (in this specific case) the "add account v2" device scan -> then a success or warning screen -> when we choose an account -> redirection to the address/QR code page. This case should be part of the receive plugin, however, I only handled the case where the "add account" button is on the account selection screen in cases of accounts already existing on the receive.

Solution: 

When the user try to receive funds on currency where there is no persisted account, the redirection should be done to Add account v2 flow (FF ON) and on the success/warning page when an account is selected , the redirection should be to the receive confirmation page with QR code. 

Full scenario here ⬇️

https://github.com/user-attachments/assets/0a8cb91a-b94a-4e7a-b419-f2df29eff4b6

I also made a refactor for the use of string literal `"addAccounts"` or `"receiveFunds"` across the codebase and replaced it with common enum based type. 

Please make sure your review covers the scope of the changes. 

### ❓ Context

- **JIRA or GitHub link**: 

https://ledgerhq.atlassian.net/browse/LIVE-16522
---

### DIFF
--- a/apps/ledger-live-mobile/src/components/RootNavigator/types/ReceiveFundsNavigator.ts
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/types/ReceiveFundsNavigator.ts
@@ -6,9 +6,9 @@ import {
 } from "@ledgerhq/types-cryptoassets";
 import { Device, DeviceModelId } from "@ledgerhq/types-devices";
 import { AccountLike } from "@ledgerhq/types-live";
-import { NavigatorScreenParams } from "@react-navigation/core";
+import { NavigatorScreenParams } from "@react-navigation/native";
 import type { NavigatorName, ScreenName } from "~/const";
-import { DeviceSelectionNavigatorParamsList } from "~/newArch/features/DeviceSelection/types";
+import { DeviceSelectionNavigatorParamsList } from "LLM/features/DeviceSelection/types";
 
 export type ReceiveFundsStackParamList = {
   [ScreenName.ReceiveSelectCrypto]:
--- a/apps/ledger-live-mobile/src/components/RootNavigator/types/ReceiveFundsNavigator.ts
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/types/ReceiveFundsNavigator.ts
@@ -6,7 +6,9 @@ import {
 } from "@ledgerhq/types-cryptoassets";
 import { Device, DeviceModelId } from "@ledgerhq/types-devices";
 import { AccountLike } from "@ledgerhq/types-live";
-import type { ScreenName } from "~/const";
+import { NavigatorScreenParams } from "@react-navigation/core";
+import type { NavigatorName, ScreenName } from "~/const";
+import { DeviceSelectionNavigatorParamsList } from "~/newArch/features/DeviceSelection/types";
 
 export type ReceiveFundsStackParamList = {
   [ScreenName.ReceiveSelectCrypto]:
@@ -76,4 +78,7 @@ export type ReceiveFundsStackParamList = {
     onSuccess?: (_?: string) => void;
     onError?: () => void;
   };
+  [NavigatorName.DeviceSelection]?: Partial<
+    NavigatorScreenParams<DeviceSelectionNavigatorParamsList>
+  >;
 };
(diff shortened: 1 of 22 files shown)


jiyuzhuang (dismissed):

cgrellard-ledger (dismissed):

apps/ledger-live-mobile/src/components/RootNavigator/types/ReceiveFundsNavigator.ts
```diff
@@ -6,7 +6,9 @@ import {
 } from "@ledgerhq/types-cryptoassets";
 import { Device, DeviceModelId } from "@ledgerhq/types-devices";
 import { AccountLike } from "@ledgerhq/types-live";
-import type { ScreenName } from "~/const";
+import { NavigatorScreenParams } from "@react-navigation/core";
```
KVNLS, line 9: On documentation they use `@react-navigation/native` to import `NavigatorScreenParams`, is there a reason to use `@react-navigation/core` instead? What's the difference?

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccount/enums.ts
```diff
@@ -0,0 +1,4 @@
+export enum AddAccountContexts {
```
KVNLS, line 1: Are we sure to use enums?
- https://fettblog.eu/tidy-typescript-avoid-enums/
- https://thoughtbot.com/blog/the-trouble-with-typescript-enums

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccountWarning/index.tsx
```diff
@@ -29,17 +30,27 @@ export default function AddAccountsWarning({ route }: Props) {
...
+  const { emptyAccount, emptyAccountName, currency, context } = route.params || {};
+
+  const statusColor = colors.warning.c70;
+
   const goToAccounts = useCallback(
     (accountId: string) => () => {
-      navigation.navigate(ScreenName.Account, {
-        accountId,
-      });
+      if (context === AddAccountContexts.AddAccounts)
```
KVNLS, line 39: Can you use braces as we do most of the time in the codebase, to keep it harmonized, please?
Can we add an ESLint rule?

CremaFR (dismissed):

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccount/enums.ts
themooneer: i like enums , newer versions of typescripts support it very well in term of typing/inference(in the ends they are objects), in term of system design there is a huge difference between an enumeration and an Object to define limited set of values for a type. It's readable. There will be no issues in transpilation/ build / performance time. Why don't we ?

apps/ledger-live-mobile/src/newArch/features/Accounts/screens/AddAccountWarning/index.tsx
themooneer: sure ! it's called `curly` rule , it depends on developer's XP/taste , oneshot block is less readable i agree, i will change it . Thanks

apps/ledger-live-mobile/src/components/RootNavigator/types/ReceiveFundsNavigator.ts
themooneer: good question, i need to figure out

themooneer: Well, `@react-navigation/native` import everything from `@react-navigation/core` where the type `NavigatorScreenParams` truly exists, and while there is this import, there is a reason to decouple everything from core and make native as a façade , which justifies why the recommend to use the `native` , well seen ! just changed.

cgrellard-ledger (approved):

CremaFR (approved):

KVNLS (approved):

themooneer:
Valentin has already approved this PR. No major updates after the last push , I'm proceeding to merge it.

PR #9076, opened 2025-01-31
feat(coin:xrp): add memo and destination tag in crafting

jprudent:
integration tests: https://github.com/LedgerHQ/ledger-live/actions/runs/13076078720
(failing but I think I'm not accountable)

### DIFF
--- a/libs/coin-framework/src/api/types.ts
+++ b/libs/coin-framework/src/api/types.ts
@@ -24,14 +24,9 @@ export type Transaction = {
   recipient: string;
   amount: bigint;
   fee: bigint;
-  supplement?: unknown;
-};
+} & Record<string, unknown>; // Field containing dedicated value for each blockchain
 
-// TODO rename start to minHeight
-//       and add a `token: string` field to the pagination if we really need to support pagination
-//       (which is not the case for now)
-//       for now start is used as a minHeight from which we want to fetch ALL operations
-//       limit is unused for now
+// TODO add a `token: string` field to the pagination if we really need to support pagination (which is not the case for now)
 //       see design document at https://ledgerhq.atlassian.net/wiki/spaces/BE/pages/5446205788/coin-modules+lama-adapter+APIs+refinements
 export type Pagination = { minHeight: number };
 export type Api = {
--- a/libs/coin-modules/coin-xrp/src/logic/craftTransaction.ts
+++ b/libs/coin-modules/coin-xrp/src/logic/craftTransaction.ts
@@ -8,20 +8,34 @@ import { UINT32_MAX, validateTag } from "./utils";
 const LEDGER_OFFSET = 20;
 
 const { TRANSACTION_TYPES } = XrplDefinitions;
+type Memo = {
+  MemoData?: string;
+  MemoFormat?: string;
+  MemoType?: string;
+};
+type MemoWrapper = {
+  Memo: Memo;
+};
 type XrplTransaction = {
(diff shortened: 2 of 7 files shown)


hedi-edelbloute (approved):

libs/coin-framework/src/api/types.ts
```diff
@@ -24,14 +24,9 @@ export type Transaction = {
   recipient: string;
   amount: bigint;
   fee: bigint;
-  supplement?: unknown;
-};
+} & Record<string, unknown>; // Field containing dedicated value for each blockchain
```
sprohaszka-ledger, line 27: By doing this, we will not isolate dedicate fields into a substructure 🤔 
I think it's a shame :(

sprohaszka-ledger (commented):
I would have keep dedicated field into a separate substructure.

libs/coin-framework/src/api/types.ts
jnicoulaud-ledger: @sprohaszka-ledger this data structure uses the flattened form in lama-adapter API, then transformed with `supplement` just for here, then back to flattened form internally, does this really bring any value ?

Suggestion: also add a `type` field and turn this into a tagged union so we can have an exact type for each family

jnicoulaud-ledger: discussed with Stéphane and Jérôme => agreed to keep current API until we have a better idea of how we want to handle this with tron integration

libs/coin-modules/coin-xrp/src/logic/craftTransaction.ts
```diff
@@ -31,31 +45,49 @@ export async function craftTransaction(
...
+    xrplTransaction.Memos = transaction.memos.map(memoMapper);
+  }
+
+  if (transaction.destinationTag) {
     invariant(
-      validateTag(new BigNumber(tag)),
+      validateTag(new BigNumber(transaction.destinationTag)),
       `tag is set but is not in a valid format, should be between [0 - ${UINT32_MAX.toString()}]`,
     );
+    xrplTransaction.DestinationTag = transaction.destinationTag;
```
hedi-edelbloute, line 90: If we move transaction.tag to transaction.destinationTag, we need to update UI

https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/families/xrp/SendRecipientFields.tsx#L37

(same for LLM and basically every link to this variable in UI)

hedi-edelbloute (changes requested):

libs/coin-modules/coin-xrp/src/logic/craftTransaction.ts
jprudent: Looks like the UI has a clean separation between views and services:
- the bridge defines its own Transaction type https://github.com/LedgerHQ/ledger-live/blob/8ed2cd3f6219cea2ee22851468477bb6e2cbb1c3/libs/coin-modules/coin-xrp/src/types/bridge.ts#L22
- and its mapped from `tag` to `destinationTag` when calling `crafsTransaction` https://github.com/LedgerHQ/ledger-live/blob/8ed2cd3f6219cea2ee22851468477bb6e2cbb1c3/libs/coin-modules/coin-xrp/src/bridge/signOperation.ts#L40

hedi-edelbloute: Indeed object is only used there

PR #9126, opened 2025-02-05
fix(coin:tezos): handle missing recipients or senders cleanly

jprudent:
What's changed:
- when recipients or senders is not found, it return an empty array instead of `[""]`
- for delegate/undelegate operations, it adds the target adress as recipients

int tests: https://github.com/LedgerHQ/ledger-live/actions/runs/13325154125

### DIFF
--- a/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
@@ -1,4 +1,5 @@
 import { tzkt } from "../network";
+import { log } from "@ledgerhq/logs";
 import {
   type APIDelegationType,
   type APITransactionType,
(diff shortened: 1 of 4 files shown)


libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -33,25 +34,40 @@ export async function listOperations(
...
 ): Operation {
   const { amount, hash, storageFee, sender, timestamp, type, counter } = operation;
-  let targetAddress = "";
-  if (isAPITransactionType(operation) && operation.target) {
-    targetAddress = operation.target.address;
+
+  let targetAddress = undefined;
+  if (isAPITransactionType(operation)) {
+    targetAddress = operation?.target?.address;
+  } else if (isAPIDelegationType(operation)) {
```
hedi-edelbloute, line 56: Add a comment if you find it necessary

hedi-edelbloute (approved):

hedi-edelbloute (approved):

PR #9134, opened 2025-02-06
:bug: fix(llm) account item wrong balance

LucasWerey:
### ✅ Checklist

  - accountItem

### 📝 Description

Spendable balance was used instead of balance. User wouldn't see the all balance (including staked)
Test added to ensure we well display the balance and not the spendable balance

### ❓ Context

- **JIRA or GitHub link**: [LIVE-16846](https://ledgerhq.atlassian.net/browse/LIVE-16846)

---

[LIVE-16846]: https://ledgerhq.atlassian.net/browse/LIVE-16846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/components/SelectableAccountsList.tsx
```diff
@@ -319,7 +319,7 @@ const SelectableAccount = ({
             padding="8px"
             columnGap={8}
           >
-            <AccountItem account={account as Account} balance={account.spendableBalance} />
+            <AccountItem account={account as Account} balance={account.balance} />
```
mcayuelas-ledger, line 322: Are we sure we want that?

Because into those flows, the aim is to send/receive, we want what we currently is available, not staking value for exemple no?

LucasWerey: I've switch for `useFullBalance ? account.balance : account.spendableBalance` because today with the ff off we use accountCard that has this logic => ` <CurrencyUnitValue
              showCode
              unit={unit}
              value={useFullBalance ? account.balance : getAccountSpendableBalance(account)}
            />`

LucasWerey: I will need to check again my bad I am not sure about what we should use because the currencyUnitValue as not the same behaviour as Countervalue component inside AccountCard for the balance

LucasWerey: Ok so today in the send flow we display the all balance and not only the spendable balance. So I think we can keep this `<CurrencyUnitValue showCode unit={unit} value={useFullBalance ? account.balance : getAccountSpendableBalance(account)} />` wdyt ?

mcayuelas-ledger: ok with final thought

mcayuelas-ledger (approved):

PR #9141, opened 2025-02-06
skip build-and-test-pr if already run on this file set

angusbayley:
Skips merge queue run of `build-and-test-pr` if:
- the PR's pre-merge-queue CI was run against latest `develop`
- the PR is being added to an empty merge queue

In this case, the CI check from before entering the merge queue is sufficient, as it tests an identical repo state to what a merge_group check would test.

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -12,11 +12,28 @@ permissions:
   id-token: write
   contents: read
   pull-requests: write
+  actions: write
 
 jobs:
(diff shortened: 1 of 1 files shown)


valpinkman:
> In this case, the CI check from before entering the merge queue is sufficient, as it tests an identical repo state to what a merge_group check would test.

I don't believe this is true, it's only true when you are the only one in the merge queue, but it won't be if there are others. Also it won't be true if your branch is not up to date.

cc: @KVNLS what do you think ?

PR #9158, opened 2025-02-10
feat: change permit 2 modal design

Canestin:
### ✅ Checklist

  - ...

### 📝 Description

We want to change the design of the Uniswap sign message modal according to this [Figma design](https://www.figma.com/design/smuoyRXju0LmIQ9CDNd7iW/Live-App-MVP?node-id=4314-31508&t=RoMi4uk6HheDS0N1-0)

We still display non `Permit2.PermitSingle` messages with all their fields: 

Shottr 2025-02-17 10 44 59

### ❓ Context

- **JIRA**:  https://ledgerhq.atlassian.net/browse/LIVE-15595 — https://ledgerhq.atlassian.net/browse/LIVE-15596

---


Canestin:
I am facing two problems :

- I used the existing component FlowStepper.Indexed from @ledgerhq/react-ui to show steps, but it doesn’t display well like the others; it is too big. I spent a lot of time trying to figure out why. Do you have any idea?
  
img1

- Do we have an implemented method or library to convert USD to ETH?

Wozacosta:
Be careful, by modifying apps/ledger-live-desktop/src/renderer/components/Stepper/index.tsx you change all steppers: 

image

It's probably not what you're aiming for.

qperrot (commented):
need changeset

libs/evm-tools/src/message/EIP712/index.ts
```diff
@@ -219,10 +243,22 @@ export const getEIP712FieldsDisplayedOnNano = async (
...
+      displayedInfos.push({
+        label: "Amount",
+        value: getValueFromPath(field.path, messageData),
+      });
+    } else if (field.path.includes("expiration")) {
+      displayedInfos.push({
+        label: "Approval expires",
+        value: formatDate(getValueFromPath(field.path, messageData) as string),
+      });
+    }
```
qperrot, line 261: Can't we do something like 
`const labelMap: Record<string, string> = {
  token: "Token",
  amount: "Amount",
  expiration: "Approval expires",
};

const key = Object.keys(labelMap).find(keyword => field.path.includes(keyword));

if (key) {
  displayedInfos.push({
    label: labelMap[key],
    value: key === "expiration"
      ? formatDate(getValueFromPath(field.path, messageData) as string)
      : getValueFromPath(field.path, messageData),
  });
}`

apps/ledger-live-desktop/src/renderer/modals/SignTransaction/steps/StepSummary.tsx
```diff
@@ -83,6 +83,8 @@ const StepSummary = (props: StepProps) => {
 
   const memo = "memo" in transaction ? transaction.memo : undefined;
 
+  console.log("NOW > totalSpent de chez totalSpent", totalSpent);
+  console.log("NOW > unit de chez unit ", unit);
```
Wozacosta, line 87: logs to remove 🧹 ;)

apps/ledger-live-desktop/src/renderer/components/Stepper/index.tsx
```diff
@@ -4,12 +4,12 @@ import { withTranslation } from "react-i18next";
...
 import { ModalBody } from "~/renderer/components/Modal";
 import { useDeviceBlocked } from "~/renderer/components/DeviceAction/DeviceBlocker";
-import Breadcrumb from "./Breadcrumb";
 import { useTrackAddAccountModal } from "~/renderer/analytics/hooks/useTrackAddAccountModal";
+import Breadcrumb from "./Breadcrumb";
 import { useSelector } from "react-redux";
 import { trackingEnabledSelector } from "~/renderer/reducers/settings";
 import { getCurrentDevice } from "~/renderer/reducers/devices";
-import { LedgerError } from "../DeviceAction";
+import { LedgerError } from "~/renderer/components/DeviceAction";
```
Wozacosta, line 12: maybe undo the code changes in this file as it does nothing

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -12,29 +12,50 @@ import useTheme from "~/renderer/hooks/useTheme";
...
-  text-align: right;
-  max-width: 50%;
-`;
+}))``;
 
 export type FieldComponentProps = {
-  account: AccountLike;
+  account: Account;
   field: DeviceTransactionField;
+  contractAddress: string;
```
Wozacosta, line 27: you can't be 100% there will be a contractAddress in the message to sign, other message.sign actions could be triggered without that field, either make it optional or accept undefined

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -60,43 +63,78 @@ const AdvancedMessageArea = styled.pre`
...
-  );
-});
+const MessageProperty = memo(
+  ({
+    label,
+    value,
+    account,
+    contractAddress,
+  }: MessageProperties[0] & { account?: Account; contractAddress?: string | string[] }) => {
+    if (!value || !account || !contractAddress) return null;
```
Wozacosta, line 73: we should probably handle the case when what's signed is not a permit2: 

Here is how it looks like now (in production) when signing on p2p.org (liveapp within discover) 
Shottr 2025-02-13 17 54 27 

And on your branch: 
Shottr 2025-02-13 17 54 45

Wozacosta: When removing that conditional, we get :

image


Maybe some stilling is still needed for those messages (word-wrap/ flex-wrap)

libs/evm-tools/src/message/EIP712/index.ts
Canestin: Hi @qperrot thanks for proposition. I chose the first solution for its clarity and readability. Each condition is explicit and easy to modify if needed, while the second approach can make the code less transparent, especially when adding new fields. It is also less costly.

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
Wozacosta: FYI: fixed the styling in https://github.com/LedgerHQ/ledger-live/pull/9158/commits/53580003ffdb960c5384ac074bb43b72729eef4a

libs/evm-tools/src/message/EIP712/index.ts
```diff
@@ -219,10 +243,22 @@ export const getEIP712FieldsDisplayedOnNano = async (
...
+    } else if (field.path.includes("amount")) {
+      displayedInfos.push({
+        label: "Amount",
+        value: getValueFromPath(field.path, messageData),
+      });
+    } else if (field.path.includes("expiration")) {
+      displayedInfos.push({
+        label: "Approval expires",
+        value: formatDate(getValueFromPath(field.path, messageData) as string),
+      });
```
Wozacosta, line 260: missing an `else` clause, as for now we remove all fields and this is not what we want, we might want to rethink how fields are included if it's a sign message for a permit 2 or not.

more context: 

https://github.com/LedgerHQ/ledger-live/actions/runs/13326569019/job/37221052447?pr=9158

In this test we check for the presence of many fields we want to be displayed, it's not a permit2 signature but it's still filtering out fields

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -12,29 +12,50 @@ import useTheme from "~/renderer/hooks/useTheme";
...
+
+  const unit = isPropertyAmount
+    ? account.subAccounts?.find(a => a.token.contractAddress === contractAddress)?.token.units[0]
+    : undefined;
+
   return field.type === "text" ? (
     <SignMessageConfirmField label={field.label}>
-      <FieldText>{field.value}</FieldText>
+      <FieldText>
+        {isPropertyAmount ? (
```
Wozacosta, line 42: should make a check on `unit` instead, we could end up in this situation, as `<FormattedVal>` has an invariant that checks for `unit`: 

image
```suggestion
        {unit ? (
```

Wozacosta: I've fixed it 6aae17f679200545860862bb70fde2421677caee

libs/evm-tools/src/message/EIP712/index.ts
Wozacosta: Fixed in 6aae17f679200545860862bb70fde2421677caee 

Before: 
Code 2025-02-17 10 27 28


After: 
Shottr 2025-02-17 10 44 59

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -128,19 +176,19 @@ export default function StepSummary({ account, message: messageData }: StepProps
...
+              <CryptoCurrencyIcon size={14} currency={account.currency} />
             </div>
             <Text ff="Inter" color="palette.text.shade100" fontSize={4} style={{ flex: 1 }}>
               {accountName}
             </Text>
           </Box>
         </Box>
       </Box>
-      <Separator />
+      <Separator color="red" />
```
Wozacosta, line 187: no need for the color prop here

Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/c60afc46f434fb4ab1733daebee59c66948fb091

Wozacosta (approved):

Wozacosta (approved):

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -97,9 +118,18 @@ const SignMessageConfirm = ({ device, account, parentAccount, signMessageRequest
         <Spinner size={30} />
       ) : (
         <>
-          <Box style={{ width: "100%" }} px={30} mb={20}>
+          <Box style={{ width: "100%", rowGap: 7 }} mb={20}>
```
LucasWerey, line 121: The rowGap doesn't have any impact here. I am not sure what's the goal here since I can't reach the modified screen but to make it works use `<Flex> `component or add display: "flex"

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -60,43 +63,88 @@ const AdvancedMessageArea = styled.pre`
...
+
+    const isPropertyAmount = label.includes("Amount") && account && contractAddress;
+
+    const tokenUnit = isPropertyAmount
+      ? account.subAccounts?.find(a => a.token.contractAddress === contractAddress)?.token.units[0]
+      : undefined;
+
+    return (
+      <Box
+        style={{ flexDirection: "row", justifyContent: "space-between", maxWidth: "100%" }}
```
LucasWerey, line 83: same here use display flex or` <Flex>` component

LucasWerey (commented):
Not tested 
In general, I find that you use a lot of style={{}}. For most use cases, you can pass props directly to the component without using style props.

LucasWerey (dismissed):
Actually LGTM I thought it was with the `<Box>` from the lib

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Canestin: It does have an impact here.

Wozacosta (approved):

LucasWerey (dismissed):

LucasWerey (dismissed):

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -60,43 +63,88 @@ const AdvancedMessageArea = styled.pre`
...
+    account,
+    contractAddress,
+  }: MessageProperties[0] & { account?: Account; contractAddress?: string | string[] }) => {
+    if (!value) return null;
+
+    const isPropertyAmount = label.includes("Amount") && account && contractAddress;
+
+    const tokenUnit = isPropertyAmount
+      ? account.subAccounts?.find(a => a.token.contractAddress === contractAddress)?.token.units[0]
+      : undefined;
```
mcayuelas-ledger, line 79: You are using this 2 times, you can extract those functions I think :) 

and use parameters

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -60,43 +63,88 @@ const AdvancedMessageArea = styled.pre`
...
+const MessagePropertiesComp = memo(
+  (props: { properties: MessageProperties | null } & { account: Account }) => {
+    const { properties, account } = props;
+    const contractAddress = properties?.find(p => p.label === "Token")?.value;
+
+    return properties ? (
+      <Box flex="1">
+        {properties.map((p, i) => {
+          const props = { ...p, account, contractAddress };
+          return <MessageProperty key={i} {...props} />;
```
Justkant, line 142: It seems quite strange to do this kind of data transformation here
Why contractAddress and account aren't already in the properties prop ?

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -12,29 +12,50 @@ import useTheme from "~/renderer/hooks/useTheme";
...
 
 const FieldText = styled(Text).attrs(() => ({
   ml: 1,
   ff: "Inter|Medium",
   color: "palette.text.shade80",
   fontSize: 3,
-}))`
-  word-break: break-all;
-  text-align: right;
-  max-width: 50%;
```
Justkant, line 24: Why don't we keep the `word-break` rule ?

Do you remove this so you can get a FieldText on a full width ?
Doesn't it break the current layout ?

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -12,29 +12,50 @@ import useTheme from "~/renderer/hooks/useTheme";
...
 
 export type FieldComponent = React.ComponentType<FieldComponentProps>;
 
-const TextField = ({ field }: FieldComponentProps) => {
+const TextField = ({ field, account, contractAddress }: FieldComponentProps) => {
+  const isPropertyAmount = field.label.includes("Amount") && contractAddress;
+
+  const unit = isPropertyAmount
+    ? account.subAccounts?.find(a => a.token.contractAddress === contractAddress)?.token.units[0]
+    : undefined;
```
Justkant, line 37: Can't we pass the unit above, it feels wrong to have a TextField with all this logic about account and contractAddress with no memo too

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Justkant: @Canestin Why did we resolve this while it's still using the `isPropertyAmount` ?

Edit: looks like it was only updated for this file `apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx`

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -12,29 +12,50 @@ import useTheme from "~/renderer/hooks/useTheme";
...
+              color={"palette.neutral.c100"}
+              val={Number(field.value)}
+              unit={unit}
+              fontSize={3}
+              disableRounding
+              alwaysShowValue
+              showCode
+              inline
+            />
+          </>
```
Justkant, line 54: The fragment seems unnecessary 
```suggestion
          <FormattedVal
            color={"palette.neutral.c100"}
            val={Number(field.value)}
            unit={unit}
            fontSize={3}
            disableRounding
            alwaysShowValue
            showCode
            inline
          />
```

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -12,29 +12,50 @@ import useTheme from "~/renderer/hooks/useTheme";
...
+  const isPropertyAmount = field.label.includes("Amount") && contractAddress;
+
+  const unit = isPropertyAmount
+    ? account.subAccounts?.find(a => a.token.contractAddress === contractAddress)?.token.units[0]
+    : undefined;
+
   return field.type === "text" ? (
     <SignMessageConfirmField label={field.label}>
-      <FieldText>{field.value}</FieldText>
+      <FieldText>
```
Justkant, line 41: I don't think I ever saw a `FormattedVal` in a `FieldText`, I think we usually just use it as a `Text` component, is there any benefit here to nest it in the `FieldText` as it should already be using the same font

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -97,9 +118,18 @@ const SignMessageConfirm = ({ device, account, parentAccount, signMessageRequest
         <Spinner size={30} />
       ) : (
         <>
-          <Box style={{ width: "100%" }} px={30} mb={20}>
+          <Box style={{ width: "100%", rowGap: 7 }} mb={20}>
             {fields.map((field, i) => {
-              return <TextField key={i} field={field} account={account} />;
+              const contractAddress = fields?.find(f => f.label === "Token") as { value: string };
```
Justkant, line 123: If you really need to do a find please don't do it in a map and memoize it

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
Justkant: @Canestin please create a util fonction as you are using the same logic twice but still duplicating the code

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -60,43 +63,88 @@ const AdvancedMessageArea = styled.pre`
...
+                    color={"palette.neutral.c100"}
+                    val={Number(value)}
+                    unit={tokenUnit}
+                    fontSize={3}
+                    disableRounding
+                    alwaysShowValue
+                    showCode
+                    inline
+                  />
+                </>
```
Justkant, line 111: Same as the other one, not sure why we have a fragment here and `FormattedVal` is usually used outside a `Text` component

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Justkant: Also why 7 here and not a round number ?
Is it on our design system and 7 actually refers to a round number ?

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
```diff
@@ -175,7 +226,8 @@ export function StepSummaryFooter({ transitionTo }: StepProps) {
         onClick={() => {
           transitionTo("sign");
         }}
-        primary
+        style={{ borderRadius: 20, backgroundColor: "white" }}
+        color="neutral.c00"
```
Justkant, line 230: You cannot change the footer button only for this step while the other steps keep the old design, it would be inconsistent across the flow

Not sure why the ticket asks to update the design on a single step here, it feels wrong and should have been picked up when refining the ticket
IMO you do it totally or you don't

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/1d9344cac302be9d4dbe3ec3c9d0a804569684d5#diff-15a820817a850003c96032d59991c4b0648434571bf22737f00d331d1842c5f0L33-R41

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/1d9344cac302be9d4dbe3ec3c9d0a804569684d5#diff-f57747013989b40db64bce6d5f0b186372ca88adedd041b19323f319eb3bb4d2

apps/ledger-live-desktop/src/renderer/modals/SignMessage/steps/StepSummary.tsx
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/1d9344cac302be9d4dbe3ec3c9d0a804569684d5#diff-bf82e4739f76f64cd12a8231302a0ad4163bbaf69df8258435af31e5e8a5cfa1R130-R136

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/1d9344cac302be9d4dbe3ec3c9d0a804569684d5#diff-15a820817a850003c96032d59991c4b0648434571bf22737f00d331d1842c5f0R24-R28

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/1d9344cac302be9d4dbe3ec3c9d0a804569684d5#diff-15a820817a850003c96032d59991c4b0648434571bf22737f00d331d1842c5f0R120-R127

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/1d9344cac302be9d4dbe3ec3c9d0a804569684d5#diff-15a820817a850003c96032d59991c4b0648434571bf22737f00d331d1842c5f0R109-R114

libs/evm-tools/src/message/EIP712/index.ts
```diff
@@ -148,6 +148,30 @@ export const getValueFromPath = (path: string, eip721Message: EIP712Message): st
...
+    minute: "2-digit",
+    second: "2-digit",
+    timeZone: "UTC",
+    hour12: false,
+  });
+
+  const parts = formatter.formatToParts(date);
+  const p = (type: string) => parts.find(p => p.type === type)?.value || "00";
+
+  return `${p("year")}-${p("month")}-${p("day")} ${p("hour")}:${p("minute")}:${p("second")} UTC`;
```
Justkant, line 172: Can't we just use the format function of the formatter ?

apps/ledger-live-desktop/src/renderer/utils/getTokenUnit.ts
```diff
@@ -0,0 +1,14 @@
+import { Account } from "@ledgerhq/types-live";
+
+export const getTokenUnit = (
+  label: string,
+  account?: Account,
+  contractAddress?: string | string[],
```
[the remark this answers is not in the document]
Canestin: Normally this is not likely to happen when the label is `amount`, but I covered this case anyway.
[`eda8de1` (#9158)](https://github.com/LedgerHQ/ledger-live/pull/9158/commits/eda8de1f36a6a3bbd3c5f5295db497abde7f648b)

apps/ledger-live-desktop/src/renderer/utils/getTokenUnit.ts
```diff
@@ -0,0 +1,14 @@
...
+) => {
+  const unit =
+    label.includes("Amount") && account && contractAddress
+      ? account.subAccounts?.find(a => a.token.contractAddress === contractAddress)?.token.units[0]
+      : undefined;
+
+  return unit;
+};
```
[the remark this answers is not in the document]
Canestin: https://github.com/LedgerHQ/ledger-live/pull/9158/commits/eda8de1f36a6a3bbd3c5f5295db497abde7f648b

apps/ledger-live-desktop/src/renderer/components/SignMessageConfirm/index.tsx
```diff
@@ -91,16 +106,22 @@ const SignMessageConfirm = ({ device, account, parentAccount, signMessageRequest
     }
   }
 
+  const contractAddress = (fields?.find(f => f.label === "Token") as { value: string })?.value;
```
mcayuelas-ledger, line 109: You shouldn't cast it 

```suggestion
 const contractAddress = fields?.find(f => f.label === "Token" && typeof f.value === 'string')?.value;
```

apps/ledger-live-desktop/src/renderer/utils/getTokenUnit.test.ts
```diff
@@ -0,0 +1,165 @@
+import { Account } from "@ledgerhq/types-live";
+import BigNumber from "bignumber.js";
+import { getTokenUnit } from "./getTokenUnit"; // Remplace par le bon chemin
```
mcayuelas-ledger, line 3: comment ^^

apps/ledger-live-desktop/src/renderer/utils/getTokenUnit.test.ts
mcayuelas-ledger, line 1: You should had a test for Array.isArray in this case, as you check if string or string[]

mcayuelas-ledger (approved):

Justkant (approved):

PR #9175, opened 2025-02-10
fix(coin:stellar): list operations api

jprudent:
Co-authored-by: Salim Belkhir <salim.belkhir@ledger.fr>

integration tests: https://github.com/LedgerHQ/ledger-live/actions/runs/13495943095


jprudent:
about: https://github.com/LedgerHQ/ledger-live/pull/9175#issuecomment-2677903080
It adds a test suite on api that wasn't present before, so I would say 0% is a bit harsh

hedi-edelbloute (approved):

PR #9201, opened 2025-02-12
TEST - try caching an echo statement 

angusbayley:
...and demonstrate recalled from cache by turbo.

This PR is a CI test only/should not be reviewed or merged.

It should
- [x] trigger the new turbo-test.yml workflow
- [x] that workflow should use turbo to run the new echo command in desktop
- [x] on second run, it should restore that echo statement from cache (not run it fresh)


angusbayley:
This worked. Cached echo [here](https://github.com/LedgerHQ/ledger-live/actions/runs/13284552144/job/37090240216#step:5:36)

PR #9202, opened 2025-02-12
feat(LIVE-16730): enable Swap deeplinks on LLM

ejaycoleman:
### ✅ Checklist

  - ...

### 📝 Description

To enable swap app LLM deeplinks, two problems needed to be fixed:
- an infinite useEffect loop within the useDappAccountLogic function
- the deeplink parameter has different casing on LLM (`customDappURL`) compared to LLD (`customDappUrl`) 

Kapture 2025-02-17 at 10 14 12

### ❓ Context

- **JIRA or GitHub link**: 
[LIVE-16730](https://ledgerhq.atlassian.net/browse/LIVE-16730) Prevent infinite useEffect loop on Swap deeplinks 

---

### DIFF
--- a/apps/ledger-live-mobile/src/components/RootNavigator/types/BaseNavigator.ts
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/types/BaseNavigator.ts
@@ -117,6 +117,7 @@ export type BaseNavigatorStackParamList = {
     defaultCurrencyId?: string;
     defaultTicker?: string;
     customDappURL?: string;
+    customDappUrl?: string;
     uri?: string;
     requestId?: string;
     sessionTopic?: string;
--- a/apps/ledger-live-mobile/src/screens/Platform/LiveApp.tsx
+++ b/apps/ledger-live-mobile/src/screens/Platform/LiveApp.tsx
@@ -21,7 +21,8 @@ type Props = StackNavigatorProps<BaseNavigatorStackParamList, ScreenName.Platfor
 
 export function LiveApp({ route }: Props) {
   const { theme } = useTheme();
-  const { platform: appId, ...params } = route.params || {};
+  // For consistency with LLD, customDappUrl casing should be allowed for Swap
+  const { platform: appId, customDappURL, customDappUrl, ...params } = route.params || {};
   const { setParams } = useNavigation<Props["navigation"]>();
   const localManifest = useLocalLiveAppManifest(appId);
   const remoteManifest = useRemoteLiveAppManifest(appId);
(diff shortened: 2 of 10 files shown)


CremaFR (approved):

apps/ledger-live-mobile/src/screens/Platform/LiveApp.tsx
```diff
@@ -21,7 +21,7 @@ type Props = StackNavigatorProps<BaseNavigatorStackParamList, ScreenName.Platfor
 
 export function LiveApp({ route }: Props) {
   const { theme } = useTheme();
-  const { platform: appId, customDappURL, ...params } = route.params || {};
+  const { platform: appId, ...params } = route.params || {};
```
Justkant, line 24: Careful adding stuff back in the `params` rest value, it's only meant for the `inputs` passed to the webview, please revert the change and add the `customDappUrl` here too

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
```diff
@@ -107,6 +107,10 @@ function useDappAccountLogic({
   }, [currentAccount, accounts]);
 
   const firstAccountAvailable = useMemo(() => {
+    // Return an account for manifests with wildcard currencyIds
+    // Prevents infinite useEffect loop seen on Swap deep links
+    if (currencyIds.includes("**")) return accounts[0];
```
Justkant, line 112: This is not a proper fix IMO
- First currencyIds could also include a single `*`
- Second we don't select a proper account a could actually select an account that is not supported
- Third if there is no `firstAccountAvailable` it should just work already and be supported with the current logic and if that's not the case something else broke recently when we did some changes like this one (https://github.com/LedgerHQ/ledger-live/commit/5f579300621c8f72679f4d5d4d043bda394d9e31)

Justkant (changes requested):

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
ejaycoleman: thanks, ill investigate what changed from that commit, but a couple of questions:
- What does the wildcard currencyId imply? Does it not mean all accounts are supported? that's why I assumed it would be sufficient to return the first account with no additional checks
- Also `["**"]` comes from usePermission so I assumed it was the only possible value in the case of the wildcard?

Justkant: Thanks for the details, I think we can move with this for now and investigate why the `useEffect` below loops forever in that case
I suspect a missing `setInitialAccountSelected` but better to keep it as is for now to avoid breaking other stuff here as we clearly miss automated tests to cover us

Justkant (dismissed):

Justkant (approved):

CremaFR (approved):

apps/ledger-live-mobile/src/components/RootNavigator/types/BaseNavigator.ts
```diff
@@ -117,6 +117,7 @@ export type BaseNavigatorStackParamList = {
     defaultCurrencyId?: string;
     defaultTicker?: string;
     customDappURL?: string;
+    customDappUrl?: string;
```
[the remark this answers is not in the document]
ejaycoleman: I think it has to be because it's case-sensitive

PR #9206, opened 2025-02-12
bugfix: re-introduce SpamFilter for TX (Nfts)

mcayuelas-ledger:
### ✅ Checklist

  - ...

### 📝 Description

How it works : 
Screenshot 2025-02-12 at 15 33 39

🏛️ We have a centralised hook in common for reuse in LLM and LLD
- a hook to retrieve `NFT_IN` tx only
- a hook to filter and hide spam (hiddenCollections object)
- a hook to ‘rebuild’ the data 

ℹ️ Key points: 
- Add spamScore to collections and NFTs directly via our Backend
- We take advantage of the network lib cache + the backend cache

🤓 N.B: There are several network calls that go out because we need to process each NFT (collection) that we have in the operations
The calls go out as soon as the operations (which are recalculated to filter the operations on the NFTs) are ready, which is why we sometimes see 5-6 requests go out. 

**Demo :** 

https://github.com/user-attachments/assets/6b6536a5-d11c-459a-94ae-15b34c08458f

And then added in `hiddenCollections`

https://github.com/user-attachments/assets/11655c5d-9e0d-4027-bbdf-8db0f34d854c

### ❓ Context

- **JIRA or GitHub link**:[LIVE-15621] 

---

[LIVE-15621]: https://ledgerhq.atlassian.net/browse/LIVE-15621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/live-nft-react/src/NftMetadataProvider.tsx
mcayuelas-ledger, line 1: Needed to handle batch like **useNftMetadataBatch**

Maybe it's not the best way to do but I just copied hook above and adapt it for Collection

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +117,46 @@ export function useNftCollectionMetadata(
...
+    const key = hasAllProperties(item) ? getNftCollectionKey(item.contract!, item.currencyId!) : "";
+    const cachedData = cache[key] as NFTResource<
+      NonNullable<NFTCollectionMetadataResponse["result"]>
+    >;
+    return { key, cachedData };
+  });
+
+  useEffect(() => {
+    data.forEach(({ key, cachedData }) => {
+      const item = items.find(
```
thesan, line 141: shouldn't this start by:
```ts
if (cachedData && !isOutdated(cachedData)) return;
```
instead of going though the find first ?

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +117,46 @@ export function useNftCollectionMetadata(
...
+    contract: string | undefined;
+    currencyId: string | undefined;
+  }>,
+): Array<NFTResource<NonNullable<NFTCollectionMetadataResponse["result"]>>> {
+  const { cache, loadCollectionMetadata } = useContext(NftMetadataContext);
+
+  const hasAllProperties = (item: { contract?: string; currencyId?: string }) =>
+    item.contract && item.currencyId;
+
+  const data = items.map(item => {
```
thesan, line 131: the map will create a new object on each render which invalidate the useEffect dependencies. I think it should be wrapped in a `useMemo(..., [items])`

libs/live-nft-react/src/hooks/useSpamTxFiltering.ts
```diff
@@ -0,0 +1,130 @@
...
+  markNftAsSpam?: (collectionId: string, blockchain: BlockchainsType, spamScore: number) => void,
+  threshold?: number,
+) {
+  const nftOperations = useNftOperations({
+    accountsMap,
+    groupedOperations,
+    spamFilteringTxEnabled,
+  });
+
+  const metadatas = useNftCollectionMetadataBatch(Object.values(nftOperations));
```
thesan, line 115: `Object.values` will also create a copy on each render. So I think this should be:
```suggestion
  const metadatas = useNftCollectionMetadataBatch(useMemo(() => Object.values(nftOperations), nftOperations));
```

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +117,46 @@ export function useNftCollectionMetadata(
...
+        i => hasAllProperties(i) && getNftCollectionKey(i.contract!, i.currencyId!) === key,
+      );
+      if (!item) return;
+      if (!cachedData || isOutdated(cachedData)) {
+        loadCollectionMetadata(item.contract!, item.currencyId!);
+      }
+    });
+  }, [items, loadCollectionMetadata, cache, data]);
+
+  return items.map(item => {
```
thesan, line 151: IMO hooks are expected to return memoized values, which can be safely used as dependencies without creating unnecessary re-renders. So I think the returned value should also be wrapped in a `useMemo`

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +117,46 @@ export function useNftCollectionMetadata(
...
+export function useNftCollectionMetadataBatch(
+  items: Array<{
+    contract: string | undefined;
+    currencyId: string | undefined;
+  }>,
+): Array<NFTResource<NonNullable<NFTCollectionMetadataResponse["result"]>>> {
+  const { cache, loadCollectionMetadata } = useContext(NftMetadataContext);
+
+  const hasAllProperties = (item: { contract?: string; currencyId?: string }) =>
+    item.contract && item.currencyId;
```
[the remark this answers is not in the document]
mcayuelas-ledger: Nice catch :)

libs/live-nft-react/src/hooks/useSpamTxFiltering.ts
```diff
@@ -0,0 +1,133 @@
...
+                const account = accountsMap[operation.accountId];
+                const contract = operation.contract || "";
+                acc[contract] = {
+                  contract,
+                  currencyId: getAccountCurrency(account).id,
+                };
+              }
+              return acc;
+            }, {} as NftOperations) || ({} as NftOperations)
+        : ({} as NftOperations),
```
thesan, line 47: Can't we get rid of all the `as` by typing the functions passed to the useMemo and the reducer:
```ts
return useMemo(
  (): NftOperations =>
```
```ts
.reduce((acc: NftOperations, operation): NftOperations => {
```
It should also make the ` || {}` part not needed:
```suggestion
            }, {})
        : {},
```

libs/live-nft-react/src/hooks/useSpamTxFiltering.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export function filterSection(
+  operations: Operation[],
+  metadataMap: Map<string | undefined, number>,
+  accountsMap: AccountMap,
+  markNftAsSpam?: (collectionId: string, blockchain: BlockchainsType, spamScore: number) => void,
+  threshold?: number,
+) {
+  const filtered: Operation[] = [];
```
thesan, line 59: Why not just using a Map from the start:
```suggestion
  const filtered = new Map<string, Operation>();
```
and instead of `filtered.push(item)`: `filtered.set(item.id, item)`

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -41,157 +28,113 @@ const ShowMore = styled(Box).attrs(() => ({
...
+          <TableHeader
+            title={title}
+            titleProps={{
+              "data-e2e": "dashboard_OperationList",
+            }}
+          />
+        )}
+        {groupedOperations?.sections?.map(
+          group =>
+            group.data.length > 0 && (
```
themooneer, line 80: to avoid a condition on each and every group, i would use a filtered sections before. 
Maybe not on the component level, i think here `useOperationsList` when you build  the `groupedOperations`, make sure the sections in it (if you don't use) don't contain empty data group.

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -41,157 +28,113 @@ const ShowMore = styled(Box).attrs(() => ({
...
-                      parentAccount={parentAccount}
-                      key={`${account.id}_${operation.id}`}
-                      onOperationClick={this.handleClickOperation}
-                      t={t}
-                      withAccount={withAccount}
-                      editable={account && isEditableOperation({ account: mainAccount, operation })}
-                    />
-                  );
-                })}
+                    let parentAccount;
```
themooneer, line 90: pay attention here to `Shadow variable declaration` or `variable shaddowing`, you have already a declared `parentAccount`. (if it's a legacy code, please can you put a comment here , anyways sonarqube will point it)

apps/ledger-live-desktop/src/renderer/components/OperationsList/useOperationsList.ts
```diff
@@ -0,0 +1,120 @@
...
+export function useOperationsList({
+  account,
+  parentAccount,
+  accounts,
+  withSubAccounts,
+  filterOperation,
+}: Props) {
+  const spamFilteringTxFeature = useFeature("lldSpamFilteringTx");
+  const nftsFromSimplehashFeature = useFeature("nftsFromSimplehash");
+  const thresold = Number(nftsFromSimplehashFeature?.params?.threshold);
```
themooneer, line 52: no fallback in case of `nftsFromSimplehashFeature` is undefined ?

apps/ledger-live-desktop/src/renderer/components/OperationsList/useOperationsList.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  //both features must be enabled to enable spam filtering
+  const spamFilteringTxEnabled =
```
themooneer, line 55: `nftsFromSimplehashFeature?.enabled && spamFilteringTxFeature?.enabled` is sufficient expression

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -41,157 +28,113 @@ const ShowMore = styled(Box).attrs(() => ({
...
+                    if (account.type !== "Account") {
+                      const pa =
+                        accountsMap[account.parentId] ||
+                        allAccounts?.find(a => a.id === account.parentId);
+                      if (pa && pa.type === "Account") {
+                        parentAccount = pa;
+                      }
+                      if (!parentAccount) {
```
themooneer, line 98: this return early can be done way before this line (if not possible to fixed it now, i know it's legacy, you can put a TODO comment please, it will be easy to detect by sonar)

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
mcayuelas-ledger: Yes it's legacy code :) 
i'll comment it

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -41,157 +28,113 @@ const ShowMore = styled(Box).attrs(() => ({
     text-decoration: underline;
   }
 `;
-const mapDispatchToProps = {
-  openModal,
-};
+
 type Props = {
```
themooneer, line 32: it's the same as `apps/ledger-live-desktop/src/renderer/components/OperationsList/useOperationsList.ts` well exported here , you can import it i guess

apps/ledger-live-desktop/src/renderer/components/OperationsList/useOperationsList.ts
```diff
@@ -0,0 +1,120 @@
...
+  accounts?: AccountLike[];
+  withAccount?: boolean;
+  withSubAccounts?: boolean;
+  title?: string;
+  filterOperation?: (b: Operation, a: AccountLike) => boolean;
+};
+
+const INITIAL_TO_SHOW = 20;
+
+function usePagination(initialCount: number) {
```
themooneer, line 32: it's a very interesting custom hook that can be externalised in the newArch shared hooks.

mcayuelas-ledger: Indeed

libs/live-nft-react/src/hooks/useSpamTxFiltering.ts
mcayuelas-ledger: Indeed, I just saw that while reading code x)

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +122,54 @@ export function useNftCollectionMetadata(
   }
 }
 
+export function useNftCollectionMetadataBatch(
+  items: Array<Item>,
+): Array<NFTResource<NonNullable<NFTCollectionMetadataResponse["result"]>>> {
+  const { cache, loadCollectionMetadata } = useContext(NftMetadataContext);
+
+  const hasAllProperties = (item: Item): item is Required<Item> =>
+    !!item.contract && !!item.currencyId;
```
themooneer, line 131: you didn't check on `tokenId` ?

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +122,54 @@ export function useNftCollectionMetadata(
...
+  );
+
+  useEffect(() => {
+    data.forEach(({ key, cachedData }) => {
+      if (cachedData && !isOutdated(cachedData)) return;
+      const item = items.find(
+        i => hasAllProperties(i) && getNftCollectionKey(i.contract, i.currencyId) === key,
+      );
+      if (!item) return;
+      hasAllProperties(item) && loadCollectionMetadata(item.contract, item.currencyId);
```
themooneer, line 154: if block is more readable here

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +122,54 @@ export function useNftCollectionMetadata(
...
+    });
+  }, [items, loadCollectionMetadata, cache, data]);
```
themooneer, line 156: these dependencies scare me, but cachedData can redirect the performance issue that can be here , nice approach

libs/live-nft-react/src/NftMetadataProvider.tsx
```diff
@@ -117,6 +122,54 @@ export function useNftCollectionMetadata(
...
+
+  return useMemo(
```
themooneer, line 158: is this a dupplication of line 133 ? if so you can refactor it

libs/live-nft-react/src/NftMetadataProvider.tsx
mcayuelas-ledger: Not needed for Collection Items :)

libs/live-nft-react/src/NftMetadataProvider.tsx
mcayuelas-ledger: I was thinking the same, I'll try to check if it's possible

themooneer (commented):
A very nice to review PR ! 👏

apps/ledger-live-desktop/src/renderer/components/OperationsList/useOperationsList.ts
mcayuelas-ledger: not sufficient because`spamFilteringTxEnabled ` can undefined with your expression :)

apps/ledger-live-desktop/src/renderer/components/OperationsList/index.tsx
```diff
@@ -41,157 +28,113 @@ const ShowMore = styled(Box).attrs(() => ({
...
-                      withAccount={withAccount}
-                      editable={account && isEditableOperation({ account: mainAccount, operation })}
-                    />
-                  );
-                })}
+                    let parentAccount;
+                    if (account.type !== "Account") {
+                      const pa =
+                        accountsMap[account.parentId] ||
+                        allAccounts?.find(a => a.id === account.parentId);
```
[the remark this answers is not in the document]
mcayuelas-ledger: Indeed I put a comment on this, normally it will be not that much but you're right 

Should happen only in few edge case

libs/live-nft-react/src/NftMetadataProvider.tsx
mcayuelas-ledger: Yes indeed I'm not a huge fan of this 
I'll try to migrate it with TanStackQuery

PR #9222, opened 2025-02-13
ci: Adjust concurrency to 25% to avoid memory issues

lewisd5:
Reducing memory pressure on runners during test to avoid inside-container flakes (such as https://github.com/LedgerHQ/ledger-live/actions/runs/13310399962/job/37171255396)


angusbayley (approved):
On [agreement](https://ledger.slack.com/archives/C07KP3A8CNA/p1739462407138599?thread_ts=1739460644.834319&cid=C07KP3A8CNA) that @lewisd5 will monitor the changes, LGTM 👍

PR #9274, opened 2025-02-18
CI: reliable build desktop failures

lewisd5:
Currently successful mac builds are being misreported in certain cases. This fix ensures that the individual build steps fail and the reporting step passes

Partial failure (linux) https://github.com/LedgerHQ/ledger-live/actions/runs/13397330840
Partial failure (mac) https://github.com/LedgerHQ/ledger-live/actions/runs/13397562809
Timeout: https://github.com/LedgerHQ/ledger-live/actions/runs/13397620348/job/37420477018

Problem job: https://github.com/LedgerHQ/ledger-live/actions/runs/13291952727/job/37145026465?pr=9209

### DIFF
--- a/.github/workflows/build-desktop-reusable.yml
+++ b/.github/workflows/build-desktop-reusable.yml
@@ -43,7 +43,6 @@ jobs:
       win: ${{ steps.status.outputs.win }}
       mac: ${{ steps.status.outputs.mac }}
 
-    continue-on-error: true
     env:
       NODE_OPTIONS: "--max-old-space-size=7168"
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-desktop-reusable.yml
```diff
@@ -43,7 +43,6 @@ jobs:
       win: ${{ steps.status.outputs.win }}
       mac: ${{ steps.status.outputs.mac }}
 
-    continue-on-error: true
```
angusbayley, line 46: Does this mean that we don't want the report step to go ahead if the build doesn't complete?

.github/workflows/build-desktop-reusable.yml
```diff
@@ -113,14 +112,17 @@ jobs:
           private_key: ${{ secrets.GH_BOT_PRIVATE_KEY }}
 
       - uses: actions/download-artifact@v4
+        if: ${{ needs.build-desktop-app.outputs.linux == 'success' }}
```
angusbayley, line 115: Confirming my understanding:

- this previously causing an error at this point? So that the report creation was being halted?
- And now we do create the report but we correctly include in the report that the build for linux was a failure?

.github/workflows/build-desktop-reusable.yml
angusbayley, line 46: Does this mean that we don't want the report step to go ahead if the build doesn't complete? Can you explain the thinking?

angusbayley (commented):
Couple of questions.

Also, have you got an example a workflow run displaying the problem this PR solves?

.github/workflows/build-desktop-reusable.yml
lewisd5: The if statement on the report means it always goes ahead irrespective. This just prevents the job being shown as succeeded when it failed

angusbayley (approved):

PR #9335, opened 2025-02-25
Verified commits in all remaining parts of the release workflows

angusbayley:
Ticket [here](https://ledgerhq.atlassian.net/browse/LIVE-17241)

WIP


.github/workflows/release-prerelease.yml
```diff
@@ -87,30 +87,42 @@ jobs:
         with:
           from_level: major
           to_level: patch
+      - name: Add changed files
```
angusbayley, line 90: does this step still require the Setup Git User step? If not, remove?

lewisd5 (approved):

PR #9336, opened 2025-02-25
:thread: (llm): optimize launch

LucasWerey:
### ✅ Checklist

  - ledger store

### 📝 Description

Rework the way we initialise the ledger store to make it faster. 

#### Today 
Previously, the code was executing promises sequentially using await statements one after another:

```
const bleData = await getBle();
const settingsData = await getSettings();
const cachedCurrencyIds = await listCachedCurrencyIds();
```

This approach was slower as each promise had to complete before the next one could start.

#### Promise concurrency 

The new implementation uses promise concurrency by starting all promises simultaneously:

```
const {getBle, getSettings, listCachedCurrencyIds} = await Promise.all([
        retry(getBle, MAX_RETRIES, RETRY_DELAY),
        retry(getSettings, MAX_RETRIES, RETRY_DELAY),
        retry(listCachedCurrencyIds, MAX_RETRIES, RETRY_DELAY),
])
```

This way we parallelise promise execution and reduce init time.  Also a retry function is added so we ensure to retry the init if needed max 3 times until an error is send

documentation : https://www.builder.io/blog/promises

first POC : https://github.com/LedgerHQ/ledger-live/pull/8866

#### TLDR
```
// Sequential approach
const bleData = await getBle();          // Wait 100ms
const settings = await getSettings();     // Wait another 100ms
const accounts = await getAccounts();     // Wait another 100ms
// Total time: ~300ms

// Concurrent approach
const blePromise = getBle();             // Start immediately
const settingsPromise = getSettings();    // Start immediately
const accountsPromise = getAccounts();    // Start immediately
// Total time: ~100ms
``` 

#### Results

The start up time will of course differs on accounts number but we can expect a reduction of ~30% on the initialisation of the Store Provider. In the videos below I've used the same accounts and as we can see promise concurrency is quite faster. 

The risks should be minor since we only parallelise execution but we still do all of them before starting the app

How did I measure ? => by adding a start time at the beginning of the init method and a stop time at the end. For more precision we could do as it's done in the POC https://github.com/LedgerHQ/ledger-live/pull/8866

This measurement will depends on many conditions to be fully reliable (accounts number etc...) but it's still an improvement

|https://github.com/user-attachments/assets/bfdd583a-134f-46b0-a8bc-04ee18297d02|https://github.com/user-attachments/assets/2dfb4db2-650f-4869-8306-cb7d4567fcdc|

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17265](https://ledgerhq.atlassian.net/browse/LIVE-17265)

---

[LIVE-17265]: https://ledgerhq.atlassian.net/browse/LIVE-17265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/context/LedgerStore.tsx
+++ b/apps/ledger-live-mobile/src/context/LedgerStore.tsx
@@ -1,6 +1,6 @@
-import React, { Component } from "react";
+import React, { useEffect, useState, ReactNode, useCallback } from "react";
 import { Provider } from "react-redux";
-import { type StoreType } from "./store";
+import { Store } from "redux";
 import { importPostOnboardingState } from "@ledgerhq/live-common/postOnboarding/actions";
 import { CounterValuesStateRaw } from "@ledgerhq/live-countervalues/types";
 import { findCryptoCurrencyById } from "@ledgerhq/live-common/currencies/index";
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/context/LedgerStore.tsx
```diff
@@ -41,22 +41,34 @@ export default class LedgerStoreProvider extends Component<
...
   componentDidMount() {
     return this.init();
   }
-
   componentDidCatch(e: Error) {
     console.error(e);
     throw e;
   }
 
   async init() {
```
themooneer, line 52: I saw your PR description and especially the following section : 
```
The new implementation uses promise concurrency by starting all promises simultaneously:

const blePromise = getBle();
const settingsPromise = getSettings();
const cachedCurrencyIdsPromise = listCachedCurrencyIds();
const supportedFiatsPromise = listSupportedFiats();

```

it's true, but you need to pay attention here, if you start them all like this, while awaiting you need to wrap it up with a try catch block, because when one of your promise is `rejected` you will `await` the error, and it will crash your app. 

Second, catching error is good, but a retry strategy is better too. 

So for instance `getSettings();` is rejected, it should reter a timeout in order to have the expected settings object. This will add a resilience layer to your implementation in addition to the performance gain. 

To experiment a real scenario, try to explicitly `throw new Error("blabla")` inside one of your parallel promises. and you will see the app behavior. 

I hope it helps

apps/ledger-live-mobile/src/context/LedgerStore.tsx
```diff
@@ -71,71 +83,67 @@ export default class LedgerStoreProvider extends Component<
...
       settingsData.counterValue &&
-      !supportedCV.find(({ ticker }) => ticker === settingsData.counterValue)
+      !supportedCounterValues.find(({ ticker }) => ticker === settingsData.counterValue)
     ) {
       settingsData.counterValue = settingsState.counterValue;
     }
 
     this.props.store.dispatch(importSettings(settingsData));
-    const accountsData = await getAccounts();
+
```
themooneer, line 115: in the await used in lines 116 / 119 / 129 / 134 / 139 / 145 , you just moved the sequential behaviour that pauses the thread for each and every line. And i don't know if you gain time by moving the awaits here. 
In the other hand, you can use promise All for all those awaits: 

```

// retry helpers 

function delay(ms: number) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

async function retry<T>(fn: () => Promise<T>, retries: number, delayMs: number): Promise<T> {
  for (let i = 0; i < retries; i++) {
    try {
      return await fn();
    } catch (error) {
      if (i === retries - 1) {
        throw error;
      }
      await delay(delayMs);
    }
  }
  throw new Error("Max retries reached");


  const [accountsData, postOnboardingState, marketState, trustchainStore, walletStore, protect, initialCountervalues] = await Promise.all([
    retry(accountsPromise, 3, 1000),
    retry(postOnboardingStatePromise, 3, 1000),
    retry(marketStatePromise, 3, 1000),
    retry(trustchainStorePromise, 3, 1000),
    retry(walletStorePromise, 3, 1000),
    retry(protectPromise, 3, 1000),
    retry(initialCountervaluesPromise, 3, 1000),
  ]);

```

PS yu can tweak the timeout , 500 ms is sufficient in web env, to see in mobile env

themooneer: sorry if there are some mini syntax errors

apps/ledger-live-mobile/src/context/LedgerStore.tsx
```diff
@@ -26,57 +26,80 @@ import { importMarket } from "~/actions/market";
...
+        supportedFiats,
+        accountsData,
+        postOnboardingState,
+        marketState,
+        trustchainStore,
+        walletStore,
+        protect,
+        initialCountervalues,
+      ] = await Promise.all([
+        retry(getBle, 3, 500),
```
themooneer, line 69: you can store 3 and 500 timeouts in a top constant

LucasWerey: Yep, I pushed this to make another PR I will update it! Thanks 🙏

themooneer (approved):
LGTM

PR #9357, opened 2025-02-26
:bug: fix(lld) add a check before size and position events

LucasWerey:
### ✅ Checklist

  - window events

### 📝 Description

We need to ensure that the app well exists before checking its size or position to avoid typeError in the case the user closes the app during the debounce effect

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17339](https://ledgerhq.atlassian.net/browse/LIVE-17339)

---

[LIVE-17339]: https://ledgerhq.atlassian.net/browse/LIVE-17339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/main/index.ts
+++ b/apps/ledger-live-desktop/src/main/index.ts
@@ -150,6 +150,7 @@ app.on("ready", async () => {
   window.on(
     "resize",
     debounce(() => {
+      if (!window || window.isDestroyed()) return;
       const [width, height] = window.getSize();
       db.setKey("windowParams", `${window.name}.dimensions`, {
         width,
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/main/index.ts
```diff
@@ -150,6 +150,7 @@ app.on("ready", async () => {
   window.on(
     "resize",
     debounce(() => {
+      if (!window || window.isDestroyed()) return;
```
themooneer, line 153: how that can happen ? an event listener on undefined window instance ?

LucasWerey: I don't have the exact reason, but all the events on Sentry have the same issue: the typerError is triggered during the app close. I guess it could be that the app is still running until it is fully off but the window is already destroyed. And since we have a debounce of 300ms it could lead to this issue. Otherwise, I don't see any reason to have an issue on this part precisely a TyperError

themooneer (approved):

PR #9374, opened 2025-02-27
chore(ldmk-in-llm): add featureFlag

fAnselmi-Ledger:
### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/live-dmk-mobile/src/hooks/useBleDevicesScanning.ts
```diff
@@ -4,7 +4,7 @@ import { useDeviceManagementKit } from "./useDeviceManagementKit";
 import { Device } from "@ledgerhq/types-devices";
 import { dmkToLedgerDeviceIdMap } from "@ledgerhq/live-dmk-shared";
 
-const defaultMapper = (device: DiscoveredDevice): Device => ({
+export const useBleDevicesScanningDefaultMapper = (device: DiscoveredDevice): Device => ({
```
jdabbech-ledger, line 7: no need to export as this is the default value, you should probably change the hooks arguments to have separate var instead of an object

apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
```diff
@@ -88,7 +92,19 @@ export default function SelectDevice({
...
+  const { scannedDevices: DMKscannedDevices } = useDMKBleDevicesScanning({
+    mapper: useBleDevicesScanningDefaultMapper,
+    enabled: isLDMKEnabled,
+  });
+  const { scannedDevices: PREDMKscannedDevices } = useBleDevicesScanning({
+    bleTransportListen: getBLETransport({ isLDMKEnabled }).listen,
+    stopBleScanning,
+    enabled: !isLDMKEnabled,
+  });
+  const scannedDevices = isLDMKEnabled ? DMKscannedDevices : PREDMKscannedDevices;
```
jdabbech-ledger, line 107: [naming] `legacy` instead of `predmk`

libs/live-dmk-mobile/src/hooks/useBleDevicesScanning.ts
fAnselmi-Ledger: thanks, this is a left-over from a prev. implementation

apps/ledger-live-mobile/src/react-native-hw-transport-ble/index.ts
```diff
@@ -36,6 +50,15 @@ const transport = Config.MOCK
...
-    })
-  : DeviceManagementKitTransport;
-export default transport;
+    });
+  } else {
+    // when not in MOCK mode, return DeviceManagementKitTransport if enabled,
+    // otherwise BleTransport. Asserting the type of DeviceManagementKitTransport
+    // to match BleTransport.
+    return isLDMKEnabled
+      ? (DeviceManagementKitTransport as unknown as typeof BleTransport)
```
valpinkman, line 59: [WOULD] I don't think it's a good idea to force a type to this transport

apps/ledger-live-mobile/src/screens/PairDevices/index.tsx
```diff
@@ -203,7 +206,9 @@ function PairDevicesInner({ navigation, route }: NavigationProps) {
         } finally {
           transport.close();
```
valpinkman, line 207: [NOTE]: This part here should probably be disabled in the context of ldmk as we don't need to close or disconnect the device.

valpinkman (approved):

apps/ledger-live-mobile/src/react-native-hw-transport-ble/index.ts
ofreyssinet-ledger: [SHOULD] +1
this is not a good idea, and it will already break, as you can see as `observeState` (called in another file) is not defined on `DeviceManagementKitTransport`

PR #9438, opened 2025-03-05
ci: exit prerelease, manually fully exiting this time

angusbayley:
https://github.com/LedgerHQ/ledger-live/pull/9435 did not fully exit the prerelease, instead it prepared the prerelease for exiting - exiting only happens on subsequent versioning, which we don't want.

This PR instead manually backs out of the prerelease state by reverting the prerelease version names and deleting the `pre.json` file.


angusbayley:
This problem was solved manually outside this PR. Closing

PR #9440, opened 2025-03-05
Feat/tron alpaca estimate fees

qperrot:
### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17432](https://ledgerhq.atlassian.net/browse/LIVE-17432)

---

### DIFF
--- a/libs/coin-framework/src/api/types.ts
+++ b/libs/coin-framework/src/api/types.ts
@@ -31,6 +31,16 @@ export type Asset = {
   native: bigint;
 };
 
+export type Intent = {
+  sender: string;
+  recipient: string;
+  amount: bigint;
+  contractAddress?: string;
+  standard?: "trc10" | "trc20";
+};
+
+export type Fees = { [feeOption: string]: bigint };
+
 // TODO rename start to minHeight
 //       and add a `token: string` field to the pagination if we really need to support pagination
 //       (which is not the case for now)
@@ -42,7 +52,7 @@ export type Api = {
   broadcast: (tx: string) => Promise<string>;
   combine: (tx: string, signature: string, pubkey?: string) => string;
   craftTransaction: (address: string, transaction: Transaction, pubkey?: string) => Promise<string>;
-  estimateFees: (addr: string, amount: bigint) => Promise<bigint>;
+  estimateFees: (intent: Intent) => Promise<Fees>;
   getBalance: (address: string) => Promise<Asset | bigint>;
   lastBlock: () => Promise<BlockInfo>;
   listOperations: (address: string, pagination: Pagination) => Promise<[Operation[], string]>;
--- a/libs/coin-modules/coin-tezos/src/api/index.integ.test.ts
+++ b/libs/coin-modules/coin-tezos/src/api/index.integ.test.ts
@@ -36,7 +36,7 @@ describe("Tezos Api", () => {
       const amount = BigInt(100);
 
       // When
-      const result = await module.estimateFees(address, amount);
+      const result = await module.estimateFees({ sender: address, recipient: "address", amount });
 
       // Then
       expect(result).toEqual(BigInt(287));
--- a/libs/coin-modules/coin-tezos/src/api/index.ts
+++ b/libs/coin-modules/coin-tezos/src/api/index.ts
@@ -1,5 +1,7 @@
 import {
+  Fees,
   IncorrectTypeError,
+  Intent,
   Operation,
   Pagination,
   type Api,
@@ -16,8 +18,8 @@ import {
   lastBlock,
   rawEncode,
 } from "../logic";
-import api from "../network/tzkt";
 import { log } from "@ledgerhq/logs";
+import api from "../network/tzkt";
 
 export function createApi(config: TezosConfig): Api {
   coinConfig.setCoinConfig(() => ({ ...config, status: { type: "active" } }));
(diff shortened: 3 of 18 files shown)


libs/coin-framework/src/api/types.ts
```diff
@@ -31,6 +31,16 @@ export type Asset = {
...
 
+export type Intent = {
+  sender: string;
+  recipient: string;
+  amount: bigint;
+  contractAddress?: string;
+  standard?: "trc10" | "trc20";
+};
+
+export type Fees = { [feeOption: string]: bigint };
```
francois-guerin-ledger, line 42:
```suggestion
type FeeOption = "standard" | "gross";
export type Fees = Partial<Record<FeeOption, bigint>>;
```

Ideally we would like to control the type of keys.

francois-guerin-ledger (approved):

libs/coin-framework/src/api/types.ts
hedi-edelbloute: Fee option will be scoped for each coin module

I don't think we want to list keys for now

libs/coin-modules/coin-tron/src/logic/estimateFees.ts
```diff
@@ -1,3 +1,14 @@
-export function estimateFees(_address: string, _amount: bigint): Promise<bigint> {
-  throw Error("Not implemented yet");
+import { ACTIVATION_FEES, ACTIVATION_FEES_TRC_20 } from "./constants";
+import type { Intent, Fees } from "@ledgerhq/coin-framework/api/index";
+
+export async function estimateFees(intent: Intent): Promise<Fees> {
```
sprohaszka-ledger, line 4: When I am looking at the `bridge/getEstimateFees` the fess computation seems more complex.
I don't understand the `gross` notion here 🤔

sprohaszka-ledger (changes requested):
I didn't check everything, because there are 2 points that raise some questions:
* the fees computation seems very light considering what is done in the `bridge` implementation
* the `gross` notion that I don't understand. If this is something specific to Tron, it should be "known" by the CoinFmk

libs/coin-modules/coin-tron/src/logic/estimateFees.ts
hedi-edelbloute: See slack message in vault channel

libs/coin-framework/src/api/types.ts
```diff
@@ -31,6 +31,14 @@ export type Asset = {
   native: bigint;
 };
 
+export type TransactionIntent = {
+  sender: string;
+  recipient: string;
+  amount: bigint;
+  contractAddress?: string;
+  standard?: "trc10" | "trc20";
```
sprohaszka-ledger, line 39: The type cannot be tight to Tron implementation.
The easiest way is to use `string` (not sure that Generic will be efficient here)

libs/coin-framework/src/api/types.ts
```diff
@@ -31,6 +31,14 @@ export type Asset = {
   native: bigint;
 };
 
+export type TransactionIntent = {
+  sender: string;
+  recipient: string;
+  amount: bigint;
+  contractAddress?: string;
```
sprohaszka-ledger, line 38: `contractAddress` is for the token smartcontract address?
It may be more explicit to call it `tokenAddress` even if for trc10 it is not an address.

libs/coin-modules/coin-tezos/src/api/index.ts
```diff
@@ -51,18 +52,22 @@ async function craft(
   return rawEncode(contents);
 }
 
-async function estimate(addr: string, amount: bigint): Promise<bigint> {
-  const accountInfo = await api.getAccountByAddress(addr);
+export async function estimate(transactionIntent: TransactionIntent): Promise<bigint> {
```
sprohaszka-ledger, line 55: Why do you need to export this function?

libs/coin-modules/coin-tezos/src/api/index.ts
```diff
@@ -51,18 +52,22 @@ async function craft(
...
+  const accountInfo = await api.getAccountByAddress(transactionIntent.recipient);
   if (accountInfo.type !== "user") throw new Error("unexpected account type");
 
   const estimatedFees = await estimateFees({
     account: {
-      address: addr,
+      address: transactionIntent.recipient,
```
sprohaszka-ledger, line 61: It should be the `sender` address

libs/coin-modules/coin-tron/src/logic/estimateFees.test.ts
```diff
@@ -0,0 +1,31 @@
+import { estimateFees } from "./estimateFees";
+import { ACTIVATION_FEES, ACTIVATION_FEES_TRC_20 } from "./constants";
+import { TransactionIntent } from "@ledgerhq/coin-framework/api/index";
+
+describe("estimateFees", () => {
```
sprohaszka-ledger, line 5: To be exhaustive, you can be called without a "standard" value meaning you are using the native coin (TRX).
A trc10 need to be identified with its "id" (which can be provide through the `contractAddress` field)

sprohaszka-ledger (changes requested):
Last modification regarding Tron "notion" inside CoinFmk.

libs/coin-framework/src/api/types.ts
qperrot: in the case of `Tron` we do not use it (or in any other coin modules right now) but `Intent` was defined like this in the doc. Should I keep it and just rename it `tokenAddress`

libs/coin-modules/coin-tezos/src/api/index.ts
qperrot: my bad

libs/coin-modules/coin-tezos/src/api/index.integ.test.ts
```diff
@@ -36,10 +36,15 @@ describe("Tezos Api", () => {
...
+      const result = await module.estimateFees({
+        type: "send",
+        sender: address,
+        recipient: "tz1heMGVHQnx7ALDcDKqez8fan64Eyicw4DJ",
+        amount,
+      });
 
       // Then
-      expect(result).toEqual(BigInt(287));
+      expect(result).toBeGreaterThanOrEqual(BigInt(200));
```
francois-guerin-ledger:
```suggestion
      expect(result).toBeGreaterThanOrEqual(BigInt(0));
```

francois-guerin-ledger (approved):

sprohaszka-ledger (approved):
👍  Let's move on.
We will need to do some global clean up for consistency between all the different Tron PR (ex: `TransactionIntent` to use in `craftTransaction`)

PR #9450, opened 2025-03-06
✨ LLD - Add thai language support

cgrellard-ledger:
### ✅ Checklist

  - Add thai language behind feature flag on the Welcome screen and in the settings
  - It will automatically ask to install the language on the device when switching language once provided by the API
  - External links fallback to english as they are not yet supported in thai
  - When starting the app, it will fallback to english if the user was previously in thai and the FF has been disabled

### 📝 Description

Thai language added

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17156]

---


apps/ledger-live-desktop/src/renderer/reducers/application.ts
```diff
@@ -14,7 +14,7 @@ export type ApplicationState = {
   };
 };
 const { language, region } = getParsedSystemLocale();
-const osLangSupported = LanguageIds.includes(language);
+const osLangSupported = LanguageIdsNotFeatureFlagged.includes(language);
```
mcayuelas-ledger, line 17: `LanguageIds` no more used ?

[ASK] why do we need this `LanguageIdsNotFeatureFlagged` ?

apps/ledger-live-desktop/src/renderer/reducers/settings.ts
```diff
@@ -143,7 +144,7 @@ export const getInitialLanguageAndLocale = (): { language: Language; locale: Loc
   const systemLocal = getAppLocale();
 
   // Find language from system locale (i.e., en, fr, es ...)
-  const languageId = LanguageIds.find(lang => systemLocal.startsWith(lang));
+  const languageId = LanguageIdsNotFeatureFlagged.find(lang => systemLocal.startsWith(lang));
```
mcayuelas-ledger, line 147: can't be thai?

apps/ledger-live-desktop/src/renderer/reducers/application.ts
cgrellard-ledger: `LanguageIds` is still used as it contains all languages that Ledger Live can support.
`LanguageIdsNotFeatureFlagged` is necessary due to our current implementation. For example, when opening the app for the first time and setting the initial state of the app, we get the system language of the user and set it directly as the initial value. It is done in a function so we can't use the `useFeature` hook to check if a language is enabled or not. So for the initial state of the app we have to only use the languages that are not feature flagged otherwise it could cause some issues.

apps/ledger-live-desktop/src/renderer/reducers/settings.ts
cgrellard-ledger: It can't be thai because at this point we can't know if the feature flag is enabled or not. It's a limitation of our current implementation that we can improve later on

themooneer (approved):
LGTM

mcayuelas-ledger (approved):

PR #9478, opened 2025-03-10
:adhesive_bandage: (llm) rework netinfo usage

LucasWerey:
### ✅ Checklist

  - NetInfo lib call

### 📝 Description

1- useAppStateListener is called inside index.ts and is just used to display an error message on a specific account in specific cases so we can easily delay that and do it only in this specific case and in the account component. I've cut the component into multiple ones so we don't do unnecessary calls 

2- counterValuesProvider we useNetInfo but it's not really necessary, even without network or 3G it will just perform the polling without any issue

I've cleaned the app state too because it was only use to trigger this error so not really essential to have it in the store

NetInfo is also use to trigger a fallback screen inside market and to know if a webview can be opened or not so we can't totally remove the package but we can avoid to use it too much

| Wifi off        | Wifi on         |
| <video src="https://github.com/user-attachments/assets/5e1f1e47-410a-4d15-9c4f-39600d1cefa8"> | <video src="https://github.com/user-attachments/assets/ae10e4c8-8c44-46de-b18b-e5e913489afe"> |

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17467](https://ledgerhq.atlassian.net/browse/LIVE-17467)

---

[LIVE-17467]: https://ledgerhq.atlassian.net/browse/LIVE-17467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thesan (commented):
Everything else looks good but I think Sonar has a point not to use "Error to declare a variable": https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&pullRequest=9478&id=LedgerHQ_ledger-live&open=AZWQdmd4zxZFJvQm_fTx

I also named components Error in the past, but it's true that it's a JS built-in so it's best to avoid it.

thesan (approved):

PR #9610, opened 2025-03-20
Fix/btc handle replaced txs

Wozacosta:
### ✅ Checklist

  - ...

### 📝 Description

When syncing Bitcoin accounts, it's possible to fetch multiple versions of the same transaction due to Replace-By-Fee (RBF). This can lead to:

Duplicate or conflicting operations in the UI
Displaying transactions that were never confirmed and already replaced

image

✅ What this PR does
- Introduces the `removeReplaced` function to filter out replaced/unconfirmed Bitcoin transactions
- Ensures only the most relevant version of each transaction is kept
Handles the following cases:
- Keeps confirmed transactions over unconfirmed ones
- Keeps the latest confirmed transaction (based on block height, then date)
- Keeps the newest unconfirmed transaction if none are confirmed
- Coinbase transactions are always preserved
- Transactions without inputs (e.g. OUT txs) are not affected
- Preserves the original operation order in the final result to avoid UI diff noise

This PR also adds unit tests to cover `removeReplaced`

### Before/After

https://github.com/user-attachments/assets/bfe96101-c9a2-408e-a20d-12e197bf8408

### ❓ Context

- **JIRA or GitHub link**: LIVE-16927

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/bitcoin/CoinControlRow.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/bitcoin/CoinControlRow.tsx
@@ -74,6 +74,7 @@ export const CoinControlRow = ({
   const unconfirmed = utxoStatus === "pickPendingUtxo";
   const last = !s.excluded && totalExcludedUTXOS + 1 === account.bitcoinResources?.utxos.length; // make sure that at least one utxo is selected
   const disabled = unconfirmed || last;
+  const confirmations = utxo.blockHeight ? account.blockHeight - utxo.blockHeight : 0;
   const onClick = () => {
     if (disabled) return;
     const patch = {
@@ -138,7 +139,7 @@ export const CoinControlRow = ({
         />
         {utxo.blockHeight ? (
           <Text ff="Inter|Medium" fontSize={3} color={"palette.text.shade50"}>
-            {account.blockHeight - utxo.blockHeight + " confirmations"}
+            {confirmations + ` confirmation${confirmations > 1 ? "s" : ""}`}
           </Text>
         ) : (
           <Text ff="Inter|Medium" fontSize={3} color={"alertRed"}>
--- a/libs/coin-modules/coin-bitcoin/src/logic.ts
+++ b/libs/coin-modules/coin-bitcoin/src/logic.ts
@@ -7,13 +7,14 @@ import type {
   Transaction,
   NetworkInfo,
   UtxoStrategy,
+  BtcOperation,
 } from "./types";
 import { $Shape } from "utility-types";
 import type { TX, Input as WalletInput, Output as WalletOutput } from "./wallet-btc";
 import { BigNumber } from "bignumber.js";
 import { encodeOperationId } from "@ledgerhq/coin-framework/operation";
 import type { CryptoCurrency, CryptoCurrencyId } from "@ledgerhq/types-cryptoassets";
-import type { Account, Operation, OperationType } from "@ledgerhq/types-live";
+import type { Account, OperationType } from "@ledgerhq/types-live";
 
 // correspond ~ to min relay fees but determined empirically for a tx to be accepted by network
 const minFees: Partial<Record<CryptoCurrencyId | "LBRY" | "groestcoin" | "osmo", number>> = {
@@ -169,8 +170,8 @@ export const mapTxToOperations = (
   accountId: string,
   accountAddresses: Set<string>,
   changeAddresses: Set<string>,
-): $Shape<Operation[]> => {
-  const operations: Operation[] = [];
+): $Shape<BtcOperation[]> => {
+  const operations: BtcOperation[] = [];
   const txId = tx.id;
   const fee = new BigNumber(tx.fees ?? 0);
   const blockHeight = tx.block?.height;
@@ -184,8 +185,10 @@ export const mapTxToOperations = (
   const accountInputs: WalletInput[] = [];
   const accountOutputs: WalletOutput[] = [];
   const syncReplaceAddress = perCoinLogic[currencyId]?.syncReplaceAddress;
+  const inputs = new Set<`${string}-${number}`>(); // txid-outputIndex
 
   for (const input of tx.inputs) {
+    inputs.add(`${input.output_hash}-${input.output_index}`);
     if (input.address) {
       senders.add(syncReplaceAddress ? syncReplaceAddress(input.address) : input.address);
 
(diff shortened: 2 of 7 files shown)


apps/ledger-live-desktop/src/renderer/families/bitcoin/CoinControlRow.tsx
```diff
@@ -138,7 +139,7 @@ export const CoinControlRow = ({
         />
         {utxo.blockHeight ? (
           <Text ff="Inter|Medium" fontSize={3} color={"palette.text.shade50"}>
-            {account.blockHeight - utxo.blockHeight + " confirmations"}
+            {confirmations + ` confirmation${confirmations > 1 ? "s" : ""}`}
```
palra, line 142: Is there an intl library in use handling pluralization in multiple locales?

Wozacosta: There's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules (not used anywhere else in the codebase but still is polyfilled on LLM side with@formatjs/intl-pluralrules) but ideally this thing would move on to be fully internationalized in another PR.

libs/coin-modules/coin-bitcoin/src/synchronisation.ts
```diff
@@ -52,11 +52,101 @@ const fromWalletUtxo = (utxo: WalletOutput, changeAddresses: Set<string>): Bitco
...
+          )
+        ) {
+          uniqueOperations.set(op.hash, op);
+          continue; // ✅ Skip processing further, but KEEP it
+        }
+        const existingOp = txByInput.get(input);
+        if (existingOp) {
+          const isExistingConfirmed =
+            existingOp.blockHeight !== null && existingOp.blockHeight !== undefined;
+          const isNewOpConfirmed = op.blockHeight !== null && op.blockHeight !== undefined;
```
francois-guerin-ledger, line 103:
```suggestion
          const isExistingConfirmed = typeof existingOp.blockHeight === "number";
          const isNewOpConfirmed = typeof op.blockHeight === "number";
```

libs/coin-modules/coin-bitcoin/src/types.ts
```diff
@@ -172,3 +173,8 @@ export type BitcoinAccount = Account & { bitcoinResources: BitcoinResources };
 export type BitcoinAccountRaw = AccountRaw & {
   bitcoinResources: BitcoinResourcesRaw;
 };
+
+export type BtcOperationExtra = {
+  inputs?: string[] | undefined;
```
francois-guerin-ledger, line 178:
```suggestion
  inputs: string[];
```

See my comment above

libs/coin-modules/coin-bitcoin/src/synchronisation.test.ts
```diff
@@ -0,0 +1,462 @@
...
+      date: new Date("2024-01-01"),
+      extra: {}, // No inputs
+    };
+
+    const result = removeReplaced([noInputTx]);
+    expect(result).toContain(noInputTx); // ✅ Should always be kept
+  });
+});
+
+describe("removeReplaced - Order Preservation", () => {
```
francois-guerin-ledger, line 310: Method `Array.prototype.filter` does preserve the order for sure. Maybe these tests can be merged with the ones above ?

libs/coin-modules/coin-bitcoin/src/logic.ts
```diff
@@ -184,8 +185,10 @@ export const mapTxToOperations = (
   const accountInputs: WalletInput[] = [];
   const accountOutputs: WalletOutput[] = [];
   const syncReplaceAddress = perCoinLogic[currencyId]?.syncReplaceAddress;
+  const inputs: Set<`${string}-${number}`> = new Set(); // txid-outputIndex
```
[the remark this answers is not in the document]
Wozacosta: Good taste 👍 Changed in f583161353602588bf98d3de63fe35928d88410a

libs/coin-modules/coin-bitcoin/src/synchronisation.test.ts
```diff
@@ -0,0 +1,462 @@
...
+      date: new Date("2024-01-01"),
+      extra: { inputs: ["input1"] },
+    };
+
+    const higherHeightTx: BtcOperation = {
+      ...baseTx,
+      id: "highHeight",
+      hash: "tx2",
+      blockHeight: 100, // Higher block height
+      date: new Date("2024-01-02"),
```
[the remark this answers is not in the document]
Wozacosta: Indeed, changed in f583161353602588bf98d3de63fe35928d88410a 👍

libs/coin-modules/coin-bitcoin/src/types.ts
Wozacosta: This one is problematic (undoable) as of now as: 

- Old operations have no inputs, this update is not retroactive on old operations as some information (input ids ) available on the original transaction is lost when mapping it to operations here (https://github.com/LedgerHQ/ledger-live/blob/f583161353602588bf98d3de63fe35928d88410a/libs/coin-modules/coin-bitcoin/src/logic.ts#L167). 

Not a big deal imo as fixing the behavior starting in the next release is more crucial, for users that have dangling operations they are still asked to clear cache and resync.

libs/coin-modules/coin-bitcoin/src/logic.ts
```diff
@@ -303,7 +306,7 @@ export const mapTxToOperations = (
         accountId,
         date,
         hasFailed,
-        extra: {},
+        extra: inputs.size > 0 ? { inputs: Array.from(inputs) } : {},
```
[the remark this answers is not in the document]
Wozacosta: Cleaned it up a bit in f583161353602588bf98d3de63fe35928d88410a

libs/coin-modules/coin-bitcoin/src/synchronisation.test.ts
Wozacosta: Merged 👍

libs/coin-modules/coin-bitcoin/src/synchronisation.ts
Wozacosta: Changed in latest commit 👍

francois-guerin-ledger (approved):
👍

PR #9662, opened 2025-03-25
fix(llm): 🐛 cosmos transactions failing due to undefined memo

thesan:
### ✅ Checklist

  - ...

### 📝 Description

This fix is based on [this comment](https://ledgerhq.atlassian.net/browse/LIVE-17876?focusedCommentId=535342) however I wasn't able to reproduce the bug.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17876]

---

[LIVE-17876]: https://ledgerhq.atlassian.net/browse/LIVE-17876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/families/cosmos/MemoTagInput.tsx
+++ b/apps/ledger-live-mobile/src/families/cosmos/MemoTagInput.tsx
@@ -5,8 +5,5 @@ import type { MemoTagInputProps } from "LLM/features/MemoTag/types";
 import { GenericMemoTagInput } from "LLM/features/MemoTag/components/GenericMemoTagInput";
 
 export default (props: MemoTagInputProps<CosmosTransaction>) => (
-  <GenericMemoTagInput
-    {...props}
-    valueToTxPatch={value => tx => ({ ...tx, memo: value || undefined })}
-  />
+  <GenericMemoTagInput {...props} valueToTxPatch={value => tx => ({ ...tx, memo: value })} />
 );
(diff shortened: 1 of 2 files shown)


mcayuelas-ledger (approved):

apps/ledger-live-mobile/src/families/cosmos/MemoTagInput.tsx
```diff
@@ -5,8 +5,5 @@ import type { MemoTagInputProps } from "LLM/features/MemoTag/types";
 import { GenericMemoTagInput } from "LLM/features/MemoTag/components/GenericMemoTagInput";
 
 export default (props: MemoTagInputProps<CosmosTransaction>) => (
-  <GenericMemoTagInput
-    {...props}
-    valueToTxPatch={value => tx => ({ ...tx, memo: value || undefined })}
-  />
+  <GenericMemoTagInput {...props} valueToTxPatch={value => tx => ({ ...tx, memo: value })} />
```
Justkant, line 8: I don't think this is the root cause of the issue, especially because memo can still be undefined or null looking at the types
For me the issue is probably where we use the memo and don't expect it to be undefined somehow while the type seems correct

Also looking at other `MemoTagInput` for other coins, we also use the fallback to `undefined` so I would prefer another fix if possible, especially if we cannot reproduce with or without this change

thesan: No this is **not** the root cause you're 100% right. After more tests I managed to reproduce the bug Yesterday and it has nothing to do with the memo. I updated the ticket but I forgot to close this PR sorry.

PR #9711, opened 2025-03-28
refactor(coin-tron): adapt bridge to use logic methods

Canestin:
### ✅ Checklist

  - ...

### 📝 Description

Methods to refactor : 

- sync
- broadcast
- getEstimatedFees
- signOperation

### ❓ Context

- **JIRA**: https://ledgerhq.atlassian.net/browse/LIVE-15802

---

### DIFF
--- a/libs/coin-modules/coin-tron/src/bridge/getEstimateFees.ts
+++ b/libs/coin-modules/coin-tron/src/bridge/getEstimateFees.ts
@@ -6,10 +6,10 @@ import {
   STANDARD_FEES_NATIVE,
   STANDARD_FEES_TRC_20,
 } from "../logic/constants";
-import { fetchTronAccount } from "../network";
 import type { AccountTronAPI } from "../network/types";
 import { Transaction } from "../types";
 import { extractBandwidthInfo, getEstimatedBlockSize } from "./utils";
+import { estimateFees, getAccount } from "../logic";
 
 // see : https://developers.tron.network/docs/bandwith#section-bandwidth-points-consumption
 // 1. cost around 200 Bandwidth, if not enough check Free Bandwidth
@@ -35,7 +35,7 @@ const getFeesFromAccountActivation = async (
   transaction: Transaction,
   tokenAccount?: TokenAccount | null,
 ): Promise<BigNumber> => {
-  const recipientAccounts = await fetchTronAccount(transaction.recipient);
+  const recipientAccounts = await getAccount(transaction.recipient);
   const recipientAccount: AccountTronAPI | undefined = recipientAccounts[0];
   const { gainedUsed, gainedLimit } = extractBandwidthInfo(transaction.networkInfo);
   const available = gainedLimit.minus(gainedUsed);
--- a/libs/coin-modules/coin-tron/src/bridge/getEstimateFees.ts
+++ b/libs/coin-modules/coin-tron/src/bridge/getEstimateFees.ts
@@ -1,15 +1,12 @@
 import { Account, TokenAccount } from "@ledgerhq/types-live";
 import BigNumber from "bignumber.js";
-import {
-  ACTIVATION_FEES,
-  ACTIVATION_FEES_TRC_20,
-  STANDARD_FEES_NATIVE,
-  STANDARD_FEES_TRC_20,
-} from "../logic/constants";
+import { ACTIVATION_FEES, STANDARD_FEES_NATIVE, STANDARD_FEES_TRC_20 } from "../logic/constants";
 import type { AccountTronAPI } from "../network/types";
 import { Transaction } from "../types";
 import { extractBandwidthInfo, getEstimatedBlockSize } from "./utils";
 import { estimateFees, getAccount } from "../logic";
+import type { TransactionIntent } from "@ledgerhq/coin-framework/api/index";
+import { TronToken } from "../types";
 
 // see : https://developers.tron.network/docs/bandwith#section-bandwidth-points-consumption
 // 1. cost around 200 Bandwidth, if not enough check Free Bandwidth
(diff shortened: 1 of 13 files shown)


libs/coin-modules/coin-tron/src/logic/getAccount.ts
```diff
@@ -0,0 +1,6 @@
+import { fetchTronAccount } from "../network";
+import { AccountTronAPI } from "../network/types";
+
+export async function getAccount(address: string): Promise<AccountTronAPI[]> {
+  return await fetchTronAccount(address);
+}
```
Canestin, line 6: Created this function because `fetchTronAccount`is needed in the bridge. Since we dont want to directly call network methods in the bridge. Or if no need to a have it in the common logic we can keep network calls ?

libs/coin-modules/coin-tron/src/logic/utils.ts
```diff
@@ -38,6 +39,29 @@ export async function decodeTransaction(rawTx: string): Promise<{
...
+
+export async function encodeTransaction(
+  rawData: Record<string, unknown>,
+  signature: string,
+): Promise<string> {
+  const { Transaction } = (globalThis as unknown as any).TronWebProto;
+  const transaction = new Transaction();
+  transaction.raw = convertTxToRaw(rawData);
+
+  const serializedHex = Buffer.from(transaction.raw.serializeBinary()).toString("hex");
```
Canestin, line 60: Returns an empty string here! Trying to figure out why...

libs/coin-modules/coin-tron/src/bridge/signOperation.ts
```diff
@@ -87,6 +87,7 @@ export const buildSignOperation =
       );
     });
 
+// Should we keep this here or move it?
 const prepareTransactionForSignature = (
```
Canestin, line 91: I think the way to sign a transaction is different depending on the platform (Ledger Live & Vault, for instance).
So, I don’t think this needs to be in the common logic. However, it uses network calls, so should we keep it here as well? Or create dedicated functions in the common logic?

libs/coin-modules/coin-tron/src/bridge/synchronization.ts
```diff
@@ -80,6 +81,7 @@ export const getAccountShape: GetAccountShape<TronAccount> = async (
   const balance = computeBalanceBridge(acc);
 
   const parentTxs = txs.filter(isParentTx);
+  // why don't i use listOperations here? Because two types TronOperation and Operation<TronToken> do not have the same structure
```
hedi-edelbloute, line 84: @sprohaszka-ledger we need to have logic method to output backend raw response and then adapt in bridge / api folder ?

libs/coin-modules/coin-tron/src/logic/getAccount.ts
hedi-edelbloute: We use fetchTronAccount in alpaca api, so you need to replace alpaca api call to this common getAccount method

hedi-edelbloute (changes requested):
Add tests please 🙏

libs/coin-modules/coin-tron/src/logic/getAccount.ts
hedi-edelbloute: You need to use getAccount in alpaca api

libs/coin-modules/coin-tron/src/bridge/getEstimateFees.ts
```diff
@@ -1,15 +1,12 @@
...
-  STANDARD_FEES_TRC_20,
-} from "../logic/constants";
-import { fetchTronAccount } from "../network";
+import { ACTIVATION_FEES, STANDARD_FEES_NATIVE, STANDARD_FEES_TRC_20 } from "../logic/constants";
 import type { AccountTronAPI } from "../network/types";
 import { Transaction } from "../types";
 import { extractBandwidthInfo, getEstimatedBlockSize } from "./utils";
+import { estimateFees, getAccount } from "../logic";
+import type { TransactionIntent } from "@ledgerhq/coin-framework/api/index";
+import { TronToken } from "../types";
```
francois-guerin-ledger, line 9:
```suggestion
import type { TronToken } from "../types";
```

francois-guerin-ledger (dismissed):
LGTM

libs/coin-modules/coin-tron/src/bridge/getEstimateFees.ts
sprohaszka-ledger: Moreover, use one import per file. You already have a `import { Transaction } from "../types";` on line 5

libs/coin-modules/coin-tron/src/bridge/synchronization.ts
sprohaszka-ledger: The best again is to abstract the backend response. Exposing the backend raw response is not a good idea as it may evolve and therefore have incidence on both `bridge` and `api`.
The `logic`folder will expose its own model to the `bridge` and `api` folder.

libs/coin-modules/coin-tron/src/logic/getAccount.ts
sprohaszka-ledger: @Canestin , I think Hedi's point is to replace this [call](https://github.com/LedgerHQ/ledger-live/blob/2ad754d1cb450a07390f88c9b18a82cc449e2d43/libs/coin-modules/coin-tron/src/logic/getBalance.ts#L19) with your new method

sprohaszka-ledger (commented):
A few comments

hedi-edelbloute (dismissed):

francois-guerin-ledger (approved):

PR #9717, opened 2025-03-28
fix: retry behavior set for post requests within solana calls

Wozacosta:
### ✅ Checklist

  - ...

### 📝 Description

Default request behavior made through ky skipped retrying on POST request, problem is all solana requests are made using POST requests. 

### ❓ Context

- **JIRA or GitHub link**: LIVE-17972

---

### DIFF
--- a/libs/coin-modules/coin-solana/src/api/chain/index.ts
+++ b/libs/coin-modules/coin-solana/src/api/chain/index.ts
@@ -129,7 +129,24 @@ const remapErrorsWithRetry = <P extends Promise<T>, T>(callback: () => P, times 
   };
 };
 
-const kyNoTimeout = ky.create({ timeout: false });
+/*
+NOTE: https://github.com/sindresorhus/ky?tab=readme-ov-file#retry
+defaults values are set here https://github.com/sindresorhus/ky/blob/b49cd03d8673ea522a29bae4ef6b4672cf23201b/source/utils/normalize.ts#L14
+retrying on 404 is debatable, but due to current solana node behavior, it's necessary
+*/
+const kyNoTimeout = ky.create({
+  timeout: false,
+  retry: {
+    limit: 3,
+    statusCodes: [404, 408, 413, 429, 500, 502, 503, 504],
+    methods: ["get", "post"],
+  },
+});
+
+const kyNoRetry = ky.create({
+  timeout: false,
+  retry: { limit: 0 },
+});
 
 export function getChainAPI(
   config: Config,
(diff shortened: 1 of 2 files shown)


libs/coin-modules/coin-solana/src/api/chain/index.ts
```diff
@@ -129,7 +129,24 @@ const remapErrorsWithRetry = <P extends Promise<T>, T>(callback: () => P, times
   };
 };
 
-const kyNoTimeout = ky.create({ timeout: false });
+/*
+NOTE: https://github.com/sindresorhus/ky?tab=readme-ov-file#retry
+defaults values are set here https://github.com/sindresorhus/ky/blob/b49cd03d8673ea522a29bae4ef6b4672cf23201b/source/utils/normalize.ts#L14
+retrying on 404 is debatable, but due to current solana node behavior, it's necessary
```
Justkant, line 135: IMO it's necessary if retrying here would give us a response 200 after we received a 404, but if we just get 404 non stop with the current solana node behavior you are talking about I think it's better to not retry

libs/coin-modules/coin-solana/src/api/chain/index.ts
```diff
@@ -129,7 +129,24 @@ const remapErrorsWithRetry = <P extends Promise<T>, T>(callback: () => P, times
...
+*/
+const kyNoTimeout = ky.create({
+  timeout: false,
+  retry: {
+    limit: 3,
+    statusCodes: [404, 408, 413, 429, 500, 502, 503, 504],
+    methods: ["get", "post"],
```
Justkant, line 142: We should keep the rest of the defaults as options request can also fail sometimes
```suggestion
    methods: ["get", "post", "put", "head", "delete", "options", "trace"],
```

libs/coin-modules/coin-solana/src/api/chain/index.ts
```diff
@@ -257,14 +276,15 @@ export function getChainAPI(
 
     sendRawTransaction: (buffer: Buffer, recentBlockhash?: BlockhashWithExpiryBlockHeight) => {
       return (async () => {
-        const conn = connection();
+        const connNoRetry = connectionNoRetry(); // Prevent retries on sendTransaction
 
         const commitment = "confirmed";
 
-        const signature = await conn.sendRawTransaction(buffer, {
+        const signature = await connNoRetry.sendRawTransaction(buffer, {
```
Justkant, line 283: If it's the only call we don't retry, I think it's pretty safe to call it again actually and we can just keep one connection

The best would be to do our own retry logic too as shown here: https://solana.com/developers/guides/advanced/retry#customizing-rebroadcast-logic

libs/coin-modules/coin-solana/src/api/chain/index.ts
Wozacosta: indeed, removed the 404 comment and status from the retryable status code.

libs/coin-modules/coin-solana/src/api/chain/index.ts
Wozacosta: done ✅

libs/coin-modules/coin-solana/src/api/chain/index.ts
Wozacosta: Removed the specific connNoRetry logic for sendRawTransaction, custom retry logic would be best in a separate PR later on wdyt ?

libs/coin-modules/coin-solana/src/api/chain/index.ts
Justkant: I can still see the comment, maybe you forgot to push a commit ?

libs/coin-modules/coin-solana/src/api/chain/index.ts
Justkant: Yes it should be done in another PR

Justkant:
I would green the PR like that but you said that you'd remove the 404 comment and status from the list above https://github.com/LedgerHQ/ledger-live/pull/9717#discussion_r2021244077
Do we want to keep it after all ?

Justkant (approved):

Justkant (approved):

PR #9722, opened 2025-03-28
fix(lld): unwanted rendering when clicking the 'close' button from the recover screen just after onboarding

ysitbon:
### ✅ Checklist

- [x] **Impact of the changes:**  I modified the routing that was happening inside the Dashboard component instead of being dispatched like other routes. Maybe it was done for a good reason, as a workaround for example, but I don't envision why atm so don't hesitate to enlighten me ;). Also, for this same reason I encourage the QA to ensure the routing experience doesn't provide any regression 🙏. 

### 📝 Description

After the onboarding, when clicking the `close` button, it was displaying the Portfolio before the PostOnboarding. It was caused by a re-routing called during the `Dashboard` rendering. So, it was logically processing, first the `Dashboard` rendering and then the one from the component provided by the re-routing. 

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-17336

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/hooks/useAutoRedirectToPostOnboarding/index.ts
+++ b/apps/ledger-live-desktop/src/renderer/hooks/useAutoRedirectToPostOnboarding/index.ts
@@ -9,6 +9,9 @@ import { useHistory } from "react-router";
 /**
  * Navigates to the post onboarding or to the Ledger Recover upsell if needed
  * */
+// TODO: Rename and refactor this function and use algebraic data types like Result
+// to avoid the need of a fallback redirection and to make the function more readable
+// and maintainable. (see neverthrow library for example)
 export function useRedirectToPostOnboardingCallback() {
   const history = useHistory();
   const lastOnboardedDevice = useSelector(lastOnboardedDeviceSelector);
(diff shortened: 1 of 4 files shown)


apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking/useDeepLinkHandler.ts
```diff
@@ -364,11 +366,21 @@ export function useDeepLinkHandler() {
         }
         case "portfolio":
```
themooneer, line 367: no reason for this case i guess while default will be trigger

ysitbon: Absolutely right! Let's remove that too :D

apps/ledger-live-desktop/src/renderer/hooks/useAutoRedirectToPostOnboarding/index.ts
```diff
@@ -29,10 +29,14 @@ export function useRedirectToPostOnboardingCallback() {
       if (shouldRedirectToRecoverUpsell) {
         openRecoverUpsell({ fallbackRedirection });
         onDone();
-      } else if (shouldRedirectToPostOnboarding && lastOnboardedDevice) {
+        return true;
```
themooneer, line 32: if this hook will return a boolean , then the naming should change , or you can use another approach for that.

themooneer (approved):

PR #9742, opened 2025-03-31
:pencil2: (lld) add fallback to sync error message

LucasWerey:
### ✅ Checklist

  - account row sync indicator

### 📝 Description

I've not been able to reproduce the issue since it's more like an infra error. But from my understanding we can add this fallback this way we won't display the message if there is no corresponding translation but "please, try again" instead. What's better than [Object Object] or "no network" I guess

image

### ❓ Context

- **JIRA or GitHub link**: [LIVE-16756](https://ledgerhq.atlassian.net/browse/LIVE-16756)

---

[LIVE-16756]: https://ledgerhq.atlassian.net/browse/LIVE-16756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/accounts/AccountSyncStatusIndicator.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/accounts/AccountSyncStatusIndicator.tsx
@@ -85,7 +85,10 @@ class StatusError extends PureComponent<{
               maxWidth: 250,
             }}
           >
-            <TranslatedError error={error} />
+            <TranslatedError
+              fallback={<Trans i18nKey="errors.AccountNeedResync.title" />}
+              error={error}
+            />
           </Box>
         }
       >
--- a/.changeset/chilly-hairs-cough.md
+++ b/.changeset/chilly-hairs-cough.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": minor
+---
+
+Sync error add a fallback message to avoid non ux error message


apps/ledger-live-desktop/src/renderer/screens/accounts/AccountSyncStatusIndicator.tsx
```diff
@@ -85,7 +85,10 @@ class StatusError extends PureComponent<{
               maxWidth: 250,
             }}
           >
-            <TranslatedError error={error} />
+            <TranslatedError
+              fallback={<Trans i18nKey="errors.AccountNeedResync.title" />}
```
thesan, line 89: It feels like a waist to loose the error message when it might not even be solving the issue.

I'm don't know if it's the issue either but at [TranslatedError.tsx line 95](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/components/TranslatedError/TranslatedError.tsx#L95) there is: `const errorTranslation = t(errorKey, args);` where `args` is `{  ...error, message: error?.message, returnObjects: true }`.
What do you think of changing the line 95 by
```ts
const errorTranslation = t(errorKey, { ...args, returnObjects: false });
```
?

LucasWerey: I agree that error message can be useful, but as a user, it's not really helpful to have an unreadable error message. Besides, we only display the error message if there's no corresponding translation, so I guess it's ok to have a fallback text with "Please try again" will be more user friendly wdyt ?

cgrellard-ledger (approved):

PR #9777, opened 2025-04-02
fix(BACK-8611): fix tezos fee estimation silently catching errors

jnicoulaud-ledger:
### ✅ Checklist

  - ...

### 📝 Description

ensure error is finally thrown, or `0` is returned, silently eating the error

### ❓ Context

https://ledgerhq.atlassian.net/browse/BACK-8611

---

### DIFF
--- a/.changeset/six-avocados-breathe.md
+++ b/.changeset/six-avocados-breathe.md
@@ -1,5 +1,5 @@
 ---
-"@ledgerhq/coin-tezos": patch
+"@ledgerhq/coin-tezos": minor
 ---
 
 fix tezos fee estimation silently catching errors
--- a/.changeset/six-avocados-breathe.md
+++ b/.changeset/six-avocados-breathe.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-tezos": patch
+---
+
+fix tezos fee estimation silently catching errors
--- a/libs/coin-modules/coin-tezos/src/api/index.test.ts
+++ b/libs/coin-modules/coin-tezos/src/api/index.test.ts
@@ -114,6 +114,12 @@ describe("Testing craftTransaction function", () => {
 describe("Testing estimateFees function", () => {
   beforeEach(() => jest.clearAllMocks());
 
+  it("should return estimation from logic module", async () => {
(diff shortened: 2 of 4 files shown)


Justkant (commented):
Hey, I think this was the expected behaviour as `estimation.taquitoError` is set and we then return `estimation` in the only case we don't throw the error

If it's used in the prepareTransaction we don't want it to throw at all usually and instead we want to return the errors to the UI for the different fields/components impacted

Also you can see the `taquitoError` used here: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-tezos/src/bridge/getFeesForTransaction.ts#L47
I also see it in the transaction itself: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-tezos/src/types/bridge.ts#L42
And in the getTransactionStatus that will set errors for the UI: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-tezos/src/bridge/getTransactionStatus.ts#L68-L91

Justkant (commented):
If it's only about the api exposed for alpaca maybe we can just check for the `taquitoError` but you'll still be missing the rest of the error with the current code: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-tezos/src/api/index.ts#L61-L79

jnicoulaud-ledger:
@Justkant OK I see, I moved it to `api.estimateFees` only

.changeset/six-avocados-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/coin-tezos": patch
```
Justkant, line 2: Small detail sorry for that but we now only use `minor` for standard bump and reserve the patch for hotfixes 
```suggestion
"@ledgerhq/coin-tezos": minor
```

Justkant (changes requested):

Justkant (approved):

hedi-edelbloute (approved):

Moustafa-Koterba (approved):

PR #9801, opened 2025-04-04
feat(LIVE-17984): add support for `memo` and `expiration` craft parameters for tron coin module

jnicoulaud-ledger:
### ✅ Checklist

  - ...

### 📝 Description

add support for `memo` and `expiration` craft parameters for tron coin module

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-17984

---

### DIFF
--- a/libs/coin-modules/coin-tron/src/logic/craftTransaction.integ.test.ts
+++ b/libs/coin-modules/coin-tron/src/logic/craftTransaction.integ.test.ts
@@ -188,4 +188,111 @@ describe("Testing craftTransaction function", () => {
       }),
     );
   });
+
+  it("should create a valid transaction when setting a memo", async () => {
+    const amount = BigInt(20);
+    const sender = "TRqkRnAj6ceJFYAn2p1eE7aWrgBBwtdhS9";
+    const recipient = "TPswDDCAWhJAZGdHPidFg5nEf8TkNToDX1";
+
+    // WHEN
+    const result = await craftTransaction({
+      type: "send",
+      sender,
+      recipient,
+      amount,
+      memo: "this is a test",
+    });
+
+    const decodeResult = await decodeTransaction(result);
+
+    expect(decodeResult).toEqual(
+      expect.objectContaining({
+        raw_data: expect.objectContaining({
+          data: Buffer.from("this is a test"),
+        }),
+      }),
+    );
+  });
+
+  it("throw expected error when setting a memo on a smart contract transaction", async () => {
+    const amount = BigInt(20);
+    const sender = "TRqkRnAj6ceJFYAn2p1eE7aWrgBBwtdhS9";
+    const recipient = "TPswDDCAWhJAZGdHPidFg5nEf8TkNToDX1";
+
+    await expect(
+      craftTransaction({
+        type: "send",
+        asset: {
+          standard: "trc20",
+          contractAddress: "TLa2f6VPqDgRE67v1736s7bJ8Ray5wYjU7",
+        },
+        sender,
+        recipient,
+        amount,
+        memo: "test",
+      }),
+    ).rejects.toThrow("Memo cannot be used with smart contract transactions");
+  });
+
+  it("should set an expiration of 10 minutes by default", async () => {
+    const amount = BigInt(20);
+    const sender = "TRqkRnAj6ceJFYAn2p1eE7aWrgBBwtdhS9";
+    const recipient = "TPswDDCAWhJAZGdHPidFg5nEf8TkNToDX1";
+
+    const before = Date.now();
+
+    // WHEN
+    const result = await craftTransaction({
+      type: "send",
+      sender,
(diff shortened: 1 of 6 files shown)


libs/coin-modules/coin-tron/src/logic/craftTransaction.ts
```diff
@@ -5,11 +5,20 @@ import { decode58Check } from "../network/format";
...
+type TransactionIntentExtra = {
+  /** Memo value. */
+  memo?: string;
+
+  /** Expiration in seconds after crafting time. */
+  expiration?: number;
+};
+
 export async function craftTransaction(
   transactionIntent: TransactionIntent<TronToken>,
```
francois-guerin-ledger, line 17:
```suggestion
  transactionIntent: TransactionIntent<TronToken, TransactionIntentExtra>,
```

libs/coin-modules/coin-tron/src/logic/craftTransaction.ts
```diff
@@ -5,11 +5,20 @@ import { decode58Check } from "../network/format";
...
   customFees?: bigint,
 ): Promise<string> {
   const { asset, recipient, sender, amount } = transactionIntent;
+  const { memo, expiration } = transactionIntent as TransactionIntentExtra;
```
francois-guerin-ledger, line 21: After updating the generic, you should be able to destructure the object without casting

francois-guerin-ledger (changes requested):

libs/coin-modules/coin-tron/src/logic/craftTransaction.integ.test.ts
```diff
@@ -188,4 +188,111 @@ describe("Testing craftTransaction function", () => {
...
+        },
+        sender,
+        recipient,
+        amount,
+        memo: "test",
+      }),
+    ).rejects.toThrow("Memo cannot be used with smart contract transactions");
+  });
+
+  it("should set an expiration of 10 minutes by default", async () => {
```
Moustafa-Koterba, line 237:
```suggestion
  it("should set an expiration of 10 minutes by default when setting is not provided", async () => {
```
wdyt

libs/coin-modules/coin-tron/src/network/index.ts
```diff
@@ -275,19 +279,23 @@ export const createTronTransaction = async (
     );
   }
 };
-async function extendTronTxExpirationTimeBy10mn(
+
+/** Default expiration, in seconds after crafting time. */
```
Moustafa-Koterba, line 283:
```suggestion
/** Default expiration of 10 minutes (in seconds) after crafting time. */
```

libs/coin-modules/coin-tron/src/network/index.ts
```diff
@@ -275,19 +279,23 @@ export const createTronTransaction = async (
...
   const HttpProvider = TronWeb.providers.HttpProvider;
   const fullNode = new HttpProvider(getBaseApiUrl());
   const solidityNode = new HttpProvider(getBaseApiUrl());
   const eventServer = new HttpProvider(getBaseApiUrl());
   const tronWeb = new TronWeb(fullNode, solidityNode, eventServer);
-  //FIXME: test it and rewrite it
   return tronWeb.transactionBuilder.extendExpiration(
     preparedTransaction,
-    VAULT_EXPIRATION_TIME,
+    value,
```
Moustafa-Koterba, line 298:
```suggestion
    expiration ?? DEFAULT_EXPIRATION,
```

libs/coin-modules/coin-tron/src/logic/craftTransaction.ts
jnicoulaud-ledger: OK, done. I would expect `createApi` to break to be honest, because now it uses `craftTransaction` with unrelated types, but it seems the type checker does not catch it... Or maybe valid because `Record` is a supertype of `TransactionIntentExtra` ? 

Proper fix would be to add the generic type up to `Api`, but I don't want to do in this PR as it will impact framework + all modules.

francois-guerin-ledger: > Or maybe valid because `Record` is a supertype of `TransactionIntentExtra` ?

Yes, `Record<string, unknown>` is a supertype.

> Proper fix would be to add the generic type up to `Api`, but I don't want to do in this PR as it will impact framework + all modules.

I think keeping this `Extra` is fine

francois-guerin-ledger (approved):

libs/coin-modules/coin-tron/src/logic/craftTransaction.ts
jnicoulaud-ledger: Yes, but it means we can't use correct typing in alpaca (or anyone using `Api`). Not a big deal, but probably we want to plan this

PR #9837, opened 2025-04-08
feat(coin-ton): expose ton payload for all possible actions [LIVE-17777]

Justkant:
### ✅ Checklist

  - none, as it's only a new optional field for wallet-api flows

### 📝 Description

We couldn't access some of the coin-ton functionality as it was not exposed through the transaction type

I'm patching `@ton/core` to make the classes work with jest expect

I'm update the wallet-api right now and will update the version in this PR before merging

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17777] [LIVE-13067]

---

[LIVE-17777]: https://ledgerhq.atlassian.net/browse/LIVE-17777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-13067]: https://ledgerhq.atlassian.net/browse/LIVE-13067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-modules/coin-ton/src/transaction.ts
```diff
@@ -28,6 +29,109 @@ SEND ${
...
+        address: Address.parse(payload.address),
+      };
+    case "single-nominator-withdraw":
+      return {
+        type: payload.type,
+        queryId: payload.queryId ? BigInt(payload.queryId) : null,
+        amount: BigInt(payload.amount),
+      };
+    case "single-nominator-change-validator":
+      return {
```
qperrot, line 83: case "single-nominator-change-validator" is the same as case "add-whitelist":

libs/coin-modules/coin-ton/src/transaction.ts
```diff
@@ -36,9 +140,127 @@ export const fromTransactionRaw = (tr: TransactionRaw): Transaction => {
...
+        address: payload.address.toRawString(),
+      };
+    case "single-nominator-withdraw":
+      return {
+        type: payload.type,
+        queryId: payload.queryId ? payload.queryId.toString() : null,
+        amount: payload.amount.toString(),
+      };
+    case "single-nominator-change-validator":
+      return {
```
qperrot, line 210: Same as "add-whitelist" here too

libs/coin-modules/coin-ton/src/transaction.ts
Justkant: Nice catch, I'll update

libs/coin-modules/coin-ton/src/transaction.ts
Justkant: Nice catch I'm updating

libs/coin-modules/coin-ton/src/transaction.ts
Moustafa-Koterba, line 1: can you add some unit tests ?

Moustafa-Koterba, line 1: (sorry for the wall of text 😂 😂 😂 )

This is `optional`
wdyt of splitting the code, instead of having a big `switch`, maybe have a map of `payload.type` by mapper function. Thus, if we need to add new type in the future, we will just add a new mapper function to this map and it will be picked automatically

Like
```ts
const mapperByType = new Map<string, (payload: TonPayloadFormatRaw) => TonPayloadFormat>();
mapperByType.set("unsafe", toUnsafe TonPayloadFormat)
// etc.

const fromTransactionPayloadRaw = (payload: TonPayloadFormatRaw): TonPayloadFormat => {
  const mapperFunction = mapperByType.get(payload.type);
  if (mapperFunction) {
    return mapperFunction(payload);
  }

  throw new Error();
}
```
It is part of the SOLID principle in OOP, the `O` which is `Open-closed Principle`:
> software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification

Moustafa-Koterba (approved):

libs/coin-modules/coin-ton/src/transaction.ts
Justkant: Isn't this already covered by the types ?
Seems fairly redundant to write tests for this but I'll take a look

libs/coin-modules/coin-ton/src/transaction.ts
```diff
@@ -28,6 +29,104 @@ SEND ${
...
+        forwardPayload: payload.forwardPayload ? Cell.fromBase64(payload.forwardPayload) : null,
+      };
+    case "jetton-burn":
+      return {
+        type: payload.type,
+        queryId: payload.queryId ? BigInt(payload.queryId) : null,
+        amount: BigInt(payload.amount),
+        responseDestination: Address.parse(payload.responseDestination),
+        // Handle customPayload could be a buffer, we should try/catch the fromBase64
+        customPayload: payload.customPayload ? Cell.fromBase64(payload.customPayload) : null,
```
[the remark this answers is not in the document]
Justkant: I'm checking again if we need to handle it here or not, I'll update when I have more info

libs/coin-modules/coin-ton/src/transaction.ts
Justkant: Also the `Open-Closed Principle` was really design to solve a problem that we don't have here mainly extending an already built library

> At the time Meyer was writing, adding fields or functions to a library inevitably required changes to any programs depending on that library.[[citation needed](https://en.wikipedia.org/wiki/Wikipedia:Citation_needed)] Meyer's proposed solution to this problem relied on the notion of object-oriented [inheritance](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)) (specifically [implementation inheritance](https://en.wikipedia.org/wiki/Implementation_inheritance)):[[1]](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle#cite_note-:0-1): 229 
> 
> A class is closed, since it may be compiled, stored in a library, [baselined](https://en.wikipedia.org/wiki/Baseline_(configuration_management)), and used by client classes. But it is also open, since any new class may use it as parent, adding new features. When a descendant class is defined, there is no need to change the original or to disturb its clients.

I'm not sure this change would solve an issue in our case

Justkant: You can check this PR for the exhaustive check I'm adding here too, also the return type was already forcing us to handle all cases, it just adds a default case in case it's used with something the type isn't expecting
https://github.com/LedgerHQ/wallet-api/pull/449

Justkant: Also you can see an example of the issue of typing the params when using a record or a map here: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/transaction/index.ts#L13-L22

And if you instead type as the general type you need to narrow in all functions adding more code

Moustafa-Koterba (approved):

Moustafa-Koterba (approved):

mohammed-ledger (approved):

PR #9859, opened 2025-04-09
chore(llm): add base MMKV implementation

ysitbon:
### ✅ Checklist

- [x] **Impact of the changes:** None

### 📝 Description

Implement MMKVSAsyncWrapper and base application Storage. Adds unit test for all wrappers Async / MMKV and for base application Storage.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-18087

---


apps/ledger-live-mobile/src/newArch/storage/index.ts
```diff
@@ -1,34 +1,83 @@
...
+  const state: StorageState = {
+    storageType: "AsyncStorage",
+  };
+
+  initializer(state);
   return {
     keys() {
-      return asyncStorageWrapper.keys();
+      try {
+        return state.storageType === "MMKV"
```
mcayuelas-ledger, line 19:
```suggestion
        return state.storageType === STORAGE_TYPE.MMKV
```
This works ? More efficient for safety type

apps/ledger-live-mobile/src/newArch/storage/mmkvStorageWrapper/index.ts
```diff
@@ -0,0 +1,129 @@
+import { MMKV, Mode } from "react-native-mmkv";
+import merge from "lodash/merge";
+
+const mmkv = new MMKV({
+  id: "ledger-live",
```
mcayuelas-ledger, line 5: we may want to extract this in config file maybe?

ysitbon: Totally agree, I had the same while implementing. We should first answer if it's necessary? And if yes, where to store this configuration, when it is set (runtime, env var with ci, etc)?

ysitbon: Following the documentation this id has nothing to do with security. It is only used if you need to have multiple MMKV instances that read and write in different places. So it can be configured in code directly without any issue.

I think that creating a `constant.ts` file with `CONFIG_PARAMS` for the MMKV wrapper should be enough.

apps/ledger-live-mobile/src/newArch/storage/index.ts
```diff
@@ -1,34 +1,83 @@
...
+      } catch (e) {
+        console.error("Error getting keys from storage", e);
+        return Promise.reject(e);
+      }
     },
 
     get(key) {
-      return asyncStorageWrapper.get(key);
+      try {
+        return state.storageType === "MMKV"
```
themooneer, line 30: STORAGE_TYPE.MMKV

apps/ledger-live-mobile/src/newArch/storage/index.ts
```diff
@@ -1,34 +1,83 @@
-import type { Storage } from "./types";
+import type { Storage, StorageInitializer, StorageState } from "./types";
 import asyncStorageWrapper from "./asyncStorageWrapper";
+import mmkvStorageWrapper from "./mmkvStorageWrapper";
+import { STORAGE_TYPE } from "./constants";
 
 /** Singleton reference to the global application storage object. */
 export default createStorage();
```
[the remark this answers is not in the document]
ysitbon: `AsyncStorage` is a singleton so for this part we don't have any issue. On the contrary for MMKV it requires an initialization, but since it is supposed to become our defacto standard that is always used the current impl is intended. It is always being instantiated and always use the same singleton instance created in `mmkvStorageWrapper`. 

That's the main reasons why I didn't implemented an initialization strategy. In any case it can be easily added now or later depending on recommandation, requirement, necessity, etc.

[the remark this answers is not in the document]
themooneer: so clear

themooneer (approved):
LGTM

PR #9867, opened 2025-04-09
[LIVE-18179] `Something went wrong` error briefly appears when reopening My Ledger while device app is open

jiyuzhuang:
### ✅ Checklist

  - ...

### 📝 Description

Remove unexpected page when returning to manager page.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/.changeset/twelve-hats-accept.md
+++ b/.changeset/twelve-hats-accept.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": minor
+"@ledgerhq/live-common": minor
+---
+
+Fix unexpected warning when return to manager page
--- a/apps/ledger-live-desktop/src/renderer/screens/manager/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/manager/index.tsx
@@ -21,7 +21,6 @@ const Manager = () => {
   const { setDrawer } = useContext(context);
   const [result, setResult] = useState<Result | null>(null);
   const [hasReset, setHasReset] = useState(false);
-
   const onReset = useCallback(
     (apps?: string[] | null) => {
       setRestoreApps(apps ?? []);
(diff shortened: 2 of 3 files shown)


libs/ledger-live-common/src/hw/actions/manager.ts
```diff
@@ -249,6 +249,7 @@ export const createAction = (
 
     return {
       ...state,
+      unresponsive: false,
```
jdabbech-ledger, line 252: it looks like an override of `unresponsive` field of reducer state, not a good solution if the device is really unresponsive imo. 
It could have side effect on LLM as well.

jdabbech-ledger (changes requested):
It could result with an infinite loader on LLM to force `unresponsive: false` in connectManager

fAnselmi-Ledger (changes requested):
I think this can be solved by delaying the `UnresponsiveDeviceError` `renderError()` in `DeviceAction/index.tsx` as it's a timing issue. 

Device briefly disconnects when app is switched and `DeviceAction` immediately renders the `UnresponsiveDeviceError`. (you can also narrow down the `DeviceAction` by checking on `location` prop, dashboard should match with `HOOKS_TRACKING_LOCATIONS.managerDashboard `, same for mobile)

fAnselmi-Ledger (commented):
>Delaying the render without check the location is not Ok, this component is widely used, I cannot make sure the delaying is expected in each use case. Checking the location with the tracking property is not a good solution either, the component behaviour should not depend on any tracking issues. I will try to find out another solution.

Yeah those are good points, I wasn't specifically suggesting to delay every time, you can also do it via `DeviceAction` prop, I was just suggesting you that the issue lies on `DeviceAction` being too responsive in that instance:)

jdabbech-ledger (approved):

PR #9891, opened 2025-04-10
ci: Disable smartling content update runs

lewisd5:
Smarling gives us content as PRs are created and also on-demand for release.

When done against a PR they are prefixed with "smartling-content-updated"
https://github.com/LedgerHQ/ledger-live/pull/9816

When done against an on demand they are prefixed with "smartling-translations"
https://github.com/LedgerHQ/ledger-live/pull/9728

This PR should preserve the on-demand PR runs while disabling the runs for content updated

### DIFF
--- a/.github/workflows/test-integration-pr.yml
+++ b/.github/workflows/test-integration-pr.yml
@@ -3,8 +3,8 @@ name: "[Test] Integration"
 on:
   pull_request:
     branches-ignore:
-      - 'smartling-content-updated*'
-      - 'smartling-translation-completed*'
+      - 'smartling-content-updated**'
+      - 'smartling-translation-completed**'
 
 concurrency:
   group: ci-integration-${{ github.ref }}


.github/workflows/build-and-test-pr.yml
```diff
@@ -3,6 +3,9 @@ on:
   merge_group:
     types: [checks_requested]
   pull_request:
+    branches-ignore:
+        - 'smartling-content-updated*'
+        - 'smartling-translation-completed*'
```
angusbayley, line 8: should be `**` not `*`

From the [documentation](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet):

> `*`: Matches zero or more characters, but does not match the / character

The branch for https://github.com/LedgerHQ/ledger-live/pull/9816 is `smartling-translation-completed-aimeothvv6qo-from-feat/aptos-token-integration`

angusbayley (changes requested):
One change needed (then it's a 👍)

angusbayley (approved):

PR #9892, opened 2025-04-10
feat: Modular Asset Drawer redirection into the Add Account flow

mcayuelas-ledger:
### ✅ Checklist

  - ...

### 📝 Description

This PR introduces logic to handle redirection based on the return value of **isModularDrawerVisible**.

Some existing components related to drawers have been extracted into a new component to facilitate testing and to begin the separation between the old and new flows.

The following files:

- `apps/ledger-live-desktop/src/newArch/features/ModularDrawer/utils/selectCurrency.ts`
- `apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectCurrencyDrawer.tsx`

are expected to evolve in the future to accommodate new designs and integrate with the upcoming API.

https://github.com/user-attachments/assets/fce158e7-21f1-40ee-9223-359422037a18

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18128] 

---

[LIVE-18128]: https://ledgerhq.atlassian.net/browse/LIVE-18128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cgrellard-ledger (approved):

apps/ledger-live-desktop/src/renderer/screens/accounts/AccountRowItem/Placeholder.tsx
```diff
@@ -26,14 +26,11 @@ const AddAccountButton = styled(Box)`
   }
 `;
 const Placeholder = () => {
-  const dispatch = useDispatch();
   const { t } = useTranslation();
-  const openAddAccounts = useCallback(() => {
-    dispatch(openModal("MODAL_ADD_ACCOUNTS", undefined));
-  }, [dispatch]);
+  const { openAssetFlow } = useOpenAssetFlow(ModularLocation.ADD_ACCOUNT);
```
thesan, line 30: Suggestion:
```suggestion
  const openAddAccounts = useCallback((asset: CryptoOrTokenCurrency) => {
    dispatch(openModal("MODAL_ADD_ACCOUNTS", { currency }));
  }, [dispatch]);
  const { openAssetFlow } = useOpenAssetFlow(ModularLocation.ADD_ACCOUNT, openAddAccounts);
```

apps/ledger-live-desktop/src/renderer/screens/accounts/AccountsHeader.tsx
```diff
@@ -1,20 +1,17 @@
...
+import { useOpenAssetFlow } from "LLD/features/ModularDrawer/hooks/useOpenAssetFlow";
+import { ModularLocation } from "LLD/features/ModularDrawer/enums";
 
 const AccountsHeader = () => {
   const { t } = useTranslation();
-  const dispatch = useDispatch();
-  const openAddAccounts = useCallback(() => {
-    dispatch(openModal("MODAL_ADD_ACCOUNTS", undefined));
-  }, [dispatch]);
+  const { openAssetFlow } = useOpenAssetFlow(ModularLocation.ADD_ACCOUNT);
```
thesan, line 14: Same thing here

apps/ledger-live-desktop/src/renderer/reducers/modals.ts
```diff
@@ -3,6 +3,8 @@ import { State } from "~/renderer/reducers";
 import { ModalData } from "~/renderer/modals/types";
 import { Handlers } from "./types";
 
+export const INITIAL_STATE: ModalsState = {};
```
thesan, line 6: Where is this used ?

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectCurrencyDrawer.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }, [assetIds, searchValue, sortedCurrencies]);
+
+  const handleCurrencySelected = useCallback(
+    (currency: CryptoOrTokenCurrency) => {
+      onAssetSelected(currency);
+      dispatch(
+        openModal("MODAL_ADD_ACCOUNTS", {
+          currency,
+        }),
+      );
```
thesan, line 85: Oh ok this is where the modal is called. But IMO this component should be more general. WDYT ?

mcayuelas-ledger: I just extracted what was on Jira Ticket, This will be reworked as I explained

but to be more general, I don't what to do honestly

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/utils/selectCurrency.ts
```diff
@@ -0,0 +1,32 @@
+import { Account, AccountLike } from "@ledgerhq/types-live";
+import { listAndFilterCurrencies } from "@ledgerhq/live-common/platform/helpers";
+
+import SelectCurrencyDrawer from "../components/SelectCurrencyDrawer";
+import { setDrawer } from "~/renderer/drawers/Provider";
+type SelectAccountAndCurrencyResult = {
+  account: AccountLike;
+  parentAccount?: Account;
+};
+export function selectCurrency(
```
thesan, line 10: Why should this function be extracted from `useOpenAssetFlow` how will it be reused in a different context ? also why does it return a Promise ?

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectCurrencyDrawer.tsx
thesan: Yes I think extracting the part of `SelectAccountAndCurrencyDrawer` needed here is exactly what needed to be done in this PR. And creating a hook to open this drawer is the best solution IMO.

That said in the scope I suggested to extract the logic into a temporary component called `SelectCurrencyDrawer` and create a `SelectAssetFlow` component in new Arch which would call this first temporary component. I think your solution of creating just one drawer component which will be updated later is much better, but in this case the component should be called "SelectAssetFlow" instead of "SelectCurrencyDrawer" to be consistent with the rest of the scope.

But here the issue I have with this temporary version of the drawer is that it is in charge of opening the add account modal. This PR is supposed to wire the drawer to the add account flow so even though the implementation of drawer is not in this PR scope, I think the hook that opens it and the changes in components that call this hook should be implemented fully in this PR.

Earlier today I suggested that the hook opening the drawer should be reusable in other contexts of the add account. I still think that approach is fine but actually TBH I was out of context and I had forgotten that initially the scope was that only the add account flow would open the `SelectCurrencyDrawer`. In this case your initial approach of opening the `SelectCurrencyDrawer` from a hook dedicated to the add account flow (`useOpenAddAccounts`) is would be fine too. I'm sorry for the confusion.

So to summarise I see 2 approachs:
1. `useOpenAddAccounts`: opens the `SelectCurrencyDrawer` and pass it a `onAssetSelected` callback which opens the `MODAL_ADD_ACCOUNTS`.
2. `useOpenAssetFlow`: is called with a callback which opens the `MODAL_ADD_ACCOUNTS` and it opens the `SelectCurrencyDrawer` with this callback passed as the `onAssetSelected` prop.
 
The approach (1) is simpler and creates less code, it's less reusable but we might not reuse this hook anyway. The approach requires more code (i.e the suggestions I added in components with add account buttons). That said there is a risk of potentially needing to reuse `useOpenAssetFlow` for the send flow I really need to clarify this.

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/hooks/useOpenAssetFlow.tsx
```diff
@@ -0,0 +1,57 @@
+import { useModularDrawerVisibility } from "./useModularDrawerVisibility";
+import { openModal } from "~/renderer/actions/modals";
+import { useDispatch } from "react-redux";
+import { useCallback } from "react";
+import { ModularDrawerLocation } from "../enums";
+import { listAndFilterCurrencies } from "@ledgerhq/live-common/platform/helpers";
+import SelectCurrencyDrawer from "../components/SelectAssetFlow";
```
thesan, line 7:
```suggestion
import SelectAssetFlow from "../components/SelectAssetFlow";
```

PR #9931, opened 2025-04-15
📈 (dmk): Add analytics for opt-in for transaction checks

aussedatlo:
### ✅ Checklist

  - ...

### 📝 Description

Add analytics on LLD when the opt-in transaction checks screen appear on the device

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17708]

---

[LIVE-17708]: https://ledgerhq.atlassian.net/browse/LIVE-17708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/analytics/hooks/useTrackTransactionChecksFlow.ts
+++ b/apps/ledger-live-desktop/src/renderer/analytics/hooks/useTrackTransactionChecksFlow.ts
@@ -1,0 +1,47 @@
+import { Device } from "@ledgerhq/types-devices";
+import { track } from "../segment";
+import { CONNECTION_TYPES } from "./variables";
+import { useRef } from "react";
+import { DeviceInfo } from "@ledgerhq/types-live";
+import { AppAndVersion } from "@ledgerhq/live-common/hw/connectApp";
+
+export type UseTrackTransactionChecksFlow = {
+  location: string | undefined;
+  device: Device;
+  deviceInfo: DeviceInfo | undefined | null;
+  appAndVersion: AppAndVersion | undefined | null;
+  transactionChecksOptInTriggered: boolean | undefined | null;
+  isTrackingEnabled: boolean;
+};
+
+export const useTrackTransactionChecksFlow = ({
+  location,
+  device,
+  deviceInfo,
+  appAndVersion,
+  transactionChecksOptInTriggered,
+  isTrackingEnabled,
+}: UseTrackTransactionChecksFlow) => {
+  const triggered = useRef(false);
+
+  if (transactionChecksOptInTriggered && !triggered.current) {
+    triggered.current = true;
+
+    const defaultPayload = {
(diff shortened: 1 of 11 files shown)


cgrellard-ledger (dismissed):

jdabbech-ledger (dismissed):

apps/ledger-live-desktop/src/renderer/analytics/hooks/useTrackTransactionChecksFlow.ts
```diff
@@ -0,0 +1,47 @@
...
+      appFlags: appAndVersion?.flags.toString(),
+      deviceInfoHardwareVersion: deviceInfo?.hardwareVersion,
+      deviceInfoMcuVersion: deviceInfo?.mcuVersion,
+      deviceInfoBootloaderVersion: deviceInfo?.bootloaderVersion,
+      deviceInfoProviderName: deviceInfo?.providerName,
+      deviceInfoLanguageId: deviceInfo?.languageId,
+    };
+
+    track("Transaction Check Opt-in Triggered", defaultPayload, isTrackingEnabled);
+  }
```
Justkant, line 46: Why do we prefer using a ref instead of a useEffect here ?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -318,6 +321,15 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
     isTrackingEnabled: useSelector(trackingEnabledSelector),
   });
 
+  useTrackTransactionChecksFlow({
+    location,
+    device,
+    deviceInfo,
+    appAndVersion,
+    transactionChecksOptInTriggered,
+    isTrackingEnabled: useSelector(trackingEnabledSelector),
```
Justkant, line 330: Could we move the `useSelector` in a const and only call the useSelector once please ?

apps/ledger-live-desktop/src/renderer/analytics/hooks/useTrackTransactionChecksFlow.ts
aussedatlo: this component does not need to re render as it's only tracking, also we need to ensure that the event is fired only once so `useRef` is working great in this situation

Justkant: Yes but it's just that in react you usually don't do side effects in a render phase and should use a `useEffect` with no dependencies for an side effect that needs to run only once

Justkant (approved):

ofreyssinet-ledger (approved):

cgrellard-ledger (approved):

PR #9949, opened 2025-04-15
chore(LLD): patch WebHidSendReportError with LDMK enabled

fAnselmi-Ledger:
### ✅ Checklist

  - ...

### 📝 Description

Patched WebHidSendReportError error when LDMK transport is enabled

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18413](https://ledgerhq.atlassian.net/browse/LIVE-18413)

---

[LIVE-18413]: https://ledgerhq.atlassian.net/browse/LIVE-18413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
@@ -15,6 +15,7 @@ import {
   LockedDeviceError,
   UpdateYourApp,
   WrongDeviceForAccount,
+  DisconnectedDevice,
 } from "@ledgerhq/errors";
 import {
   DeviceNotOnboarded,
@@ -72,6 +73,7 @@ import { isSyncOnboardingSupported } from "@ledgerhq/live-common/device/use-case
 import NoSuchAppOnProviderErrorComponent from "./NoSuchAppOnProviderErrorComponent";
 import Image from "~/renderer/components/Image";
 import Nano from "~/renderer/images/nanoS.v4.svg";
+import { isWebHidSendReportError } from "@ledgerhq/live-dmk-desktop";
 
 export const AnimationWrapper = styled.div`
   width: 600px;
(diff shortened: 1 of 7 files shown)


mbertin-ledger (approved):
Ok for LLD but it could had been forgotten for LLM.

libs/live-dmk-desktop/src/errors.ts
valpinkman, line 1: [COULD] Add tests for this function

valpinkman (dismissed):

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -825,8 +827,9 @@ export const renderError = ({
         learnMoreTextKey={learnMoreTextKey}
       />
     );
+  } else if (isWebHidSendReportError(tmpError)) {
+    tmpError = new DisconnectedDevice();
```
ofreyssinet-ledger, line 831: [ASK] isn't there something in LL to remap errors rather than doing it at rendering time?

ofreyssinet-ledger: NB: this solution works so it is kind of ok (except that we create a new object at rendering time and it's used directly as a component prop) but we could find something more elegant to remap all such common transport errors in one place

fAnselmi-Ledger: Not the prettiest indeed, we will need to return more detailed error from DMK first (I'd say before v1), then we can handle it better in LL!

ofreyssinet-ledger (approved):

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
ofreyssinet-ledger: yep it's good for now, we have to define a strategy for error remapping in LL and indeed already remap this kind of error inside the DMK

ypolishchuk-ledger (approved):

cgrellard-ledger (approved):

PR #10003, opened 2025-04-18
feat(react:ui): 🔎 implement the new search component UI

thesan:
### ✅ Checklist

  - ...

### 📝 Description

Implement the new search component UI.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18121]

---

[LIVE-18121]: https://ledgerhq.atlassian.net/browse/LIVE-18121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
+++ b/libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
@@ -1,0 +1,39 @@
+import React, { ChangeEvent, useMemo, useRef } from "react";
+import { Icons } from "../../../assets";
+import { useDebouncedCallback } from "../../hooks";
+import { Input } from "..";
+
+type InputProps = React.ComponentProps<"input">;
+type Props = Readonly<
+  InputProps & {
+    onDebouncedChange?: (current: string, prev: string) => void;
+    debounceTime?: number;
+  }
+>;
+
+export function Search({ onDebouncedChange, debounceTime = 500, onChange, ...props }: Props) {
+  const initialValue = props.value ?? props.defaultValue ?? "";
+  const prevValue = useRef(String(initialValue));
+
+  const handleDebouncedChange = useDebouncedCallback(
+    useMemo(() => {
+      if (!onDebouncedChange) return;
+      return (event: ChangeEvent<HTMLInputElement>) => {
+        const current = event.target.value;
+        onDebouncedChange(current, prevValue.current);
+        prevValue.current = current;
+      };
+    }, [onDebouncedChange]),
+    debounceTime,
+  );
+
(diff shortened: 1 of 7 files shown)


libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
```diff
@@ -0,0 +1,39 @@
...
+        prevValue.current = current;
+      };
+    }, [onDebouncedChange]),
+    debounceTime,
+  );
+
+  const handleChange = useMemo(() => {
+    if (!handleDebouncedChange && !onChange) return;
+    return (event: ChangeEvent<HTMLInputElement>) => {
+      onChange?.(event);
```
themooneer, line 33: no need for the `?` here

libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
```diff
@@ -0,0 +1,39 @@
...
+      handleDebouncedChange?.(event);
```
themooneer, line 34: same here

libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
thesan: @themooneer both `?` needed `handleDebouncedChange` (x)or `onChange` could be `undefined` just not both

cgrellard-ledger:
Should we be adding unit tests for the useDebounced ?

libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
themooneer: you already checked on them before no ? `if (!handleDebouncedChange && !onChange) return;`

cgrellard-ledger (approved):

ejaycoleman (approved):

libs/ui/packages/react/src/pre-ldls/components/Search/Search.tsx
thesan: Yes but `!handleDebouncedChange && !onChange` early return if both are undefined but doesn't when only one of the 2 is undefined (actually it's not a the xor since both could also be defined sorry for the confusion).

E.g in the case where `handleDebouncedChange` is set but `onChange` is not:
`!handleDebouncedChange && !onChange` is `false` so there is no early return (we need to pass a `handleChange`) however `onChange?.(event)` is needed to not call `undefined` but `handleDebouncedChange` will be call

PR #10032, opened 2025-04-23
ci: Naming updates to workflows and jobs

lewisd5:
Updating workflow naming for consistency
https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/5775295007/Workflow+Job+Naming

Ticket: https://ledgerhq.atlassian.net/browse/LIVE-18178

Green Run: https://github.com/LedgerHQ/ledger-live/actions/runs/14730101641

### DIFF
--- a/.github/workflows/test-ui-e2e-only-desktop.yml
+++ b/.github/workflows/test-ui-e2e-only-desktop.yml
@@ -1,4 +1,4 @@
-name: "@Desktop • UI e2e • Test App"
+name: "[Desktop] - E2E Test - Scheduled"
 run-name: "@Desktop • UI e2e • Test App triggered by ${{ inputs.login || github.actor }} ${{ format('on ref {0}', github.ref_name) }}"
 
 on:
@@ -62,7 +62,7 @@ permissions:
 
 jobs:
   e2e-tests-linux:
-    name: "Desktop Tests E2E (Ubuntu)"
+    name: "Ubuntu E2E"
     env:
       NODE_OPTIONS: "--max-old-space-size=7168"
       INSTRUMENT_BUILD: true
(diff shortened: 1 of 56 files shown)


.github/workflows/test-ui-e2e-only-desktop.yml
```diff
@@ -1,4 +1,4 @@
-name: "@Desktop • UI e2e • Test App"
+name: "[Desktop] - E2E Test - Scheduled"
```
ypolishchuk-ledger, line 1: but it's not only scheduled job, it is as well manually triggered one for example for release validation

Couto (approved):

.github/workflows/test-ui-e2e-only-desktop.yml
ypolishchuk-ledger: I'm just afraid it will confuse people wanting to run this workflow manually for feature validation or for release validation

hedi-edelbloute: I agree with @ypolishchuk-ledger, there will always be multiple triggers and it can be confusing

I'm ok with testing it that way 🤷 maybe it will come naturally with usage

hedi-edelbloute (approved):
Is there way to lint test it ? It's nice to add a convention but if it's not applied long term we'll have to do another round later

.github/workflows/test-ui-e2e-only-desktop.yml
Couto: The previous naming had nothing in relation to how the action was triggered. So in order to know if it could be manual or not, devs would have to either check the actions page, or open the workflow file.

Now it states the primary intention. The option to trigger manually is still there. 
I understand that it might cause confusion, but it's one that's easily fixed by just doing the steps that devs had to do before anyway.

KVNLS (approved):
Thanks a lot!
Maybe you will need to "Disable workflows" once merged in Actions tabs as I think we will have duplicates otherwise (tbc, not sure of myself).

Couto (approved):

PR #10137, opened 2025-05-05
:sparkles: (lld/ui) MAD asset selection drawer

LucasWerey:
### ✅ Checklist

  - MAD project

### 📝 Description

Currently, the `src/newArch/features/ModularDrawer/components/SelectAsset.tsx` component isn't used; it will be in future PRs that implement the logic. 

This PR has some small UI fixes inside the pre-llds lib, implementation of the search component and the asset List component. It's a bit tricky as it's a virtual list

Also skeleton component has been moved from Collectibles/ to global Components/ as we need it for MAD too

I made a change on `apps/ledger-live-desktop/src/renderer/screens/market/components/Table.tsx` since it was making my test fail for an unknown reason. This [topic](https://stackoverflow.com/questions/71876240/jest-cannot-create-styled-component-for-component-undefined) was the solution, but still not sure why the market component makes the test fail. 

https://github.com/user-attachments/assets/3566d7d0-9634-4509-bcee-1863bdb21c60

https://github.com/user-attachments/assets/938bd4e1-b92e-43f0-8040-e1147645ab9e

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18149](https://ledgerhq.atlassian.net/browse/LIVE-18149)

---

[LIVE-18149]: https://ledgerhq.atlassian.net/browse/LIVE-18149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectAsset.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const onVisibleItemsScrollEnd = () => {
+    //TODO: Add logic to handle scroll end event
+  };
+
+  if (shouldDisplayLoading) {
+    return (
+      <Flex flexDirection="column" flex="1 1 auto" overflow="hidden" rowGap="8px">
+        {Array.from({ length: 10 }, (_, index) => (
+          <Skeleton key={index} barHeight={64} minHeight={64} />
+        ))}
```
mcayuelas-ledger, line 52: Am I missing something, or couldn’t the list itself handle the skeleton?
For example, the component could take a skeletonType prop and apply it based on the list size and virtualization?

LucasWerey: This skeleton it's still in discussion with design, as there is no design for this, and I am not sure of the expected purpose of this, which is why I put it like this 
However, there is already an infinite loader for infinite/paginated lists

But I assume there is a way to do it

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/Search/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  const { searchQuery, handleSearch } = useSearch({
+    setItemsToDisplay,
+    items,
+    source,
+    flow,
+  });
+
+  return (
+    <Input
```
thesan, line 18: Please use the Search component from the `@ledgerhq/react-ui/pre-ldls` https://react-ui-storybook.vercel.app/?path=/story/preldls-components-search--default (the main difference is the `onDebouncedChange` which is not usefull rn but it should be later).

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectNetwork.tsx
```diff
@@ -3,14 +3,22 @@ import { CryptoOrTokenCurrency } from "@ledgerhq/types-cryptoassets";
 import { NetworkList } from "@ledgerhq/react-ui/pre-ldls";
 import { track } from "~/renderer/analytics/segment";
 import TrackPage from "~/renderer/analytics/TrackPage";
+import { Box, Flex } from "@ledgerhq/react-ui/index";
```
thesan, line 6: We should avoid relying on the legacy DS for this new components. Since we won't be using Tailwind in LLD directly maybe it would make sense to export a Flex component from the pre-ldls 🤔

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectAssetFlow.tsx
```diff
@@ -64,25 +42,22 @@ function SelectAssetFlow(props: SelectAccountAndCurrencyDrawerProps) {
...
           {t("drawers.selectCurrency.title")}
         </Text>
       </HeaderContainer>
       <SelectorContent>
         <SearchInputContainer>
-          <SearchInput
-            data-testid="select-asset-drawer-search-input"
-            value={searchValue}
-            onChange={setSearchValue}
+          <SearchItem
```
thesan, line 52: It's fine to replace the search here so we see it, but the search should be on the SelectAsset component. Because it only shows durring the asset selection not the network selection

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/Search/useSearch.ts
```diff
@@ -0,0 +1,56 @@
...
+export type SearchProps = {
+  setItemsToDisplay: React.Dispatch<React.SetStateAction<CryptoOrTokenCurrency[]>>;
+  source: string;
+  flow: string;
+  items: CryptoOrTokenCurrency[];
+};
+
+export type SearchResult = {
+  handleSearch: (query: string) => void;
+  setSearchQuery?: React.Dispatch<React.SetStateAction<string>>;
```
thesan, line 16: Is it necessary to expose `setSearchQuery` ? It looks to me that the only correct way to update the `searchQuery` is by calling `handleSearch`. Am I missing something ?

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/tests/Search/useSearch.test.ts
```diff
@@ -0,0 +1,108 @@
+import { renderHook, act } from "tests/testSetup";
+import { useSearch } from "../../Search/useSearch";
+import { track } from "~/renderer/analytics/segment";
+import { CryptoOrTokenCurrency } from "@ledgerhq/types-cryptoassets";
+
+describe("useSearch", () => {
```
thesan, line 6: It's not so clear in the guidelines but my understanding was that unit tests should be nested within the component folder (so here inside Search next to the hook it tests). Is it wrong ?

LucasWerey: LGTM moved ✅

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/Search/useSearch.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    if (query.trim().length < 2) {
+      return setItemsToDisplay(items);
+    }
+
+    const results = fuse.search(query).map(result => result.item);
+
+    setItemsToDisplay(results);
+
+    track("asset_searched", { query, page: source, flow });
```
thesan, line 47: Sorry I missed that before: I forgot we wanted to track the search query. As long as the search remains local it's fine not to debounce it (since fuse.js is so efficient) but if we send a request on every character maybe it would be best to debounce the search a bit. WDYT ?

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectAsset.tsx
```diff
@@ -0,0 +1,76 @@
...
+  source,
+  flow,
+  onAssetSelected,
+}: SelectAssetProps) => {
+  const shouldDisplayLoading = !assetTypes || assetTypes.length === 0;
+
+  const onClick = (asset: AssetType) => {
+    track("asset_clicked", { asset, page: CURRENT_PAGE, flow });
+
+    const selectedAsset = assetsToDisplay.find(({ ticker }) => ticker === asset.ticker);
```
thesan, line 37: It seems tickers are not unique in practice is there a different prop we could use ?

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectAsset.tsx
```diff
@@ -0,0 +1,76 @@
...
+export const SelectAsset = ({
+  assetTypes,
+  assetsToDisplay,
+  source,
+  flow,
+  onAssetSelected,
+}: SelectAssetProps) => {
+  const shouldDisplayLoading = !assetTypes || assetTypes.length === 0;
+
+  const onClick = (asset: AssetType) => {
```
thesan, line 34: I think it's worth wrapping this in a useCallback to avoid unnecessary `AssetList` re-renders

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/SelectAsset.tsx
LucasWerey: I can use the id of the asset to do that. I will update AssetType type

LucasWerey: Done ✅

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/Search/useSearch.ts
LucasWerey: Done ✅

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/components/Search/useSearch.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const fuse = useMemo(() => new Fuse(items, FUZE_OPTIONS), [items]);
+
+  const trackSearch = useCallback(
+    (current: string, previous: string) => {
+      const query = current.trim();
+      const prevQuery = previous.trim();
+      if (query === prevQuery) return;
+
+      if (query && query.trim() !== "") {
```
thesan, line 43: For clarity lets just have one early return:
```suggestion
      if (!query || query === prevQuery) return;

```
Also query doesn't need to be trimmed twice

LucasWerey: done ✅

apps/ledger-live-desktop/src/renderer/screens/market/components/Table.tsx
thesan, line 1: It looks fine but why is this change in this PR ?

PR #10153, opened 2025-05-06
fix: use base58 encoding for Solana signMessage signature

Justkant:
### ✅ Checklist

  - solana sign message

### 📝 Description

The signature returned by signMessage for the Solana was not in base58 as expected by a lot of apps

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/ledger-live-common/src/wallet-api/react.ts
+++ b/libs/ledger-live-common/src/wallet-api/react.ts
@@ -446,7 +446,11 @@ export function useWalletAPIServer({
                 if (done) return;
                 done = true;
                 tracking.signMessageSuccess(manifest);
-                resolve(Buffer.from(signature.replace("0x", ""), "hex"));
+                resolve(
+                  signature.startsWith("0x")
+                    ? Buffer.from(signature.replace("0x", ""), "hex")
+                    : Buffer.from(signature),
+                );
               },
               onCancel: () => {
                 if (done) return;
--- a/.changeset/fresh-fishes-change.md
+++ b/.changeset/fresh-fishes-change.md
@@ -1,0 +1,6 @@
+---
+"@ledgerhq/coin-solana": minor
+"@ledgerhq/live-common": minor
+---
+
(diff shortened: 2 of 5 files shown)


libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -446,7 +446,11 @@ export function useWalletAPIServer({
                 if (done) return;
                 done = true;
                 tracking.signMessageSuccess(manifest);
-                resolve(Buffer.from(signature.replace("0x", ""), "hex"));
+                resolve(
+                  signature.startsWith("0x")
+                    ? Buffer.from(signature.replace("0x", ""), "hex")
+                    : Buffer.from(signature),
```
francois-guerin-ledger, line 452: Question : Is is OK to teat `signature` as if is utf8 encoded, where `bs58.decode` might be more appropriated in the very specific scenario of a Solana signature ?
My understanding is that it becomes a guessing game if we start to deal with multiple encodings, maybe we want to be explicit on which encoding is used in the future

francois-guerin-ledger (approved):

Wozacosta (approved):

libs/ledger-live-common/src/wallet-api/react.ts
Justkant: > Is is OK to teat `signature` as if is utf8 encoded, where bs58.decode might be more appropriated in the very specific scenario of a Solana signature ?

I think it's correct to treat `signature` as utf8 encoded

Here it's more evm that was doing something special IMO, now we just pass it as it was encoded by the hw-signMessage of each coin if not explicitly passed as hex

> My understanding is that it becomes a guessing game if we start to deal with multiple encodings, maybe we want to be explicit on which encoding is used in the future

Totally agree on that part, we don't want to jump between encodings and that's why I chose to change this here
The alternative would have been to send this from Solana hw-signMessage and feels worse for me
```ts
return { signature: "0x" + bs58.encode(result.signature).toString("hex") };
```

PR #10192, opened 2025-05-13
fix(live-18992): retry fetching incorrect trc20 ops 

jprudent:
### ✅ Checklist

  - now, if we can't get a well formed trc20 operations, we prefer to crash instead of skipping the operation
  - a retry mechanism (limited to 3) has been implemented, that could slow down a bit the retrieval of operations. 

### 📝 Description

TRC20 operations are fetched with that call: `/v1/accounts/${addr}/transactions/trc20?${queryParams}&get_detail=true`
ex: https://tron.coin.ledger.com/v1/accounts/TRRYfGVrzuUvJYRe9UaA8KqxjgVSwU9m6L/transactions/trc20?get_detail=true&limit=1&min_timestamp=1709131368000&order_by=block_timestamp,desc&get_detail=true

unfortunately, sometimes, the result is not “complete”, we miss an important part, the `detail.ret[0]` path
This path contains the fees
A new call *can* return the result with expected information

consequence is that when it’s missing, the logic layer will skip  the operation from the result:
- unprotected access that raises an exception https://github.com/LedgerHQ/ledger-live/blob/c342be772cdd3d321fa5556c180a12ed07a1bceb/libs/coin-modules/coin-tron/src/network/format.ts#L19
- an here is the catch that returns undefined for the whole operation  https://github.com/LedgerHQ/ledger-live/blob/c342be772cdd3d321fa5556c180a12ed07a1bceb/libs/coin-modules/coin-tron/src/network/format.ts#L53
- => we are missing some operations sometimes

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-18992

---

### DIFF
--- a/libs/coin-modules/coin-tron/src/network/index.ts
+++ b/libs/coin-modules/coin-tron/src/network/index.ts
@@ -554,25 +554,26 @@ export async function fetchTronAccountTxs(
     return Array.isArray(ret) && ret.length > 0;
   }
 
-  function flagInvalids(txs: Trc20API[]): number[] {
+  function getInvalidTxIndexes(txs: Trc20API[]): number[] {
     const invalids: number[] = [];
     for (let i = 0; i < txs.length; i++) {
       if (!isValid(txs[i])) {
         invalids.push(i);
       }
     }
+    txs.filter(tx => !isValid(tx)).map((tx, index) => index);
     return invalids;
   }
 
-  function pre(pred: boolean, message: string) {
-    if (!pred) {
+  function assert(predicate: boolean, message: string) {
+    if (!predicate) {
       throw new Error(message);
     }
   }
 
   // Merge the two results
   function mergeAccs(acc1: Acc, acc2: Acc): Acc {
-    pre(acc1.txs.length == acc2.txs.length, "accs should have the same length");
+    assert(acc1.txs.length == acc2.txs.length, "accs should have the same length");
     const accRet: Acc = { txs: acc1.txs, invalids: [] };
     acc1.invalids.forEach(invalidIndex => {
       acc2.invalids.includes(invalidIndex)
@@ -584,14 +585,14 @@ export async function fetchTronAccountTxs(
 
   // see LIVE-18992 for an explanation to why we need this
   async function getTrc20TxsWithRetry(acc: Acc | null, times: number): Promise<Trc20API[]> {
-    pre(
+    assert(
       times > 0,
       "getTrc20TxsWithRetry: couldn't fetch trc20 transactions after several attempts",
     );
     const ret = await callTrc20Endpoint();
     const thisAcc: Acc = {
       txs: ret,
-      invalids: flagInvalids(ret),
+      invalids: getInvalidTxIndexes(ret),
     };
     const newAcc = acc ? mergeAccs(acc, thisAcc) : thisAcc;
     if (newAcc.invalids.length == 0) {


libs/coin-modules/coin-tron/src/network/index.test.ts
```diff
@@ -64,3 +74,86 @@ describe("fetchTronAccountTxs", () => {
...
 });
+
+describe("fetchTronAccountTxs with invalid TRC20 (see LIVE-18992)", () => {
+  const tx1Hash = "1237889e91c0ebbe389436c341865df09921f8f0c029d9286102372cbaadc585";
+  const tx2Hash = "154164dd04482ae78f930033d0ad95730b8b19fde171a33c3920d18c228426ab";
+  let counterGetTrc20 = 0;
+  const invalidTrc20Handler = http.get(
+    `${TRON_BASE_URL_TEST}/v1/accounts/:addr/transactions/trc20`,
+    () => {
+      const ret: any = JSON.parse(JSON.stringify(TRC20_FIXTURE));
```
Moustafa-Koterba, line 85: why do you need `stringify` then `parse` ?

libs/coin-modules/coin-tron/src/network/index.ts
```diff
@@ -536,13 +536,76 @@ export async function fetchTronAccountTxs(
...
+    const invalids: number[] = [];
+    for (let i = 0; i < txs.length; i++) {
+      if (!isValid(txs[i])) {
+        invalids.push(i);
+      }
+    }
+    return invalids;
+  }
+
+  function pre(pred: boolean, message: string) {
```
Moustafa-Koterba, line 567: wdyt
```suggestion
    function assert(predicate: boolean, message: string) {
```

Canestin (approved):

libs/coin-modules/coin-tron/src/network/index.ts
```diff
@@ -536,13 +536,76 @@ export async function fetchTronAccountTxs(
...
+    invalids: number[];
+  };
+
+  function isValid(tx: Trc20API): boolean {
+    const ret = tx?.detail?.ret;
+    return Array.isArray(ret) && ret.length > 0;
+  }
+
+  function flagInvalids(txs: Trc20API[]): number[] {
+    const invalids: number[] = [];
```
Moustafa-Koterba, line 558: I think you can do
```suggestion
    return txs.filter(tx => !isValid(tx))
      .map((tx, index) => index);
```

libs/coin-modules/coin-tron/src/network/index.ts
```diff
@@ -536,13 +536,76 @@ export async function fetchTronAccountTxs(
...
+    txs: Trc20API[];
+    invalids: number[];
+  };
+
+  function isValid(tx: Trc20API): boolean {
+    const ret = tx?.detail?.ret;
+    return Array.isArray(ret) && ret.length > 0;
+  }
+
+  function flagInvalids(txs: Trc20API[]): number[] {
```
Moustafa-Koterba, line 557: wdyt 
```suggestion
  function getInvalidTxIndexes(txs: Trc20API[]): number[] {
```

Moustafa-Koterba (dismissed):

libs/coin-modules/coin-tron/src/network/index.test.ts
jprudent: it's for making a deep copy of the object

I had a strange bug because of mutation. Do you know something more idiomatic ?

libs/coin-modules/coin-tron/src/network/index.ts
Moustafa-Koterba: yes my bad, `filter` create [a new array](https://www.w3schools.com/jsref/jsref_filter.asp)

libs/coin-modules/coin-tron/src/network/index.test.ts
Moustafa-Koterba: Unfortunately right now, no 😞 
What kind of bug ? The variable is a constant and you just read values as I see in the file

Moustafa-Koterba (approved):

libs/coin-modules/coin-tron/src/network/index.test.ts
jprudent: no, a few lines behind I change it to artificially make bad trc20 operations
my change is nested, and when I copied with `{...TRC20_FIXTURE}` it actually changed the `TRC20_FIXTURE` even when I added the `as const` (which have a very limited usage actually ...)

PR #10197, opened 2025-05-13
[CoinFmk] Remove generated transaction

sprohaszka-ledger:
### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18301]

---


libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -59,7 +140,276 @@ export function getAccountBridgeByFamily(family: string, accountId?: string): Ac
...
+      return jsBridges.ton;
+    case "tron":
+      return jsBridges.tron;
+    case "vechain":
+      return jsBridges.vechain;
+    case "xrp":
+      return jsBridges.xrp;
+    default:
+      throw new CurrencyNotSupported("bridge not found " + family);
+  }
```
francois-guerin-ledger, line 238:
```suggestion
  const bridge = jsBridges[family];
  if (!bridge) {
      throw new CurrencyNotSupported("bridge not found " + family);
  }
  return bridge;
```

libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -59,7 +140,276 @@ export function getAccountBridgeByFamily(family: string, accountId?: string): Ac
...
+  return BigNumber.isBigNumber(tx.amount);
+}
+export function callSerializeBridgeFunc(
+  tx: TransactionCommon | TransactionCommonRaw,
+  func: string,
+) {
+  return isTransaction(tx)
+    ? callSerializeBridgeFuncOnTransaction(tx, func)
+    : callSerializeBridgeFuncOnTransactionRaw(tx, func);
+}
```
francois-guerin-ledger, line 255: I don't understand the need of such a distinction. `callSerializeBridgeFuncOnTransaction` and `callSerializeBridgeFuncOnTransactionRaw` do call the same methods under the hood.

The only difference is calling `isXXXTransaction` vs `isXXXTransactionRaw` predicates. We can simplify with

```typescript
if (isAlgorandTransaction(tx) || isAlgorandTransactionRaw(tx)) {
    return jsBridges.algorand.serializationBridge[func](tx);
}
...
```

below and keep only one function.

Also why don't you use `getBridgeByTransaction` here ?

francois-guerin-ledger (commented):
I get the overall idea, it seems fair. I suppose you also plan to remove the other files that are in the "generated" folder, correct ?

Apart from my few comments, we must check why the CI fails.

PR #10211, opened 2025-05-14
fix: optimize the "Add account" flow in Hedera

md-blockydevs:
### ✅ Checklist

  - Hedera account scanning is much faster

### 📝 Description

This pull request fixes slow account scanning when adding a new Hedera account. 

Here is a demo of how it works now:
https://drive.google.com/file/d/1xOLbOj8AEUaMnO_UsjWl23hD9nreaCdZ/view. 

You can notice that it takes more than 300 requests per account to get the list of transactions from the Mirror Node API. This can be easily reproduced by going to: https://mainnet-public.mirrornode.hedera.com/api/v1/transactions?account.id=0.0.8835924&timestamp=gt:0 and trying to fetch the transactions until `links.next = null`. 

The proposed fix is to simply break the loop when the transactions array is empty, avoiding unnecessary pagination when we know no more data can be returned.

Here is a demo of how it works after the change is made: https://drive.google.com/file/d/1M4Fyq2ebZJixOpv3NtrpD9ggvE9B4DQs/view

---

### DIFF
--- a/libs/coin-modules/coin-hedera/src/bridge/synchronisation.ts
+++ b/libs/coin-modules/coin-hedera/src/bridge/synchronisation.ts
@@ -36,16 +36,14 @@ export const getAccountShape: GetAccountShape<Account> = async (
   // grab latest operation's consensus timestamp for incremental sync
   const oldOperations = initialAccount?.operations ?? [];
   const latestOperationTimestamp = oldOperations[0]
-    ? Math.floor(oldOperations[0].date.getTime() / 1000)
+    ? new BigNumber(Math.floor(oldOperations[0].date.getTime() / 1000))
     : null;
 
   // merge new operations w/ previously synced ones
   const newOperations = await getOperationsForAccount(
     liveAccountId,
     address,
-    typeof latestOperationTimestamp === "number"
-      ? new BigNumber(latestOperationTimestamp).toString()
-      : null,
+    latestOperationTimestamp ? latestOperationTimestamp.toString() : null,
   );
   const operations = mergeOps(oldOperations, newOperations);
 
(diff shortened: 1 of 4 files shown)


libs/coin-modules/coin-hedera/src/bridge/synchronisation.ts
```diff
@@ -37,13 +37,15 @@ export const getAccountShape: GetAccountShape<Account> = async (
   const oldOperations = initialAccount?.operations ?? [];
   const latestOperationTimestamp = oldOperations[0]
     ? Math.floor(oldOperations[0].date.getTime() / 1000)
```
may01, line 39: Can we do it `BigNumber` here instead of line 47 and avoid type check on line 46?

may01 (approved):
LGTM

libs/coin-modules/coin-hedera/src/bridge/synchronisation.ts
md-blockydevs: Sure, nice way to avoid type checking caused by `0`

may01 (approved):

Moustafa-Koterba (dismissed):

francois-guerin-ledger (commented):
The early `break` certainly helps, but adding this transaction fetch limit is also part of the optimisation solution.

Maybe you can a unit test to reflect it ?

md-blockydevs:
> The early `break` certainly helps, but adding this transaction fetch limit is also part of the optimisation solution.
> 
> Maybe you can a unit test to reflect it ?

It's not a bad idea, I've just pushed the `libs/coin-modules/coin-hedera/src/api/mirror.test.ts` that verifies both `break` and `limit` changes.

I also wrote down in my notes that `"@ledgerhq/live-network/network"` is deprecated, but I didn't want to create a mess in this PR.

may01 (approved):

PR #10225, opened 2025-05-14
core(coin-solana,hw-app-solana): support SPL tokens dynamically

francois-guerin-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Provide dynamic token information to Solana app, downloaded from the CAL.

Note we need to fetch a specific certificate from the CAL as well for this use case.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/tools/actions/composites/setup-coin-tester-env/action.yml
+++ b/tools/actions/composites/setup-coin-tester-env/action.yml
@@ -3,10 +3,12 @@ description: "Installs deps, pulls docker, builds modules"
 runs:
   using: "composite"
   steps:
-    - run: docker pull ghcr.io/ledgerhq/speculos:latest
+    # Option `-p` of Speculos is only available on `master` tag
+    - run: docker pull ghcr.io/ledgerhq/speculos:master
       shell: bash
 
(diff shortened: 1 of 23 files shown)


Justkant (approved):

Justkant (commented):
The only detail we didn't check or talked about is if we want this code to still support older versions of the app
It means we don't need to force update the app for all users on send flows

Justkant (approved):

Justkant (dismissed):

Justkant (approved):

tools/actions/composites/setup-coin-tester-env/action.yml
```diff
@@ -3,10 +3,12 @@ description: "Installs deps, pulls docker, builds modules"
 runs:
   using: "composite"
   steps:
-    - run: docker pull ghcr.io/ledgerhq/speculos:latest
```
lewisd5, line 6: This looks like there needs to be some changes in versioning here to get proper version pinning going.

That being said as this is consumed only in coin tester, you can add this composite to be owned by coin-integration. I don't think the CI team need to input on this action

PR #10305, opened 2025-05-21
feat: add custom webview loader for Buy/Sell

mwiercinska:
### ✅ Checklist

  - ...

### 📝 Description

Adding a custom loader to enable custom transition between Buy/Sell and providers in LLM and LLD.

**Demo**
| LLD | LLM iOS | LLM Android |
| ----- | ----- | ----- | 
| https://github.com/user-attachments/assets/0996f9eb-cf00-425f-9615-b84c6d79c9a7 | https://github.com/user-attachments/assets/b2e74904-afdb-4f85-9623-7d2fced2d90e | https://github.com/user-attachments/assets/d3ca0833-d075-4285-9b28-f0e9ad98be82 |

### ❓ Context

- **JIRA or GitHub link**: [LIVE-17147](https://ledgerhq.atlassian.net/browse/LIVE-17147)

---

[LIVE-17147]: https://ledgerhq.atlassian.net/browse/LIVE-17147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
@@ -398,7 +398,7 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
       onStateChange,
       hideLoader,
       webviewStyle: customWebviewStyle,
-      onWidgetLoadedChange,
+      Loader = DefaultLoader,
     },
     ref,
   ) => {
@@ -448,10 +448,6 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
       customWebviewStyle,
     );
 
-    useEffect(() => {
-      onWidgetLoadedChange?.(widgetLoaded);
-    }, [onWidgetLoadedChange, widgetLoaded]);
-
     const isDapp = !!manifest.dapp;
     const preloader = isDapp ? "webviewDappPreloader" : "webviewPreloader";
 
@@ -489,14 +485,20 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
           {...webviewProps}
           {...webviewPartition}
         />
-        {!widgetLoaded && !hideLoader ? (
-          <Loader>
-            <BigSpinner size={50} />
-          </Loader>
-        ) : null}
+        {!hideLoader ? <Loader manifest={manifest} isLoading={!widgetLoaded} /> : null}
       </>
     );
   },
 );
+
+function DefaultLoader({ isLoading }: { isLoading: boolean }) {
+  if (!isLoading) return null;
+
+  return (
+    <Loader>
+      <BigSpinner size={50} />
+    </Loader>
+  );
+}
 
 WalletAPIWebview.displayName = "WalletAPIWebview";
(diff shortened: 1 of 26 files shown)


sarneijim (dismissed):

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/index.tsx
```diff
@@ -38,7 +39,21 @@ export const WebViewWrapper = styled.div<WebViewWrapperProps>`
...
+export default function WebPTXPlayer({
+  manifest,
+  inputs,
+  CustomLoader,
+}: WebviewProps & {
+  CustomLoader?: React.ComponentType<{
+    manifest: LiveAppManifest;
+    isLoading: boolean;
+  }>;
+}) {
```
Justkant: I think we should be able to directly use the `Loader` props from `WebviewProps` instead of adding a new `CustomLoader` with the same type
```suggestion
  Loader,
}: WebviewProps) {
```

hhumphrey-ledger: Good shout! Update added here [cc5d404](https://github.com/LedgerHQ/ledger-live/pull/10305/commits/cc5d4049d6233eb1d31998ea6c3f85d28409ea28)

apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
```diff
@@ -290,6 +297,7 @@ export const WebPTXPlayer = ({
         onStateChange={setWebviewState}
         customHandlers={customHandlers}
       />
+      {CustomLoader ? <CustomLoader manifest={manifest} isLoading={webviewState.loading} /> : null}
```
Justkant: Should we also use the `Loader` prop here instead of putting the Loader under the webview ?

hhumphrey-ledger: Mobile is working slightly differently to desktop in that the loader is sent to the [react native webview component](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx#L72), meaning that that component internally handles when to show / hide the loader. Because we meed the partner loader to be on screen for a minimum amount of time we need full control over when we show and hide it meaning we can't use this original pattern.

HOWEVER, it turns out the PTX player had a custom loader which was just overlaying the WalletAPI one so I have:
- Sent this as a prop to `WalletApiWebview` so we are no longer rendering two loaders
- Renamed the custom loader in this PR to `ProviderInterstitial` to try and make it clear why this isn't being used as a loader...it has additionally requirements that a loader does not usually have.

I've also renamed the Loader in desktop. It is a component that is meant to be used only for provider transitions so we might as well be explicit about it.

Justkant (approved):

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
ooke-ledger, line 1: I noticed that the Loader component receives both manifest and isLoading props, but the current DefaultLoader implementation only utilizes isLoading. Is the manifest prop intended for future use, or could it be removed?

hhumphrey-ledger: That's because the Loader that we are potentially injecting to that component needs both of those props. True the DefaultLoader doesn't need both but that is a fallback

ooke-ledger (approved):

PR #10395, opened 2025-06-02
fix: improve solana raw tx handling [LIVE-19975]

Justkant:
### ✅ Checklist

  - Wallet-API solana raw tx

### 📝 Description

- Implemented handling of raw Solana transactions in the SendFunds summary screen.
- Updated transaction building and signing processes to accommodate raw transaction types.
- Introduced new types and command descriptors for raw transactions in the Solana module.
- Enhanced error handling and user feedback for raw transactions.
- Added mock data and tests to ensure proper functionality of raw transaction features.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-19307]

---

[LIVE-19307]: https://ledgerhq.atlassian.net/browse/LIVE-19307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/modals/SignTransaction/steps/StepSummary.tsx
```diff
@@ -84,9 +84,16 @@ const StepSummary = (props: StepProps) => {
 
   const memo = "memo" in transaction ? transaction.memo : undefined;
 
+  const isSolanaRawTransaction = "raw" in transaction && transaction.raw;
+
   return (
     <Box flow={4} mx={40}>
       <TrackPage category="Sign Flow" name="Step Summary" />
+      {isSolanaRawTransaction ? (
```
Moustafa-Koterba, line 92: same for others in the file
```suggestion
      {isSolanaRawTransaction ?? (
```

apps/ledger-live-desktop/tests/models/LiveAppWebview.ts
```diff
@@ -166,6 +166,11 @@ export class LiveAppWebview {
     return webview.getByTestId("transaction-sign-solana").click();
   }
 
+  async transactionSignRawSolana() {
```
Moustafa-Koterba, line 169: wdyt
```suggestion
  async clickOnSignSolanaRawTransaction() {
```

apps/ledger-live-mobile/src/families/solana/SendRowsCustom.tsx
```diff
@@ -28,11 +28,17 @@ type Props = {
   transaction: Transaction;
 } & Navigation;
 
+export function isModelSupported(model: TransactionModel) {
+  if (model.kind === "transfer" || model.kind === "token.transfer") return model;
```
Moustafa-Koterba, line 32:
```suggestion
  return model.kind === "transfer" || model.kind === "token.transfer" ? return model : undefined;
```

apps/ledger-live-mobile/src/screens/SendFunds/04-Summary.tsx
```diff
@@ -276,6 +279,25 @@ function SendSummary({ navigation, route }: Props) {
             />
           </>
         ) : null}
+
+        {isSolanaRawTransaction ? (
```
Moustafa-Koterba, line 283: same here

apps/ledger-live-mobile/e2e/specs/wallet-api.spec.ts
Moustafa-Koterba, line 1: Why did you comment code here ? Something is missing ?

apps/ledger-live-mobile/src/families/solana/SendRowsCustom.tsx
Justkant: The early return looks cleaner, what's the upside of using a ternary ?

apps/ledger-live-mobile/e2e/specs/wallet-api.spec.ts
Justkant: Like I said the other day, it somehow doesn't work at all on detox, I compared with the other test we have for transaction sign and somehow I get into a strange flaky issue where detox doesn't wait for the text to be inputed before opening the transaction sign but it somehow still fill the inputs

apps/ledger-live-desktop/src/renderer/modals/SignTransaction/steps/StepSummary.tsx
Justkant: Why would you use nullish coalescing on a boolean ? it would never return the jsx this way

apps/ledger-live-desktop/tests/models/LiveAppWebview.ts
Justkant: we'll keep the same convention as the other actions for now

libs/coin-modules/coin-solana/src/prepareTransaction.test.ts
```diff
@@ -33,6 +22,7 @@ jest.mock("./estimateMaxSpendable", () => {
 describe("testing prepareTransaction", () => {
   it("packs a 'NotEnoughGas' error if the sender can not afford the fees during a token transfer", async () => {
     const preparedTransaction = await prepareTransaction(
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
francois-guerin-ledger, line 25: I suggest we mute this rule in tests in the future

libs/coin-modules/coin-solana/src/rawTransaction.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export async function deriveRawCommandDescriptor(tx: Transaction, api: ChainAPI) {
+  if (!tx.raw) {
+    throw new Error("Raw transaction is required to derive command descriptor");
+  }
+
+  const liveTx = await toLiveTransaction(api, tx.raw);
+
+  return (
+    liveTx.model.commandDescriptor || {
```
francois-guerin-ledger, line 49:
```suggestion
    liveTx.model.commandDescriptor ?? {
```
No need for `||` here, since we are not trying to fallback on `""` or `0` for instance

libs/coin-modules/coin-solana/src/transaction.ts
```diff
@@ -247,6 +249,13 @@ function formatStakeSplit(mainAccount: Account, tx: Transaction, command: StakeS
   return "\n" + str;
 }
 
+function formatRaw(tx: Transaction) {
+  const str = [`  SEND RAW: ${tx.useAllAmount ? " (ALL)" : ""}`, `  TO: ${tx.recipient}`]
+    .filter(Boolean)
```
francois-guerin-ledger:
```suggestion
  const str = [`  SEND RAW: ${tx.useAllAmount ? " (ALL)" : ""}`, `  TO: ${tx.recipient}`]
```
Strings in your array are never empty

francois-guerin-ledger (approved):
Hopping we got this covered by the coin tester too

apps/ledger-live-desktop/src/renderer/modals/SignTransaction/steps/StepSummary.tsx
Moustafa-Koterba: I am tired, I was thinking of 
```suggestion
      {isSolanaRawTransaction && (
```

apps/ledger-live-mobile/src/families/solana/SendRowsCustom.tsx
Moustafa-Koterba: Ternary is more concise, you will have only one expression in the function and you can avoid multiple returns. But it is a matter of taste actually, both solution works

libs/coin-modules/coin-solana/src/transaction.ts
Justkant: True, wdyt, should I update also all the other ones that don't need it ?

libs/coin-modules/coin-solana/src/rawTransaction.ts
Justkant: `commandDescriptor` is an object that can be `undefined` or `null`, in that case there is no difference with nullish coalescing and in fact if it was `""` or `0` we'd prefer to fallback on the actual expect object

Justkant: Also in this case the issue is mainly that the typing system cannot infer or know that we return a `commandDescriptor` from the function for sure because the `Transaction` type for solana defines `commandDescriptor` as possibly `undefined`

libs/coin-modules/coin-solana/src/transaction.ts
francois-guerin-ledger: Yes, let's remove some confusion in our codebase

libs/coin-modules/coin-solana/src/prepareTransaction.test.ts
Justkant: Good point

apps/ledger-live-mobile/src/screens/SendFunds/04-Summary.tsx
```diff
@@ -172,8 +172,9 @@ function SendSummary({ navigation, route }: Props) {
 
   const displayedError = mergeErrors();
 
-  // FIXME: why is recipient sometimes empty?
-  if (!account || !transaction || !transaction.recipient || !currencyOrToken) {
+  const isSolanaRawTransaction = "raw" in transaction && transaction.raw;
+
+  if (!account || !transaction || !currencyOrToken) {
```
KVNLS, line 177: Are you sure we don't break something by removing the check on `transaction.recipient`? It's hard to understand the use case we had with the comment.

Justkant: Yes I did check all the usage for `transaction.recipient` and there is a check now on the component that uses it
I also tested the change with the empty recipient that I can have with the wallet-api, it was probably fixed in this way before instead of showing the UI for such transaction (only possible with the wallet-api)

libs/coin-modules/coin-solana/src/transaction.ts
```diff
@@ -157,7 +157,7 @@ function formatCreateATA(mainAccount: Account, command: TokenCreateATACommand) {
   if (!token) {
     throw new Error(`token for mint "${command.mint}" not found`);
   }
-  const str = [`  OPT IN TOKEN: ${token.ticker}`].filter(Boolean).join("\n");
+  const str = [`  OPT IN TOKEN: ${token.ticker}`].join("\n");
```
francois-guerin-ledger, line 160: Weird to `.join` with only 1 item in the array

francois-guerin-ledger (approved):

libs/coin-modules/coin-solana/src/transaction.ts
Justkant: Yes I had the same feeling but it's used on the other ones and if we want to add lines it's simple like that
I was going to even remove the array and the concat with the new line below but then I don't want to break the new line so I need to be careful using template literals
Or I could keep the concat and just remove the array and the join

PR #10418, opened 2025-06-03
feat(LIVE-19141): update swap screen header navigation

philipptpunkt:
### ✅ Checklist

  - on new swap tab screen the navigation inside the webview has changed. it is now navigating the webview pages rather than closing the live app.

### 📝 Description

Update swap screen header and webview navigation. Replace static navigation screen title with header title based on webview and native navigation. Use native back button for in webview back navigation.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-19141

Updated navigation header - title per page and in-webview navigation with back button

---

### DIFF
--- a/apps/ledger-live-mobile/src/screens/Swap/LiveApp/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/Swap/LiveApp/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from "react";
+import React, { useCallback, useMemo, useRef, useState } from "react";
 
 import { DEFAULT_FEATURES } from "@ledgerhq/live-common/featureFlags/defaultFeatures";
 import useFeature from "@ledgerhq/live-common/featureFlags/useFeature";
@@ -12,13 +12,17 @@ import { Flex, InfiniteLoader } from "@ledgerhq/native-ui";
 import { useTranslation } from "react-i18next";
 import GenericErrorView from "~/components/GenericErrorView";
 import { initialWebviewState } from "~/components/Web3AppWebview/helpers";
-import { WebviewState } from "~/components/Web3AppWebview/types";
+import { WebviewAPI, WebviewState } from "~/components/Web3AppWebview/types";
 import { WebView } from "./WebView";
 import { DefaultAccountSwapParamList, DetailsSwapParamList } from "../types";
 import { StackNavigatorProps } from "~/components/RootNavigator/types/helpers";
 import { SwapNavigatorParamList } from "~/components/RootNavigator/types/SwapNavigator";
 import { ScreenName } from "~/const";
 import { useNetInfo } from "@react-native-community/netinfo";
+import { useSwapNavigationHelper } from "./navigationHandlers/useSwapNavigationHelper";
+import { useNavigation } from "@react-navigation/native";
+import { useSwapAndroidHardwareBackPress } from "./navigationHandlers/useSwapAndroidHardwareBackPress";
+import { useSwapHeaderNavigation } from "./navigationHandlers/useSwapHeaderNavigation";
 
 // set the default manifest ID for the production swap live app
 // in case the FF is failing to load the manifest ID
@@ -43,6 +47,9 @@ export function SwapLiveApp({
   const [webviewState, setWebviewState] = useState<WebviewState>(initialWebviewState);
   const isWebviewError = webviewState?.url.includes("/unknown-error");
 
+  const webviewRef = useRef<WebviewAPI>(null);
+  const navigation = useNavigation();
+
   const swapLiveAppManifestID =
     (ptxSwapLiveAppMobile?.params?.manifest_id as string) || DEFAULT_MANIFEST_ID;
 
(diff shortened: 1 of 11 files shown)


lpaquet-ledger (dismissed):

apps/ledger-live-mobile/src/components/Web3AppWebview/useSwapHeaderNavigation.tsx
```diff
@@ -0,0 +1,119 @@
...
+              onPress={navigateToSwapHistory}
+            >
+              <Icons.Clock color={"neutral.c100"} />
+            </Touchable>
+          </Flex>
+        ),
+        headerTitle: t("transfer.swap2.form.title"),
+      });
+    }
+  }
```
Justkant: I'm not sure we're supposed to call setOptions in the render pass, we should probably move all of it in a useEffect, wdyt ?

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -53,6 +54,8 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
 
     const internalAppIds = useInternalAppIds() || INTERNAL_APP_IDS;
 
+    useSwapHeaderNavigation(webviewRef);
```
ComradeAERGO: WalletAPIWebview is supposed to be agnostic of usage. Shouldn't this be in the PTX child Webview?
`https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx`

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -53,6 +54,8 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
+
```
Justkant: Why do we need to add this swap specific code in the common WalletAPIWebview code ?
Especially because you have access to the ref above in the PTX specific webview code

apps/ledger-live-mobile/src/components/Web3AppWebview/useIsSwapTab.ts
```diff
@@ -0,0 +1,17 @@
+import { useNavigation } from "@react-navigation/native";
```
ComradeAERGO: Web3AppWebview should stay agnostic of usage. This file should be in the WebPTXPlayer folder

apps/ledger-live-mobile/src/components/Web3AppWebview/useSwapHeaderNavigation.tsx
```diff
@@ -0,0 +1,119 @@
+import React, { useCallback } from "react";
```
ComradeAERGO: Same for this file

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
philipptpunkt: fair point, moved the file

apps/ledger-live-mobile/src/components/Web3AppWebview/useSwapHeaderNavigation.tsx
philipptpunkt: You're right - wrapped it in a useEffect

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
philipptpunkt: moved 😇

apps/ledger-live-mobile/src/components/Web3AppWebview/useIsSwapTab.ts
philipptpunkt: moved 😃

apps/ledger-live-mobile/src/components/Web3AppWebview/useSwapHeaderNavigation.tsx
philipptpunkt: moved 😃

lpaquet-ledger (approved):

lpaquet-ledger (approved):

apps/ledger-live-mobile/src/screens/Swap/LiveApp/index.tsx
```diff
@@ -63,6 +70,26 @@ export function SwapLiveApp({
     [params],
   );
 
+  useSwapHeaderNavigation(webviewRef);
+
+  useSwapAndroidHardwareBackPress({
+    webviewRef,
+    canGoBack: webviewState.canGoBack,
+  });
```
Justkant, line 78: Did you checked on the WebView itself for a props allowing you to enable this ?

I think we had a param to do this directly on react-native-webview, it's also fine like this I suppose but it might be better integrated and could also probably support swiping back on both platforms

philipptpunkt: So in general `allowsBackForwardNavigationGestures` is allowed from our side in the webviews and it works fine for iOS and I think for Android "back-swipe" - however it doesn't cover the "hardware" button on Android as far as I know 🤔

PR #10420, opened 2025-06-03
Feat/generic bridge token support stellar

Wozacosta:
### ✅ Checklist

  - ...

### 📝 Description

Remove Stellar bridge and create a new  `bridge` folder into `leger-live-common` this folder should be shared for every Alpaca coin.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18558](https://ledgerhq.atlassian.net/browse/LIVE-18558)

---

[LIVE-18558]: https://ledgerhq.atlassian.net/browse/LIVE-18558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
```diff
@@ -0,0 +1,195 @@
...
+    });
+
+    const blockInfo = await getAlpacaApi(network, kind).lastBlock();
+    console.log("blockInfo", blockInfo);
+    const balanceRes = await getAlpacaApi(network, kind).getBalance(address);
+    console.log("balanceRes", balanceRes);
+    const nativeAsset = balanceRes.find(b => b.asset.type === "native");
+    const nativeBalance = BigInt(nativeAsset?.value ?? "0");
+    const lockedBalance = BigInt(nativeAsset?.locked ?? "0");
+    let spendableBalance = nativeBalance > lockedBalance ? nativeBalance - lockedBalance : 0n;
```
Wozacosta: TODO: recheck

libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
```diff
@@ -3,60 +3,193 @@ import { GetAccountShape, mergeOps } from "@ledgerhq/coin-framework/bridge/jsHel
...
+        creationDate: new Date(), // Optional: replace if available
+        operations: newCoreOps.filter(
+          op =>
+            op.asset?.assetCode === b.asset.assetCode &&
+            op.asset?.assetIssuer === b.asset.assetIssuer,
+        ),
+      }));
 
-      const blockHeight = oldOperations.length ? (oldOperations[0].blockHeight ?? 0) + 1 : 0;
+    const accountInfo = await getAlpacaApi(network, kind).getAccountInfo(address);
```
Wozacosta: TODO: check if we could get rid of getaccountinfo and just recompute `spendableBalance` here,
It's also needed to get assets, (`accountInfo.assets`), check if `tokenAssets` above is enough

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -94,8 +94,12 @@ function compose(tx: string, signature: string, pubkey?: string): string {
   return combine(envelopeFromAnyXDR(tx, "base64"), signature, pubkey);
 }
 
-async function estimate(): Promise<FeeEstimation> {
-  const value = await estimateFees();
+async function estimate(
```
Wozacosta: TODO: document this logic somewhere

libs/coin-framework/src/api/types.ts
```diff
@@ -9,8 +11,27 @@ export type BlockInfo = {
   time?: Date;
 };
 
-type TokenInfoCommon = Record<string, unknown>;
-// TODO add a `token: string` field to the pagination if we really need to support pagination (which is not the case for now)
+export type TokenInfoCommon = {
```
hedi-edelbloute: I do not get this type, we probably need to remove most of those attributes

libs/coin-framework/src/api/types.ts
```diff
@@ -62,6 +88,9 @@ export type Account = {
   address: string;
   balance: bigint;
   currencyUnit: Unit;
+  pendingOperations: number; // NOTE: can get away with only the number of pending operations?
+  spendableBalance: bigint; // NOTE:: check if we can get rid of this one
+  subAccount?: TokenAccount;
```
hedi-edelbloute: We need to remove this

libs/coin-framework/src/api/types.ts
```diff
@@ -62,6 +88,9 @@ export type Account = {
   address: string;
   balance: bigint;
   currencyUnit: Unit;
+  pendingOperations: number; // NOTE: can get away with only the number of pending operations?
```
hedi-edelbloute: I think account type should be removed

libs/coin-framework/src/api/types.ts
```diff
@@ -137,13 +167,16 @@ export type FeeEstimation = {
...
-export type Pagination = { minHeight: number };
+export type Pagination = { minHeight: number } & { pagingToken?: string; limit?: number }; // For evm, XRP, etc. // NOTE: For Stellar
+// NOTE: future proof export type Pagination = Record<string, unknown>;
 
 export type AccountInfo = {
   isNewAccount: boolean;
   balance: string;
   ownerCount: number;
   sequence: number;
+  assets?: BalanceAsset[]; // Optional, depending on the API
```
hedi-edelbloute: This should be removed, coin module need to be stateless

libs/coin-framework/src/api/types.ts
```diff
@@ -163,10 +196,26 @@ export type AlpacaApi<
...
   listOperations: (
     address: string,
-    pagination: Pagination,
+    pagination?: Pagination,
+    lastPagingToken?: string,
   ) => Promise<[Operation<AssetInfo>[], string]>;
 };
 
+// NOTE: taken from coin-stellar/bridge/types
+export type BalanceAsset = {
```
hedi-edelbloute: Stellar specific in coin-framework, nogo

libs/coin-modules/coin-module-boilerplate/src/common-logic/history/listOperations.ts
```diff
@@ -13,9 +13,10 @@ import { BoilerplateAsset } from "../../types";
  */
 export async function listOperations(
   address: string,
-  page: Pagination,
+  page?: Pagination,
```
hedi-edelbloute: Can't we just use page.minheight to 0 ? no need to put it optional everywhere and duplicate the logic that sets default value to 0

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -31,22 +35,36 @@ export function createApi(config: StellarConfig): AlpacaApi<StellarAsset, Stella
     getBalance,
     lastBlock,
     listOperations: operations,
+    validateIntent: getTransactionStatus,
+    getAccountInfo: async (address: string) => {
```
hedi-edelbloute: This needs to be removed, you can add getSequence if required

libs/ledger-live-common/src/bridge/generic-alpaca/buildSubAccounts.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    if (token && !blacklistedTokenIds.includes(token.id)) {
+      tokenAccounts.push(
+        buildStellarTokenAccount({
+          parentAccountId: accountId,
+          stellarAsset: asset,
+          token,
+          operations: operations.filter(
+            op =>
+              op.extra.assetCode === asset.asset_code &&
```
hedi-edelbloute: This is stellar specific, we probably need to move those info in main opeation type, and no related to asset code at all

libs/ledger-live-common/src/bridge/generic-alpaca/createTransaction.ts
```diff
@@ -1,27 +1,66 @@
...
+        amount: BigNumber(0),
+        recipient: "",
+        fee: null,
+        tag: undefined,
+        networkInfo: null,
+        feeCustomUnit: null, // NOTE: XRP does not use custom units for fees anymore
+      };
+    case "stellar":
+      return {
+        family: currency.family,
```
hedi-edelbloute, line 42: Can't we abstract this and just return one object for all ?

libs/ledger-live-common/src/bridge/generic-alpaca/estimateMaxSpendable.ts
```diff
@@ -6,21 +6,50 @@ import { transactionToIntent } from "./utils";
...
+      {
+        type: "PAYMENT",
+        recipient: draftTransaction.recipient,
+        amount: BigInt(draftTransaction.amount?.toString() ?? "0"),
+        fee: BigInt(draftTransaction["fees"]?.toString() ?? "0"),
+        useAllAmount: !!draftTransaction.useAllAmount,
+        subAccountId: transaction?.subAccountId || "",
+      },
+    );
+    if (network === "stellar") {
```
hedi-edelbloute, line 30: we need to remove this

libs/ledger-live-common/src/bridge/generic-alpaca/estimateMaxSpendable.ts
```diff
@@ -6,21 +6,50 @@ import { transactionToIntent } from "./utils";
...
+      useAllAmount: true,
     };
     const fees = await getAlpacaApi(network, kind).estimateFees(
       transactionToIntent(mainAccount, draftTransaction),
     );
-
+    const { freshAddress, balance, currency, pendingOperations, spendableBalance, subAccounts } =
+      account;
+    // FIXME: hardcoding type here
+    const { amount } = await getAlpacaApi(network, kind).validateIntent(
```
hedi-edelbloute, line 27: validateIntent needs to take an intent as input

libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
```diff
@@ -3,60 +3,120 @@ import { GetAccountShape, mergeOps } from "@ledgerhq/coin-framework/bridge/jsHel
...
+//         minHeight: isInitSync ? 0 : parseInt(lastPagingToken || "0", 10),
+//       };
+//   }
+// }
+
+export function genericGetAccountShape(network: string, kind: string): GetAccountShape {
+  return async (info, syncConfig) => {
+    const { address, initialAccount, currency, derivationMode } = info;
+
+    if (address === STELLAR_BURN_ADDRESS) {
```
hedi-edelbloute: To be checked if it's still necessary, if yes, it needs to be abstracted

libs/ledger-live-common/src/bridge/generic-alpaca/getTransactionStatus.ts
```diff
@@ -4,35 +4,89 @@ import { getAlpacaApi } from "./alpaca";
...
+      memoType?: string;
+      memoValue?: string;
+    },
+  ) => {
+    const { freshAddress, balance, currency, pendingOperations, subAccounts, spendableBalance } =
+      account;
+
     const alpacaApi = getAlpacaApi(network, kind);
-    const { errors, warnings } = await alpacaApi.validateIntent(
+    let transactionType = "PAYMENT"; // NOTE: assuming payment by default here, can be changed based on transaction.mode
```
hedi-edelbloute: We can see stellar specific logic here, it needs to be in coin-stellar, and validated through validateIntent that takes intent as input, not account/transaction live objects

libs/ledger-live-common/src/bridge/generic-alpaca/prepareTransaction.ts
```diff
@@ -9,18 +9,41 @@ function bnEq(a: BigNumber | null | undefined, b: BigNumber | null | undefined):
...
     if (!bnEq(transaction.fees, bnFee)) {
-      return { ...transaction, fees: bnFee };
+      return { ...transaction, fees: bnFee, assetCode, assetIssuer };
     }
 
     return transaction;
   };
 }
+
+export function getAssetCodeIssuer(
```
hedi-edelbloute: Nogo in generic alpaca

libs/ledger-live-common/src/bridge/generic-alpaca/signOperation.ts
```diff
@@ -40,16 +84,48 @@ export const genericSignOperation =
...
+          if (transaction["memoType"] && transaction["memoValue"]) {
+            const txMemoType = String(transaction["memoType"]);
+            const txMemoValue = String(transaction["memoValue"]);
+            txWithMemo.memo = {
+              type: txMemoType as "NO_MEMO" | "MEMO_TEXT" | "MEMO_ID" | "MEMO_HASH" | "MEMO_RETURN",
+              value: txMemoValue,
+            };
+          }
+          const txWithAsset = transactionIntent as TransactionIntent<any>;
+          if (transaction["assetCode"] && transaction["assetIssuer"]) {
```
hedi-edelbloute: no go in generic alpaca

libs/ledger-live-common/src/bridge/generic-alpaca/signer/types.ts
```diff
@@ -1,4 +1,5 @@
 export type SignTransactionOptions = {
   rawTxHex: string;
   path: string;
+  transaction: Buffer;
```
hedi-edelbloute, line 4: what is the difference with rawtxhex ? why do we need two types for the same thing ?

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -4,16 +4,39 @@ import {
...
+  if (op.details?.ledgerOpType !== undefined) {
+    extra.ledgerOpType = op.details.ledgerOpType as string;
+  }
+
+  if (op.details?.assetAmount !== undefined) {
+    extra.assetAmount = op.details.assetAmount as string;
+  }
+
+  if (op.asset?.type === "token") {
+    extra.assetCode = op.asset.assetCode;
```
hedi-edelbloute: Nogo here

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -4,16 +4,39 @@ import {
...
 import { fromBigNumberToBigInt } from "@ledgerhq/coin-framework/utils";
 
 export function adaptCoreOperationToLiveOperation(
   accountId: string,
-  op: CoreOperation<Asset>,
+  op: CoreOperation<Asset<TokenInfoCommon>>,
 ): Operation {
-  return {
-    id: encodeOperationId(accountId, op.tx.hash, op.type),
+  const extra: {
```
hedi-edelbloute, line 23: stellar specific

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -25,33 +48,69 @@ export function adaptCoreOperationToLiveOperation(
...
+    assetIssuer?: string;
+    assetCode?: string;
+    mode?: string;
+    fees?: BigNumber | null | undefined;
+  },
 ): TransactionIntent<any> {
-  return {
-    type: "Payment",
+  // NOTE: why Payment here and not PAYMENT like in getTransactionStatus
+  let transactionType = "Payment"; // NOTE: assuming payment by default here, can be changed based on transaction.mode
```
hedi-edelbloute, line 103: stellar specific

libs/coin-framework/src/api/types.ts
```diff
@@ -54,6 +75,11 @@ export type Transaction = {
   recipient: string;
   amount: bigint;
   fee: bigint;
+  baseReserve?: bigint; // NOTE: used for changeTrust mode in stellar
```
hedi-edelbloute: To remove

libs/coin-framework/src/api/types.ts
```diff
@@ -54,6 +75,11 @@ export type Transaction = {
...
+  networkInfo?: {
```
hedi-edelbloute, line 74: To remove

libs/coin-framework/src/api/types.ts
qperrot: all of it ? Or only `subAccount` ?

libs/coin-framework/src/api/types.ts
```diff
@@ -1,4 +1,6 @@
 import { Unit } from "@ledgerhq/types-cryptoassets";
+import { TokenAccount } from "@ledgerhq/types-live";
+import BigNumber from "bignumber.js";
```
sprohaszka-ledger: Not a big fan of adding this library in the Alpaca interface

libs/coin-framework/src/api/types.ts
sprohaszka-ledger: This information is not a blockchain info and therefore should not exists in this interface.

libs/coin-framework/src/api/types.ts
sprohaszka-ledger: Again, this info is not a blockchain info. It is a Ledger Live invention. I agree with @hedi-edelbloute .

libs/coin-framework/src/api/types.ts
```diff
@@ -101,7 +117,7 @@ export interface TypedMapMemo<KindToValueMap extends Record<string, unknown>> ex
 type MaybeMemo<MemoType extends Memo> = MemoType extends MemoNotSupported ? {} : { memo: MemoType };
 
 export type TransactionIntent<
-  AssetInfo extends Asset<TokenInfoCommon>,
+  // AssetInfo extends Asset<TokenInfoCommon>,
```
sprohaszka-ledger: We remove this comment ;)

libs/coin-framework/src/api/types.ts
```diff
@@ -137,43 +154,57 @@ export type FeeEstimation = {
 //       for now start is used as a minHeight from which we want to fetch ALL operations
 //       limit is unused for now
 //       see design document at https://ledgerhq.atlassian.net/wiki/spaces/BE/pages/5446205788/coin-modules+lama-adapter+APIs+refinements
-export type Pagination = { minHeight: number };
+export type Pagination = { minHeight: number } & { pagingToken?: string; limit?: number }; // For evm, XRP, etc. // NOTE: For Stellar
```
sprohaszka-ledger: This comment highlight a bad pattern: if properties are specific to some blockchain, they shouldn't be defined at this level.

libs/coin-framework/src/api/types.ts
```diff
@@ -137,43 +154,57 @@ export type FeeEstimation = {
...
+// NOTE: future proof export type Pagination = Record<string, unknown>;
 
 export type AccountInfo = {
   isNewAccount: boolean;
   balance: string;
   ownerCount: number;
   sequence: number;
+  assets?: AssetInfo[]; // BalanceAsset[]; // Optional, depending on the API
+  spendableBalance?: string; // Optional, depending on the API
```
sprohaszka-ledger: Why optional?

libs/coin-framework/src/api/types.ts
```diff
@@ -137,43 +154,57 @@ export type FeeEstimation = {
...
 };
 
 export type AlpacaApi<
-  AssetInfo extends Asset<TokenInfoCommon>,
+  // AssetInfo extends Asset<TokenInfoCommon>,
```
sprohaszka-ledger: You can remove this comment ;)

libs/coin-framework/src/api/types.ts
```diff
@@ -137,43 +154,57 @@ export type FeeEstimation = {
...
     customFees?: bigint,
   ) => Promise<string>;
-  getBalance: (address: string) => Promise<Balance<AssetInfo>[]>;
+  getBalance: (address: string) => Promise<Balance[]>;
   lastBlock: () => Promise<BlockInfo>;
   listOperations: (
     address: string,
-    pagination: Pagination,
-  ) => Promise<[Operation<AssetInfo>[], string]>;
+    pagination?: Pagination,
```
sprohaszka-ledger: I think it is better to keep this property as non optional and add a default value as it will always be needed in the implementation.

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -31,22 +35,36 @@ export function createApi(config: StellarConfig): AlpacaApi<StellarAsset, Stella
     getBalance,
     lastBlock,
     listOperations: operations,
+    validateIntent: getTransactionStatus,
```
sprohaszka-ledger: IMHO this is a mistake.
`getTransactionStatus` required an `Account`, which is not an Alpaca notion but Ledger Live notion. For me this is prohibitive

libs/coin-modules/coin-stellar/src/logic/getTransactionStatus.ts
sprohaszka-ledger, line 1: We should at least rename the file according to what it does in accordance with Alpaca api

libs/coin-modules/coin-stellar/src/logic/getTransactionStatus.ts
```diff
@@ -116,10 +133,12 @@ export const getTransactionStatus: AccountBridge<Transaction>["getTransactionSta
     }
 
     // Asset payment
-    if (isAssetPayment) {
-      const asset = findSubAccountById(account, transaction.subAccountId || "");
+    if (isAssetPayment && account.subAccount) {
```
sprohaszka-ledger: `subAccount` is a Live notion :(

libs/coin-modules/coin-stellar/src/logic/tokens.ts
sprohaszka-ledger, line 1: Why move this file here?
It is so LedgerLive specific :(

libs/coin-modules/coin-stellar/src/logic/utils.ts
sprohaszka-ledger, line 1: Not sure this is the correct place for this file too

libs/coin-modules/coin-stellar/src/network/horizon.ts
```diff
@@ -4,7 +4,6 @@ import { LedgerAPI4xx, LedgerAPI5xx, NetworkDown } from "@ledgerhq/errors";
 import type { CacheRes } from "@ledgerhq/live-network/cache";
 import { makeLRUCache } from "@ledgerhq/live-network/cache";
 import { log } from "@ledgerhq/logs";
-import type { Account } from "@ledgerhq/types-live";
```
sprohaszka-ledger, line 7: 👍

libs/coin-modules/coin-stellar/src/transaction.ts
sprohaszka-ledger, line 1: I disagree with this file move.
It is important to keep files in their "domain". Root folder is not one.

libs/coin-framework/src/api/types.ts
Wozacosta: Removed now, the whole Account type is not needed as all information required is located in `TransactionIntent`

https://github.com/LedgerHQ/ledger-live/blob/feat/generic-bridge-token-support-stellar/libs/coin-framework/src/api/types.ts#L144

libs/coin-framework/src/api/types.ts
Wozacosta: This whole type is removed as well 👍

libs/coin-framework/src/api/types.ts
Wozacosta: This whole type is removed as well 👍

libs/coin-modules/coin-module-boilerplate/src/common-logic/history/listOperations.ts
Wozacosta: This argument is no more optional

libs/coin-modules/coin-stellar/src/api/index.ts
Wozacosta: Removed as well, in its place a getSequence is defined: https://github.com/LedgerHQ/ledger-live/blob/feat/generic-bridge-token-support-stellar/libs/coin-framework/src/api/types.ts#L187-L187

You'll also find a   getChainSpecificRules route, 
This one is used  here https://github.com/LedgerHQ/ledger-live/blob/feat/generic-bridge-token-support-stellar/libs/ledger-live-common/src/bridge/generic-alpaca/getTransactionStatus.ts#L66-L66 and here https://github.com/LedgerHQ/ledger-live/blob/feat/generic-bridge-token-support-stellar/libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts#L38-L38 

And allows chain-specific behavior in gettransactionstatus and getaccountshape. (Which before required access to the ledger account in the coin-module)

For stellar its implem is here: https://github.com/LedgerHQ/ledger-live/blob/feat/generic-bridge-token-support-stellar/libs/coin-modules/coin-stellar/src/api/index.ts#L46-L46

libs/ledger-live-common/src/bridge/generic-alpaca/estimateMaxSpendable.ts
Wozacosta: Done 👍

libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
Wozacosta: Even though I still have a hard time figuring out how a user could end up having the burn address, it's critical enough (loss of funds happened before this check https://ledgerhq.atlassian.net/browse/LIVE-4014) that we still keep it. 
Abstracted here https://github.com/LedgerHQ/ledger-live/blob/feat/generic-bridge-token-support-stellar/libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts#L38-L38

libs/coin-framework/src/api/types.ts
Wozacosta: Gone 👍

libs/coin-framework/src/api/types.ts
Wozacosta: Gone as well

libs/coin-framework/src/api/types.ts
Wozacosta: Indeed, we removed its import and its use 👍

libs/coin-modules/coin-stellar/src/logic/getTransactionStatus.ts
Wozacosta: Indeed, no more account arg and subaccount access now 👍

libs/coin-framework/src/api/types.ts
Wozacosta: This whole type is removed now as we don't need it anymore.

libs/coin-framework/src/api/types.ts
Wozacosta: Done 👍

libs/coin-modules/coin-stellar/src/api/index.ts
Wozacosta: Adapted the arguments (validateIntent -> getTransactionStatus of xrp and stellar are now only taking in a single transactionIntent argument)

libs/coin-modules/coin-stellar/src/logic/tokens.ts
Wozacosta: It was kept there whilst ledger live specific logic was moved, will be fully removed in next commit 👍

libs/ledger-live-common/src/bridge/generic-alpaca/createTransaction.ts
```diff
@@ -1,27 +1,61 @@
...
 
-export function createTransaction(account: Account): TransactionCommon & {
+export function createTransaction(account: Account | TokenAccount): TransactionCommon & {
   family: string;
-  fee: BigNumber | null | undefined;
-  networkInfo: NetworkInfo | null | undefined;
-  tag: number | null | undefined;
-  feeCustomUnit: Unit | null | undefined;
+  fee?: BigNumber | null | undefined;
+  fees?: BigNumber | null;
```
hedi-edelbloute, line 17: Do we still need both fees / fee ?

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -40,30 +43,52 @@ export function createApi(config: StellarConfig): AlpacaApi<StellarMemo> {
...
   customFees?: FeeEstimation,
 ): Promise<string> {
-  const fees = customFees?.value ?? (await estimateFees());
-
+  let fees = transactionIntent.fees;
+  if (customFees?.value) {
+    fees = customFees?.value ?? (await estimateFees(transactionIntent.sender));
+  } else {
+    fees = transactionIntent.fees ?? (await estimateFees(transactionIntent.sender));
+  }
```
francois-guerin-ledger: Really weird to instanciate a variable and re-assign it right after. Can we have a ternary ?

```typescript
const fees = customFees?.value ? ... : ... ;
```

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -86,16 +111,25 @@ function compose(tx: string, signature: string, pubkey?: string): string {
   return combine(envelopeFromAnyXDR(tx, "base64"), signature, pubkey);
 }
 
-async function estimate(): Promise<FeeEstimation> {
-  const value = await estimateFees();
+async function estimate(transactionIntent: TransactionIntent): Promise<FeeEstimation> {
+  const value = transactionIntent?.fees
+    ? BigInt(transactionIntent?.fees.toString())
+    : await estimateFees(transactionIntent.sender);
```
francois-guerin-ledger: `TransactionIntent.fees` is already a `bigint`.
```suggestion
  const value = transactionIntent.fees ?? (await estimateFees(transactionIntent.sender));
```

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -86,16 +111,25 @@ function compose(tx: string, signature: string, pubkey?: string): string {
...
 }
 
-async function operations(
-  address: string,
-  { minHeight }: Pagination,
-): Promise<[Operation[], string]> {
-  return operationsFromHeight(address, minHeight);
+async function operations(address: string, pagination: Pagination): Promise<[Operation[], string]> {
+  const minHeight = pagination.minHeight as number | undefined;
+  const lastPagingToken = pagination.lastPagingToken as string | undefined;
```
francois-guerin-ledger:
```suggestion
  const minHeight = pagination.minHeight;
  const lastPagingToken = pagination.lastPagingToken ?? "";
```

libs/coin-modules/coin-stellar/src/logic/estimateFees.ts
```diff
@@ -1,10 +1,10 @@
...
 /**
  * Estimate the fees for one transaction
  * @see {@link https://developers.stellar.org/docs/learn/fundamentals/fees-resource-limits-metering#inclusion-fee}
  */
-export async function estimateFees(): Promise<bigint> {
-  const baseFee = await fetchBaseFee();
-  return BigInt(baseFee.recommendedFee);
+export async function estimateFees(account: string): Promise<bigint> {
+  const { fees } = await fetchAccountNetworkInfo(account);
+  return BigInt(fees.toString());
```
francois-guerin-ledger, line 9: Be really careful with `toString`. See

```js
> BigInt((new BigNumber("100000000000000000000000000")).toString())
```

```
Uncaught SyntaxError: Cannot convert 1e+26 to a BigInt
```

Always prefer `toFixed`.

```js
> BigInt((new BigNumber("100000000000000000000000000")).toFixed(0))
100000000000000000000000000n
```

libs/coin-modules/coin-stellar/src/logic/getBalance.ts
```diff
@@ -1,7 +1,40 @@
...
+import BigNumber from "bignumber.js";
 
 export async function getBalance(addr: string): Promise<Balance[]> {
-  const { balance } = await fetchAccount(addr);
-  return [{ value: BigInt(balance.toString()), asset: { type: "native" } }];
+  const { balance, assets, spendableBalance } = await fetchAccount(addr);
+  const locked = BigInt(balance.toString()) - BigInt(spendableBalance.toString());
+  const nativeRes = [
+    {
+      value: BigInt(balance.toString()),
```
francois-guerin-ledger, line 13: Same, feels risky. Also I realize we use it already

libs/coin-modules/coin-stellar/src/logic/getTransactionStatus.ts
```diff
@@ -0,0 +1,230 @@
...
+  StellarNotEnoughNativeBalanceToAddTrustline,
+  StellarMuxedAccountNotExist,
+  StellarSourceHasMultiSign,
+  StellarMemo,
+} from "../types";
+import BigNumber from "bignumber.js";
+import { getBalance } from "./getBalance";
+import { fetchAccount } from "../network/horizon";
+
+export const getTransactionStatus = async (
```
francois-guerin-ledger: Seems more accurate
```suggestion
export const validateIntent = async (
```

libs/ledger-live-common/src/bridge/generic-alpaca/buildSubAccounts.ts
```diff
@@ -0,0 +1,110 @@
+import BigNumber from "bignumber.js";
+import { emptyHistoryCache } from "@ledgerhq/coin-framework/account/index";
+import type { CryptoCurrency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
+import type { Operation, SyncConfig, TokenAccount } from "@ledgerhq/types-live";
+import { encodeOperationId } from "@ledgerhq/coin-framework/operation";
+import { listTokensForCryptoCurrency } from "@ledgerhq/cryptoassets";
+import { AssetInfo, Balance } from "@ledgerhq/coin-framework/lib/api/types";
```
francois-guerin-ledger:
```suggestion
import { AssetInfo, Balance } from "@ledgerhq/coin-framework/api/types";
```

libs/coin-modules/coin-stellar/src/network/horizon.ts
```diff
@@ -33,6 +32,7 @@ import {
   rawOperationsToOperations,
 } from "./serialization";
 import { patchHermesTypedArraysIfNeeded, unpatchHermesTypedArrays } from "../polyfill";
+import { AssetInfo } from "@ledgerhq/coin-framework/lib/api/types";
```
francois-guerin-ledger:
```suggestion
import { AssetInfo } from "@ledgerhq/coin-framework/api/types";
```

libs/coin-modules/coin-stellar/src/network/horizon.ts
```diff
@@ -189,6 +189,41 @@ export async function fetchAccount(addr: string): Promise<{
...
+            currentAsset.type !== "native" &&
+            "assetReference" in currentAsset &&
+            asset.asset_code === currentAsset?.assetReference &&
+            asset.asset_issuer === currentAsset?.assetOwner
+          );
+        })
+        .map(asset => {
+          return parseCurrencyUnit(currency.units[0], asset.selling_liabilities || "0");
+        });
+      return assetSellingLiabilities[0];
```
francois-guerin-ledger: Maybe add a security in case `assetSellingLiabilities` is empty ?

libs/coin-modules/coin-stellar/src/transaction.ts
```diff
@@ -43,13 +44,12 @@ export function formatTransaction(
...
   const { networkInfo } = tr;
-  const [assetCode, assetIssuer] = getAssetCodeIssuer(tr);
-
+  const [assetReference, assetOwner] = getAssetCodeIssuer(tr);
   return {
     ...common,
     family: tr.family,
     fees: tr.fees ? new BigNumber(tr.fees) : null,
-    baseReserve: tr.baseReserve ? new BigNumber(tr.baseReserve) : null,
+    // baseReserve: tr.baseReserve ? new BigNumber(tr.baseReserve) : null,
```
francois-guerin-ledger: Commented line should be removed.

francois-guerin-ledger (commented):
Barely started to review it, will continue

libs/coin-modules/coin-stellar/src/network/horizon.ts
qperrot: this is not use I can remove it

libs/coin-modules/coin-stellar/src/logic/estimateFees.ts
qperrot: I don't think this is even possible

hedi-edelbloute (approved):

libs/coin-framework/src/api/types.ts
```diff
@@ -11,12 +10,28 @@ export type BlockInfo = {
...
+// NOTE: from crypto-asset
+export type Unit = {
+  // display name of a given unit (example: satoshi)
+  name: string;
+  // string to use when formatting the unit. like 'BTC' or 'USD'
+  code: string;
+  // number of digits after the '.'
+  magnitude: number;
+  // should it always print all digits even if they are 0 (usually: true for fiats, false for cryptos)
+  showAllDigits?: boolean;
```
sprohaszka-ledger, line 22: This `showAllDigits` and `prefixCode` properties are useless here as it is only useful for UI client.
Or have I missed something?

libs/coin-framework/src/api/types.ts
```diff
@@ -381,10 +409,20 @@ export type AlpacaApi<MemoType extends Memo = MemoNotSupported> = {
   getRewards: (address: string, cursor?: Cursor) => Promise<Page<Reward>>;
 };
 
-export type BridgeApi = {
-  validateIntent: (account: Account, transaction: Transaction) => Promise<TransactionValidation>;
-  // TODO: make it available on alpacaApi
-  getAccountInfo: (address: string) => Promise<AccountInfo>;
+export type ChainSpecificRules = {
+  getAccountShape: (address: string) => any;
```
sprohaszka-ledger, line 413: Let's hope this `any` return is temporary...

sprohaszka-ledger (commented):
Just a few comments as I focused this quick review only on the CoinFmk.

libs/coin-framework/src/api/types.ts
qperrot: yes but we are using the `Unit` types in `export const parseCurrencyUnit = (unit: Unit, valueString: string)` and we juste wanted to use the exacte same type even if we do note use`showAllDigits` and `prefixCode` 
Also this fonction only use `magnitude` we can change that in another PR

libs/coin-modules/coin-stellar/src/logic/validateIntent.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    if (!errors.amount && amount === 0n) {
+      errors.amount = new AmountRequired();
+    }
+  }
+
+  if (await isAccountMultiSign(transactionIntent.sender)) {
+    errors.recipient = new StellarSourceHasMultiSign();
+  }
```
francois-guerin-ledger, line 213: Is it expected to use the key `recipient` after a test on the `sender` ?

qperrot: The original was like this[Stellar]( https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-stellar/src/bridge/getTransactionStatus.ts#L186)

libs/coin-framework/src/api/types.ts
qperrot: yes it is

sprohaszka-ledger (approved):
For the sake of moving on, I've just watched the API interface definition and let's say that we can go on for this PR.
We will still need to continue to improve it soon

hedi-edelbloute (approved):

libs/coin-framework/src/api/types.ts
jprudent: I also acknowledge, seems that doesn't belongs here ?
or is that the network fees at the time of the operation ?

libs/coin-framework/src/api/types.ts
```diff
@@ -260,12 +279,12 @@ export interface StringMemo<Kind extends string = "text"> extends Memo {
 }
 
 export interface MapMemo<Kind extends string, Value> extends Memo {
-  type: "map";
+  type: string;
```
jprudent, line 282: why change the type here ?

libs/coin-framework/src/api/types.ts
```diff
@@ -282,6 +301,8 @@ export type TransactionIntent<MemoType extends Memo = MemoNotSupported> = {
   expiration?: number;
   recipient: string;
   amount: bigint;
+  fees?: bigint | null | undefined; // Optional, depending on the API
+  useAllAmount?: boolean;
```
jprudent, line 305: please provide documentation to explain what this is

libs/coin-framework/src/api/types.ts
```diff
@@ -282,6 +301,8 @@ export type TransactionIntent<MemoType extends Memo = MemoNotSupported> = {
   expiration?: number;
   recipient: string;
   amount: bigint;
+  fees?: bigint | null | undefined; // Optional, depending on the API
```
jprudent, line 304: not a TS expert, but I would have simplified this as:

```suggestion
  fees?: bigint // Optional, depending on the API
```

seems like the `undefined` or `null` values doesn't bring better semantic than an absent attribute 

wdyt?

libs/coin-framework/src/api/types.ts
```diff
@@ -306,8 +327,14 @@ export type FeeEstimation = {
 //       for now start is used as a minHeight from which we want to fetch ALL operations
 //       limit is unused for now
 //       see design document at https://ledgerhq.atlassian.net/wiki/spaces/BE/pages/5446205788/coin-modules+lama-adapter+APIs+refinements
-export type Pagination = { minHeight: number };
-
+// Future-proof pagination type as suggested in the comment
+export type Pagination = {
```
jprudent, line 331: 🤔 couldn't find this data structure in https://ledgerhq.atlassian.net/wiki/spaces/BE/pages/5446205788/coin-modules+lama-adapter+APIs+refinements#Pagination

what is `pagingToken` vs `lastPagingToken`

libs/coin-framework/src/api/types.ts
```diff
@@ -323,6 +350,7 @@ export type AccountInfo = {
   ownerCount: number;
   sequence: number;
 };
+// NOTE: future proof export type Pagination = Record<string, unknown>;
```
jprudent, line 353: comment already on line 337

libs/coin-framework/src/api/types.ts
jprudent, line 304: why add the fees here ? intent fees are passed as separated datastructure.

libs/coin-framework/src/api/types.ts
```diff
@@ -306,8 +327,14 @@ export type FeeEstimation = {
 //       for now start is used as a minHeight from which we want to fetch ALL operations
 //       limit is unused for now
 //       see design document at https://ledgerhq.atlassian.net/wiki/spaces/BE/pages/5446205788/coin-modules+lama-adapter+APIs+refinements
-export type Pagination = { minHeight: number };
-
+// Future-proof pagination type as suggested in the comment
+export type Pagination = {
+  minHeight: number;
+  lastPagingToken?: string;
+  pagingToken?: string;
```
estrauser-ledger, line 334: Why do we need 2 tokens ?
`Pagination` is used by alpaca as an api input, I think it's confusing for the user, which one should he use ? Reading the rest of the PR it seems to me that `pagingToken` is just used internally and not supposed to be exposed, maybe you can extract it from `Pagination` ?

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -50,30 +53,47 @@ export function createApi(config: StellarConfig): AlpacaApi<StellarMemo> {
...
   };
 }
 
 async function craft(
   transactionIntent: TransactionIntent<StellarMemo>,
   customFees?: FeeEstimation,
 ): Promise<string> {
-  const fees = customFees?.value ?? (await estimateFees());
+  const fees =
+    customFees?.value || transactionIntent.fees || (await estimateFees(transactionIntent.sender));
```
jprudent, line 81: related to comment https://github.com/LedgerHQ/ledger-live/pull/10420/files#r2266740937

I think this is starting to be very confusing

why not use the FeeEstimation ?

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -50,30 +53,47 @@ export function createApi(config: StellarConfig): AlpacaApi<StellarMemo> {
...
-
   const tx = await craftTransaction(
     { address: transactionIntent.sender },
     {
       type: transactionIntent.type,
       recipient: transactionIntent.recipient,
       amount: transactionIntent.amount,
       fee: fees,
-      ...(transactionIntent.asset.type === "token"
+      ...(transactionIntent.asset.type !== "native" && "assetReference" in transactionIntent.asset
```
jprudent, line 96: in which case we have non native whithout assetReference ?

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -96,16 +116,25 @@ function compose(tx: string, signature: string, pubkey?: string): string {
...
+  const lastPagingToken = pagination.lastPagingToken ?? "";
+  if (minHeight) {
+    return operationsFromHeight(address, minHeight);
+  }
+  const isInitSync = lastPagingToken === "";
+  // FIXME: why bother creating limit and pagingToken here, something is off?!
+  const newPagination = isInitSync
+    ? { limit: getEnv("API_STELLAR_HORIZON_INITIAL_FETCH_MAX_OPERATIONS"), minHeight: 0 }
+    : { pagingToken: lastPagingToken, minHeight: 0 };
+  return operationsFromHeight(address, newPagination.minHeight);
```
jprudent, line 137: note that right now, you can use operation height to iterate over all operations 

this behavior is in used by lama-adapter, in production since 2 months. 

Also @estrauser-ledger is working on implementing proper pagination token. Looks like our respective teams are not synced on that.

jprudent (commented):
Here's what I checked as a lama boy:
- coin-framework API is retro compatible
- implementation of changes won't break anything already working
- the code is still aligned with our team vision

libs/coin-framework/src/api/types.ts
qperrot: NeworkInfo.fees is used on Front end in case you are manually change it `custome fees`

libs/coin-modules/coin-stellar/src/logic/estimateFees.ts
```diff
@@ -1,10 +1,10 @@
...
 
 /**
  * Estimate the fees for one transaction
  * @see {@link https://developers.stellar.org/docs/learn/fundamentals/fees-resource-limits-metering#inclusion-fee}
  */
-export async function estimateFees(): Promise<bigint> {
-  const baseFee = await fetchBaseFee();
-  return BigInt(baseFee.recommendedFee);
+export async function estimateFees(account: string): Promise<bigint> {
+  const { fees } = await fetchAccountNetworkInfo(account);
```
jprudent, line 8: looking at the impem, we are doing 1 extra call to the node to get that info now

libs/coin-modules/coin-stellar/src/api/index.ts
qperrot: Because it depends of what user do
In your case you have `customFees?.value` but on our side use can custome fees directly on LLD and this will modifie the value of `transactionIntent.fees`.
We want to call `estimateFees` only if `transactionIntent.fees`  is not set

libs/coin-modules/coin-stellar/src/api/index.ts
qperrot: This is to fix a typescript error.
If I do not check that then I will have the error
`Property 'assetReference' does not exist on type 'AssetInfo'.
  Property 'assetReference' does not exist on type '{ type: "native"; name?: string; unit?: Unit; }'.`

libs/coin-framework/src/api/types.ts
qperrot: This is set to true if user check the box `Send Max` on the front end side.

libs/coin-framework/src/api/types.ts
qperrot: Because `map` was too complex to handle for each coin, it added more complexity and just a `string` was enough here.
Also type should be `Memo_text` or `Memo_ID` for stellar and I don't think that other coin have multiples types of MEMO

libs/coin-framework/src/api/types.ts
jprudent: man, I'm unsure about all that UI concerns in the API
see also https://github.com/LedgerHQ/ledger-live/pull/10420/files#r2262293794

I don't have all your background, maybe I'm wrong but
it looks like this is a high level feature that could  be composed of a call to getBalance() followed by the crafting process ? Maybe you need to breakdown UX interactions in smaller unit / aggregation layer ?

jprudent: or maybe draining an account is a different kind of intent than a traditional tx

libs/coin-framework/src/api/types.ts
qperrot: Again this  used if user had a `custome fees`  we don't want to use estimateFees but the one customed by the user 
Also this is only for stellar right now but it will be useable by other coin

libs/coin-modules/coin-stellar/src/logic/getBalance.ts
```diff
@@ -1,7 +1,40 @@
 import { Balance } from "@ledgerhq/coin-framework/api/types";
 import { fetchAccount } from "../network";
+import { findTokenById } from "@ledgerhq/cryptoassets/tokens";
```
jprudent, line 3: do we want a dependency to cryptoasset lib ?

libs/coin-modules/coin-stellar/src/api/index.ts
```diff
@@ -96,16 +116,25 @@ function compose(tx: string, signature: string, pubkey?: string): string {
...
 
-async function operations(
-  address: string,
-  { minHeight }: Pagination,
-): Promise<[Operation[], string]> {
-  return operationsFromHeight(address, minHeight);
+async function operations(address: string, pagination: Pagination): Promise<[Operation[], string]> {
+  const minHeight = pagination.minHeight;
+  const lastPagingToken = pagination.lastPagingToken ?? "";
+  if (minHeight) {
```
jprudent, line 129: should pass the if for minHeight == 0

libs/coin-modules/coin-xrp/src/api/index.ts
```diff
@@ -157,9 +160,17 @@ async function operationsFromHeight(
   return [state.accumulator, state.apiNextCursor ?? ""];
 }
 
-async function operations(
-  address: string,
-  { minHeight }: Pagination,
-): Promise<[Operation[], string]> {
-  return await operationsFromHeight(address, minHeight);
+// NOTE: double check
```
jprudent, line 163: double check :)

jprudent (approved):
Approved, fixes will be done in subsequent PRs

hedi-edelbloute (approved):

hedi-edelbloute (approved):

jprudent (approved):

PR #10430, opened 2025-06-04
ci: Mock test sharding

lewisd5:
- Making use of Jests parallelisation features to run macOS and Android tasks in parallel
- Separating out builds to ensure a build only happens once for a test suite
- Separating out JS and native builds to ensure that only that which is required, is built
- New cache upload function to work around networking issues between macOS runners and S3

Full details of changes: https://ledgerhq.atlassian.net/wiki/x/uICZYAE

Green Run: https://github.com/LedgerHQ/ledger-live/actions/runs/15822516598

Tickets: https://ledgerhq.atlassian.net/browse/LIVE-14409
https://ledgerhq.atlassian.net/browse/LIVE-18849
https://ledgerhq.atlassian.net/browse/LIVE-18811

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -63,13 +63,6 @@ jobs:
     secrets: inherit
 
   # LLM
-  build-mobile:
-    name: "Build Mobile"
-    needs: determine-affected
-    if: ${{contains(needs.determine-affected.outputs.paths, 'ledger-live-mobile') && !github.event.pull_request.head.repo.fork}}
-    uses: LedgerHQ/ledger-live/.github/workflows/build-mobile-reusable.yml@develop
-    secrets: inherit
-
   test-mobile:
     name: "Test Mobile"
     needs: determine-affected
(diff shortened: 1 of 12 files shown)


.github/workflows/build-and-test-pr.yml
```diff
@@ -63,25 +63,26 @@ jobs:
...
-  test-mobile:
-    name: "Test Mobile"
+  # LLM
+  build-mobile:
+    name: "Build Mobile (Native)"
     needs: determine-affected
     if: ${{contains(needs.determine-affected.outputs.paths, 'ledger-live-mobile') && !github.event.pull_request.head.repo.fork}}
-    uses: LedgerHQ/ledger-live/.github/workflows/test-mobile-reusable.yml@develop
+    uses: LedgerHQ/ledger-live/.github/workflows/build-mobile-reusable.yml@develop
     secrets: inherit
```
KVNLS, line 79: [ASK] Can we drop build-mobile-reusable as it's done in test-mobile-e2e? Is there a reason to keep it?

.github/workflows/test-mobile-build-android-reusable.yml
```diff
@@ -0,0 +1,233 @@
...
+
+      - name: setup Android SDK
+        uses: android-actions/setup-android@v2.0.10
+
+        # https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/
+      - name: Enable KVM group perms
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
```
KVNLS, line 121: Maybe we should do it on the image directly, no?

.github/workflows/build-and-test-pr.yml
lewisd5: This is on the to-do. I'm currently discussing with @aboissiere-ledger how we still facilitate the native build artifact in order to remove this.

.github/workflows/test-mobile-build-android-reusable.yml
lewisd5: These are using the GHA hosted runners because timing reliability on the ARC runners was entirely unreliable (I believe due to the underlying CPU differences within the fleet - even if the same label). Even with these steps the runners here outperformed the ARC runners by a factor of over 100%. Definitely one to further tune though

KVNLS: Thanks

KVNLS (approved):
Great improvements!

PR #10535, opened 2025-06-13
fix: make sure all transactions are fetched from mirror node

md-blockydevs:
### ✅ Checklist

  - Prevents incorrect termination of transaction pagination when using the Hedera mirror node API

### 📝 Description

This change fixes a bug in the transaction pagination logic where the loop was prematurely exited when `res.data.transactions` was empty. That condition was originally added as an optimization, based on the assumption that an empty transactions array with a links.next field indicated a mirror node pagination bug.

However, this assumption was incorrect. The API documentation does not mention that the mirror node enforces a 60-day implicit limit on result ranges - even when a custom timestamp parameter is provided.

Due to this behavior, responses may contain an empty transactions array while still including a valid `links.next`, which can cause the loop to exit too early and miss data. Here is an example of account.id that won't see any operations in LL: https://mainnet-public.mirrornode.hedera.com/api/v1/transactions?account.id=0.0.751515&order=desc&limit=100

This behavior is undocumented in the API itself. I found it in the following Hedera blog post:
https://hedera.com/blog/changes-to-the-hedera-operated-mirror-node

The fix removes the early break and adds an explanatory comment to prevent future confusion.

---

### DIFF
--- a/.changeset/chilly-olives-give.md
+++ b/.changeset/chilly-olives-give.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-hedera": patch
+---
+
+prevent breaking loop when mirror node returns empty transactions with links.next
--- a/libs/coin-modules/coin-hedera/src/api/mirror.test.ts
+++ b/libs/coin-modules/coin-hedera/src/api/mirror.test.ts
@@ -19,7 +19,7 @@ describe("getAccountTransactions", () => {
     jest.clearAllMocks();
   });
 
-  it("should include 'account.id', 'limit=100' and 'order=desc' query params", async () => {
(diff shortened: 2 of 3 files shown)


may01 (dismissed):

hedi-edelbloute (dismissed):

.changeset/chilly-olives-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/coin-hedera": patch
```
Justkant: We usually use `minor` now instead of `patch` for everything in LL and reserve `patch` for hotfixes

md-blockydevs: Sure, rebased

PR #10567, opened 2025-06-16
feat: back navigation in earn app deposit flows

beths-ledger:
### ✅ Checklist

  - Entrypoints for depositing USDT, USDC, USDS & Dai stablecoins to earn yield via the earn live app.

### 📝 Description

1. Allow user to open the earn dashboard without the bottom nav when entering via the "deposit" / "withdraw" stablecoin yield entry points
2. Closing the earn dashboard from any entry point redirects to where the user originated

back - iPhone 16 Pro - 2025-06-16 at 17 57 07

### ❓ Context

- **JIRA or GitHub link**: [LIVE-19179](https://ledgerhq.atlassian.net/browse/LIVE-19179)
- [LIVE-19506](https://ledgerhq.atlassian.net/browse/LIVE-19506)

---

[LIVE-19179]: https://ledgerhq.atlassian.net/browse/LIVE-19179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/screens/PTX/Earn/EarnMenuDrawer.tsx
+++ b/apps/ledger-live-mobile/src/screens/PTX/Earn/EarnMenuDrawer.tsx
@@ -1,5 +1,7 @@
 import { Button, Flex, Text } from "@ledgerhq/native-ui";
 import { Theme } from "@ledgerhq/native-ui/lib/styles/theme";
+import { useRoute } from "@react-navigation/native";
+import { StackNavigationProp } from "@react-navigation/stack";
 import React, { useCallback, useEffect, useState } from "react";
 import { Linking } from "react-native";
 import { useDispatch, useSelector } from "react-redux";
@@ -7,13 +9,23 @@ import styled from "styled-components/native";
 import { makeSetEarnMenuModalAction } from "~/actions/earn";
 import { track } from "~/analytics";
 import QueuedDrawer from "~/components/QueuedDrawer";
+import { NavigatorName, ScreenName } from "~/const";
 import { earnMenuModalSelector } from "~/reducers/earn";
+
+function isValidIntent(intent?: string): intent is "deposit" | "withdraw" {
+  return ["deposit", "withdraw"].includes(intent ?? "");
+}
 
 /** TODO Should be a shared constant throughout the app for all events */
 const BUTTON_CLICKED_TRACK_EVENT = "button_clicked";
 
-export function EarnMenuDrawer() {
+export function EarnMenuDrawer({
+  navigation,
+}: {
+  navigation: StackNavigationProp<{ [key: string]: object | undefined }>;
+}) {
   const dispatch = useDispatch();
+  const route = useRoute();
   const [modalOpened, setModalOpened] = useState(false);
 
   const openModal = useCallback(() => setModalOpened(true), []);
@@ -39,14 +51,42 @@ export function EarnMenuDrawer() {
           </Text>
         ) : null}
         <Flex rowGap={16}>
-          {modal?.options.map(({ label, metadata: { link, ...tracked } }) =>
+          {modal?.options.map(({ label, metadata: { link, live_app, ...tracked } }) =>
             link ? (
               <OptionButton
                 key={label}
                 onPress={async () => {
-                  await track(BUTTON_CLICKED_TRACK_EVENT, tracked);
+                  await track(BUTTON_CLICKED_TRACK_EVENT, { live_app, ...tracked });
                   closeDrawer();
-                  await Linking.openURL(link);
+                  if (live_app === "earn" || live_app === "earn-stg") {
+                    const pathSegments = link.split("?");
+                    const earnSearchParams = new URLSearchParams(pathSegments.pop());
+                    const intent = earnSearchParams.get("intent") ?? undefined;
+                    const accountId = earnSearchParams.get("accountId");
+                    const earnParams = Object.fromEntries(earnSearchParams.entries());
+
+                    if (!isValidIntent(intent)) {
+                      console.warn(
+                        `Invalid earn flow intent: ${intent}. Expected "deposit" or "withdraw".`,
+                      );
+                    }
+                    // Use the base navigator to allow back navigation and hide the main navigation bar from the bottom
+                    navigation.navigate(NavigatorName.Base, {
+                      screen: NavigatorName.Earn,
+                      params: {
+                        screen: ScreenName.Earn,
+                        ...route.params,
+                        platform: "earn",
+                        params: {
+                          ...earnParams,
+                          intent: isValidIntent(intent) ? intent : undefined,
+                          accountId: accountId,
+                        },
+                      },
+                    });
+                  } else {
+                    await Linking.openURL(link);
+                  }
                 }}
               >
                 {label}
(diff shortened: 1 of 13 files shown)


apps/ledger-live-mobile/src/screens/PTX/Earn/EarnWebview/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        handleHardwareBackPress,
+      );
+
+      return () => {
+        subscription.remove();
+      };
+    }
+  }, [handleHardwareBackPress]);
+
+  useEffect(() => {
```
philipptpunkt, line 68: Just for my understanding: Here you're intercepting the navigation to use the `webview.goBack` instead of the native back navigation? Basically what I'm doing for Swap when `setWebviewState` is called?

beths-ledger: Yes. setWebviewState is called when we push a new query param to the webviewState.url. Then canGoBack is updated. Then this handler overrides the native back nav, unless the user is on the initial page

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnMenuDrawer.tsx
```diff
@@ -1,19 +1,31 @@
...
 import styled from "styled-components/native";
 import { makeSetEarnMenuModalAction } from "~/actions/earn";
 import { track } from "~/analytics";
 import QueuedDrawer from "~/components/QueuedDrawer";
+import { NavigatorName, ScreenName } from "~/const";
 import { earnMenuModalSelector } from "~/reducers/earn";
 
+function isValidIntent(intent?: string): intent is "deposit" | "withdraw" {
+  return ["deposit", "withdraw"].includes(intent ?? "");
+}
```
chrisduma-ledger, line 17: What is the value in saying `intent is "deposit" | "withdraw"` 

vs just returning `boolean`

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnWebview/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (Platform.OS === "android") {
+      const subscription = BackHandler.addEventListener(
+        "hardwareBackPress",
+        handleHardwareBackPress,
+      );
+
+      const url = safeUrl(webviewState.url);
+      const isInitialWebviewStep = url?.searchParams.get("view") === "amount";
+      const isFinalWebviewStep = url?.searchParams.get("view") === "confirmation";
```
chrisduma-ledger, line 56: This code is duplicated. You can extract it into a separate function.

chrisduma-ledger: The strings also can be extracted in an object:

```
const WEBVIEW_STEPS = {
  AMOUNT: 'amount',
  CONFIRMATION: 'confirmation'
} as const;
```

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnWebview/useEarnWebviewCustomHandlers.ts
```diff
@@ -0,0 +1,48 @@
...
+          "custom.exchange.start": () => {},
+          "custom.exchange.complete": () => {},
+          "custom.exchange.error": () => {},
+          "custom.isReady": async () => {
+            if (Config.DETOX) {
+              sendEarnLiveAppReady();
+            }
+          },
+        },
+      })["custom.isReady"],
```
sergiubreban, line 45: Why are we creating an object here and then using only the custom.isReady value? _(I'm not sure I understand the TODO)_

apps/ledger-live-mobile/src/screens/PTX/Earn/index.tsx
```diff
@@ -38,9 +38,14 @@ function Earn({ route }: Props) {
...
-    ? new URL("ledgerlive://" + route.path).searchParams
-    : new URLSearchParams();
+  const searchParams = useMemo(
+    () => (route.path ? new URL("ledgerlive://" + route.path).searchParams : new URLSearchParams()),
+    [route.path],
+  );
+  const hideMainNavigator = useMemo(
+    () => ["deposit", "withdraw"].includes(params?.intent ?? ""),
+    [params],
+  );
```
chrisduma-ledger, line 48: I don't think this is not worth memoing. These are cheap operations.

chrisduma-ledger (dismissed):

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnMenuDrawer.tsx
beths-ledger: it's for TypeScript, the navigator only accepts `"deposit" | "withdraw"` type, and TS cannot infer from just the logic itself that we have narrowed the input to that type.

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnWebview/index.tsx
beths-ledger: this is duplication of the below, but a helper fn wouldn't be much cleaner and risks being less sure that all vars are created inside the scope of each useEffect

apps/ledger-live-mobile/src/screens/PTX/Earn/index.tsx
beths-ledger: Indeed. But it needs referential identity if it's going to be passed as a prop

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnMenuDrawer.tsx
chrisduma-ledger: Yes, but can't we just pass an `enum` to rather than `string` to the param and then it's all fine?

apps/ledger-live-mobile/src/screens/PTX/Earn/index.tsx
chrisduma-ledger: Are `searchParams` passed as a prop to a memoized component or is it used in the dependency array of other hooks or used in callbacks that needs a stable reference @beths-ledger ?

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnWebview/index.tsx
chrisduma-ledger: I do not agree @beths-ledger 
These are pure calculations, there are no side effects or closures involved, also the variables are local, they're not escaping the `useEffect` hook. A helper function would be equally safe in terms of scope.

apps/ledger-live-mobile/src/screens/PTX/Earn/EarnWebview/useEarnWebviewCustomHandlers.ts
beths-ledger: it's an interim solution -- this is a clone of the Exchange player but we don't have time to create a custom earn one right now.

PR #10694, opened 2025-06-26
:sparkles: (MAD) incorrect account name assignment

LucasWerey:
### ✅ Checklist

  - account selection

### 📝 Description

Account Names were sometimes wrongly assigned as we had dependencies on two different sorts. I've fixed that by moving the logic inside the hook and before the sort. Then we override the default accountName of account object 

### ❓ Context

- **JIRA or GitHub link**: [LIVE-19899](https://ledgerhq.atlassian.net/browse/LIVE-19899)

---


apps/ledger-live-desktop/src/newArch/features/ModularDrawer/hooks/useDetailedAccounts.tsx
```diff
@@ -52,11 +53,23 @@ export const useDetailedAccounts = (
 
   const accounts = useMemo(() => {
     const accountTuples = getAccountTuplesForCurrency(asset, nestedAccounts, accountIds);
-    return accountTuples.sort((a, b) => sortAccountsByBalance(a.account, b.account));
+    return orderBy(accountTuples, [(tuple: AccountTuple) => tuple.account.balance], ["desc"]);
   }, [asset, nestedAccounts, accountIds]);
 
+  const overridedAccountName = useBatchMaybeAccountName(accounts.map(({ account }) => account));
```
ejaycoleman, line 59: thinking out loud here but I'm wondering how feasible it would be to instead use `getAccountName` that's used by `useBatchMaybeAccountName`, and call it directly within the accounts map below instead 
it sounds a bit complex but it just doesn't feel _right_ to create this array and then reference it by index

ejaycoleman (approved):

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/hooks/useDetailedAccounts.tsx
LucasWerey: I agree that's the way we found the accountName is a bit complex rn. But I think if we want to properly do it we should mix both wallet.accountNames, and accounts objects. But rn it needs to much stored data to do it easily from here

PR #10765, opened 2025-07-01
Feat/deprecate nanoS

Antoine-bls83:
### ✅ Checklist

  - ...

### 📝 Description

This PR introduces support for dynamically configuring Ledger Nano S (LNS) deprecation and blind signing warnings via Firebase Remote Config.

✅ What’s implemented:
A new config entry (as an example) is have to be added in firebase
Enables configuration of:
 - Info screens (deprecation warnings shown before deprecation date).
 - Error screens (feature blocked after deprecation date).
 - Clear signing Warning Screens for unsupported Clear Signing  on LNS.

Example of config:

```
config_nanoapp_tron: {
  type: "object",
  default: {
  "deviceDeprecated": [
      {
        "deviceModelId": "flex",
        "infoScreen": {
          "date": "2025-01-31",
          "deprecatedFlow": [
            "receive"
          ]
        },
        "errorScreen": {
          "date": "2025-10-31",
          "deprecatedFlow": [
            "receive"
          ]
        },
        "warningClearSigningScreen": {
          "date": "2025-01-01",
          "exception": [
            "WINK",
            "Tron"
          ],
          "deprecatedFlow": [
            "receive"
          ]
        }
      }
    ]
  }
}
```

## Warning screen:

<p>
Screenshot 2025-08-05 at 09 07 45
Screenshot 2025-08-05 at 09 08 11
Screenshot 2025-08-05 at 11 51 23

</p>

## Warning screen Clear Signing:
<p>
  Screenshot 2025-08-05 at 09 09 48
  Screenshot 2025-08-05 at 09 09 58
  Screenshot 2025-08-05 at 09 34 20
</p>

## Error screen:

<p>
  Screenshot 2025-08-05 at 09 12 08
  Screenshot 2025-08-05 at 09 15 14
  Screenshot 2025-08-05 at 09 48 16
</p>

### ❓ Context

- **JIRA or GitHub link**:  [LIVE-19339](https://ledgerhq.atlassian.net/browse/LIVE-19339) 

---

[LIVE-19339]: https://ledgerhq.atlassian.net/browse/LIVE-19339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/actions/settings.ts
+++ b/apps/ledger-live-desktop/src/renderer/actions/settings.ts
@@ -231,6 +231,11 @@ export const hideOrdinalsAsset = (inscriptionId: string) => ({
   payload: inscriptionId,
 });
 
+export const deprecateWarningReminder = (coinName: string) => ({
+  type: "DEPRECATION_DO_NOT_REMIND",
+  payload: coinName,
+});
+
 export const setLastSeenCustomImage = (lastSeenCustomImage: {
   imageSize: number;
   imageHash: string;
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
@@ -1,0 +1,218 @@
+import React, { useCallback, useState } from "react";
+import { Trans } from "react-i18next";
+import { useDispatch } from "react-redux";
+import styled from "styled-components";
+
+import CheckBox from "~/renderer/components/CheckBox";
+import Box from "~/renderer/components/Box";
+import { Button, Flex, Icons, IconsLegacy, Link, Text } from "@ledgerhq/react-ui";
+import { openURL } from "~/renderer/linking";
+import { rgba } from "~/renderer/styles/helpers";
+import { deprecateWarningReminder } from "~/renderer/actions/settings";
+import { useDateFormatted } from "~/renderer/hooks/useDateFormatter";
+import { urls } from "~/config/urls";
+import { useLocalizedUrl } from "~/renderer/hooks/useLocalizedUrls";
+
+export const Wrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  align-items: center;
+  justify-content: center;
+  min-height: 260px;
+  max-width: 100%;
+  margin: auto ${p => p.theme.space[5]}px;
+`;
+
+const CircleIcon = styled.div`
+  padding: 16px;
+  border-radius: 100px;
+  background-color: ${p => rgba(p.theme.colors.palette.primary.c80, 0.1)};
+  align-items: center;
+  gap: 10px;
+  display: flex;
+  align-self: center;
+  justify-content: center;
+  width: fit-content;
+  height: fit-content;
+`;
+
+export const enum DeviceDeprecationScreens {
+  clearSigningScreen,
+  warningScreen,
+  errorScreen,
+}
+
+export const DeviceDeprecationScreen = ({
+  coinName,
+  date,
+  onContinue,
+  productName,
+  screenName,
+  displayClearSigningWarning = false,
+}: {
+  coinName: string;
+  date?: Date;
+  onContinue?: () => void;
+  productName: string;
+  screenName: DeviceDeprecationScreens;
+  displayClearSigningWarning?: boolean;
+}) => {
+  const dispatch = useDispatch();
+  const [isChecked, setIsChecked] = useState(false);
+  const dateFormatted = useDateFormatted(date);
+  const toggleCheck = useCallback(() => {
+    setIsChecked(prev => !prev);
+  }, []);
+  const shopUrl = useLocalizedUrl(urls.deviceDeprecation.shop);
+  const learnMoreUrl = useLocalizedUrl(urls.deviceDeprecation.learnMore);
+  const [currentScreenName, setCurrentScreenName] = useState(screenName);
+  const handleUgradeClick = () => {
+    openURL(shopUrl);
+  };
+  const handleLearnMoreClick = () => {
+    openURL(learnMoreUrl);
+  };
+  const handleContinue = () => {
+    if (isChecked) {
+      dispatch(deprecateWarningReminder(coinName));
+    }
+    if (
+      displayClearSigningWarning &&
+      currentScreenName === DeviceDeprecationScreens.warningScreen
+    ) {
+      setCurrentScreenName(DeviceDeprecationScreens.clearSigningScreen);
+    } else if (onContinue) {
+      onContinue();
+    }
+  };
+
+  const mappedIcons = {
+    [DeviceDeprecationScreens.clearSigningScreen]: (
+      <Icons.WarningFill size="L" color="palette.warning.c70" />
+    ),
+    [DeviceDeprecationScreens.warningScreen]: (
+      <Icons.InformationFill size="L" color="primary.c80" />
+    ),
+    [DeviceDeprecationScreens.errorScreen]: (
+      <IconsLegacy.CircledCrossSolidMedium size={40} color="error.c60" />
+    ),
+  };
+  const mappedTitles = {
+    [DeviceDeprecationScreens.clearSigningScreen]: (
+      <Trans
+        i18nKey={`deviceDeprecation.warning.title`}
+        values={{
+          device: productName,
+        }}
+      />
+    ),
+    [DeviceDeprecationScreens.warningScreen]: (
+      <Trans
+        i18nKey={`deviceDeprecation.info.title`}
+        values={{
+          device: productName,
+          date: dateFormatted,
+          coinName: coinName,
+        }}
+      />
+    ),
+    [DeviceDeprecationScreens.errorScreen]: (
+      <Trans
+        i18nKey={`deviceDeprecation.error.title`}
+        values={{
+          device: productName,
+        }}
+      />
+    ),
+  };
+  const mappedSubTitleKeys = {
+    [DeviceDeprecationScreens.clearSigningScreen]: "deviceDeprecation.warning.subtitle",
+    [DeviceDeprecationScreens.warningScreen]: "deviceDeprecation.info.subtitle",
+    [DeviceDeprecationScreens.errorScreen]: "deviceDeprecation.error.subtitle",
+  };
+  return (
+    <Wrapper>
+      <Flex
+        mx={20}
+        rowGap={5}
+        style={{
+          flexDirection: "column",
+        }}
+      >
+        <Box
+          flow={3}
+          style={{
+            width: "100%",
+          }}
+        >
+          <CircleIcon>{mappedIcons[currentScreenName]}</CircleIcon>
+          <Text
+            fontWeight="600"
+            data-testid="warning-deprecation-title"
+            textAlign="center"
+            fontSize={24}
+          >
+            {mappedTitles[currentScreenName]}
+          </Text>
+          <Text color="palette.text.shade60" textAlign="center" fontWeight="500" fontSize={14}>
+            <Trans i18nKey={mappedSubTitleKeys[currentScreenName]} />
+          </Text>
+        </Box>
+        <Box
+          flow={4}
+          style={{
+            width: "100%",
+          }}
+        >
+          <Button variant="main" size="large" onClick={handleUgradeClick}>
+            <Trans i18nKey={`deviceDeprecation.update`} />
+          </Button>
+          {currentScreenName !== DeviceDeprecationScreens.errorScreen ? (
+            <Button
+              size="large"
+              variant="shade"
+              data-testid="warning-deprecation-continue"
+              onClick={handleContinue}
+            >
+              <Trans
+                i18nKey={
+                  currentScreenName === DeviceDeprecationScreens.clearSigningScreen
+                    ? `deviceDeprecation.warning.continue`
+                    : `deviceDeprecation.continue`
+                }
+              />
+            </Button>
+          ) : null}
+        </Box>
+        <Link color="palette.text.shade60" onClick={handleLearnMoreClick}>
+          <Trans i18nKey={`deviceDeprecation.learnMore`} />
+        </Link>
+        {currentScreenName === DeviceDeprecationScreens.warningScreen ? (
+          <Box
+            horizontal
+            alignItems="flex-start"
+            onClick={toggleCheck}
+            style={{
+              flex: 1,
+              cursor: "pointer",
+            }}
+          >
+            <CheckBox isChecked={isChecked} data-testid="dismiss-disclaimer" />
+            <Text
+              ff="Inter|SemiBold"
+              fontSize={4}
+              style={{
+                marginLeft: 8,
+                overflowWrap: "break-word",
+                flex: 1,
+              }}
+            >
+              <Trans i18nKey={`deviceDeprecation.info.reminder`} />
+            </Text>
+          </Box>
+        ) : null}
+      </Flex>
+    </Wrapper>
+  );
+};
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
@@ -1,0 +1,256 @@
+import React from "react";
+import { render, screen, fireEvent } from "tests/testSetup";
+import DeviceAction from "./index";
+import { HOOKS_TRACKING_LOCATIONS } from "~/renderer/analytics/hooks/variables";
+import { DeviceDeprecationRules } from "@ledgerhq/live-common/hw/connectApp";
+import { DeviceModelId } from "@ledgerhq/types-devices";
+import { DmkError } from "@ledgerhq/live-dmk-desktop";
+import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets/currencies";
+import { genAccount } from "@ledgerhq/coin-framework/lib/mocks/account";
+
+const ethereumCurrency = getCryptoCurrencyById("ethereum");
+const ETH_ACCOUNT = genAccount("ethereum-1", {
+  currency: ethereumCurrency,
+});
+
+describe("DeviceAction - Deprecation", () => {
+  it("should render deprecation warning and deprecation clear signing warning", () => {
+    const mockOnContinue = jest.fn();
+    const deviceDeprecationRules: DeviceDeprecationRules = {
+      modelId: DeviceModelId.nanoS,
+      warningScreenVisible: true,
+      clearSigningScreenVisible: true,
+      warningScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      errorScreenVisible: false,
+      errorScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      clearSigningScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      date: new Date("2023-12-31"),
+      onContinue: mockOnContinue,
+    };
+
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account: ETH_ACCOUNT }}
+        location={HOOKS_TRACKING_LOCATIONS.receiveModal}
+      />,
+    );
+
+    expect(
+      screen.getByText(
+        "As of 12/31/2023, Ethereum support on Ledger Nano S™ via Ledger Live™ will end",
+      ),
+    ).toBeVisible();
+    expect(screen.getByText("Continue")).toBeVisible();
+
+    const continueButton = screen.getByTestId("warning-deprecation-continue");
+    fireEvent.click(continueButton);
+    expect(screen.getByText("Ledger Nano S™ can not Clear Sign this transaction")).toBeVisible();
+    expect(screen.getByText("Proceed at your own risk")).toBeVisible();
+
+    fireEvent.click(continueButton);
+    expect(mockOnContinue).toHaveBeenCalled();
+  });
+
+  it("should render deprecation warning only", () => {
+    const mockOnContinue = jest.fn();
+    const deviceDeprecationRules: DeviceDeprecationRules = {
+      modelId: DeviceModelId.nanoS,
+      warningScreenVisible: true,
+      clearSigningScreenVisible: false,
+      warningScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      errorScreenVisible: false,
+      errorScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      clearSigningScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      date: new Date("2023-12-31"),
+      onContinue: mockOnContinue,
+    };
+
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account: ETH_ACCOUNT }}
+        location={HOOKS_TRACKING_LOCATIONS.receiveModal}
+      />,
+    );
+
+    expect(
+      screen.getByText(
+        "As of 12/31/2023, Ethereum support on Ledger Nano S™ via Ledger Live™ will end",
+      ),
+    ).toBeVisible();
+    expect(screen.getByText("Continue")).toBeVisible();
+
+    const continueButton = screen.getByTestId("warning-deprecation-continue");
+    fireEvent.click(continueButton);
+    expect(mockOnContinue).toHaveBeenCalled();
+  });
+
+  it("should render deprecation clear signing warning only", () => {
+    const mockOnContinue = jest.fn();
+    const deviceDeprecationRules: DeviceDeprecationRules = {
+      modelId: DeviceModelId.nanoS,
+      warningScreenVisible: false,
+      clearSigningScreenVisible: true,
+      warningScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      errorScreenVisible: false,
+      errorScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      clearSigningScreenRules: {
+        exception: [],
+        deprecatedFlow: ["send", "receive"],
+      },
+      date: new Date("2023-12-31"),
+      onContinue: mockOnContinue,
+    };
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account: ETH_ACCOUNT }}
+        location={HOOKS_TRACKING_LOCATIONS.receiveModal}
+      />,
+    );
+
+    expect(screen.getByText("Ledger Nano S™ can not Clear Sign this transaction")).toBeVisible();
+    expect(screen.getByText("Proceed at your own risk")).toBeVisible();
+
+    const continueButton = screen.getByTestId("warning-deprecation-continue");
+    fireEvent.click(continueButton);
+    expect(mockOnContinue).toHaveBeenCalled();
+  });
+
+  it("should not render deprecation warning if deprecate is false", () => {
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules: null,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account: ETH_ACCOUNT }}
+        location={HOOKS_TRACKING_LOCATIONS.sendModal}
+      />,
+    );
+
+    expect(screen.queryByText("Ledger Nano S™")).toBeNull();
+  });
+
+  it("should call onContinue immediately if skipDeprecation is true", () => {
+    const mockOnContinue = jest.fn();
+    const deviceDeprecationRules: DeviceDeprecationRules = {
+      modelId: DeviceModelId.nanoS,
+      date: new Date("2023-12-31"),
+      warningScreenVisible: false,
+      clearSigningScreenVisible: false,
+      errorScreenVisible: false,
+      warningScreenRules: {
+        deprecatedFlow: ["send"],
+        exception: [],
+      },
+      errorScreenRules: {
+        deprecatedFlow: ["send"],
+        exception: [],
+      },
+      clearSigningScreenRules: {
+        deprecatedFlow: ["send"],
+        exception: [],
+      },
+      onContinue: mockOnContinue,
+    };
+
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules,
+        displayDeprecateWarning: false,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account: ETH_ACCOUNT, tokenCurrency: { name: "someToken" } }}
+        location={HOOKS_TRACKING_LOCATIONS.sendModal}
+      />,
+    );
+
+    expect(mockOnContinue).toHaveBeenCalled();
+  });
+
+  it("should display Error Screen", () => {
+    const device: Device = {
+      modelId: DeviceModelId.nanoS,
+      deviceId: "nanoS",
+      wired: true,
+    };
+    const err: DmkError = {
+      message: "device-deprecation",
+      _tag: "DeviceDeprecationError",
+    };
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules: null,
+        error: err,
+        displayDeprecateWarning: false,
+        device: device,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account: ETH_ACCOUNT }}
+        location={HOOKS_TRACKING_LOCATIONS.sendModal}
+      />,
+    );
+    expect(screen.getByText("Ledger Nano S™ does not support this feature")).toBeVisible();
+  });
+});
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
@@ -1,4 +1,4 @@
-import React, { Component, useEffect } from "react";
+import React, { Component, useEffect, useState } from "react";
 import BigNumber from "bignumber.js";
 import { Trans, useTranslation } from "react-i18next";
 import { useDispatch, useSelector } from "react-redux";
@@ -22,7 +22,7 @@ import {
   storeSelector as settingsSelector,
   trackingEnabledSelector,
 } from "~/renderer/reducers/settings";
-import { DeviceModelId } from "@ledgerhq/devices";
+import { DeviceModelId, getDeviceModel } from "@ledgerhq/devices";
 import AutoRepair from "~/renderer/components/AutoRepair";
 import TransactionConfirm from "~/renderer/components/TransactionConfirm";
 import SignMessageConfirm from "~/renderer/components/SignMessageConfirm";
@@ -73,11 +73,17 @@ import {
   InitSwapResult,
 } from "@ledgerhq/live-common/exchange/swap/types";
 import { Transaction, TransactionStatus } from "@ledgerhq/live-common/generated/types";
-import { AppAndVersion } from "@ledgerhq/live-common/hw/connectApp";
+import { AppAndVersion, DeviceDeprecationRules } from "@ledgerhq/live-common/hw/connectApp";
 import { Device } from "@ledgerhq/types-devices";
 import { LedgerErrorConstructor } from "@ledgerhq/errors/lib/helpers";
 import { TokenCurrency } from "@ledgerhq/types-cryptoassets";
-import { getNoSuchAppProviderLearnMoreMetadataPerApp, isDeviceNotOnboardedError } from "./utils";
+import {
+  FlowName,
+  getCurrencyName,
+  getFlowName,
+  getNoSuchAppProviderLearnMoreMetadataPerApp,
+  isDeviceNotOnboardedError,
+} from "./utils";
 import { useKeepScreenAwake } from "~/renderer/hooks/useKeepScreenAwake";
 import { walletSelector } from "~/renderer/reducers/wallet";
 import { useTrackManagerSectionEvents } from "~/renderer/analytics/hooks/useTrackManagerSectionEvents";
@@ -90,6 +96,10 @@ import { useTrackSyncFlow } from "~/renderer/analytics/hooks/useTrackSyncFlow";
 import { useTrackGenericDAppTransactionSend } from "~/renderer/analytics/hooks/useTrackGenericDAppTransactionSend";
 import { useTrackTransactionChecksFlow } from "~/renderer/analytics/hooks/useTrackTransactionChecksFlow";
 import { useTrackDmkErrorsEvents } from "~/renderer/analytics/hooks/useTrackDmkErrorsEvents";
+import {
+  DeviceDeprecationScreen,
+  DeviceDeprecationScreens,
+} from "./Screen/DeviceDeprecationScreen";
 
 export type LedgerError = InstanceType<LedgerErrorConstructor<{ [key: string]: unknown }>>;
 
@@ -157,6 +167,7 @@ type States = PartialNullable<{
   manifestId: string;
   transactionChecksOptInTriggered: boolean;
   transactionChecksOptIn: boolean;
+  deviceDeprecationRules?: DeviceDeprecationRules;
 }>;
 
 type InnerProps<P> = {
@@ -244,6 +255,7 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
     signMessageRequested,
     manifestId,
     manifestName,
+    deviceDeprecationRules,
   } = hookState;
 
   const dispatch = useDispatch();
@@ -252,6 +264,7 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
   const stateSettings = useSelector(settingsSelector);
   const walletState = useSelector(walletSelector);
   const isTrackingEnabled = useSelector(trackingEnabledSelector);
+  const [hasDisplayDeprecateWarning, setHasDisplayDeprecateWarning] = useState(false);
 
   useTrackManagerSectionEvents({
     location: location === HOOKS_TRACKING_LOCATIONS.managerDashboard ? location : undefined,
@@ -338,7 +351,6 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
   useTrackDmkErrorsEvents({ error });
 
   const type = useTheme().colors.palette.type;
-
   const modelId = device ? device.modelId : overridesPreferredDeviceModel || preferredDeviceModel;
 
   useEffect(() => {
@@ -366,6 +378,71 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
       dispatch(addNewDeviceModel({ deviceModelId: lastSeenDevice.modelId }));
     }
   }, [dispatch, device, deviceInfo, latestFirmware]);
+
+  const currencyName: string = getCurrencyName(request);
+  if (deviceDeprecationRules && request) {
+    const { clearSigningScreenRules, warningScreenRules, errorScreenRules, date, modelId } =
+      deviceDeprecationRules;
+
+    const currentFlow = getFlowName(location, request);
+    const skippedDeprecations = stateSettings.deprecationDoNotRemind;
+    const doSkipDeprecation = (exception: string[], deprecatedFlow: string[]) => {
+      return (
+        exception.includes(currencyName) ||
+        !deprecatedFlow.includes(currentFlow) ||
+        currentFlow === FlowName.unknown
+      );
+    };
+    const alreadyDismissed = skippedDeprecations.includes(currencyName);
+    const skipWarningScreen =
+      doSkipDeprecation(
+        warningScreenRules?.exception || [],
+        warningScreenRules?.deprecatedFlow || [],
+      ) ||
+      alreadyDismissed ||
+      hasDisplayDeprecateWarning;
+    const skipClearSigningScreen =
+      doSkipDeprecation(
+        clearSigningScreenRules?.exception || [],
+        clearSigningScreenRules?.deprecatedFlow || [],
+      ) || hasDisplayDeprecateWarning;
+    const skipErrorScreen = doSkipDeprecation(
+      errorScreenRules?.exception || [],
+      errorScreenRules?.deprecatedFlow || [],
+    );
+    const handleContinue = () => {
+      setHasDisplayDeprecateWarning(true);
+      deviceDeprecationRules.onContinue(false);
+    };
+
+    if (deviceDeprecationRules.errorScreenVisible) {
+      deviceDeprecationRules.onContinue(!skipErrorScreen);
+    } else if (deviceDeprecationRules.warningScreenVisible && !skipWarningScreen) {
+      return (
+        <DeviceDeprecationScreen
+          coinName={currencyName}
+          date={date}
+          onContinue={handleContinue}
+          productName={getDeviceModel(modelId).productName}
+          displayClearSigningWarning={
+            deviceDeprecationRules.clearSigningScreenVisible && !skipClearSigningScreen
+          }
+          screenName={DeviceDeprecationScreens.warningScreen}
+        />
+      );
+    } else if (deviceDeprecationRules.clearSigningScreenVisible && !skipClearSigningScreen) {
+      return (
+        <DeviceDeprecationScreen
+          onContinue={handleContinue}
+          productName={getDeviceModel(modelId).productName}
+          screenName={DeviceDeprecationScreens.clearSigningScreen}
+          coinName={currencyName}
+          date={date}
+        />
+      );
+    }
+    deviceDeprecationRules.onContinue(false);
+  }
 
   if (displayUpgradeWarning && appAndVersion && passWarning) {
     return renderWarningOutdated({ appName: appAndVersion.name, passWarning });
@@ -607,7 +684,6 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
     if ((error as unknown) instanceof UserRefusedDeviceNameChange) {
       withDescription = false;
     }
-
     return renderError({
       t,
       error,
@@ -617,6 +693,7 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
       device: device ?? undefined,
       inlineRetry,
       withDescription,
+      currencyName: currencyName,
     });
   }
 
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
@@ -76,6 +76,10 @@ import Nano from "~/renderer/images/nanoS.v4.svg";
 import { DmkError } from "@ledgerhq/live-dmk-desktop";
 import { isDmkError } from "@ledgerhq/live-common/deviceSDK/tasks/core";
 import { isDisconnectedWhileSendingApduError } from "@ledgerhq/live-dmk-desktop";
+import {
+  DeviceDeprecationScreen,
+  DeviceDeprecationScreens,
+} from "./Screen/DeviceDeprecationScreen";
 
 export const AnimationWrapper = styled.div`
   width: 600px;
@@ -787,6 +791,7 @@ export const renderError = ({
   learnMoreTextKey,
   Icon,
   stretch,
+  currencyName = "",
 }: {
   error: Error | ErrorConstructor | DmkError;
   t: TFunction;
@@ -808,6 +813,7 @@ export const renderError = ({
   withDescription?: boolean;
   stretch?: boolean;
   Icon?: (props: { color?: string | undefined; size?: number | undefined }) => JSX.Element;
+  currencyName?: string;
 }) => {
   let tmpError = error;
   // Redirects from renderError and not from DeviceActionDefaultRendering because renderError
@@ -822,6 +828,14 @@ export const renderError = ({
     if (tmpError.title === "userRefused") {
       tmpError = new TransactionRefusedOnDevice();
     }
+  } else if (isDmkError(error) && error._tag === "DeviceDeprecationError") {
+    return (
+      <DeviceDeprecationScreen
+        productName={getDeviceModel(device?.modelId as DeviceModelId)?.productName}
+        screenName={DeviceDeprecationScreens.errorScreen}
+        coinName={currencyName}
+      />
+    );
   } else if (tmpError instanceof NoSuchAppOnProvider) {
     return (
       <NoSuchAppOnProviderErrorComponent
@@ -917,6 +931,7 @@ export const renderInWrongAppForAccount = ({
 }: {
   t: TFunction;
   onRetry?: (() => void) | null | undefined;
+  passWarning?: () => void;
 }) =>
   renderError({
     t,
--- a/apps/ledger-live-desktop/static/i18n/en/app.json
+++ b/apps/ledger-live-desktop/static/i18n/en/app.json
@@ -1598,6 +1598,26 @@
       }
     }
   },
+  "deviceDeprecation": {
+      "warning": {
+        "title": "{{device}}™ can not Clear Sign this transaction",
+        "subtitle": "Use any other Ledger device to verify this transaction’s details before signing. Without Clear Signing, you risk losing all your assets.",
+        "continue": "Proceed at your own risk"
+  
+      },
+      "info": {
+        "title": "As of {{date}}, {{coinName}} support on {{device}}™ via Ledger Live™ will end",
+        "subtitle": "Upgrade to a more recent Ledger device today to ensure access to all Ledger Live service providers and optimal customer support",
+        "reminder": "Do not remind me again"
+      },
+      "error": {
+        "title": "{{device}}™ does not support this feature",
+        "subtitle": "Upgrade to a more recent Ledger device today to ensure access to all the latest Ledger Live features."
+      },
+      "update": "Discover your Upgrade Program",
+      "learnMore": "Learn more",
+      "continue": "Continue"
+    },
   "currentAddress": {
     "title": "Current address",
     "for": "Address for <1><0>{{name}}</0></1>",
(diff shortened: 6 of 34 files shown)


apps/ledger-live-desktop/src/renderer/components/DeviceAction/RenderDeviceDeprecation.tsx
```diff
@@ -0,0 +1,123 @@
...
+            ? t("lnsDeprecation.warning.subtitle")
+            : t("lnsDeprecation.error.subtitle")
+          : t("lnsDeprecation.info.subtitle")}
+      </Text>
+      <Button
+        variant="main"
+        size="large"
+        onClick={e => {
+          e.preventDefault();
+          openURL("https://shop.ledger.com/pages/ledger-nano-s-upgrade-program");
```
francois-guerin-ledger: We prefer the use of handlers
```suggestion
        onClick={openDeprecationPage}
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +984,176 @@ export const renderConnectYourDevice = ({
...
+      >
+        <Trans i18nKey={`lnsDeprecation.warning.continue`} />
+      </ButtonV3>
+      <Link
+        href="https://support.ledger.com/article/Ledger-Nano-S-Limitations"
+        color="palette.text.shade60"
+        onClick={e => {
+          e.preventDefault();
+          openURL("https://support.ledger.com/article/Ledger-Nano-S-Limitations");
+        }}
```
francois-guerin-ledger: Weird to have `href` **and** `onClick`

francois-guerin-ledger (commented):
You may want to get rid of the conflict, to avoid having too much later.

Didn't review the whole thing yet

apps/ledger-live-desktop/src/renderer/actions/settings.ts
```diff
@@ -296,6 +302,8 @@ export const setExchangePairsAction: SetExchangePairs = pairs => ({
   type: "SETTINGS_SET_PAIRS",
   payload: pairs,
 });
+
+// disclaimer type
```
francois-guerin-ledger: Comment to remove

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -1597,6 +1597,26 @@
       }
     }
   },
+  "DeviceDeprecation": {
+      "warning": {
+        "title": "Ledger Nano S™ can not Clear Sign this transaction",
```
francois-guerin-ledger: Maybe the device can be a parameter as well ? To anticipate further deprecations
```suggestion
        "title": "{{device}}™ can not Clear Sign this transaction",
```

libs/ledger-live-common/src/__tests__/hw/deviceDeprectation.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  it("should not throw an error if passDeprecation is true", () => {
+    expect(() =>
+      throwErrorWhenDeviceDeprecated(mockDeviceManagementKit, mockSessionId, true, "Polkadot"),
+    ).not.toThrow();
+  });
+
+  it("should not throw an error if no deprecation config is found", () => {
+    jest.spyOn(LiveConfig, "getValueByKey").mockReturnValue(undefined);
```
francois-guerin-ledger: Instead of mocking, could you use `LiveConfig.setConfig( ... )`

libs/ledger-live-common/src/__tests__/hw/deviceDeprectation.ts
```diff
@@ -0,0 +1,134 @@
...
+import { DeviceDeprecationError } from "../../errors";
+import { throwErrorWhenDeviceDeprecated } from "../../hw/deviceDeprecation";
+
+jest.mock("@ledgerhq/device-management-kit");
+jest.mock("@ledgerhq/live-config/LiveConfig");
+
+describe("throwErrorWhenDeviceDeprecated", () => {
+  const mockDeviceManagementKit = {
+    getConnectedDevice: jest.fn(),
+  } as unknown as DeviceManagementKit;
```
francois-guerin-ledger: Mocking a complete module is overkill. See comment below

libs/ledger-live-common/src/__tests__/hw/deviceDeprectation.ts
```diff
@@ -0,0 +1,134 @@
...
+        {
+          deviceModelId: "nanoS",
+          errorScreen: { date: "2023-01-01" },
+        },
+      ],
+    };
+    jest.spyOn(LiveConfig, "getValueByKey").mockReturnValue(mockConfig);
+    (mockDeviceManagementKit.getConnectedDevice as jest.Mock).mockReturnValue({
+      modelId: "nanoS",
+    });
```
francois-guerin-ledger: You can create an instance of `DeviceManagementKit` using the `DeviceManagementKitBuilder`.
```typescript
const mockDeviceManagementKit = new DeviceManagementKitBuilder().build();

jest.spyOn(mockDeviceManagementKit, "getConnectedDevice").mockReturnValue({
    modelId: DeviceModelId.NANO_S,
} as unknown as ConnectedDevice);
```

libs/ledger-live-common/src/hw/actions/app.ts
```diff
@@ -489,6 +503,12 @@ export const createAction = (
               firmwareResolvedRef.current = true;
             } else if (e.type === "has-outdated-app") {
               outdatedAppRef.current = e.outdatedApp as AppAndVersion;
+            } else if (
+              e.type === "deprecation" &&
+              "onContinue" in e &&
+              typeof e.onContinue === "function"
```
francois-guerin-ledger:
```suggestion
              e.type === "deprecation"
```
This is enough for the type checking

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -39,6 +39,16 @@ import {
 } from "@ledgerhq/device-management-kit";
 import { ConnectAppDeviceAction } from "@ledgerhq/live-dmk-shared";
 import { ConnectAppEventMapper } from "./connectAppEventMapper";
+import { CryptoCurrencyId } from "@ledgerhq/types-cryptoassets";
+import { DeviceDeprecationError } from "../errors";
+import { throwErrorWhenDeviceDeprecated } from "./deviceDeprecation";
+
+type DeviceDeprecationErrorType = Error & {
```
francois-guerin-ledger: See this file that shows how we deal with custom errors
```
libs/ledger-live-common/src/errors.ts
```

francois-guerin-ledger (changes requested):

libs/ledger-live-common/src/hw/connectApp.ts
Antoine-bls83: thats what I use
`export const DeviceDeprecationError = createCustomErrorClass("DeviceDeprecationError");
`

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -364,7 +374,67 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
...
+        default:
+          return "other";
+      }
+    })();
+    const skippedDeprecations = stateSettings.deprecationDoNotRemind;
+    const alreadyDismissed = skippedDeprecations.includes(coin);
+    const currentToken =
+      request && typeof request === "object" && "tokenCurrency" in request && request.tokenCurrency
+        ? (request.tokenCurrency as TokenCurrency).name
+        : "undefined";
```
francois-guerin-ledger: Pretty uncommon to hardcode the string`"string"`. How about simply
```suggestion
        : "";
```

francois-guerin-ledger (commented):
I just merged https://github.com/LedgerHQ/ledger-live/pull/11252 to `develop`. Could you please rebase and check if you get the correct metadata in your error after an `instanceof` ?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,182 @@
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
+import { Provider } from "react-redux";
+import { createStore, combineReducers } from "redux";
+import { DefaultTheme, ThemeProvider } from "styled-components";
+
+import DeviceAction from "./index";
+import { HOOKS_TRACKING_LOCATIONS } from "~/renderer/analytics/hooks/variables";
+import { DeviceModelId } from "@ledgerhq/types-devices";
+import { account } from "~/newArch/features/Collectibles/__integration__/mocks/mockedAccount";
```
LucasWerey:
```suggestion
import { account } from "LLD/features/Collectibles/__integration__/mocks/mockedAccount";
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,182 @@
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
```
LucasWerey:
```suggestion
import { render, screen, fireEvent } from "tests/testSetup";
```

LucasWerey: You can remove all the providers you mock in your test this way

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+  onContinue: () => void;
+  deviceId: DeviceModelId;
+  isSwap?: boolean;
+}) => {
+  return (
+    <Box
+      flow={6}
+      mx={20}
+      style={{
+        display: "flex",
```
LucasWerey: you can use Flex component directly instead

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+          size="large"
+          variant="shade"
+          onClick={() => {
+            onContinue();
+          }}
+        >
+          <Trans i18nKey={`DeviceDeprecation.warning.continue`} />
+        </ButtonV3>
+        <Link
+          href="https://support.ledger.com/article/Ledger-Nano-S-Limitations"
```
LucasWerey: why href and openUrl?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+          >
+            <Trans i18nKey={`DeviceDeprecation.info.reminder`} />
+          </Text>
+        </Box>
+      </Box>
+    </Box>
+  );
+};
+
+export const DeprecationDeviceError = ({
```
LucasWerey: Can you divide this under multiple folders, please? It's getting quite complex to read

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -1597,6 +1597,26 @@
       }
     }
   },
+  "DeviceDeprecation": {
```
LucasWerey:
```suggestion
  "deviceDeprecation": {
```

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -153,6 +167,289 @@ export function renderRequestQuitApp({
...
+          onPress={() => openURL("https://support.ledger.com/article/Ledger-Nano-S-Limitations")}
+          testID="learn-more-link"
+          style={{
+            marginTop: 10,
+          }}
+        >
+          <Text
+            style={{
+              marginTop: 24, // équivaut à mt={4}
+              color: "#999999", // remplace par ton vrai code couleur du thème si possible
```
LucasWerey: to be changed

apps/ledger-live-mobile/src/locales/en/common.json
```diff
@@ -1520,6 +1520,25 @@
     "bannerCta": "Begin setup",
     "continue": "Continue"
   },
+  "DeviceDeprecation": {
```
LucasWerey:
```suggestion
  "deviceDeprecation": {
```

LucasWerey (dismissed):

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
mcayuelas-ledger: You can use `from "@testing-library/react"` — no need to mock the theme or Redux in this case.
You’ll also be able to initialize the state easily.

I recommend checking out this example for reference:
`apps/ledger-live-desktop/src/newArch/features/ModularDrawer/__tests__/modules.test.tsx`

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -364,7 +374,69 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
...
-  if (displayUpgradeWarning && appAndVersion && passWarning) {
+  if (deprecate && deprecateData && onContinue && request) {
+    const { account } = request as unknown as {
+      account: Account;
+    };
+    const { deprecatedFlowExceptions, tokenExceptions, date, modelId } = deprecateData;
+    const currentFlow = (() => {
+      switch (location) {
+        case HOOKS_TRACKING_LOCATIONS.genericDAppTransactionSend:
+          return "stake";
```
mcayuelas-ledger: is this typed somewhere?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+        flow={4}
+        style={{
+          width: "100%",
+        }}
+      >
+        <ButtonV3
+          variant="main"
+          size="large"
+          onClick={(e: { preventDefault: () => void }) => {
+            e.preventDefault();
```
LucasWerey: is the preventDefault necessary?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -805,7 +826,9 @@ export const renderError = ({
   inlineRetry?: boolean;
   withDescription?: boolean;
   stretch?: boolean;
+  passWarning?: (() => void) | null | undefined;
```
mcayuelas-ledger:
```suggestion
  passWarning?: () => void;
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -156,6 +158,10 @@ type States = PartialNullable<{
   manifestId: string;
   transactionChecksOptInTriggered: boolean;
   transactionChecksOptIn: boolean;
+  deprecate?: boolean;
+  deprecateData?: DeviceDeprecation;
```
mcayuelas-ledger: could be combined no?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -364,7 +374,69 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
...
   }, [dispatch, device, deviceInfo, latestFirmware]);
 
-  if (displayUpgradeWarning && appAndVersion && passWarning) {
+  if (deprecate && deprecateData && onContinue && request) {
+    const { account } = request as unknown as {
+      account: Account;
+    };
+    const { deprecatedFlowExceptions, tokenExceptions, date, modelId } = deprecateData;
+    const currentFlow = (() => {
+      switch (location) {
```
LucasWerey: you can do a mapping instead 

```
const flowMapping = {
  [HOOKS_TRACKING_LOCATIONS.sendFlow]: "send",
  [HOOKS_TRACKING_LOCATIONS.receiveFlow]: "receive",
  [HOOKS_TRACKING_LOCATIONS.swapFlow]: "swap",
};

const currentFlow = flowMapping[location] || "other";
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+          }}
+        >
+          <Trans i18nKey={`DeviceDeprecation.update`} />
+        </ButtonV3>
+        <ButtonV3
+          size="large"
+          variant="shade"
+          onClick={() => {
+            onContinue();
+          }}
```
mcayuelas-ledger:
```suggestion
          onClick={onContinue}
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -912,15 +941,18 @@ export const renderError = ({
 export const renderInWrongAppForAccount = ({
   t,
   onRetry,
+  passWarning,
 }: {
   t: TFunction;
   onRetry?: (() => void) | null | undefined;
+  passWarning?: (() => void) | null | undefined;
```
LucasWerey:
```suggestion
  passWarning?: (() => void)
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+          textAlign="center"
+          fontSize={24}
+          data-testid="warning-deprecation-title"
+        >
+          <Trans
+            i18nKey={`DeviceDeprecation.info.title`}
+            values={{
+              device: getDeviceModel(deviceId).productName,
+              date: dateFormatted,
+              coinName: coinName.charAt(0).toUpperCase() + coinName.slice(1),
```
LucasWerey: What's the result of that? Looks weird

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
```diff
@@ -308,6 +320,66 @@ export function DeviceActionDefaultRendering<R, H extends Status, P>({
...
   const walletState = useSelector(walletSelector);
   const settingsState = useSelector(settingsStoreSelector);
+  if (deprecate && deprecateData && onContinue) {
+    const { account } = request as unknown as {
+      account: Account;
+    };
+    const { deprecatedFlowExceptions, tokenExceptions, date, modelId } = deprecateData;
+    const currentFlow = (() => {
+      switch (location) {
+        case HOOKS_TRACKING_LOCATIONS.sendFlow:
```
LucasWerey: Same here you can do a mapping

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
mcayuelas-ledger, line 1: I think you could make this a more generic component that takes the displayed data as a parameter.
Right now, there’s quite a bit of duplication.

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
mcayuelas-ledger, line 1: Same as LLD

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -153,6 +167,289 @@ export function renderRequestQuitApp({
...
```
mcayuelas-ledger: 🤖

use colors from useTheme() :)

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
Antoine-bls83: I wanted to do that:
`onClick={openURL("https://shop.ledger.com/pages/ledger-nano-s-upgrade-program")}`

but I get this error:

```
Type 'void' is not assignable to type '((event: SyntheticEvent<HTMLButtonElement, Event>) => void) | undefined'.ts(2322)
index.d.ts(25, 5): The expected type comes from property 'onClick' which is declared here on type 'IntrinsicAttributes & ButtonProps'
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
Antoine-bls83: So how can I know if the action is a staking or other?

francois-guerin-ledger (dismissed):

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
LucasWerey: Yep but you can either create a function like handleClick (or more specific) or just to () => openURL("https://shop.ledger.com/pages/ledger-nano-s-upgrade-program")

Also I just noticed but urls should be accessed via `urls.` so you need to add them inside urls file for LLD and LLM

libs/ledger-live-common/src/__tests__/hw/deviceDeprectation.ts
OlivierFreyssinet, line 1: Typo in file name (should be **deviceDeprecation**, not deviceDeprec*t*ation)

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -820,6 +843,12 @@ export const renderError = ({
     if (tmpError.title === "userRefused") {
       tmpError = new TransactionRefusedOnDevice();
     }
+  } else if (error instanceof DeviceDeprecationError) {
+    const castDeviceModelId =
+      error.modelId.toString() === "flex"
+        ? DeviceModelId.europa
+        : (error.modelId.toString() as DeviceModelId);
```
OlivierFreyssinet: Two issues here:
- The way this "cast" is done is not type safe at all. You are using a hardcoded string and in case the value of it changes, or the value of DeviceModelId.europa changes, this will be broken.
- This remapping should have been done upstream, i.e. where the `DeviceDeprecationError` is instantianted. That way, you only do the remapping once, and it scales. Here you do it 4 times if I'm counting well, and it's 4 times approximately the same code, which multiplies the risk of introducing an error down the road when people maintain the codebase.

I will put a comment in `deviceDeprecation.ts` explaining how to properly do this remapping.

libs/ledger-live-common/src/hw/deviceDeprecation.ts
```diff
@@ -0,0 +1,66 @@
...
+  const config =
+    appName === "Exchange" && dependencies && dependencies.length > 0
+      ? LiveConfig.getValueByKey(
+          `config_nanoapp_${dependencies[0].toLowerCase().replace(/ /g, "_")}`,
+        )
+      : LiveConfig.getValueByKey(`config_nanoapp_${appName.toLowerCase().replace(/ /g, "_")}`);
+  if (!config || !config.deviceDeprecated) {
+    return;
+  }
+  const { modelId } = dmk.getConnectedDevice({ sessionId });
```
OlivierFreyssinet: This should be remapped to a LL `DeviceModelId` type.
You can use `dmkToLedgerDeviceIdMap` from the `"@ledgerhq/live-dmk-shared"` package to do this.

OlivierFreyssinet: Advice for the future, don't hesitate to ask help from the device team members before modifying things in our scope, chan #team-devices on Slack 😉

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -959,6 +991,289 @@ export const renderConnectYourDevice = ({
...
+  return (
+    <Box
+      flow={6}
+      mx={20}
+      alignSelf="center"
+      style={{
+        display: "flex",
+        flexDirection: "column",
+        alignItems: "center",
+        justifyContent: "center",
```
OlivierFreyssinet: There's a component `Flex` from `"@ledgerhq/react-ui"` that does what you need. And 99% of the time you don't need to use inline styles in LL.
Don't hesitate to check the rest of the codebase to see how components from this UI library are supposed to be used (thousands of examples).

And when in doubt, you can ask the live-hub team. It's usually much better if you ask other teams for help during your implementation process rather than by asking for a PR review after 1 month of work. It will be much smoother and save a lot of time for everyone involved.

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
OlivierFreyssinet: 1. Have you tried in dark and light mode ? A hardcoded color is a bad idea for many reasons, one of them being that there are several themes in the app. Another one is that we want to keep a consistent design in the app, and if the designers decide to change the color of some text, it should change everywhere in the app. And here it will stay at a wrong value.
2. Again, referring to my other comment, don't use inline styles, use the styling props from the components of the UI lib. There is a storybook available and many many examples (thousands) across the LL codebase. Live Hub is the team to reach out to for this kind of advice. In that specific case, you can use: `mt={4}` (no hardcoding of spacing values, we have a predefined scale), `textAlign="center"`, `color="whateverColorIsInFigma"` but not a hardcoded one, it should use a color shade from the theme.
3. Maybe self review your PR and delete comments from your AI before submitting a PR for review...

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -57,6 +66,16 @@ export type ConnectAppRequest = {
...
+};
+
+export type DeviceDeprecation = {
+  warningScreen: boolean;
+  warningClearSigning: boolean;
+  modelId: DeviceModelId;
+  date: string;
+  tokenExceptions: string[];
+  deprecatedFlowExceptions: string[];
 };
```
OlivierFreyssinet: Consider adding tsdoc comments to your types so everyone reading the code can understand exactly what it's for, how to use it etc.

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -57,6 +66,16 @@ export type ConnectAppRequest = {
   requireLatestFirmware?: boolean;
   outdatedApp?: AppAndVersion;
   allowPartialDependencies: boolean;
+  passDeprecation?: boolean;
+};
+
+export type DeviceDeprecation = {
+  warningScreen: boolean;
```
OlivierFreyssinet: Consider using a different naming for booleans. For instance `"warningScreenVisible"` (if this is what this property means) is much easier to understand.

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -57,6 +66,16 @@ export type ConnectAppRequest = {
...
+  warningClearSigning: boolean;
+  modelId: DeviceModelId;
+  date: string;
```
OlivierFreyssinet: What is the formatting of this date string ? Is is a unix timestamp ? ISO 8601 ?
I don't know in your case, but using a `Date` object might be much wiser. No risk of badly encoding or decoding the date.

.changeset/little-fishes-run.md
```diff
@@ -0,0 +1,7 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
+"@ledgerhq/live-common": minor
+---
+
+Device Deprecation
```
mbertin-ledger: [MUST] Indicate Nano S

apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const toggleCheck = useCallback(() => {
+    setIsChecked(prev => !prev);
+  }, []);
+
+  const [currentScreenName, setCurrentScreenName] = useState(screenName);
+  const handleUgradeClick = () => {
+    openURL("https://shop.ledger.com/pages/ledger-nano-s-upgrade-program");
+  };
+  const handleLearnMoreClick = () => {
+    openURL("https://support.ledger.com/article/Ledger-Nano-S-Limitations");
```
mbertin-ledger: [MUST] Use constant for URL.

mbertin-ledger (changes requested):
This deprecation feature should belong to the ConnectAppDeviceAction directly and not bring a tape using observable, as it will result in the same understandable pattern that was used for legacy connect app.  

This could be a new state here `checkDeprecation` : `./libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts`

More over, I know it's not your fault but touching the `connectApp` is really critical so I would like the team-device involve in the spec for this. Regarding the criticity of it I would like to have a full QA on the connect app.

libs/live-dmk-shared/src/device-action/ConnectApp/types.ts
```diff
@@ -8,32 +8,57 @@ import type {
...
   readonly deviceMetadata?: GetDeviceMetadataDAOutput;
   readonly installResult?: InstallOrUpdateAppsDAOutput;
   readonly derivation?: ConnectAppDerivation;
+  readonly deviceDeprecation?: DeviceDeprecationPayload;
 };
 
 export type ConnectAppDAInput = OpenAppWithDependenciesDAInput & {
   readonly allowMissingApplication: boolean;
   readonly requiredDerivation?: () => Promise<ConnectAppDerivation>;
+  readonly appConfig?: any;
```
paoun-ledger: is it possible to have a better typing? We usually prefer to avoid `any`, especially since XState allow to properly type inputs, outputs, and intermediate values

libs/live-dmk-shared/src/device-action/ConnectApp/types.ts
```diff
@@ -8,32 +8,57 @@ import type {
...
+export type DeviceDeprecationPayload = {
+  warningScreenVisible: boolean;
+  clearSigningScreenVisible: boolean;
+  errorScreenVisible: boolean;
+  modelId: LLDeviceModelId;
+  date: Date;
+  warningScreenConfig?: DeviceDeprecationConfig;
+  clearSigningScreenConfig?: DeviceDeprecationConfig;
+  errorScreenConfig?: DeviceDeprecationConfig;
+  onContinue: (value?: string) => void;
```
paoun-ledger: I find it weird to take a `value?: string` in parameter, and then only verify if it equals `error`. It would be clearer to take a boolean instead, something like `abort` or `error`, wdyt?

libs/ledger-live-common/src/hw/connectAppEventMapper.ts
```diff
@@ -132,7 +132,7 @@ export class ConnectAppEventMapper {
   }
 
   private handlePendingEvent(intermediateValue: ConnectAppDAIntermediateValue): void {
-    switch (intermediateValue.requiredUserInteraction) {
+    switch (intermediateValue.requiredUserInteraction.toString()) {
```
paoun-ledger: you should not need to convert it to string, you loose the typing here and therefore the switch-case is not exhaustive anymore

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -39,6 +39,44 @@ import {
...
+ *
+ * @property warningScreenVisible - Whether the generic deprecation warning screen should be shown.
+ * @property clearSigningScreenVisible - Whether the clear signing deprecation warning screen should be shown.
+ * @property errorScreenVisible - Whether the deprecation error screen should be shown (blocking usage).
+ * @property productName - The name of the affected product (e.g., "Nano S").
+ * @property date - The date when the deprecation becomes effective.
+ * @property warningScreenConfig - Optional configuration for the warning screen.
+ * @property clearSigningScreenConfig - Optional configuration for the clear signing screen.
+ * @property errorScreenConfig - Optional configuration for the error screen.
+ * @property onThrowError - Callback function triggered when an error should be thrown.
```
paoun-ledger: the doc is not up-to-date. Moreover it look like a dupplicate of the device action type, maybe you could just re-export it?

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.test.ts
```diff
@@ -103,13 +103,15 @@ describe("OpenAppWithDependenciesDeviceAction", () => {
             intermediateValue: {
```
paoun-ledger, line 132: you should test your modifications, the state machine was fully tested before

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -95,6 +100,67 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
...
+        clearSigningScreenVisible: false,
+        errorScreenVisible: false,
+        modelId: dmkToLedgerDeviceIdMap[modelId],
+        date: fallbackDate,
+        warningScreenConfig: { exception: [], deprecatedFlow: [] as string[] },
+        clearSigningScreenConfig: { exception: [], deprecatedFlow: [] as string[] },
+        errorScreenConfig: { exception: [], deprecatedFlow: [] as string[] },
+        onThrowError: () => {},
+        onContinue: () => {},
+      };
```
paoun-ledger: this function don't seem to be safe since `appconfig` is of type `any` and here you access some fileds without question mark, and assign them with no type verification. That's why it would be better to properly type it and verify the firebase config format beforehand. Or at least use `unknown` instead and then verify the typing here, but it make the device action API not clear

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -128,6 +194,33 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
...
+          const cfg = context.input.appConfig;
+          const today = new Date();
+          const modelId = context._internalState.deviceModel;
+          if (!cfg?.deviceDeprecated?.length) return false;
+          return cfg.deviceDeprecated.some((dd: any) => {
+            if (dd.deviceModelId !== modelId) return false;
+            if (dd.errorScreen) {
+              const errorDate = dd.errorScreen?.date ? new Date(dd.errorScreen.date) : null;
+              if (errorDate && errorDate < today)
+                return true && !context._internalState.deprecationShown;
```
paoun-ledger: not sure why the `true &&` is needed?

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -235,6 +347,61 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
...
+            src: fromPromise(async ({ input }) => {
+              return new Promise<void>((resolve, reject) => {
+                const payload = input.deviceDeprecation;
+                if (!payload) {
+                  resolve();
+                  return;
+                }
+                payload.onContinue = (value?: string) => {
+                  if (value === "error") {
+                    reject(new Error("device-deprecation"));
```
paoun-ledger: ou should create a new error and add it to `ConnectAppDAError`

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -284,8 +451,8 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
         GetDeviceMetadataCheck: {
           always: [
             {
-              target: "Error",
-              guard: "hasError",
+              guard: "isDeprecation",
+              target: "DeviceDeprecation",
```
paoun-ledger: I think that modification should not be done, or did I miss something?

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
paoun-ledger: by the way, wouldn't it be simpler and clearer to early-return if `deprecationShown` is set?

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -235,6 +347,61 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
...
+        DeviceDeprecation: {
+          entry: assign(({ context }) => {
+            const cfg = context.input.appConfig;
+            const modelId = context._internalState?.deviceModel;
+
+            if (!modelId) {
+              throw new Error("Device model ID is undefined");
+            }
+            const deviceDeprecation = computeDeviceDeprecation(cfg, modelId);
+            return {
```
paoun-ledger: maybe a `...context` is needed, or you may loose fields from the context such as inputs, no? Also, is it really needed to add the `deviceDeprecation` in both internalState and intermediateValue? Maybe assigning the intermediateValue could be enough instead?

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -235,6 +347,61 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
...
+                deviceDeprecation: deviceDeprecation,
+              },
+              _internalState: {
+                ...context._internalState,
+                deviceDeprecation: deviceDeprecation,
+                deprecationShown: true,
+              },
+            };
+          }),
+          always: "waitForUiAnswer",
```
paoun-ledger: not sure it's safe to always go to `waitForUiAnswer` since there is some logic in `entry`, with even exception thrown? It would be cleaner tu put the logic in a `src`, then assign the intermediateValue in `onDone`, and catch errors

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
```diff
@@ -119,6 +124,7 @@ type Status = PartialNullable<{
   imageCommitRequested: boolean;
   transactionChecksOptInTriggered: boolean;
   transactionChecksOptIn: boolean;
+  deprecateData?: DeviceDeprecation;
```
mbertin-ledger: [COULD] Rename `DeviceDeprecationConfig`

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
```diff
@@ -308,6 +316,90 @@ export function DeviceActionDefaultRendering<R, H extends Status, P>({
...
+    tokenCurrency?: TokenCurrency | null;
+  }
+
+  function extractCoin(request?: Request): string {
+    if (!request) return "";
+    const { account, tokenCurrency } = request;
+    if (tokenCurrency) return tokenCurrency.name;
+    if (account) return account.currency.name;
+    return "";
+  }
```
mbertin-ledger: [MUST] Refactor and isolate this function  (+ TU) it's already existing in 
`apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx`

libs/ledger-live-common/src/apps/support.ts
```diff
@@ -54,3 +54,13 @@ export function getMinVersion(appName: string, model?: DeviceModelId): string |
   const minVersion = config?.[`${model?.toLowerCase()}MinVersion`] ?? config?.minVersion;
   return minVersion ? semver.coerce(minVersion)?.version : undefined;
 }
+
+export function getDeprectationConfig(appName: string, dependencies?: string[]) {
```
mbertin-ledger: [MUST] Typo `Deprecation` not  `Deprectation`

libs/ledger-live-common/src/hw/connectApp.ts
```diff
@@ -39,6 +39,43 @@ import {
...
+ * @property clearSigningScreenVisible - Whether the clear signing deprecation warning screen should be shown.
+ * @property errorScreenVisible - Whether the deprecation error screen should be shown (blocking usage).
+ * @property modelId - The modeID of the affected product
+ * @property date - The date when the deprecation becomes effective.
+ * @property warningScreenConfig - Optional configuration for the warning screen.
+ * @property clearSigningScreenConfig - Optional configuration for the clear signing screen.
+ * @property errorScreenConfig - Optional configuration for the error screen.
+ * @property onContinue - Callback invoked when the user chooses to continue or throw an error despite the deprecation warning.
+ */
+export type DeviceDeprecation = {
```
mbertin-ledger: [COULD] Like I said above, type should be a little more explicit, so name the type `DeviceDepecationConfig` could be better.

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -115,6 +122,16 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
           }
           return await this.input.requiredDerivation();
         }),
+        computeDeprecation: fromPromise(async ({ input }) => {
```
mbertin-ledger: [COULD] Rename `isDeviceDeprecated`

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -192,7 +227,7 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
               }),
             },
             onDone: {
-              target: "GetDeviceStatusCheck",
+              target: "DeprecationCheck",
```
mbertin-ledger: [COULD] Same here, `IsDeviceDeprecatedCheck`

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -216,12 +251,31 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
             },
           },
         },
+        DeprecationCheck: {
+          always: [
+            {
+              guard: "isDeprecation",
+              target: "DeviceDeprecation",
```
mbertin-ledger: [COULD] As it's a state it is better to use `DeviceDeprecated`

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -128,6 +145,24 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
         }) => ConnectAppDeviceAction.isAppOpened(application, deviceStatus!, deviceModel!),
         requiresDerivation: ({ context }) => context.input.requiredDerivation !== undefined,
         hasError: ({ context }) => context._internalState.error !== null,
+        isDeprecation: ({ context }) => {
```
mbertin-ledger, line 148: [SHOULD] Rename `shouldCheckDeprecation` ?

libs/live-dmk-shared/src/device-action/ConnectApp/ConnectAppDeviceAction.ts
```diff
@@ -235,6 +289,65 @@ export class ConnectAppDeviceAction extends XStateDeviceAction<
...
+                }),
+              }),
+            },
+            onError: {
+              target: "Error",
+              actions: "assignErrorFromEvent",
+            },
+          },
+        },
+        waitForUiAnswer: {
```
mbertin-ledger: [SHOULD] Be more specific like `waitForDeprecationAcknowledgment` because if we add a new state which wait also for a user acknowlegdment it would bring confusion.

libs/live-dmk-shared/src/device-action/ConnectApp/deprecation.ts
```diff
@@ -0,0 +1,71 @@
+import { dmkToLedgerDeviceIdMap } from "../../config/dmkToLedgerDeviceIdMap";
+import type {
+  DeviceDeprecated,
+  DeviceDeprecatedEntry,
+  DeviceDeprecationPayload,
+  DeviceDeprecationConfig,
+} from "./types";
+import type { DeviceModelId } from "@ledgerhq/device-management-kit";
+
+export function computeDeviceDeprecation(
```
mbertin-ledger: [SHOULD] Same Naming `isDeviceDeprecated` ?

libs/live-dmk-shared/src/device-action/ConnectApp/types.ts
```diff
@@ -8,32 +8,80 @@ import type {
...
+  deprecatedFlow: string[];
+  exception?: string[];
+};
+
+export type DeviceDeprecatedEntry = {
+  deviceModelId: string;
+  infoScreen: DeprecationScreen;
+  errorScreen: DeprecationScreen;
+  warningClearSigningScreen: DeprecationScreen;
 };
```
mbertin-ledger, line 76: [ASK] Can you explain me what are the purpose of each type declared above and when they are used ?

Antoine-bls83: i change by:
```
export type DeviceDeprecationRules = {
  warningScreenVisible: boolean;
  clearSigningScreenVisible: boolean;
  errorScreenVisible: boolean;
  modelId: LLDeviceModelId;
  date: Date;
  warningScreenRules?: DeviceDeprecationScreenRules;
  clearSigningScreenRules?: DeviceDeprecationScreenRules;
  errorScreenRules?: DeviceDeprecationScreenRules;
  onContinue: (isError?: boolean) => void;
};
```
All of this data are sending to the DeviceAction in LLD/LLM 
The boolean are here to say to the `deviceAction` into LLD/LLM to display the different deprecation Screen, the date is the date when the deprecation will be effective (when the error screen should be display) and the rules  because the PO want to except some flow (send, receive or swap...) and want to except some currency 
onContinue is a callback it will be call if the frontend decide to not display the screen because of the exception or just when the user has click on continue

I also change a type by DeviceDeprecationConfigs.
It's the config in firebase sending to the ConnectAppDeviceAction as input

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account }}
+        location={HOOKS_TRACKING_LOCATIONS.sendModal}
+      />,
+    );
+
+    screen.debug();
```
LucasWerey:
```suggestion
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,207 @@
+import React, { useCallback, useState } from "react";
+import { Trans } from "react-i18next";
+import { useDispatch } from "react-redux";
+import styled from "styled-components";
+
+import CheckBox from "~/renderer/components/CheckBox";
+import Box from "~/renderer/components/Box";
+import { Button as ButtonV3, Flex, Icons, IconsLegacy, Link, Text } from "@ledgerhq/react-ui";
```
LucasWerey:
```suggestion
import { Button, Flex, Icons, IconsLegacy, Link, Text } from "@ledgerhq/react-ui";
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,252 @@
+import React from "react";
+import { render, screen, fireEvent } from "tests/testSetup";
+import DeviceAction from "./index";
+import { HOOKS_TRACKING_LOCATIONS } from "~/renderer/analytics/hooks/variables";
+import { account } from "LLD/features/Collectibles/__integration__/mocks/mockedAccount";
```
LucasWerey: It's ok but a good practice would be to use a dedicated mock. 
FYI you can do it like this : 

```
const ethereumCurrency = getCryptoCurrencyById("ethereum");
const ETH_ACCOUNT = genAccount("ethereum-1", {
  currency: ethereumCurrency,
});
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account }}
+        location={HOOKS_TRACKING_LOCATIONS.receiveModal}
+      />,
+    );
+
+    screen.debug();
```
LucasWerey:
```suggestion
```

apps/ledger-live-desktop/src/renderer/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+    expect(screen.getByTestId("warning-deprecation-title")).toBeInTheDocument();
```
LucasWerey: Instead of testing based on test id we should test what the user actually sees. Something like

getByText(/idk what he sees/i).toBeVisible

Because when we use toBeInTheDocument we can't be sure it's visible. Also, testing text is a good thing to avoid unexpected regression (like i18next key removal)

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -367,6 +378,75 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
...
+    );
+    const skipErrorScreen = doSkipDeprecation(
+      errorScreenRules?.exception || [],
+      errorScreenRules?.deprecatedFlow || [],
+    );
+    const handleContinue = () => {
+      setHasDisplayDeprecateWarning(true);
+      deviceDeprecationRules.onContinue(false);
+    };
+    if (deviceDeprecationRules.errorScreenVisible) {
```
LucasWerey, line 432: [ASK] Perhaps we could make it easier to read instead of having so many `if` overlaps

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -367,6 +378,75 @@ export const DeviceActionDefaultRendering = <R, H extends States, P>({
...
+            isSwap={currentFlow === "swap"}
+            screenName={DeviceDeprecationScreens.warningScreen}
+          />
+        );
+      } else if (deviceDeprecationRules.clearSigningScreenVisible && !skipClearSigningScreen) {
+        return (
+          <DeviceDeprecationScreen
+            onContinue={handleContinue}
+            productName={getDeviceModel(modelId).productName}
+            isSwap={currentFlow === "swap"}
```
LucasWerey: [RECO] Use an enum instead

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -822,6 +830,17 @@ export const renderError = ({
     if (tmpError.title === "userRefused") {
       tmpError = new TransactionRefusedOnDevice();
     }
+  } else if (isDmkError(error) && error._tag === "DeviceDeprecationError") {
+    return (
+      <DeviceDeprecationScreen
+        isSwap={isSwap}
+        productName={getDeviceModel(device?.modelId as DeviceModelId)?.productName}
+        onContinue={() => {}}
```
LucasWerey: maybe make the param optional it doesn't look good to have this

apps/ledger-live-mobile/src/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,212 @@
...
+      displayClearSigningWarning &&
+      currentScreenName === DeviceDeprecationScreens.warningScreen
+    ) {
+      setCurrentScreenName(DeviceDeprecationScreens.clearSigningScreen);
+    } else {
+      onContinue();
+    }
+  }, [isChecked, displayClearSigningWarning, currentScreenName, onContinue, dispatch, coinName]);
+
+  const mapedIcon = {
```
LucasWerey, line 83:
```suggestion
  const mappedIcons = {
```

apps/ledger-live-mobile/src/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,212 @@
...
+    [DeviceDeprecationScreens.errorScreen]: (
+      <Trans
+        i18nKey={`deviceDeprecation.error.title`}
+        values={{
+          device: productName,
+        }}
+      />
+    ),
+  };
+  const mapedSubTitleKey = {
```
LucasWerey, line 122:
```suggestion
  const mappedSubTitleKeys = {
```

apps/ledger-live-mobile/src/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,259 @@
+import React from "react";
+import { render, screen, fireEvent } from "@tests/test-renderer";
+import DeviceAction from "./index";
+import { HOOKS_TRACKING_LOCATIONS } from "~/analytics/hooks/variables";
+import { MockedAccounts } from "LLM/features/Accounts/__integrations__/mockedAccounts";
```
LucasWerey: Same as before you can generate them directly

apps/ledger-live-mobile/src/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+    render(
+      <DeviceAction
+        action={mockAction}
+        request={{ account }}
+        device={nanoS}
+        location={HOOKS_TRACKING_LOCATIONS.receiveFlow}
+      />,
+    );
+
+    screen.debug();
```
LucasWerey:
```suggestion

```

apps/ledger-live-mobile/src/components/DeviceAction/deviceAction.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+    };
+
+    const mockAction = {
+      useHook: jest.fn(() => ({
+        deviceDeprecationRules,
+      })),
+      mapResult: jest.fn(),
+    };
+
+    render(
```
LucasWerey: you can do `const { getByText, getAllByText, user } = render(`

and then use user.press instead of fireEvent

apps/ledger-live-mobile/src/components/DeviceAction/rendering.tsx
```diff
@@ -668,6 +674,17 @@ export function renderError({
   if (error instanceof PeerRemovedPairing) {
     showRetryIfAvailable = false;
   }
+  if (error.message === "device-deprecation") {
+    return (
+      <DeviceDeprecationScreen
+        coinName={currencyName}
+        date={new Date()}
+        onContinue={() => {}}
```
LucasWerey: same here doesn't look good to have that

apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const dispatch = useDispatch();
+  const [isChecked, setIsChecked] = useState(false);
+  const dateFormatted = useCalendarFormatted(date);
+  const toggleCheck = useCallback(() => {
+    setIsChecked(prev => !prev);
+  }, []);
+
+  const [currentScreenName, setCurrentScreenName] = useState(screenName);
+  const handleUgradeClick = () => {
+    openURL(urls.deviceDeprecation.shop);
```
mcayuelas-ledger: You can use `useLocalizedUrls` from`apps/ledger-live-desktop/src/renderer/hooks/useLocalizedUrls/index.ts` to localize your urls

apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,207 @@
...
+      <Trans
+        i18nKey={`deviceDeprecation.error.title`}
+        values={{
+          device: productName,
+        }}
+      />
+    ),
+  };
+  const mapedSubTitleKey = {
+    [DeviceDeprecationScreens.clearSigningScreen]: `deviceDeprecation.warning.subtitle`,
```
mcayuelas-ledger:
```suggestion
    [DeviceDeprecationScreens.clearSigningScreen]: "deviceDeprecation.warning.subtitle",
```

apps/ledger-live-mobile/src/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const toggleCheck = useCallback(() => {
+    setIsChecked(prev => !prev);
+  }, []);
+
+  const dateFormatted = formatDate(date);
+  const openURL = useCallback((url: string) => {
+    Linking.openURL(url);
+  }, []);
+  const handleUgradeClick = () => {
+    openURL(urls.deviceDeprecation.shop);
```
mcayuelas-ledger: same here for localized url

Justkant:
It should be reopened and rebased at some point

gre-ledger (approved):
ok for platform scope

apps/ledger-live-desktop/src/renderer/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,218 @@
...
+  const dispatch = useDispatch();
+  const [isChecked, setIsChecked] = useState(false);
+  const dateFormatted = useDateFormatted(date);
+  const toggleCheck = useCallback(() => {
+    setIsChecked(prev => !prev);
+  }, []);
+  const shopUrl = useLocalizedUrl(urls.deviceDeprecation.shop);
+  const learnMoreUrl = useLocalizedUrl(urls.deviceDeprecation.learnMore);
+  const [currentScreenName, setCurrentScreenName] = useState(screenName);
+  const handleUgradeClick = () => {
```
mcayuelas-ledger: little typo

libs/ledger-live-common/src/device-action/utils.ts
```diff
@@ -0,0 +1,56 @@
+import { TransportStatusError } from "@ledgerhq/errors";
+import { DeviceNotOnboarded } from "../errors";
+import { TokenCurrency } from "@ledgerhq/types-cryptoassets";
+import { Account } from "@ledgerhq/types-live";
```
mcayuelas-ledger, line 4: Seems relevant

apps/ledger-live-mobile/src/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,216 @@
...
+    ) {
+      setCurrentScreenName(DeviceDeprecationScreens.clearSigningScreen);
+    } else if (onContinue) {
+      onContinue();
+    }
+  }, [isChecked, displayClearSigningWarning, currentScreenName, onContinue, dispatch, coinName]);
+
+  const mapedIcon = {
+    [DeviceDeprecationScreens.clearSigningScreen]: (
+      <Icons.WarningFill size="L" color="palette.warning.c70" />
```
mcayuelas-ledger: this color not exists anymore should be
```suggestion
      <Icons.WarningFill size="L" color="warning.c70" />
```

apps/ledger-live-mobile/src/components/DeviceAction/Screen/DeviceDeprecationScreen.tsx
```diff
@@ -0,0 +1,216 @@
...
+      <Link
+        onPress={handleLearnMoreClick}
+        style={{
+          marginTop: 10,
+        }}
+      >
+        <Text
+          mt={4}
+          style={{
+            color: "palette.text.shade60",
```
mcayuelas-ledger: This color not exists anymore should be something equivalent `neutral.c60` or `neutral.c70`

PR #10781, opened 2025-07-02
Fix: changing debit account no longer disables continue button

Canestin:
### ✅ Checklist

  - ...

### 📝 Description

Fixes a bug where changing the debit account in the send flow clears the recipient internally, disabling the "Continue" button even though the input looks valid.

#### Solution Demo

https://github.com/user-attachments/assets/41577d21-d25c-46c6-aa9e-ffdc896df629

### ❓ Context

- **JIRA**: https://ledgerhq.atlassian.net/browse/LIVE-19341

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.tsx
@@ -46,7 +46,7 @@ const RecipientField = <T extends Transaction, TS extends TransactionStatus>({
       onChangeTransaction(bridge.updateTransaction(transaction, { recipient: value }));
       resetInitValue && resetInitValue();
     }
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [transaction]); // eslint-disable-line react-hooks/exhaustive-deps
 
   const onChange = useCallback(
     async (recipient: string, maybeExtra?: OnChangeExtra | null) => {
--- a/.changeset/angry-waves-beg.md
+++ b/.changeset/angry-waves-beg.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": minor
+---
+
+fix: changing debit account no longer disables continue button


apps/ledger-live-desktop/src/renderer/modals/Send/fields/RecipientField.tsx
```diff
@@ -46,7 +46,7 @@ const RecipientField = <T extends Transaction, TS extends TransactionStatus>({
       onChangeTransaction(bridge.updateTransaction(transaction, { recipient: value }));
```
Justkant, line 46: Should we also update the code as seen below in the `onChange` ?
We are missing the currency that comes from the maybeExtra
```suggestion
      const invalidRecipient = currency && currency.scheme !== account.currency.scheme;
      onChangeTransaction(
        bridge.updateTransaction(transaction, {
          recipient: invalidRecipient ? "" : value,
        }),
      );
```

Also we could remove the `onChangeTransaction` part from the `onChange` handler if we do it in the `useEffect` but I suppose that if we only did the first set in the `useEffect` it's to avoid setting the recipient if it's an incorrect `currency.scheme` (This change might introduce a sneaky bug)

Canestin: Helloo! In practice, `maybeExtra` is always `undefined` (except when scanning an address via QR code — and even then, it's not guaranteed). You can see it in the onChange call [here](https://github.com/LedgerHQ/ledger-live/blob/e28c3ab39d6920878105a3b5b5b9e1f7cd2f8828/apps/ledger-live-desktop/src/renderer/components/RecipientAddress.tsx#L111). So every time a user pastes or types an address, the transaction gets updated with that recipient.

Also, if we don't pass the **recipient** to the transaction, we can't trigger the error `This is not a valid "Currency" address`, because `transaction.recipient` is used by `getTransactionStatus` to generate that error:
For example, see [EVM](https://github.com/LedgerHQ/ledger-live/blob/e28c3ab39d6920878105a3b5b5b9e1f7cd2f8828/libs/coin-modules/coin-evm/src/getTransactionStatus.ts#L68) or [Hedera](https://github.com/LedgerHQ/ledger-live/blob/e28c3ab39d6920878105a3b5b5b9e1f7cd2f8828/libs/coin-modules/coin-hedera/src/bridge/getTransactionStatus.ts#L29).

So in reality, we've always passed the recipient (even if it's not the right currency scheme) because it's necessary for the user to understand why the "Continue" button is disabled.

One last thing: when the currency changes, a new transaction object is created. That's why I placed this logic at this level.

* We want to trigger `onChangeTransaction` not only when the currency changes, but also when the account to debit changes.

* We can’t remove `onChangeTransaction` from `onChange`, because we need to update the transaction recipient whenever the user types in an address.

Justkant: > We want to trigger `onChangeTransaction` not only when the currency changes, but also when the account to debit changes.

That was my understanding

> We can’t remove `onChangeTransaction` from `onChange`, because we need to update the transaction recipient whenever the user types in an address. 

It should be possible because we call `setValue` and value is used in the `useEffect` so if we just add `value` in the deps of the `useEffect` it should just work too

> In practice, `maybeExtra` is always `undefined` (except when scanning an address via QR code — and even then, it's not guaranteed).

Too bad we can't just remove the `maybeExtra`

Justkant (approved):

PR #10793, opened 2025-07-03
fix: Sui alpaca integration

estrauser-ledger:
### 📝 Description

To allow Aplaca to integrate with the SUI coin-module, a slight change in the config is needed. It's now the same as other alpaca coins like Tron or Tezos.


tools/actions/turbo-affected/build/main.js
jprudent, line 1: we should revert this file

Canestin (approved):

PR #10823, opened 2025-07-04
fix: [xrp] getbalance returns early if new account + fix integ tests

Wozacosta:
### ✅ Checklist

  - ...

### 📝 Description

Previously, the getBalance function always computed a locked balance based on XRP reserve rules — even for accounts that were inactive (not yet created on-chain). This led to confusing or incorrect results in cases like:

`{ value: 0n, locked: 1_000_000n }`

when the account had no transaction history and wasn't funded.

#### What the PR Does

- Adds a check in `getBalance.ts`:

If `accountInfo.isNewAccount` is `true` (account doesn't exist):

Return `value: 0n, locked: 0n`

- Updates the integration test:

Adjusts the expected output to include `locked: 0n` explicitly

Fixes the test asserting `result.length` to match actual behavior

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/coin-modules/coin-xrp/src/logic/getBalance.test.ts
+++ b/libs/coin-modules/coin-xrp/src/logic/getBalance.test.ts
@@ -40,4 +40,79 @@ describe("getBalance", () => {
     expect(mockGetAccountInfo.mock.lastCall[0]).toEqual(address);
     expect(result).toEqual([{ value: balance, asset: { type: "native" }, locked: 23000000n }]);
   });
+
+  it("returns 0 balance and no locked amount for a new (inactive) account", async () => {
+    mockGetAccountInfo.mockResolvedValue({
+      isNewAccount: true,
+      balance: "0",
+      ownerCount: 0,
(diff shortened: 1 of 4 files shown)


libs/coin-modules/coin-xrp/src/logic/getBalance.test.ts
```diff
@@ -40,4 +40,79 @@ describe("getBalance", () => {
...
+      sequence: 5,
+    });
+
+    const result = await getBalance("ZERO_BALANCE_ACTIVE_ACCOUNT");
+
+    expect(result).toEqual([
+      {
+        value: 0n,
+        asset: { type: "native" },
+        locked: 10_000_000n,
```
Moustafa-Koterba, line 114: wdyt to put a calculation instead ?
```suggestion
        locked: (server.info.validated_ledger.reserve_base_xrp + (server.info.validated_ledger.reserve_base_xrp * account.ownerCount)) * 100_000n,
```

PR #10838, opened 2025-07-07
feat(BACK-9222): add `getBlock` method to coin framework + SUI implementation

jnicoulaud-ledger:
### ✅ Checklist

  - only alpaca impact

### 📝 Description

- This PR adds a `getBlock`/`getBlockInfo` methods to ALPACA API
- It returns the block metadata + full list of transactions
- It will be used by our backends systems (A4) to perform live block indexation
- Note: a different model is used than `listOperations`, due to:
   - `listOperations` is specialized to a single account/address
   - `listOperations` flattens transactions with multiple transactions (backend would need to group them again)
   - `listOperations` includes block info in each operation, leading to high volume unnecessarily for `getBlock`

Tests are passing locally, added them as commented out to align with current state of SUI module.

### ❓ Context

- https://ledgerhq.atlassian.net/browse/BACK-9222

---

### DIFF
--- a/libs/coin-modules/coin-sui/src/network/sdk.test.ts
+++ b/libs/coin-modules/coin-sui/src/network/sdk.test.ts
@@ -1328,12 +1328,51 @@ describe("filterOperations", () => {
       ]);
     });
 
-    test("toBlockOperation should ignore transfers from unsupported owner types", () => {
+    test("toBlockOperation should ignore transfers from shared owner", () => {
       expect(
         sdk.toBlockOperation({
           owner: {
             Shared: {
               initial_shared_version: "0",
+            },
+          },
+          coinType: sdk.DEFAULT_COIN_TYPE,
+          amount: "-10000000000",
+        }),
+      ).toEqual([]);
+    });
+
+    test("toBlockOperation should ignore transfers from object owner", () => {
+      expect(
+        sdk.toBlockOperation({
+          owner: {
+            ObjectOwner: "test",
+          },
+          coinType: sdk.DEFAULT_COIN_TYPE,
+          amount: "-10000000000",
+        }),
+      ).toEqual([]);
+    });
+
+    test("toBlockOperation should ignore transfers from immutable owner", () => {
+      expect(
+        sdk.toBlockOperation({
+          owner: "Immutable",
+          coinType: sdk.DEFAULT_COIN_TYPE,
+          amount: "-10000000000",
+        }),
+      ).toEqual([]);
+    });
+
+    test("toBlockOperation should ignore transfers from consensus owner", () => {
+      expect(
+        sdk.toBlockOperation({
+          owner: {
+            ConsensusV2: {
+              authenticator: {
+                SingleOwner: "test",
+              },
+              start_version: "1",
             },
           },
           coinType: sdk.DEFAULT_COIN_TYPE,


libs/coin-framework/src/api/types.ts
```diff
@@ -163,6 +245,8 @@ export type AlpacaApi<
   ) => Promise<string>;
   getBalance: (address: string) => Promise<Balance<AssetInfo>[]>;
   lastBlock: () => Promise<BlockInfo>;
```
sprohaszka-ledger, line 247: I think for consistency we need to rename this function: `lastBlockInfo`

libs/coin-modules/coin-sui/src/api/index.integration.test.ts
```diff
@@ -101,4 +101,26 @@ describe("sui api integration test", () => {
 //       expect(result.time).toBeInstanceOf(Date);
 //     });
 //   });
+
+// describe("getBlockInfo", () => {
```
sprohaszka-ledger, line 105: Maybe we can remove them if they are unused.

libs/coin-modules/coin-sui/src/logic/index.ts
```diff
@@ -4,4 +4,5 @@ export { broadcast } from "./broadcast";
 export { combine } from "./combine";
 export { getBalance } from "./getBalance";
 export { lastBlock } from "./lastBlock";
+export { getBlock, getBlockInfo } from "./getBlock";
```
sprohaszka-ledger, line 7: A little nitpicking, but we could regroup all "block" functions (`lastBlock`, `getblock` and `getBlockInfo`) into one file in the logic directory. Let say: `block.ts`

libs/coin-framework/src/api/types.ts
```diff
@@ -57,7 +58,88 @@ export type Transaction = {
...
+/** A asset transfer that occurred in a {@link BlockTransaction}. */
+export type TransferBlockOperation<AssetType extends Asset<TokenInfoCommon>> = {
+  /** Operation type discriminator. */
+  type: "transfer";
+
+  /** The impacted address (can be sender or recipient based on signum of <code>amount</code>}. */
+  address: string;
+
+  /** The peer participant in the transfer (optional as it may be not known). */
+  peer?: string;
```
sprohaszka-ledger, line 121: I don't see it used right now.
It is not clear for me the purpose of this field, but if we can add it later...

libs/coin-framework/src/api/types.ts
```diff
@@ -57,7 +58,88 @@ export type Transaction = {
...
+   */
+  amount: bigint;
+};
+
+/**
+ * An unclassified type of operation that occurred in a {@link BlockTransaction}.
+ *
+ * Implementations are free to partially/completely omit this kind of operations.
+ */
+export type OtherBlockOperation = {
```
sprohaszka-ledger, line 138: I am not fond of leaving an "unknown" type :/ 
What could be its usage?

sprohaszka-ledger (commented):
Maybe we can remove all the commented code?

Also, we agree the new type (BlockOperation and BlockTransaction) could be transient and be replaced with a unique Operation and Transaction type (used both by "block" functions and listTransactions function)?

libs/coin-framework/src/api/types.ts
jnicoulaud-ledger: Agree in principle, but this would be a breaking change

libs/coin-modules/coin-sui/src/api/index.integration.test.ts
jnicoulaud-ledger: They are not, I only commented them out because the RPC was disabled because of the incident. All tests should be enabled again when situation is resolved

libs/coin-framework/src/api/types.ts
jnicoulaud-ledger: It's sender/recipient. Unfortunately on SUI, the easiest way is to generate operations from balance changes => we don't have peer info. To have full info, we would need to parse the transaction effects and it becomes a lot more complicated to maintain

libs/coin-framework/src/api/types.ts
jnicoulaud-ledger: - Debugging
- Display
- Being able to support network specific features without changing the API, and consolidate afterwards

libs/coin-modules/coin-sui/src/logic/index.ts
jnicoulaud-ledger: Yes sure, I'll do it in separate PR though, there are many concurrent PRs on SUI coin module right now, conflict resolution is difficult

jnicoulaud-ledger:
> Also, we agree the new type (BlockOperation and BlockTransaction) could be transient and be replaced with a unique Operation and Transaction type (used both by "block" functions and listTransactions function)?

@sprohaszka-ledger I am not sure to be honest, for instance the `listOperations` type contains the block info => we don't want to repeat it for every transaction in `getBlock`

libs/coin-framework/src/api/types.ts
jnicoulaud-ledger: Note: I will do it in a separate PR after SUi coin module situation has stabilized and tests are enabled again

libs/coin-modules/coin-sui/src/logic/index.ts
jnicoulaud-ledger: To be done in followup PR

libs/coin-modules/coin-sui/src/api/index.integration.test.ts
francois-guerin-ledger: Maybe `describe.skip` is more appropriate then ?

libs/coin-modules/coin-sui/src/network/sdk.ts
```diff
@@ -260,6 +278,83 @@ function transactionToOp(address: string, transaction: SuiTransactionBlockRespon
   };
 }
 
+/**
+ * Convert a SUI RPC checkpoint info to a {@link BlockInfo}.
+ *
+ * @param checkpoint SUI RPC checkpoint info
+ */
+export function suiCheckpointToBlockInfo(checkpoint: Checkpoint): BlockInfo {
```
Moustafa-Koterba: optional wdyt (same for other function)
```suggestion
export function toBlockInfo(checkpoint: Checkpoint): BlockInfo {
```

libs/coin-modules/coin-sui/src/network/sdk.test.ts
Moustafa-Koterba, line 1: you did not check this two cases:
```ts
if (typeof change.owner == "string" || !("AddressOwner" in change.owner)) return [];
```
for `balanceChangeToBlockOperation`

libs/coin-modules/coin-sui/src/network/sdk.ts
jnicoulaud-ledger: ok, changed it all

libs/coin-modules/coin-sui/src/network/sdk.test.ts
jnicoulaud-ledger: ok thanks, test added

PR #10839, opened 2025-07-07
chore: toggle sentry initialization based on llmSentry

themooneer:
### ✅ Checklist

  - ...

### 📝 Description

While we are monitoring Datdog on 20% of our users. We need to prepare a sentry switch off when satisfied with the datadog remoted data and error tracking. 

`llmSentry` feature flag is switched on by default so that we continue to run Sentry for this moment. 

### ❓ Context

- **JIRA or GitHub link**: 
https://ledgerhq.atlassian.net/browse/LIVE-19811

---


apps/ledger-live-mobile/src/datadog.ts
```diff
@@ -82,13 +82,18 @@ const isExcludedErrorDescription = (errorDescription: string): boolean => {
  * @param event The RUMErrorEvent object.
  * @returns The modified RUMErrorEvent object, or null if the event should be dropped.
  */
-export const customErrorEventMapper: ErrorEventMapper = event => {
-  if (isExcludedErrorName(event.stacktrace) || isExcludedErrorDescription(event.message)) {
-    return null; // Return null to drop the event
-  }
+export const customErrorEventMapper: (disableErrorTracking: boolean) => ErrorEventMapper =
```
mcayuelas-ledger, line 85:
```suggestion
export const customErrorEventMapper = (
  disableErrorTracking: boolean
): ErrorEventMapper => event => {
  const shouldDropEvent =
    disableErrorTracking ||
    isExcludedErrorName(event.stacktrace) ||
    isExcludedErrorDescription(event.message);

  return shouldDropEvent ? null : event;
};
```

I would have wrote something like this, It's more readable I think ? WDYT?

themooneer: maybe if the use of the `shouldDropEvent` will be in the business logic of the entire function, but for this case, it's a free variable reference in the memory for each and everytime the function is called , so why not a oneshot return. Anyways, the error mapper function is well documented to understand that null return will not send the error.

mcayuelas-ledger (approved):

PR #10874, opened 2025-07-09
fix: truncated Hedera addresses in LLD

md-blockydevs:
### ✅ Checklist

  - Hedera address should be fully visible in operation details drawer

### 📝 Description

This PR adjusts the layout fix introduced in [this commit](https://github.com/LedgerHQ/ledger-live/commit/2cd31146db355cb08dafabee809135b5561c7351#diff-4b25388664652813745e23bb81f5513bd7952a8590d998a10f4a0618b3f9e910) to address data overflow in the operation details screen. The issue with the current implementation is that the container with left part of the address has a width of 0px (at least with Hedera operations), making the prefix invisible. As a result, values like `0.0.98` incorrectly appear as `0.98`:

image

To solve this, I explored how the layout behaves without any `max-width` constrain with long values:

Pasted Graphic 6

Current behavior on `develop` (notice 0.3 instead of 0.0.3):

Pasted Graphic 7
Pasted Graphic 8

Behavior after this PR:

324 02382 HBAR
324 92382 HBAR

Ticket: [LIVE-20245](https://ledgerhq.atlassian.net/browse/LIVE-20245)
---

[LIVE-20245]: https://ledgerhq.atlassian.net/browse/LIVE-20245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/OperationsList/AddressCell.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/OperationsList/AddressCell.tsx
@@ -51,7 +51,7 @@ export const Address = ({ value }: { value: string }) => (
 );
 const Left = styled.div`
   overflow: hidden;
-  max-width: calc(100% - 50px);
+  max-width: calc(100% - 20px);
   white-space: nowrap;
   font-kerning: none;
   letter-spacing: 0px;
--- a/.changeset/dull-peaches-exist.md
+++ b/.changeset/dull-peaches-exist.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": minor
+---
+
+LLD - fixed truncated address in transaction details


apps/ledger-live-desktop/src/renderer/components/OperationsList/AddressCell.tsx
```diff
@@ -51,7 +51,8 @@ export const Address = ({ value }: { value: string }) => (
 );
 const Left = styled.div`
   overflow: hidden;
-  max-width: calc(100% - 50px);
+  max-width: 100%;
+  flex-shrink: 1;
```
may01: Do we need `flex-shrink: 1;`
For some coins (ex. Aptos) it lead to cutting the symbols that looks like a glitch:
Screenshot 2025-07-10 at 10 39 10

md-blockydevs: Ahh, in that case it's a no-go, I will try to get around it and find another way to fix Hedera without glitching other coins. I'll tag you here once pushed

md-blockydevs: @may01 Ok, rebased. This is how it looks now with mocked aptos addresses:

image

Transaction ID is unfortunately already getting cut on `develop`. I've just lowered max-width calc from `50px` to `20px`, matching the padding:

Memo

PR #10950, opened 2025-07-15
:sparkles: (LLM-UI) search component

LucasWerey:
### ✅ Checklist

  - native ui

### 📝 Description

Add useDebouncedCallback hook and search component to pre ldls UI for RN
It's the same as LLD but with rn event

image

https://github.com/user-attachments/assets/e16f9ec3-53ec-453c-9574-067b5b2eb964

image

image

### ❓ Context

- **JIRA or GitHub link**: [LIVE-19655](https://ledgerhq.atlassian.net/browse/LIVE-19655)

---

[LIVE-19655]: https://ledgerhq.atlassian.net/browse/LIVE-19655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui/packages/native/src/pre-ldls/hooks/useDebouncedCallback.ts
+++ b/libs/ui/packages/native/src/pre-ldls/hooks/useDebouncedCallback.ts
@@ -1,0 +1,22 @@
+import { useEffect, useRef, useState } from "react";
+
+export function useDebouncedCallback<T extends unknown[]>(
+  callback: ((...args: T) => void) | undefined,
+  delay?: number,
+) {
+  const [debouncedCallback, setDebouncedCallback] = useState<(...args: T) => void>();
+  const timeout = useRef<ReturnType<typeof setTimeout>>();
+
+  useEffect(() => {
+    if (!callback) return setDebouncedCallback(undefined);
+
+    setDebouncedCallback(() => (...args: T) => {
+      clearTimeout(timeout.current);
+      timeout.current = setTimeout(() => callback(...args), delay);
+    });
+
+    return () => clearTimeout(timeout.current);
+  }, [callback, delay]);
+
+  return debouncedCallback;
+}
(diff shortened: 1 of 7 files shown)


libs/ui/packages/native/src/pre-ldls/hooks/useDebouncedCallback.ts
marcotoniut-ledger, line 1: To keep the mental model consistent with React’s useCallback, could we adopt the same parameter order and semantics?

```
export function useDebouncedCallback<T extends unknown[]>(
  callback: ((...args: T) => void) | undefined,
  deps: unknown[],
  delay?: number,
) {

// put callback in a ref and avoid retriggering useEffect unless deps change.
```

Use:

```
useDebouncedCallback(() => {
}, deps, delay?);
```

cgrellard-ledger (dismissed):

libs/ui/packages/native/src/pre-ldls/hooks/useDebouncedCallback.ts
LucasWerey: Not convinced as we'll need to send deps every time, but the useEffect will always depend on the same deps

marcotoniut-ledger: As you prefer, though this puts the onus on the developer to make sure they memoise their callback. Perhaps we  should call it `useCallbackDebounce`, instead?

LucasWerey: We don't really need to memoize the function, actually, it's up to the dev. And for the name, I'm iso with LLD so I think we can stay on the same naming (I don't think we will use it anywhere except inside the DS, as we already have it on each platform-

PR #10952, opened 2025-07-15
feature/LIVE-20311 fix account for e2e swap test

Moustafa-Koterba:
### ✅ Checklist

- [] **Impact of the changes:** 
  - No impact

### 📝 Description

fix account for e2e swap test

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-20311?atlOrigin=eyJpIjoiNTExZWJkNmQ2MjMxNDZkNGIxYmJmY2E4ZWJiOTc0YWQiLCJwIjoiaiJ9

---


e2e/desktop/tests/page/index.ts
```diff
@@ -63,4 +63,13 @@ export class Application extends PageHolder {
   public liveApp = new LiveApp(this.page);
   public buyAndSell = new BuyAndSellPage(this.page, this.electronApp);
   public redux = new Redux(this.page);
+
+  public async overrideNetworkCallPayload(url: string, payloadHandler: (json: any) => any) {
```
ypolishchuk-ledger: can you please explain why this needs to be done? just using sanctioned address app json is not enough? also can you please move it to utils class?

Moustafa-Koterba: > can you please explain why this needs to be done? just using sanctioned address app json is not enough?

The test was failing because the app.json contained an account that wasn’t linked to the device seed, which caused a mismatch during the swap flow. To avoid further investigation into the seed/account linkage. I’ve re-written the test to mock the relevant network call, which should make it more reliable

> also can you please move it to utils class?

done on this [commit](https://github.com/LedgerHQ/ledger-live/pull/10952/commits/c8c6c187a25aec6b5bce80699aaa8247aa93e518)

ypolishchuk-ledger: thanks for explanation, it's clear now, one more question - does it mean we don't need app json anymore?

Moustafa-Koterba: We still need it, but with regular accounts (same used with other tests)

ypolishchuk-ledger: we don't need app json for regular accounts, we use CLI to create them on the fly, here is the example of swap e2e tests, we just load app json to skip onboarding and then create accounts: https://github.com/LedgerHQ/ledger-live/blob/develop/e2e/desktop/tests/specs/swap.send.spec.ts#L174

Moustafa-Koterba: you absolutely right indeed, I removed the file and setup the accounts with the CLI, see [commit](https://github.com/LedgerHQ/ledger-live/pull/10952/commits/a0fb0b0d38948e97c8d550cebd2fdb1cb1ed7056)

ypolishchuk-ledger (approved):

PR #10990, opened 2025-07-17
build: upgrade build target for android with corresponding required fixes

ysitbon:
### ✅ Checklist

- [x] **Impact of the changes:**  Upgrade Android build target. 

### 📝 Description

Upgrade Android build target from `34` to `35`. The upgrade creates a regression on Android when interacting with text inputs and the keyboard is displayed. This upgrade includes a fix for these regressions.

Tested manually on:
- Android 15
- Android 14  to ensure regression fixes do not impact lower versions.
- iOS 18 to ensure regression fixes do not impact iOS versions.

Should be monitored carefully since an SDK upgrade can cause many sneaky, "hard-to-find" or invisible modifications.  

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-20066

---

- **The PR description documents the changes** made and explains any technical trade-offs or design decisions.

### DIFF
--- a/apps/ledger-live-mobile/src/components/KeyboardView.tsx
+++ b/apps/ledger-live-mobile/src/components/KeyboardView.tsx
@@ -20,19 +20,24 @@ const KeyboardView = React.memo<Props>(
   }: Props) => {
     const isExperimental = useExperimental();
     const headerHeight = React.useContext(HeaderHeightContext) || 0;
-    const keyboardVerticalOffset = isExperimental || Config.DETOX ? ExperimentalHeaderHeight : 0;
+    const isAndroid35 = Platform.OS === "android" && Platform.Version >= 35;
+    const experimentalHeaderHeight = isExperimental || Config.DETOX ? ExperimentalHeaderHeight : 0;
+    const keyboardVerticalOffset = isAndroid35
+      ? headerHeight + experimentalHeaderHeight
+      : headerHeight + (StatusBar.currentHeight || 0) + experimentalHeaderHeight;
     const behaviorParam = behavior ?? "height";
-    const behaviorProp = Platform.select({
+    const behaviorProp: KeyboardAvoidingViewProps["behavior"] = Platform.select({
       ios: behaviorParam,
-      android: undefined,
+      // On Android 34 or below, we leave it undefined to use the default behavior
+      // otherwise it can cause the keyboard avoiding to re-render infinitely in
+      // some cases like going back from a modal.
+      android: isAndroid35 ? "height" : undefined,
     });
 
     return (
       <KeyboardAvoidingView
-        style={style}
-        keyboardVerticalOffset={
-          headerHeight + (StatusBar.currentHeight || 0) + keyboardVerticalOffset
-        }
+        style={[style]}
+        keyboardVerticalOffset={keyboardVerticalOffset}
         behavior={behaviorProp}
         enabled
       >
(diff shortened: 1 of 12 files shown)


themooneer (approved):

mcayuelas-ledger:
I saw that we have a patch` /patches/react-native-fast-pbkdf2@0.3.1.patch` using v34

ysitbon:
@KVNLS  You're right to check for consistency! However, it's not a critical issue because:
1. `AndroidManifest.xml` - The `targetSdkVersion="33"` (which was already obsolete) is ignored by the build system. As you can see in the build [logs](https://github.com/LedgerHQ/ledger-live/actions/runs/16346956678/job/46183225056?pr=10990#step:11:1972), it shows: 
    ``` 
    uses-sdk:targetSdkVersion value (33) specified in the manifest file is ignored. 
    It is overridden by the value declared in the DSL or the variant API. Current value is (35).
    ```
2. `react-native-hid` library - It uses `safeExtGet("targetSdkVersion", 29)` which automatically inherits the target from the main project's Gradle configuration (35). The fallback value of 29 is only used if the main project doesn't define targetSdkVersion.

@mcayuelas-ledger  Good catch on the patch! While it doesn't cause build issues (as evidenced by the successful build), updating it to API 35 would be better for consistency across the entire codebase.

**Bottom line:** The Gradle configuration in the main project is what matters for Play Store compliance. The other locations are either overridden or inherit the correct values, but updating them for consistency is good practice. 

I updated `AndroidManifest.xml` and the patch for this PR, as the library, I would prefer the direct maintainers to do it.

apps/ledger-live-mobile/src/screens/Platform/v2/Catalog/Layout.tsx
```diff
@@ -74,32 +73,28 @@ export function Layout({
...
-          <Flex style={[styles.marginHorizontal]}>{bodyContent}</Flex>
-        </ScrollContainer>
-      </KeyboardView>
+      <ScrollContainer
+        onScroll={scrollHandler}
+        scrollEventThrottle={10}
+        showsVerticalScrollIndicator={false}
+        stickyHeaderIndices={listStickyElement}
+      >
+        {titleHeaderContent && <Flex style={[styles.marginHorizontal]}>{titleHeaderContent}</Flex>}
```
mcayuelas-ledger, line 82: indentation issue no?

ysitbon: Mmmh :thinking:, neither eslint, neither prettier warn me or re-format it... 
It reaches 100 chars, which I guess is our column limit, right?

apps/ledger-live-mobile/src/components/KeyboardView.tsx
```diff
@@ -20,19 +20,24 @@ const KeyboardView = React.memo<Props>(
   }: Props) => {
     const isExperimental = useExperimental();
     const headerHeight = React.useContext(HeaderHeightContext) || 0;
-    const keyboardVerticalOffset = isExperimental || Config.DETOX ? ExperimentalHeaderHeight : 0;
+    const isAndroid35 = Platform.OS === "android" && Platform.Version >= 35;
```
mcayuelas-ledger, line 23: as you use it twice and we may need to use it somewhere else, maybe we can extract this?

ysitbon: We could, yes, with a version parameter since this check is often used in the codebase. What I'm afraid of is adding a utility function, used only once, because this utility function will not be obvious to know/find. What are your thoughts about it?

mcayuelas-ledger (approved):

KVNLS (approved):
Great work!

mbertin-ledger (approved):

Moustafa-Koterba (approved):

PR #11083, opened 2025-07-25
fix: check ff before openning drawer

sergiubreban:
Make Earn entry points use the new drawer based on the feature flag value (`lldModularDrawer`)

https://github.com/user-attachments/assets/d3717965-ce11-4ce7-827e-5b9a48784605

### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: [LIVE-20555](https://ledgerhq.atlassian.net/browse/LIVE-20555)

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/screens/stake/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/screens/stake/index.tsx
@@ -11,6 +11,11 @@ import { getDefaultAccountName } from "@ledgerhq/live-wallet/accountName";
 import { useStake } from "LLD/hooks/useStake";
 import { walletSelector } from "~/renderer/reducers/wallet";
 import { Account, AccountLike } from "@ledgerhq/types-live";
+import {
+  ModularDrawerLocation,
+  openAssetAndAccountDrawer,
+  useModularDrawerVisibility,
+} from "LLD/features/ModularDrawer";
 
 type StakeFlowProps = {
   currencies?: string[];
@@ -28,6 +33,12 @@ const useStakeFlow = () => {
   const walletState = useSelector(walletSelector);
   const { enabledCurrencies, partnerSupportedAssets, getRouteToPlatformApp } = useStake();
   const list = enabledCurrencies.concat(partnerSupportedAssets);
+
+  const { isModularDrawerVisible } = useModularDrawerVisibility({
+    modularDrawerFeatureFlagKey: "lldModularDrawer",
+  });
+
+  const modularDrawerVisible = isModularDrawerVisible(ModularDrawerLocation.LIVE_APP);
 
   const handleAccountSelected = useCallback(
     (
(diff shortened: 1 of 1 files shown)


Valentin-Ledger (approved):

Valentin-Ledger (dismissed):

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -108,6 +181,94 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"], account
...
+            alwaysShowNoFunds: false,
+            entryPoint: "get-funds",
+            source,
+            returnTo: "",
+          });
+          return { success: true };
+        }
+        throw new Error("Unknown navigation action");
+      },
+      "custom.getFunds": request => {
```
Valentin-Ledger, line 222: [Not a blocker]
Dont' we use an enum or a constant for this?

Valentin-Ledger (dismissed):

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
sergiubreban: I followed the current convention. Changing it to an enum will imply too many modifications in this PR and it will change the scope of the PR. We can consider doing it in a separate ticket

Justkant (dismissed):

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -40,6 +53,66 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"], account
...
+      const {
+        account,
+        parentAccount,
+        alwaysShowNoFunds,
+        entryPoint,
+        source,
+        shouldRedirect,
+        returnTo,
+      } = props;
+      const platformAppRoute = getRouteToPlatformApp(account, walletState, parentAccount);
```
beths-ledger:
```suggestion
      const platformAppRoute = getRouteToPlatformApp(account, walletState, parentAccount, returnTo);
```

If you pass `returnTo` to the fn then it will return that, but if it's undefined then it will return a `returnTo` param that navigates to the account or token account page for the account in question. This is especially good for token accounts.

Then you won't need to keep adding `returnTo` from params below - it'll be taken care of by the returned `platformAppRoute.state`.

apps/ledger-live-desktop/src/renderer/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -108,6 +181,94 @@ export function usePTXCustomHandlers(manifest: WebviewProps["manifest"], account
           },
         },
       }),
+      "custom.navigate": async request => {
+        const { action, queryParams, currencyId, accountId } = request.params || {};
```
beths-ledger: why is there a nested `queryParams` object inside this action? If any of these query params are needed, could they instead just be flattened onto the same level?

sergiubreban: I wanted to add a flexible way that does not require strict typing, in case we need something in the future, to be able to do it without a ledger-live release. I don't exactly know why we have queryParams attribute inside the feature flag StakeProgram. But basically, if we need to use those from the FF value and send them back to LL, we can send them here using the queryParams attribute

beths-ledger: ahh so these are the query params from the feature flag that we add to the customDappUrl and spread in to send to earn, right?

beths-ledger (approved):
Nice one

francois-guerin-ledger (changes requested):
We should have ledger live mobile in the changeset

francois-guerin-ledger (approved):

libs/ledger-live-common/src/wallet-api/helpers.ts
```diff
@@ -127,3 +127,19 @@ export const stripHexPrefix = (str: string): string => {
...
+
+export function objectToURLSearchParams(obj: Record<string, unknown>): URLSearchParams {
+  const searchParams = new URLSearchParams();
+
+  Object.entries(obj).forEach(([key, value]) => {
+    if (value !== undefined && value !== null) {
+      if (typeof value === "object") {
+        searchParams.append(key, JSON.stringify(value));
+      } else {
+        searchParams.append(key, String(value));
```
beths-ledger, line 139: 'value' will use Object's default stringification format ('[object Object]') when stringified.

PR #11120, opened 2025-07-28
fix(BACK-9407): [coin-modules][stellar] fix memo details lost

jnicoulaud-ledger:
### ✅ Checklist

  - stellar coin module
  - stellar operation details UI LLM+LLD

### 📝 Description

Memo details for stellar were lost during memo refactoring, but LES requires it

### ❓ Context

- https://ledgerhq.atlassian.net/browse/BACK-9407

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/stellar/operationDetails.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/stellar/operationDetails.tsx
@@ -11,20 +11,11 @@ import { SplitAddress } from "~/renderer/components/OperationsList/AddressCell";
 import { Account } from "@ledgerhq/types-live";
 import { OperationDetailsExtraProps } from "../types";
 import { StellarOperation } from "@ledgerhq/live-common/families/stellar/types";
+import { formatMemo } from "@ledgerhq/live-common/families/stellar/ui";
 
 const OperationDetailsExtra = ({
   operation: { extra },
 }: OperationDetailsExtraProps<Account, StellarOperation>) => {
-  let memo = undefined;
-  switch (extra.memo?.type) {
-    case "MEMO_ID":
-    case "MEMO_TEXT":
-    case "MEMO_HASH":
-    case "MEMO_RETURN":
-      memo = extra.memo.value;
-      break;
-  }
-
   return (
     <>
       {Object.keys(extra).map(key => {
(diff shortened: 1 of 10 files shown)


apps/ledger-live-desktop/src/renderer/families/stellar/operationDetails.tsx
```diff
@@ -15,6 +15,16 @@ import { StellarOperation } from "@ledgerhq/live-common/families/stellar/types";
...
 }: OperationDetailsExtraProps<Account, StellarOperation>) => {
+  let memo = undefined;
+  switch (extra.memo?.type) {
+    case "MEMO_ID":
+    case "MEMO_TEXT":
+    case "MEMO_HASH":
+    case "MEMO_RETURN":
+      memo = extra.memo.value;
+      break;
+  }
```
Justkant: We should probably share the logic between desktop and mobile in a `getMemo` util or something similar in common, but not a big deal if we keep it like this

libs/coin-modules/coin-stellar/src/network/serialization.ts
```diff
@@ -90,11 +90,39 @@ async function formatOperation(
...
-    ? transaction.memo_type === "hash" || transaction.memo_type === "return"
-      ? Buffer.from(transaction.memo, "base64").toString("hex")
-      : transaction.memo
-    : null;
+
+  let memo: StellarMemo | undefined = undefined;
+  switch (transaction.memo_type) {
+    case "none":
+      memo = { type: "NO_MEMO" };
+      break;
```
Justkant, line 98: Should we have a default case when `memo_type` is `undefined` or something not supported to also set `memo` to `{ type: "NO_MEMO" }` ?

apps/ledger-live-desktop/src/renderer/families/stellar/operationDetails.tsx
jnicoulaud-ledger: I am not super familiar, where would you put it ?

libs/coin-modules/coin-stellar/src/network/serialization.ts
jnicoulaud-ledger: This could cause problems on LES side, "memo absent" is different from "memo = NO_MEMO", HSM currently does not support memo so we need to differentiate it. With this implementation it defaults memo = undefined

apps/ledger-live-desktop/src/renderer/families/stellar/operationDetails.tsx
jnicoulaud-ledger: Can't really find a place that fits :sweat_smile:

jnicoulaud-ledger: Maybe I should create a new package in https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledger-live-common/src ?

jnicoulaud-ledger: I put it in `@ledgerhq/live-common/families/stellar/ui`

PR #11150, opened 2025-07-30
feat: swap intent wip

CremaFR:
### ✅ Checklist

  - ...

### 📝 Description

This PR introduces 2 changes:
 - SwapIntent logs to DD to know retries based on user account amount and provider (or without) to filter out retries on swap failures, and know if users manages to swap (not perfect but could tell us blocker errors vs non blocking ones)
 - Sending seedId to /cancel to know better about our main error. SeedId isn't really a seedId, no match doesn't mean different seed. (my Solana and Eth accounts are on the same seed but different seedId)

example of new payload 

```
{
  "provider": "lifi",
  "swapId": "0xb38f7ed27af26c8663082968a8078953659355fab7dcd8c40767b2bd74a6f46a",
  "swapIntentWithProvider": "0fd1bec72f44a265d91bc1c673546758c4af3d96bf9277f14177943f39a87101",
  "swapIntentWithoutProvider": "da624a7fa5aa499a6ecbc42f9aeea86b083956897c6de9c7babc49fbaa606d2a",
  "payloadAddressMatchAccountAddress": true,
  "fromAccountId": "js:2:ethereum:0xe197a5A43521Fc8f0842E63C232Dc2d616A7A308:",
  "toAccountId": "js:2:ethereum:0x4496974953B8643b274B364068AF9aF4FeBd5969:+0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
  "payloadRefundAddress": "0xe197a5A43521Fc8f0842E63C232Dc2d616A7A308",
  "payloadPayoutAddress": "0x4496974953B8643b274B364068AF9aF4FeBd5969",
  "maybeSeedMatch": false,
  "swapStep": "CHECK_REFUND_ADDRESS",
  "statusCode": "WrongDeviceForAccountRefund",
  "sourceCurrencyId": "ethereum",
  "targetCurrencyId": "ethereum",
  "hardwareWalletType": "europa",
  "swapType": "float",
  "refundAddress": "0xe197a5A43521Fc8f0842E63C232Dc2d616A7A308",
  "payoutAddress": "0x4496974953B8643b274B364068AF9aF4FeBd5969"
}
```

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/ledger-live-common/src/exchange/swap/postSwapState.ts
+++ b/libs/ledger-live-common/src/exchange/swap/postSwapState.ts
@@ -155,7 +155,7 @@ export const postSwapCancelled: PostSwapCancelled = async ({
       toAccountId: shouldIncludeAddresses ? toAccountId : undefined,
       payloadRefundAddress: shouldIncludeAddresses ? refundAddress : undefined,
       payloadPayoutAddress: shouldIncludeAddresses ? payoutAddress : undefined,
-      maybeSeedMatch: seedIdFrom === seedIdTo, // should only matters for EVM to EVM
+      maybeSeedMatch: seedIdFrom === seedIdTo, // Only true if both accounts are from the same seed and from the same chain type
       ...rest,
     };
 


libs/ledger-live-common/src/exchange/swap/postSwapState.ts
```diff
@@ -21,6 +25,36 @@ export const postSwapAccepted: PostSwapAccepted = async ({
   if (!swapId) {
     return null;
   }
+
+  // for example '2025-08-01' used to add a one day unique nonce to the swap intent hash
+  const currentday = new Date().toISOString().split("T")[0];
```
zaruen, line 20: random thought but should we try to make the day logic timezone-specific?
I don't know if we have a library for that like `moment` or `luxon` or something like that.
Might be an overkill but it might help get cleaner results, no?

zaruen: Also we should definitely use `luxon` for date formatting, comparison and all if we don't already. It's very convenient.
https://moment.github.io/luxon/#/

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
```diff
@@ -21,6 +25,36 @@ export const postSwapAccepted: PostSwapAccepted = async ({
...
+    .createHash("sha256")
+    .update(
+      JSON.stringify({
+        fromAccount,
+        toAccount,
+        amount,
+        currentday,
+      }),
+    )
+    .digest("hex");
```
zaruen: Should we extract that into an helper function since we need it in both places?

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
```diff
@@ -57,17 +116,42 @@ export const postSwapCancelled: PostSwapCancelled = async ({
...
       ...(swapAppVersion ? { "x-swap-app-version": swapAppVersion } : {}),
     };
 
+    const requestData = {
+      provider,
+      swapId,
+      swapIntentWithProvider,
+      swapIntentWithoutProvider,
+      payloadAddressMatchAccountAddress,
+      seedMatch: seedIdFrom === seedIdTo,
```
zaruen: Since the `seedIdentifier` is chain specific, we expect the `from` and `to` to be different for cross-chain swaps, right? Should we be more explicit in the naming then? `chainSeedMatch` or something like that?

libs/ledger-live-common/src/wallet-api/Exchange/server.ts
```diff
@@ -571,6 +571,9 @@ export const handlers = ({
               targetCurrencyId: toCurrency?.id,
               hardwareWalletType: deviceInfo?.modelId as DeviceModelId,
               swapAppVersion,
+              fromAccount: fromAccount.id,
+              toAccount: toAccount?.id!,
```
zaruen: Can we change to `fromAccountId` and `toAccountId` just for clarity 🙏

CremaFR: Yuupp nice catch, done

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
CremaFR: it's not really chain too. it's chain categories. This will be only used for a one time investigation on DD and hopefully scrapped later. I'd say this name doesn't really matter

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
CremaFR: we don't have moment nor luxon inside our app. There's no way I'm adding a npm package to use a simple toISOString to generate a single reusable nonce. The result will be the same as I don't expect a user to change his timezone before retrying a swap.

zaruen: > There's no way I'm adding a npm package to use a simple toISOString

If there is no other usage, maybe it's an overkill indeed. That was just in case we had other work around dates in other places.

> I don't expect a user to change his timezone before retrying a swap.

The idea was more for the logic to work for all timezones. As is, it would be accurate around GMT timezones. In New York for example, if you place a swap at 7:59pm and another one at 8:01pm, the day would have changed in UTC so the `swapIntent` hashes will be different.

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
zaruen: > This will be only used for a one time investigation on DD and hopefully scrapped later.

I agree but I still think the name is misleading and it probably doesn't have to be. Anyone reading `seedMatch=false` would think that the accounts come from 2 different seeds, which is not necessarily the case as you told me.

CremaFR: I can add a comment but no naming will tell us 
"it's indeed a seed check" and also "not a reliable one" 
In the past we used in LL the term "maybe" so maybeSeedMatch would that be ok?

zaruen (dismissed):

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
```diff
@@ -57,17 +116,51 @@ export const postSwapCancelled: PostSwapCancelled = async ({
...
+      provider,
+      swapId,
+      swapIntentWithProvider,
+      swapIntentWithoutProvider,
+      payloadAddressMatchAccountAddress,
+      fromAccountId: shouldIncludeAddresses ? fromAccountId : undefined,
+      toAccountId: shouldIncludeAddresses ? toAccountId : undefined,
+      payloadRefundAddress: shouldIncludeAddresses ? rest.refundAddress : undefined,
+      payloadPayoutAddress: shouldIncludeAddresses ? rest.payoutAddress : undefined,
+      maybeSeedMatch: seedIdFrom === seedIdTo, // should only matters for EVM to EVM
```
zaruen: Ah even SPL tokens seed id wouldn't match?

libs/ledger-live-common/src/exchange/swap/postSwapState.ts
```diff
@@ -57,17 +116,51 @@ export const postSwapCancelled: PostSwapCancelled = async ({
...
 
+  const { swapIntentWithProvider, swapIntentWithoutProvider } = createSwapIntentHashes({
+    provider,
+    fromAccountId,
+    toAccountId,
+    amount,
+  });
+
+  // Check if the refundAddress and payoutAddress match the account addresses, just to eliminate this supposition
+  // returns true if any value is missing to not send false in error cases where the values are not set
```
zaruen: Going to be annoying, but could you rephrase the comment a bit 🙏 
I found it a bit unclear for people that doesn't have any context.

zaruen: > returns true if any value is missing

Also, is that possible? An address can be missing?

zaruen (approved):

PR #11205, opened 2025-08-04
chore(deps): upgrade react-navigation from v6 to v7

ysitbon:
### ✅ Checklist

  - upgrading WS for e2e (necessary for new types)
  - patch update on many react-native libraries notably `react-native: 0.77.2 ->  0.77.3`
  - major update of `react-navigation` libraries so **all screens navigation are impacted**
  - drop `react-native-tab-view`
  - drop `@shopify/react-native-performance`
  - the Portfolio is no longer unmounted when switching tabs

### 📝 Description

We have migrated react-navigation dependency from v6 to v7. The migration guide can be found here: https://reactnavigation.org/docs/upgrading-from-6.x/

Essentially for us:
- we need to protect some goBack() by first checking `canGoBack()`
- sometimes it was necessary to overrides the implementation of `layout` to prevent the navigator to re-render the screen making it losing the state and breaking navigation location.
- tabBarTestID becomes tabBarButtonTestID
- we reintroduce concept of `UnmountOnBlur` which was previous behaviour of React Navigation. typically in our tabs based root navigation, we want some screens to unmount to release performance for other tabs.
- some `navigation.navigate` where changed to `navigation.pop` / `navigation.popTo`
- sometimes we need to explicit the tree of screens to preserve navigation
-  `popTo` usage in swap ( https://reactnavigation.org/docs/upgrading-from-6.x#the-navigate-method-no-longer-goes-back-use-popto-instead )

### @Mobile E2E Results

Workflow [run](https://github.com/LedgerHQ/ledger-live/actions/runs/17232230402)

- [android-e2e Allure report](https://ledger-live.allure.green.ledgerlabs.net/allure/reports/30cb3527-b6eb-49ae-84eb-a77beec72c18/)
- [ios-e2e Allure report](https://ledger-live.allure.green.ledgerlabs.net/allure/reports/74b0460b-a341-487f-acd7-21a1c58c4758/)

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-17445

---


apps/ledger-live-mobile/package.json
```diff
@@ -322,6 +319,6 @@
     "ts-jest": "29.1.2",
     "ts-node": "10.9.2",
     "tslib": "2.6.2",
-    "ws": "7.5.10"
+    "ws": "8.18.3"
```
gre-ledger, line 322: small feedback here: according to https://github.com/websockets/ws/releases/tag/8.0.0 , some small breaking change could happen like the onMessage may no longer always be a String, could it be a problem in mobile e2e implementation?

ysitbon: Just to keep track, we talked about that in DM and in this case it's only used by e2e tests and the communication actually work as expected so it should be fine for LLM.

themooneer (approved):

ypolishchuk-ledger (approved):

ofreyssinet-ledger (approved):

PR #11245, opened 2025-08-05
feat: optimize aptos queries (LIVE-20766)

semeano:
### ✅ Checklist

  - ...

### 📝 Description

- Remove unused queries
- Improve queries performance
- Add cache

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-20766

---

### DIFF
--- a/libs/coin-modules/coin-aptos/src/network/client.ts
+++ b/libs/coin-modules/coin-aptos/src/network/client.ts
@@ -309,9 +309,11 @@ export class AptosAPI {
     });
 
     return Promise.all(
-      queryResponse.data.account_transactions.map(({ transaction_version }) => {
-        return this.richItemByVersion(transaction_version);
-      }),
+      queryResponse.data.account_transactions
+        .sort((a, b) => b.transaction_version - a.transaction_version)
+        .map(({ transaction_version }) => {
+          return this.richItemByVersion(transaction_version);
+        }),
     );
   }
 
--- a/libs/coin-modules/coin-aptos/src/network/graphql/queries.ts
+++ b/libs/coin-modules/coin-aptos/src/network/graphql/queries.ts
@@ -2,11 +2,7 @@ import { gql } from "@apollo/client";
 
 export const GetAccountTransactionsData = gql`
   query GetAccountTransactionsData($address: String, $limit: Int) {
-    account_transactions(
-      where: { account_address: { _eq: $address } }
-      order_by: { transaction_version: desc }
-      limit: $limit
-    ) {
+    account_transactions(where: { account_address: { _eq: $address } }, limit: $limit) {
       transaction_version
       __typename
     }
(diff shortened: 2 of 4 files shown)


libs/coin-modules/coin-aptos/src/network/graphql/queries.ts
```diff
@@ -1,26 +1,8 @@
...
-      transaction_version
-    }
-  }
-`;
-
 export const GetAccountTransactionsData = gql`
   query GetAccountTransactionsData($address: String, $limit: Int) {
-    account_transactions(
-      where: { account_address: { _eq: $address } }
-      order_by: { transaction_version: desc }
```
pvoliveira, line 21: Why are we ordering the results locally instead of using the query? limitations?

libs/coin-modules/coin-aptos/src/__tests__/network/validators.test.ts
```diff
@@ -83,23 +83,31 @@ describe("getValidators", () => {
...
+    }));
     mockedApolloClient.mockImplementation(() => ({
-      query: async () => ({
-        data: {
-          current_delegator_balances,
-        },
-      }),
+      query: mockedGetValidatorsQuery,
     }));
+    const mockedGetValidatorsQuerySpy = jest.spyOn(
```
Moustafa-Koterba: I do not really get why you have a spy on a mock, can you explain this ?
Plus, I think you can do something like
```ts
expect(mockedGetValidatorsQuery.mock.calls.length).toBe(1);
```

pvoliveira (commented):
Check my comment

libs/coin-modules/coin-aptos/src/network/client.ts
```diff
@@ -309,9 +309,12 @@ export class AptosAPI {
     });
 
     return Promise.all(
-      queryResponse.data.account_transactions.map(({ transaction_version }) => {
-        return this.richItemByVersion(transaction_version);
-      }),
+      queryResponse.data.account_transactions
+        .slice()
```
Moustafa-Koterba, line 313: why do you need to slice it ?

libs/coin-modules/coin-aptos/src/network/graphql/queries.ts
semeano: By adding an `order_by` on the query will increase the cost to run the query on the node, leading to higher charges, and that's what we want to reduce as much as possible.

libs/coin-modules/coin-aptos/src/network/client.ts
semeano: `queryResponse.data.account_transactions` is a [frozen array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze). I'm using `slice` to create a copy to be able to sort it.

libs/coin-modules/coin-aptos/src/__tests__/network/validators.test.ts
semeano: You're right. I'll update that.

libs/coin-modules/coin-aptos/src/network/graphql/queries.ts
pvoliveira: As we are not ensuring the order at the source, we'll not be able to page the results properly, so we need to fetch all results to the local sort work correctly. If it is not included, the results from the queries may vary significantly.

pvoliveira (approved):

PR #11309, opened 2025-08-08
feat(coin-evm): implement `validateIntent`

francois-guerin-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Implement `validateIntent` for EVM.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/coin-modules/coin-evm/src/logic/common.ts
+++ b/libs/coin-modules/coin-evm/src/logic/common.ts
@@ -1,6 +1,58 @@
 import { ethers } from "ethers";
 import { TransactionTypes } from "ethers/lib/utils";
+import { FeeEstimation } from "@ledgerhq/coin-framework/api/types";
 import ERC20ABI from "../abis/erc20.abi.json";
+import { ApiFeeData, ApiGasOptions } from "../types";
+
+export function isApiGasOptions(options: unknown): options is ApiGasOptions {
+  if (!options || typeof options !== "object") return false;
+
+  return (
+    "slow" in options &&
+    isApiFeeData(options.slow) &&
+    "medium" in options &&
+    isApiFeeData(options.medium) &&
+    "fast" in options &&
+    isApiFeeData(options.fast)
+  );
+}
+
+export function isApiFeeData(fees: unknown): fees is ApiFeeData {
+  if (!fees || typeof fees !== "object") return false;
+
+  const isBigIntOrNull = (value: unknown): boolean => value === null || typeof value === "bigint";
+
+  return (
+    "maxFeePerGas" in fees &&
+    isBigIntOrNull(fees.maxFeePerGas) &&
+    "maxPriorityFeePerGas" in fees &&
+    isBigIntOrNull(fees.maxPriorityFeePerGas) &&
+    "gasPrice" in fees &&
+    isBigIntOrNull(fees.gasPrice) &&
+    "nextBaseFee" in fees &&
+    isBigIntOrNull(fees.nextBaseFee)
+  );
+}
+
+type LegacyFeeEstimation = FeeEstimation & { parameters: { gasPrice: bigint } };
+type Eip1559FeeEstimation = FeeEstimation & {
+  parameters: { maxFeePerGas: bigint; maxPriorityFeePerGas: bigint };
+};
+
+export function isLegacyFeeEstimation(
+  estimation: FeeEstimation,
+): estimation is LegacyFeeEstimation {
+  return typeof estimation.parameters?.gasPrice === "bigint";
+}
+
+export function isEip1559FeeEstimation(
+  estimation: FeeEstimation,
+): estimation is Eip1559FeeEstimation {
+  return (
+    typeof estimation.parameters?.maxFeePerGas === "bigint" &&
+    typeof estimation.parameters?.maxPriorityFeePerGas === "bigint"
+  );
+}
 
 export function getTransactionType(intentType: string): TransactionTypes {
   if (!["send-legacy", "send-eip1559"].includes(intentType)) {
--- a/libs/coin-modules/coin-evm/src/logic/validateIntent.ts
+++ b/libs/coin-modules/coin-evm/src/logic/validateIntent.ts
@@ -1,0 +1,274 @@
+import { ethers } from "ethers";
+import BigNumber from "bignumber.js";
+import type {
+  AssetInfo,
+  Balance,
+  FeeEstimation,
+  TransactionIntent,
+  TransactionValidation,
+} from "@ledgerhq/coin-framework/api/types";
+import {
+  AmountRequired,
+  ETHAddressNonEIP,
+  FeeNotLoaded,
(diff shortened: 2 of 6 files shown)


libs/coin-modules/coin-evm/src/logic/validateIntent.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  // eip1559 specific
+  if (transactionType === TransactionTypes.eip1559 && isEip1559FeeEstimation(estimatedFees)) {
+    const gasTracker = getGasTracker(currency);
+    const { maxFeePerGas, maxPriorityFeePerGas } = estimatedFees.parameters;
+    const gasOptions = await gasTracker?.getGasOptions({ currency, options: { useEIP1559: true } });
+    const maximalPriorityFee = BigInt(gasOptions?.fast?.maxPriorityFeePerGas?.toFixed(0) ?? "0");
+    const minimalPriorityFee = BigInt(gasOptions?.slow?.maxPriorityFeePerGas?.toFixed(0) ?? "0");
+    const recommandedNextBaseFee = BigInt(gasOptions?.medium?.nextBaseFee?.toFixed(0) ?? "0");
```
dilaouid: `recommendedNextBaseFee`*
Just a little typo nothing much

libs/coin-modules/coin-evm/src/logic/validateIntent.ts
```diff
@@ -0,0 +1,241 @@
...
+      cryptoName: currency.name,
+      links: ["ledgerlive://buy"],
+    });
+  }
+
+  // eip1559 specific
+  if (transactionType === TransactionTypes.eip1559 && isEip1559FeeEstimation(estimatedFees)) {
+    const gasTracker = getGasTracker(currency);
+    const { maxFeePerGas, maxPriorityFeePerGas } = estimatedFees.parameters;
+    const gasOptions = await gasTracker?.getGasOptions({ currency, options: { useEIP1559: true } });
```
dilaouid: As you releved in slack, this may be a perf concern: calling `gasTracker.getGasOptions()` on every validation can be expensive? Or not, idk ?
Consider adding an optional `gasOptions` parameter to avoid repeated network calls when validating multiple intents with the same gas conditions (like we do in the manual tests for ex)
Something like:

```ts
export async function validateIntent(
  currency: CryptoCurrency,
  intent: TransactionIntent,
  customFees?: FeeEstimation,
  gasOptions?: GasOptions, // This one, also consider import it
): Promise<TransactionValidation>
```

Then use provided `gasOptions` or fetch as fallback

dilaouid: (you said that gasOptions is not possible to implement here, but i don't understand why)

francois-guerin-ledger: > As you releved in slack, this may be a perf concern: calling gasTracker.getGasOptions() on every validation can be expensive? Or not, idk ?

Technically speaking it is just one HTTP GET call, and the result payload is pretty small (cf example below)

```json
{"low":"30529409","medium":"130286067","high":"2199353445","next_base":"216696957"}
```

But `validateIntent` can be called quite often during the transaction flow (maybe after each keystroke 😬) so better avoiding it.

> (you said that gasOptions is not possible to implement here, but i don't understand why)

I said it is not possible to have it in the intent (without asking cloud wallet their consent for this small addition). I think having `gasOptions` as a key of `customFees.parameters` is legit. Remember the signature of `validateIntent` in the `coin-framework`

```ts
  validateIntent: (
    transactionIntent: TransactionIntent<MemoType>,
    customFees?: FeeEstimation,
  ) => Promise<TransactionValidation>;
```

libs/coin-modules/coin-evm/src/logic/validateIntent.ts
francois-guerin-ledger: Good catch

libs/coin-modules/coin-evm/src/logic/validateIntent.ts
dilaouid: OK you're right sorry i went so fast
But maybe as you said, using a key of `customFees.parameters` may be a good solution:
```ts
const providedGasOptions = estimatedFees.parameters?.gasOptions as GasOptions | undefined;
const gasOptions = providedGasOptions || 
  await gasTracker?.getGasOptions({ currency, options: { useEIP1559: true } });
```

Less http requests that may give the same results is always better, i think, but if it's mandatory because of the function signature or issue with the fw, c'est la vie

libs/coin-modules/coin-evm/src/logic/validateIntent.ts
```diff
@@ -0,0 +1,275 @@
...
+    const { maxFeePerGas, maxPriorityFeePerGas, gasOptions } = estimatedFees.parameters;
+    const { maximalPriorityFee, minimalPriorityFee, recommendedNextBaseFee } = await (async (
+      options: unknown,
+    ): Promise<{
+      maximalPriorityFee: bigint;
+      minimalPriorityFee: bigint;
+      recommendedNextBaseFee: bigint;
+    }> => {
+      if (isApiGasOptions(options)) {
+        console.log("Hallo");
```
dilaouid: Hallo 👋

libs/coin-modules/coin-evm/src/logic/common.ts
```diff
@@ -1,6 +1,56 @@
...
+    "slow" in options &&
+    isApiFeeData(options.slow) &&
+    "medium" in options &&
+    isApiFeeData(options.medium) &&
+    "fast" in options &&
+    isApiFeeData(options.fast)
+  );
+}
+
+export function isApiFeeData(fees: unknown): fees is ApiFeeData {
```
qperrot, line 20: Maybe it is juste me but perhaps we can improve readability 
something like 
```export function isApiFeeData(fees: unknown): fees is ApiFeeData {
  if (
    !fees ||
    typeof fees !== "object"
  ) {
    return false;
  }

  const obj = fees as Record<string, unknown>;

  const isBigIntOrNull = (value: unknown) =>
    value === null || typeof value === "bigint";

  return (
    "maxFeePerGas" in obj && isBigIntOrNull(obj.maxFeePerGas) &&
    "maxPriorityFeePerGas" in obj && isBigIntOrNull(obj.maxPriorityFeePerGas) &&
    "gasPrice" in obj && isBigIntOrNull(obj.gasPrice) &&
    "nextBaseFee" in obj && isBigIntOrNull(obj.nextBaseFee)
  );
}```

qperrot (commented):
Don't you need to have a validateNFT in your `validateIntent` ?

`ValidateIntent` is some kind of  replacement for  `getTransactionStatus`

libs/coin-modules/coin-evm/src/logic/common.ts
francois-guerin-ledger: We can definitely do that

PR #11418, opened 2025-08-21
Support[icp]: Retire rosetta usage and move to icp canisters for API

lawRathod:
### ✅ Checklist

  - ...

### 📝 Description

Replaces rosetta api with canister support 

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/coin-modules/coin-internet_computer/src/bridge/broadcast.ts
+++ b/libs/coin-modules/coin-internet_computer/src/bridge/broadcast.ts
@@ -10,20 +10,28 @@ interface BroadcastRawData {
   encodedSignedCallBlob: string;
 }
 
+// Type guard to validate rawData shape
+function isBroadcastRawData(data: unknown): data is BroadcastRawData {
+  return (
+    typeof data === "object" &&
+    data !== null &&
+    "encodedSignedCallBlob" in data &&
+    typeof (data as any).encodedSignedCallBlob === "string"
+  );
+}
+
 // Main broadcast function for handling Internet Computer transactions
 export const broadcast: AccountBridge<Transaction>["broadcast"] = async ({
   signedOperation: { operation, rawData },
 }) => {
   log("debug", "[broadcast] Internet Computer transaction broadcast initiated");
 
-  // Type assertion and validation for rawData
-  const rawDataTyped = rawData as unknown as BroadcastRawData;
-  invariant(rawDataTyped, "[ICP](broadcast) Missing rawData");
-  invariant(rawDataTyped.encodedSignedCallBlob, "[ICP](broadcast) Missing encodedSignedCallBlob");
+  // Validate rawData with type guard
+  invariant(isBroadcastRawData(rawData), "[ICP](broadcast) Invalid rawData format");
   invariant(operation.extra, "[ICP](broadcast) Missing operation extra");
 
   await broadcastTxn(
-    Buffer.from(rawDataTyped.encodedSignedCallBlob, "hex"),
+    Buffer.from(rawData.encodedSignedCallBlob, "hex"),
     MAINNET_LEDGER_CANISTER_ID,
     "call",
   );


libs/coin-modules/coin-internet_computer/src/bridge/broadcast.ts
```diff
@@ -1,15 +1,32 @@
...
 export const broadcast: AccountBridge<Transaction>["broadcast"] = async ({
-  signedOperation: { signature, operation },
+  signedOperation: { operation, rawData },
 }) => {
-  // log("debug", "[broadcast] start fn");
+  log("debug", "[broadcast] Internet Computer transaction broadcast initiated");
 
-  await broadcastTxn(signature);
+  // Type assertion and validation for rawData
+  const rawDataTyped = rawData as unknown as BroadcastRawData;
```
scavassa-yld: It might be worth considering typing rawData explicitly or introducing a type guard instead of using as unknown as BroadcastRawData, to reduce runtime checks and let the compiler enforce more invariants.

libs/coin-modules/coin-internet_computer/src/bridge/signOperation.ts
```diff
@@ -1,53 +1,103 @@
...
+import { Cbor } from "@zondax/ledger-live-icp/agent";
 import { buildOptimisticOperation } from "./buildOptimisticOperation";
 import { Transaction } from "../types";
 import { SignerContext } from "@ledgerhq/coin-framework/signer";
 import { ICPSigner } from "../types";
 import { getPath } from "../common-logic";
+import { log } from "@ledgerhq/logs";
+import invariant from "invariant";
+
+const signICPTransaction = async (
```
scavassa-yld, line 19: **Just as a thought**: since there are quite a few invariant runtime checks here, would it make sense to explore whether some of these could be enforced at compile time through stricter types or interfaces? 
Not a requirement, just wondering if that could simplify things.

libs/coin-modules/coin-internet_computer/src/bridge/signOperation.ts
```diff
@@ -1,53 +1,103 @@
...
+      content: unsignedTxn,
+      sender_pubkey: pubkeyToDer(account.xpub),
+      sender_sig: signatures.signatureRS,
+    },
+  };
+};
 
 export const buildSignOperation =
-  (signerContext: SignerContext<ICPSigner>): AccountBridge<Transaction, Account>["signOperation"] =>
+  (signerContext: SignerContext<ICPSigner>): AccountBridge<Transaction>["signOperation"] =>
```
scavassa-yld, line 45: since the async flow here is wrapped in an Observable, would it be worth looking at making the error handling more explicit (for example with o.error()), to ensure errors propagate as expected? Totally optional, just raising it for consideration.

libs/coin-modules/coin-internet_computer/src/consts.ts
```diff
@@ -9,3 +9,13 @@ export const ICP_FEES = 1e4;
 
 // Max Memo value on ICP network
 export const MAX_MEMO_VALUE = Number.MAX_SAFE_INTEGER;
+
+// API limits
+export const FETCH_TXNS_LIMIT = 100;
+
+export {
```
scavassa-yld, line 16: Just as a note: since these constants are being re-exported from this module, would it be worth checking if all existing usages have been updated to use the new exports, to avoid any potential breaking changes?

libs/coin-modules/coin-internet_computer/src/hw-signMessage.ts
```diff
@@ -19,11 +18,7 @@ export const signMessage =
     if (typeof message !== "string") throw new Error("Message must be a string");
 
     const { r } = await signerContext(deviceId, async signer => {
-      const r = await signer.sign(
-        account.freshAddressPath,
-        getBufferFromString(message),
-        ICP_SEND_TXN_TYPE,
-      );
+      const r = await signer.sign(account.freshAddressPath, getBufferFromString(message));
```
scavassa-yld, line 21: As we have a change for function signature for sing, can we be sure on checking possible braking changes?

libs/coin-modules/coin-internet_computer/src/types/signer.ts
```diff
@@ -18,5 +27,5 @@ export type ICPGetAddrResponse = {
 export interface ICPSigner {
   showAddressAndPubKey(path: string): Promise<ICPGetAddrResponse>;
   getAddressAndPubKey(path: string): Promise<ICPGetAddrResponse>;
-  sign(path: string, message: Buffer, txtype: number): Promise<ICPSignature>;
+  sign(path: string, message: Buffer): Promise<ICPSignature>;
```
scavassa-yld, line 30: just part of my question on hw-signMessage.ts

libs/env/src/env.ts
```diff
@@ -57,11 +57,6 @@ const envDefinitions = {
     parser: boolParser,
     desc: "Show theme debug overlay UI",
   },
```
scavassa-yld, line 59: As removing API ICP ENDPOINT  could affect other parts of LL, are we sure references are cleared?

libs/coin-modules/coin-internet_computer/src/bridge/signOperation.ts
lawRathod: The error should propagate, we are using o.error() [here](https://github.com/LedgerHQ/ledger-live/pull/11418/files/4db81da95ee8b887db3ed69c16e1958db6ce9ebe#diff-668c386d3681339894e6c02149ebab0b7d966a95968c9c6a52202530154f9ca8R107)

libs/coin-modules/coin-internet_computer/src/consts.ts
lawRathod: I have double checked but so far all usage of these consts are covered by integration tests and bot tests. These const are used in network calls.

libs/env/src/env.ts
lawRathod: Double checked, I didn't find any existing ref to `API_ICP_ENDPOINT`

libs/coin-modules/coin-internet_computer/src/bridge/signOperation.ts
lawRathod: I agree, hard assumptions through invariants can look ugly. Tho, the signer needs to match the types coming from hw package and I am not sure it will be easy to impose these assertions at compile time.

libs/coin-modules/coin-internet_computer/src/bridge/broadcast.ts
lawRathod: Agreed, added the typeguard but I am not sure how can we enforce compile time check using it?

scavassa-yld: My comment about “reducing runtime checks” was mainly about typing rawData as BroadcastRawData earlier so the compiler can enforce it.
With your fix adding the type guard we’re fine — that covers the runtime validation, and it’s a good approach.

libs/coin-modules/coin-internet_computer/src/bridge/signOperation.ts
scavassa-yld: I see. It was just a thought. Implementation seems fine :)

PR #11485, opened 2025-08-25
fix(ton): syncing when publicKey and addr is same

lawRathod:
### ✅ Checklist

  - ...

### 📝 Description

When account is imported using account id, the address is missing in account objest as it was never present in the accountId and is equal to public key. This fails sync since parsing address isn't possible, the fix is to derive address from pubkey.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-20982

---

### DIFF
--- a/libs/coin-modules/coin-ton/src/synchronisation.ts
+++ b/libs/coin-modules/coin-ton/src/synchronisation.ts
@@ -50,6 +50,8 @@ export const getAccountShape: GetAccountShape<TonAccount> = async (
       workchain: 0,
       publicKey: Buffer.from(publicKey, "hex"),
     }).address.toString({ bounceable: false, urlSafe: true });
+    // update the account info with the correct address
+    info.address = address;
   }
 
   const blockHeight = await fetchLastBlockNumber();


libs/coin-modules/coin-ton/src/synchronisation.ts
```diff
@@ -31,17 +31,27 @@ import {
...
 
   const publicKey = reconciliatePubkey(rest?.publicKey, initialAccount);
 
+  // handle when address is pubkey, can happen when accounts imported using accountID
+  if (publicKey === address) {
+    address = WalletContractV4.create({
+      workchain: 0,
+      publicKey: Buffer.from(publicKey, "hex"),
+    }).address.toString({ bounceable: false, urlSafe: true });
+  }
```
Canestin:
```suggestion
    }).address.toString({ bounceable: false, urlSafe: true });
    // update the account info with the correct address
    info.address = address;
  }
```

Canestin: If we don't update `info.address`, we will encounter the same issue in other functions coming from:
https://github.com/ton-org/ton-core/blob/main/src/address/Address.ts#L115

```markdown

Error: Unknown address type: 9e2c90eba44812043c01af90732af9552888687ed6d13ab51b086703f248daa5
at Address.parse (chunk-QNVZM2VR.js?v=cb1095b4:141:17)
at fetchJettonWallets (api.ts:113:29)
at getSubAccountShape (synchronisation.ts:162:29)
at getSubAccounts (synchronisation.ts:212:7)
at getAccountShape (synchronisation.ts:127:31)
at async main (jsHelpers.ts:212:25)

```

How I reproduced it: I used [this app.json](https://api.jsonbin.io/v3/b/68af0414ae596e708fd8ea5b), where the particularity is that i put the publicKey in `data.freshAddress`.

lawRathod: I see, updating the address is definitely important. Great catch! Thanks 🙌🏻

PR #11633, opened 2025-09-03
[LIVE-21242] fix(Custom Lock Screen LLD): implement reconstruction of picture from device for Apex

ofreyssinet-ledger:
### ✅ Checklist

  - preview of the Custom Lock Screen when restoring it in the OS update, for Apex

### 📝 Description

Context: when we do an OS update, we download the lock screen picture from the device as it gets deleted.
And then, when we upload it back to the device, we display a preview of that picture.
This algo is for reconstructing the picture from the raw picture data pulled from the device.
As pictures on Stax/Flex are 4 bits per pixel, and pictures on Apex are 1 bit per pixel, the data is serialised differently, so a new algo is necessary.
It was already implemented on LLM, this is the LLD implementation.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-21242] 

---

[LIVE-21242]: https://ledgerhq.atlassian.net/browse/LIVE-21242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/CustomImage/TestImage.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/CustomImage/TestImage.tsx
@@ -2,8 +2,11 @@ import React, { useEffect, useState } from "react";
 import { Alert, Flex, Text } from "@ledgerhq/react-ui";
 import { ImageProcessingError } from "@ledgerhq/live-common/customImage/errors";
 import { CLSSupportedDeviceModelId } from "@ledgerhq/live-common/device/use-cases/isCustomLockScreenSupported";
-import { getScreenVisibleAreaDimensions } from "@ledgerhq/live-common/device/use-cases/screenSpecs";
 import { ProcessorResult } from "~/renderer/components/CustomImage/dithering/types";
+import {
+  getScreenSpecs,
+  getScreenVisibleAreaDimensions,
+} from "@ledgerhq/live-common/device/use-cases/screenSpecs";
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/renderer/components/CustomImage/TestImage.tsx
```diff
@@ -2,16 +2,20 @@ import React, { useEffect, useState } from "react";
...
 import { ImageBase64Data } from "~/renderer/components/CustomImage/types";
 import { createCanvas } from "~/renderer/components/CustomImage/imageUtils";
 
 export function reconstructImage({
   hexData,
   width,
   height,
-}: ProcessorResult["rawResult"]): ImageBase64Data {
+  bitsPerPixel,
+}: ProcessorResult["rawResult"] & { bitsPerPixel: 1 | 4 }): ImageBase64Data {
```
aussedatlo, line 18: [COULD] use a type for this one, something like 
```ts
type Example: { result: ProcessorResult["rawResult"], bitsPerPixel: BitPerPixel}
```

with a `BitPerPixel` enum ?

OlivierFreyssinet: @aussedatlo good idea, this type is in many places, I'll refactor it later 👍

PR #11654, opened 2025-09-04
refacto: improve call to FF feature_cal_lazy_loading + CAL

dilaouid:
### ✅ Checklist

  - ...

### 📝 Description

Refactor [this ticket](https://ledgerhq.atlassian.net/browse/LIVE-20069) following [this issue](https://ledgerhq.atlassian.net/browse/LIVE-21162).
Now successfully check the correct value for the cal feature flag.

### ❓ Context

- **[JIRA or GitHub link](https://ledgerhq.atlassian.net/browse/LIVE-20069)**: 

---

### DIFF
--- a/libs/coin-framework/.unimportedrc.json
+++ b/libs/coin-framework/.unimportedrc.json
@@ -21,8 +21,7 @@
     "src/transaction/common.ts",
     "src/utils.ts",
     "src/sanction/index.ts",
-    "src/sanction/errors.ts",
-    "src/crypto-assets/utils.ts"
+    "src/sanction/errors.ts"
   ],
   "ignoreUnimported": [
     "src/config.ts",
--- a/libs/coin-framework/.unimportedrc.json
+++ b/libs/coin-framework/.unimportedrc.json
@@ -21,7 +21,8 @@
     "src/transaction/common.ts",
     "src/utils.ts",
     "src/sanction/index.ts",
-    "src/sanction/errors.ts"
+    "src/sanction/errors.ts",
+    "src/crypto-assets/utils.ts"
   ],
   "ignoreUnimported": [
     "src/config.ts",
--- a/libs/coin-framework/src/crypto-assets/utils.ts
+++ b/libs/coin-framework/src/crypto-assets/utils.ts
@@ -1,0 +1,28 @@
+import type { CryptoAssetsStore } from "@ledgerhq/types-live";
+import type { TokenCurrency } from "@ledgerhq/types-cryptoassets";
+import { LiveConfig } from "@ledgerhq/live-config/LiveConfig";
+
+export type AddTokens = (tokens: TokenCurrency[]) => void;
+
+// Resolves an addTokens function from a CryptoAssetsStore getter, if available
+// Returns null if the store is not set yet or does not support dynamic additions
+export function resolveTokenAdder(getStore: () => CryptoAssetsStore): AddTokens | null {
+  try {
+    const store = getStore();
+    return store?.addTokens?.bind(store) || null;
+  } catch (_) {
+    // store not set yet
+  }
+  return null;
+}
+
+export function isCalLazyLoadingActive(getStore: () => CryptoAssetsStore): boolean {
+  try {
+    const value = LiveConfig.getValueByKey("feature_cal_lazy_loading");
+    const enabled = value === true || value === "true";
+    const hasDynamic = resolveTokenAdder(getStore) !== null;
+    return enabled && hasDynamic;
+  } catch (e) {
+    return false;
+  }
+}
--- a/libs/coin-framework/src/crypto-assets/utils.ts
+++ b/libs/coin-framework/src/crypto-assets/utils.ts
@@ -1,6 +1,5 @@
 import type { CryptoAssetsStore } from "@ledgerhq/types-live";
 import type { TokenCurrency } from "@ledgerhq/types-cryptoassets";
-import { LiveConfig } from "@ledgerhq/live-config/LiveConfig";
 
 export type AddTokens = (tokens: TokenCurrency[]) => void;
 
@@ -14,15 +13,4 @@ export function resolveTokenAdder(getStore: () => CryptoAssetsStore): AddTokens 
     // store not set yet
   }
   return null;
-}
-
-export function isCalLazyLoadingActive(getStore: () => CryptoAssetsStore): boolean {
-  try {
-    const value = LiveConfig.getValueByKey("feature_cal_lazy_loading");
-    const enabled = value === true || value === "true";
-    const hasDynamic = resolveTokenAdder(getStore) !== null;
-    return enabled && hasDynamic;
-  } catch (e) {
-    return false;
-  }
 }
--- a/libs/coin-modules/coin-evm/package.json
+++ b/libs/coin-modules/coin-evm/package.json
@@ -85,6 +85,7 @@
   "devDependencies": {
     "@ledgerhq/disable-network-setup": "workspace:^",
     "@ledgerhq/hw-app-eth": "workspace:^",
+    "@ledgerhq/live-config": "workspace:^",
     "@ledgerhq/types-cryptoassets": "workspace:^",
     "@ledgerhq/types-live": "workspace:^",
     "@types/bluebird": "^3.5.42",
--- a/libs/coin-modules/coin-evm/package.json
+++ b/libs/coin-modules/coin-evm/package.json
@@ -85,7 +85,6 @@
   "devDependencies": {
     "@ledgerhq/disable-network-setup": "workspace:^",
     "@ledgerhq/hw-app-eth": "workspace:^",
-    "@ledgerhq/live-config": "workspace:^",
     "@ledgerhq/types-cryptoassets": "workspace:^",
     "@ledgerhq/types-live": "workspace:^",
     "@types/bluebird": "^3.5.42",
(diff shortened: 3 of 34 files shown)


libs/coin-modules/coin-evm/src/bridge/preload.ts
```diff
@@ -84,16 +85,18 @@ export const fetchERC20Tokens: (
...
   log("evm/preload", "preload " + erc20.length + " tokens");
-  const store = getCryptoAssetsStore();
-  if ("addTokens" in store && typeof store.addTokens === "function") {
+  const addTokensToStore = resolveTokenAdder(getCryptoAssetsStore);
+  if (addTokensToStore) {
     const convertedTokens = erc20.map(convertERC20).filter(Boolean);
-    store.addTokens(convertedTokens);
+    if (convertedTokens.every(token => token !== undefined)) {
+      addTokensToStore(convertedTokens);
+    }
```
Justkant: If you did this change for the types, you should be able to change the `filter(Boolean)` with a predicate that TS can infer so you get the proper types

You could test `filter(token => (!!token))`

libs/coin-modules/coin-solana/src/preload.ts
```diff
@@ -149,32 +137,13 @@ export function hydrate(data: SolanaPreloadData | undefined, currency: CryptoCur
 }
 
 function hydrateV1(data: SolanaPreloadDataV1, currency: CryptoCurrency) {
+  chooseCorrectStore();
```
Justkant: Carefull here you are not using `data.splTokens`

You should pass the array in the params IMO

libs/ledgerjs/packages/types-live/src/crypto-assets/index.ts
```diff
@@ -6,6 +6,7 @@ export type CryptoAssetsStore = {
   findTokenById(id: string): TokenCurrency | undefined;
   findTokenByAddressInCurrency(address: string, currencyId: string): TokenCurrency | undefined;
   findTokenByTicker(ticker: string): TokenCurrency | undefined;
+  addTokens: (tokens: TokenCurrency[]) => void;
```
Justkant: Why do we need to add this here now ?

pnpm-lock.yaml
Justkant, line 1: You can revert this change or simple checkout the file from develop as we are not touching the deps

Justkant (commented):
Looks better now at least the feature flag should be up to date when checked

We might need to also update the default value for the feature flag and coin config to avoid being true until we receive the feature flag from firebase, WDYT ?

libs/coin-framework/src/crypto-assets/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+      value,
+      "enabled=",
+      enabled,
+      "hasDynamicStore=",
+      hasDynamic,
+    );
+    return enabled && hasDynamic;
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.log("[CAL] feature_cal_lazy_loading error:", e);
```
Justkant: Could you remove the logs please ?
And now that I look at it we could even remove the try/catch I think

dilaouid: Fixed

libs/coin-modules/coin-evm/package.json
```diff
@@ -85,6 +85,7 @@
   "devDependencies": {
     "@ledgerhq/disable-network-setup": "workspace:^",
     "@ledgerhq/hw-app-eth": "workspace:^",
+    "@ledgerhq/live-config": "workspace:^",
```
Justkant: Well it won't impact us much with the apps and the way it's bundled but we should probably have this in the `dependencies` object and not the dev one

Also I don't think we should add this dep in the coin modules, usually if you check the config file in coin-solana for example we pass in the data from common

libs/coin-framework/.unimportedrc.json
```diff
@@ -21,7 +21,8 @@
     "src/transaction/common.ts",
     "src/utils.ts",
     "src/sanction/index.ts",
-    "src/sanction/errors.ts"
+    "src/sanction/errors.ts",
+    "src/crypto-assets/utils.ts"
```
Justkant: We could also export it in the `crypto-assets/index.js` if it makes sense as it's listed below in `ignoreUnimported` already

I don't mind the current solution

pnpm-lock.yaml
Justkant, line 1: We should revert this, you can use `git checkout origin/develop ./pnpm-lock.yaml`

libs/coin-modules/coin-solana/src/config.ts
```diff
@@ -4,6 +4,7 @@ export type SolanaConfig = {
   token2022Enabled: boolean;
   queuedInterval: number;
   legacyOCMSMaxVersion: string;
+  calLazyLoading?: boolean;
```
Justkant: I think we might be missing the default values defined in common config files for evm and solana

You should also remove the optional

libs/coin-modules/coin-solana/src/preload.ts
```diff
@@ -17,18 +17,41 @@ import { getCALHash, setCALHash } from "./logic";
...
+    addTokens(tokens.map(convertSplTokens));
+  }
+};
 
 export const PRELOAD_MAX_AGE = 15 * 60 * 1000; // 15min
 
+const isLazyLoadingEnabled = (): boolean => {
+  try {
+    const override = coinConfig.getCoinConfig()?.calLazyLoading;
+    if (override !== undefined) return override;
```
Justkant: By providing a default value in the config in common and removing the optional you can simplify this code

I think we should also remove the try catch, the error is meant to be seen in dev and it's better to see the error to correct it instead of silently doing a fallback

libs/coin-modules/coin-solana/src/config.ts
dilaouid: There is a lot of different config for `evmConfig`, should I apply `calLazyLoading: false` to all of them ?
Also, this changes made several impactes:, having to update all calls of coinconfig for solana / evm, including in the coin-tester

Maybe the FF can be the appropriate option since this is a temporary PR, avoiding too much lines changes so it could be easily reverted later?

libs/coin-framework/.unimportedrc.json
dilaouid: Fixed !

Justkant (approved):

Justkant (dismissed):

libs/coin-modules/coin-solana/src/preload.ts
```diff
@@ -17,18 +17,39 @@ import { getCALHash, setCALHash } from "./logic";
...
+
+const storeTokens = (tokens: SPLToken[]) => {
+  const addTokensToStore = resolveTokenAdder(getCryptoAssetsStore);
+  if (addTokensToStore) {
+    const convertedTokens = tokens.map(convertSplTokens);
+    addTokensToStore(convertedTokens);
+  } else {
+    addTokens(tokens.map(convertSplTokens));
+  }
+};
```
francois-guerin-ledger, line 31: No big deal, just simpler
```suggestion
const storeTokens = (tokens: SPLToken[]) => {
  const addTokensToStore = resolveTokenAdder(getCryptoAssetsStore);
  const convertedTokens = tokens.map(convertSplTokens);
  if (addTokensToStore) {
    addTokensToStore(convertedTokens);
  } else {
    addTokens(convertedTokens);
  }
};
```

libs/ledger-live-common/src/bridge/crypto-assets/index.test.ts
```diff
@@ -96,16 +50,9 @@ describe("Testing CryptoAssetStore", () => {
     expect(store).toBe(mockCALStore);
   });
 
-  it("should prioritize CAL integration over feature flags", () => {
+  it("prioritizes CAL integration over any custom store", () => {
```
francois-guerin-ledger, line 53: +1 for imperative tense

francois-guerin-ledger (dismissed):

KVNLS (approved):
I have not tested and only reviewed files owned by Live Hub

ejaycoleman (approved):

libs/coin-modules/coin-evm/src/bridge/preload.ts
```diff
@@ -8,13 +8,21 @@ import {
...
 
-let shouldSkipTokenLoading = false;
-export function setShouldSkipTokenLoading(skip: boolean): void {
-  shouldSkipTokenLoading = skip;
-}
+const isDefined = <T>(x: T | undefined | null): x is T => x !== undefined && x !== null;
+
+const isCalEnabled = (currency: CryptoCurrency): boolean => {
+  try {
+    return getCoinConfig(currency).info.calLazyLoading ?? false;
```
francois-guerin-ledger, line 20:
```suggestion
    return !!getCoinConfig(currency).info.calLazyLoading;
```

libs/coin-modules/coin-solana/src/preload.ts
```diff
@@ -17,18 +17,39 @@ import { getCALHash, setCALHash } from "./logic";
...
 
 export const PRELOAD_MAX_AGE = 15 * 60 * 1000; // 15min
 
+const isLazyLoadingEnabled = (): boolean => {
+  try {
+    return coinConfig.getCoinConfig().calLazyLoading ?? false;
```
francois-guerin-ledger, line 37:
```suggestion
    return !!coinConfig.getCoinConfig().calLazyLoading;
```

francois-guerin-ledger (dismissed):

francois-guerin-ledger:
Could we close this PR now that the related ticket has been trashed ?

PR #11751, opened 2025-09-10
feat(coin-bitcoin): support psbt v2 signature

Canestin:
### ✅ Checklist

  - ...

https://github.com/user-attachments/assets/2c479d00-3d71-4830-a2c4-d9e8ae567e54

### 📝 Description

### ❓ Context

- **JIRA**: https://ledgerhq.atlassian.net/browse/LIVE-21077

---


libs/coin-modules/coin-bitcoin/src/getTransactionStatus.ts
```diff
@@ -123,6 +124,11 @@ export const getTransactionStatus: AccountBridge<
 
   if (!errors.amount && !amount.gt(0)) {
     errors.amount = useAllAmount ? new NotEnoughBalance() : new AmountRequired();
+
+    // For a PSBT tx we do not provide the amount so we do not throw AmountRequired error
+    if (transaction.psbt && !useAllAmount) {
```
Justkant: Why do we need check for `useAllAmout` when `transaction.psbt` is present ?

libs/coin-modules/coin-bitcoin/src/signOperation.ts
```diff
@@ -81,56 +82,100 @@ export const buildSignOperation =
...
-          },
-        });
+        if (transaction.psbt) {
+          const psbtBuffer = Buffer.from(transaction.psbt, "base64");
+
+          const psbtResult = await signerContext(deviceId, currency, signer =>
+            signer.signPsbtV2Buffer
+              ? signer.signPsbtV2Buffer(psbtBuffer, {
+                  // @ts-expect-error: x
+                  finalizePsbt: transaction.finalizePsbt,
```
Justkant: Why do we expect an error ? Can't we fix the type ?
Also please add a reason instead of `x` if you need it

libs/coin-modules/coin-bitcoin/src/signOperation.ts
```diff
@@ -81,56 +82,100 @@ export const buildSignOperation =
...
+
+          o.next({ type: "device-signature-granted" });
+
+          // Build the optimistic operation pour PSBT
+          const operation: Operation = {
+            id: encodeOperationId(account.id, "", "OUT"),
+            hash: "",
+            type: "OUT",
+            value: new BigNumber(transaction.amount).plus(fee),
+            fee,
```
Justkant: Calculate fees might not work as expected with a raw transaction as it uses buildTransaction to calculate the fees, we might want to look into it

Did you test the sync and make sure the history is synced correctly ?

libs/hw-ledger-key-ring-protocol/package.json
```diff
@@ -19,17 +19,18 @@
...
     "@ledgerhq/hw-transport": "workspace:*",
-    "@ledgerhq/logs": "workspace:*",
     "@ledgerhq/live-env": "workspace:*",
-    "bip32": "^4.0.0",
+    "@ledgerhq/logs": "workspace:*",
     "@noble/curves": "^1.9.7",
-    "create-hmac": "^1.1.7"
+    "bip32": "^4.0.0",
+    "create-hmac": "^1.1.7",
+    "secp256k1": "link:@noble/curves/secp256k1"
```
Justkant: Is there a specific reason to not use the package directly instead of doing a link ?
I know that @gre removed some libs too by using noble directly

libs/ledgerjs/packages/hw-app-btc/package.json
```diff
@@ -62,15 +62,16 @@
...
     "@ledgerhq/hw-transport": "workspace:^",
     "@ledgerhq/logs": "workspace:^",
+    "@noble/curves": "1.9.7",
     "bip32-path": "^0.4.2",
     "bitcoinjs-lib": "^5.2.0",
     "bs58": "^4.0.1",
     "bs58check": "^2.1.2",
     "invariant": "^2.2.4",
     "ripemd160": "2",
+    "secp256k1": "link:@noble/curves/secp256k1",
```
Justkant: Same on this one

libs/coin-modules/coin-bitcoin/src/getTransactionStatus.ts
Canestin: > Because on the previous line we want to throw `NotEnoughBalance` when `useAllAmount` is `false`.
> We only want to avoid an `AmountRequired()` error. Since we don't provide an amount anyway, we can simply skip this check when `transaction.psbt` is present.

libs/coin-modules/coin-bitcoin/src/signOperation.ts
Canestin: It was an oversight, fixed

libs/coin-modules/coin-bitcoin/src/signOperation.ts
Canestin: Interesting, I'll look into it.

libs/hw-ledger-key-ring-protocol/package.json
Canestin: No reason, conflict resolution,  I synced with develop.

libs/ledgerjs/packages/hw-app-btc/package.json
Canestin: No reason, conflict resolution, I synced with develop.

libs/coin-modules/coin-bitcoin/src/getTransactionStatus.ts
Justkant: > Because on the previous line we want to throw NotEnoughBalance when useAllAmount is false.

I think it's the inverse but doesn't really matter

You might want to simply bypass the above check (`!errors.amount && !amount.gt(0)`) when `psbt` is present, because you always want to bypass the error for psbt as we always set the amout to 0 and we don't really let the user set useAllAmount too

libs/ledger-live-common/src/wallet-api/logic.ts
```diff
@@ -259,6 +261,59 @@ export const bitcoinFamilyAccountGetAddressLogic = (
...
+
+  // For the PSBT we cannot provide the amount and recipient at the moment
+  const liveTx: Partial<BitcoinTransaction> = {
+    family: "bitcoin",
+    amount: new BigNumber(0),
+    utxoStrategy: {
+      strategy: 0,
+      excludeUTXOs: [],
+    },
+    recipient: "bc1qtthqrseeu4d2yw9yh0clw95066klvcyvqdk7fv",
```
Wozacosta: Pretty sure we don't want to hardcode the recipient here, even if rewritten.

libs/coin-modules/coin-bitcoin/src/signRawOperation.ts
```diff
@@ -1,43 +1,40 @@
...
 
-        const feesRes = await calculateFees({ account, transaction });
-        senders = new Set(feesRes.txInputs.map(i => i.address).filter(Boolean) as string[]);
-        recipients = feesRes.txOutputs
-          .filter(o => o.address && !o.isChange)
-          .map(o => o.address) as string[];
-        fee = feesRes.fees;
+        // const txInfo = craftPsbtTransaction({ psbt });
+
+        // const feesRes = await calculateFees({ account, transaction: txInfo });
```
Canestin: This is used as a fallback in case `feeFromPsbtBase64()` returns `null` on line 87.
However, the `FeeNotLoaded` error occurs because `feePerByte` is `null`.
I tested it, and it correctly returns the fee from the PSBT.
If that’s fine, I’ll remove the comments.

libs/coin-modules/coin-bitcoin/src/signRawOperation.ts
```diff
@@ -55,147 +52,64 @@ export const buildSignRawOperation =
...
-            signedOperation: {
-              operation,
-              // Ensure non-empty signature: if not finalized, fall back to the PSBT (base64)
-              signature: psbtResult.tx || psbtResult.psbt.toString("base64"),
-              rawData: { psbtSigned: psbtResult.psbt.toString("base64") },
-            },
-          });
-          return;
+        if (!psbtResult) {
+          // TODO: find proper name, also, do we want to really throw here or handle it differently
```
Canestin: It’s okay to throw an error here, right?

libs/coin-modules/coin-bitcoin/src/signRawOperation.ts
Justkant: IMO If it's only used as a fallback you should first call `feeFromPsbtBase64` and only if it returns null you would calculateFees and everything

libs/ledger-live-common/src/families/algorand/__snapshots__/bridge.integration.test.ts.snap
Justkant, line 1: Shouldn't we revert this snapshot change as it's not linked to our current change ? And maybe fix it in another PR ? Are you rebased on develop ?

Canestin: Done!

libs/coin-modules/coin-bitcoin/src/psbtFees.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (/^[0-9a-fA-F]+$/.test(s) && s.length % 2 === 0) {
+    try {
+      return Buffer.from(s, "hex");
+    } catch {
+      /* ignore */
+    }
+  }
+
+  // Treat as base64: strip whitespace and convert URL-safe to standard
+  const b64 = s.replace(/\s+/g, "").replace(/-/g, "+").replace(/_/g, "/");
```
Justkant, line 20: We can follow the sonarcloud analysis and use replaceAll to make sure we don't forget the /g

libs/coin-modules/coin-bitcoin/src/signRawOperation.ts
```diff
@@ -0,0 +1,94 @@
...
+    new Observable(o => {
+      async function main() {
+        const { currency } = account;
+        const walletAccount = getWalletAccount(account);
+
+        log("hw", `signRawTransaction ${currency.id} for account ${account.id}`);
+
+        const networkParams = getNetworkParameters(currency.id);
+        const sigHashType = networkParams.sigHash;
+        if (isNaN(sigHashType)) {
```
Justkant: As suggested by SonarCloud
```suggestion
        if (Number.isNaN(sigHashType)) {
```

libs/coin-modules/coin-bitcoin/package.json
```diff
@@ -48,6 +48,7 @@
     "@ledgerhq/cryptoassets": "workspace:^",
     "@ledgerhq/devices": "workspace:*",
     "@ledgerhq/errors": "workspace:^",
+    "@ledgerhq/hw-app-btc": "workspace:^",
```
Justkant, line 51: Why do we need this added ?

If it's only for a type I would prefer to find another way to share it

Canestin: We need it in [libs/coin-modules/coin-bitcoin/src/psbtFees.ts](https://github.com/LedgerHQ/ledger-live/pull/11751/files#diff-0f4771bf0964ce0ece8f173f5e6ea6978a53e350a403feccc848e9268aa23ed1R2)

Justkant: I think it would still be good to avoid have hw-app-btc in the coin-module, we should probably move this file in a shared library for both hw-app-btc and the coin-module, wdyt ?

Justkant: Also did you test this lib: https://github.com/caravan-bitcoin/caravan/tree/main/packages/caravan-psbt

Looks like it would support v0 to v2 transformation for us: https://github.com/caravan-bitcoin/caravan/tree/main/packages/caravan-psbt#static-psbtv2fromv0

This site seems to use it too:
- https://bip370.org/
- https://github.com/Shadouts/bip370org

Justkant:
Follow up PR https://github.com/LedgerHQ/ledger-live/pull/13340

PR #11788, opened 2025-09-12
fix(coin:sui): large accounts are missing some operations

jprudent:
### ✅ Checklist

  - coin module API now returns all operations for address
  - duplicated operations may happen for most of out operations
  - the previous algorithm still exists and nothing has changed for the bridge sync

### 📝 Description

- Current algorithm has some cons:
   - was missing many operations in ascending mode
   - cursor wasn't returned in some cases
   - somehow inefficient (filtering of ops, no concurrency)
   - is difficult to debug/maintain (recursion with mutable parameter, handling of single cursor with timestamp filtering, lot of tests to migrate when refactoring)
- but has some pros   
   - is working in descending mode
   - never return duplicated operations 

- We implemented a simpler algorithm:
   - we just call "in" and "out" operations concurrently and 
   - the API cursor is the concatenation of both cursors.
   - a big cons: it may have duplicated operations
      - it's a pragmatic choice: we prefer a simpler algo, handling dups at caller site rather than wondering why some operations are missing in the lib

### ❓ Context

https://ledgerhq.atlassian.net/browse/BACK-9674

---

### DIFF
--- a/libs/coin-modules/coin-sui/src/network/sdk.ts
+++ b/libs/coin-modules/coin-sui/src/network/sdk.ts
@@ -582,14 +582,14 @@ export const getListOperations = async (
     const { out: outCursor, in: inCursor } = deserializeCursor(cursor);
 
     const [opsOut, opsIn] = await Promise.all([
-      await queryTransactions({
+      queryTransactions({
         api,
         addr,
         type: "OUT",
         cursor: toSdkCursor(outCursor),
         order: rpcOrder,
       }),
-      await queryTransactions({
+      queryTransactions({
         api,
         addr,
         type: "IN",
--- a/libs/coin-modules/coin-sui/src/network/sdk.ts
+++ b/libs/coin-modules/coin-sui/src/network/sdk.ts
@@ -41,6 +41,7 @@ import { getEnv } from "@ledgerhq/live-env";
 import { SUI_SYSTEM_STATE_OBJECT_ID } from "@mysten/sui/utils";
 import { getCurrentSuiPreloadData } from "../bridge/preload";
 import { ONE_SUI } from "../constants";
+import bs58 from "bs58";
 
 const apiMap: Record<string, SuiClient> = {};
 type AsyncApiFunction<T> = (api: SuiClient) => Promise<T>;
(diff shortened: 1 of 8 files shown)


estrauser-ledger (approved):

libs/coin-modules/coin-sui/src/network/sdk.ts
```diff
@@ -520,44 +525,94 @@ export const filterOperations = (
...
+type Cursor = {
+  out?: string;
+  in?: string;
+};
+
+function serializeCursor(cursor: Cursor): string {
+  return bs58.encode(Buffer.from(JSON.stringify(cursor)));
+}
+
+function deserializeCursor(b58cursor: string | undefined): Cursor {
```
semeano, line 541: Rename argument to have the same name as module.

```suggestion
function deserializeCursor(bs58cursor: string | undefined): Cursor {
```

ishaba (approved):

libs/coin-modules/coin-sui/src/network/sdk.ts
```diff
@@ -520,44 +525,94 @@ export const filterOperations = (
...
-    if (order) {
-      rpcOrder = order === "asc" ? "ascending" : "descending";
-    } else {
-      rpcOrder = cursor ? "ascending" : "descending";
-    }
+    const rpcOrder = convertApiOrderToSdkOrder(order);
+    const { out: outCursor, in: inCursor } = deserializeCursor(cursor);
+
+    const [opsOut, opsIn] = await Promise.all([
+      await queryTransactions({
```
jonezque: we have `await Promise.all` so `await queryTransactions` is not needed, both for L585 and L592

libs/coin-modules/coin-sui/src/network/sdk.ts
jprudent: Thanks for the review.

I wrote `b58` like base58. The lib is named `bs58` but what the `s` stands for ?

semeano: `bs` also stands for base. My suggestion was just to keep the same naming so it's easier to identify that it's a base58 cursor. Having "bs58" and "b58" could suggest they're different things and create confusion.

semeano (approved):

francois-guerin-ledger (approved):

PR #11826, opened 2025-09-16
ci: move desktop tests to 16CPU runners and optimize execution

lewisd5:
- Use larger Ubuntu runners for mock tests
- Enable turborepo for linting

Ticket: https://ledgerhq.atlassian.net/browse/LIVE-21435
Test flow: https://github.com/LedgerHQ/ledger-live/actions/runs/17712539818

### DIFF
--- a/tools/actions/composites/setup-test-desktop/action.yml
+++ b/tools/actions/composites/setup-test-desktop/action.yml
@@ -46,7 +46,7 @@ runs:
     - name: Install dependencies
       env:
         LANG: en_US.UTF-8
-      run: pnpm i --filter="ledger-live-desktop..." --filter="ledger-live" --filter="@ledgerhq/dummy-*-app..." --unsafe-perm
+      run: pnpm i --filter="ledger-live-desktop..." --filter="ledger-live" --filter="@ledgerhq/dummy-*-app..." --filter="live-cli*..." --unsafe-perm
       shell: bash
     - name: Install playwright dependencies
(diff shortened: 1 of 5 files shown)


couto (approved):

ypolishchuk-ledger (approved):

tools/actions/composites/setup-test-desktop/action.yml
```diff
@@ -46,7 +46,7 @@ runs:
     - name: Install dependencies
       env:
         LANG: en_US.UTF-8
-      run: pnpm i --filter="ledger-live-desktop..." --filter="ledger-live" --filter="@ledgerhq/dummy-*-app..." --unsafe-perm
+      run: pnpm i --filter="ledger-live-desktop..." --filter="ledger-live" --filter="@ledgerhq/dummy-*-app..." --filter="live-cli*..." --unsafe-perm
```
abdurrahman-ledger, line 49: I am wondering if we actually need live-cli for mock tests. I think it was not cleaned when we separated speculos execution into another workflow

gre-ledger (approved):

tools/actions/composites/setup-test-desktop/action.yml
lewisd5: I'll put an item in the backlog to investigate this for now as the runtime and flakiness of these tests is quite high priority. Are you happy with the approach?

PR #11883, opened 2025-09-17
feat: Add new seed confirmation and illustration to InstallSetOfApps …

mitchellv-ledger:
…component

### ✅ Checklist

  - ...

### 📝 Description

Adds a new component to display in onboarding when the user is going through onboarding with a new seed. Add `flow` field with value `onboarding` to tracking in InstallSetOfApps component.

Screenshot 2025-09-17 at 17 32 47

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-20734

---


OlivierFreyssinet:
Is it possible to have a video of the new behaviour please ?

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
```diff
@@ -178,30 +205,76 @@ const InstallSetOfApps = ({
...
+        seedConfiguration={seedConfiguration}
+      />
+      <NewSeedConfirmation
+        onConfirm={() => {
+          track("button_clicked", {
+            button: "Secure my crypto",
+            flow: "onboarding",
+            seedConfiguration,
+          });
+          isSyncIncr1Enabled ? onResult(true) : setUserConfirmed(true);
```
OlivierFreyssinet, line 247: So if I understand correctly, when the flag is enabled, it will just call onResult(true), which I believe just makes the sync onboarding move to the next step ?
And when it's not, it proceeds to install the apps ? But this behaviour does not match what is displayed in `NewSeedConfirmation`, so it seems a bit strange. Can you clarify please?

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
```diff
@@ -27,6 +29,8 @@ type Props = {
   onResult: (done: boolean) => void;
   onError?: (error: Error) => void;
   debugLastSeenDeviceModelId?: DeviceModelId;
+  isNewSeed?: boolean;
```
OlivierFreyssinet, line 32: This new prop is never used, it it part of another PR that comes later?

OlivierFreyssinet:
So I'm looking at the Figma to understand the context a bit better.
I'm not sure why you're reusing the "Install set of apps" component for a new component which purpose does not seem to be about installing a set of apps, but rather redirecting to an add account flow 🤔 .

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
mitchellv-ledger: This new prop will be used in another PR where there is a new version of the sync onboarding companion.  In this way it doesn't affect the current sync onboarding companion.  In this PR we use this `isNewSeed` prop to show the new seed panel when it is a completely fresh new seed, as show in the PR description.

OlivierFreyssinet: But this `isNewSeed` prop means that when it's true, this component is no longer meant for installing apps.

This component is called `InstallSetOfApps`, thus its responsibility is to display a UI proposing the user to install a set of apps.
Here you're introducing a new responsibility that is totally unrelated. Those two things should not be coupled, it will just make the code harder to move around or maintain.

In my opinion instead of modifying this component for that new behaviour, it's up to the parent of that component to either render `InstallSetOfApps` (if `isNewSeed`) or whatever else you want to do.

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
```diff
@@ -56,11 +63,16 @@ const InstallSetOfApps = ({
 
   const dependenciesToInstall = useMemo(() => {
     if (shouldRestoreApps && lastSeenDevice) {
+      if (lastSeenDevice.apps.length === 0) {
+        return dependencies;
+      }
```
OlivierFreyssinet, line 68: Can you explain this change of behaviour ?

ofreyssinet-ledger (changes requested):
This "NewSeedConfirmation" which is meant for redirecting to an add account flow has no place being rendered by the `InstallSetOfApps` component.
From what I see on Figma, it should be rendered _instead of_ the `InstallSetOfApps` component by its parents.

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
mitchellv-ledger: Thank you for pointing this one out, I will update the code as it should always be truthy in this case so will update and remove the ternary. We want to skip the device actions in this case to not do the install of apps and go straight to performing the onResult callback - which for the new onboarding companion will be opening the receive funds flow (as you can see in the new video in the description)

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
mitchellv-ledger: That makes total sense, and after looking at the code it would be the case if we are not installing the apps. I am double checking with the PM that we don't want to install the apps, and if so then that is perfectly correct to move it out.

apps/ledger-live-mobile/src/components/DeviceAction/InstallSetOfApps/index.tsx
mitchellv-ledger: I will double check on this, but in the case when we're restoring a device that has no apps on it then we should install the default apps instead of leaving the device empty.

OlivierFreyssinet: Good catch with that edge case 👍 but in that case I think it would make more sense to take this into account to determine `shouldRestoreApps`: `shouldRestoreApps = restore && !!lastSeenDeviceModelId && lastSeenDevice.app.length > 0`

mitchellv-ledger: I had to put it in the if clause instead as adding it to the shouldRestoreApps variable seemed to directly impact the behaviour for debug.

mitchellv-ledger:
Split the logic into two other PRs and made the changes requested from this PR in the new ones: https://github.com/LedgerHQ/ledger-live/pull/11914

PR #11918, opened 2025-09-19
Feat/live 20733 two step sync companion no actions

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

With the feature flag `llmSyncOnboardingIncr1` we switch on a new sync onboarding companion that is in two steps as shown in the video below.  We modified the animated completion screens for the devices so that they do not close unless the user presses the cta button, what this meant was that we needed to ensure the animation looped without exiting. The user will also be directed to the receive flow when a new seed is created which means that the animation has to start once it is in focus otherwise the user will see the animation in a different state than intended.

https://github.com/user-attachments/assets/51c8be71-4e3a-4994-b438-573529541002

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-20751

---


mbertin-ledger:
Is it normal to not have rounded corner with black border ? It looks cropped. 
image

mitchellv-ledger:
> Is it normal to not have rounded corner with black border ? It looks cropped. image

The header is covering the top. It's in a scrollbar and has been scrolled to the bottom so part of the top is covered up by the header.

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/FirstStepSyncOnboarding.tsx
OlivierFreyssinet, line 1: [ASK] Why this choice of copying the logic of the original SyncOnboardingCompanion rather than adapting the existing one, if the end purpose is to delete the old one?
What are the main differences apart from ?
In my opinion by copying a big part of the original logic, we lose the git history and the capacity to understand the reason for some choices of implementation.
Even for this review it makes it complicated to see what's new and needs proper examination.

mitchellv-ledger: The logic was all in one component and ~1k lines of code, and because of that code structure there are dependencies and code flows I could not easily refactor and at the same time modify for the new structure of code.  It also added a further risk that if I changed the old code I could easily break it and not be able to revert it back to original form for current users.  

I would take the new code for what it is which is new code and keep the two as practically distinct versions of similar logic/code.

The main differences are the code structures - I have separated out concerns such as synchronisation of device and step state, animation state. And there are a couple of new UI components such as CollapsibleStep and the device seeded success panel.

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/FirstStepSyncOnboarding.tsx
```diff
@@ -0,0 +1,512 @@
...
+import BackgroundGreen from "../assets/BackgroundGreen";
+import Animated, {
+  useAnimatedStyle,
+  useDerivedValue,
+  useSharedValue,
+  withTiming,
+} from "react-native-reanimated";
+import { LayoutChangeEvent, ScrollView } from "react-native";
+import { SEED_STATE } from "./TwoStepSyncOnboardingCompanion";
+
```
OlivierFreyssinet: There are cyclical dependencies in those files, maybe make sure to properly isolate types and such atomic elements like this in separate files? This is just one example

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/useFirstStepCompanionState.ts
```diff
@@ -0,0 +1,119 @@
+import {
+  OnboardingState,
+  OnboardingStep as DeviceOnboardingStep,
+} from "@ledgerhq/live-common/hw/extractOnboardingState";
+import { useEffect, useRef } from "react";
+import { FirstStepCompanionStepKey } from "./useCompanionSteps";
+import { SeedPathStatus } from "./FirstStepSyncOnboarding";
+import { SeedOriginType } from "@ledgerhq/types-live";
```
OlivierFreyssinet: cyclical dependency (both files import stuff from each other)

apps/ledger-live-mobile/src/screens/SyncOnboarding/CompletionScreen.tsx
```diff
@@ -73,13 +120,76 @@ const CompletionScreen = ({ route }: Props) => {
...
+                  </LinearGradient>
+                )}
+              </Defs>
+              <Text fill="url(#grad)">
+                <TSpan fontSize="24" x="0" y="20">
+                  {t("onboarding.completionScreen.title")}
+                </TSpan>
+              </Text>
+            </Svg>
+              </Flex> */}
```
OlivierFreyssinet, line 154: Is this for later or not needed at all?

mitchellv-ledger: This is for the later. I was not keen on leaving commented out code but I was told to comment it out for now as we'll be using it eventually

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/FirstStepSyncOnboarding.tsx
mitchellv-ledger: have moved to a types file

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/useFirstStepCompanionState.ts
mitchellv-ledger: have moved to a types file

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/types.ts
```diff
@@ -0,0 +1,32 @@
...
+export type SeedPathStatus =
+  | "choice_new_or_restore"
+  | "new_seed"
+  | "choice_restore_direct_or_recover"
+  | "restore_seed"
+  | "recover_seed"
+  | "backup_charon"
+  | "restore_charon";
+
+// Because of https://github.com/typescript-eslint/typescript-eslint/issues/1197
```
OlivierFreyssinet, line 25: Is this comment still necessary?

mitchellv-ledger: Good to have just to be safe in case someone tries to modify it by adding another step

apps/ledger-live-mobile/src/screens/SyncOnboarding/TwoStepStepper/FirstStepSyncOnboarding.tsx
```diff
@@ -0,0 +1,503 @@
...
+    );
+    setIsFinishedStep(true);
+    setTimeout(() => {
+      companionSteps.setStep(FirstStepCompanionStepKey.Exit);
+      setIsFinishedStep(false);
+      setHasFinishedExitAnimation(true);
+    }, 400);
+  }, [
+    handleFinishStep,
+    companionSteps,
```
OlivierFreyssinet: You should use `companionSteps.setStep` as a dependency instead of this object.
Due to the implem of `useCompanionSteps`, this object is new at each render, so the useCallback will have a new value on each render.

mitchellv-ledger: I had to destructure because otherwise it was still complaining about exhaustive-deps on not having the whole object

ofreyssinet-ledger (approved):

PR #11929, opened 2025-09-19
feature(coin-evm): pass calldata to intent

francois-guerin-ledger:
### ✅ Checklist

  - Minor refactoring should **not** introduce regression on send operations

### 📝 Description

Pass data inside EVM intent, needed while swaping

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -78,6 +78,19 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
   return res;
 }
 
+function computeIntentType(transaction: GenericTransaction): string {
+  if (!transaction.mode) return "Payment"; // NOTE: assuming payment by default here, can be changed based on transaction.mode
+
+  if (transaction.mode === "send" && transaction.family === "evm") {
+    return transaction.type === 2 ? "send-eip1559" : "send-legacy";
+  }
```
francois-guerin-ledger: During a swap, the plain transaction (coming from the swap live app ?) does not pack `"send-eip1559"` or `"send-legacy"` that are expected intent types for EVMs. Instead, it has `"send"` as its mode and `2` or `0` as its type to make the distinction. Hence this addition in the mapping

qperrot: thx for this comment I was wondering if we can avoid this function in the generic adapter !
But since you have mode set to `send` and this type can't we do this in evm coin-module ? 
I mean in ou alpaca function we can have a check if type exist and we have mode set to send then we change the mode depending of the type ?

libs/coin-framework/src/api/types.ts
```diff
@@ -264,18 +264,31 @@ export interface Memo {
   type: string;
 }
 
+export interface CallData {
```
hedi-edelbloute: I would use Data / TransactionData / TxData

CallData is eth specific

hedi-edelbloute (approved):

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
francois-guerin-ledger: I guess the cleanest would be, as usual, an optional method in `BridgeApi` that performs the needed transformation.

libs/coin-framework/src/api/types.ts
```diff
@@ -264,18 +264,31 @@ export interface Memo {
...
+}
+
 // Specialized version, not extending the above
 export interface StringMemo<Kind extends string = "text"> extends Memo {
   type: "string";
   kind: Kind;
   value: string;
 }
 
+export interface BufferTxData extends TxData {
```
jnicoulaud-ledger: Can we have some JSDoc explaining what this data is ? Is it EVM calldata, or a raw tx ? Does it take precedence over other parameters ?

libs/coin-modules/coin-evm/src/logic/common.ts
```diff
@@ -90,16 +91,22 @@ export function getErc20Data(recipient: string, amount: bigint): Buffer {
   return Buffer.from(data.slice(2), "hex");
 }
 
+function getCallData(intent: TransactionIntent<MemoNotSupported, BufferTxData>): Buffer {
```
jprudent, line 94: I feel like, the getErc20Data should be left to the caller to provide
- because the lib is sometimes computing it (erc20)
- sometimes it must be provided by caller

libs/coin-framework/src/api/types.ts
```diff
@@ -264,18 +264,31 @@ export interface Memo {
   type: string;
 }
 
+export interface TxData {
+  type: string;
```
jprudent, line 288: I don't know TS very well.
With other typed language, this would be a pleonasm because we can retrieve the type though the instance, like `isInstanceOf`

francois-guerin-ledger: We are dealing with `interface`s here, there is no such thing as `class`es instances.

libs/coin-modules/coin-evm/src/logic/common.ts
francois-guerin-ledger: Sorry I did not get it. What are you proposing ?

libs/coin-framework/src/api/types.ts
francois-guerin-ledger: > Is it EVM calldata, or a raw tx ?

It is **not** the final raw transaction. It is a computational payload used by blockchains, such as calldata on EVMs and instructions on Solana.

> Does it take precedence over other parameters ?

It does not take precedence over any parameters. It is just an extra optional property, really, very similar to the memo.

Note also that this is temporary solution, before it gets reworked [here](https://ledgerhq.atlassian.net/browse/LIVE-21638).

libs/coin-framework/src/api/types.ts
jprudent: yes interface would work too, but maybe not with TS (I don't know enough TS)

libs/coin-modules/coin-evm/src/logic/common.ts
jprudent: ok, sorry, what I'm proposing is 

```
function getCallData(intent: TransactionIntent<MemoNotSupported, BufferTxData>): Buffer {
  const data = intent.data?.value;
  return (Buffer.isBuffer(data) && data.length) ? data : Buffer.from([]);
```

and the `getErc20Data(intent.recipient, intent.amount)` must be provided by the caller.

Why ?
Because it's strange that in some cases the lib compute the data, and sometimes it doesn't.
If that doesn't make sense for you, let's Resolve the conversation

libs/coin-framework/src/api/types.ts
jnicoulaud-ledger: OK, got it (still think a little JSDoc would be nice ^^)

libs/coin-modules/coin-evm/src/logic/common.ts
francois-guerin-ledger: I think I see but my use case it really about sometimes we let `ethers` and sometimes we keep the input calldata. I keep this part as it.

jnicoulaud-ledger (approved):

PR #11774, opened 2025-09-11
chore(lld): Complete migration to WebHID and remove internal process architecture

gre-ledger:
tech debt / dead code

### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  - LLD only

### 📝 Description

This PR completes the migration from native `node-hid`/`usb` modules to WebHID (DeviceManagementKit) and eliminates the internal Electron process architecture. The motivation is to improve performance by removing blocking I/O operations, modernize to use WebHID standards, and significantly simplify the codebase architecture.

By doing this, we finally can remove the Internal process mechanism.

LLD is about 30 seconds faster to build (sub 2min on my mac).

## Changes Made

**Core Architecture**
- Force-enabled DeviceManagementKit (WebHID) as the only USB transport, removing the `ldmkTransport` feature flag
- Completely removed the internal Electron process and IPC communication layer
- Moved Speculos and HTTP proxy transports from internal process to renderer process with direct registration

**Dependencies & Build System**
- Removed all native module dependencies: `@ledgerhq/hw-transport-node-hid-singleton`, `@sentry/node`, `tree-kill`, and `@ledgerhq/native-modules-tools`
- Eliminated `beforePack.js` script and updated build configuration to remove native module processing
- Simplified `tools/main.js` and all `electron-builder*.yml` configurations

**Code Cleanup**
- Deleted entire internal process directory (`src/internal/*`) and related IPC infrastructure
- Removed transport channels, internal lifecycle management, and process communication code
- Simplified device listening hooks to use WebHID exclusively
- Cleaned up "kill internal process" concepts and unused Sentry exports

### ❓ Context

- **JIRA or GitHub link**: LIVE-21565

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #11955, opened 2025-09-22
chore(lld): remove unused node-hid and usb

gre-ledger:
a safer version of https://github.com/LedgerHQ/ledger-live/pull/11774 without reconsidering yet where Speculos runs: it still run from a node process.

### ✅ Checklist

  - LLD only

### 📝 Description

This PR modernizes Ledger Live Desktop by entirely removing the `usb` and `node-hid` dependency from the project, as we already migrated support of USB devices to WebHID, months ago. This approach delivers the performance benefits of removing native modules while maintaining proven IPC mechanisms for Speculos and HTTP proxy transports.

**The hybrid approach avoids the connectivity issues encountered in the previous complete migration attempt while still achieving the core performance goals.**

LLD is about 30 seconds faster to build (sub 2min on my mac).

### ❓ Context

- **JIRA or GitHub link**: LIVE-21565

---


gre-ledger:
https://github.com/LedgerHQ/ledger-live/actions/runs/17926702889

gre-ledger:
@LucasWerey what you saw in earlier comment of the bot was a false positive. building main bundle without the correct config tends to not bundle the actual lib (it was thinking everything is a node lib but then it would fail on start)
now everything should be good !

gre-ledger:
LLB builds https://github.com/LedgerHQ/ledger-live-build/actions/runs/17937524494

apps/ledger-live-desktop/src/main/transportHandler.ts
```diff
@@ -0,0 +1,179 @@
...
+      if (descriptor?.includes("speculos") || process.env.SPECULOS_API_PORT) {
+        // Speculos transport
+        const req: SpeculosHttpTransportOpts = {
+          apiPort: String(process.env.SPECULOS_API_PORT || "5000"),
+        };
+        transport = await SpeculosHttpTransport.open(req);
+      } else if (process.env.DEVICE_PROXY_URL) {
+        // HTTP proxy transport
+        const TransportHttpProxy = TransportHttp(process.env.DEVICE_PROXY_URL.split("|"));
+        transport = await TransportHttpProxy.create(timeout || 3000, timeout || 5000);
```
Justkant, line 37: I'm wondering if having the transport on the main thread could create some freezes in the UI when some interactions are running, but I suppose also that it could be fine if it's only an http server that node can handle in async, wdyt ?

gre-ledger: main thread isn't busy a lot, it's not managing the UI right? and speculos transport is mostly asynchronous.
forking an internal process was made because HID was blocking as well as libcore back at the time.

apps/ledger-live-desktop/src/renderer/components/MainSideBar/index.tsx
```diff
@@ -222,21 +222,9 @@ const TagContainerFeatureFlags = ({ collapsed }: { collapsed: boolean }) => {
...
-        pathname: "/settings/developer",
-      }}
-      onClick={() => setTrackingSource("sidebar")}
-    >
-      <Icons.UsbC size="S" color="primary.c80" />
-      <TagText collapsed={collapsed}>{t("common.ldmkEnabled")}</TagText>
-    </Tag>
-  ) : null;
+// WebHID is now the default transport, no need for warning tag
+const TagContainerLDMK = ({ collapsed: _collapsed }: { collapsed: boolean }) => {
```
OlivierFreyssinet, line 226: This component can be removed

PR #11981, opened 2025-09-24
fkeat: LLD Mixpanel session replay swap flow

RobinVncnt:
### ✅ Checklist

- [ ] **Covered by automatic tests.** If we run tests, session replays will be sent every time, and if we mock them, the tests wouldn’t be relevant
  -  Add feature flag `lldSessionReplay`  
  - Start recording from the app launch  
  - Send session replays to Mixpanel  

### 📝 Description

This PR introduces a new feature flag lldSessionReplay that enables session replay recording.
When enabled, the app will start recording sessions from the very beginning of the app lifecycle and send them to Mixpanel.
The main goal is to record the swap flow. Since it cannot be fully recorded from Ledger Live alone, we will link two recordings together with distinctId : one from Ledger Live and one from the Swap Live app.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-21530]

https://github.com/user-attachments/assets/a8d6dafa-75ac-4c5d-8ef2-9a83cad049f5


libs/ledgerjs/packages/types-live/src/feature.ts
```diff
@@ -229,6 +230,9 @@ export type Features = CurrencyFeatures & {
   lldNetworkBasedAddAccount: DefaultFeature;
   llmOfacGeoBlocking: DefaultFeature;
   lldOfacGeoBlocking: DefaultFeature;
+  lldSessionReplay: Feature<{
+    sampling: number;
+  }>;
```
mcayuelas-ledger: You can type it like that 

`export type Feature_LldSessionReplay = Feature<{
  sampling: number;
}>;`

apps/ledger-live-desktop/src/renderer/Default.tsx
```diff
@@ -220,6 +222,10 @@ export default function Default() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dmk]);
 
+  useEffect(() => {
+    initMixpanel(mixpanelFF?.params?.sampling || 100);
```
mcayuelas-ledger: instead of doing like that, you should add default value on `initMixpanel(sampling=100)`

Also you never use `mixpanelFF.enable`?

mcayuelas-ledger (changes requested):

apps/ledger-live-desktop/src/renderer/analytics/mixpanel.ts
```diff
@@ -0,0 +1,12 @@
+import mixpanel from "mixpanel-browser";
+
+const API_HOST = "https://api.mixpanel.com";
```
mcayuelas-ledger: you can add it in `libs/env/src/env.ts`

apps/ledger-live-desktop/src/renderer/analytics/mixpanel.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const TOKEN = "9a6672bcb9778cb9ea6e6ba671690d66";
```
mcayuelas-ledger: We should never add token like that, it can introduce security issue, 

you need to add it to the project and use `env` var

mcayuelas-ledger (changes requested):

mcayuelas-ledger (approved):

apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/SwapWebViewDemo3.tsx
```diff
@@ -151,6 +152,14 @@ const SwapWebView = ({ manifest }: SwapWebProps) => {
   const isOffline = networkStatus === NetworkStatus.OFFLINE;
   // Remove after KYC AB Testing
   const ptxSwapLiveAppKycWarning = useFeature("ptxSwapLiveAppKycWarning")?.enabled;
+  const [distinctId, setDistinctId] = useState<string>("");
+
+  useEffect(() => {
+    const id = getMixpanelDistinctId();
+    if (id) {
+      setDistinctId(id);
```
liviuciulinaru: Since `getMixpanelDistinctId()` is synchronous and fast, using `useState` and `useEffect` is unnecessary and adds a slight performance overhead due to the extra re-render it causes.

The cleanest solution is to call the function directly.

```js 
const distinctId = getMixpanelDistinctId() || "";
```

PR #12010, opened 2025-09-25
fix: issue not showing correct storage limit value on Tezos

dilaouid:
### ✅ Checklist

  - ...

### 📝 Description

Tezos was not showing the proper storage limit value on a transaction (always frozen on 300)

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-alpaca/prepareTransaction.ts
+++ b/libs/ledger-live-common/src/bridge/generic-alpaca/prepareTransaction.ts
@@ -33,8 +33,8 @@ export function genericPrepareTransaction(
     const { assetReference, assetOwner } = getAssetFromToken
       ? getAssetInfos(transaction, account.freshAddress, getAssetFromToken)
       : assetInfosFallback(transaction);
-
-    let fees: BigNumber | bigint | null = transaction.customFees?.parameters?.fees || null;
+    const customParametersFees = transaction.customFees?.parameters?.fees;
+    let fees: BigNumber | bigint | null = customParametersFees || null;
     if (fees === null) {
       fees = (
         await estimateFees(
@@ -57,7 +57,7 @@ export function genericPrepareTransaction(
         assetOwner,
         customFees: {
           parameters: {
-            fees: new BigNumber(fees.toString()),
+            fees: customParametersFees ? new BigNumber(fees.toString()) : undefined,
           },
         },
       };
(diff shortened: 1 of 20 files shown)


libs/ledger-live-common/src/bridge/generic-alpaca/prepareTransaction.ts
```diff
@@ -57,7 +57,7 @@ export function genericPrepareTransaction(
         assetOwner,
         customFees: {
           parameters: {
-            fees: new BigNumber(fees.toString()),
+            fees: customParametersFees ? new BigNumber(fees.toString()) : undefined,
```
francois-guerin-ledger, line 60: Ternaries with expressions that do not depend on the condition are hard to read.
```suggestion
            fees: fees ? new BigNumber(fees.toString()) : undefined,
```

Also I don't think you need to introduce a new variable `customParametersFees` if it used only once.

Finally it is not obvious it fixes the storage limit issue if you are modifying the `fees` properties. A test could help here.

francois-guerin-ledger (changes requested):

libs/ledger-live-common/src/bridge/generic-alpaca/prepareTransaction.ts
dilaouid: Actually `customParametersFees` is used twice (ref line 37 and line 60).

The ternary is not checking the `fees` variable but the original `customFees` parameter to preserve the correct behavior: when `customFees` doesn't exist for a coin module (like Tezos in this case), we set `customFees.parameters.fees` to `undefined`.
This ensures we don't skip the condition `if (!bnEq(transaction.fees, new BigNumber(fees.toString())))` and allows `estimateFees` to be called twice - which is essential for Tezos to get the correct `storageLimit` from the second estimation with real recipient/amount values.

Added some tests demonstrate this fixes the storageLimit issue by showing difference between fallback  / real  estimation.

libs/ledger-live-common/src/bridge/generic-alpaca/tests/prepareTransaction.test.ts
```diff
@@ -91,4 +91,82 @@ describe("genericPrepareTransaction", () => {
 
     expect(result).toBe(baseTransaction); // still same reference
   });
+
+  it("propagates storageLimit from second estimation", async () => {
```
francois-guerin-ledger, line 95: The only difference between those two tests is second parameters being `{ storageLimit: 0 }` or `{ storageLimit: 277 }`. Could we compact them with `it.each` ?

libs/ledger-live-common/src/bridge/generic-alpaca/prepareTransaction.ts
Moustafa-Koterba: @francois-guerin-ledger not really agree on Ternary condition, but it is a matter of taste I think

The logic here is to say that if we have a custom fees (here `customParametersFees`), we take the estimated fees (here `fees`) if I am not wrong @dilaouid ? If so, the code look good to me

PR #12023, opened 2025-09-26
Feat/live 21658 zondax icp retire rosetta

scavassa-yld:
### ✅ Checklist

  - ...

### 📝 Description

Replaces rosetta api with canister support 
(development by Zondax)

### ❓ Context

- [**JIRA or GitHub link**: ](https://ledgerhq.atlassian.net/browse/LIVE-21658)

---


LucasWerey (approved):
✅ WXP scope

libs/coin-modules/coin-internet_computer/src/api/api.ts
```diff
@@ -1,92 +1,148 @@
...
-    account_identifier: {
-      address: accountId,
-      metadata: {},
+export const broadcastTxn = async (
+  payload: Buffer,
+  canisterId: string,
+  type: "call" | "read_state",
+) => {
+  log("debug", `[ICP] Broadcasting ${type} to ${canisterId}, body: ${payload.toString("hex")}`);
+  const res = await fetch(`${ICP_NETWORK_URL}/api/v2/canister/${canisterId}/${type}`, {
```
acewf: it seems icp has available v3, any particular reason to go for v2?

not related with this specific PR, but seems v3 has openapi `https://ic-api.internetcomputer.org//api/v3/openapi.json` which is great for using with https://openapi-ts.dev/

libs/coin-modules/coin-internet_computer/src/api/api.ts
```diff
@@ -1,92 +1,148 @@
...
+    headers: {
+      "Content-Type": "application/cbor",
     },
-  };
-  const data = await ICPFetchWrapper<ICPRosettaGetBalancesResponse>("account/balance", body);
-  return data;
+  });
+
+  // If the status is not 2XX, throw an error
+  if (res.status >= 400) {
```
Moustafa-Koterba: we should check here the documentation and see the possible responses
Is is better to handle possible responses from documentation and throw an error otherwise
to avoid any non expected behaviour
```suggestion
 if (res.status === 200) {
   return await res.arrayBuffer();
 } 
 
  throw new Error(`Failed to broadcast transaction: ${res.text()}`);
```
If you have a 3XX response or other, you may return a not handled answer

libs/coin-modules/coin-internet_computer/src/bridge/getTransactionStatus.ts
Moustafa-Koterba, line 1: sorry for the dumb question, why did you add this imports ? same for next file

libs/coin-modules/coin-internet_computer/src/types/signer.ts
```diff
@@ -6,6 +6,15 @@ export type ICPSignature = {
   signatureDER?: Buffer;
 };
 
+export type ICPSignUpdateCall = {
```
Moustafa-Koterba: same for this, I don't see it used somewhere, did I miss something ?

libs/coin-modules/coin-internet_computer/src/bridge/getTransactionStatus.ts
lawRathod: No not a dumb question at all, you mean imports from the new `@zondax/ledger-live-icp` package? 

It's because some @dfinity dependencies are used for validation and since there are already available on the dep it made sense to keep all utils there. There's a little quirk which we are handling in the package which therefore lets this integration depend only on one dep and none of the dep of dep. This will be a lot more apparent in future staking PR for ICP.

libs/coin-modules/coin-internet_computer/src/types/signer.ts
lawRathod: Nice catch, this signer call isn't being used yet. I can remove this type.

libs/coin-modules/coin-internet_computer/src/api/api.ts
lawRathod: I agree, I can incorporate your suggestion.

libs/coin-modules/coin-internet_computer/src/api/api.ts
lawRathod: We can use v3, it's pretty much the same behavior. We don't want to keep using endpoint directly and use agent lib, this will be changed in an upcoming PR very shortly for supporting staking on ICP.

scavassa-yld: @acewf  could you please review ?

libs/coin-modules/coin-internet_computer/src/api/api.ts
```diff
@@ -1,92 +1,148 @@
...
+    method: "POST",
+    headers: {
+      "Content-Type": "application/cbor",
     },
-  };
-  const data = await ICPFetchWrapper<ICPRosettaGetBalancesResponse>("account/balance", body);
-  return data;
+  });
+
+  // If the status is not 2XX, throw an error
```
Moustafa-Koterba: you can remove the comment, the code is readable

pnpm-lock.yaml
```diff
@@ -31497,6 +31602,19 @@ packages:
...
+  request-promise-core@1.1.4:
+    resolution: {integrity: sha512-TTbAfBBRdWD7aNNOoVOBH4pN/KigV6LyapYNNlAPA8JwbovRti1E88m3sYAwsLi5ryhPKsE9APwnjFTgdUjTpw==}
+    engines: {node: '>=0.10.0'}
+    peerDependencies:
+      request: ^2.34
+
+  request-promise-native@1.0.9:
+    resolution: {integrity: sha512-wcW+sIUiWnKgNY0dqCpOZkUbF/I+YPi+f09JZIDa39Ec+q82CpSYniDp+ISgTTbKmnpJWASeJBPZmoxH84wt3g==}
+    engines: {node: '>=0.12.0'}
+    deprecated: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142
```
gre-ledger: could you explain why there are so many new library added in deps? also note that this one is deprecated

libs/coin-modules/coin-internet_computer/src/bridge/getTransactionStatus.ts
Moustafa-Koterba: Ok my bad, I did not see that they were used, in the file. Github only displayed the import change
So they were missing actually, right ? look good to me

Moustafa-Koterba (approved):

libs/coin-modules/coin-internet_computer/src/api/api.ts
```diff
@@ -1,92 +1,148 @@
...
-  if (!baseUrl) throw new Error("API base URL not available");
-
-  return `${baseUrl}${path ? path : ""}`;
-};
-
-const ICPFetchWrapper = async <T>(path: string, body: any) => {
-  const url = getICPURL(path);
+import {
+  ledgerIdlFactory as ledgerIdlFactory,
+  indexIdlFactory as indexIdlFactory,
```
LucasWerey: Both named imports aren't needed, I think

libs/coin-modules/coin-internet_computer/src/bridge/getTransactionStatus.ts
lawRathod: Yes exactly. We are trying to incorporate all ICP specific logic on `@zondax/ledger-live-icp` so we can reduce clutter in ledger-live codebase.

pnpm-lock.yaml
lawRathod: sorry most came from `browser-env` that wasn't being used in the dependency on this PR. Removed it from dep and there should be way less now.

libs/coin-modules/coin-internet_computer/src/api/api.ts
lawRathod: done!

libs/coin-modules/coin-internet_computer/src/api/api.ts
lawRathod: yes, crazy it was there 😓. Removed!

pnpm-lock.yaml
gre-ledger: I still see it in your current PR, is it expected? thanks

gre-ledger (approved):

pnpm-lock.yaml
lawRathod: ahh, we are struggling to merge the fix, it's in https://github.com/LedgerHQ/ledger-live/pull/12195

libs/coin-modules/coin-internet_computer/src/api/api.ts
```diff
@@ -1,92 +1,147 @@
...
+import invariant from "invariant";
 
-  // We force data to this way as network func is not using the correct param type. Changing that func will generate errors in other implementations
-  const opts: AxiosRequestConfig = {
-    method: "POST",
-    data: body,
-    url,
+export const fetchBlockHeight = async (): Promise<BigNumber> => {
+  const canisterId = Principal.fromText(MAINNET_LEDGER_CANISTER_ID);
+  const queryBlocksRawRequest = {
```
acewf, line 26: this is a constant that could be defined outside of function

gre-ledger:
> ⚠️ renderer bundle size significantly increased: 46.3mb -> 54.9mb. Please check if this is expected.

oops, sadly i didn't see at the time. but this is not acceptable IMO. any way to solve this regression?
until we have lazy loading for coin modules, this means it will likely create performance regression at boot time :'(

francois-guerin-ledger:
> any way to solve this regression?

The underlying package does support modular imports: https://www.npmjs.com/package/@zondax/ledger-live-icp#quick-start. Let's check if it helps first

PR #12042, opened 2025-09-29
feat: (LIVE-21558) Add A/B test for desktop ‘Confirm on Device’ flow (details vs no-details)

ooke-ledger:
### ✅ Checklist

  - ...

### 📝 Description

This PR Implements A/B test: transaction details vs simplified confirm prompt (desktop)

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-21558

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
@@ -6,6 +6,7 @@ import { connect, useDispatch } from "react-redux";
 import { useHistory } from "react-router-dom";
 import styled from "styled-components";
 import { CryptoCurrency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
+import { ABTestingVariants } from "@ledgerhq/types-live";
 import ProviderIcon from "~/renderer/components/ProviderIcon";
 import { Transaction } from "@ledgerhq/live-common/generated/types";
 import { ExchangeRate, ExchangeSwap } from "@ledgerhq/live-common/exchange/swap/types";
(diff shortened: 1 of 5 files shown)


apps/ledger-live-desktop/src/renderer/screens/exchange/Swap2/Form/DrawerFooter.tsx
```diff
@@ -32,7 +32,9 @@ export function DrawerFooter({ provider }: { provider: string }) {
   const ptxSwapLiveAppKycWarning = useFeature("ptxSwapLiveAppKycWarning");
   const url = providerData?.termsOfUseUrl;
   const providerName = getProviderName(provider);
-  const isDetailedViewEnabled = useFeature("ptxSwapDetailedView");
+  const ptxSwapDetailedView = useFeature("ptxSwapDetailedView");
+  const isDetailedViewEnabled =
+    ptxSwapDetailedView?.enabled && ptxSwapDetailedView?.params?.variant === "ptxdrawerdetails";
```
philipptpunkt: I don't think the double check here is necessary 🤔 Not sure how ab-tests are set up on LL, but technically you should get a `true`/`false` from `enabled`, which toggles the feature and the variant is only used for tracking as a parameter in the event 🤔

libs/ledger-live-common/src/featureFlags/defaultFeatures.ts
```diff
@@ -127,6 +126,12 @@ export const DEFAULT_FEATURES: Features = {
   brazePushNotifications: initFeature(),
   stakeAccountBanner: initFeature(),
 
+  ptxSwapDetailedView: initFeature({
+    enabled: false,
+    params: {
+      variant: "ptxdrawerdetails",
```
LucasWerey: You can use ABTestingVariants for this

ooke-ledger: ABTestingVariants is currently restricted to "A" and "B", but "ptxdrawerdetails" was requested as the variant name for clearer analytics tracking.

ofreyssinet-ledger (changes requested):
Please update the PR description and the changeset to include more context and explain exactly what this PR is doing, this is too vague.

libs/ledger-live-common/src/featureFlags/defaultFeatures.ts
philipptpunkt: You could use the default `ABTestingVariants` and then map to the actual variant name in the tracking event.
If you go with the custom variant names, I'm not sure I understand the naming for the default. `ptxdrawerdetails` sounds like you show the details, put the flag is set to `false`.
What's the other variant called?

apps/ledger-live-desktop/src/renderer/components/DeviceAction/rendering.tsx
```diff
@@ -1170,9 +1171,19 @@ const SwapDeviceConfirmation: React.FC<SwapConfirmationProps> = ({
...
   walletState,
 }) => {
-  const isDetailedViewEnabled = useFeature("ptxSwapDetailedView")?.enabled;
+  const ptxSwapDetailedView = useFeature("ptxSwapDetailedView");
+  const isDetailedViewEnabled = !!ptxSwapDetailedView?.enabled;
+  const variant = ptxSwapDetailedView?.params?.variant ?? ABTestingVariants.variantA;
   const sourceAccountCurrency = exchange.fromCurrency;
   const targetAccountCurrency = exchange.toCurrency;
+
+  useEffect(() => {
```
philipptpunkt: before we add logic here to make sure we fire this event only once, I think it might make sense to just track the variant on the Page Tracking event, no?! Maybe check with Product people, but imho this would be the cleanest solution

philipptpunkt (approved):

PR #12057, opened 2025-09-29
:bug: (LLM) Cannot read property 'params' of undefined sentry

LucasWerey:
### ✅ Checklist

- [] **Covered by automatic tests.** 
  - accountsNavigator

### 📝 Description

I'm not able to repro nor find a common behavior. It looks like it happens only on android
I've added a typeguard so we don't throw this error. 
As the error mentions params undefined, it shouldn't be an issue in prod it will just change the navigation

### ❓ Context

- **JIRA or GitHub link**: [LIVE-21864](https://ledgerhq.atlassian.net/browse/LIVE-21864)

---

### DIFF
--- a/apps/ledger-live-mobile/src/components/RootNavigator/AccountsNavigator.tsx
+++ b/apps/ledger-live-mobile/src/components/RootNavigator/AccountsNavigator.tsx
@@ -37,6 +37,11 @@ type ParamsType = {
   params?: { specificAccounts?: object[] };
 };
 
+const isParamsType = (value: unknown): value is ParamsType =>
+  typeof value === "object" &&
+  value !== null &&
+  Object.prototype.hasOwnProperty.call(value, "params");
+
 export default function AccountsNavigator() {
   const { colors } = useTheme();
   const stackNavConfig = useMemo(() => getStackNavigatorConfig(colors), [colors]);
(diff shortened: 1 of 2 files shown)


apps/ledger-live-mobile/src/components/RootNavigator/AccountsNavigator.tsx
```diff
@@ -37,6 +37,11 @@ type ParamsType = {
   params?: { specificAccounts?: object[] };
 };
 
+const isParamsType = (value: unknown): value is ParamsType =>
+  typeof value === "object" &&
+  value !== null &&
+  Object.prototype.hasOwnProperty.call(value, "params");
```
ejaycoleman, line 43: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

Prefer `Object.hasOwn(obj, prop)`?

ejaycoleman (approved):

apps/ledger-live-mobile/src/components/RootNavigator/AccountsNavigator.tsx
LucasWerey: We'll still be in Chrome, so I guess we don't really have to worry, no?

PR #12060, opened 2025-09-29
:package: (llm) remove react-native-extra-dimensions-android

LucasWerey:
### ✅ Checklist

  - recover webview
  - buy device in live webview
  - all getWindowDimensions usage

### 📝 Description

react-native-extra-dimensions-android lib is 6 years old and not maintained. It needs a patch to be compatible with gradle8 
As we use it for only for few use cases, it can be dropped and replaced simply with a safeareview and flexGrow / react native Dimensions

This is needed, otherwise we can't build locally on android :( 

Concerned screens: 

Ledger Live shop 

image

image

Recover 

image

image

### ❓ Context

- **JIRA or GitHub link**: [LIVE-21867](https://ledgerhq.atlassian.net/browse/LIVE-21867)

---

[LIVE-21867]: https://ledgerhq.atlassian.net/browse/LIVE-21867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/logic/getWindowDimensions.android.ts
+++ b/apps/ledger-live-mobile/src/logic/getWindowDimensions.android.ts
@@ -1,7 +1,9 @@
-import ExtraDimensions from "react-native-extra-dimensions-android";
+import { Dimensions } from "react-native";
 
-export default () => ({
-  width: ExtraDimensions.get("REAL_WINDOW_WIDTH"),
-  height: ExtraDimensions.get("REAL_WINDOW_HEIGHT"),
-});
-export const softMenuBarHeight = () => ExtraDimensions.get("SOFT_MENU_BAR_HEIGHT");
+export default () => {
+  const screen = Dimensions.get("screen");
+  return {
+    width: screen.width,
+    height: screen.height,
+  };
+};
--- a/.changeset/itchy-spiders-drum.md
+++ b/.changeset/itchy-spiders-drum.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": minor
+---
+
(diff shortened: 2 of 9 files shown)


apps/ledger-live-mobile/src/logic/getWindowDimensions.android.ts
```diff
@@ -1,7 +1,9 @@
...
-  width: ExtraDimensions.get("REAL_WINDOW_WIDTH"),
-  height: ExtraDimensions.get("REAL_WINDOW_HEIGHT"),
-});
-export const softMenuBarHeight = () => ExtraDimensions.get("SOFT_MENU_BAR_HEIGHT");
+export default () => {
+  const screen = Dimensions.get("screen");
+  return {
+    width: screen.width,
+    height: screen.height,
+  };
```
Justkant, line 8: We could directly return screen here to avoid new allocations
```suggestion
  return screen;
```

We could also rename this as getScreenDimensions to not confuse with `Dimensions.get("window")` but it means changing a lot of imports and I understand if we want to keep the PR smaller

Justkant (approved):
LGTM, I added a small suggestion

apps/ledger-live-mobile/src/logic/getWindowDimensions.android.ts
LucasWerey: Indeed but for IOS it uses window because it's not the same behaviour on both platforms
So it will be confusing also with IOS 

The dif between iOS and Android is explained here https://stackoverflow.com/a/75234700

PR #12067, opened 2025-09-29
LIVE-21715 Investigate 500Mb coin monitoring reports and fix

Moustafa-Koterba:
### ✅ Checklist

  - No impact

### 📝 Description

Investigate 500Mb coin monitoring reports and fix

## The issue
The issue come mainly from the imports, which take like 300Mb of the memory

## The solution
Make a difference on the memory between the start of the process and the execution of the sync/scan

## What changed
- The CLI take now 3 parameters: currencies, account type and isolated. We can pass now an array of arguments like
```sh
--currencies algorand,bitcoin,ethereum --account-types pristine,average --isolated
```
- The measurement run now by currency and account type: we run the script by combination sequentially. The aim is to not have any false measure with caches or future modifications
- The script run on a transpiled version
- The measurement of the memory run at higher rate to have better statistics (every 30ms, like 33 FPS 🎮) 

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-21715?atlOrigin=eyJpIjoiMWFkODQ3ZTJlNWFiNDZmM2E2ZjUzMzE5ZTkzMWE4MGMiLCJwIjoiaiJ9

---

### DIFF
--- a/libs/coin-modules-monitoring/src/cli.ts
+++ b/libs/coin-modules-monitoring/src/cli.ts
@@ -160,11 +160,16 @@ async function runMonitorCommand(options: CommandLineOptions) {
   }
 
   if (options.isolated) {
+    let atLeastOneProcessFailed = false;
     for (const currency of monitoredCurrencies) {
       for (const accountType of accountTypes) {
         try {
           await runIsolatedMonitor(currency, accountType);
         } catch (error) {
+          if (!atLeastOneProcessFailed) {
+            atLeastOneProcessFailed = true;
+          }
+
           console.error(
             `Process failed for currency ${currency} and account type ${accountType} with error ${error}`,
           );
@@ -173,7 +178,7 @@ async function runMonitorCommand(options: CommandLineOptions) {
       }
     }
 
-    process.exit(0);
+    process.exit(atLeastOneProcessFailed ? 1 : 0);
   }
 
   try {


.github/workflows/coin-monitoring-manual.yml
```diff
@@ -5,6 +5,9 @@ permissions:
 on:
   workflow_dispatch:
     inputs:
+      account_types:
+        description: comma-separated account types, separated with commas (,) (pristine, average, big)
```
francois-guerin-ledger: I would keep the same wording as below for homogeneity purposes

.github/workflows/coin-monitoring-scheduled.yml
```diff
@@ -15,14 +15,15 @@ jobs:
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-caches@develop
         with:
           install-proto: true
-      - uses: LedgerHQ/ledger-live/tools/actions/composites/monitor@develop
+      - uses: LedgerHQ/ledger-live/tools/actions/composites/monitor
```
francois-guerin-ledger: Why this change ? Which branch is used then ?

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,50 +1,98 @@
-import currencies from "./currencies";
+import { Command } from "commander";
```
francois-guerin-ledger: I prefer avoiding the extra dependency of `commander`. None of our CLI use it, and having it just for the coin module monitoring does not seem justified.

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,50 +1,98 @@
...
-    ? supportedCurrencies
-    : argv2
-        .split(",")
-        .map(c => c.trim())
-        .filter(c => supportedCurrencies.includes(c));
-const monitoredCurrencies = filteredCurrencies.filter(c => {
-  const currency = currencies[c];
-  if (currency?.skip) {
-    console.log(`Skipping "${c}". Reason: ${currency.skip}`);
+const VALID_ACCOUNT_TYPES: string[] = ["pristine", "average", "big"] as const;
```
francois-guerin-ledger: Why specifying `as const` to restrict the values and type with the more flexible `string[]` ?

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,50 +1,98 @@
...
+  const types = value.split(",").map(t => t.trim());
+  const invalidTypes = types.filter(t => !VALID_ACCOUNT_TYPES.includes(t));
+
+  if (invalidTypes.length > 0) {
+    throw new Error(
+      `Invalid account types: ${invalidTypes.join(", ")}. Valid types are: ${VALID_ACCOUNT_TYPES.join(", ")}`,
+    );
+  }
+
+  return types as AccountType[];
```
francois-guerin-ledger: Consider using a guard instead of manually casting

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,50 +1,98 @@
...
-          operation: log.operationType,
-          "duration (ms)": log.duration,
-          "network calls": log.totalNetworkCalls,
-        })),
-      );
+const program = new Command();
+program
+  .name("coin-modules-monitoring")
+  .description("Monitor cryptocurrency modules and push metrics to Datadog")
+  .version("2.3.1");
```
francois-guerin-ledger: If we _really_ want to output the version, better reading the one from `package.json`.

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,50 +1,98 @@
...
+    "-t, --account-types <types>",
+    `comma-separated account types (${VALID_ACCOUNT_TYPES.join(", ")})`,
+    "pristine,average,big",
+  )
+  .option(
+    "-c, --currencies <currencies>",
+    `comma-separated currencies (${VALID_CURRENCIES.slice(0, 5).join(", ")}...)`,
+    "all",
+  )
+  .action(async options => {
```
francois-guerin-ledger: Your callback is `async`, maybe use `await` instead of the `then ... catch`. pattern ?

libs/coin-modules-monitoring/src/measure.ts
```diff
@@ -63,11 +63,13 @@ function setImmediateInterval(fn: () => unknown, ms: number) {
 
 async function measureDuring<T>(fn: () => Promise<T>) {
   const cpuSamples: number[] = [];
-  const memSamples: number[] = [];
+  const memSamples: Set<number> = new Set();
```
francois-guerin-ledger: Why a `Set` ? Won't it have an impact in percentile calculuses ?

libs/coin-modules-monitoring/src/measure.ts
```diff
@@ -77,22 +79,22 @@ async function measureDuring<T>(fn: () => Promise<T>) {
...
+
+    const { rss } = process.memoryUsage();
+    memSamples.add(bytesToMB(Math.max(rss - previousMemory, 0)));
 
     prevCpu = curCpu;
     prevTs = now;
   };
 
-  const timer = setImmediateInterval(tick, 100);
+  const timer = setImmediateInterval(tick, 30);
```
francois-guerin-ledger, line 97: Every 100ms is not frequent enough you think ?

libs/coin-modules-monitoring/src/run.test.ts
```diff
@@ -42,10 +42,14 @@ describe("Coin Modules Monitoring", () => {
...
 
-      const logs = await run(["solana"]);
+      const logs = await run(["solana"], ["pristine", "average", "big"]);
 
       expect(logs).toEqual({
-        entries: [
+        entries: expect.any(Array),
+        failed: true,
+      });
+      expect(logs.entries).toEqual(
```
francois-guerin-ledger: We can merge the two `expect`ations.

francois-guerin-ledger (changes requested):
I don't think we address the real problem here. If we simply run
```sh
% pnpm coin:modules:monitoring start monitor --currencies algorand

> @ledgerhq/ledger-live@0.0.0 coin:modules:monitoring /Users/francois.guerin/Documents/ledger-live
> pnpm --filter coin-modules-monitoring "start" "monitor" "--currencies" "algorand"


> @ledgerhq/coin-modules-monitoring@2.4.0 start /Users/francois.guerin/Documents/ledger-live/libs/coin-modules-monitoring
> node lib/cli.js "monitor" "--currencies" "algorand"

Monitoring 3 account(s) within algorand

[1 / 3] Start (currency = "algorand" account = "pristine")
(node:11354) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

[1 / 3] ✅ Completed in 338 ms
 ┌─ 🔎 Scan
 │  • Calls: Total: 2 - algorand.coin.ledger.com: 2
 │  • CPU  : min=0%, median=7.342%, max=127.365%, p90=127.365%, p99=127.365%
 │  • MEM  : min=0 MB, median=6.594 MB, max=7.438 MB, p90=7.438 MB, p99=7.438 MB
 └─ 🔄 Sync
    • Calls: Total: 2 - algorand.coin.ledger.com: 2
    • CPU  : min=0%, median=9.139%, max=16.097%, p90=16.097%, p99=16.097%
    • MEM  : min=0 MB, median=0.094 MB, max=0.203 MB, p90=0.203 MB, p99=0.203 MB


[2 / 3] Start (currency = "algorand" account = "average")

[2 / 3] ✅ Completed in 1.387s
 ┌─ 🔎 Scan
 │  • Calls: Total: 8 - algorand.coin.ledger.com: 8
 │  • CPU  : min=0%, median=0.511%, max=45.442%, p90=22.803%, p99=45.442%
 │  • MEM  : min=0 MB, median=0.523 MB, max=1.672 MB, p90=1.578 MB, p99=1.672 MB
 └─ 🔄 Sync
    • Calls: Total: 2 - algorand.coin.ledger.com: 2
    • CPU  : min=0%, median=6.944%, max=34.503%, p90=34.503%, p99=34.503%
    • MEM  : min=0 MB, median=1.781 MB, max=1.797 MB, p90=1.797 MB, p99=1.797 MB


[3 / 3] Start (currency = "algorand" account = "big")

[3 / 3] ✅ Completed in 2.575s
 ┌─ 🔎 Scan
 │  • Calls: Total: 19 - algorand.coin.ledger.com: 19
 │  • CPU  : min=0%, median=0.494%, max=67.052%, p90=28.459%, p99=67.052%
 │  • MEM  : min=0 MB, median=0.484 MB, max=2.641 MB, p90=2.516 MB, p99=2.641 MB
 └─ 🔄 Sync
    • Calls: Total: 2 - algorand.coin.ledger.com: 2
    • CPU  : min=0%, median=0.483%, max=28.665%, p90=28.665%, p99=28.665%
    • MEM  : min=0 MB, median=1.25 MB, max=1.266 MB, p90=1.266 MB, p99=1.266 MB
```
we still have the current behaviour, correct ? That is, the first account takes more memory than the other ones..
Instead, I am afraid we more likely count on the fact the script is consumed in a specific way to print trusty results.

Moustafa-Koterba:
> I don't think we address the real problem here

Actually, it does when running by coin and account type on the CI
That what we agreed on the discussing on Slack, or I have misunderstood may be ? 😄 
The PR fix 2 issues:
- Have the correct value for memory (in measure.ts)
```ts
memSamples.add(bytesToMB(Math.max(rss - previousMemory, 0)));
```
- Avoid cache issue (the problem you said), the workflows on CI have been updated to run by combination of `currency` and `account type` 

So, I just updated the script to be able to run a combination of these parameters, but you can still run everything at the same time, and indeed you will have the cache issue

Do you want me to avoid this and block to run everything ? I don't think this is a big deal actually, this library is just used to collect data on coins

We can specify in documentation why we did this 👍

.github/workflows/coin-monitoring-manual.yml
Moustafa-Koterba: Can you precise please ? Sorry I dont get it
You want something like
```suggestion
        description: Account types to monitor, separated with commas (,), or "pristine, average, big" (default)
```

.github/workflows/coin-monitoring-scheduled.yml
Moustafa-Koterba: an error left from my test, fixed

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: My bad, you're right, fixed

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: done

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: fixed

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: done

libs/coin-modules-monitoring/src/measure.ts
Moustafa-Koterba: a leftover from previous test, fixed

libs/coin-modules-monitoring/src/measure.ts
Moustafa-Koterba: I wanted more values to have more accurate data
Do you want me to rollback ?

libs/coin-modules-monitoring/src/run.test.ts
Moustafa-Koterba: done

libs/coin-modules-monitoring/src/measure.ts
francois-guerin-ledger: No need to rollback, the motivation you gave is sufficient 👍

.github/workflows/coin-monitoring-manual.yml
francois-guerin-ledger: Yes exactly. Otherwise the wording mismatch is weird within the same contextual menu

francois-guerin-ledger:
> Actually, it does when running by coin and account type on the CI
> [...]
> [...] but you can still run everything at the same time, and indeed you will have the cache issue

That's the point. It does under certain conditions. That module should act the same, independently of how it is used.

> That what we agreed on the discussing on Slack, or I have misunderstood may be ?

Looking at our Slack channel, I don't think we ever agreed on anything. I just gave indications on what to do to run the script once per account, as it was a suggestion of yours. Sorry if it was unclear.

> Do you want me to avoid this and block to run everything ?

No, I think it would be a regression to add such a big restriction

To summarize, my opinion did not change (and is even stronger now that I am aware of the inconsistency is creates)

- We want trustable output from this module, no matter how we use it
- Allowing only one account per run is very restrictive. Monitoring multiple accounts will become painful

We need a solution to address the real problem in any cases.

Moustafa-Koterba:
Oh my bad, I understand our conversation as you were agree on this, and I did the work, sorry for that
So I understand your point and it is true, it is better to have a consistent behaviour but in our case, I thought it was better to not handle the cache issue and avoid future problem on it if the scan and sync evolve

May be it is better to ask our team members to choose between this 2 solutions (or just @Justkant as he is Staff engineer)
I will start a Slack conversation with him and you and explain the issue

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: Removed, use `command-line-args` package used in LL CLI

.github/workflows/coin-monitoring-manual.yml
francois-guerin-ledger: You prefer keeping this wording ? Fine but at least explain please.

.vscode/launch.json
```diff
@@ -1,6 +1,7 @@
 {
   "version": "0.2.0",
   "configurations": [
+
```
francois-guerin-ledger:
```suggestion
```

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,30 +1,176 @@
...
+    }
+    return !currency.skip;
+  });
+
+  if (!monitoredCurrencies.length) {
+    console.log("No currencies to monitor. Exit now.");
+    process.exit(0);
+  }
+
+  if (options.isolated === true) {
```
francois-guerin-ledger:
```suggestion
  if (options.isolated) {
```

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,30 +1,176 @@
...
+
+function isValidAccountType(value: string): value is AccountType {
+  return VALID_ACCOUNT_TYPES.includes(value as AccountType);
+}
+
+function validateAccountTypes(value: string): AccountType[] {
+  const types = value.split(",").map(t => t.trim());
+  const invalidTypes = types.filter(t => !isValidAccountType(t));
+
+  if (types.some(t => !isValidAccountType(t))) {
```
francois-guerin-ledger: Can we be homogeneous with the similar function below ?
```suggestion
  if (invalidTypes.length > 0) {
```

libs/coin-modules-monitoring/src/run.ts
```diff
@@ -134,7 +134,14 @@ export default async function (currencyIds: string[]) {
...
     const sync = getSync(currency);
 
-    for (const [accountType, info] of Object.entries(monitored.accounts)) {
+    for (const accountType of accountTypes) {
+      const info = monitored.accounts[accountType as AccountType]!;
+
+      if (!info) {
+        console.log(`\nSkipping currency = "${currencyId}", no account = "${accountType}"`);
+        continue;
+      }
```
francois-guerin-ledger, line 143: Why this change ? Looping over entries ensures `monitored.accounts[accountType]` exists.

.github/workflows/coin-monitoring-manual.yml
Moustafa-Koterba: I forgot to do it, done

.vscode/launch.json
Moustafa-Koterba: done

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: done

libs/coin-modules-monitoring/src/cli.ts
Moustafa-Koterba: done

libs/coin-modules-monitoring/src/run.ts
Moustafa-Koterba: I dont want to run all account of a currency when I pass `--account-types` option
So I need to filter on the account
Some currency have not all account type (example bitcoin, which does not `big` account)

Moustafa-Koterba:
> While running isolated, the command finishes as soon as we have one child failing. This is a regression compared to what we have, and we will be missing logs on Datadog

Should be good

francois-guerin-ledger (changes requested):
Almost done!
Remaining task is to exit the global process with `1` when at least one isolated child fails.

Moustafa-Koterba:
> Almost done!
Remaining task is to exit the global process with 1 when at least one isolated child fails.

Done

libs/coin-modules-monitoring/src/cli.ts
```diff
@@ -1,30 +1,188 @@
...
+  if (options.isolated) {
+    let atLeastOneProcessFailed = false;
+    for (const currency of monitoredCurrencies) {
+      for (const accountType of accountTypes) {
+        try {
+          await runIsolatedMonitor(currency, accountType);
+        } catch (error) {
+          if (!atLeastOneProcessFailed) {
+            atLeastOneProcessFailed = true;
+          }
```
francois-guerin-ledger, line 171: Useless `if`. Let's try to be homogeneous with the non-isolated counterpart [here](https://github.com/LedgerHQ/ledger-live/blob/eecc52f1350596d8057e878d5230f690ba17b007/libs/coin-modules-monitoring/src/run.ts#L211-L214).
```suggestion
            atLeastOneProcessFailed = true;
```

francois-guerin-ledger (approved):
Ideally add a unit test in `run.test.ts` with a custom `accountTypes`, and address my last comment.
I am not blocking that PR since you tried hard already.

PR #12081, opened 2025-09-30
fix(coin:sui): return positive amount for undelegate operations

jprudent:
### ✅ Checklist

  - alpaca api for coin:sui returns positive amounts, even for staking
  - no impact for bridge (different code path)

### 📝 Description

IN or OUT operations always return positive amounts
but UNSTAKING was returning a negative

we believe this is inconsistent API. It's UI to job to negate the amount if that makes sense for user.

### ❓ Context

https://ledgerhq.atlassian.net/browse/BACK-9784

---

### DIFF
--- a/libs/coin-modules/coin-sui/src/network/sdk.test.ts
+++ b/libs/coin-modules/coin-sui/src/network/sdk.test.ts
@@ -185,7 +185,7 @@ const mockTransaction = {
   ],
   timestampMs: "1742294454878",
   checkpoint: "313024",
-};
+} as SuiTransactionBlockResponse;
 
 const mockApi = new SuiClient({ url: "mock" }) as jest.Mocked<SuiClient>;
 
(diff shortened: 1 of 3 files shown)


libs/coin-modules/coin-sui/src/network/sdk.test.ts
```diff
@@ -185,7 +185,7 @@ const mockTransaction = {
   ],
   timestampMs: "1742294454878",
   checkpoint: "313024",
-};
+} as SuiTransactionBlockResponse;
```
acewf, line 188: Always odd to me using this type os assertions, telling the [compiler](https://www.totaltypescript.com/concepts/type-assertions#:~:text=Type%20assertions%20let%20you%20lie%20to%20TypeScript%20about%20what%20type%20a%20value%20is) something that might not be true. 
But i have seen it on other parts of the code base, but think we should avoid this approach overall.

PR #12154, opened 2025-10-03
feat: alpaca integration in coin-hedera (standard api + hts)

md-blockydevs:
### ✅ Checklist

  - @ledgerhq/coin-hedera
  - @ledgerhq/live-common

### 📝 Description

This PR introduces Alpaca API integration for Hedera, covering both the standard API and HTS. It refactors the `coin-hedera` library to establish a shared logic layer that can be reused across old bridge and Alpaca.

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
```diff
@@ -0,0 +1,321 @@
...
+import { FeeEstimation, Operation } from "@ledgerhq/coin-framework/api/types";
+import { createApi } from "../api";
+import { HEDERA_TRANSACTION_MODES, TINYBAR_SCALE } from "../constants";
+import { base64ToUrlSafeBase64 } from "../logic/utils";
+import { MAINNET_TEST_ACCOUNTS } from "../test/fixtures/account.fixture";
+import type { HederaMemo } from "../types";
+
+describe("createApi", () => {
+  const api = createApi({});
+
```
md-blockydevs, line 11: Docs (https://developers.ledger.com/docs/ledger-live/accounts/integration/blockchain/api#crafttransaction) mentions "Send max tx => crafted tx with max amount" test for `craftTransaction` method. 

I observed that in `coin-aptos`, the `useAllAmount` logic is done by checking `amount === BigInt(0)`. 
Is this the confirmed and intended behavior that I should replicate in `coin-hedera`?

libs/coin-modules/coin-hedera/src/network/client.ts
```diff
@@ -0,0 +1,26 @@
...
+function broadcastTransaction(transaction: HederaTransaction): Promise<TransactionResponse> {
+  return transaction.execute(getClient());
+}
+
+let _hederaClient: Client | null = null;
+
+function getClient(): Client {
+  _hederaClient ??= Client.forMainnet().setMaxNodesPerTransaction(1);
+
+  //_hederaClient.setNetwork({ mainnet: "https://hedera.coin.ledger.com" });
```
md-blockydevs: This comment was copied from old `src/api/network.ts`. I'm not sure whether it should be left or if it's safe to remove at this point

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
```diff
@@ -0,0 +1,321 @@
...
+    it("returns the last block information", async () => {
+      const lastBlock = await api.lastBlock();
+
+      expect(lastBlock).toMatchObject({
+        height: expect.any(Number),
+      });
+    });
+  });
+
+  describe("listOperations", () => {
```
jprudent, line 268: please could you add a test for a very big account ?

It should showcase that:
- if minHeight=0 and order=DESC, it returns a page with a token
  - passing the token for retrieving the page 2, it must assert that the operations:
      - are distinct from operations of page 1
      - doesn't contain operations where height > operations of page 1
- adapt same test for minHeight=MAX_INT and order=ASC 
Thanks

md-blockydevs: @jprudent Sure thing, this method worries me the most + I don't have any good way to test it, so every input means a lot. I've just pushed small refactor of `listOperations` and better tests. Just please note that Hedera has no blocks, `minHeight` is provided to satisfy the types

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
```diff
@@ -317,5 +319,45 @@ describe("createApi", () => {
...
+
+        // NOTE: this won't be equal to limit, because single Hedera transaction can generate multiple operations
+        expect(page1.length).toBeGreaterThanOrEqual(commonPagination.limit);
+        expect(page2.length).toBeGreaterThanOrEqual(commonPagination.limit);
+        expect(pagingToken1).not.toBeNull();
+        expect(pagingToken2).not.toBeNull();
+        expect(hasOverlap).toBe(false);
+        expect(firstPage1Timestamp).toBeInstanceOf(Date);
+        expect(firstPage2Timestamp).toBeInstanceOf(Date);
+        expect(firstPage1Timestamp > firstPage2Timestamp).toBe(order === "desc");
```
jprudent: We could be more severe checking that 

```suggestion
expect(lastPage1Timestamp > firstPage2Timestamp).toBe(order === "desc");
expect(firstPage1Timestamp < lastPage2Timestamp).toBe(order === "asc);
```

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
jprudent: Thanks for the tests! I just made a comment to improve them.

`limit` is not that important, what is important for listOperations is:
- no missing ops
- chronological order in page and between pages (desc or asc)
- no overlapping ops between pages

I've to do some research about `minHeight`, I'll tell you soon

libs/coin-modules/coin-hedera/src/network/client.ts
scavassa-yld: I think it's ok to remove that

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
scavassa-yld: Have checked with team and we should go with  useAllAmount property instead. 
So, the check on amount would not be the best approach.

md-blockydevs: Thanks for confirming! I'm leaving `craftTransaction` as it is now for coin-hedera then

libs/coin-modules/coin-hedera/src/api/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  coinConfig.setCoinConfig(() => ({ ...config, status: { type: "active" } }));
+  const currency = getCryptoCurrencyById("hedera");
+
+  return {
+    broadcast: async tx => {
+      const response = await logicBroadcast(tx);
+      const base64Hash = Buffer.from(response.transactionHash).toString("base64");
+      const base64HashUrlSafe = base64ToUrlSafeBase64(base64Hash);
+
+      return base64HashUrlSafe;
```
jprudent: does that mean that as a user of coin-module I can't use the pristine return value of broadcast ? I have to convert it back to "real" base64 ?

As a user I need to be able to compare hashes on a public explorer for instance. That's a common practice, I don't want an extra step of conversion for a technical reason.

if so, I would suggest:
- do not use encoding at all, except if it's a binary format, in that case an hexstring is the way to go (but looks like it's b64 which is fine)
- make necessary urlSafe conversion at caller site: in the bridge or LL code somewhere (I'm not familiar with that codebase)

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
```diff
@@ -0,0 +1,149 @@
...
+    type: "js",
+    version: "2",
+    currencyId: currency.id,
+    xpubOrAddress: address,
+    derivationMode: "hederaBip44",
+  });
+
+  for (const rawTx of mirrorResult.transactions) {
+    const timestamp = new Date(parseInt(rawTx.consensus_timestamp.split(".")[0], 10) * 1000);
+    const hash = base64ToUrlSafeBase64(rawTx.transaction_hash);
```
jprudent: same problem here, we need the original tx hash

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
```diff
@@ -0,0 +1,149 @@
...
+          recipients,
+          senders,
+          hash,
+          fee,
+          date: timestamp,
+          blockHeight,
+          blockHash,
+          hasFailed,
+          extra,
+        });
```
jprudent, line 95: in other coin modules, we don't have a FEE operation, 
the fees are set on the Operation (IN/OUT)

is there a good reason why we diverge from that ?

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
```diff
@@ -0,0 +1,149 @@
...
+        hasFailed,
+        extra,
+      });
+    } else if (transfers.length > 0) {
+      const { type, value, senders, recipients } = parseTransfers(transfers, address);
+      const extra = { ...commonExtra };
+      let operationType = type;
+
+      // try to enrich ASSOCIATE_TOKEN operation with extra.associatedTokenId
+      // this value is used by custom OperationDetails components in Hedera family
```
jprudent, line 121:
```suggestion
      // try to enrich ASSOCIATE_TOKEN operation with extra.associatedTokenId
      // this value is used by custom OperationDetails components in Hedera family
      // Accounts or contracts must first associate with an HTS token before they can receive or send that token; without association, token transfers fail
```

libs/coin-modules/coin-hedera/src/network/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const senders: string[] = [];
+  const recipients: string[] = [];
+
+  for (const transfer of mirrorTransfers) {
+    const amount = new BigNumber(transfer.amount);
+    const accountId = AccountId.fromString(transfer.account);
+
+    if (transfer.account === address) {
+      value = amount.abs();
```
jprudent, line 37: should the value be accumulated ?
if there are several transfers, I must have the sum of them 
why? because we compute balances based on amounts, and that must correct value

If you have a concrete example of multitransfer, I could check if what I'm saying is correct or no (I'm unsure)

libs/coin-modules/coin-hedera/src/network/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+      type = amount.isNegative() ? "OUT" : "IN";
```
jprudent, line 38: this will likely be a mix, (I don't have suggestion yet)

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
jprudent: unfortunately, we need the `minHeight`
I'll raise the issue to the team and try to find a reasonable solution.

libs/coin-modules/coin-hedera/src/api/index.ts
md-blockydevs: That's a good point - I'm not sure why this was added. I found following note next to the `base64ToUrlSafeBase64` definition:

>// NOTE: convert from the non-url-safe version of base64 to the url-safe version (that the explorer uses)

But the `getTransactionExplorer` logic in Hedera is based on consesusTimestamp/transactionId. I will try to remove it and see if it doesn't break any flow.

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
md-blockydevs: This is basically a copy of the old bridge behavior as I wasn't sure what Alpaca expects. In current LL version, the main account has "FEES" operation visible in the history, with "OUT" token operation linked to it. 

If I understood correctly, this is something that the bridge should solve and it's not needed on the API level, so I will have to refactor this a bit.

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
md-blockydevs: Sure thing, thanks for checking it. I'm not sure how the `minHeight` is going to be used, but if this is meant for pagination only, then we also have the `lastPagingToken` option. We just have to keep the `minHeight: 0` to satisfy the TS like it's done now.

libs/coin-modules/coin-hedera/src/network/utils.ts
md-blockydevs: Here is an example of multi transfer made from `0.0.8835924`: https://hashscan.io/mainnet/transaction/1760020254.898270421

```
const recipient1 = AccountId.fromString("0.0.9124531");
const recipient2 = AccountId.fromString("0.0.9169746");
const recipient3 = AccountId.fromString("0.0.9213430");

  const tx = new TransferTransaction()
    .addHbarTransfer(operatorId, new Hbar(0.03).negated())
    .addHbarTransfer(recipient1, new Hbar(0.01))
    .addHbarTransfer(recipient2, new Hbar(0.01))
    .addHbarTransfer(recipient3, new Hbar(0.01));
```

libs/coin-modules/coin-hedera/src/network/utils.ts
md-blockydevs: Hmm, could you clarify this one? Maybe multi-transfer example will be helpful: https://hashscan.io/mainnet/transaction/1760020254.898270421

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
jprudent: Hello, I've got more explanations:

- we maintain address operations state polling this endpoint with order=DESC
- without minHeight we receive a whole page that may contains operations we already synced
- with minHeight filter implemented, we only receive operations we don't have
- so minHeight is an optimisation 
- in some coin modules this is important because the implem has to do N+1 calls to the node in order to retrieve operations with all mandatory data
- I from what I read, it's not the case for this implem
- So we can live without it !

You can resolve that conversation. Thanks

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
jprudent: I agree with your conclusion :+1:

libs/coin-modules/coin-hedera/src/api/index.ts
md-blockydevs: I've just pushed a commit with two "compatibility" flags added to common `listOperations` logic:
- skipFeesForTokenOperations
- useEncodedHash

I feel like this is the easiest solution, because hash encoding cannot be easily removed from the old bridge. Hash is used during "operation.id" generation, so duplicated operations could occur 😕 Right now the old bridge will behave in the same way and Alpaca API will return raw base64 hashes + no FEES operations for tokens.

libs/coin-modules/coin-hedera/src/network/utils.ts
jprudent: Thanks for example!

In that case this implem would return:
- senders 0.0.8835924
- recipients 0.0.9124531, 0.0.9169746, 0.0.9213430, 0.0.28, 0.0.801
- amount 0.00045298 (last amounts of transfers)
- type OUT

So the final op (output in api layer) would contain 0.00045298 instead of 0.03047215
which is not correct (any balance computation would be incorrect)

libs/coin-modules/coin-hedera/src/network/utils.ts
scavassa-yld: @jprudent 
Hi, just to mention - Blocky devs can't merge to develop branch and in order to proceed with PR Review process, we will move this PR to a new Feature  Branch that I've created and then I'll open the PR.  So, **your review on this PR is not worthless, but can be done or continued in the next one**, which I'll share the link with you.  :)

libs/coin-modules/coin-hedera/src/network/utils.ts
md-blockydevs: Hmm, this logic hasn't changed and it's just a copy of what we have for old bridge behavior. I haven't noticed anything weird in the operations history 🤔 

For this transaction: https://hashscan.io/mainnet/transaction/1760020254.898270421, this function will output:
```
{
  type: "OUT",
  value: 3047215,
  senders: [ "0.0.8835924", "0.0.800" ],
  recipients: [ "0.0.9213430", "0.0.9169746", "0.0.9124531", "0.0.28" ],
}
```

And the value for Alpaca operation is calculated here: https://github.com/LedgerHQ/ledger-live/blob/81c2c6ae6ffe2eab3c9559aaec16e8911a5c408b/libs/coin-modules/coin-hedera/src/api/index.ts#L103

So without running it and just looking at the code, I would say that the final operation would have the `value = 3047215 - 47215 (charged_tx_fee) = 3000000`

libs/coin-modules/coin-hedera/src/network/utils.ts
jprudent: ~~based on my comment here https://github.com/LedgerHQ/ledger-live/pull/12154/files#r2419158408~~
~~yes, the correct implementation should return the type based on the sign of the sum of the transfer amounts~~
~~ie: if there is a IN = 100 and an OUT = 1000, then  type=OUT and amount=900~~

~~Does that make sense for you too ?~~

libs/coin-modules/coin-hedera/src/network/utils.ts
jprudent: Ok I may have misunderstood, if final value is correct, let's close this thread.

libs/coin-modules/coin-hedera/src/network/utils.ts
jprudent: ok based on what you commented above let's forget my last comment then.

md-blockydevs: Ahh, sorry, it would be probably easier to understand with raw data: https://mainnet-public.mirrornode.hedera.com/api/v1/transactions/0.0.8835924-1760020247-065844637. This is what both hashscan and LL use to process transactions and for our example of multi-transfer we have:

image

Never mind then, should be fine 😄

libs/coin-modules/coin-hedera/src/api/index.ts
jprudent: ok thansk

PR #12159, opened 2025-10-03
feat: LIVE-21973 add provider session id to noah

hhumphrey-ledger:
### ✅ Checklist

  - ...

### 📝 Description

In order for tracking to be correctly sent from the Noah live app we need to supply them with the providerSessionId. For Noah this is sent as `externalId`.

Below a test live app that is outputting the url params. Mobile doesn't have locale or theme - I'm looking at that separately:

Screenshot 2025-10-03 at 10 06 53

Screenshot 2025-10-03 at 10 42 19

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-21973

---

### DIFF
--- a/libs/env/src/env.ts
+++ b/libs/env/src/env.ts
@@ -651,6 +651,11 @@ const envDefinitions = {
     parser: stringParser,
     desc: "(dev feature) seed to be used by speculos (device simulator)",
   },
+  PROVIDER_SESSION_ID_ENDPOINT: {
+    def: "https://buy.api.live.ledger.com/session",
+    parser: stringParser,
+    desc: "Request provider session id",
+  },
   SHOW_LEGACY_NEW_ACCOUNT: {
     def: false,
     parser: boolParser,
--- a/libs/ledger-live-common/src/hooks/useManifestWithSessionId.ts
+++ b/libs/ledger-live-common/src/hooks/useManifestWithSessionId.ts
@@ -1,0 +1,83 @@
+import React from "react";
+import { appendQueryParamsToManifestURL } from "../wallet-api/utils/appendQueryParamsToManifestURL";
+import { LiveAppManifest } from "../platform/types";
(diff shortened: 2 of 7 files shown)


libs/env/src/env.ts
```diff
@@ -651,6 +651,11 @@ const envDefinitions = {
     parser: stringParser,
     desc: "(dev feature) seed to be used by speculos (device simulator)",
   },
+  PROVIDER_SESSION_ID_ENDPOINT: {
```
hhumphrey-ledger, line 654: Do I need to add this variable anywhere else?

mcayuelas-ledger: No it's ok like this

libs/ledger-live-common/src/hooks/useManifestWithSessionId.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+type HookResult = {
+  manifest?: LiveAppManifest | null;
+  loading: boolean;
+};
+
+export function useManifestWithSessionId({ manifest, shareAnalytics }: Options): HookResult {
+  const [id, setId] = React.useState<string | null>(null);
+  const [loading, setLoading] = React.useState<boolean>(false);
```
mcayuelas-ledger, line 18: You can maybe use [RTK Query](https://redux-toolkit.js.org/rtk-query/usage/queries) for this? (you can look at `libs/ledger-live-common/src/dada-client/state-manager/api.ts`) 

It will helps you to get all the state Management :)

hhumphrey-ledger: That looks super nice - but it does seem overkill for this scenario? I don't want the request to be cached and we are only using this request in this one hook so for simplicity it seems best to just handle it in this way?

dada-client makes sense as there is a lot more complexity there.

PR #12173, opened 2025-10-06
feat: canton full transaction signing

ishaba:
### ✅ Checklist

  - ledger-live-desktop, coin-canton, hw-app-canton, live-signer-canton

### 📝 Description

Replace hash signing with full transaction signing 

### ❓ Context

- **JIRA or GitHub link**: 
- [LIVE-21689](https://ledgerhq.atlassian.net/browse/LIVE-21689)

---

[LIVE-21689]: https://ledgerhq.atlassian.net/browse/LIVE-21689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepAuthorize.tsx
```diff
@@ -14,9 +14,25 @@ import { TransactionConfirm } from "../components/TransactionConfirm";
...
+  accountName,
+  authorizeStatus,
+  device,
+  onboardingResult,
+  error,
+}: StepProps) => {
   invariant(onboardingResult?.completedAccount, "canton: completed account is required");
 
+  const getErrorMessage = (error: Error | null) => {
+    if (error?.name === "DeviceLockedError") {
```
acewf: looking into other places that have similar conditions, the implementation is slightly different 
https://github.com/LedgerHQ/ledger-live/blob/267f8f8be0453e29e43452d18f3a9508486b1de3/libs/ledger-live-common/src/hw/connectManagerEventMapper.ts#L95

What i think it needs to be improved here, is the last connection of possible error types and the comparisons, at this moment nothing on typescript gives us confidence the code is going to work.

acewf: there are other instances of this along of your PR, but hope you can catch and address them.

libs/coin-modules/coin-canton/package.json
```diff
@@ -108,21 +108,24 @@
...
     "dotenv": "^16.4.5",
     "expect": "^27.4.6",
     "jest": "^29.7.0",
+    "protobufjs-cli": "1.1.2",
     "ts-jest": "^29.1.1",
     "typescript": "^5.4.5"
   },
   "scripts": {
     "clean": "rimraf lib lib-es",
+    "generate-proto": "node scripts/generate.js",
```
acewf, line 128: when are we meant to use this script? 
cause isn't available on turborepo, and also we don't have any specific information on the README file for this coin-module

acewf: would it be better to keep the file in typescript and run it with something like `ts-node`?
`npx ts-node scripts/generate.ts`

libs/coin-modules/coin-canton/src/bridge/sync.integ.test.ts
```diff
@@ -201,6 +201,6 @@ describe("sync (devnet)", () => {
       });
 
       mockGetOperations.mockRestore();
-    });
+    }, 30000);
```
acewf: not sure what is the particular reason to change the timeout, but instead of repetition, could we just have a constant?

ishaba: Request times increased significantly after the latest gateway updates — sure, I’ll move it to a constant.

libs/coin-modules/coin-canton/package.json
ishaba: It’s meant to be run after protobuf updates on the gateway and application sides. I’ll add a description to the README.
Not sure there’s much benefit in converting it to TypeScript, but I can do it if you prefer.

apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepAuthorize.tsx
ishaba: thanks, changed it

jonezque (approved):

.changeset/late-games-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ledgerhq/hw-app-canton": minor
+"@ledgerhq/coin-canton": minor
+"@ledgerhq/live-signer-canton": minor
```
francois-guerin-ledger, line 4: `ledger-live-desktop` is missing here

libs/coin-modules/coin-canton/src/bridge/onboard.integ.test.ts
```diff
@@ -221,25 +221,5 @@ describe("onboard (devnet)", () => {
...
-      // WHEN & THEN
-      try {
-        await firstValueFrom(
-          preapprovalObservable(mockCurrency, mockDeviceId, mockAccount, "invalid-party-id-123"),
-        );
-        expect(true).toBe(true);
-      } catch (error) {
-        expect(error).toBeDefined();
-      }
-    }, 30000);
```
francois-guerin-ledger, line 243: Why removing this test ? Are you confident it is useless or it is a duplicate ?

libs/coin-modules/coin-canton/src/bridge/signOperation.test.ts
```diff
@@ -0,0 +1,149 @@
...
+          }
+        },
+        error: reject,
+      });
+    });
+
+    // THEN
+    expect(mockCraftTransaction).toHaveBeenCalled();
+    expect(result).toBeDefined();
+    expect((result as any).signedOperation.signature).toContain("prepared-transaction-signature-");
```
francois-guerin-ledger: Combine this assertion with
```ts
expect(result).toMatchObject({ ... });
```

libs/coin-modules/coin-canton/src/bridge/signOperation.test.ts
```diff
@@ -0,0 +1,149 @@
...
+            commandId: "test-command-id",
+          },
+          synchronizerId: "test-synchronizer-id",
+          transactionUuid: "test-transaction-uuid",
+          submissionTime: "1234567890",
+          inputContracts: [],
+        },
+      },
+      serializedTransaction: "serialized-transaction",
+      hash: "mock-hash",
```
francois-guerin-ledger: Do you need to return such a massive object for your single test ?

ishaba: can be optimised actually, thanks

.changeset/late-games-tie.md
ishaba: good spot, thanks!

libs/coin-modules/coin-canton/src/bridge/onboard.integ.test.ts
ishaba: Previously, there was no validation on the gateway side — now there is.
Also, the test itself was poorly designed; there’s no need for an extra check for the thrown error since that’s the default behavior.

libs/coin-modules/coin-canton/src/network/gateway.ts
```diff
@@ -428,73 +428,10 @@ export async function getOperations(
...
-  json: null;
-  hash: string;
-};
-
 enum TransactionType {
-  TAP_REQUEST = "tap-request",
   TRANSFER_PRE_APPROVAL_PROPOSAL = "transfer-pre-approval-proposal",
 }
 
-export async function prepareTapRequest(
```
hedi-edelbloute, line 447: Can't we keep using it for devnet ?

libs/coin-modules/coin-canton/scripts/generate.js
hedi-edelbloute, line 1: This might need to move to hw-app-canton, wdyt @sprohaszka-ledger ?

Also, @jnicoulaud-ledger spotted this: https://docs.daml.com/app-dev/bindings-ts/daml2js.html would this help ?

libs/coin-modules/coin-canton/README.md
```diff
@@ -6,6 +6,50 @@ This repository contains the Coin Canton module for Ledger Live.
...
+# Navigate to the coin-canton directory
+cd libs/coin-modules/coin-canton
+
+# Run the protobuf generation script
+pnpm generate-proto
+```
+
+### Description
+
+The script performs the following operations:
```
hedi-edelbloute: can you explain the full flow in documentation ? where / why those files are used / what it contains, not just how to generate them

libs/coin-modules/coin-canton/scripts/generate.js
ishaba: but actual binding is used inside coin-canton or you suggest to move also split.ts logic to hw-app-canton?

libs/coin-modules/coin-canton/src/network/gateway.ts
ishaba: yes we can

libs/coin-modules/coin-canton/README.md
ishaba: ok i will add more info

libs/coin-modules/coin-canton/scripts/generate.js
ishaba: we can get some benefits using daml2js but i think it is better to keep it outside of this pr

libs/coin-modules/coin-canton/package.json
acewf: converting to typescript was to have unified way how we write our code, with the type safety benefit. But was more a suggestion then something to be enforced.

libs/coin-modules/coin-canton/scripts/generate.js
hedi-edelbloute: If split.ts is also related to device app, yes

or maybe I got it wrong ? But in coin-canton it feels weird to have code related to device app

ishaba: You’re right — it’s more related to the device. Ideally, we’d move this part of the code to hw-app-canton, but I just checked and there’s a dependency in the integration tests that also uses the proto definitions. So, it won’t be an easy change right now. I suggest postponing it for later.

hedi-edelbloute: are you talking about `libs/coin-modules/coin-canton/src/common-logic/transaction/split.ts` ? It needs to move to hw-app library too right ?

ishaba: Yes but i prefer to do it in the upcoming pr's

jonezque (approved):

PR #11905, opened 2025-09-18
refine CryptoAssetsStore to be async

gre-ledger:
TLDR:

```diff
- const token = findTokenById(id);
+ const token = await getCryptoAssetsStore().findTokenById(id);

- const token = getTokenById(id);
+ const token = await getCryptoAssetsStore().findTokenById(id);
+ if (!token) return null; // or do something else

- const token = findTokenByAddress(address);
- const token = findTokenByAddressInCurrency(address, currency);
+ const token = await getCryptoAssetsStore().findTokenByAddressInCurrency(address, currency);

- const token = findTokenByTicker(ticker);
⚠️  no more usecase identified. Feature dropped.
```

**Recommendation for code reviewers: focus on the part owned by you:**

Screenshot 2025-09-29 at 20 31 06

### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** (Existing tests updated for async patterns)
- [x] Builds
  - LLD: https://github.com/LedgerHQ/ledger-live-build/actions/runs/18102818564
  - iOS: https://github.com/LedgerHQ/ledger-live-build/actions/runs/18102824726
  - Android: https://github.com/LedgerHQ/ledger-live-build/actions/runs/18102821547
- [x] E2E runs
  - LLD https://github.com/LedgerHQ/ledger-live/actions/runs/18121309613
  - LLM https://github.com/LedgerHQ/ledger-live/actions/runs/18130617435
- [x] **Impact of the changes:**
  - **Breaking change** for devs: see migration guide above.
  - **Data serialization**: `fromAccountRaw` and `DataModel.decode` now async
  - Features to focus on during manual testing:
    - Account management (import/export, synchronization)
      - Add Account flow - verify tokens still appear when adding accounts
      - Receive Token flow - test receiving specific tokens (not just main currency)
      - Asset Selection - any dropdown/picker showing available tokens
      - Portfolio display - ensure all holdings/tokens load correctly
    - Specific Coin Flows:
      - Stellar: Add Assets flow
       - Algorand: Select Token flow
       - Hedera: Token association flows
       - EVM & token operations (the biggest impact area)
     - Market section - token prices/data loading
     - Swap/Exchange - token selection and pricing
     - Settings > Accounts - blacklisted tokens display
  - LLM specifics
    - ReadOnly Assets - portfolio view for read-only accounts

### 📝 Description

**Problem**: Current token system uses hardcoded static lists, preventing dynamic token loading from APIs.

[...]

PR #12196, opened 2025-10-06
chore: drop hasToken and replace getTokenById by findTokenById

gre-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** All existing tests pass, including cryptoassets package tests.
  - LLD and LLM are touched but no impact for users.
  - it's mostly tests that are actually altered here.

### 📝 Description

- `hasToken` is an unused function that was dropped.
- `getTokenById` usages has been inlined to use `findTokenById`. hopefully there weren't a lot of remaining usages (mostly in tests)
  - throwing an untyped exception wasn't a good pattern and it's never really safe to expect a token to be available by id
  - so it's up to dev to manage that possibly it can be undefined. the falsy case is to be handled on the usage side.
  - this simplifies the CryptoAssetStore interface which should only focus implementing the canonical need
  - in a very near future (#11905), `store.findTokenById` is becoming asynchronous, it will only "fail/throw exception" if the networking fails, so it's a better paradigm to differentiate between "unavailable token" and "temporarily impossible to retrieve it" cases.
- we take this opportunity to introduce `import { legacyCryptoAssetsStore } from "@ledgerhq/cryptoassets/tokens";` in order to reduce the previous repetitive pattern we had everywhere.

**The CryptoAssetsStore interface finally have its final shape (minus the async future changes):**

```ts
export type CryptoAssetsStore = {
  findTokenById(id: string): TokenCurrency | undefined;
  findTokenByAddressInCurrency(address: string, currencyId: string): TokenCurrency | undefined;
};
```

### ❓ Context

- **JIRA or GitHub link**:  facilitator of #11905 

---

### DIFF
--- a/.changeset/remove-getTokenById-hasTokenId.md
+++ b/.changeset/remove-getTokenById-hasTokenId.md
@@ -1,0 +1,16 @@
+---
+"@ledgerhq/cryptoassets": minor
+"@ledgerhq/live-common": minor
+"@ledgerhq/coin-evm": minor
+"@ledgerhq/coin-solana": minor
+"@ledgerhq/coin-vechain": minor
+"@ledgerhq/coin-ton": minor
+"@ledgerhq/coin-tester-evm": minor
+"@ledgerhq/coin-tester-solana": minor
+"@ledgerhq/live-countervalues": minor
+"live-mobile": minor
+---
+
+Remove deprecated getTokenById and hasTokenId functions
+
+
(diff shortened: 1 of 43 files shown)


gre-ledger:
@thesan reviewed them. it's all LOW. I think we should take a look at SonarQube in another PR (notably per feature to improve), here getTokenById were strictly replaced by the equivalent code: the function is inlined with a findTokenById that throws if undefined => this is the naive equivalent, in future, we can refine this, but we need getTokenById to be removed for the CAL removal.

PR #12197, opened 2025-10-06
feat(LIVE-21181): add sponsored transactions

philipptpunkt:
add sponsored transactions to evm based swaps

### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-21181

---


libs/coin-modules/coin-evm/src/bridge/getTransactionStatus.ts
```diff
@@ -115,7 +115,7 @@ const validateAmount = (
   // if no amount or 0
   if ((!transaction.amount || transaction.amount.isZero()) && !canHaveZeroAmount) {
     errors.amount = new AmountRequired(); // "Amount required"
-  } else if (totalSpent.isGreaterThan(account.balance)) {
+  } else if (totalSpent.isGreaterThan(account.balance) && !transaction.sponsored) {
     // if not enough to make the transaction
     errors.amount = new NotEnoughBalance(); // "Sorry, insufficient funds"
   }
```
Justkant: I'm not sure we just want to allow any transaction spending more than the balance but only the ones that would not cover the fees

Or even better the totalSpent should probably be correct for sponsored transactions

libs/ledger-live-common/src/families/evm/walletApiAdapter.ts
```diff
@@ -37,6 +37,13 @@ const convertToLiveTransaction: ConvertToLiveTransaction<
     liveTx.data = walletApiTransaction.data;
   }
 
+  // Propagate sponsored flag if provided by the Wallet API transaction
+  // This enables fee validation logic to skip NotEnoughGas for sponsored txs
+  const maybeSponsored = (walletApiTransaction as Record<string, unknown>).sponsored;
+  if (typeof maybeSponsored === "boolean") {
+    liveTx.sponsored = maybeSponsored;
+  }
```
Justkant: I think you could simplify this a bit
```suggestion
  if ("sponsored" in walletApiTransaction) {
    liveTx.sponsored = walletApiTransaction.sponsored;
  }
```

libs/coin-modules/coin-evm/src/bridge/getTransactionStatus.ts
philipptpunkt: In the calculation for `totalAmount`
```
const totalSpent = isTokenTransaction ? tx.amount : tx.amount.plus(totalFees);
```
we can skip adding the fees when the `sponosred` is `true` - for the approval the amount should be 0, no?! so without adding the fees, we are not above the user balance.

Just not sure if we can savly return a `totalAmount` of `0` or if I should only remove the fees from the `totalAmount` for the balance validation check.

Justkant: You would need to allow sending 0 for `sponsored` transactions because we have a check for it too I think
I think it's better if we don't add fees to the amount if `sponsored` is `true`

libs/ledger-live-common/src/families/evm/walletApiAdapter.ts
philipptpunkt: That's correct 😎

Justkant (commented):
LGTM you just need to bypass the `AmountRequired` error for sponsored transactions, there is already a `canHaveZeroAmount` that you can edit to handle this case

PR #12202, opened 2025-10-07
Live 20263 earn modular asset network drawer with apy

sergiubreban:
Enable earn drawer configuration based on FF value.
FF key: feature_ptx_earn_drawer_configuration

## Demo

https://github.com/user-attachments/assets/257fd344-4e45-49e6-8a9c-83d8bc22031b

https://github.com/user-attachments/assets/d317b0a3-b451-4b55-921b-3e33ad955b43

### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

[LIVE-20812](https://ledgerhq.atlassian.net/browse/LIVE-20812)

---

[LIVE-20812]: https://ledgerhq.atlassian.net/browse/LIVE-20812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
@@ -95,13 +95,6 @@ function useUiHook(manifest: AppManifest, tracking: TrackingAPI): UiHook {
 
           // We agree that for useCase, we should send max 25 currencies if provided else use only useCase (e.g. buy)
           const shouldUseCurrencies = (useCase && currencies.length <= 25) || !useCase;
-          const earnAppDrawerConfig: {
-            assets?: Record<string, unknown>;
-            networks?: Record<string, unknown>;
-          } =
-            earnDrawerConfigurationFlag?.enabled && earnDrawerConfigurationFlag.params
-              ? earnDrawerConfigurationFlag.params
-              : {};
 
           const finalDrawerConfiguration = createDrawerConfiguration(drawerConfiguration, useCase);
 
--- a/libs/ledger-live-common/src/dada-client/hooks/useDrawerConfiguration.ts
+++ b/libs/ledger-live-common/src/dada-client/hooks/useDrawerConfiguration.ts
@@ -49,12 +49,12 @@ export function useDrawerConfiguration() {
 
       return {
         assets: {
-          ...(useCaseConfig?.assets || {}),
-          ...(config?.assets || {}),
+          ...useCaseConfig?.assets,
+          ...config?.assets,
         },
         networks: {
-          ...(useCaseConfig?.networks || {}),
-          ...(config?.networks || {}),
+          ...useCaseConfig?.networks,
+          ...config?.networks,
         },
       };
     },


apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -91,9 +93,23 @@ function useUiHook(manifest: AppManifest, tracking: TrackingAPI): UiHook {
...
+          const finalDrawerConfiguration = {
+            assets: {
+              ...(earnAppDrawerConfig?.assets || {}),
+              ...(drawerConfiguration?.assets || {}),
+            },
+            networks: {
+              ...(earnAppDrawerConfig?.networks || {}),
+              ...(drawerConfiguration?.networks || {}),
+            },
+          };
```
mcayuelas-ledger: I don't like the fact we transform the data from "earn" here , because Webview should be the genreric entry

Why don't send the config from liveApp directly?

apps/ledger-live-desktop/src/renderer/screens/earn/index.tsx
```diff
@@ -79,7 +78,6 @@ const Earn = () => {
           stakeCurrenciesParam: stakeCurrenciesParam
             ? JSON.stringify(stakeCurrenciesParam)
             : undefined,
-          earnDrawerApyFlag: earnDrawerApyFlag ? JSON.stringify(earnDrawerApyFlag) : undefined,
```
mcayuelas-ledger, line 82: You were doing it here no?

sergiubreban: This was intended to be passed down to the earn app, which is supposed to send the config back when using wallet-api to request an account. But I don't think we need to pass down the data and then pass it back to use it in wallet-api handler. We can hook directly in the wallet api handler to specify a default behaviour for earn, and anything that will be sent as `drawerConfiguration` it will override the default behaviour. is this an issue?

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
sergiubreban: it's a config that will be sent back anyway, and it's the same config that we need to use in ledger-live entrypoints. I am thinking that having all the setup in ledger-live would be easier to maintain

sergiubreban: there was a bug actually here, I forgot to add the check if the useCase === 'earn'. I pushed a new commit. is this approach okay @mcayuelas-ledger ?

Justkant (approved):
LGTM

Valentin-Ledger (approved):

libs/ledger-live-common/src/dada-client/hooks/useDrawerConfiguration.ts
```diff
@@ -0,0 +1,65 @@
...
+        networks: {
+          ...useCaseConfig?.networks,
+          ...config?.networks,
+        },
+      };
+    },
+    [earnDrawerConfigurationFlag],
+  );
+
+  return { createDrawerConfiguration };
```
marcotoniut-ledger, line 64: It's a bit of a nitpick, I suppose, shouldn't this hook be called something like useCreateDrawerConfiguration instead?

marcotoniut-ledger (approved):
Nitpick comment aside, it looks good to me.

PR #12209, opened 2025-10-07
:package: (i18next) upgrade

LucasWerey:
### ✅ Checklist

  - all LLM translations

### 📝 Description

Upgrade i18next packages on LLM. I put the versions in the catalog to have the same ones on LLM and LLD for consistency

This way we can use _other instead of _plural, which is deprecated  

small example: 

image

image

image

### ❓ Context

- **JIRA or GitHub link**: [LIVE-22038](https://ledgerhq.atlassian.net/browse/LIVE-22038)

---

[LIVE-22038]: https://ledgerhq.atlassian.net/browse/LIVE-22038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre-ledger (dismissed):

Moustafa-Koterba (dismissed):

apps/ledger-live-mobile/src/components/TranslatedError/TranslatedError.tsx
```diff
@@ -52,10 +52,10 @@ export function TranslatedError({ error, field = "title" }: Props): JSX.Element
   if (error.name) {
     // the string id of a simple error message
     const simpleTranslationStringId = `errors.${error.name}.${field}`;
-    let translation = t(simpleTranslationStringId, arg);
+    let translation = t(simpleTranslationStringId, { ...arg });
```
Justkant, line 55: Why do we need to spread now ?

LucasWerey: Because arg isn't an object but an Error and an object but the t function only accepts an object

apps/ledger-live-mobile/src/locales/en/common.json
```diff
@@ -56,9 +56,9 @@
     "usb": "USB",
     "add": "Add",
     "token": "Token",
-    "token_plural": "Tokens",
+    "token_other": "Tokens",
```
Justkant: The [doc on plurals](https://www.i18next.com/translation-function/plurals) shows that we should use `_one` for the singular case instead of no suffix, wdyt ?

I think the advantage is that you would get correct 0 and 1 cases but I'm not sure about that
From the doc they also share a tool to do the migration
> With [v21.0.0](https://www.i18next.com/misc/migration-guide#json-format-v4-pluralization) a new [JSON format v4](https://www.i18next.com/misc/json-format#i-18-next-json-v4) was introduced that changed the suffixes.
To convert your existing translations to the new v4 format, have a look at [i18next-v4-format-converter](https://github.com/i18next/i18next-v4-format-converter) or [this web tool](https://i18next.github.io/i18next-v4-format-converter-web/).

aussedatlo (dismissed):

apps/ledger-live-mobile/src/locales/en/common.json
Justkant: Also are we sure that we now always use/pass a count param for the plural to work ?

> Note: The variable name must be count.
And it must be present: i18next.t('key', {count: 1});
There will be no fallback to the 'key' value if count is not provided.

apps/ledger-live-mobile/src/components/TranslatedError/TranslatedError.tsx
Justkant: Got it thx

philipptpunkt (dismissed):

apps/ledger-live-mobile/src/locales/en/common.json
LucasWerey: - Yep I'll change to use _one
- I'll check 👁️

LucasWerey: Most of the keys I changed are not used actually, so the risk is limited. But all the used ones have the count param

philipptpunkt (approved):

PR #12210, opened 2025-10-07
remove dependency on getTokenOrCryptoCurrencyById in getProviderCurrency

gre-ledger:
### ✅ Checklist

  - MAD

### 📝 Description

- This removes the last dependency that MAD has on the CAL tokens
- this remove the previous mechanism that was mapping the extracted currencies into the hardcoded tokens from the bundle.
  - note that it means, we will always have a new ref of a `token` object. so we must not assume `tokenA.id === tokenB.id` means `tokenA === tokenB`. this is the main impact this could have here, but i didn't experience issue so far. that said, the problem could have exist if the backend would return token unknown from the client, as the fallback case was to use [0], in any case we now converge to the same approach.
- I can't find a valid reason that we needed to `toLowerCase` on ids. it's generally a bad practice and that usually need a fix backend side.
- by testing many search term on my end, i never have cases of seeing the console.warn that I've introduced so I think this rework is functional.

### ❓ Context

- **JIRA or GitHub link**: LIVE-21847

---

### DIFF
--- a/.changeset/refactor-provider-currency-lookup.md
+++ b/.changeset/refactor-provider-currency-lookup.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": minor
+---
+
+refactor: remove dependency on getTokenOrCryptoCurrencyById in getProviderCurrency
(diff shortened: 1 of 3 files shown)


gre-ledger:
decision has been to wait for full enabling of MAD before reworking this => the code will actually be dropped as no more flow will need this part in a near future.

PR #12254, opened 2025-10-10
Feat/live 20546 blockydevs hedera alpaca

scavassa-yld:
### ✅ Checklist

  - ...

### 📝 Description
Implement Alpaca Integration to Coin-Hedera

### ❓ Context

[JIRA ](https://ledgerhq.atlassian.net/browse/LIVE-20546)

---

### DIFF
--- a/libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
+++ b/libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
@@ -6,11 +6,11 @@ import {
   TokenAssociateTransaction,
   TransferTransaction,
 } from "@hashgraph/sdk";
-import { FeeEstimation, Operation, Pagination } from "@ledgerhq/coin-framework/api/types";
+import type { FeeEstimation, Pagination } from "@ledgerhq/coin-framework/api/types";
 import { createApi } from "../api";
 import { HEDERA_TRANSACTION_MODES, TINYBAR_SCALE } from "../constants";
+import { getSyntheticBlock } from "../logic/utils";
 import { MAINNET_TEST_ACCOUNTS } from "../test/fixtures/account.fixture";
-import type { HederaMemo } from "../types";
 
 describe("createApi", () => {
   const api = createApi({});
@@ -269,9 +269,9 @@ describe("createApi", () => {
     it("returns the last block information", async () => {
       const lastBlock = await api.lastBlock();
 
-      expect(lastBlock).toMatchObject({
-        height: expect.any(Number),
-      });
+      expect(lastBlock.height).toBeGreaterThan(0);
+      expect(lastBlock.hash?.length).toBe(64);
+      expect(lastBlock.time?.getTime()).toBeGreaterThan(0);
     });
   });
 
@@ -287,16 +287,33 @@ describe("createApi", () => {
       expect(operations.length).toBe(0);
     });
 
-    it("returns operations for real account with tokens", async () => {
+    it("returns operations with valid synthetic block info", async () => {
+      const lastPagingToken = "1753099264.927988000";
       const block = await api.lastBlock();
-      const [operations] = await api.listOperations(MAINNET_TEST_ACCOUNTS.withTokens.accountId, {
+      const [ops] = await api.listOperations(MAINNET_TEST_ACCOUNTS.withTokens.accountId, {
+        minHeight: block.height,
+        order: "asc",
+        limit: 4,
+        lastPagingToken,
+      });
+
+      const expectedSyntheticBlock = getSyntheticBlock(lastPagingToken);
+      const blockHeights = ops.map(o => o.tx.block.height);
+
+      expect(blockHeights).toHaveLength(4);
+      expect(blockHeights.every(h => h >= expectedSyntheticBlock.blockHeight)).toBe(true);
+    });
+
+    it("returns operations for real account with tokens", async () => {
+      const lastPagingToken = "1753099264.927988000";
+      const block = await api.lastBlock();
+      const [ops] = await api.listOperations(MAINNET_TEST_ACCOUNTS.withTokens.accountId, {
         minHeight: block.height,
         order: "desc",
         limit: 100,
-        lastPagingToken: "1753099264.927988000",
+        lastPagingToken,
       });
 
-      const ops = operations as unknown as Operation<HederaMemo>[];
       const memoTxHash = "WvMcFERtxRsGJqxqGVDYa6JR5PqLgFeJxiSVoimayaWra/AMEJMzC09LhdRLTZ/M";
       const operationWithMemo = ops.find(op => op.tx.hash === memoTxHash);
       const firstTokenAssociateOperations = ops.find(op => op.type === "ASSOCIATE_TOKEN");
--- a/libs/coin-modules/coin-hedera/src/api/index.ts
+++ b/libs/coin-modules/coin-hedera/src/api/index.ts
@@ -71,6 +71,7 @@ export function createApi(config: Record<string, never>): Api<HederaMemo> {
         fetchAllPages: false,
         skipFeesForTokenOperations: true,
         useEncodedHash: false,
+        useSyntheticBlocks: true,
       });
 
       const liveOperations = [
--- a/libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
@@ -1,9 +1,23 @@
 import type { BlockInfo } from "@ledgerhq/coin-framework/api/index";
+import { apiClient } from "../network/api";
+import { getSyntheticBlock } from "./utils";
 
-// NOTE: there are no "blocks" in hedera
-// Set a value just so that operations are considered confirmed according to isConfirmedOperation
+/**
+ * Gets the latest "block" information for Hedera.
+ *
+ * Hedera doesn't have actual blocks - it uses a timestamp-based consensus model.
+ * To make Hedera compatible with block-based architecture:
+ * 1. We fetch the most recent transaction from the mirror node
+ * 2. Extract its consensus timestamp
+ * 3. Convert this timestamp into a synthetic block using a hardcoded time window (10 seconds by default)
+ */
 export async function lastBlock(): Promise<BlockInfo> {
+  const latestTransaction = await apiClient.getLatestTransaction();
+  const syntheticBlock = getSyntheticBlock(latestTransaction.consensus_timestamp);
+
   return {
-    height: 10,
+    height: syntheticBlock.blockHeight,
+    hash: syntheticBlock.blockHash,
+    time: syntheticBlock.blockTime,
   };
 }
--- a/libs/coin-modules/coin-hedera/src/logic/listOperations.test.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/listOperations.test.ts
@@ -49,6 +49,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(apiClient.getAccountTransactions).toHaveBeenCalledTimes(1);
@@ -101,6 +102,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(result.coinOperations).toHaveLength(1);
@@ -171,6 +173,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(result.coinOperations).toHaveLength(1);
@@ -234,6 +237,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(result.coinOperations).toHaveLength(1);
@@ -294,6 +298,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(result.coinOperations).toHaveLength(0);
@@ -323,6 +328,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(apiClient.getAccountTransactions).toHaveBeenCalledTimes(1);
@@ -373,6 +379,7 @@ describe("listOperations", () => {
       fetchAllPages: true,
       skipFeesForTokenOperations: false,
       useEncodedHash: false,
+      useSyntheticBlocks: false,
     });
 
     expect(result.coinOperations).toHaveLength(1);
--- a/libs/coin-modules/coin-hedera/src/logic/listOperations.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/listOperations.ts
@@ -8,17 +8,20 @@ import { encodeOperationId } from "@ledgerhq/coin-framework/operation";
 import { apiClient } from "../network/api";
 import { parseTransfers } from "../network/utils";
 import type { HederaMirrorToken, HederaMirrorTransaction, HederaOperationExtra } from "../types";
-import { base64ToUrlSafeBase64, getMemoFromBase64 } from "./utils";
+import { base64ToUrlSafeBase64, getMemoFromBase64, getSyntheticBlock } from "./utils";
 
-function getCommonOperationData(rawTx: HederaMirrorTransaction, useEncodedHash: boolean) {
+function getCommonOperationData(
+  rawTx: HederaMirrorTransaction,
+  useEncodedHash: boolean,
+  useSyntheticBlocks: boolean,
+) {
   const timestamp = new Date(Number.parseInt(rawTx.consensus_timestamp.split(".")[0], 10) * 1000);
   const hash = useEncodedHash
     ? base64ToUrlSafeBase64(rawTx.transaction_hash)
     : rawTx.transaction_hash;
   const fee = new BigNumber(rawTx.charged_tx_fee);
   const hasFailed = rawTx.result !== "SUCCESS";
-  const blockHeight = 10;
-  const blockHash = null;
+  const syntheticBlock = getSyntheticBlock(rawTx.consensus_timestamp);
   const memo = getMemoFromBase64(rawTx.memo_base64);
   const extra: HederaOperationExtra = {
     pagingToken: rawTx.consensus_timestamp,
@@ -31,8 +34,8 @@ function getCommonOperationData(rawTx: HederaMirrorTransaction, useEncodedHash: 
     hash,
     fee,
     hasFailed,
-    blockHeight,
-    blockHash,
+    blockHeight: useSyntheticBlocks ? syntheticBlock.blockHeight : 10,
+    blockHash: useSyntheticBlocks ? syntheticBlock.blockHash : null,
     extra,
   };
 }
@@ -173,6 +176,7 @@ export async function listOperations({
   fetchAllPages,
   skipFeesForTokenOperations,
   useEncodedHash,
+  useSyntheticBlocks,
 }: {
   currency: CryptoCurrency;
   address: string;
@@ -182,6 +186,7 @@ export async function listOperations({
   fetchAllPages: boolean;
   skipFeesForTokenOperations: boolean;
   useEncodedHash: boolean;
+  useSyntheticBlocks: boolean;
 }): Promise<{
   coinOperations: Operation<HederaOperationExtra>[];
   tokenOperations: Operation<HederaOperationExtra>[];
@@ -205,7 +210,7 @@ export async function listOperations({
   });
 
   for (const rawTx of mirrorResult.transactions) {
-    const commonData = getCommonOperationData(rawTx, useEncodedHash);
+    const commonData = getCommonOperationData(rawTx, useEncodedHash, useSyntheticBlocks);
 
     // process token transfers
     const tokenResult = processTokenTransfers({
--- a/libs/coin-modules/coin-hedera/src/logic/utils.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/utils.ts
@@ -1,4 +1,5 @@
 import BigNumber from "bignumber.js";
+import { createHash } from "crypto";
 import invariant from "invariant";
 import { AccountId, Transaction as HederaSDKTransaction } from "@hashgraph/sdk";
 import { AssetInfo, TransactionIntent } from "@ledgerhq/coin-framework/api/types";
@@ -9,7 +10,11 @@ import { InvalidAddress } from "@ledgerhq/errors";
 import { makeLRUCache, seconds } from "@ledgerhq/live-network/cache";
 import type { Currency, ExplorerView, TokenCurrency } from "@ledgerhq/types-cryptoassets";
 import type { AccountLike, Operation as LiveOperation } from "@ledgerhq/types-live";
-import { HEDERA_OPERATION_TYPES, HEDERA_TRANSACTION_MODES } from "../constants";
+import {
+  HEDERA_OPERATION_TYPES,
+  HEDERA_TRANSACTION_MODES,
+  SYNTHETIC_BLOCK_WINDOW_SECONDS,
+} from "../constants";
 import { apiClient } from "../network/api";
 import type {
   HederaAccount,
@@ -224,3 +229,27 @@ export const safeParseAccountId = (
     return [new InvalidAddress("", { currencyName }), null];
   }
 };
+
+/**
+ * Calculates a synthetic block height based on Hedera consensus timestamp.
+ *
+ * @param consensusTimestamp - Hedera consensus timestamp
+ * @param blockWindowSeconds - Duration of one synthetic block in seconds (default: 10)
+ * @returns Deterministic synthetic block (height and hash)
+ */
+export function getSyntheticBlock(
+  consensusTimestamp: string,
+  blockWindowSeconds = SYNTHETIC_BLOCK_WINDOW_SECONDS,
+) {
+  const seconds = Math.floor(Number(consensusTimestamp));
+
+  if (Number.isNaN(seconds) || seconds === 0) {
+    throw new Error(`Invalid consensus timestamp: ${consensusTimestamp}`);
+  }
+
+  const blockHeight = Math.floor(seconds / blockWindowSeconds);
+  const blockHash = createHash("sha256").update(blockHeight.toString()).digest("hex");
+  const blockTime = new Date(seconds * 1000);
+
+  return { blockHeight, blockHash, blockTime };
+}
--- a/libs/coin-modules/coin-hedera/src/network/api.ts
+++ b/libs/coin-modules/coin-hedera/src/network/api.ts
@@ -132,9 +132,28 @@ async function getAccountTokens(address: string): Promise<HederaMirrorToken[]> {
   return tokens;
 }
 
+async function getLatestTransaction(): Promise<HederaMirrorTransaction> {
+  const params = new URLSearchParams({
+    limit: "1",
+    order: "desc",
+  });
+
+  const res = await fetch<HederaMirrorTransactionsResponse>(
+    `/api/v1/transactions?${params.toString()}`,
+  );
+  const transaction = res.data.transactions[0];
+
+  if (!transaction) {
+    throw new Error("No transactions found on the Hedera network");
+  }
+
+  return transaction;
+}
+
 export const apiClient = {
   getAccountsForPublicKey,
   getAccount,
   getAccountTokens,
   getAccountTransactions,
+  getLatestTransaction,
 };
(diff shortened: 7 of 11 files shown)


libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
```diff
@@ -0,0 +1,9 @@
+import type { BlockInfo } from "@ledgerhq/coin-framework/api/index";
+
+// NOTE: there are no "blocks" in hedera
+// Set a value just so that operations are considered confirmed according to isConfirmedOperation
+export async function lastBlock(): Promise<BlockInfo> {
+  return {
+    height: 10,
```
acewf: don't follow why we have a function call, to get a constant value.

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
```diff
@@ -0,0 +1,158 @@
+import BigNumber from "bignumber.js";
+import type { Operation } from "@ledgerhq/types-live";
+import type { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import type { Pagination } from "@ledgerhq/coin-framework/api/types";
+import { findTokenByAddressInCurrency } from "@ledgerhq/cryptoassets/tokens";
+import { encodeAccountId, encodeTokenAccountId } from "@ledgerhq/coin-framework/account/accountId";
+import { encodeOperationId } from "@ledgerhq/coin-framework/operation";
+import { hederaMirrorNode } from "../network/mirror";
```
acewf: i get that mirror node is a specific hedera language, but at the same time feel code on our application should use more less same terms across the different chains.

libs/coin-modules/coin-hedera/src/network/mirror.ts
```diff
@@ -0,0 +1,140 @@
...
+  return network<Result>({
+    method: "GET",
+    url: `${getMirrorApiUrl()}${path}`,
+  });
+};
+
+async function getAccountsForPublicKey(publicKey: string): Promise<HederaMirrorAccount[]> {
+  let r;
+  try {
+    r = await fetch<HederaMirrorAccountsResponse>(
```
acewf: in the following function `getAccount` we have 
```
const res = await fetch()
```

but on here we have just an `r`, can we make consistent please.

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
md-blockydevs: Yup, I agree, it will be much easier to read for anyone who is not familiar with Hedera. 
Would `hederaMirrorNode -> hederaAPI` and `hederaClient -> hederaSDK` work?

libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
md-blockydevs: Yeah, Hedera doesn't have blocks and I've added this in a separate file for consistency. Another option would be to inline it in `api/index.ts`, but I'm not sure which approach is better

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
acewf: we have already the context of the folder, that is the `coin-hedera`.
It seems there isn't much an aligned pattern between coins, but makes more sense to me be something like `Node` if we are just doing rcp calls.

libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
jnicoulaud-ledger: Not a blocker for this PR, but this will be problematic, as this is what we use to perform incremental block synchronization. My understanding is that Hedera does have a notion of block, even if it's a bit artificial:
 - https://docs.hedera.com/hedera/sdks-and-apis/rest-api/blocks
 - https://mainnet.mirrornode.hedera.com/api/v1/docs/#/blocks/getBlocks
 - https://hedera.com/blog/introducing-block-streams-a-unified-data-stream-capturing-the-comprehensive-history-of-the-hedera-network

md-blockydevs: Ok, so `pagination.lastPagingToken` won't solve anything here? I will report this to our internal team as well, because here is an example of what data we are getting now: https://mainnet-public.mirrornode.hedera.com/api/v1/transactions?account.id=0.0.8835924 (incremental sync was built around `consensus_timestamp`)

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
md-blockydevs: Ok, I just went with `apiClient` and `rpcClient` (will be there once https://github.com/LedgerHQ/ledger-live/pull/12270/files is merged)

libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
jnicoulaud-ledger: Yes, it will help, we have 2 systems built on top of alpaca API:
 - lama-adapter, which does synchronization per account using listOperations => `lastPagingToken` will work, all we need is to be able to resume the stream of operations using some cursor. However, the operations contain block info, and we must make sure it is consistent with `lastBlock`, even if we decide to use a fake block height (such as timestamp)
 - A4, which indexes blocks using `getBlock`. `lastBlock` is used here to detect if we have new blocks to index, so the block height must be consistent between `getBlock` and `lastBlock` as well
I need to dig a bit more how Hedera works so we can have a solution to covers both... Either we use the Block Streams numbers, or a fake height / timestamps, with 1 block = 1 to N transactions

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
```diff
@@ -0,0 +1,374 @@
...
+
+      expect(rawTx).toBeInstanceOf(TransferTransaction);
+      invariant(rawTx instanceof TransferTransaction, "TransferTransaction type guard");
+
+      const sendTransfer = rawTx.hbarTransfers.get(MAINNET_TEST_ACCOUNTS.withoutTokens.accountId);
+      const receiveTransfer = rawTx.hbarTransfers.get(MAINNET_TEST_ACCOUNTS.withTokens.accountId);
+      const sendAmount = sendTransfer?.toBigNumber().toString();
+      const receiveAmount = receiveTransfer?.toBigNumber().toString();
+
+      expect(rawTx.hbarTransfers);
```
Moustafa-Koterba: Using `expect` alone can be vague and not complete testing (only truthiness here). May be something like:
```ts
expect(rawTx.hbarTransfers).toBeDefined();
expect(rawTx.hbarTransfers.size).toBe(2);
expect(rawTx.hbarTransfers.keys).toContain(MAINNET_TEST_ACCOUNTS.withoutTokens.accountId);
expect(rawTx.hbarTransfers.keys).toContain(MAINNET_TEST_ACCOUNTS.withTokens.accountId);
``` 
same for other tests

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
```diff
@@ -0,0 +1,374 @@
...
+      expect(sendAmount).toBe("-1");
```
Moustafa-Koterba: I think you don't need to convert to string here, `sendAmount` has the type `Hbar` which can be tested like:
```ts
expect(sendTransfer).toEqual(Hbar.from(-1, HbarUnit.Hbar));
```
same for others

libs/coin-modules/coin-hedera/src/api/index.ts
```diff
@@ -0,0 +1,145 @@
...
+import type { HederaMemo } from "../types";
+
+export function createApi(config: HederaConfig): Api<HederaMemo> {
+  coinConfig.setCoinConfig(() => ({ ...config, status: { type: "active" } }));
+  const currency = getCryptoCurrencyById("hedera");
+
+  return {
+    broadcast: async tx => {
+      const response = await logicBroadcast(tx);
+      const base64Hash = Buffer.from(response.transactionHash).toString("base64");
```
Moustafa-Koterba:
```suggestion
      return Buffer.from(response.transactionHash).toString("base64");
```

libs/coin-modules/coin-hedera/src/logic/broadcast.ts
```diff
@@ -0,0 +1,10 @@
+import type { TransactionResponse } from "@hashgraph/sdk";
+import { hederaClient } from "../network/client";
+import { deserializeTransaction } from "./utils";
+
+export const broadcast = async (txWithSignature: string): Promise<TransactionResponse> => {
+  const hederaTransaction = deserializeTransaction(txWithSignature);
+  const response = await hederaClient.broadcastTransaction(hederaTransaction);
```
Moustafa-Koterba:
```suggestion
  return await hederaClient.broadcastTransaction(hederaTransaction);
```

libs/coin-modules/coin-hedera/src/logic/combine.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    const mockSerializedTxWithSignature = "serialized-transaction-with-signature";
+
+    (deserializeTransaction as jest.Mock).mockReturnValue(mockDeserializedTx);
+    (deserializeSignature as jest.Mock).mockReturnValue(mockBufferSignature);
+    (PublicKey.fromString as jest.Mock).mockReturnValue(mockBufferPublicKey);
+    (serializeTransaction as jest.Mock).mockReturnValue(mockSerializedTxWithSignature);
+
+    const result = combine(tx, signature, publicKey);
+
+    expect(deserializeTransaction).toHaveBeenCalledWith(tx);
```
Moustafa-Koterba, line 31: Usually, we check times call with mock, you can add this to ensure that your function called one time (or more) the mock
```suggestion
    expect(deserializeTransaction).toHaveBeenCalledTimes(1);
```

libs/coin-modules/coin-hedera/src/logic/craftTransaction.test.ts
```diff
@@ -0,0 +1,214 @@
...
+          type: "string",
+          value: "Test memo with custom fee",
+        },
+      },
+      customFees,
+    );
+
+    expect(result.tx).toBeInstanceOf(sdk.TransferTransaction);
+    invariant(result.tx instanceof sdk.TransferTransaction, "TransferTransaction type guard");
+    expect(result.tx.maxTransactionFee?.toTinybars().toString()).toBe(customFees.value.toString());
```
Moustafa-Koterba: Conversion to string is not necessary and can confuse reader (why doing it), values from the SDK are comparable (like what it was said in a previous file). Please update your PR to avoid conversion to string in your tests

libs/coin-modules/coin-hedera/src/logic/estimateFees.test.ts
Moustafa-Koterba, line 1: same here, no need for conversion
```ts
expect(result.isEqualTo(expected)).toEqual(true);
```

libs/coin-modules/coin-hedera/src/logic/estimateFees.ts
```diff
@@ -0,0 +1,27 @@
...
+  let fee: BigNumber | undefined;
+  const usdRate = await getCurrencyToUSDRate(currency).catch(() => null);
+
+  if (usdRate) {
+    fee = new BigNumber(BASE_USD_FEE_BY_OPERATION_TYPE[operationType])
+      .dividedBy(new BigNumber(usdRate))
+      .integerValue(BigNumber.ROUND_CEIL)
+      .multipliedBy(ESTIMATED_FEE_SAFETY_RATE);
+  } else {
+    fee = new BigNumber("150200").multipliedBy(ESTIMATED_FEE_SAFETY_RATE);
```
Moustafa-Koterba: Is `150200` a constant ? Please put it into a variable

libs/coin-modules/coin-hedera/src/logic/getAssetFromToken.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      units: [
+        {
+          name: "Test",
+          code: "TEST",
+          magnitude: 8,
+        },
+      ],
+    });
+
+    expect(getAssetFromToken(token, owner)).toEqual({
```
Moustafa-Koterba, line 19:
```suggestion
    expect(getAssetFromToken(token, owner)).toEqual({
    assetReference: token.contractAddress,
      assetOwner: token.owner,
      name: token.name,
      type: "hts",
      unit: {
        code: token.unit[0].code,
        magnitude: token.unit[0].magnitude,
        name: token.unit[0].name,
      },
    });
  });
```

libs/coin-modules/coin-hedera/src/logic/getAssetFromToken.ts
```diff
@@ -0,0 +1,12 @@
+import type { AssetInfo } from "@ledgerhq/coin-framework/api/types";
+import type { TokenCurrency } from "@ledgerhq/types-cryptoassets";
+
+export function getAssetFromToken(token: TokenCurrency, owner: string): AssetInfo {
```
Moustafa-Koterba, line 4: wdyt
```suggestion
export function toAsset(token: TokenCurrency, owner: string): AssetInfo {
```

libs/coin-modules/coin-hedera/src/logic/getBalance.test.ts
```diff
@@ -0,0 +1,190 @@
...
+        type: "hts",
+        assetReference: "0.0.7890",
+        assetOwner: address,
+        name: "Test Token 1",
+        unit: { name: "TT1", code: "tt1", magnitude: 6 },
+      },
+    });
+  });
+
+  it("should handle errors from hederaMirrorNode.getAccount", async () => {
```
Moustafa-Koterba:
```suggestion
  it("should throw an error when getAccount throws an error", async () => {
```
same for others

libs/coin-modules/coin-hedera/src/logic/getBalance.test.ts
```diff
@@ -0,0 +1,190 @@
...
+        balance: "1000000000",
+      },
+    };
+
+    (hederaMirrorNode.getAccount as jest.Mock).mockResolvedValue(mockMirrorAccount);
+    (hederaMirrorNode.getAccountTokens as jest.Mock).mockResolvedValue([]);
+
+    const result = await getBalance(mockCurrency, address);
+
+    expect(hederaMirrorNode.getAccount).toHaveBeenCalledWith(address);
```
Moustafa-Koterba: add `.toHaveBeenCalledTimes(1);` for your mocks 
same for other tests cases

libs/coin-modules/coin-hedera/src/logic/getBalance.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    expect(hederaMirrorNode.getAccountTokens).toHaveBeenCalledWith(address);
+    expect(result).toHaveLength(1);
+    expect(result[0]).toEqual({
```
Moustafa-Koterba: Try may be to let just `toEqual`, you should not need `toHaveLength`
```suggestion
    expect(result).toEqual([{
```
same for others cases

---
for other cases, you may need to check on longer array
You can do something like (to avoid sorted check)
```ts
expect(result).toEqual(expect.arrayContaining([{ /* your array */ }]);
```

libs/coin-modules/coin-hedera/src/logic/getBalance.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    expect(result).toHaveLength(3);
+    expect(result[0]).toEqual({
+      asset: { type: "native" },
+      value: BigInt("1000000000"),
+    });
+    expect(result[1]).toEqual({
+      value: BigInt("5000"),
+      asset: {
+        type: "hts",
+        assetReference: "0.0.7890",
```
Moustafa-Koterba:
```suggestion
        assetReference: mockToken1.contractAddress,
```
same for others fields and other tests

libs/coin-modules/coin-hedera/src/logic/getTokenFromAsset.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  it("returns token from USDC asset", async () => {
+    const asset1 = { type: "hts", assetReference: "0.0.5022567" };
+    const asset2 = { type: "hts", assetReference: "0.0.456858" };
+
+    expect(await getTokenFromAsset(mockCurrency, asset1)).toMatchObject({
+      id: "hedera/hts/hbark_0.0.5022567",
+      contractAddress: "0.0.5022567",
+      name: "hBARK",
+    });
+    expect(await getTokenFromAsset(mockCurrency, asset2)).toMatchObject({
```
Moustafa-Koterba: May be a dumb question, but why 2 cases ? One is enough

libs/coin-modules/coin-hedera/src/logic/getTokenFromAsset.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getTokenFromAsset } from "./getTokenFromAsset";
+import { getMockedCurrency } from "../test/fixtures/currency.fixture";
+
+describe("getTokenFromAsset", () => {
```
Moustafa-Koterba, line 4: you did not test the case when `getTokenFromAsset` return `undefined`

libs/coin-modules/coin-hedera/src/logic/listOperations.test.ts
```diff
@@ -0,0 +1,370 @@
...
+      mirrorTokens: [],
+      fetchAllPages: true,
+      skipFeesForTokenOperations: false,
+      useEncodedHash: false,
+    });
+
+    expect(result.coinOperations).toHaveLength(1);
+    expect(result.tokenOperations).toHaveLength(0);
+
+    expect(result.coinOperations[0]).toMatchObject({
```
Moustafa-Koterba:
```suggestion
    expect(result.coinOperations).toEqual([{
```

libs/coin-modules/coin-hedera/src/logic/listOperations.test.ts
```diff
@@ -0,0 +1,370 @@
...
+      address,
+      pagination,
+      mirrorTokens: [],
+      fetchAllPages: true,
+      skipFeesForTokenOperations: false,
+      useEncodedHash: false,
+    });
+
+    expect(result.coinOperations).toHaveLength(1);
+    expect(result.tokenOperations).toHaveLength(0);
```
Moustafa-Koterba:
```suggestion
    expect(result.tokenOperations).toEqual([]);
```

libs/coin-modules/coin-hedera/src/logic/utils.ts
```diff
@@ -0,0 +1,226 @@
...
+  if (txIntent.type === HEDERA_TRANSACTION_MODES.Send && txIntent.asset.type !== "native") {
+    return HEDERA_OPERATION_TYPES.TokenTransfer;
+  }
+
+  return HEDERA_OPERATION_TYPES.CryptoTransfer;
+};
+
+export const getMemoFromBase64 = (memoBase64: string | undefined): string | null => {
+  try {
+    if (typeof memoBase64 !== "string") return null;
```
Moustafa-Koterba:
```suggestion
    if (memoBase64 === undefined) return null;
```

libs/coin-modules/coin-hedera/src/network/client.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  };
+});
+
+describe("getClient", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    hederaClient._resetClient();
+  });
+
+  it("reuses the same client instance for multiple calls", async () => {
```
Moustafa-Koterba: Good test, but better to check the parameter passed to the `execute` function of mocked transaction
Get a reference for the returned mocked module at the top:
```ts
const client = {
        setMaxNodesPerTransaction: jest.fn(() => mockClients[key]),
        setNetwork: jest.fn(() => mockClients[key]),
      } as unknown as Client;

jest.mock("@hashgraph/sdk", () => {
  return {
    Transaction: jest.fn(),
    Client: {
      forMainnet: () => client,
    },
  };
});
```

Then assert in your tests that it is the same instance called by execute functions:
```ts
const mockedExecute = jest.fn();
const mockTransaction1 = {
      execute: mockedExecute, // no need to return value I think here, you check only client
} as unknown as Transaction;

// ...
expect(mockedExecute).toHaveBeenCalledTimes(1);
expect(mockedExecute).toHaveBeenCalledWith(client);
```

Something like that, from my mind, can be wrong on syntax

libs/coin-modules/coin-hedera/src/network/client.test.ts
```diff
@@ -0,0 +1,65 @@
+import { Client, Transaction, TransactionResponse } from "@hashgraph/sdk";
+import { hederaClient } from "./client";
+
+jest.mock("@hashgraph/sdk", () => {
+  const mockClients: Record<string, Client> = {};
```
Moustafa-Koterba: why did you store this ?

libs/coin-modules/coin-hedera/src/network/api.ts
```diff
@@ -0,0 +1,140 @@
...
+async function getAccountTokens(address: string): Promise<HederaMirrorToken[]> {
+  const tokens: HederaMirrorToken[] = [];
+  const params = new URLSearchParams({
+    limit: "100",
+  });
+
+  let nextUrl: string | null = `/api/v1/accounts/${address}/tokens?${params.toString()}`;
+
+  while (nextUrl) {
+    const res: LiveNetworkResponse<HederaMirrorAccountTokensResponse> = await fetch(nextUrl);
```
Moustafa-Koterba, line 126: Dumb question, so here, the `limit` parameters handle the size automatically ? There is no `nextUrl` once we reach that number ?

libs/coin-modules/coin-hedera/src/network/api.ts
```diff
@@ -0,0 +1,140 @@
...
+    nextCursor = lastTx?.consensus_timestamp ?? null;
+  }
+
+  return { transactions, nextCursor };
+}
+
+async function getAccountTokens(address: string): Promise<HederaMirrorToken[]> {
+  const tokens: HederaMirrorToken[] = [];
+  const params = new URLSearchParams({
+    limit: "100",
```
Moustafa-Koterba, line 120: Should we put this as parameter with a default value ? I dont know if it is part of an interface or something that should respect this signature 
```ts
getAccountTokens(address: string)
```

Moustafa-Koterba:
Thank you for your PR @scavassa-yld 
I left some comments, mainly on tests
Please review them to improve some points:
- do not convert to string numbers to compare/assert on them, there is no need for this. It can confuse the reader to understand the behaviour and add an overhead (even small) to the tests
- When asserting a mock function call, a good thing is to check the number of calls with `.toHaveBeenCalledTimes` to ensure the behaviour and the pass used in the function
- mock only what is necessary, and do not return things if they are not used, it can confuse the reader also

libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
md-blockydevs: Sure, `getBlock` just throws an error for now, so it's probably another thing that must be adjusted. When it comes to the last block information, maybe we could also use the timestamp of last account's transaction (or its creation_date if there are no transactions yet)? We would just have to round this up to make sure transactions gets confirmed. Example data: https://mainnet-public.mirrornode.hedera.com/api/v1/accounts/0.0.8835924

acewf (dismissed):

libs/coin-modules/coin-hedera/src/api/index.integ.test.ts
md-blockydevs: Yup, it's leftover and shouldn't be there at all - good catch, thanks

libs/coin-modules/coin-hedera/src/network/api.ts
md-blockydevs: I don't think we would need it as a parameter right now. This function just fetches token balances, and we always need to get everything, so the limit is fixed to the mirror node's maximum

libs/coin-modules/coin-hedera/src/network/api.ts
md-blockydevs: We can have 10k transactions, but there is a maximum limit of 100 transactions per page in mirror node (REST api for Hedera). Previously we were fetching everything, until the `res.data.links.next` is null, but with Alpaca we need to have a paginated mode. This is why I've added `fetchAllPages` logic and early break if we have enough data, depending if old bridge is used or the Alpaca API is called.

Does this answer your question? 😄

libs/coin-modules/coin-hedera/src/logic/getAssetFromToken.ts
md-blockydevs: Looks good, but I tried to keep it consistent with generic interface:
image

libs/coin-modules/coin-hedera/src/logic/estimateFees.ts
md-blockydevs: It's basically a magic number, previous implementation:

```
// as fees are based on a currency conversion, we stay
// on the safe side here and double the estimate for "max spendable"
return new BigNumber("150200").multipliedBy(ESTIMATED_FEE_SAFETY_RATE); // 0.001502 ℏ (as of 2023-03-14)
```

Good idea, I will move this to `DEFAULT_TINYBAR_FEE`

libs/coin-modules/coin-hedera/src/network/client.test.ts
md-blockydevs: This is a leftover actually that will be replaced with your idea from https://github.com/LedgerHQ/ledger-live/pull/12254#discussion_r2428489827. Previously I tried to work with `testnet` and `mainnet` and this was created to cache different instances + test whether the `getClient` (or `getInstance` now) behaves correctly, as a singleton

libs/coin-modules/coin-hedera/src/logic/lastBlock.ts
jnicoulaud-ledger: So it seems you already use `consensus_timestamp` as paging token. Couldn't that fit ? It must be absolute (not specific to an account). Client calls `lastBlock`, then determines which account to sync based on that.

For `getBlock`, we can do virtual blocks (using fixed window of `[consensus_timestamp, consensus_timestamp+ window size]`) so that it's all consistent

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (tokenTransfers.length > 0) {
+      const tokenId = tokenTransfers[0].token_id;
+      const token = findTokenByAddressInCurrency(tokenId, currency.id);
+      if (!token) continue;
+
+      const encodedTokenId = encodeTokenAccountId(ledgerAccountId, token);
+      const { type, value, senders, recipients } = parseTransfers(tokenTransfers, address);
+      const extra = { ...commonExtra };
+
+      // add main FEES coin operation for send token transfer
```
jnicoulaud-ledger: I don't understand why fees are split to a separate operation, it should be on same operation (fee is assumed to be native coin)

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
```diff
@@ -0,0 +1,158 @@
...
+        senders,
+        hash,
+        fee,
+        date: timestamp,
+        blockHeight,
+        blockHash,
+        hasFailed,
+        extra,
+      });
+    } else if (transfers.length > 0) {
```
jnicoulaud-ledger: Can't we have both transfers and token_transfers ?

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
md-blockydevs: This is the old bridge behavior. I've added a `skipFeesForTokenOperations` flag to disable this with Alpaca. This is what happens in LLD when `Account 1` sends HTS token to `Account 2`:

- Account 1 renders "FEES" operation in the history, with token "OUT" operation linked (suboperation)
image  

- Account 1 > hBARK renders token "OUT" operation
image  

- Account 2 > hBARK renders token "IN" operation
image

libs/coin-modules/coin-hedera/src/logic/listOperations.ts
md-blockydevs: Actually this is a very good catch and edge case that may happen:
```
const tx = new TransferTransaction()
    .addHbarTransfer(operatorId, new Hbar(0.01).negated())
    .addHbarTransfer(recipient, new Hbar(0.01))
    .addTokenTransfer(token, operatorId, -1)
    .addTokenTransfer(token, recipient, 1);
 ```

 I have never encountered this with normal wallet usage, but the SDK supports this feature and it is worth including. I will fix this. Example of multi-transfer transaction (only-token transfers have `transfers` defined as well, but only fees were there, so this is why it was ignored): https://mainnet-public.mirrornode.hedera.com/api/v1/transactions/0.0.8835924-1760508445-680524572

File: libs/coin-modules/coin-hedera/src/logic/estimateFees.test.ts
Comment (acewf) [reply libs/coin-modules/coin-hedera/src/logic/estimateFees.test.ts:1]: discussed this with @md-blockydevs on this [PR](https://github.com/LedgerHQ/ledger-live/pull/12305#discussion_r2431559452) what is benefit of having a longer and non-standard test, when the result seems to be the same as the shorter version.

```
expect(result).toEqual(expected);
```

libs/coin-modules/coin-hedera/src/network/api.ts
Moustafa-Koterba: Thank you for the explanation

PR #12302, opened 2025-10-14
Feat/LIVE-21984 receive flow onboarding

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

For the new onboarding flow we will be sending the user to the receive flow.  We have to add state to allow the onboarding flow to determine when the user has successfully completed the receive flow, ie. has reached the point where they have verified the address, and we also have to show only receive flow meaning no staking flow or modal asking if the user wants to buy.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-21984

---

### DIFF
--- a/.changeset/lazy-wolves-retire.md
+++ b/.changeset/lazy-wolves-retire.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": minor
+---
+
+Add receive flow onboarding action and logic


apps/ledger-live-desktop/src/renderer/modals/Receive/index.tsx
```diff
@@ -27,24 +28,30 @@ function getInitialState(
...
+    stepId: !isOnboardingFlow && shouldUseReceiveOptions ? "receiveOptions" : "account",
     isAddressVerified: null,
     verifyAddressError: null,
   };
 }
 
 const ReceiveModal = (props: GlobalModalData["MODAL_RECEIVE"]) => {
   const { enabled } = useFeature("noah") || {};
-  const initialState = getInitialState(enabled === true, props?.account, props?.parentAccount);
+  const isOnboardingReceiveFlow = useSelector(onboardingReceiveFlowSelector);
```
thesan, line 48: Here the `initialState` is calculated based on `isOnboardingReceiveFlow` which is reactive. Is it safe ?

mitchellv-ledger: Yes it is safe :) it is initialised to false for both state variables by default and resets the main variable which changes the flow back to false whenever the modal is closed in `apps/ledger-live-desktop/src/renderer/modals/Receive/Body.tsx`

PR #12277, opened 2025-10-13
feat: make CryptoCurrency dynamically loadable (usecase: Market)

gre-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  - once https://github.com/LedgerHQ/ledger-live/pull/12276 is merged, this will have no impact for users

### 📝 Description

any usage of dada-client can now pull CryptoCurrency that aren't even yet in our local data, allowing to integrate them in Market DaDa rework.

We must have https://github.com/LedgerHQ/ledger-live/pull/12276 landed first.

### ❓ Context

- **JIRA or GitHub link**: LIVE-22244

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #12354, opened 2025-10-17
type: improve CryptoCurrencyId type

gre-ledger:
### ✅ Checklist

  - only type system

### 📝 Description

*unfortunately this PR looks huge but it's only just changing types. i won't have a lot of budget to keep it, so merge it or lose it.*

We have recently switched CryptoCurrencyId to `string` type to make it dynamic ( https://github.com/LedgerHQ/ledger-live/pull/12277 ), however by doing this we lose some type-safety on using .id. 

This PR is a follow up of feedback from @francois-guerin-ledger  on the fact we should improve CryptoCurrencyId to make the type stronger using branded types.

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/ledgerjs/packages/types-cryptoassets/src/index.ts
```diff
@@ -1,6 +1,17 @@
...
+ * Branded type for CryptoCurrencyId to prevent mixing with regular strings
+ * and ensure type safety across different ID types.
+ */
+export type CryptoCurrencyId = string & { [cryptoCurrencyIdTag]: "CryptoCurrencyId" };
+
+/**
+ * Helper function to create a CryptoCurrencyId from a string.
+ * Use this when you need to create or cast a string to CryptoCurrencyId.
+ */
+export const toCryptoCurrencyId = (id: string): CryptoCurrencyId => id as CryptoCurrencyId;
```
francois-guerin-ledger, line 14: Maybe this wording is too specific. If we want to add more in the future, a solution like this scales better
```ts
export const asId = <Id extends string>(id: string): Id => id as Id;
```
What do you think ?

gre-ledger: yeah this could be an idea. but would it infer correctly or do i have to `asId<CryptoCurrencyId>(id)` on usage?

francois-guerin-ledger: It is on usage since you are passing a `string` anyway

gre-ledger:
I don't have time to maintain this :(

PR #12357, opened 2025-10-17
Feat/onboarding arrow buttons

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Updates the header buttons to be the back arrow for onboarding. This ticket was previously merged and found to have issues with settings which came from `apps/ledger-live-mobile/src/navigation/navigatorConfig.tsx`. I have added a fix to prevent the issue from occuring.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-20224

---

### DIFF
--- a/apps/ledger-live-mobile/src/screens/SyncOnboarding/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/SyncOnboarding/index.tsx
@@ -23,6 +23,8 @@ import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
 import { NavigationHeaderBackButton } from "~/components/NavigationHeaderBackButton";
 import { NavigatorName, ScreenName } from "~/const";
 import { RootNavigation } from "~/components/RootNavigator/types/helpers";
+import { hasCompletedOnboardingSelector } from "~/reducers/settings";
+import { useSelector } from "react-redux";
 
 const POLLING_PERIOD_MS = 1000;
 const DESYNC_TIMEOUT_MS = 20000;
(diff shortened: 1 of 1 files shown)


ofreyssinet-ledger (approved):

apps/ledger-live-mobile/src/screens/SyncOnboarding/index.tsx
```diff
@@ -70,9 +74,36 @@ export const SyncOnboarding = ({ navigation, route }: SyncOnboardingScreenProps)
...
+                        name: ScreenName.OnboardingDeviceSelection,
+                      },
+                    ],
+                  },
+                },
+              ],
+            },
+          },
+        ],
+      });
```
OlivierFreyssinet: Does that work if LW is already onboarded and the sync onboarding is started by adding an unseeded device via My Ledger ?

OlivierFreyssinet: I don't remember in that case what's the navigation route but this hardcoded route might cause issues in that case

mitchellv-ledger: You are correct, I have fixed the issue by checking if there is a device already onboarded and if so does the popToTop instead. This code is to fix the bug where you end up seeing an empty screen when you press the back button in no devices onboarded onboarding flow

vcluzeau-ledger (approved):

PR #12402, opened 2025-10-21
fix: use BigNumber and bigint for sequence

qperrot:
### ✅ Checklist

  - ...

### 📝 Description

Stellar sequence is larger than the type `Number` is allowed.
We replace it by bigint for alpaca and BigNumber on LL

### ❓ Context

- **JIRA or GitHub link**: [LIVE-20362](https://ledgerhq.atlassian.net/browse/LIVE-20362)

---

[LIVE-20362]: https://ledgerhq.atlassian.net/browse/LIVE-20362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-framework/src/account/pending.test.ts
```diff
@@ -279,7 +279,7 @@ export function createOperation(
     recipients: [],
     blockHeight: undefined,
     blockHash: undefined,
-    transactionSequenceNumber: sequenceNumber,
+    transactionSequenceNumber: new BigNumber(sequenceNumber?.toString() ?? 0),
```
Justkant: Maybe we should not default to 0 and respect the undefined we receive as you did in `groupOperations.test.ts`

libs/coin-framework/src/serialization/operation.ts
```diff
@@ -168,7 +168,10 @@ export const fromOperationRaw = (
   };
 
   if (transactionSequenceNumber !== undefined) {
-    res.transactionSequenceNumber = transactionSequenceNumber;
+    const transactionSequenceNumberBN = new BigNumber(transactionSequenceNumber);
+    res.transactionSequenceNumber = transactionSequenceNumberBN.isNaN()
+      ? new BigNumber(0)
+      : transactionSequenceNumberBN;
```
Justkant: Maybe we shouldn't set the sequence number if it's NaN, wdyt ?

ishaba (approved):

libs/coin-modules/coin-cosmos/src/synchronisation.ts
```diff
@@ -102,7 +102,9 @@ const getBlankOperation = (tx: CosmosTx, fees: BigNumber, accountId: string): Co
     accountId,
     date: new Date(tx.timestamp),
     extra: {},
-    transactionSequenceNumber: parseInt(tx?.tx?.auth_info?.signer_infos[0]?.sequence || "0"),
+    transactionSequenceNumber: new BigNumber(
+      parseInt(tx?.tx?.auth_info?.signer_infos?.[0]?.sequence ?? "0"),
```
Justkant: You might want to keep the `||` because it means an empty string will not fallback to `"0"` now and I don't know what parseInt will return

We could also probably drop the parseInt as BigNumber already works with string

Justkant: Also this parseInt doesn't use the base 10 param and it means the radix will be inferred from the string, it can be something else than 10

libs/coin-modules/coin-evm/src/adapters/etherscan.ts
```diff
@@ -223,7 +223,7 @@ export const etherscanERC1155EventToOperations = (
         recipients: [to],
         blockHeight: parseInt(event.blockNumber, 10),
         blockHash: event.blockHash,
-        transactionSequenceNumber: parseInt(event.nonce, 10),
+        transactionSequenceNumber: new BigNumber(parseInt(event.nonce, 10)),
```
Justkant: We can probably drop all the parseInt now

libs/coin-modules/coin-evm/src/bridge/prepareTransaction.ts
```diff
@@ -264,15 +264,15 @@ export const prepareForSignOperation = async ({
...
+    : Number(await getSequence(account.currency, account.freshAddress));
 
   if (isNftTransaction(transaction)) {
     return {
       ...transaction,
       amount: new BigNumber(0), // amount set to 0 as we're interacting with a smart contract
       recipient: transaction.nft.contract, // recipient is then the NFT smart contract
       // data as already been added by the `prepareTokenTransaction` method
-      nonce,
+      nonce: Number(nonce),
```
Justkant: Why do you only update the nonce here and not below for the token or the last return ?

It might be better to just update the nonce before the ifs in the ternary above

libs/coin-modules/coin-evm/src/logic/estimateFees.ts
```diff
@@ -152,7 +152,7 @@ export async function estimateFees(
   const unsignedTransaction: TransactionLike = {
     type,
     to,
-    nonce,
+    nonce: Number(nonce),
```
Justkant: Maybe we should update `TransactionLike` to use a bigint instead of keeping it a number, wdyt ?

libs/coin-modules/coin-evm/src/logic/getSequence.ts
```diff
@@ -1,6 +1,7 @@
 import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
 import { getNodeApi } from "../network/node";
 
-export function getSequence(currency: CryptoCurrency, address: string): Promise<number> {
-  return getNodeApi(currency).getTransactionCount(currency, address);
+export async function getSequence(currency: CryptoCurrency, address: string): Promise<bigint> {
+  const txCount = await getNodeApi(currency).getTransactionCount(currency, address);
+  return BigInt(txCount?.toString() || 0);
```
Justkant: You should be able to simply use the number here
```suggestion
  return BigInt(txCount);
```

libs/coin-modules/coin-icon/src/logic.ts
```diff
@@ -56,8 +56,8 @@ export const getNonce = (account: IconAccount): number => {
 
   const nonce = Math.max(
     account.iconResources?.nonce || 0,
-    lastPendingOp && typeof lastPendingOp.transactionSequenceNumber === "number"
-      ? lastPendingOp.transactionSequenceNumber + 1
+    lastPendingOp && BigNumber.isBigNumber(lastPendingOp.transactionSequenceNumber)
+      ? Number(lastPendingOp.transactionSequenceNumber.plus(1))
```
Justkant: Not sure if it's better but I would call the `.toNumber()` method of BigNumber.js
```suggestion
      ? lastPendingOp.transactionSequenceNumber.plus(1).toNumber()
```

libs/coin-modules/coin-multiversx/src/api/sdk.ts
```diff
@@ -225,7 +225,9 @@ function transactionToEGLDOperation(
     senders: (type === "OUT" || type === "IN") && transaction.sender ? [transaction.sender] : [],
     recipients:
       (type === "OUT" || type === "IN") && transaction.receiver ? [transaction.receiver] : [],
-    transactionSequenceNumber: isSender(transaction, addr) ? transaction.nonce : undefined,
+    transactionSequenceNumber: isSender(transaction, addr)
+      ? new BigNumber((transaction.nonce ?? 0).toString())
```
Justkant: You can just use the nonce directly
```suggestion
      ? new BigNumber(transaction.nonce)
```

libs/coin-modules/coin-solana/src/signOperation.ts
```diff
@@ -42,9 +42,9 @@ const buildOptimisticOperation = (account: Account, transaction: Transaction): S
   const lastOpSeqNumber =
     account.pendingOperations[0]?.transactionSequenceNumber ??
     account.operations[0]?.transactionSequenceNumber ??
-    0;
+    BigNumber(0);
```
Justkant: not a big deal
```suggestion
    new BigNumber(0);
```

libs/coin-modules/coin-stacks/src/bridge/utils/misc.ts
```diff
@@ -94,7 +96,7 @@ export const mapPendingTxToOps =
       accountId: accountID,
       senders: [sender_address],
       recipients: [token_transfer.recipient_address],
-      transactionSequenceNumber: nonce,
+      transactionSequenceNumber: new BigNumber(nonce ?? 0),
```
Justkant: Maybe we shouldn't default to 0 and instead correctly respect the undefined if present and support by the object we create here
```suggestion
      transactionSequenceNumber: nonce ? new BigNumber(nonce) : undefined,
```

libs/coin-modules/coin-stacks/src/bridge/utils/misc.ts
```diff
@@ -165,7 +167,7 @@ export const mapTxToOps =
         fee: feeToUse,
         accountId: accountID,
         senders: [sender_address],
-        transactionSequenceNumber: nonce,
+        transactionSequenceNumber: new BigNumber(nonce ?? 0),
```
Justkant: Same as above

libs/coin-modules/coin-stacks/src/bridge/utils/misc.ts
Justkant: Looking at the type it shouldn't even be possible to be undefined or null so you should be good to use nonce directly
```suggestion
      transactionSequenceNumber: new BigNumber(nonce),
```

libs/coin-modules/coin-stacks/src/bridge/buildOptimisticOperation.ts
```diff
@@ -26,7 +26,7 @@ export const buildOptimisticOperation = (
     blockHash: null,
     blockHeight: null,
     date: new Date(),
-    transactionSequenceNumber: transaction.nonce?.toNumber() || 0,
+    transactionSequenceNumber: new BigNumber(transaction.nonce?.toString() || 0),
```
Justkant: Isn't nonce already a BigNumber ?
```suggestion
    transactionSequenceNumber: transaction.nonce ?? new BigNumber(0),
```

libs/coin-modules/coin-stellar/src/logic/listOperations.ts
```diff
@@ -49,7 +49,7 @@ const convertToCoreOperation = (operation: StellarOperation): Operation => {
       date: operation.date,
     },
     details: {
-      sequence: operation.transactionSequenceNumber,
+      sequence: operation.transactionSequenceNumber?.toString(),
```
Justkant, line 52: Are we sure we don't break anything here going from a number to a string ?

Especially because the typing for details is `Record<string, unknown>` so you won't catch the change were it's used

libs/coin-modules/coin-xrp/src/logic/craftRawTransaction.ts
```diff
@@ -74,7 +74,7 @@ export async function craftRawTransaction(
   if (xrplTransaction.TicketSequence) {
     xrplTransaction.Sequence = 0;
   } else if (!xrplTransaction.Sequence) {
-    xrplTransaction.Sequence = sequence;
+    xrplTransaction.Sequence = sequence.toString();
```
Justkant: Sequence looks to be a number not a string

libs/ledger-live-common/src/bridge/generic-alpaca/alpaca/network/network-alpaca.ts
```diff
@@ -124,8 +124,8 @@ const buildGetBalance = (networkFamily: string) =>
   };
 
 const buildGetSequence = (networkFamily: string) =>
-  async function getSequence(address: string): Promise<number> {
-    const { data } = await network<number, unknown>({
+  async function getSequence(address: string): Promise<bigint> {
+    const { data } = await network<bigint, unknown>({
```
Justkant, line 128: Careful only saying the type is bigint doesn't make the lib (or network function) do the conversion for you

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -114,7 +114,9 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
     senders: extra.parentSenders ?? op.senders,
     recipients: extra.parentRecipients ?? op.recipients,
     date: op.tx.date,
-    transactionSequenceNumber: op.details?.sequence as number,
+    transactionSequenceNumber: op.details?.sequence
+      ? new BigNumber(op.details?.sequence.toString())
```
Justkant, line 126: It should be fine to do this
```suggestion
      ? new BigNumber(op.details?.sequence)
```

libs/ledger-live-common/src/bridge/generic-alpaca/alpaca/network/network-alpaca.ts
qperrot: This part of the code is actually not use No working end point yet

libs/coin-modules/coin-evm/src/logic/estimateFees.ts
qperrot: TransactionLike is from `ethers` library so unfortunately I can not update it

libs/coin-modules/coin-stellar/src/logic/listOperations.ts
qperrot: I added some test and also checked `app.json` and all is good

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
qperrot: `op.details?.sequence` is of type `unknow` because `details?: Record<string, unknown>;`
that is why I am doing `toString()`

libs/coin-framework/src/serialization/operation.ts
qperrot: I add 
```
res.transactionSequenceNumber = transactionSequenceNumberBN.isNaN()
      ? undefined
      : transactionSequenceNumberBN;
      ```

File: libs/coin-framework/src/serialization/operation.ts [hunk 18]
```diff
@@ -168,7 +168,10 @@ export const fromOperationRaw = (
   };
 
   if (transactionSequenceNumber !== undefined) {
-    res.transactionSequenceNumber = transactionSequenceNumber;
+    const transactionSequenceNumberBN = new BigNumber(transactionSequenceNumber);
+    res.transactionSequenceNumber = transactionSequenceNumberBN.isNaN()
+      ? undefined
+      : transactionSequenceNumberBN;
```
francois-guerin-ledger: Do we really want an actual `undefined` instead of simply not setting the key ? We have this snippet to avoid `undefined`
```ts
if (transactionSequenceNumber !== undefined) {
    res.transactionSequenceNumber = transactionSequenceNumber;
}
```
```suggestion
    if (!transactionSequenceNumberBN.isNaN()) {
      res.transactionSequenceNumber = transactionSequenceNumberBN;
    }
```

qperrot: done

Justkant (approved):

jnicoulaud-ledger (approved):

libs/ledgerjs/packages/types-live/src/operation.ts
```diff
@@ -130,7 +130,7 @@ export type OperationRaw<ExtraRaw = OperationExtraRaw> = {
   recipients: string[];
   blockHeight: number | null | undefined;
   blockHash: string | null | undefined;
-  transactionSequenceNumber?: number;
+  transactionSequenceNumber?: string;
```
LucasWerey, line 133: [ASK] I don't have the full context but shouldn't it be BigNumber (like the one line 92) ?

qperrot: No if you look fee line 128 is string but line 82 is BigNumber
this is because BigNumber is not convertible when you do `JSON.stringify(toOperationRaw(obj))),`

LucasWerey (approved):

PR #12431, opened 2025-10-23
chore: add ethers on pnpm catalog to dedup 3 similar versions

gre-ledger:
### ✅ Checklist

  - evm and filecoin implementations

### 📝 Description

In our bundles, we see 3 times the ethers library, each taking 1.2MB –in dev mode at least–, and they are very similar in versions (one minor apart each), so it doesn't cost a lot to try to make them converge to the highest one using **pnpm catalog** and see what it means on the bundle size.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-21989

---


package.json
```diff
@@ -251,7 +251,8 @@
       "@ethersproject/providers>ws": "7.5.10",
       "elliptic": "6.6.1",
       "use-latest-callback": "0.1.9",
-      "@sentry/react-native>metro": "0.81.5"
+      "@sentry/react-native>metro": "0.81.5",
+      "ethers": "6.15.0"
```
francois-guerin-ledger, line 256: Naive question: why no using `catalog:` here ?

gre-ledger: it's a good point, i actually wonder if we need it at the top level package.json at all!

gre-ledger: ok, actually after trying to remove, it will starts to dup again.
probably because       `"@ethersproject/providers>ws": "7.5.10",` above

 i'll try to do catalog: now

PR #12440, opened 2025-10-23
[LIVE-13366][LLM][Android][iOS][BLE] New BLE Re-Pairing screen

jiyuzhuang:
### ✅ Checklist

  - New screen to tell users to forget their devices on the phone.

### 📝 Description
- Add new drawer to tell users to forget device on their phone then pairing;
- There are two use cases:
  - One is trying to connect a know device in the device list of My Ledger;
  - The other is trying to connect a discovered device in the scanning screen.
- The two use cases are not totally using the same component;
- When trying to connect in My Ledger device list, there will be a information banner which is rendered by parent component, to hide this and have exactly the same behaviour with the other use case, the banner will be hidden when the special error is detected.
- The old `BleDevicePeerRemoved` component is still in the code base and used by pairing without DMK, this is intended to keep a backup and will be removed in the future. As the DMK transports are now activated by default, so this won't impact the final result.
- There are 2 commits in this PR, the first one only contains code improvement and UX adjustment, the second one is the main commit of this work.

**Demo on Google Pixel 8 with Android 16**

https://github.com/user-attachments/assets/33a8dbc2-4404-4af1-a4bb-5fe73faa4db7

### ❓ Context

- **JIRA or GitHub link**: [LIVE-13366]

---

[LIVE-13366]: https://ledgerhq.atlassian.net/browse/LIVE-13366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jiyuzhuang:
@jiyuzhuang

[Generating screenshots: workflow started](https://github.com/LedgerHQ/ledger-live/actions/runs/18773821640)

CremaFR (dismissed):

mcayuelas-ledger (dismissed):

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/BleForgetDeviceIllustration.tsx
```diff
@@ -0,0 +1,79 @@
...
+    marginBottom: 16,
+    textAlign: "center",
+  },
+  description: {
+    color: "#6a6a6a",
+    fontWeight: "500",
+    fontSize: 14,
+    fontStyle: "normal",
+    marginHorizontal: 20,
+  },
```
OlivierFreyssinet, line 83: Should use the design system here (especially for stuff like the text color)

apps/ledger-live-mobile/src/locales/en/common.json
```diff
@@ -4587,7 +4593,6 @@
     }
   },
   "manager": {
-    "tabTitle": "My Ledger",
```
OlivierFreyssinet, line 4592: did you double check it's not used?

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/BleForgetDeviceIllustration.tsx
jiyuzhuang: There is no `<Image />` in `native-ui`, the `<Text variant='h4' />` and `<Text variant='body' />` seems like the most relavant to the Figma, but still needs some adjustments. Our DS is not good enough to use and also their components have overridden some native properties which should not be done.

apps/ledger-live-mobile/src/locales/en/common.json
jiyuzhuang: All checked.

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
```diff
@@ -149,10 +150,13 @@ export default function DeviceAction<R, H extends Status, P>({
   request,
   device: selectedDevice,
   location,
+  setShowInfo,
   ...props
 }: Omit<Props<H, P>, "status"> & {
   action: Action<R, H, P>;
   request: R;
+} & {
+  setShowInfo?: (show: boolean) => void;
```
OlivierFreyssinet: This naming is way too vague, just looking at this file I have no idea what it's for

jiyuzhuang: There is a banner at the bottom of this page, which is display when the device is connected using BLE, but in the Figma of Forget Device drawer, there is not. In the description, there is an explanation about this.
The older name is `showAlert` which is even worse because this is not an alert, and there is an information icon on it.

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/BleForgetDeviceIllustration.tsx
OlivierFreyssinet: ok but at least the color should definitely not be hardcoded

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
OlivierFreyssinet: So several thing, "this page" does not apply because this is a component DeviceAction that can be used in a bunch of places.

The explanation should be in the code, and also in the naming, or at least in TSDoc.

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
```diff
@@ -545,6 +553,13 @@ export function DeviceActionDefaultRendering<R, H extends Status, P>({
       });
     }
 
+    if (
+      error instanceof PeerRemovedPairing ||
+      (error.name === "BleError" && error.message === "Operation was cancelled")
```
OlivierFreyssinet: this matching by message is not necessary since you made the changes in RN BLE, is it?

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
OlivierFreyssinet: I think a naming like `onPeerRemovedPairingError` is what we should go with here as it leaves the responsibility to the consumer of the component to do what it wants with it.

jiyuzhuang: I really don't understand why all the device actions have been put in the same place and cannot even think out one single advantage by doing this. The banner has been added in upper component which make it difficult to control by use case. I cannot change all the component DeviceAction to achieve this error drawer just for this case. So the solution is to adapt the existing one.

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
jiyuzhuang: Yes, you're right, but I have no confidence about every time the error can be emitted the right way on Android.

OlivierFreyssinet: if you don't have confidence in that then the matching should be done that way in RN BLE, here in LL we should assume that RN BLE does things properly, otherwise it's chaos if we do a patch here and there haha

apps/ledger-live-mobile/src/components/DeviceAction/index.tsx
OlivierFreyssinet: Yes yes but what I mean is that if you add a prop `onPeerRemovedPairingError`, which is the _only_  reason why this function would be called, at least it's crystal clear what is the purpose of the prop. If the name is `setShowInfo` it's extremely blurry.

jiyuzhuang: I think I've got your point. The children just informe the parent to do something in specific case, the specific case is `onError` and there is already a callback for this. When an error happens, check it and do hide if necessary (`instanceof PeerRemovedPairing`). No need to update the children component, or even make it simpler.

jiyuzhuang: And about the passing setting function, it's not criticised, [example](https://react.dev/learn/thinking-in-react#step-5-add-inverse-data-flow). Personally, I don't like to add specific props for only one use case in a general component, this will make the props list bigger and bigger. If the `onThis` and `onThat` must be done in the same component in each specific case, the component should be refactored.

jiyuzhuang:
> Good, just 3 things:
> 
> * `setShowInfo` prop to change into a `onBlePeerRemovedPairingError` prop
> * color should not be hardcoded, use a color from the DS
> * matching the pairing error by name should not be necessary anymore since the remapping has been done in RN BLE

1. Found a better solution based on this code base;
2. Ok, the colours seem like the single thing we can relay on the actual DS;
3. It's not necessary, just want to prevent some non previewed BLE pairing issues from showing a general error page. It's removed.

apps/ledger-live-mobile/src/components/DeviceActionModal.tsx
```diff
@@ -79,11 +93,11 @@ export default function DeviceActionModal<Req, Stt, Res>({
         ? null
         : device && (
             <Flex alignItems="center">
-              <DeviceActionContainer marginBottom={showAlert ? "16px" : 0}>
+              <DeviceActionContainer marginBottom={showInfo ? "16px" : 0}>
                 <DeviceAction
                   action={action as unknown as Action<Req, PartialNullable<Stt>, Res>}
                   device={device}
-                  onError={onError}
+                  onError={onDeviceActionError}
```
OlivierFreyssinet, line 100: 👌

ofreyssinet-ledger (approved):
thanks for the changes!

CremaFR (dismissed):

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/assets/forget-device-illustration.png
LucasWerey, line 1: Can you use a WebP image, please 🙏 
You can follow this guide if needed https://developers.google.com/speed/webp

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/hooks/useForgetDeviceCta.ts
```diff
@@ -0,0 +1,40 @@
+import { useCallback } from "react";
+import { Linking } from "react-native";
+import { track } from "~/analytics";
+
+const HOW_TO_FORGET_DEVICE_URL =
+  "https://support.ledger.com/fr/article/How-to-remove-a-Ledger-device-from-your-phone-s-Bluetooth-settings";
```
LucasWerey: We should use urls file instead

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/hooks/useForgetDeviceCta.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Hook to handle navigation actions in forgetting BLE device drawer.
+ */
+export function useForgetDeviceCta({ onRetry }: UseForgetDeviceCtaProps) {
+  const onLearnHowToFixPress = useCallback(() => {
+    track("button_clicked", {
+      button: BleForgetDeviceDrawerCta.LearnHowToFix,
+      source: "MyLedger",
+      page: "unpair_required",
+    });
+    Linking.openURL(HOW_TO_FORGET_DEVICE_URL);
```
LucasWerey: You can use useLocalizedUrl for that

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/BleDevicePeerRemoved.tsx
```diff
@@ -1,40 +1,38 @@
-import { Button, Flex, IconsLegacy } from "@ledgerhq/native-ui";
-import { BleError } from "@ledgerhq/live-common/ble/types";
-import { Trans } from "react-i18next";
 import React from "react";
+import { useTranslation } from "react-i18next";
+import { Button, Flex } from "@ledgerhq/native-ui";
+import { BleError } from "@ledgerhq/live-common/ble/types";
 import GenericErrorView from "~/components/GenericErrorView";
+import ExternalLink from "~/components/ExternalLink";
```
LucasWerey, line 6: Can we use Link component of native-ui please 🙏

apps/ledger-live-mobile/src/components/SelectDevice2/index.tsx
```diff
@@ -384,7 +384,7 @@ export default function SelectDevice({
                   px={16}
                 >
                   <Text variant="h5" fontWeight="semiBold">
-                    <Trans i18nKey="manager.selectDevice.title" />
+                    {t(`manager.selectDevice.title`)}
```
LucasWerey:
```suggestion
                    {t("manager.selectDevice.title")}
```

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/hooks/useForgetDeviceCta.ts
jiyuzhuang: There is no localised url for this.

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/BleDevicePeerRemoved.tsx
jiyuzhuang: This one is the `<Link />` component with the icon that is asked in Figma.

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/assets/forget-device-illustration.png
jiyuzhuang: Sorry, there is no asset about this, what is offered in Figma is the `.png` file.

LucasWerey: I know, but it would be better to use WebP as they are more compressed and have the same quality (I changed all the PNGs for WebP a few months ago 🦀 )

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/BleDevicePairingContent/hooks/useForgetDeviceCta.ts
LucasWerey: but there is the /fr in the hardcoded url

jiyuzhuang: Yes, I've noticed that, but this site is in English and I'm not sure if there is a French site about this or not.

LucasWerey: But if in the future there is one we'll need to do a fix in LLM instead of just doing it now I think

jiyuzhuang: Moved

apps/ledger-live-mobile/src/components/BleDevicePairingFlow/assets/forget-device-illustration.png
jiyuzhuang: I'm reading the article, but I don't think for an image like this, this will save more space for us. Btw, I'm not sure the `.webp` format image will be good at all platform and `.png` is the most widely used one for now.

jiyuzhuang: @LucasWerey If this change is intended, can you please inform and make agreement with design team that for the next time, the assets should be in format of `.webp` instead of `.png` or whatever else ?

ofreyssinet-ledger (approved):

PR #12443, opened 2025-10-23
fix(LLD): shallow equal device to fix e2e speculos crash

ejaycoleman:
### ✅ Checklist

  - ...

### 📝 Description
Fixes Speculos crash on E2E adding ADA account by stabalising device

https://github.com/user-attachments/assets/4866a27a-011c-4fd7-bd21-9e7c8632c6cc

### ❓ Context

- **JIRA or GitHub link**: [LIVE-22475](https://ledgerhq.atlassian.net/browse/LIVE-22475)

---

[LIVE-22475]: https://ledgerhq.atlassian.net/browse/LIVE-22475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
@@ -1,7 +1,7 @@
 import React, { Component, useEffect } from "react";
 import BigNumber from "bignumber.js";
 import { Trans, useTranslation } from "react-i18next";
-import { shallowEqual, useDispatch, useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import { Action } from "@ledgerhq/live-common/hw/actions/types";
 import {
   EConnResetError,
@@ -740,7 +740,7 @@ export default function DeviceAction<R, H extends States, P>({
   action: Action<R, H, P>;
   request: R;
 }): JSX.Element {
-  const device = useSelector(getCurrentDevice, shallowEqual);
+  const device = useSelector(getCurrentDevice);
   const hookState = action.useHook(device, request);
   const payload = action.mapResult(hookState);
   useKeepScreenAwake(true);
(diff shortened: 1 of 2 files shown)


ofreyssinet-ledger (dismissed):

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
```diff
@@ -740,7 +740,7 @@ export default function DeviceAction<R, H extends States, P>({
   action: Action<R, H, P>;
   request: R;
 }): JSX.Element {
-  const device = useSelector(getCurrentDevice);
+  const device = useSelector(getCurrentDevice, shallowEqual);
```
OlivierFreyssinet: Can you explain the full reasoning for using `shallowEqual` here?
What's the reason for the reference of the in-store device to change ?
IMO it would be better to fix the issue at the source (what part of the code is updating the store with a different device object). So look at all redux dispatches and try to identify what's wrong.
This change impacts all device actions across all transports, and that needs to be properly assessed, and I think that making a fix specific to Speculos here might be a bit overkill, and it could have unwanted side effects.

ejaycoleman: Without spending significantly more time on this to specifically determine the difference between Cardano and Bitcoin, I believe Cardano has more frequent state changes which trigger getCurrentDevice updates (due to its reliance on the entire state object). I've committed a change to make getCurrentDevice only respond to relevant state change (replacing shallowEqual), which should be safer than shallowEqual since it's more targeted. This was a problem specifically due to one of the envConditions as the object is being created within the getCurrentDevice function, creating new references on each state change. 

If you need a more specific reason then I'll need to spend more time investigating.

ofreyssinet-ledger (approved):

apps/ledger-live-desktop/src/renderer/components/DeviceAction/index.tsx
OlivierFreyssinet: Ok thanks, seems better like this 👍

PR #12551, opened 2025-10-29
refactor: icp stake banner improvement

acewf:
@lawRathod just making this PR to illustrate potentially code refactor we could have on the StakeBanner component


lawRathod:
Thanks a lot for this. I think it's great improvement.

PR #12562, opened 2025-10-30
[LIVE-22504][DMK][LLM] RefusedByUserDAError when user refuses secure connection before an open app

jiyuzhuang:
### ✅ Checklist

  - Now `RefusedByUserDAError` and `DeviceLockedError` can behave as expected.

### 📝 Description

In these two videos, I initialised a receive flow then:
1. Refuse on the device;
2. Lock device.

The `instanceof` keyword cannot identify the errors emitted by DMK in this case, so checking the `_tag` property instead.

**Before:**

https://github.com/user-attachments/assets/b8fa815a-e717-43aa-b5ed-e8945d7903d8

**After:**

https://github.com/user-attachments/assets/f7e9a4f4-1d14-42c3-9eb9-e63414628ceb

### ❓ Context

- **JIRA or GitHub link**: [LIVE-22504]

---

[LIVE-22504]: https://ledgerhq.atlassian.net/browse/LIVE-22504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/hw/connectAppEventMapper.ts
+++ b/libs/ledger-live-common/src/hw/connectAppEventMapper.ts
@@ -12,11 +12,9 @@ import type {
 import {
   DeviceActionStatus,
   DeviceDisconnectedWhileSendingError,
-  DeviceLockedError,
   DeviceSessionStateType,
   UserInteractionRequired,
   OutOfMemoryDAError,
-  SecureChannelError,
   UnsupportedFirmwareDAError,
 } from "@ledgerhq/device-management-kit";
 import type {
@@ -254,10 +252,10 @@ export class ConnectAppEventMapper {
             deviceState.firmwareUpdateContext!.currentFirmware.version,
         }),
       );
-    } else if (error instanceof DeviceLockedError) {
+    } else if ("_tag" in error && error._tag === "DeviceLockedError") {
       this.eventSubject.next({ type: "lockedDevice" });
       this.eventSubject.complete();
-    } else if (error instanceof SecureChannelError) {
+    } else if ("_tag" in error && error._tag === "RefusedByUserDAError") {
       this.eventSubject.error(new UserRefusedAllowManager());
     } else if (error instanceof DeviceDisconnectedWhileSendingError) {
       this.eventSubject.next({ type: "disconnected", expected: false });
--- a/.changeset/poor-spoons-do.md
+++ b/.changeset/poor-spoons-do.md
@@ -1,0 +1,6 @@
+---
+"@ledgerhq/live-common": minor
+"@ledgerhq/live-dmk-mobile": minor
+---
+
+Fix an error recognition bug in ConnectApp DA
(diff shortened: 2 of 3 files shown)


libs/ledger-live-common/src/hw/connectAppEventMapper.ts
```diff
@@ -254,10 +252,10 @@ export class ConnectAppEventMapper {
             deviceState.firmwareUpdateContext!.currentFirmware.version,
         }),
       );
-    } else if (error instanceof DeviceLockedError) {
+    } else if ("_tag" in error && error._tag === "DeviceLockedError") {
       this.eventSubject.next({ type: "lockedDevice" });
       this.eventSubject.complete();
-    } else if (error instanceof SecureChannelError) {
```
OlivierFreyssinet, line 260: [ASK] so you no longer remap `SecureChannelError` ?

libs/ledger-live-common/src/hw/connectAppEventMapper.ts
```diff
@@ -254,10 +252,10 @@ export class ConnectAppEventMapper {
...
+    } else if ("_tag" in error && error._tag === "RefusedByUserDAError") {
```
OlivierFreyssinet, line 258: [COMMENT] Maybe on DMK side we should rename this error to show that it's specific to secure channel interactions, as "refused by user" could apply to a lot of interactions (open app, sign etc.).
Just to make it clearer.

libs/ledger-live-common/src/hw/connectAppEventMapper.ts
jiyuzhuang: I think that only the `RefusedByUserDAError` should be remapped to `UserRefusedAllowManager` in LW where the dedicated drawer shows the relative information. For other cases in `SecureChannelError`, we don't have the drawers in LW, so treat them using default (showing info from error). The previous version tried to map all `SecureChannelError` to `UserRefusedAllowManager` is not correct, but we never enter this condition, so no bug reported.

libs/ledger-live-common/src/hw/connectAppEventMapper.ts
jiyuzhuang: Yes, totally agree. 
I've checked in DMK, the `RefusedByUserDAError` is dedicated to the error of refusing secure connection, other refusing cases will emit `ActionRefusedError`.
I add this renaming task in the ticket issued by this one. Thanks !

OlivierFreyssinet: Thank you !

libs/ledger-live-common/src/hw/connectAppEventMapper.ts
OlivierFreyssinet: Ok 👍

PR #12512, opened 2025-10-28
feat(cryptoassets): make legacy token initialization explicit

gre-ledger:
#### ✅ Checklist
- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.**
  - [x] E2E tests to run https://github.com/LedgerHQ/ledger-live/actions/runs/18872375736
  - [x] integration tests to run https://github.com/LedgerHQ/ledger-live/actions/runs/18872372459
- [x] **Impact of the changes:**
  - Removes automatic `initializeLegacyTokens(addTokens)` from `@ledgerhq/cryptoassets/tokens`
  - Adds explicit initialization of tokens

#### 📝 Description
Makes legacy token initialization explicit instead of automatic.

**Why:**
- Prepares for async migration by making store initialization explicit
- Gives applications control over when tokens are loaded

#### ❓ Context
- **GitHub link**: https://github.com/LedgerHQ/ledger-live/pull/11905

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #12572, opened 2025-10-30
CryptoAssetsStore async rework

gre-ledger:
This PR contains all remaining work to deliver CryptoAssetsStore async rework ( 4 PRs were kept separate at some point – to ease code review / make more atomic commits – but it will now be better to just deliver this PR in one time and validate only this PR https://github.com/LedgerHQ/ledger-live/pull/12512 https://github.com/LedgerHQ/ledger-live/pull/12539 https://github.com/LedgerHQ/ledger-live/pull/12540 https://github.com/LedgerHQ/ledger-live/pull/12541 )

### ✅ Checklist

  - [LLD e2e](https://github.com/LedgerHQ/ledger-live/actions/runs/19036325171)
  - [LLM e2e](https://github.com/LedgerHQ/ledger-live/actions/runs/19036330586)
  - [LLD builds](https://github.com/LedgerHQ/ledger-live-build/actions/runs/19036302098)
  - LLM builds : [ios](https://github.com/LedgerHQ/ledger-live-build/actions/runs/19036306048) , [android](https://github.com/LedgerHQ/ledger-live-build/actions/workflows/build-apk.yml)
  - **as a user, there is no changes at all. bundled tokens are still used (NB: this is NOT the Remove CAL)**
  - **Breaking change** for devs: see migration guide below.
  - **Data serialization**: `fromAccountRaw` and `DataModel.decode` now async
  - Features to focus on during manual testing:
    - App initialization and account loading
    - Token operations and display
    - Deep linking
    - Wallet API/DApps
    - Add Account (Modal)
    - Receive Flow

### 📝 Description

**Migration guidelines**

```diff
- const token = findTokenById(id);
+ const token = await getCryptoAssetsStore().findTokenById(id);

- const token = getTokenById(id);
+ const token = await getCryptoAssetsStore().findTokenById(id);
+ if (!token) return null; // or do something else

- const token = findTokenByAddress(address);
- const token = findTokenByAddressInCurrency(address, currency);
+ const token = await getCryptoAssetsStore().findTokenByAddressInCurrency(address, currency);

- const token = findTokenByTicker(ticker);
⚠️  no more usecase identified. Feature dropped.
```

**Problem**: Current token system uses hardcoded static lists, preventing dynamic token loading from APIs.

**Solution**: This first step migrates the crypto assets system to async patterns while maintaining full backwards compatibility:

- **Core API changes**: Made `findTokenById` and `findTokenByAddressInCurrency` asynchronous and the only necessary functions for CryptoAssetsStore to work
  - `getTokenById` is dropped because we need to treat an async failure differently from a "missing token" case.
  - `findTokenByTicker` weren't used anymore
  - `findTokenByAddress` were safely replaced by `findTokenByAddressInCurrency`
  - they are also only exposed through `getCryptoAssetsStore()`
- **Legacy strategy**: Moved all tokens implementations related to bundled data into `legacy/` . make it harder to have the data bundled in our stack (it's only when you start importing from legacy/legacy-data)
- **tests** : a lot of the tests has been adapted due to the async changes of the APIs.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-21488

---


libs/coin-modules/coin-multiversx/src/buildSubAccounts.ts
```diff
@@ -41,27 +42,114 @@ async function buildMultiversXESDTTokenAccount({
   return tokenAccount;
 }
 
+function buildExistingAccountMaps(
+  existingAccount: Account | null | undefined,
+  blacklistedTokenIds: string[],
```
acewf, line 47: question, should the type be more specific?
```
type MultiversxToken = `erd1${string}`
const blacklistedTokenIds: MultiversxToken[],
```
Probably not important to this PR, but mentioned anyway

acewf (approved):

libs/coin-framework/src/account/__snapshots__/index.test.ts.snap
```diff
@@ -1,14 +1,6 @@
...
 
-exports[`accountWithMandatoryTokens ethereum 1`] = `
-[
-  "mock:1:ethereum::|0",
-  "mock:1:ethereum::|1",
-  "mock:1:ethereum::|2",
-  "mock:1:ethereum::|3",
-  "mock:1:ethereum::|4",
-]
-`;
```
Justkant, line 11: This looks like an issue, wdyt ?

I would expect the snapshot to not change as it points to a behaviour change on the tests expectation

gre-ledger: Indeed. good catch.
I'm not understand how it even worked before: i see that `accountWithMandatoryTokens` do `t.parentCurrency === account.currency` which can be dangerous to assume currency refs are preserved (esp when currency will soon be dynamic – (they already are: https://github.com/LedgerHQ/ledger-live/pull/12277 ) ). in context of the current test, we use a `./__fixtures__/ethereum-erc20-0x_project.json` json fixture which contains `parentCurrency` being a new object which is why the test fails and ref equality can never work.
indeed the snapshot was wrongly regenerated.

I would think a proper fix is to do `t.parentCurrency.id === account.currency.id` wdyt?

libs/coin-modules/coin-multiversx/src/buildSubAccounts.ts
gre-ledger: mmh I'm not sure because `blacklistedTokenIds` is a global parameter passed to all coins in order for the user to filter out spam tokens, it's an array of ids. at the moment Token Id are represented as string (even tho this could be improved in future, i tried to do this for just CryptoCurrencyId and impact is already very important: https://github.com/LedgerHQ/ledger-live/pull/12354 but if we do this for both Crypto and Token Ids, this could be a type improvement )

apps/ledger-live-desktop/src/config/bridge-setup.ts
```diff
@@ -1,7 +1,16 @@
...
+import { addTokens } from "@ledgerhq/cryptoassets/legacy/legacy-utils";
+import { setup } from "@ledgerhq/live-common/bridge/impl";
+import { legacyCryptoAssetsStore } from "@ledgerhq/cryptoassets/legacy/legacy-store";
+
+initializeLegacyTokens(addTokens);
 
 export const cryptoAssetsHooks = createCryptoAssetsHooks({});
 
-setCryptoAssetsStoreForCoinFramework(getCryptoAssetsStore());
+export function setupCryptoAssetsStore(_store: Store) {
```
mcayuelas-ledger, line 13: Why do we keep unused var? :)

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking/useDeepLinkHandler.ts
```diff
@@ -273,55 +273,59 @@ export function useDeepLinkHandler() {
...
 
-          if (!currency || !foundCurrency) {
-            // we fallback to default add account flow with asset selection
-            openAssetFlow();
-            return;
-          }
-          const found = getAccountsOrSubAccountsByCurrency(foundCurrency, accounts || []);
+            if (!currency || !foundCurrency) {
+              // we fallback to default add account flow with asset selection
+              openAssetFlow(true);
```
mcayuelas-ledger: you should not add `true` here, it's for drawerConfig and should be an object

apps/ledger-live-mobile/ios/ledgerlivemobile.xcodeproj/project.pbxproj
mcayuelas-ledger, line 1: You can drop this

apps/ledger-live-mobile/src/components/FabActions/ReadOnlyFabActions.tsx
```diff
@@ -34,6 +34,8 @@ function ReadOnlyFabActions() {
   const pressBuy = useCallback(() => handleOnPress("+ Buy"), [handleOnPress]);
   const pressReceive = useCallback(() => handleOnPress("Receive"), [handleOnPress]);
 
+  if (!currency) return null; // loading
```
mcayuelas-ledger, line 37: what happens if `catch` error?

apps/ledger-live-mobile/src/config/bridge-setup.ts
```diff
@@ -1,7 +1,17 @@
...
+import { setup } from "@ledgerhq/live-common/bridge/impl";
+import { legacyCryptoAssetsStore } from "@ledgerhq/cryptoassets/legacy/legacy-store";
+import { StoreType } from "../context/store";
+
+initializeLegacyTokens(addTokens);
 
 export const cryptoAssetsHooks = createCryptoAssetsHooks({});
 
-setCryptoAssetsStoreForCoinFramework(getCryptoAssetsStore());
+export function setupCryptoAssetsStore(_store: StoreType) {
```
mcayuelas-ledger, line 14: same here for `_store`

apps/ledger-live-mobile/src/newArch/features/AssetSelection/screens/SelectCrypto/useAssets.ts
mcayuelas-ledger, line 1: [NOTE] This file will be remove one Receive is on MAD, we have to  keep it in mind :)

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking/useDeepLinkHandler.ts
mcayuelas-ledger: It will break the flow ^^

libs/ledger-live-common/src/deposit/helper.ts
```diff
@@ -135,11 +131,11 @@ export const searchByNameOrTicker = (list: MappedAsset[], nameOrTicker: string)
       elem.ticker.toLowerCase().includes(nameOrTicker.toLowerCase()),
   );
 
-export const getTokenOrCryptoCurrencyById = (id: string): CryptoOrTokenCurrency => {
+export const getTokenOrCryptoCurrencyById = async (id: string): Promise<CryptoOrTokenCurrency> => {
```
mcayuelas-ledger, line 134: We need to test Receive flow, see if we don't send lot of network request

apps/ledger-live-desktop/src/config/bridge-setup.ts
gre-ledger: @mcayuelas-ledger this is to prepare the actual implementation so we minimize the changes of the next PR that will start needing it https://github.com/LedgerHQ/ledger-live/blob/feat/live-21490-without-persistence/apps/ledger-live-desktop/src/config/bridge-setup.ts#L11

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking/useDeepLinkHandler.ts
gre-ledger: ok it must be a bad conflict resolution that happened at some point. i'll rework.

apps/ledger-live-mobile/src/components/FabActions/ReadOnlyFabActions.tsx
gre-ledger: it would log it but there is no error handling at the moment. ( https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/helpers/useCurrency.tsx#L15 ) room for improvement on this but beyond this PR I think, we should probably create a task?

apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking/useDeepLinkHandler.ts
```diff
@@ -273,55 +273,59 @@ export function useDeepLinkHandler() {
...
-            foundCurrency = getTokenOrCryptoCurrencyById(
-              typeof currency === "string" ? currency : "",
-            );
-          } catch (error) {
-            foundCurrency = null;
-          }
+          async function handleSendDeepLink() {
+            let foundCurrency;
+            try {
+              foundCurrency = await getTokenOrCryptoCurrencyById(
```
gre-ledger: self note: use the CryptoAssetsStore directly instead because we shall take this opportunity to drop the usage of deposit

libs/ledger-live-common/src/deposit/helper.ts
gre-ledger: yes, at the moment the async is actually synchronously resolving the local token data. but we will have to drop this before enabling networking for sure.

apps/ledger-live-mobile/src/config/bridge-setup.ts
gre-ledger: same :)  ( this is to prepare the actual implementation so we minimize the changes of the next PR that will start needing it https://github.com/LedgerHQ/ledger-live/blob/feat/live-21490-without-persistence/apps/ledger-live-mobile/src/config/bridge-setup.ts )

libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -376,35 +372,42 @@ export function useWalletAPIServer({
...
-            }
-            return prev;
-          }, []);
+          (async () => {
+            try {
+              // handle no curencies selected case
+              const currencyList: CryptoOrTokenCurrency[] = [];
+              for (const { id } of currencies) {
+                const currency =
+                  findCryptoCurrencyById(id) || (await getCryptoAssetsStore().findTokenById(id));
```
gre-ledger, line 381: ⚠️  TBD how we can avoid this, we don't want to go through this for each entry. 
it's fine for the async rework, but a possible blocker for the time we switch to networking.
will create JIRA task to track this.

cc @Justkant

libs/coin-framework/src/account/__snapshots__/index.test.ts.snap
```diff
@@ -2,11 +2,7 @@
 
 exports[`accountWithMandatoryTokens ethereum 1`] = `
 [
-  "mock:1:ethereum::|0",
-  "mock:1:ethereum::|1",
-  "mock:1:ethereum::|2",
-  "mock:1:ethereum::|3",
-  "mock:1:ethereum::|4",
+  "mock:1:ethereum::+ethereum%2Ferc20%2F0x~!underscore!~project",
```
gre-ledger: @gre-ledger to double check why the changes remain

libs/coin-framework/src/bot/specs.ts
```diff
@@ -78,13 +78,36 @@ export function deviceActionFlow<T extends TransactionCommon>(
...
+                .catch(error => {
+                  console.error(`Async validation failed for step '${currentStepTitle}':`, error);
+                });
+            } else {
+              botTest("deviceAction confirm step '" + stepTitle + "'", () => {
+                expect({
+                  [stepTitle]: stepValueTransform(stepValue),
+                }).toMatchObject({
+                  [stepTitle]: expectedResult.trim(),
+                });
```
Justkant: We could also just await as it works fine on sync functions

```suggestion
            const expectedResult = await expectedValue(arg, acc);

            botTest("deviceAction confirm step '" + stepTitle + "'", () => {
              expect({
                [stepTitle]: stepValueTransform(stepValue),
              }).toMatchObject({
                [stepTitle]: expectedResult.trim(),
              });
```

libs/coin-framework/src/currencies/__snapshots__/formatCurrencyUnit.test.ts.snap
Justkant, line 1: This snapshot diff looks a bit strange too

libs/coin-framework/src/bot/specs.ts
gre-ledger: unfortunately we are not in an async context here, it would be too impacting to do (deviceActionFlow would turn async, and it's essentially a reducer), but i'll try to dedup the logic

libs/coin-framework/src/currencies/__snapshots__/formatCurrencyUnit.test.ts.snap
gre-ledger: seems that this test is failing and requires regen on develop too so let me send a PR against develop ^^

gre-ledger: https://github.com/LedgerHQ/ledger-live/pull/12614

libs/coin-framework/src/bot/specs.ts
Justkant: Correct I didn't catch that

gre-ledger:
Coin Test -bitcoin failure => known problem in develop
Integration Tests => instability of only one test  `  ● js bridge on account Stellar (js:1:stellar:GAS5NQ2VU6LA3QPDSCVBH66IHP2RE52VFCLFQKSGRF7VKMZA2KTLGI3M:sep5) › getTransactionStatus › transaction check spendable balance (selling_liabilities is not zero)
`, like on develop

gre-ledger:
(same bot report as on develop)

PR #12593, opened 2025-10-31
feat: drop deps on listTokens() for LLD Market

mcayuelas-ledger:
### ✅ Checklist

  - ...

## 🎯 Goal

Remove dependency on `listTokens()` and `internalCurrency` in Market.
Use **Dada as the single source of truth** for currency data.

---

## ✨ Summary

Replaces internal currency references with **lazy-fetched Dada data**, fully decoupling the Market from internal currency lists.

---

## 🔧 Key Changes

* **New hook:** `useLazyInternalCurrency`

  * Lazy fetch currency metadata from Dada
  * Handles crypto > token priority & meta currencies
* **Market refactor**

  * `ledgerIds` replace `internalCurrency`
  * `buy/swap/stake` actions async → fetch via Dada
* **Utility updates**
  * `isAvailableOnBuy/Swap/Stake` now use only `ledgerIds`
  * Removed old internalCurrency logic
* **Docs & tests**

  * Updated `dada-client` README

---

## 🐛 Fix

* Fixed Dada endpoint to `/assets`

https://github.com/user-attachments/assets/bfc7162a-15b1-411b-8717-752cf56d731d

### ❓ Context

- **JIRA or GitHub link**:  https://ledgerhq.atlassian.net/browse/LIVE-22804

---


libs/ledger-live-common/src/dada-client/utils/currencySelection.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  if (networks.length === 0) return undefined;
+
+  const mainCurrency = networks.find(c => c.id === metaCurrencyId);
+  if (mainCurrency) return mainCurrency;
+
+  const cryptoCurrency = networks.find(c => c.type === "CryptoCurrency");
+  if (cryptoCurrency) return cryptoCurrency;
+
+  return networks[0];
```
gre-ledger: do you know if there are a lot of `assetsIds` ? because I see we iterate 3 times on the list (indexing, then finding a main, then finding first crypto currency), but I think there is a possibility to only iterate once 🤔 but i imagine we don't often fall in the last cases?
probably disregards my comment for now, but the complexity of this logic is quite high atm and it's hard to understand why we have to handle them all.

gre-ledger (approved):
🆗  overall

libs/ledger-live-common/src/dada-client/utils/currencySelection.ts
mcayuelas-ledger: Yeah I can check but max ids I saw was a list of 10, I'll recheck :) but indeed we can do it in 1 time anyway :)

PR #12604, opened 2025-11-03
fix(BACK-9521): [coin modules][alpaca] add transaction status to operations

jnicoulaud-ledger:
### ✅ Checklist

  - coin modules

### 📝 Description

Adding transaction status to operations. Notes:
 - In several cases `status` was already in details. I leave it for backward compatibility for now
 - In most cases status was already computed for the Ledger Live op, so I didn't add more tests (except XRP which is particular)

### ❓ Context

- https://ledgerhq.atlassian.net/browse/BACK-9521
- https://ledgerhq.atlassian.net/browse/BACK-9867

---


francois-guerin-ledger:
Interesting. We may want to drop `details.status: "failed" : "success"` in core operation then, it is redundant.

francois-guerin-ledger:
Question: in case of failed operation, do we expect to keep the original `value` or should it be 0 since only the fees have been paid ? I think this is the first option, but just wanted to be sure

jnicoulaud-ledger:
@francois-guerin-ledger yes, we expect the same values, it only changes the way we aggregate balances. Regarding the one in `details`, I kept it intentionally to be backward compatible, we can remove it once we have migrated everything to use the `failed` field

MateuszPaloszBD:
In the context of its use within `coin-hedera`, it looks fine.

libs/coin-modules/coin-xrp/src/api/index.integ.test.ts
```diff
@@ -84,6 +84,25 @@ describe("Xrp Api (testnet)", () => {
...
+        minHeight: 200,
+        order: "desc",
+      });
+      expect(operations.length).toBeGreaterThanOrEqual(200);
+
+      // Then
+      // Check if status matches reference
+      for (const operation of operations) {
+        expect(operation.tx.failed).toBe(FAILED_TRANSACTIONS.has(operation.tx.hash));
+      }
```
francois-guerin-ledger, line 104: Taking https://developers.ledger.com/docs/ledger-live/accounts/integration/blockchain/api#listoperations as an example: we should not compare values but rather ensure the `failed` key is always present and is a boolean (we may want the update this documentation as well)

jnicoulaud-ledger: That's the only place where we test the value is actually correct from real blockchain data :thinking:

francois-guerin-ledger: Tests that check our values are correct are bridge integration tests, with their infamous snapshots. This is far from ideal because:
- snapshots are not resilient to some changes on chain (balance update, new operations ...)
- these tests go through coin bridges, and they are quite high in the stack
So eventually, when this new property is [used at the bridge level](https://ledgerhq.atlassian.net/browse/LIVE-23091), its value will be tested.

jonezque (approved):

libs/coin-modules/coin-xrp/src/api/index.integ.test.ts
jnicoulaud-ledger: OK then, I'll submit that change in another PR to not go through another round of validations, this PR needs to be merged, we have a production issue due to this

jnicoulaud-ledger: @francois-guerin-ledger here is the followup PR: https://github.com/LedgerHQ/ledger-live/pull/12704

PR #12608, opened 2025-11-03
feat: erc20 tokens support in coin-hedera LIVE-22366

mdomanski-ext-ledger:
### ✅ Checklist

  - added send/receive of erc20 tokens to coin-hedera
  - upgraded `@hashgraph/sdk` in coin-hedera and removed react-native patches
  - cleaned up the mix of unit/integration tests around coin-hedera utils 

### 📝 Description

This PR adds support for ERC20 tokens to coin-hedera module. Synchronisation process feels a bit complex now, because there are two data sources (thirdweb and mirror node) and this implementation handles situations when thirdweb is out of sync. It has happened during working on this actually - new erc20 transactions were not returned by Thirdweb API for more than 1 day and we had to contact support to get it fixed.

It's also worth to mention that:
- It will be a separate story, but this logic may be hard to integrate with Alpaca. 
- CAL returns ERC20 tokens as HTS now, but it should be patched by the time of release. This will cause integration tests to fail. The implementation was tested locally with `ref: "branch:next"`.

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-21380

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/hedera/StepRecipientCustomAlert.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/hedera/StepRecipientCustomAlert.tsx
@@ -6,6 +6,7 @@ import { urls } from "~/config/urls";
 import Alert from "~/renderer/components/Alert";
 import TranslatedError from "~/renderer/components/TranslatedError";
 import { openURL } from "~/renderer/linking";
+import { useLocalizedUrl } from "~/renderer/hooks/useLocalizedUrls";
 
 interface AlertProps {
   error: Error;
@@ -38,13 +39,15 @@ const UnverifiedAssociationAlert = ({ error }: AlertProps) => {
 };
 
 const UnverifiedEvmAddressAlert = () => {
+  const evmAddressVerificationUrl = useLocalizedUrl(urls.hedera.evmAddressVerification);
+
   return (
     <Alert type="warning" mt={4} style={{ whiteSpace: "pre-wrap" }}>
       <Trans i18nKey="hedera.send.warnings.evmVerificationRequired.text">
         <Link
           color="inherit"
           textProps={{ fontWeight: "medium" }}
-          onClick={() => openURL(urls.hedera.evmAddressVerification)}
+          onClick={() => openURL(evmAddressVerificationUrl)}
         />
       </Trans>
     </Alert>
--- a/apps/ledger-live-desktop/src/renderer/families/hedera/NoAssociatedAccounts.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/hedera/NoAssociatedAccounts.tsx
@@ -4,15 +4,18 @@ import { openURL } from "~/renderer/linking";
 import { urls } from "~/config/urls";
 import LinkWithExternalIcon from "~/renderer/components/LinkWithExternalIcon";
 import { StepProps } from "~/renderer/modals/AddAccounts/index";
+import { useLocalizedUrl } from "~/renderer/hooks/useLocalizedUrls";
 
 // "no associated accounts" text when adding/importing accounts
 const NoAssociatedAccounts = ({ t }: StepProps) => {
+  const createAccountSupportUrl = useLocalizedUrl(urls.hedera.supportArticleLink);
+
   return (
     <div>
       <Trans i18nKey="hedera.createHederaAccountHelp.text"></Trans>{" "}
       <LinkWithExternalIcon
         fontSize={3}
-        onClick={() => openURL(urls.hedera.supportArticleLink)}
+        onClick={() => openURL(createAccountSupportUrl)}
         label={t("hedera.createHederaAccountHelp.link")}
       />
     </div>
--- a/apps/ledger-live-mobile/src/families/hedera/NoAssociatedAccounts.tsx
+++ b/apps/ledger-live-mobile/src/families/hedera/NoAssociatedAccounts.tsx
@@ -4,13 +4,18 @@ import { urls } from "~/utils/urls";
 import { Flex, Icons, Text, Button } from "@ledgerhq/native-ui";
 import { useTheme } from "styled-components/native";
 import { useTranslation } from "react-i18next";
+import { useLocalizedUrl } from "LLM/hooks/useLocalizedUrls";
 
 // "no associated accounts" text when adding/importing accounts
 function NoAssociatedAccounts() {
   const { colors } = useTheme();
   const { t } = useTranslation();
+  const createAccountSupportUrl = useLocalizedUrl(urls.hedera.supportArticleLink);
 
-  const onPress = useCallback(() => Linking.openURL(urls.hedera.supportArticleLink), []);
+  const onPress = useCallback(() => {
+    Linking.openURL(createAccountSupportUrl);
+  }, [createAccountSupportUrl]);
+
   return (
     <>
       <Flex flex={1} alignSelf="stretch" alignItems="center">
--- a/apps/ledger-live-mobile/src/families/hedera/SendSelectRecipient.tsx
+++ b/apps/ledger-live-mobile/src/families/hedera/SendSelectRecipient.tsx
@@ -4,6 +4,7 @@ import { Linking, StyleSheet } from "react-native";
 import { Text } from "@ledgerhq/native-ui";
 import { sendRecipientCanNext } from "@ledgerhq/live-common/families/hedera/utils";
 import type { TransactionStatus } from "@ledgerhq/live-common/generated/types";
+import { useLocalizedUrl } from "LLM/hooks/useLocalizedUrls";
 import { urls } from "~/utils/urls";
 import Alert from "~/components/Alert";
 import TranslatedError from "~/components/TranslatedError";
@@ -37,9 +38,11 @@ const UnverifiedAssociationAlert = ({ error }: AlertProps) => {
 };
 
 const UnverifiedEvmAddressAlert = () => {
+  const evmAddressVerificationUrl = useLocalizedUrl(urls.hedera.evmAddressVerification);
+
   const goToLink = useCallback(() => {
-    Linking.openURL(urls.hedera.evmAddressVerification);
-  }, []);
+    Linking.openURL(evmAddressVerificationUrl);
+  }, [evmAddressVerificationUrl]);
 
   return (
     <Alert type="warning">


package.json
```diff
@@ -246,9 +246,8 @@
       "remove-flow-types-loader>loader-utils": "*",
       "@ledgerhq/devices": "workspace:*",
       "@ledgerhq/hw-transport": "workspace:*",
+      "@hashgraph/sdk>bignumber.js": "9.1.2",
       "tslib": "2.6.2",
-      "@hashgraph/sdk>@hashgraph/cryptography": "1.1.2",
```
mdomanski-ext-ledger, line 251: Running the mobile app locally works fine, but I'm not sure if this is enough to tell if that works as expected with react-native. I'm not sure why this was added and what did it break previously, but the SDK must be upgraded to have ERC20 support, so it's worth to double check :/

libs/coin-modules/coin-hedera/src/network/utils.ts
```diff
@@ -56,3 +66,67 @@ export function parseTransfers(
     recipients,
   };
 }
+
+export async function getERC20BalancesForAccount(
+  evmAccountId: string,
+): Promise<HederaERC20TokenBalance[]> {
+  const currency = getCryptoCurrencyById("hedera");
+  const availableTokens = listTokensForCryptoCurrency(currency);
```
gre-ledger: ❌  this method must no longer be used. It will not function as expected by the end of our current PI as we are about to drop the `legacy/` folder completely and we completely migrated away from all of the synchronous listTokens approaches

gre-ledger (changes requested):
⚠️  PR currently not compatible with current Remove CAL initiative
🙏  also i would advice we manage to get https://github.com/LedgerHQ/ledger-live/pull/12572 in first

libs/coin-modules/coin-hedera/src/network/utils.ts
mdomanski-ext-ledger: Thanks for taking a look @gre-ledger. Unfortunately, to get erc20 balances we need to call the `balanceOf` method of each possible contract (erc20 token address), so this is the main reason the `listTokensForCryptoCurrency` is used there. I think it won't be a problem to switch to async API call - just a question here, will there be any method, similar to `useTokensData` hook, that could be used on the coin-module level?

Another thing worth to mention is probably the `libs/ledgerjs/packages/cryptoassets/src/crypto-assets-importer/importers/hedera/index.ts` - currently the API returns ERC20 tokens as HTS standard. The plan was to patch it on LL side, as this is quicker than upgrading the API, so we can try to target the next release window. I think it also conflicts with the plan of dropping the `legacy/` folder

gre-ledger: I'm not sure to understand why you need the full list of tokens to get the balance of erc20 tokens the user have 🤔 
AFAIK many existing coins manage to solve this by using either `CryptoAssetsStore#findTokenById` / `CryptoAssetsStore#findTokenByAddress` (these functions are becoming asynchronously really soon)

Isn't it the way Hedera works too for existing tokens today? https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-hedera/src/logic/getBalance.ts#L20-L22

at the moment there is intentionally no easy way to "list all tokens" because we are moving away from this paradigm, only at the UI level we have paginated hooks to do so – some exception will exists but it's an overall intent

---

Regarding the crypto-assets-importer, the good news is that by moving completely to consuming the backend API and by dropping the local data, we should remove the necessity to even have the importers and the converters & this script is also going to be removed.
Normally we have managed to no longer need them and to reconciliate all data that comes from backend. you can verify with `pnpm run:cli backtestTokens` if the tokens coming from API are compliant with your local tokens. Some exception remained and we have introduced some edge cases in `libs/ledgerjs/packages/cryptoassets/src/api-token-converter.ts` – but at some point we will fix them all (these are more meant to be existing problem in coin implementations that needs evolution in coins, but normally we should be iso between backend data and client side data)


So regarding
```
// FIXME: remove ERC20 overrides once CAL is fixed (solution is under discussion).
// For now, API returns wrong tokenType and is missing EVM contractAddress for these tokens.
```

I think we will have to have these CAL fix in, because the company priority is to move to remote data first and no longer introduce any extra bundled tokens.

MateuszPaloszBD: I’ll try to address this as best as I can since Mateusz is currently unavailable.
In general, it’s true that for non-ERC20 tokens on Hedera, we’ve been using the approach with the mirror node request and store methods you mentioned. Unfortunately, for ERC20 tokens it’s not that straightforward.

As far as I remember, the Hedera mirror node doesn’t return the ERC20 tokens held by an account – for example:
`https://mainnet.mirrornode.hedera.com/api/v1/accounts/0.0.8835924/tokens`

Because of that, we don’t have direct access to the contract address that we could use to query the balance of a given token.
That’s why we’ve been checking the entire list of tokens, filtering it to get the ERC20 ones, and then using the resulting list of addresses to verify whether the account holds those tokens and how much, by checking their balances.

mateuszpalosz-ext: @gre-ledger A small clarification about the details I discussed with Gabriel : 
At the moment, we’re indeed checking the list of all tokens, but we only check the balance for those marked with the type erc20. So, from the list of all tokens (and after adding another supported token yesterday), we should have only 4 different tokens for which we need to send balance queries.
It’s not the best solution, since it assumes that for every supported ERC20 on Hedera (in our case 3 and soon 4), we actually have to make a balance request for each of them. However, from my earlier conversations with the other Mateusz, there doesn’t seem to be any other way to solve this. We somehow need to have access to the list of those addresses.

@gre-ledger When you find a moment, I’m waiting for your feedback regarding the issue with access to the addresses for ERC20 tokens

gre-ledger:
```
Mobile Bundle Checks
Comparing [2ac0ce4](https://github.com/LedgerHQ/ledger-live/commit/2ac0ce4686e752571445626e294fdf258756e40f) against [b8d8588](https://github.com/LedgerHQ/ledger-live/commit/b8d8588d47461269015efd9b5c249fe037bcef0e).

⚠️ main.ios.jsbundle bundle size significantly increased: 72.8mb -> 73.4mb. Please check if this is expected.
⚠️ main.android.jsbundle bundle size significantly increased: 72.8mb -> 73.4mb. Please check if this is expected.

Desktop Bundle Checks
Comparing [2ac0ce4](https://github.com/LedgerHQ/ledger-live/commit/2ac0ce4686e752571445626e294fdf258756e40f) against [b8d8588](https://github.com/LedgerHQ/ledger-live/commit/b8d8588d47461269015efd9b5c249fe037bcef0e).

⚠️ renderer bundle size significantly increased: 38.9mb -> 39.4mb. Please check if this is expected.
⚠️ @ethersproject/logger library is now duplicated in renderer (regression)
⚠️ @ethersproject/bytes library is now duplicated in renderer (regression)
⚠️ @ethersproject/bignumber library is now duplicated in renderer (regression)
⚠️ @ethersproject/properties library is now duplicated in renderer (regression)
⚠️ @ethersproject/keccak256 library is now duplicated in renderer (regression)
⚠️ @ethersproject/rlp library is now duplicated in renderer (regression)
⚠️ @ethersproject/address library is now duplicated in renderer (regression)
⚠️ @ethersproject/constants library is now duplicated in renderer (regression)
⚠️ @ethersproject/strings library is now duplicated in renderer (regression)
⚠️ @ethersproject/base64 library is now duplicated in renderer (regression)
⚠️ @ethersproject/hash library is now duplicated in renderer (regression)
⚠️ @ethersproject/abi library is now duplicated in renderer (regression)
⚠️ long library is now duplicated in renderer (regression)
```

this is a problem.

mateuszpalosz-ext:
@gre-ledger @mdomanski-ext-ledger comment about bundle size increase: "I had to upgrade the @hashgraph/sdk package and it will be required for staking as well. The version 2.74.0 gives us a possibility to remove patches/@hashgraph__sdk@2.14.2.patch as it should already include "react-native": "./src/native.js", export on the package level. This is the main benefit I would say (of course except making erc20/staking implementation possible). Maybe it would be possible to override the @ethersproject/  deps with version that we already have, but it sounds a bit hacky."

gre-ledger:
Ok, then I think we'll need to look at updating the @ethersproject/ deps on the global project and I see that it's only also used by `Celo` (well also `hw-app-eth` but that lib shall eventually be dropped)

acewf (approved):

package.json
scavassa-yld: @mdomanski-ext-ledger  can we close this comment?

mdomanski-ext-ledger: @scavassa-yld I'm not sure who would be the best person to confirm the reason behind those old overrides in package.json. I left this comment just to highlight that we can't keep them, because we need to upgrade the SDK.

Running the mobile app locally works fine for me on both Android and iOS, so I think we can close this and eventually fix with separate PR.

acewf (approved):

acewf:
@mdomanski-ext-ledger here https://github.com/LedgerHQ/ledger-live/pull/12608#issuecomment-3566943064 the code coverage is under 80% threshold, probably this needs to be added to merge this PR

Wozacosta (dismissed):

acewf (dismissed):

fAnselmi-Ledger (dismissed):

apps/ledger-live-desktop/src/config/urls.ts
```diff
@@ -171,6 +171,7 @@ export const urls = {
     supportArticleLink: "https://support.ledger.com/article/4494505217565-zd",
     tokenAssociation:
       "https://support.ledger.com/article/How-to-manage-Hedera-tokens-with-Ledger-Live",
+    evmAddressVerification: "https://ledger.com/e14",
```
mcayuelas-ledger, line 174: image

The link is wrong here no?

md-blockydevs: @mcayuelas-ledger thanks, I've already reported this, but this is the link we got some time ago and I believe we are just waiting for content update (the same link is used in the device app)

fAnselmi-Ledger (dismissed):

Wozacosta (dismissed):

acewf (dismissed):

apps/ledger-live-desktop/src/renderer/families/hedera/StepRecipientCustomAlert.tsx
```diff
@@ -35,12 +37,26 @@ const UnverifiedAssociationAlert = ({ error }: AlertProps) => {
...
 };
 
+const UnverifiedEvmAddressAlert = () => {
+  return (
+    <Alert type="warning" mt={4} style={{ whiteSpace: "pre-wrap" }}>
+      <Trans i18nKey="hedera.send.warnings.evmVerificationRequired.text">
+        <Link
+          color="inherit"
+          textProps={{ fontWeight: "medium" }}
+          onClick={() => openURL(urls.hedera.evmAddressVerification)}
```
mcayuelas-ledger: you should use `useLocalizedUrl`

mdomanski-ext-ledger: @mcayuelas-ledger That's a good point, I also see more similar errors around coin-hedera that weren't caught earlier, e.g. `NoAssociatedAccounts`.

~Can this be handled in a separate PR to avoid blocking this one?~ we need to rebase anyway, so I will address this now

PR #12610, opened 2025-11-03
feat(signer-solana): Add DMK signer solana behind FF

fAnselmi-Ledger:
### ✅ Checklist

  - ...

### 📝 Description

Added DMK signer into @ledgerhq/live-signer-solana pkg under feature flag

### ❓ Context

- **JIRA or GitHub link**: [DSDK-876](https://ledgerhq.atlassian.net/browse/DSDK-876)

---

[DSDK-876]: https://ledgerhq.atlassian.net/browse/DSDK-876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/large-dolls-pay.md
+++ b/.changeset/large-dolls-pay.md
@@ -3,8 +3,8 @@
 "@ledgerhq/types-live": patch
 "@ledgerhq/coin-bitcoin": patch
 "@ledgerhq/coin-solana": patch
-"ledger-live-desktop": patch
-"live-mobile": patch
+"ledger-live-desktop": minor
+"live-mobile": minor
 "@ledgerhq/live-common": patch
 "@ledgerhq/live-signer-solana": patch
 ---
--- a/libs/ledger-live-common/src/families/solana/setup.ts
+++ b/libs/ledger-live-common/src/families/solana/setup.ts
@@ -16,12 +16,12 @@ import { signMessage } from "@ledgerhq/coin-solana/hw-signMessage";
 import { setShouldSkipTokenLoading } from "@ledgerhq/coin-solana/preload";
 import { LiveConfig } from "@ledgerhq/live-config/LiveConfig";
 import { getCryptoAssetsStore } from "../../bridge/crypto-assets";
-import { LegacySignerSolana, DMKSignerSolana } from "@ledgerhq/live-signer-solana";
+import { LegacySignerSolana, DmkSignerSol } from "@ledgerhq/live-signer-solana";
 import { DeviceManagementKit } from "@ledgerhq/device-management-kit";
 
 let _solanaLdmkFFEnabled: boolean = false;
 
-let _dmkSignerInstance: DMKSignerSolana | null = null;
+let _dmkSignerInstance: DmkSignerSol | null = null;
 
 // temporary solution to dynamically enable/disable the Solana DMK signer,
 // waiting for LIVE-20250 to be implemented
@@ -43,7 +43,7 @@ export function getSolanaSignerInstance(
 ): SolanaSigner {
   if (canDMKSignerBeUsed(transport)) {
     if (!_dmkSignerInstance) {
-      _dmkSignerInstance = new DMKSignerSolana(transport.dmk, transport.sessionId);
+      _dmkSignerInstance = new DmkSignerSol(transport.dmk, transport.sessionId);
     }
     return _dmkSignerInstance;
   } else {
--- a/libs/ledgerjs/packages/errors/src/index.ts
+++ b/libs/ledgerjs/packages/errors/src/index.ts
@@ -51,6 +51,9 @@ export const EnpointConfigError = createCustomErrorClass("EnpointConfig");
 export const EthAppPleaseEnableContractData = createCustomErrorClass(
   "EthAppPleaseEnableContractData",
 );
+export const SolAppPleaseEnableContractData = createCustomErrorClass(
+  "SolAppPleaseEnableContractData",
+);
 export const CeloAppPleaseEnableContractData = createCustomErrorClass(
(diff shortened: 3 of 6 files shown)


libs/live-signer-solana/package.json
```diff
@@ -29,12 +29,17 @@
...
+    "@ledgerhq/device-management-kit": "catalog:",
+    "@ledgerhq/device-signer-kit-solana": "catalog:",
     "@ledgerhq/errors": "workspace:^",
     "@ledgerhq/hw-app-solana": "workspace:^",
     "@ledgerhq/hw-bolos": "workspace:^",
     "@ledgerhq/hw-transport": "workspace:*",
     "@ledgerhq/ledger-cal-service": "workspace:^",
     "@ledgerhq/ledger-trust-service": "workspace:^",
+    "@types/bs58": "^4.0.1",
+    "bs58": "^4.0.1",
```
LucasWerey: we use the same version of bs58 in common maybe we can add them to catalog wdyt?

LucasWerey (dismissed):

libs/live-signer-solana/src/DMKSignerSolana.ts
```diff
@@ -0,0 +1,192 @@
...
+  | GetAppConfigurationDAError
+  | SignMessageDAError
+  | SignTransactionDAError;
+
+/**
+ * DMK-based Solana signer using DMK signer-kit
+ */
+export class DMKSignerSolana implements SolanaSigner {
+  private dmkSigner: ReturnType<typeof SignerSolanaBuilder.prototype.build>;
+  private readonly DMKPubKeyDisplayMode = {
```
jiyuzhuang: If you want to make each property of the object `DMKPubKeyDisplayMode` read-only, this won't work. You can try to reassign a value to `long` and there will not be error.
```suggestion
  private readonly DMKPubKeyDisplayMode: { readonly long: string; readonly short: string } = {
```
or even agressive like:

```suggestion
  private readonly DMKPubKeyDisplayMode = { readonly long: "long"; readonly short: "short" } = {
```

libs/live-signer-solana/src/DMKSignerSolana.ts
```diff
@@ -0,0 +1,192 @@
...
+  | GetAddressDAError
+  | GetAppConfigurationDAError
+  | SignMessageDAError
+  | SignTransactionDAError;
+
+/**
+ * DMK-based Solana signer using DMK signer-kit
+ */
+export class DMKSignerSolana implements SolanaSigner {
+  private dmkSigner: ReturnType<typeof SignerSolanaBuilder.prototype.build>;
```
jiyuzhuang: I saw that there is a type called `SignerSolana` exported in `@ledgerhq/device-signer-kit-solana` which is the return type of the `.build()` method.

jiyuzhuang:
Suggestions:
1. Rename `DMKSignerSolana` to `DmkSignerSolana` to align with the Ethereum signer;
2. Use logs instead of the `console.log` in the code if the tracking is necessary;
3. In the `_mapError()` method, there are three kinds of handling: `6985`, `6808` and others, so it's better to give more information about the `6808` and reorganise the conditions to make it more clear. ~~The actual code may let some types of error hidden even if they may not exist, this should not happen.~~

.changeset/large-dolls-pay.md
```diff
@@ -0,0 +1,12 @@
+---
+"@ledgerhq/hw-app-btc": patch
+"@ledgerhq/types-live": patch
+"@ledgerhq/coin-bitcoin": patch
+"@ledgerhq/coin-solana": patch
+"ledger-live-desktop": patch
+"live-mobile": patch
```
francois-guerin-ledger:
```suggestion
"ledger-live-desktop": minor
"live-mobile": minor
```
I think `patch` on `ledger-live-desktop` and `live-mobile` are reserved for hotfixes

francois-guerin-ledger (commented):
I try to send some [VIRTUAL](https://solscan.io/token/3iQL8BFS2vE7mww4ehAqQHAsbmRNCrPxizWAT2Zfyr9y), and I have `???` instead of the ticker. I don't experiment this with `USDC` for instance (I can still broadcast).

Also does the DMK support ATA (https://ledgerhq.atlassian.net/browse/LIVE-21031) ? I did not test it yet.

fAnselmi-Ledger:
> I try to send some [VIRTUAL](https://solscan.io/token/3iQL8BFS2vE7mww4ehAqQHAsbmRNCrPxizWAT2Zfyr9y), and I have `???` instead of the ticker. I don't experiment this with `USDC` for instance (I can still broadcast).
> 
> Also does the DMK support ATA (https://ledgerhq.atlassian.net/browse/LIVE-21031) ? I did not test it yet.

@francois-guerin-ledger thanks for the review! both of the points you mentioned are valid and should be investigated DMK-side, the purpose of this PR is to add an adapter for DMK signer Solana and keep it off by default (with the possibility to turn it on via FF).

francois-guerin-ledger:
I have investigated a bit, the issue is at
```ts
transport.dmk instanceof DeviceManagementKit &&
typeof transport.sessionId === "string";
```
I let you check why one of those 2 assertions is `false`

jiyuzhuang:
> ```ts
> transport.dmk instanceof DeviceManagementKit
> ```

I think the issue is at the `instanceof DMK`, this will always return false.

francois-guerin-ledger:
> > ```ts
> > transport.dmk instanceof DeviceManagementKit
> > ```
> 
> I think the issue is at the `instanceof DMK`, this will always return false.

Do you know why @jiyuzhuang ?

PR #12667, opened 2025-11-06
Feat/live 22047 llm sync postonboarding

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Removes recover and adds ledger sync to post onboarding hub as an action to take for post onboarding. To enable the entry point in the postOnboarding hub the feature flag `llmLedgerSyncEntryPoints` must have `postOnboarding` param set to true.

https://github.com/user-attachments/assets/d1e29cc8-3fde-4c98-b4c3-46adb4604fb6

### ❓ Context

- **JIRA or GitHub link**: 

---


libs/ledgerjs/packages/types-live/src/postOnboarding.ts
```diff
@@ -33,17 +34,19 @@ export type WithNavigationParams = {
   }) => [screen: any] | [screen: any, navigationParams: any];
 };
 
+export interface StartActionArgs {
+  openModalCallback?: (modalName: any) => void;
```
ejaycoleman: Can't `any` just be a string?

apps/ledger-live-desktop/src/renderer/components/PostOnboardingHub/logic/index.tsx
```diff
@@ -51,12 +56,12 @@ const customImage: PostOnboardingAction = {
...
   description: "customImage.postOnboarding.description",
   actionCompletedPopupLabel: "customImage.postOnboarding.actionCompletedPopupLabel",
-  startAction: ({ deviceModelId }) =>
+  startAction: ({ deviceModelId }: StartActionArgs) =>
     setDrawer(
       CustomImage,
       {
         isFromPostOnboardingEntryPoint: true,
-        deviceModelId,
+        deviceModelId: deviceModelId as DeviceModelId | null,
```
ejaycoleman: you could avoid `as` with `deviceModelId || null`?

apps/ledger-live-mobile/src/analytics/segment.ts
```diff
@@ -321,7 +323,9 @@ const extraProperties = async (store: AppStore) => {
     isReborn,
     onboardingHasDevice,
     // For tracking receive flow events during onboarding
-    ...(isOnboardingFlow ? { flow: "onboarding" } : {}),
+    ...(isOnboardingFlow || isPostOnboardingFlow
+      ? { flow: isPostOnboardingFlow ? "post-onboarding" : "onboarding" }
+      : {}),
```
ejaycoleman: Feels a bit complex, perhaps this is simpler?
```suggestion
   ...(isOnboardingFlow ? { flow: "onboarding" } : {}),
   ...(isPostOnboardingFlow ? { flow: "post-onboarding" } : {}),
```

apps/ledger-live-mobile/src/logic/postOnboarding/actions.ts
```diff
@@ -64,28 +67,20 @@ export const customImageAction: PostOnboardingAction = {
...
+  featureFlagParamId: "postOnboarding",
+  Icon: Icons.Refresh,
+  title: "postOnboarding.actions.syncAccounts.title",
+  titleCompleted: "postOnboarding.actions.syncAccounts.titleCompleted",
+  description: "postOnboarding.actions.syncAccounts.description",
+  actionCompletedPopupLabel: "postOnboarding.actions.syncAccounts.popupLabel",
+  buttonLabelForAnalyticsEvent: "Sync accounts",
+  getIsAlreadyCompleted: async ({ isLedgerSyncActive }) => {
+    return !!isLedgerSyncActive;
+  },
```
ejaycoleman, line 82: I don't think getIsAlreadyCompleted serves much value? It at least doesn't need to be async anymore but can't isLedgerSyncActive just be used directly?

apps/ledger-live-mobile/src/analytics/segment.ts
mitchellv-ledger: I was concerned about having too many places where we add flow field but I agree it does look simpler

apps/ledger-live-mobile/src/logic/postOnboarding/actions.ts
mitchellv-ledger: I can't access it directly due to how it is structured so have added a new field called `getIsAlreadyCompletedByState` which is more defined for this use case.

ejaycoleman: But I mean the fact it's optional, rather than something like 

```
const isAlreadyCompleted = getIsAlreadyCompletedByState?.({ isLedgerSyncActive });
setIsActionCompleted(completed || !!isAlreadyCompleted);
```

couldn't you just 

`setIsActionCompleted(completed || !!isLedgerSyncActive);`

in `apps/ledger-live-mobile/src/components/PostOnboarding/PostOnboardingActionRow.tsx`

mitchellv-ledger: Unfortunately not, it's quite convoluted code because essentially this is just a wrapper that takes the action object and uses it's fields to determine logic and ui, so it needs someway of being passed the state to then check if it is correct. ie this is a generic wrapper for all actions not just for the once that needs to check isLedgerSyncActive meaning if you were to add `setIsActionCompleted(completed || !!isLedgerSyncActive);` it would affect all actions.

ejaycoleman (approved):

PR #12710, opened 2025-11-10
fix(LIVE-22854): Fix Drawer triggering /cancel after successful exchange

ooke-ledger:
### ✅ Checklist

  - ...

### 📝 Description

This PR updates the Drawer component to ensure that the /cancel endpoint is not called if a swap/exchange has already been successfully completed.

| Before        | 

https://github.com/user-attachments/assets/bf2c64fa-a369-4333-86d7-326b09242bcb

| After         |

https://github.com/user-attachments/assets/a75a4f01-3393-4889-94d3-a5c96a778f31

|               |               |

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-22854

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/LiveAppDrawer.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/LiveAppDrawer.tsx
@@ -102,7 +102,6 @@ export const LiveAppDrawer = () => {
   const [exchangeCompleted, setExchangeCompleted] = React.useState(false);
 
   const onCloseExchangeComplete = useCallback(() => {
-    setExchangeCompleted(true);
     dispatch(closePlatformAppDrawer());
   }, [dispatch]);
 


apps/ledger-live-desktop/src/renderer/components/LiveAppDrawer.tsx
```diff
@@ -98,7 +99,10 @@ export const LiveAppDrawer = () => {
     }
   }, [dismissDisclaimerChecked, dispatch, payload]);
 
+  const [exchangeCompleted, setExchangeCompleted] = React.useState(false);
+
   const onCloseExchangeComplete = useCallback(() => {
+    setExchangeCompleted(true);
```
ejaycoleman: Just to double check - is it correct to `setExchangeCompleted(true);` here? Isn't it adequate to only set it in `onResult` like you've done? It looks like `onClose` is called in `onBroadcastSuccess` in the case of a `DisabledTransactionBroadcastError` error in the `CompleteExchange` component.

ooke-ledger: Good catch! You're correct - setting it in onResult is sufficient. The flag in onCloseExchangeComplete was redundant since onResult is always called before the "View Details" button becomes available. I've removed it.

PR #12733, opened 2025-11-12
Live 23171 earn ledger live add wallet api method to get encoding account id information

sergiubreban:
This will enable us to move the ETH staking modal inside the Earn Live app.

### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**:  [LIVE-23171](https://ledgerhq.atlassian.net/browse/LIVE-23171)

---

[LIVE-23171]: https://ledgerhq.atlassian.net/browse/LIVE-23171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/wallet-api/utils/deriveAccountIdForManifest.ts
+++ b/libs/ledger-live-common/src/wallet-api/utils/deriveAccountIdForManifest.ts
@@ -5,7 +5,7 @@ import { LiveAppManifest } from "../../platform/types";
 import semver from "semver";
 
 /**
- * Determines if a manifest uses uuid (Ledger Live) account ID format or encoded (Wallet API) format.
+ * Determines if a manifest uses (Ledger Live) account ID format or UUID (Wallet API) format.
  * @param manifest - The live app manifest to check
  * @returns true if the manifest uses encoded format (v3+ dapp), false otherwise
  */


apps/ledger-live-mobile/src/components/WebPTXPlayer/CustomHandlers.ts
```diff
@@ -182,6 +203,38 @@ export function useCustomExchangeHandlers({
...
+
+        if (!liveAppIds) {
+          throw new Error("Missing liveAppIds parameter");
+        }
+
+        if (!Array.isArray(liveAppIds)) {
+          throw new Error("liveAppIds must be an array");
+        }
+
+        const results: Record<string, "uuid" | "encoded"> = {};
```
beths-ledger: I'd like to extract this `Record<string, "uuid" | "encoded">` into a re-usable type like `AccountIdFormat` and add an explanation that `/** @developer note uuid = wallet API id format, encoded = ledger wallet default format */` -- to avoid confusion. What do you think?

sergiubreban: is there any place you want to reuse the type?

libs/ledger-live-common/src/wallet-api/utils/deriveAccountIdForManifest.ts
```diff
@@ -4,14 +4,26 @@ import { TokenAccount, Account, AccountLike } from "@ledgerhq/types-live";
...
+/**
+ * Determines if a manifest uses uuid (Ledger Live) account ID format or encoded (Wallet API) format.
+ * @param manifest - The live app manifest to check
+ * @returns true if the manifest uses encoded format (v3+ dapp), false otherwise
+ */
+export function usesEncodedAccountIdFormat(manifest: LiveAppManifest): boolean {
+  return (
+    "dapp" in manifest &&
+    !!manifest.apiVersion &&
+    semver.satisfies(WALLET_API_VERSION, manifest.apiVersion)
```
Justkant, line 16: I think only live-apps (no "dapp" in manifest) using the wallet-api are on the encoded format and dApps using the dApp browser v3 ("dapp" in manifest) should be using the accounts from LL directly and I don't think we do the mapping from wallet-api to LL accounts for dApps (I might be missing something here)

```suggestion
    !"dapp" in manifest &&
    !!manifest.apiVersion &&
    semver.satisfies(WALLET_API_VERSION, manifest.apiVersion)
```

Justkant: I think the name of the function is the issue here as it seems correct for the logic below in  `deriveAccountIdForManifest`

libs/ledger-live-common/src/wallet-api/utils/deriveAccountIdForManifest.ts
```diff
@@ -4,14 +4,26 @@ import { TokenAccount, Account, AccountLike } from "@ledgerhq/types-live";
 import { LiveAppManifest } from "../../platform/types";
 import semver from "semver";
 
+/**
+ * Determines if a manifest uses uuid (Ledger Live) account ID format or encoded (Wallet API) format.
```
Justkant: wallet-api encodes the id from LL using UUID

libs/ledger-live-common/src/wallet-api/utils/deriveAccountIdForManifest.ts
sergiubreban: the previous logic was 
```
const isDapp = "dapp" in manifest;
  if (isDapp && manifest.apiVersion && semver.satisfies(WALLET_API_VERSION, manifest.apiVersion)) {
  ```
  
The intent was to reuse the logic, not to change the behaviour.

sergiubreban: I just need a method in wallet api that returns the accountId type for each live-app that I will pass in the arguments. This will be later use here: https://github.com/LedgerHQ/earn-live-app/pull/769/files


Does it make sense?

Justkant: I understand you just reused the logic below but the function is used in the custom handler too and you cannot guarantee when it's not a dApp that it will be using the wallet-api (it could be on the old platform still using LL account ids and not the wallet-api)

I suppose it should be fine in most cases but we still have live-apps using the old platform for some partners

I would say keep it like this but even the logic below was flawed IMO, we could simply return the `accountId` if `"dapp" in manifest` then only use the `walletApiAccountId` for the proper `apiVersion` and fallback to `accountId` for platform (you need to first check for dapp as it uses the same apiVersion as wallet-api)

I get it if you don't want to break anything and keep it like this

PR #12738, opened 2025-11-12
feat: Add close-app tracking event

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

We want to track what page the user is closing the app on. This PR adds a listener on the main window which prevents the close functionality and sends a message to the ipcRenderer thread where we can set off a track event. Once the track event has completed we then send another close event which this time functions as normal.

See below video for a timing of the track event, I have set the timeout to 5ms to give ample time for it to complete:

https://github.com/user-attachments/assets/ea35d9a4-6695-4500-bb84-bcd122230953

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/apps/ledger-live-desktop/src/main/index.ts
+++ b/apps/ledger-live-desktop/src/main/index.ts
@@ -33,6 +33,7 @@ import {
   REACT_DEVELOPER_TOOLS,
 } from "electron-devtools-installer";
 import { setupTransportHandlers, cleanupTransports } from "./transportHandler";
+
 // End import timing, start initialization
 console.timeEnd("T-imports");
 console.time("T-init");
@@ -194,6 +195,23 @@ app.on("ready", async () => {
     typeof applyWindowParams
   >[0];
   await applyWindowParams(windowParams, settings);
+
+  let status: number = 0;
+  window.on("close", e => {
+    if (status === 0) {
+      e.preventDefault();
+      getMainWindowAsync()
+        .then(w => {
+          status = 1;
+          if (w && "send" in w.webContents) {
+            w.webContents.send("app-close");
+          } else {
+            w.close();
+          }
+        })
+        .catch((err: unknown) => console.log(err));
+    }
+  });
 
   // Setup window event handlers
   window.on(
--- a/apps/ledger-live-desktop/src/renderer/init.tsx
+++ b/apps/ledger-live-desktop/src/renderer/init.tsx
@@ -54,9 +54,16 @@ import { importMarketState } from "./actions/market";
 import { fetchWallet } from "./actions/wallet";
 import { fetchTrustchain } from "./actions/trustchain";
 import { registerTransportModules } from "~/renderer/live-common-setup";
+import { track } from "./analytics/segment";
 
 const rootNode = document.getElementById("react-root");
 const TAB_KEY = 9;
+
+function onAppClose() {
+  console.log("app-close message received");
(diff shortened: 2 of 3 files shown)


gre-ledger:
```
  window.addEventListener("beforeunload", () => {
    // This event is triggered when we reload the app, we want it to forget what it knows
    reload();
  });
  ```
  
  do you know if it's actually where we want to add the track? I think this is supposed to be triggered when we close the app. i would prefer we don't have to touch main thread to do this.

**Comment by mitchellv-ledger** [comment]
> ```
>   window.addEventListener("beforeunload", () => {
>     // This event is triggered when we reload the app, we want it to forget what it knows
>     reload();
>   });
> ```
> 
> do you know if it's actually where we want to add the track? I think this is supposed to be triggered when we close the app. i would prefer we don't have to touch main thread to do this.

Yes this is where we want to add the track, when the user closes the app by clicking the red (on Mac) button to close the application. I couldn't see another way to achieve this as we have to listen on the close signal sent to the application.

File: apps/ledger-live-desktop/src/main/index.ts [hunk 1]
```diff
@@ -195,6 +196,23 @@ app.on("ready", async () => {
...
 
+  let status: number = 0;
+  window.on("close", e => {
+    if (status === 0) {
+      e.preventDefault();
+      getMainWindowAsync()
+        .then(w => {
+          status = 1;
+          if (w && "send" in w.webContents) {
+            w.webContents.send("track-close");
```
gre-ledger: you could name the event "app-close"

apps/ledger-live-desktop/src/renderer/init.tsx
```diff
@@ -224,6 +225,13 @@ async function init() {
     reload();
   });
 
+  ipcRenderer.addListener("track-close", () => {
+    console.log("track-close message received");
+    track("app_close");
```
gre-ledger: make a top level `function onAppClose() {` to make it more obvious as a generic use we may reuse (so we don't have too much usecase specifics in term of event names)

apps/ledger-live-desktop/src/renderer/init.tsx
```diff
@@ -224,6 +225,13 @@ async function init() {
...
+    // Wait enough time for the track function to finish, around ~3 ms
+    setTimeout(() => window.close(), 5);
```
gre-ledger: we didn't need to do it before, is it becoming necessary due to the `e.preventDefault();` done in main ? please document that then.

also i'm not super fan of the timeout approach, what happens without timeout? what if 3ms is not enough for some computer? (i wonder if there is a way to wait for track to finish. like flush the analytics and wait for it to complete)

gre-ledger (commented):
ok then, the approach on doing main->renderer is fine for me, i would just suggest a rename of things to make this a more generic "hook" that we may want for more usecase in future (i remember having similar usecase to clean up some data at close time)

apps/ledger-live-desktop/src/renderer/init.tsx
mitchellv-ledger: I also am not a fan of the timeout - I put it in just to mitigate issues with the exit signal arriving whilst it was executing. Tbh I assume it would be okay even without the timeout as I assume it would let the code finish running before exiting but it is something I would need to verify. The hard part for me is to verify that the event has been sent without the analytics console.

PR #12745, opened 2025-11-13
Feat/canton pending offer

jonezque:
### ✅ Checklist

  - ...

### 📝 Description
- show 2 tables with incoming and outgoing offers for canton
- clicking on offer, opens a drawer with detailed info
- sender can cancel(withdraw) offer
- recipient can accept offer if it's not expired yet or reject it (anytime)

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-20817

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/DeviceAppModal.test.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/DeviceAppModal.test.tsx
@@ -1,0 +1,115 @@
+import React from "react";
+import { render, screen, fireEvent, waitFor } from "tests/testSetup";
+import { setEnv } from "@ledgerhq/live-env";
+import DeviceAppModal from "./DeviceAppModal";
+
+describe("DeviceAppModal", () => {
+  const mockOnConfirm = jest.fn();
+  const mockOnClose = jest.fn();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    setEnv("MOCK", "1");
+    // Create modals container for Modal component
+    if (!document.getElementById("modals")) {
+      const modalsContainer = document.createElement("div");
+      modalsContainer.id = "modals";
+      document.body.appendChild(modalsContainer);
+    }
+  });
+
+  afterEach(() => {
+    setEnv("MOCK", "");
+    const modalsContainer = document.getElementById("modals");
+    if (modalsContainer) {
+      document.body.removeChild(modalsContainer);
+    }
+  });
+
+  describe("when modal is closed", () => {
+    it("should not render when isOpen is false", () => {
+      render(
+        <DeviceAppModal
+          isOpen={false}
+          onConfirm={mockOnConfirm}
+          action="accept"
+          onClose={mockOnClose}
+          appName="Canton"
+        />,
+      );
+
+      expect(screen.queryByTestId("canton-offer-action-modal")).not.toBeInTheDocument();
+    });
+  });
+
+  describe("when modal is open", () => {
+    it("should render modal with correct title for accept action", async () => {
+      render(
+        <DeviceAppModal
+          isOpen={true}
+          onConfirm={mockOnConfirm}
+          action="accept"
+          onClose={mockOnClose}
+          appName="Canton"
+        />,
+      );
+
+      await waitFor(() => {
+        expect(screen.getByTestId("modal-container")).toBeInTheDocument();
+      });
+    });
+
+    it("should render modal with correct title for reject action", async () => {
+      render(
+        <DeviceAppModal
+          isOpen={true}
+          onConfirm={mockOnConfirm}
+          action="reject"
+          onClose={mockOnClose}
+          appName="Canton"
+        />,
+      );
+
+      await waitFor(() => {
+        expect(screen.getByTestId("modal-container")).toBeInTheDocument();
+      });
+    });
+
+    it("should render modal with correct title for withdraw action", async () => {
+      render(
+        <DeviceAppModal
+          isOpen={true}
+          onConfirm={mockOnConfirm}
+          action="withdraw"
+          onClose={mockOnClose}
+          appName="Canton"
+        />,
+      );
+
+      await waitFor(() => {
+        expect(screen.getByTestId("modal-container")).toBeInTheDocument();
+      });
+    });
+
+    it("should call onClose when close button is clicked", async () => {
+      render(
+        <DeviceAppModal
+          isOpen={true}
+          onConfirm={mockOnConfirm}
+          action="accept"
+          onClose={mockOnClose}
+          appName="Canton"
+        />,
+      );
+
+      await waitFor(() => {
+        expect(screen.getByTestId("modal-close-button")).toBeInTheDocument();
+      });
+
+      const closeButton = screen.getByTestId("modal-close-button");
+      fireEvent.click(closeButton);
+
+      expect(mockOnClose).toHaveBeenCalledTimes(1);
+    });
+  });
+});
--- a/apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/DeviceAppModal.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/DeviceAppModal.tsx
@@ -1,0 +1,152 @@
+import React, { useCallback, useMemo, useState, useEffect, FC } from "react";
+import { useTranslation } from "react-i18next";
+import Modal, { ModalBody } from "~/renderer/components/Modal";
+import Box from "~/renderer/components/Box";
+import DeviceAction from "~/renderer/components/DeviceAction";
+import SuccessDisplay from "~/renderer/components/SuccessDisplay";
+import ErrorDisplay from "~/renderer/components/ErrorDisplay";
+import BigSpinner from "~/renderer/components/BigSpinner";
+import Text from "~/renderer/components/Text";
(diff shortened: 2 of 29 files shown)


apps/ledger-live-desktop/src/renderer/families/canton/PendingTransactions/DeviceAppModal.tsx
```diff
@@ -0,0 +1,181 @@
...
+    },
+    [onConfirm],
+  );
+
+  const handleRetry = useCallback(() => {
+    setConfirmationState("pending");
+    setError(null);
+  }, []);
+
+  const getSuccessTitle = () => {
```
acewf: To make this component clear, could extract this functions to be declared outside of it.

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransactions/DeviceAppModal.tsx
```diff
@@ -0,0 +1,181 @@
...
+  };
+
+  const getModalTitle = () => {
+    const actionLabel = getActionLabel();
+    return t("families.canton.pendingTransactions.deviceAppModal.title", {
+      action: actionLabel,
+    });
+  };
+
+  const renderContent = () => {
```
acewf: don't think having this function here helps, when you can have the function code directly on the return. If you want to extract it from the return block do a component that receives a prop and renders accordingly.

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransactions/PendingTransactionDetails.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+  const [timeRemaining, setTimeRemaining] = useState<string>("");
+
+  useEffect(() => {
+    if (!proposal || proposal.isExpired) {
+      setTimeRemaining("");
+      return;
+    }
+
+    const updateTimeRemaining = () => {
```
acewf: the logic of this function can be external to the component, improving readability.
```
const remainingTime = getRemainingTime()
if (!remainingTime) {
   setTimeRemaining("");
}
setTimeRemaining(remainingTime);
```

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransactions/PendingTransactionDetails.tsx
```diff
@@ -0,0 +1,287 @@
...
+      onOpenModal(contractId, "withdraw");
+      // Close drawer after modal state is set
+      setTimeout(() => {
+        onClose?.();
+      }, 0);
+    },
+    [onClose, onOpenModal],
+  );
+
+  const formatDate = useCallback((timestamp: number) => {
```
acewf: why do we need this function to be declared here? don't see a particular reason to "cache" this function, because declaring outside of the component achieves the same result.

apps/ledger-live-desktop/src/renderer/screens/account/index.tsx
```diff
@@ -169,6 +179,10 @@ const AccountPage = ({
             <AccountBodyHeader account={account} parentAccount={parentAccount} />
           ) : null}
 
+          {PendingTransactions && mainAccount?.currency.family === "canton" ? (
```
acewf: no need for ternary!
```
{PendingTransactions && mainAccount?.currency.family === "canton" && (
    <PendingTransactions account={account as Account} />
)}
```

libs/coin-modules/coin-canton/src/network/gateway.ts
```diff
@@ -58,6 +58,33 @@ export type PrepareTransferRequest = {
   reason?: string;
 };
 
+export type PrepareTransferInstructionRequest =
```
acewf: not following why is OR on the entire object and not only on the property type!
Doesn't this give the same result?
```
type PrepareTransferInstructionRequest = {
      type: "accept-transfer-instruction" | "reject-transfer-instruction" | "withdraw-transfer-instruction";
      contract_id: string;
      reason?: string;
    }
```

libs/ledger-live-common/src/families/canton/react.ts
```diff
@@ -0,0 +1,73 @@
...
+        contractId,
+        type,
+        reason,
+      );
+    },
+    [cantonBridge, currency, account, partyId],
+  );
+
+  // Wrapper functions for backward compatibility
+  const acceptOffer = useCallback(
```
acewf: don't think we need the`useCallback` here, whenever the the component rerenders is because some property was updated, which will make `transferInstruction` being updated also, because the list of dependencies is the same as the component.

apps/ledger-live-desktop/src/renderer/screens/account/index.tsx
semeano: In React it is preferred to not use `&&`, that could lead to some rendering issues (see [here](https://react.dev/learn/conditional-rendering#logical-and-operator-)). I it is preferred to use the ternary and return `null`.

apps/ledger-live-desktop/src/renderer/screens/account/index.tsx
```diff
@@ -93,6 +98,11 @@ const AccountPage = ({
   const specific = mainAccount ? getLLDCoinFamily(mainAccount.currency.family) : null;
   const AccountBodyHeader = specific?.AccountBodyHeader;
   const AccountSubHeader = specific?.AccountSubHeader;
+
+  const PendingTransactions =
```
hedi-edelbloute: Please use same arch as what's been done for AccountBodyHeader and AccountSubHeader

Also, is it fine for you if we rename to PendingOffers or PendingTransferProposals (name that you used in CantonResources) so we differ it from pendingOperations (which in the future might be transactions as we will move from operation to transactions) ?

libs/coin-modules/coin-canton/src/network/gateway.ts
jonezque: nice catch

apps/ledger-live-desktop/src/renderer/screens/account/index.tsx
acewf: the pitfall is only if the operator isn't a boolean, which in this case it clear is.

semeano: yes, although that could change in the future (we never know), that's why I prefer to use a ternary every time.

LucasWerey:
Can you add a description and a video please 🙏

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/DeviceAppModal.test.tsx
```diff
@@ -0,0 +1,371 @@
...
+      {onRetry && (
+        <button onClick={onRetry} data-testid="error-retry">
+          Retry
+        </button>
+      )}
+    </div>
+  ),
+}));
+
+jest.mock("~/renderer/components/BigSpinner", () => ({
```
LucasWerey: There are too many mocks can you remove the ones that shouldn't be there please

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/utils.test.ts
```diff
@@ -0,0 +1,216 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { renderHook, act } from "@testing-library/react";
```
LucasWerey:
```suggestion
import { renderHook, act } from "tests/testSetup";
```

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+    updateTimeRemaining();
+    const interval = setInterval(updateTimeRemaining, 1000);
+
+    return () => clearInterval(interval);
+  }, [expiresAtMicros, isExpired]);
+
+  return timeRemaining;
+};
+
+export const formatDate = (timestamp: number) => {
```
LucasWerey: can't we use `import { dayFormat, useDateFormatter } from "~/renderer/hooks/useDateFormatter";` ?

apps/ledger-live-desktop/src/renderer/families/canton/PendingTransferProposals/DeviceAppModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+const DeviceAppModal: FC<Props> = ({ isOpen, onConfirm, action, onClose }) => {
+  const { t } = useTranslation();
+  const [confirmationState, setConfirmationState] = useState<ConfirmationState>("pending");
+  const [error, setError] = useState<Error | null>(null);
+
+  const actionConnect = useConnectAppAction();
+  const request = useMemo(
+    () => ({
+      appName: "Canton",
```
hedi-edelbloute: Minor: Can be a reference to currency managerAppName attribute

hedi-edelbloute (dismissed):

jonezque:
https://github.com/user-attachments/assets/8335b882-ae16-4226-90a0-623f09225bb6

LucasWerey:
> Screen.Recording.2025-11-19.at.12.33.11.mov

In the video, I can see there is a misalignment between "expires in" and "amount"
image


Also, can we add a blur effect behind the modal?

jonezque:
Screenshot 2025-11-19 at 14 09 45

ishaba (approved):

PR #12865, opened 2025-11-19
feat: getBlock with tests in coin-hedera

mateuszpalosz-ext:
### ✅ Checklist

  - getBlock & getBlockInfo methods are available within the Alpaca API in coin-hedera module

### 📝 Description

This PR includes getBlock logic for Alpaca API in coin-hedera module.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/.changeset/angry-flowers-sell.md
+++ b/.changeset/angry-flowers-sell.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-hedera": minor
+---
+
+feat: getBlock in coin-hedera
--- a/libs/coin-modules/coin-hedera/src/logic/utils.test.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/utils.test.ts
@@ -460,6 +460,66 @@ describe("logic utils", () => {
     });
   });
 
+  describe("formatTransactionId", () => {
+    it("converts SDK TransactionId format to mirror node format", () => {
+      const mockTransactionId = {
+        toString: () => "0.0.8835924@1759825731.231952875",
+      } as TransactionId;
+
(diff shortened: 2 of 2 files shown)


libs/coin-modules/coin-hedera/src/network/api.ts
```diff
@@ -160,6 +160,34 @@ async function getLatestTransaction(): Promise<HederaMirrorTransaction> {
...
+  endTimestamp: string,
+): Promise<HederaMirrorTransaction[]> {
+  const transactions: HederaMirrorTransaction[] = [];
+  const params = new URLSearchParams({
+    limit: "100",
+    order: "desc",
+  });
+
+  params.append("timestamp", `gte:${startTimestamp}`);
+  params.append("timestamp", `lte:${endTimestamp}`);
```
jnicoulaud-ledger: Shouldn't we have one of the two using strict comparison ? Otherwise, a transaction exactly on a boundary will be in 2 blocks ?

MateuszPaloszBD: The current implementation works correctly because for the `end timestamp` we return the value as `end: ${endTimestamp}.999999999`, which ensures that the blocks never overlap even though we’re using a less-than-or-equal comparison. But for better readability, it would indeed look cleaner if the `endTimestamp` will return `.000000000` for nanos and later on, we can make `lt` comparison to make it cleaner

jnicoulaud-ledger: Ha, yes, so not blocker but I think it would be cleaner indeed

jprudent (approved):

PR #12867, opened 2025-11-19
feat: configure noah via ff

hhumphrey-ledger:
### ✅ Checklist

  - ...

### 📝 Description

The supported currencies for the noah flow will change in the future, instead of requiring a release every time it does change we are going to configure this via the firebase feature flag. 

Remove all the old checks and replace with a firebase check instead.

### ❓ Context

https://ledgerhq.atlassian.net/browse/LIVE-23283

---

### DIFF
--- a/libs/ledgerjs/packages/types-live/src/feature.ts
+++ b/libs/ledgerjs/packages/types-live/src/feature.ts
@@ -274,12 +274,7 @@ export type Features = CurrencyFeatures & {
   llmSyncOnboardingIncr1: DefaultFeature;
   lldSyncOnboardingIncr1: DefaultFeature;
   cantonSkipPreapprovalStep: DefaultFeature;
-  noah: {
-    enabled: boolean;
-    params: {
-      activeCurrencyIds: string[];
-    };
-  };
+  noah: Feature_Noah;
   newSendFlow: DefaultFeature;
   lldSessionReplay: Feature_LldSessionReplay;
   zcashShielded: DefaultFeature;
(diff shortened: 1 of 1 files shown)


apps/ledger-live-desktop/src/renderer/modals/Receive/index.tsx
```diff
@@ -27,14 +26,14 @@ type State = {
...
   isNoahActive: boolean,
   account: AccountLike | null | undefined,
-  parentAccount: Account | null | undefined,
   isOnboardingFlow?: boolean,
+  activeCurrencyIds?: string[],
 ): State {
-  const mainAccount = account ? getMainAccount(account, parentAccount) : null;
   const accountCurrency = account ? getAccountCurrency(account) : null;
-  const module =
-    mainAccount && getLLDCoinFamily(mainAccount.currency.family)?.shouldUseReceiveOptions;
```
gre-ledger, line 36: if `shouldUseReceiveOptions` is removed, is it still used somewhere else? if not, i would also recommend to remove it from the families/types.ts

mcayuelas-ledger (approved):

apps/ledger-live-desktop/src/renderer/modals/Receive/index.tsx
hhumphrey-ledger: It isn't used anywhere now, but it also doesn't exist in `families/types` which is a bit odd...

PR #12870, opened 2025-11-19
chore: [LIVE-23484] remove getAccountDetails redundant call

acewf:
### ✅ Checklist

  - this change the list of operations might be bigger
  - the account balance now includes the non voting Locked Balance

### 📝 Description

This work item addresses the issue of duplicated sync calls for balances, blocks, and operations in the Celo network, with ongoing efforts to optimize API usage and resolve data inconsistencies:
- It was identified that current implementation fetches data twice using getAccountDetails and evmGetAccountShape, which needs to be eliminated to avoid redundant requests.
- The team attempted to replace both methods with available API functions from the coin module.
- It was suggested to switch from celoKit to getStakes for retrieving staking data, but getStakes cannot fetch nonvotingLockedBalance, and the staking contract address needs correction to the proper LockedGold contract.

### ❓ Context

- **JIRA or GitHub link**:https://ledgerhq.atlassian.net/browse/LIVE-23483

---

### DIFF
--- a/libs/coin-modules/coin-celo/src/bridge/synchronisation.test.ts
+++ b/libs/coin-modules/coin-celo/src/bridge/synchronisation.test.ts
@@ -1,13 +1,9 @@
 import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets";
 import BigNumber from "bignumber.js";
-import { lockedGold, nonVoting, electionConfig } from "../../bridge/__mocks__/celokit.mock";
-import {
-  mockCreateApi,
-  erc20Operation,
-  nativeOperation,
-} from "../../bridge/__mocks__/operations-list.mock";
-import { mockGetCoinBalance, mockTokenEvmLogic } from "../../bridge/__mocks__/evm.mock";
-import { getAccountShape, clearEvmApiInstance } from "../../bridge/synchronisation";
+import { lockedGold, nonVoting, electionConfig } from "./__mocks__/celokit.mock";
+import { mockCreateApi, erc20Operation, nativeOperation } from "./__mocks__/operations-list.mock";
+import { mockGetCoinBalance, mockTokenEvmLogic } from "./__mocks__/evm.mock";
+import { getAccountShape } from "./synchronisation";
 
 const defaultInfo = {
   address: "0x79D5A290D7ba4b99322d91b577589e8d0BF87072",
@@ -24,9 +20,6 @@ nonVoting.mockResolvedValue(new BigNumber(0));
 electionConfig.mockResolvedValue({ maxNumGroupsVotedFor: 10 });
 
 describe("When getting the account shape", () => {
-  afterEach(() => {
-    clearEvmApiInstance();
-  });
   it("returns the account with correct balance and spendable balance", async () => {
     // Given
     mockCreateApi.mockReturnValue({
--- a/libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
+++ b/libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
@@ -38,18 +38,6 @@ const getTypeFromString = (value: string) => {
   return operationsTypes.find(item => item === value) ?? "NONE";
 };
 
-let evmAPI: ReturnType<typeof createApi> | null = null;
-const getApi = (configEvm: Parameters<typeof createApi>[0], currencyId: string) => {
-  if (!evmAPI) {
-    evmAPI = createApi(configEvm, currencyId);
-  }
-  return evmAPI;
-};
-
-export const clearEvmApiInstance = () => {
-  evmAPI = null;
-};
-
 const resolveTypesFromContracts = (item: {
   type: OperationType;
   senders: string[];
@@ -77,6 +65,7 @@ const getOperationsList = async ({
   config,
   accountId,
   contracts,
+  api,
 }: {
   info: AccountShapeInfo<CeloAccount>;
   config: SyncConfig;
@@ -85,16 +74,15 @@ const getOperationsList = async ({
     locked: string;
     election: string;
   };
+  api: ReturnType<typeof createApi>;
 }) => {
   const { address, currency, initialAccount } = info;
   const blacklistedTokenIds = config.blacklistedTokenIds || [];
-  const configEvm = getCoinConfig(currency).info;
 
   const syncHash = await getSyncHash(currency, blacklistedTokenIds);
   const shouldSyncFromScratch =
     syncHash !== initialAccount?.syncHash || initialAccount === undefined;
   const latestSyncedHeight = shouldSyncFromScratch ? 0 : initialAccount.blockHeight;
-  const api = getApi(configEvm, currency.id);
   const rawOperationsList = (
     await api.listOperations(address, {
       minHeight: Math.max(latestSyncedHeight - SAFE_REORG_THRESHOLD, 0),
@@ -240,7 +228,7 @@ export const getAccountShape: GetAccountShape<CeloAccount> = async (info, config
   const syncHash = await getSyncHash(currency, blacklistedTokenIds);
 
   const nodeApi = getNodeApi(currency);
-  const api = getApi(configEvm, currency.id);
+  const api = createApi(configEvm, currency.id);
   const blockInfo = await api.lastBlock();
   const balance = await nodeApi.getCoinBalance(currency, address);
 
@@ -249,6 +237,7 @@ export const getAccountShape: GetAccountShape<CeloAccount> = async (info, config
     config,
     accountId,
     contracts: { locked: lockedGold.address, election: election.address },
+    api,
   });
 
   const nativeOperations = operationsList.filter(({ isSubAccount }) => !isSubAccount);
--- a/libs/coin-modules/coin-celo/jest.config.js
+++ b/libs/coin-modules/coin-celo/jest.config.js
@@ -11,7 +11,7 @@ module.exports = {
       },
     ],
   },
-  testPathIgnorePatterns: ["lib/", "lib-es/", ".*\\.integration\\.test\\.[tj]s"],
+  testPathIgnorePatterns: ["lib/", "lib-es/", ".*\\.integ\\.test\\.[tj]s"],
   modulePathIgnorePatterns: ["lib/", "lib-es/"],
   workerThreads: true,
   coverageReporters: ["json", ["lcov", { file: "lcov.info", projectRoot: "../../../" }], "text"],
--- a/libs/coin-modules/coin-celo/src/bridge/synchronisation.integ.test.ts
+++ b/libs/coin-modules/coin-celo/src/bridge/synchronisation.integ.test.ts
@@ -1,5 +1,5 @@
 import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets";
-import { getAccountShape, clearEvmApiInstance } from "./synchronisation";
+import { getAccountShape } from "./synchronisation";
 import { setCoinConfig, EvmCoinConfig } from "@ledgerhq/coin-evm/config";
 import BigNumber from "bignumber.js";
 import type { CryptoAssetsStore } from "@ledgerhq/types-live";
@@ -42,15 +42,12 @@ describe("Integration — getAccountShape with real implementations", () => {
         }) as unknown as EvmCoinConfig,
     );
   });
-  afterEach(() => {
-    clearEvmApiInstance();
-  });
 
   it("should fetch account data with expected structure", async () => {
     const result = await getAccountShape(defaultInfo, defaultConfig);
 
-    expect(result.balance).toEqual(BigNumber(996309127371000000));
-    expect(result.spendableBalance).toEqual(BigNumber(696309127371000000));
+    expect(result.balance).toEqual(BigNumber(306028106371000000));
+    expect(result.spendableBalance).toEqual(BigNumber(6028106371000000));
     expect(result.blockHeight).toBeGreaterThan(0);
     expect(result.operations?.length).toBeGreaterThan(0);
     expect(result.celoResources?.lockedBalance).toEqual(BigNumber(300000000000000000));


francois-guerin-ledger:
By curiosity, do you know how hard would it be to go the other way around ? That is removing `coin-evm` dependency from `coin-celo` ?
My point is that this architecture is quite unexpected and refrains `coin-evm` from evolving without impacting `coin-celo`.

acewf:
> By curiosity, do you know how hard would it be to go the other way around ? That is removing `coin-evm` dependency from `coin-celo` ? My point is that this architecture is quite unexpected and refrains `coin-evm` from evolving without impacting `coin-celo`.

@francois-guerin-ledger sorry i missed you message, the problem with the `getAccountDetails` method it just brings the information available here, so doesn't contain the subaccounts information, and on top of that doesn't contain the full transactions list.
https://celo.coin.ledger.com/indexer/account_details/0x79D5A290D7ba4b99322d91b577589e8d0BF87072

libs/coin-modules/coin-celo/src/bridge/account-sync-helpers.ts
```diff
@@ -0,0 +1,270 @@
+import BigNumber from "bignumber.js";
```
acewf: what is on this file, is 99% a duplication of the https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-evm/src/bridge/synchronization.ts

we just stopped point to this file, because the `bridge` is going to be removed soon.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,213 @@
...
+        hash: item.tx.hash,
+        type: resolveTypesFromContracts({
+          type,
+          senders: item.senders,
+          recipients: item.recipients,
+          address,
+          value: item.value,
+          contracts: contracts,
+        }),
+        value: new BigNumber((item.details?.assetAmount as string) || item.value.toString()),
```
acewf, line 110: @hedi-edelbloute can you please pay special attention to this, during the implementation i was just trying to use `item.value` since it made sense to me this was the transaction value amount, specially also because the `Operation` type schema is the following
```
export type Operation<MemoType extends Memo = MemoNotSupported> = {
  ...
  value: bigint;
  ....
  tx: {
   ...
    fees: bigint; // network fees paid
  };
};
```

So when running the code as 
```
value: new BigNumber(item.value.toString()),
```
I was getting massive transaction values like in Millions of dollars but just on the ERC20 stable coins, and it had taken me some time to understand that in reality this value was the same as digits as the fees, but just on a different magnitude.

This seems a bug to me on the `listOperations` method on the api, but can you double check please.

libs/coin-modules/coin-evm/src/logic/craftTransaction.test.ts
```diff
@@ -356,7 +356,7 @@ describe("craftTransaction", () => {
                 intentType: "staking",
                 type: `staking-${transactionType}`,
                 sender: "0x1234567890abcdef1234567890abcdef12345678",
-                recipient: "0x55E1A0C8f376964bd339167476063bFED7f213d5",
+                recipient: "0x6cC083Aed9e3ebe302A6336dBC7c921C9f03349E",
```
acewf: correct gold contract on celo https://docs.celo.org/tooling/contracts/core-contracts#:~:text=0xf2a347F184b0Fef572C7CBd2c392359eCcf43F3c-,LockedCelo,-0x6cC083Aed9e3ebe302A6336dBC7c921C9f03349E

libs/coin-modules/coin-celo/src/bridge/synchronisation.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+    mockGetCoinBalance.mockResolvedValueOnce(new BigNumber(1010));
+
+    // When
+    const result = await getAccountShape(defaultInfo, defaultConfig);
+
+    // Then
+    expect(result).toBeDefined();
+    expect(result.balance).toEqual(BigNumber(1010));
+    expect(result.spendableBalance).toEqual(BigNumber(1010));
```
francois-guerin-ledger:
```suggestion
    expect(result).toMatchObject({
      balance: BigNumber(1010),
      spendableBalance: BigNumber(1010),
    });
```
More compact and does not rely on the weak and redundant `toBeDefined` that we are trying to remove from our codebase (cf https://github.com/LedgerHQ/ledger-live/pull/13159)

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,213 @@
...
+
+const getTypeFromString = (value: string) => {
+  const finding = operationsTypes.filter(item => {
+    return item === value;
+  });
+  let result: OperationType = "NONE";
+  if (finding.length > 0) {
+    result = finding[0];
+  }
+  return result;
```
francois-guerin-ledger: Can we use `find` instead ? Maybe
```ts
return operationsTypes.find(item => item === value) ?? "NONE";
```

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,213 @@
...
+  "WITHDRAW",
+  "REVOKE",
+  "ACTIVATE",
+  "REGISTER",
+  "VOTE",
+] as const;
+
+const SAFE_REORG_THRESHOLD = 80;
+
+type OperationType = (typeof operationsTypes)[number];
```
francois-guerin-ledger, line 35: I think an `enum` would be cleaner

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,213 @@
...
+};
+
+const getOperationsList = async ({
+  info,
+  config,
+  accountId,
+  contracts,
+}: {
+  info: Parameters<GetAccountShape<CeloAccount>>[0];
+  config: Parameters<GetAccountShape<CeloAccount>>[1];
```
francois-guerin-ledger:
```suggestion
  info: AccountShapeInfo<CeloAccount>;
  config: SyncConfig;
```

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
francois-guerin-ledger: Not a bug but definitely something to improve because this is confusing, thanks for the comment!

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: i always avoid using enums explanation here https://thoughtbot.com/blog/the-trouble-with-typescript-enums

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
hedi-edelbloute: Hello @acewf , each token can have a different magnitude, which can be different from main coin ETH magnitude, not a bug

See https://cdn.live.ledger.com/cryptoassets/evm/1/erc20.json

acewf: @francois-guerin-ledger the annoying part as also because  `item.details` type is `Record<string, unknow>` so the compiler doesn't recognise `item.details?.assetAmount` and is the reason i had to do 
```
item.details?.assetAmount as string
```

hedi-edelbloute: Also, operation.value is not tx.value, tx.value is native coin transfer related to the operation of moving funds, represented by operation.value

hedi-edelbloute: If you see a token with wrong magnitude please tell us :)

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: also have explanation from Matt here https://www.totaltypescript.com/why-i-dont-like-typescript-enums

acewf:
@hedi-edelbloute and @francois-guerin-ledger can you have another look.

The githhub pipeline isn't green due to Vercel – web-tools, but this isn't related to changes on the PR.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: @francois-guerin-ledger marking this one as resolved, but reopen again if you think it needs it.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,207 @@
...
+  const blacklistedTokenIds = config.blacklistedTokenIds || [];
+  const configEvm = {
+    ...getCoinConfig(currency).info,
+  } as const;
+
+  const syncHash = await getSyncHash(currency, blacklistedTokenIds);
+  const shouldSyncFromScratch =
+    syncHash !== initialAccount?.syncHash || initialAccount === undefined;
+  const latestSyncedHeight = shouldSyncFromScratch ? 0 : initialAccount.blockHeight;
+  const api = createApi(configEvm, currency.id);
```
hedi-edelbloute: Is there an impact on creating a new api at each sync ?

libs/coin-modules/coin-evm/src/staking/contracts.ts
```diff
@@ -19,9 +19,9 @@ export const STAKING_CONTRACTS: Record<string, StakingContractConfig> = {
   },
 
   // Celo staking
-  // Source: https://celo.blockscout.com/address/0x55E1A0C8f376964bd339167476063bFED7f213d5?tab=contract_source_code
+  // Source: https://docs.celo.org/tooling/contracts/core-contracts#celo-mainnet
   celo: {
-    contractAddress: "0x55E1A0C8f376964bd339167476063bFED7f213d5",
+    contractAddress: "0x6cC083Aed9e3ebe302A6336dBC7c921C9f03349E",
```
hedi-edelbloute: Maybe you can add an integration test to ensure alpaca api methods return what is expected ?

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: good point, probably this configuration can be static for CELO probably meaning we can create the api instance outside of the sync.

libs/coin-modules/coin-evm/src/staking/contracts.ts
acewf: this is kind out of the scope of the ticket. 
I was changing this contract address because i found the is incorrect while trying to implement the method `getStakes`, but ended not using this method (doesn't provides all data needed).

I will remove this change, and maybe we can create another ticket for adding the correct address, and have more specific tests then.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: had gone for the singleton approach!

libs/coin-modules/coin-evm/src/staking/contracts.ts
hedi-edelbloute: If you prefer you can create a synchronization.integ.test.ts that does the same thing: ensuring staking info retrieval does not break

acewf: question, since i might be missing something. 
`coin-celo` is only imported by `ledger-live-common` and i can see the integration tests are already there [bridge.integration](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/families/celo/bridge.integration.test.ts)

I couldn't find also other coin-modules with integration tests, so does it make sense to have integration tests at this level?

hedi-edelbloute: It's better to have it on the coin-module/api level hence why i asked for it in the first place, cf:

```
libs/coin-modules/coin-evm/src/api/index.integ.test.ts
libs/coin-modules/coin-stellar/src/api/index.integ.test.ts
libs/coin-modules/coin-xrp/src/api/index.integ.test.ts
libs/coin-modules/coin-tron/src/api/index.integ.test.ts
...
```

But if you don't want to add them there the standard way, you can create a sync integration test that does the same thing, eg. ensuring stakes are retrieved correctly

acewf: we are using the `coin-evm/api` methods inside Celo coin module, in this case the standard way already has it.

I can have a sync integration tests, was just trying to understand if the coin-module is the correct place to have them since they don't exist on bridges on other coin-modules.

acewf: @hedi-edelbloute added the integration tests, the tests are limited to check balances due to extra configuration for the other tokens that is necessary which i am not sure we should have.

libs/coin-modules/coin-celo/src/__tests__/bridge/synchronisation.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      listOperations: jest.fn().mockResolvedValueOnce([[nativeOperation]]),
+      lastBlock: jest.fn().mockResolvedValueOnce({ height: 4444 }),
+    });
+    mockGetCoinBalance.mockResolvedValueOnce(new BigNumber(20));
+
+    // When
+    const result = await getAccountShape(defaultInfo, defaultConfig);
+
+    // Then
+    expect(result).toBeDefined();
```
francois-guerin-ledger:
```suggestion
```
`expect(result).toMatchObject` below already ensures it is defined

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -23,33 +227,70 @@ export const getAccountShape: GetAccountShape<CeloAccount> = async (info, config
...
-    balance,
-    blockHeight,
+    balance: balance.plus(lockedBalance) || new BigNumber(0),
+    blockHeight: blockInfo.height || 0,
+    operations,
     operationsCount: operations.length,
-    spendableBalance,
-    subAccounts: getEnv("ENABLE_CELO_TOKENS") === false ? [] : fromEvm.subAccounts || [],
-    syncHash: fromEvm.syncHash,
+    spendableBalance: balance || new BigNumber(0),
```
francois-guerin-ledger:
```suggestion
    spendableBalance: balance,
```
`balance` is a `BigNumber`, it can not be falsy

libs/coin-modules/coin-celo/src/bridge/synchronisation.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    // When
+    const result = await getAccountShape(defaultInfo, defaultConfig);
+
+    // Then
+    expect(result).toMatchObject({
+      blockHeight: 4444,
+      syncHash: "0x0000000000000000000000000000000000001d00",
+    });
+    expect(result.operations?.length).toBe(0);
+  });
```
francois-guerin-ledger, line 58: You are using the same setup, it feels this can be one test
```ts
expect(result).toMatchObject({
  blockHeight: 4444,
  syncHash: "0x0000000000000000000000000000000000001d00",
  balance: BigNumber(1010),
  spendableBalance: BigNumber(1010),
 operations: []
});

File: libs/coin-modules/coin-celo/src/__tests__/bridge/synchronisation.test.ts [hunk 13]
```diff
@@ -0,0 +1,132 @@
...
+    // Then
+    expect(result).toMatchObject({
+      blockHeight: 4444,
+      celoResources: {
+        registrationStatus: true,
+        pendingWithdrawals: [],
+      },
+    });
+    expect(result.operations?.length).toBe(0);
+    expect(result.subAccounts?.length).toBe(1);
```
francois-guerin-ledger: We can have less `expect`s here, and give some info in the sub accounts. Also is it normal to have one ERC operation but 0 parent operation ?
```suggestion
    expect(result).toMatchObject({
      blockHeight: 4444,
      operations: [],
      subAccounts: [expect.objectContaining(...)],
      celoResources: {
        registrationStatus: true,
        pendingWithdrawals: [],
      },
    });
```

libs/coin-modules/coin-celo/src/__tests__/bridge/synchronisation.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    // When
+    const result = await getAccountShape(defaultInfo, defaultConfig);
+
+    // Then
+    expect(result).toMatchObject({
+      blockHeight: 4444,
+      operationsCount: 1,
+    });
+    expect(result.operations?.length).toBe(1);
```
francois-guerin-ledger: You ca, me more specific on the operation shape
```suggestion
    expect(result).toMatchObject({
      blockHeight: 4444,
      operationsCount: 1,
      operations: [expect.objectContaining(...)],
    });
```

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,219 @@
...
+  contracts: {
+    locked: string;
+    election: string;
+  };
+}) => {
+  const { address, currency, initialAccount } = info;
+  const blacklistedTokenIds = config.blacklistedTokenIds || [];
+  const configEvm = {
+    ...getCoinConfig(currency).info,
+  } as const;
```
francois-guerin-ledger: Why `as const` ?
```suggestion
  const configEvm = getCoinConfig(currency).info;
```

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,219 @@
...
+
+const SAFE_REORG_THRESHOLD = 80;
+
+type OperationType = (typeof operationsTypes)[number];
+
+const getTypeFromString = (value: string) => {
+  return operationsTypes.find(item => item === value) ?? "NONE";
+};
+
+let evmAPI: ReturnType<typeof createApi> | null = null;
```
francois-guerin-ledger:
```suggestion
let evmAPI: Api<MemoNotSupported, BufferTxData> | null = null;
```
We can be more explicit

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,219 @@
...
+const getApi = (configEvm: any, currencyId: string) => {
```
francois-guerin-ledger:
```suggestion
const getApi = (configEvm: EvmConfig, currencyId: string) => {
```
Do you think we can remove `createEvmApi` from `libs/ledger-live-common/src/families/celo/setup.ts` then ?

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -1,15 +1,219 @@
...
+  if (!evmAPI) {
+    evmAPI = createApi(configEvm, currencyId);
+  }
+  return evmAPI;
+};
+
+export const clearEvmApiInstance = () => {
+  evmAPI = null;
+};
```
francois-guerin-ledger: So this function on the production code is only for tests, right ? This is not a common pattern. Did you try to do without ?

libs/coin-modules/coin-celo/src/__tests__/bridge/synchronisation.test.ts
acewf: feels like can become quite verbose, but sure lets add it.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: the bridge version was casting the type, and with the `as const` makes it infer the type. But i don't need it anymore in this latest version of the code.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: but this also means i need to import more types, that shouldn't be in the concern of this file. 
If the `createApi` method types changes slightly will create a need to do also a fix on this file.

And in reality i just need to know the api instance contains the methods like `listOperations`. 
Feels to me the codebase should infer more then trying to be to explicit.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: don't want to import more types then the necessary ones (less dependencies better imo), that said the usage of the `any` type isn't also a good option.

But i can infer the argument type here.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: added the singleton to avoid the runtime create a new instance on every getOperationsList call, but this means for the jest tests we can't get a clean instance for each new test.

on other option would also have `jest.isolateModules` but this one also seems cumbersome.
And in the end didn't seem a bad idea to have away to reset/clean the instance with on singletons.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
francois-guerin-ledger: I don't see the issue with importing types to be honest. But no big deal since this file is temporary

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
francois-guerin-ledger: Same comment as above, fine for now

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
francois-guerin-ledger: Creating an api instance is precisely what is done in the generic adapter anyway https://github.com/LedgerHQ/ledger-live/blob/286581452662c74ff2a284d7a8a77c3940aa4ab0/libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts#L33-L36

francois-guerin-ledger:
While you are at it, could you try to remove `libs/coin-modules/coin-evm/src/bridge` ?

acewf:
> While you are at it, could you try to remove `libs/coin-modules/coin-evm/src/bridge` ?

@francois-guerin-ledger i will do it in a separate PR if that is ok.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: @francois-guerin-ledger this is what i did initially, but then @hedi-edelbloute pointed out that we should avoid creating the instances, and this is the reason i went for the singleton.

libs/coin-modules/coin-celo/src/__tests__/bridge/Untitled-1.yaml
```diff
@@ -0,0 +1,17 @@
...
+│           blockHeight: 2000,
+│           blockHash: '0xsa',
+│           accountId: 'js:2:celo:0x79D5A290D7ba4b99322d91b577589e8d0BF87072:+0xcc',
+│           date: 2026-01-09T00:00:00.000Z,
+│           hasFailed: false,
+│           fee: BigNumber { s: 1, e: 14, c: [Array] },
+│           transactionSequenceNumber: BigNumber { s: 1, e: 0, c: [Array] },
+│           extra: {},
+│           isSubAccount: true
+│         }
```
francois-guerin-ledger: What is this file ? Also this is no `yaml` 🤔

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -23,33 +225,70 @@ export const getAccountShape: GetAccountShape<CeloAccount> = async (info, config
...
-
   const votes = accountRegistrationStatus ? await getVotes(address) : [];
 
-  const fromEvm = await evmGetAccountShape(info, config);
+  const lockedBalance = await lockedGold.getAccountTotalLockedGold(address);
+  const nonvotingLockedBalance = await lockedGold.getAccountNonvotingLockedGold(address);
+
+  const configEvm = {
+    ...getCoinConfig(currency).info,
+  } as const;
```
francois-guerin-ledger, line 226: Why can't we have
```suggestion
  const configEvm = getCoinConfig(currency).info;
```

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -23,33 +225,70 @@ export const getAccountShape: GetAccountShape<CeloAccount> = async (info, config
...
+    blacklistedTokenIds,
+    accountId,
+    info,
+  });
+
+  const shouldSyncFromScratch =
+    syncHash !== initialAccount?.syncHash || initialAccount === undefined;
+  const operations = shouldSyncFromScratch
+    ? mergeOps([], nativeOperations)
+    : mergeOps(oldOperations, nativeOperations);
```
francois-guerin-ledger: Maybe ?
```suggestion
  const operations = mergeOps(shouldSyncFromScratch ? [] : oldOperations, nativeOperations);
```

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
```diff
@@ -23,33 +225,70 @@ export const getAccountShape: GetAccountShape<CeloAccount> = async (info, config
...
+
+  const initialSubAccounts = initialAccount?.subAccounts?.filter(item => {
+    return item.token.ticker !== "CELO";
+  });
```
francois-guerin-ledger, line 257: Is it expected to have a sub account that handles a token with ticker CELO ?

francois-guerin-ledger:
> > While you are at it, could you try to remove `libs/coin-modules/coin-evm/src/bridge` ?
> 
> @francois-guerin-ledger i will do it in a separate PR if that is ok.

FIne by me

libs/coin-modules/coin-celo/src/__tests__/bridge/Untitled-1.yaml
acewf: ahh was a temp file that i didn't realised i saved it and was committed, removing it.

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
acewf: this relates to behaviour we want to change, described on this ticket https://ledgerhq.atlassian.net/browse/LIVE-22821.

In short due to token duality the balances are compounded, and the decision from product was to hide the ERC20 CELO token, and just show the CELO native token

libs/coin-modules/coin-celo/src/bridge/synchronisation.ts
hedi-edelbloute: Hi @acewf , I just asked if there was an impact, check my comment
https://github.com/LedgerHQ/ledger-live/pull/12870#discussion_r2630428828

acewf: singleton is now removed, hopefully is inline with everyone expectations.

libs/coin-modules/coin-celo/src/bridge/synchronisation.integration.test.ts
```diff
@@ -0,0 +1,72 @@
+import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets";
+import { getAccountShape, clearEvmApiInstance } from "./synchronisation";
```
francois-guerin-ledger: How can it work, this function does not exist anymore ?

acewf: not sure why lint was complaining here. Let me remove it.

PR #12967, opened 2025-11-25
feature/LIVE-20764 Validate address implementation on all currencies

Moustafa-Koterba:
### ✅ Checklist

  - No impact

#### Context
We need a common function to validate the content of the clipboard in the new send flow redesign. This PR aims to add this new function `validateAddress` for the bridge. Code is not called for the moment

#### What have changed
- New function `validateAddress` for the bridge
- `validateAddress` function implementation for each coin with unit tests

### ❓ Context

- **JIRA or GitHub link**:  [LIVE-20764](https://ledgerhq.atlassian.net/browse/LIVE-20764?atlOrigin=eyJpIjoiY2NhYTYxNWI5NTBhNDRhZTgwMDYyMTdjOWEzMzNmYTgiLCJwIjoiaiJ9)

---

### DIFF
--- a/libs/coin-modules/coin-concordium/src/bridge/index.test.ts
+++ b/libs/coin-modules/coin-concordium/src/bridge/index.test.ts
@@ -15,6 +15,7 @@ describe("createBridges", () => {
         signRawOperation: expect.any(Function),
         sync: expect.any(Function),
         updateTransaction: expect.any(Function),
+        validateAddress: expect.any(Function),
       },
       currencyBridge: {
         preload: expect.any(Function),


libs/coin-framework/src/bridge/index.ts
```diff
@@ -0,0 +1 @@
+export { validateAddress as baseValidateAddress } from "./validateAddress";
```
francois-guerin-ledger: I think it is time to stop using barrel files (cf [discussion on Slack](https://ledger.slack.com/archives/C06RTDKT324/p1764345978079559))

libs/coin-framework/src/bridge/validateAddress.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import { validateAddress } from "./validateAddress";
+
+describe("validateAddress", () => {
+  it.each([undefined, ""])(
+    'should return false for undefined or empty address ("%s")',
+    async (address: string | undefined) => {
+      const result = await validateAddress(address as unknown as string, {});
+      expect(result).toEqual(false);
+    },
+  );
```
francois-guerin-ledger: This test feels wrong, having to cast parameters to call the function. See my comments below

libs/coin-framework/src/bridge/validateAddress.ts
```diff
@@ -0,0 +1,8 @@
+import { AddressValidationCurrencyParameters } from "@ledgerhq/types-live";
+
+export async function validateAddress(
+  address: string,
+  _parameters: Partial<AddressValidationCurrencyParameters>,
+): Promise<boolean> {
+  return Boolean(address && address.length > 0);
+}
```
francois-guerin-ledger, line 9:
```suggestion
export function validateAddress(address: unknown): boolean {
  return typeof address === "string" && address.length > 0
}
```

libs/coin-modules/coin-evm/src/logic/validateAddress.ts
```diff
@@ -0,0 +1,21 @@
+import { baseValidateAddress } from "@ledgerhq/coin-framework/bridge/index";
+import { AddressValidationCurrencyParameters } from "@ledgerhq/types-live";
+import { ethers } from "ethers";
+
+const ethAddressRegex = /^(0x)?[0-9a-fA-F]{40}$/;
+const ethDomainRegex = /^.+\.eth$/;
```
francois-guerin-ledger:
```suggestion
```
Better use `endsWith` at this stage

libs/coin-modules/coin-evm/src/logic/validateAddress.test.ts
```diff
@@ -0,0 +1,93 @@
+import { getAddress } from "ethers";
+import { baseValidateAddress } from "@ledgerhq/coin-framework/bridge/index";
+import { validateAddress } from "./validateAddress";
+
+jest.mock("ethers");
+jest.mock("@ledgerhq/coin-framework/bridge/index");
```
francois-guerin-ledger: Why all those mocks ? Do these functions call the network ?

libs/coin-framework/src/api/types.ts
```diff
@@ -520,6 +525,10 @@ export type AlpacaApi<
    *         does not provide cursor-based pagination.
    */
   getValidators: (cursor?: Cursor) => Promise<Page<Validator>>;
+  validateAddress(
+    address: string,
+    parameters: Partial<AddressValidationCurrencyParameters>,
+  ): Promise<boolean>;
```
francois-guerin-ledger: Looking at the ticket, we must extend the Bridge API ... not the Alpaca API.

libs/coin-framework/src/bridge/index.ts
Moustafa-Koterba: done

libs/coin-framework/src/bridge/validateAddress.test.ts
Moustafa-Koterba: You're right, I wanted to test `undefined` case but it seems not needed here
I prefer to keep the function signature as 
```ts
function validateAddress(address: string, // etc.
```
Make more sense IMO
Only empty string case has been kept 👍

libs/coin-modules/coin-evm/src/logic/validateAddress.ts
Moustafa-Koterba: done

libs/coin-modules/coin-evm/src/logic/validateAddress.test.ts
Moustafa-Koterba: They are called in the function yes
And no, no network calls

libs/coin-framework/src/bridge/validateAddress.test.ts
```diff
@@ -0,0 +1,16 @@
+import { baseValidateAddress } from "./validateAddress";
+
+describe("baseValidateAddress", () => {
+  it("should return false for empty address", async () => {
+    const result = await baseValidateAddress("", {});
+    expect(result).toEqual(false);
+  });
+
+  it.each(["a", "some random address"])(
```
francois-guerin-ledger: It does not seem relevant to have multiple input because you are testing only one path of your function

libs/coin-framework/src/bridge/validateAddress.ts
```diff
@@ -0,0 +1,10 @@
+import { AddressValidationCurrencyParameters } from "@ledgerhq/types-live";
+
+async function validateAddress(
+  address: string,
+  _parameters: Partial<AddressValidationCurrencyParameters>,
+): Promise<boolean> {
+  return Boolean(address && address.length > 0);
+}
```
francois-guerin-ledger: - `_parameters` in not used
- using the `Boolean` constructor is overkill

```suggestion
async function validateAddress(address: string): boolean {
  return address.length > 0;
}
```

libs/coin-modules/coin-algorand/src/index.ts
```diff
@@ -0,0 +1 @@
+export { validateAddress } from "./validateAddress";
```
francois-guerin-ledger: You should remove those barrel files

libs/coin-modules/coin-algorand/src/validateAddress.test.ts
```diff
@@ -0,0 +1,50 @@
+import { isValidAddress } from "algosdk";
+import { validateAddress } from "./validateAddress";
+import { baseValidateAddress } from "@ledgerhq/coin-framework/bridge/validateAddress";
+
+jest.mock("algosdk");
+jest.mock("@ledgerhq/coin-framework/bridge/validateAddress");
```
francois-guerin-ledger: Why mocking the `validateAddress` from `coin-framework` ? Looking at all these tests (I put the comment here, but this is valid for all coin modules), I feel we are testing how the functions are implemented when we should focus  on inputs and outputs.
We don't care that this `validateAddress` calls a "base validation" and the `isValidAddress` function from the `algosdk` package.

libs/coin-framework/src/bridge/validateAddress.test.ts
Moustafa-Koterba: I find multiple input relevant actually, to test that your code does not depend on it, even for the same path. In this test, I check that it work for different type of input

libs/coin-framework/src/bridge/validateAddress.ts
Moustafa-Koterba: > `_parameters` in not used

named it with underscore for it

> using the Boolean constructor is overkill

fixed

libs/coin-modules/coin-algorand/src/index.ts
Moustafa-Koterba: done

libs/coin-modules/coin-algorand/src/validateAddress.test.ts
Moustafa-Koterba: > Why mocking the validateAddress from coin-framework ? Looking at all these tests (I put the comment here, but this is valid for all coin modules), I feel we are testing how the functions are implemented when we should focus on inputs and outputs.

I mock them to not rely on their implementation. I am testing the behaviour of my function, checking that it take the path I want to test. So I check that it calls correctly the external function I am using into it. This prevent that we use correctly the `algosdk` as intented and also the `validateAddress` from the `coin-framework`. This is a very common way to make unit tests actually.

For the inputs in this function, it actually depends of external functions like from `algosdk`. I dont want to test `algosdk` behaviour, and our tests must not rely on third library. So yeah I mock them.

libs/coin-framework/src/bridge/validateAddress.test.ts
francois-guerin-ledger: I don't understand how `"a"` and `"some random address"` are different type of input. Could you please explain ?

libs/coin-framework/src/bridge/validateAddress.ts
francois-guerin-ledger: > named it with underscore for it

Yes but it is still unused, why not removing it ?

francois-guerin-ledger: Also I forgot to remove `async`, it must be
```ts
function validateAddress(address: string): boolean {
	return address.length > 0;
}
```

libs/coin-modules/coin-algorand/src/index.ts
francois-guerin-ledger: What do you mean "done", the file it still there

libs/coin-modules/coin-algorand/src/validateAddress.test.ts
francois-guerin-ledger: If you mock everything like here, how are parameter values even useful ? Because everything is mocked, any input lead to the same output

libs/coin-framework/src/bridge/validateAddress.ts
```diff
@@ -0,0 +1,10 @@
+import { AddressValidationCurrencyParameters } from "@ledgerhq/types-live";
+
+async function validateAddress(
+  address: string,
+  _parameters: Partial<AddressValidationCurrencyParameters>,
+): Promise<boolean> {
+  return address.length > 0;
+}
```
francois-guerin-ledger: Now that I think of it, why even having this function ? It is a really weak check, pretty sure it is always implicitly covered (I doubt `ethers` finds the empty string a valid address for example)

libs/coin-framework/src/bridge/validateAddress.test.ts
Moustafa-Koterba: `a` and `some random address` have different size, I wanted to show that the size does not matter.

libs/coin-framework/src/bridge/validateAddress.ts
Moustafa-Koterba: It is actually used in [accountBridge](https://github.com/LedgerHQ/ledger-live/pull/12967/files#diff-60454f79a87542a564c43ff8654f2c987613e417be6b02c975e28d1f0373e54a), I prefer to keep the same signature since it is defined like this in the bridge
```ts
validateAddress: (
    address: string,
    parameters: Partial<AddressValidationCurrencyParameters>,
  ) => Promise<boolean>;
```

libs/coin-modules/coin-aptos/src/logic/index.ts
```diff
@@ -0,0 +1 @@
+export { validateAddress } from "./validateAddress";
```
francois-guerin-ledger: No need to add this file

libs/coin-modules/coin-bitcoin/src/index.ts
```diff
@@ -0,0 +1 @@
+export { validateAddress } from "./validateAddress";
```
francois-guerin-ledger: No need to add this file

libs/coin-modules/coin-cosmos/src/validateAddress.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  if (!parameters.currency) {
+    throw new Error("Missing currency parameter on address validation for Cosmos");
+  }
+
+  let isValid = true;
+  try {
+    bech32.decode(address);
+  } catch (_) {
```
francois-guerin-ledger:
```suggestion
  } catch {
```

libs/coin-modules/coin-cardano/src/validateAddress.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  it("should throw an error when currency parameters miss networkId", async () => {
+    mockedBaseValidateAddress.mockResolvedValueOnce(true);
+
+    try {
+      await validateAddress("some random address", {});
+    } catch (error: unknown) {
+      if (!(error instanceof Error)) {
+        throw error;
+      }
```
francois-guerin-ledger: `try ... catch` must be avoided in tests. We must use the `toThrow` matcher

libs/coin-modules/coin-algorand/src/index.ts
Moustafa-Koterba: Sorry I do not see the file actually 
Screenshot 2025-12-03 at 10 47 40
You may have check before my push

libs/coin-modules/coin-algorand/src/validateAddress.test.ts
Moustafa-Koterba: I am actually testing the function logic (the flow), not `baseValidateAddress` nor `algosdk`. I mock these dependencies to isolate my test and check the behaviour. This allow me to control the execution of the function (which path to take).

I agree that mocking can hide real behaviour, but here I test the execution flow only. I dont want to verify that `algosdk` validate correctly algorand address.

A middle ground could be to add an integration tests that use really `algosdk` to validate address formats, or an E2E test that check in the send flow the recipient input.

libs/coin-framework/src/bridge/validateAddress.ts
Moustafa-Koterba: That's true, I added this to have a common validation for all coins, to not repeat the same code 
I did not check each coin library to see if it handles the case (it should be)

libs/coin-modules/coin-algorand/src/validateAddress.test.ts
francois-guerin-ledger: > I dont want to verify that `algosdk` validate correctly algorand address.

This, I agree. But me, I even don't want to know that `algosdk` is used. I don't want to be aware of any external dependencies, really.
I prefer a simple "with this input, `validateAddress` of `coin-algorand` returns this output". Simply call it with a valid address and an invalid address, ensure `true` and `false` are returned respectively.

> A middle ground could be to add an integration tests

I won't call it integration test, but we clearly need tests that don't hide anything

libs/coin-framework/src/bridge/validateAddress.ts
francois-guerin-ledger: I would be more comfortable removing it completly

libs/coin-modules/coin-cardano/src/validateAddress.test.ts
Moustafa-Koterba: done

libs/coin-modules/coin-cosmos/src/validateAddress.ts
Moustafa-Koterba: done

libs/coin-modules/coin-algorand/src/validateAddress.test.ts
Moustafa-Koterba: Got it. I’d like to keep the current tests, but I can add new ones with real addresses and no mocking—just inputs and outputs. That way we have both. Works for you?

francois-guerin-ledger: Yes, prefect 👌

libs/coin-framework/src/bridge/validateAddress.ts
Moustafa-Koterba: done

libs/coin-modules/coin-aptos/src/logic/index.ts
Moustafa-Koterba: done

libs/coin-modules/coin-bitcoin/src/index.ts
Moustafa-Koterba: done

libs/ledger-live-common/src/bridge/validateAddress.ts
```diff
@@ -0,0 +1,13 @@
+import { AddressValidationCurrencyParameters } from "@ledgerhq/types-live";
+
+export const isInvalidRecipient = (recipient: string) => {
+  if (recipient.includes("criticalcrash")) throw new Error("isInvalidRecipient_mock_criticalcrash");
```
francois-guerin-ledger, line 4: I don't understand, what is this hardcoded `"criticalcrash"` ?

libs/ledger-live-common/src/bridge/generic-alpaca/registerValidateAddress.ts
```diff
@@ -0,0 +1,10 @@
+import { registerValidateFunction } from "./getValidateAddressByCurrency";
+import { validateAddress as tezosValidateAddress } from "@ledgerhq/coin-tezos/logic/validateAddress";
+import { validateAddress as xrpValidateAddress } from "@ledgerhq/coin-xrp/logic/validateAddress";
+import { validateAddress as stellarValidateAddress } from "@ledgerhq/coin-stellar/logic/validateAddress";
+import { validateAddress as evmValidateAddress } from "@ledgerhq/coin-evm/logic/validateAddress";
+
+registerValidateFunction("stellar", stellarValidateAddress);
+registerValidateFunction("xrp", xrpValidateAddress);
+registerValidateFunction("tezoz", tezosValidateAddress);
+registerValidateFunction("evm", evmValidateAddress);
```
francois-guerin-ledger: Instead of this registration mechanism, how about homogenising with what we have in `libs/ledger-live-common/src/bridge/generic-alpaca/signer/index.ts` ?

libs/ledgerjs/packages/types-live/src/bridge.ts
```diff
@@ -212,6 +217,10 @@ interface SendReceiveAccountBridge<
   // broadcasting a signed transaction to network
   // returns an optimistic Operation that this transaction is likely to create in the future
   broadcast: BroadcastFnSignature<A>;
+  validateAddress: (
```
gre-ledger, line 220: could you document in js a bit the intention / what it does / where/when will this be used from the UI side?

as the bridge is the abstraction between the UI and the coins, what is weird to me is that in this PR i don't see any usage on the hub codeowning files other than this type, is it because it's not used by UI side yet? 🤔 

[EDIT] oh yeah, the PR description mention it already 👍

francois-guerin-ledger:
Also can we uniform path ? Depending on coin modules, we have
- `src/validateAddress.ts`
- `src/logic/validateAddress.ts`
- `src/bridge/validateAddress.ts`

Ideally it is `src/logic/validateAddress.ts`, because `src/bridge` is candidate for deletion and because we want to use this function is other contexts as well

ComradeAERGO:
> Also can we uniform path ? Depending on coin modules, we have
> 
> * `src/validateAddress.ts`
> * `src/logic/validateAddress.ts`
> * `src/bridge/validateAddress.ts`
> 
> Ideally it is `src/logic/validateAddress.ts`, because `src/bridge` is candidate for deletion and because we want to use this function is other contexts as well

@francois-guerin-ledger This is a valid point but we should tackle it in a separate task. I think we've spent enough time on this specific PR and quality seems good enough at this point.

libs/ledger-live-common/src/bridge/validateAddress.ts
Moustafa-Koterba: This function was taken from `mockHelpers` file
This validateAddress is only used for mocks actually
Screenshot 2025-12-08 at 11 24 14

libs/ledger-live-common/src/bridge/generic-alpaca/registerValidateAddress.ts
Moustafa-Koterba: done

jnicoulaud-ledger (approved):

libs/coin-modules/coin-cosmos/src/validateAddress.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  it("should throw an error when currency parameters is not provided", async () => {
+    try {
+      await validateAddress("some random address", {});
+    } catch (error) {
+      expect(error).toBeInstanceOf(Error);
+      expect((error as Error).message).toEqual(
+        "Missing currency parameter on address validation for Cosmos",
+      );
+    }
```
francois-guerin-ledger, line 31: We must use `await expect(...).rejects.toThrow("...");`

libs/ledger-live-common/.unimportedrc.json
```diff
@@ -315,9 +315,15 @@
     "src/modularDrawer/types/detailedAccount.ts",
     "src/modularDrawer/utils/sortAccountsByFiatValue.ts",
     "src/modularDrawer/index.ts",
-    "src/test-helpers/cryptoAssetsStore.ts"
+    "src/test-helpers/cryptoAssetsStore.ts",
+    "src/bridge/generic-alpaca/registerValidateAddress.ts"
```
francois-guerin-ledger, line 319:
```suggestion
```
This file does not exist

libs/ledger-live-common/src/bridge/generic-alpaca/signer/index.ts
```diff
@@ -107,3 +112,27 @@ export function getSigner(network: string): AlpacaSigner {
...
+
+const validateAddressByNetwork = new Map<string, ValidateAddressFunction>();
+validateAddressByNetwork.set("stellar", stellarValidateAddress);
+validateAddressByNetwork.set("xrp", xrpValidateAddress);
+validateAddressByNetwork.set("tezos", tezosValidateAddress);
+validateAddressByNetwork.set("evm", evmValidateAddress);
+validateAddressByNetwork.set("ethereum", evmValidateAddress); // for the coin tester
+validateAddressByNetwork.set("sonic", evmValidateAddress); // for the coin tester
+validateAddressByNetwork.set("polygon", evmValidateAddress); // for the coin tester
+validateAddressByNetwork.set("core", evmValidateAddress); // for the coin tester
```
francois-guerin-ledger, line 129: So we still use a `Map` instead of `switch .. case` to be homogeneous with `generic-alpaca/signer/index.ts` 😞 . I would have appreciated we discuss it instead muting this issue.
I don't understand `for the coin tester` comments. Does the coin tester use this bridge method ?

libs/ledger-live-common/src/bridge/mockHelpers.ts
```diff
@@ -110,10 +110,8 @@ export const signOperation: AccountBridge<any>["signOperation"] = ({ account, tr
       cancelled = true;
     };
   });
-export const isInvalidRecipient = (recipient: string) => {
-  if (recipient.includes("criticalcrash")) throw new Error("isInvalidRecipient_mock_criticalcrash");
-  return recipient.includes("invalid") || recipient.length <= 3;
-};
+
+export { isInvalidRecipient } from "./validateAddress";
```
francois-guerin-ledger, line 114: Strange to have a re-export in the middle of a file

PR #12970, opened 2025-11-25
feat: staking support in coin-hedera (LIVE-19453)

mdomanski-ext-ledger:
### ✅ Checklist

  - staking should be possible with Hedera (delegate, undelegate, redelegate, claim rewards)

### 📝 Description

This PR adds Staking support to coin-hedera module in LWD and LWM.

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-19453

---

### DIFF
--- a/libs/coin-modules/coin-hedera/src/deviceTransactionConfig.test.ts
+++ b/libs/coin-modules/coin-hedera/src/deviceTransactionConfig.test.ts
@@ -20,7 +20,7 @@ describe("getDeviceTransactionConfig", () => {
 
   describe("staking transactions", () => {
     it("should return correct fields for ClaimRewards transaction", async () => {
-      const transaction: Transaction = {
+      const transaction = {
         family: "hedera",
         mode: HEDERA_TRANSACTION_MODES.ClaimRewards,
         amount: new BigNumber(0),
@@ -36,12 +36,15 @@ describe("getDeviceTransactionConfig", () => {
         status,
       });
 
-      expect(fields).toHaveLength(2);
-      expect(fields).toMatchObject([
+      expect(fields).toEqual([
         {
           type: "text",
           label: "Method",
           value: "Claim Rewards",
+        },
+        {
+          label: "Fees",
+          type: "fees",
         },
         {
           type: "text",
@@ -52,7 +55,7 @@ describe("getDeviceTransactionConfig", () => {
     });
 
     it("should return correct fields for Delegate transaction", async () => {
-      const transaction: Transaction = {
+      const transaction = {
         family: "hedera",
         mode: HEDERA_TRANSACTION_MODES.Delegate,
         amount: new BigNumber(0),
@@ -70,12 +73,15 @@ describe("getDeviceTransactionConfig", () => {
         status,
       });
 
-      expect(fields).toHaveLength(2);
-      expect(fields).toMatchObject([
+      expect(fields).toEqual([
         {
           type: "text",
           label: "Method",
           value: "Update Account",
+        },
+        {
+          label: "Fees",
+          type: "fees",
         },
         {
           type: "text",
@@ -86,7 +92,7 @@ describe("getDeviceTransactionConfig", () => {
     });
 
     it("should not include staking node ID if not provided", async () => {
-      const transaction: Transaction = {
+      const transaction = {
         family: "hedera",
         mode: HEDERA_TRANSACTION_MODES.Undelegate,
         amount: new BigNumber(0),
@@ -102,12 +108,15 @@ describe("getDeviceTransactionConfig", () => {
         status,
       });
 
-      expect(fields).toHaveLength(1);
-      expect(fields).toMatchObject([
+      expect(fields).toEqual([
         {
           type: "text",
           label: "Method",
           value: "Update Account",
+        },
+        {
+          label: "Fees",
+          type: "fees",
         },
       ]);
     });
@@ -115,7 +124,7 @@ describe("getDeviceTransactionConfig", () => {
 
   describe("token associate transactions", () => {
     it("should return correct fields for TokenAssociate transaction", async () => {
-      const transaction: Transaction = {
+      const transaction = {
         family: "hedera",
         mode: HEDERA_TRANSACTION_MODES.TokenAssociate,
         amount: new BigNumber(0),
@@ -132,8 +141,7 @@ describe("getDeviceTransactionConfig", () => {
         status,
       });
 
-      expect(fields).toHaveLength(3);
-      expect(fields).toMatchObject([
+      expect(fields).toEqual([
         {
           type: "text",
           label: "Method",
@@ -152,7 +160,7 @@ describe("getDeviceTransactionConfig", () => {
     });
 
     it("should not include fees if they are zero", async () => {
-      const transaction: Transaction = {
+      const transaction = {
         family: "hedera",
         mode: HEDERA_TRANSACTION_MODES.TokenAssociate,
         amount: new BigNumber(0),
@@ -168,8 +176,7 @@ describe("getDeviceTransactionConfig", () => {
         status,
       });
 
-      expect(fields).toHaveLength(1);
-      expect(fields).toMatchObject([
+      expect(fields).toEqual([
         {
           type: "text",
           label: "Method",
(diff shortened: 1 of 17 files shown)


libs/coin-modules/coin-hedera/src/transaction.ts
```diff
@@ -40,6 +40,27 @@ export function fromTransactionRaw(tr: TransactionRaw): Transaction {
     };
   }
 
+  if (
+    tr.mode === HEDERA_TRANSACTION_MODES.Delegate ||
+    tr.mode === HEDERA_TRANSACTION_MODES.Undelegate ||
+    tr.mode === HEDERA_TRANSACTION_MODES.Redelegate
```
semeano, line 46: I've seen this pattern repeated a few times. You could just set a new type that takes all the staking modes and then `HEDERA_STAKING_MODES.includes(tr.mode)`.
Nothing major, just an idea.

mdomanski-ext-ledger: Thanks for reviewing 🙌 

I was thinking about the same thing, but TypeScript's type narrowing doesn't work well with `.includes()` for union types. The compiler can't infer that `txIntent.type` is one of the staking modes after the check. I would have to add something like:

```
export function isStakingMode(
  mode: string
): mode is 
  | typeof HEDERA_TRANSACTION_MODES.Delegate 
  | typeof HEDERA_TRANSACTION_MODES.Undelegate 
  | typeof HEDERA_TRANSACTION_MODES.Redelegate {
  return [
    HEDERA_TRANSACTION_MODES.Delegate,
    HEDERA_TRANSACTION_MODES.Undelegate,
    HEDERA_TRANSACTION_MODES.Redelegate,
  ].includes(mode);
}
```

so I thought it's simpler and more consistent with explicit conditions. Especially that `HEDERA_TRANSACTION_MODES.ClaimRewards` would be missing here, even if it is a staking mode. It just needs different handling (rewards are collected automatically with Hedera, but we need to make transaction and CryptoTransfer is cheaper than CryptoUpdate).

semeano (approved):

libs/coin-modules/coin-hedera/src/bridge/buildOptimisticOperation.integration.test.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { getMockedTransaction } from "../test/fixtures/transaction.fixture";
 
 describe("buildOptimisticOperation", () => {
-  test("builds optimistic operation for token association", async () => {
+  it("builds optimistic operation for token association", async () => {
```
francois-guerin-ledger, line 13: Why does this filename ends with `.integration.test.ts`, it does not call the network ?

libs/coin-modules/coin-hedera/src/deviceTransactionConfig.test.ts
```diff
@@ -0,0 +1,310 @@
...
+      const status = createMockStatus(new BigNumber(100000));
+
+      const fields = await getDeviceTransactionConfig({
+        account: mockAccount,
+        transaction,
+        status,
+      });
+
+      expect(fields).toHaveLength(2);
+      expect(fields).toMatchObject([
```
francois-guerin-ledger:
```suggestion
      expect(fields).toEqual([
```
Tip for the whole file: no need to test the length of the array if you also test it is equal or even partially equal to another array of your expected length

francois-guerin-ledger: And while comparing arrays, `toEqual` is the way to go. Items _inside_ the array may require partial equality

libs/coin-modules/coin-hedera/src/logic/listOperations.test.ts
```diff
@@ -393,4 +401,62 @@ describe("listOperations", () => {
...
+      useSyntheticBlocks: false,
+    });
+
+    const rewardTimestamp = result.coinOperations[0].date.getTime();
+    const mainTimestamp = result.coinOperations[1].date.getTime();
+
+    expect(result.coinOperations).toHaveLength(2);
+    expect(result.tokenOperations).toHaveLength(0);
+    expect(rewardTimestamp).toBe(mainTimestamp + 1);
+    expect(result.coinOperations).toMatchObject([
```
francois-guerin-ledger:
```suggestion
    expect(result.tokenOperations).toEqual([]);
    expect(rewardTimestamp).toBe(mainTimestamp + 1);
    expect(result.coinOperations).toEqual([
```

libs/coin-modules/coin-hedera/src/deviceTransactionConfig.test.ts
mdomanski-ext-ledger: > no need to test the length of the array if you also test it is equal or even partially equal to another array of your expected length

Sure thing - I also added an internal note to review the rest of `coin-hedera` for similar cases.

> And while comparing arrays, toEqual is the way to go. Items inside the array may require partial equality

I didn't want to provide every field for each operation, so I went with `.toMatchObject()` for partial comparison. Switching to `.toEqual()` without listing all values would probably end up looking like:

```
expect(result).toEqual(
  expect.arrayContaining([
    expect.objectContaining({ ... }),
    expect.objectContaining({ ... })
  ])
);
```

`.toMatchObject()` just feels cleaner for this particular case, but I'm happy to adjust if there is a better/preferable approach

libs/coin-modules/coin-hedera/src/bridge/buildOptimisticOperation.integration.test.ts
mdomanski-ext-ledger: Network call happens on `estimateFees` level: https://github.com/LedgerHQ/ledger-live/blob/976c709a7df22d9527ccd9d7fd31f6d45f6833ad/libs/coin-modules/coin-hedera/src/bridge/buildOptimisticOperation.ts#L154

francois-guerin-ledger: Is it expected to estimate fees that late in the flow ? Signing already occurred!

francois-guerin-ledger (approved):

apps/ledger-live-mobile/src/screens/Account/ListHeaderComponent.tsx
```diff
@@ -120,6 +121,7 @@ export function useListHeaderComponents({
         CosmosAccount &
         PolkadotAccount &
         MultiversXAccount &
+        HederaAccount &
         NearAccount,
```
thesan, line 125: If I understand correctly sonarCloud says that the `as ...` which would be nice if it's the case

mdomanski-ext-ledger: Unfortunately it's a false positive :/ 

Without any casting:
```
Type 'AccountLike' is not assignable to type 'Account & { celoResources: CeloResources; } & { cosmosResources: CosmosResources; } & { nearResources: NearResources; } & { ...; } & { ...; }'.
  Type 'TokenAccount' is not assignable to type 'Account & { celoResources: CeloResources; } & { cosmosResources: CosmosResources; } & { nearResources: NearResources; } & { ...; } & { ...; }'.
    Type 'TokenAccount' is missing the following properties from type 'Account': seedIdentifier, derivationMode, index, freshAddress, and 5 more.ts(2322
```

With `as Account` only:
```
Type 'Account' is not assignable to type 'Account & { celoResources: CeloResources; } & { cosmosResources: CosmosResources; } & { nearResources: NearResources; } & { ...; } & { ...; }'.
  Property 'celoResources' is missing in type 'Account' but required in type '{ celoResources: CeloResources; }'.ts(2322)
```

Switching to a less strict type with guards inside `AccountBalanceSummaryFooter` would probably be enough, but that's something better handled in a separate PR

libs/coin-modules/coin-hedera/src/bridge/buildOptimisticOperation.integration.test.ts
mdomanski-ext-ledger: Small note regarding fee estimation on this level, based on our thread on Slack:

> It's just the old logic that was copied & pasted over time: https://github.com/LedgerHQ/ledger-live/blob/%40ledgerhq/live-desktop%402.100.0/libs/coin-modules/coin-hedera/src/bridge/buildOptimisticOperation.ts. `tx.maxFee ?? new BigNumber(0)` should be used here, but it can be handled with separate PR.

mateuszpalosz-ext (approved):

PR #12973, opened 2025-11-25
feat(NotificationCenter): add HtmlTextRenderer for rendering HTML lin…

deepyjr:
…ks in notifications

### ✅ Checklist

THE NOTIFICATION IS NOT AVAILABLE ANYMORE. ONLY THE STATUS IS LEFT 

  - I've changed the status to make a clickable link inside, for example : 'Description avec seulement <a href="https://ledger.com">lien HTML</a> dans le texte, sans CTA ni link dans l\'objet.'
  
  So the html element will be clickable, you can have multiple links by the way.
- I've used the function created in this ticket to determine if the URL is safe or not, and I've put it in the common lib to have the same function for mobile and desktop

### 📝 Description
I've created functions that check for a link inside, and I've added tests for French and Russian, as it was found on a Russian device.

The data on the demo is totally fake and generated.

WITH GOOD DATA MOBILE AND DESKTOP 👇 

https://github.com/user-attachments/assets/5e33d515-6f3e-488b-a665-b03eedf574c1

https://github.com/user-attachments/assets/f7e3e778-e7e1-4f61-bcfa-9ff3fd53fd61

WITH BAD DATA MOBILE AND DESKTOP 👇 ( example redirection to google )

https://github.com/user-attachments/assets/ad7abf56-b0ae-4dde-863e-79b788b0c609

### ❓ Context

- **JIRA or GitHub [22802](https://ledgerhq.atlassian.net/browse/LIVE-22802) 

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/TranslatedError/ErrorWithAnchor.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/TranslatedError/ErrorWithAnchor.tsx
@@ -15,7 +15,11 @@ export function ErrorWithAnchorContent({
   html,
   dataTestId,
 }: ErrorWithAnchorContentProps): JSX.Element {
-  const { segments } = useHtmlLinkSegments(html);
+  const { segments }: { segments: HtmlLinkSegment[] } = useHtmlLinkSegments(html);
+
+  const handleLinkClick = (href: string) => {
+    openURL(href);
+  };
 
   return (
     <span data-testid={dataTestId}>
@@ -26,7 +30,7 @@ export function ErrorWithAnchorContent({
             color="palette.text.warning"
             alwaysUnderline
             key={uuid}
-            onClick={() => openURL(segment.href)}
+            onClick={() => handleLinkClick(segment.href)}
           >
             {segment.label}
           </Link>


apps/ledger-live-mobile/src/screens/NotificationCenter/Notifications.tsx
LucasWerey, line 1: Can be removed as it's handled via braze and they use deeplinks and not link directly

deepyjr: It's deleted now! 🚀

apps/ledger-live-desktop/src/renderer/components/TranslatedError/ErrorWithAnchor.tsx
```diff
@@ -1,55 +1,39 @@
...
+  return (
+    <span data-testid={dataTestId}>
+      {segments.map((segment: HtmlLinkSegment) => {
+        const uuid = uniqueId();
+        return segment.type === "link" ? (
+          <Link
+            color="palette.text.warning"
+            alwaysUnderline
+            key={uuid}
+            onClick={() => openURL(segment.href)}
```
LucasWerey: Can you do a proper function please? To avoid arrow functions like this it's better for perf (even it's minor)

apps/ledger-live-mobile/src/screens/NotificationCenter/HtmlTextRenderer.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <View>
+      <Text
+        style={{
+          fontSize: 14,
+          lineHeight: 20,
+          color: textColor,
+        }}
+        numberOfLines={numberOfLines}
+      >
+        {segments.map((segment: HtmlLinkSegment, index: number) => {
```
LucasWerey, line 68: Isn't there an ID directly in each segment? To use it instead of uniqueId

deepyjr: No, there is not, because it's different parts of the text. I could add one, but it will be generated too. Would you like me to do that?

apps/ledger-live-desktop/src/renderer/components/TranslatedError/ErrorWithAnchor.tsx
deepyjr: I've done the fix 🙏

PR #13010, opened 2025-11-26
[LIVE-23713] Fix SelectDevice Android permissions issues due to RN issue

OlivierFreyssinet:
### ✅ Checklist

  - Android SelectDevice component, BLE pairing flow

### 📝 Description

[Due to an issue in React Native on Android](https://github.com/facebook/react-native/issues/53887), if the user has already denied a permission, in scenarios where a call to `requestPermission` would normally resolve to `NEVER_ASK_AGAIN`, the same call now _never resolves_.

It means two things for Ledger Wallet:
- In components requiring a BLE connection, we were relying on that request call to get the current status of each permission. Now there's a risk that it does not resolve, so instead, we first call `checkPermission` in order to ensure that the state of the permissions is always determined.
- Some "permission not granted" components were offering a retry logic (which would call `requestPermission`). Now, as this call is unreliable, for lack of a better solution, in all cases we will instead redirect the user to the app's settings where he can manually change the permission.

NB: a fix is included in React Native 0.81.5, we will revert this behaviour once we upgrade to it.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-23713] 

---

[LIVE-23713]: https://ledgerhq.atlassian.net/browse/LIVE-23713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/wicked-melons-wait.md
+++ b/.changeset/wicked-melons-wait.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Fix Android permission issues in SelectDevice


apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBluetoothPermissions.ts
```diff
@@ -175,12 +175,21 @@ export const useAndroidBluetoothPermissions = (
...
+       * or the permission state might be stuck in "unknown" state.
+       */
+      const checkResult = await checkBluetoothPermissions();
+      if (cancelled) return;
+      setHasPermissions(checkResult ? "granted" : "denied");
+      if (checkResult) return;
+
+      const requestResult = await requestBluetoothPermissions();
 
       if (!cancelled) {
```
jiyuzhuang, line 189: I think this part of code can persist after the fix of RN, so it's better to reorganise them.
```suggestion
      if (cancelled) {
        return;
      } else {
        const checkResult = await checkBluetoothPermissions();
        if (checkResult) {
          setHasPermissions("granted");
          return;
        }
        const requestResult = await requestBluetoothPermissions();
```

jiyuzhuang (approved):

apps/ledger-live-mobile/src/components/RequiresBLE/hooks/useAndroidBluetoothPermissions.ts
OlivierFreyssinet: Yes this part of the code can stay like this 👍
No this refacto is not valid because we want to always use the result of `checkBluetoothPermissions` to update the state, in case `requestBluetoothPermissions` never resolves.

For instance: if a permission is not granted and it's in a "never ask again" status, then with the RN bug, `requestBluetoothPermissions` will never resolve, so we still want to update the state to "denied".

jiyuzhuang: Ok, I understand, so you can just put `setHasPermissions("denied");` in line 189 or in an else scope the `checkResult` if condition.

PR #13012, opened 2025-11-27
fix(AddAccountDrawer): The error was too explicit for the user when you had a crash on the add account

deepyjr:
…owManager

### ✅ Checklist

  - I've updated the error on the modular asset to not show the content of the error to a user (I think it was not happening frequently because I couldn't reproduce it). I added a dev tool that can emulate a 500 from a call api on the MAD to see if it was possible, and same with that I had to force it to reproduce it.

### 📝 Description

Update the error to show the user only the fact that he had an error, and not all the text of the error

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-23475

---

### DIFF
--- a/apps/ledger-live-desktop/src/mocks/init.ts
+++ b/apps/ledger-live-desktop/src/mocks/init.ts
@@ -1,7 +1,7 @@
 async function enableMocking() {
   const mswEnabled = process.env.ENABLE_MSW === "true";
 
-  if (mswEnabled) {
+  if (mswEnabled && __DEV__) {
     try {
       const { startWorker, toggleSimulate500Error } = await import("~/mocks/browser");
 
--- a/apps/ledger-live-desktop/src/newArch/features/ModularDrawer/screens/AssetSelection/index.tsx
+++ b/apps/ledger-live-desktop/src/newArch/features/ModularDrawer/screens/AssetSelection/index.tsx
@@ -18,7 +18,7 @@ export type AssetSelectionStepProps = {
   onAssetSelected: (asset: CryptoOrTokenCurrency) => void;
   loadNext?: () => void;
   error?: boolean;
-  refetch?: () => void;
+  refetch: () => void;
   assetsSorted?: AssetData[];
 };
 


apps/ledger-live-desktop/src/newArch/features/ModularDrawer/screens/AssetSelection/index.tsx
```diff
@@ -58,8 +58,8 @@ const AssetSelection = ({
         />
       )}
       <SearchInputContainer />
-      {error && refetch ? (
-        <GenericError onClick={refetch} />
+      {error ? (
+        <GenericError onClick={refetch || (() => {})} />
```
ejaycoleman: onClick is optional, i'm not sure the refetch button should appear when it doesn't do anything? 

```suggestion
        <GenericError onClick={refetch} />
```

apps/ledger-live-desktop/src/mocks/init.ts
```diff
@@ -3,9 +3,22 @@ async function enableMocking() {
 
   if (mswEnabled) {
     try {
-      const { startWorker } = await import("~/mocks/browser");
+      const { startWorker, toggleSimulate500Error } = await import("~/mocks/browser");
 
       await startWorker();
+
+      // Expose toggle function globally for debugging
+      if (__DEV__) {
```
ejaycoleman, line 11: The project setup means ENABLE_MSW implies dev mode is enabled (`"start:msw": "cross-env NODE_ENV=development ENABLE_MSW=true node ./tools/main.js",`)

Maybe it would make more sense if line 4 was `if (mswEnabled && __DEV__) { ...`?

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/hooks/useModularDrawerRemoteData.ts
```diff
@@ -60,7 +60,7 @@ export function useModularDrawerRemoteData({
   });
 
   return {
-    error,
+    error: !!error, // Convert to boolean to avoid exposing error details (HTML from 500 responses)
```
ejaycoleman, line 63: Sorry I think I'm missing something - the usage of `useModularDrawerRemoteData` in `ModularDrawerFlowManager` passes error into `AssetSelection` like `error={!!error}` already doesn't it? How does this fix the issue?

apps/ledger-live-desktop/src/mocks/init.ts
deepyjr: I agree, I will update the file! 👍

apps/ledger-live-desktop/src/newArch/features/ModularDrawer/hooks/useModularDrawerRemoteData.ts
deepyjr: For me, the type of error in this file still was: `error: SerializedError | undefined`, which could contain the message. That's why I've presented it in that way. In the previous case, it could contain a message due to the SerializedError, and I suppose that was what was shown to the user.

But to be honest, as it was pretty hard to reproduce, I've tried to resolve that part with a simulation of the 500 that you had with Cloudflare. The fix that I've done work to not show the content of the error.

apps/ledger-live-desktop/src/mocks/init.ts
```diff
@@ -1,11 +1,24 @@
 async function enableMocking() {
   const mswEnabled = process.env.ENABLE_MSW === "true";
 
-  if (mswEnabled) {
+  if (mswEnabled && __DEV__) {
     try {
-      const { startWorker } = await import("~/mocks/browser");
+      const { startWorker, toggleSimulate500Error } = await import("~/mocks/browser");
```
gre-ledger, line 6: do i read it right but i suspect it means this code will land in the bundle in the current way we do bundle. if so, i think we need to consider a different approach to avoid having the cloudflare500Html big string ending up in production bundle.

deepyjr: @gre-ledger I don't remember what I've done on this task tbh, I will restart it. 

I just remember that I couldn't reproduce it because it was during the cloud flare crash. Also i think as we are not using the modular drawer anymore, it will be different now

PR #13013, opened 2025-11-27
fix(NoFunds): handle missing account in receive navigation and improv…

deepyjr:
…e token account creation logic

### ✅ Checklist

  - The impact is on the earn section, it could be good to try to stack with funds and without then try to receive 

### 📝 Description
When I attempted to earn on a cryptocurrency that I didn't have in an account I already had, it rendered an empty page. It was due to an empty object for the account.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-23099

---

### DIFF
--- a/apps/ledger-live-mobile/src/components/NoFunds/NoFunds.tsx
+++ b/apps/ledger-live-mobile/src/components/NoFunds/NoFunds.tsx
@@ -55,7 +55,7 @@ type ButtonItem = {
 export default function NoFunds({ route }: Readonly<Props>) {
   const { t } = useTranslation();
   const { data: currenciesAll } = useFetchCurrencyAll();
-  const { account, parentAccount, entryPoint } = route.params ?? {};
+  const { account, parentAccount, entryPoint } = route.params;
   const navigation = useNavigation();
   const currency = getAccountCurrency(account);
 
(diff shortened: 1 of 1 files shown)


apps/ledger-live-mobile/src/components/NoFunds/NoFunds.tsx
```diff
@@ -83,19 +83,31 @@ export default function NoFunds({ route }: Readonly<Props>) {
   );
 
   const onReceiveFunds = useCallback(() => {
+    if (!account) {
+      console.warn("No account provided to NoFunds, skipping receive navigation");
```
ejaycoleman: I couldn't actually reproduce this bug, but is the cause `account` being undefined? TS implies it can never be undefined so wondered if you actually needed this guard (if so maybe the console.warn shouldn't be here?)

deepyjr: I think it can be undefined because of 
`const { account, parentAccount, entryPoint } = route.params ?? {};` but you are right, it should be earlier if there is this case.

I will try to run this component without the `?? {}` and I tell you if it's working!

deepyjr: I tried it, and it's working! I deleted the ?? {} and the guard too.

Thanks for the return!

PR #13091, opened 2025-12-02
feat: add descriptors + supported features for each coin

dilaouid:
### ✅ Checklist

  - New declarative feature and descriptor system for `coin-modules` based on [ADR-003](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/6125551933/Coin+modules+-+ADR-003+-+Features+support) and [ADR-012](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/6540492802/Coin+modules+-+ADR-012+-+Feature+descriptor)
  - No impact on existing UI (preparation for new Send flow)
  - **QA**: Verify that feature declarations (`supportedFeatures`) and descriptors (`memo`, `fees`, `recipient`) match actual capabilities of each coin

### 📝 Description

Inspired by this architecture: https://github.com/LedgerHQ/ledger-live/compare/develop...poc/ui-form
(with a different format, since there is no use for the `react-hook-form`
Implements a declarative feature and UI descriptor system for coin modules, following ADR-003 (Features support) and ADR-012 (Feature descriptor).

1. **Feature declarations** (`coin-modules/{coin}/src/supportedFeatures.ts`):
   - Each coin-module declares its supported features (`blockchain_txs`, `staking_txs`)
   - Features include their supported intents (`send`, `delegate`, `claimReward`)

2. **UI descriptors** (`ledger-live-common/src/families/{coin}/descriptor.ts`):
   - Defines the structure of transaction inputs (recipient, amount, memo, fees)
   - Describes field types, constraints, and options for each coin

3. **Feature status in configs** (`ledger-live-common/src/families/{coin}/config.ts`):
   - Features are declared in LiveConfig with their status (`active`/`inactive`)
   - Can be toggled dynamically via Firebase

**Examples:**
```ts
// coin-bitcoin/src/supportedFeatures.ts
export const supportedFeatures: SupportedFeatures = {
  blockchain_txs: ["send"],
};

// ledger-live-common/src/families/bitcoin/descriptor.ts
export const descriptor: CoinDescriptor = {
  send: {
    inputs: {
      recipientSupportsDomain: false,
    },
    fees: {
      hasPresets: true,
      hasCustom: true,
      hasCoinControl: true,
    },
  },
};

// ledger-live-common/src/families/bitcoin/config.ts
status: {
  type: "active",
  features: [
    { id: "blockchain_txs", type: "active" },
  ],
}
```

### ❓ Context

- **[JIRA or GitHub link](https://ledgerhq.atlassian.net/jira/software/c/projects/LIVE/boards/465?selectedIssue=LIVE-23887)**: 

---


libs/ledger-live-common/src/families/algorand/uiConfig.ts
```diff
@@ -0,0 +1,2 @@
+// Encapsulate for LWD & LWM
+export * from "@ledgerhq/coin-algorand/uiConfig";
```
francois-guerin-ledger: If we are dealing with configs, why not simply having those inside `ledger-live-common` instead ? Just like
```
libs/ledger-live-common/src/families/{family}/config.ts
```

francois-guerin-ledger (commented):
We need to check if this PR is aligned with https://ledgerhq.atlassian.net/browse/LIVE-19541 in order not to do do the work twice

libs/ledger-live-common/src/families/algorand/uiConfig.ts
dilaouid: The goal is to have the UI driven by the coin-modules itself
The config file in 
```
libs/ledger-live-common/src/families/{family}/config.ts
```

is a complete file that define a config, and not simply an encapsulation, it think it will be weird to simply add an export here, wdyt? 🤔

francois-guerin-ledger: I don't understand the "encapsulation" argument. In `libs/ledger-live-common/src/families/{family}/config.ts` we export configs, correct ? And in `libs/ledger-live-common/src/families/{family}/uiConfig.ts`, don't we export configs as well ?

dilaouid: The `config.ts` file is related to LiveConfig schemas, while the `uiConfig` is only related to the UI configuration
Those are both config but with different responsabilities, when importing the `config.ts` in `libs/ledger-live-common/src/config/sharedConfig.ts` we don't need to import the uiConfig as well

Moustafa-Koterba (approved):

apps/ledger-live-desktop/src/newArch/utils/getDescriptor.ts
```diff
@@ -0,0 +1,204 @@
...
+  zencash: bitcoinDescriptor,
+  decred: bitcoinDescriptor,
+  digibyte: bitcoinDescriptor,
+  komodo: bitcoinDescriptor,
+  pivx: bitcoinDescriptor,
+  vertcoin: bitcoinDescriptor,
+  viacoin: bitcoinDescriptor,
+  stakenet: bitcoinDescriptor,
+  peercoin: bitcoinDescriptor,
+};
```
gre-ledger: shouldn't this be part of "generated/" ? we need to avoid introducing more of these everywhere in the project, at least if we can collocate the multiplexing at the generated/ level until we have a general architecture solution addressed to get rid of the whole generated approach.
especially it means a new coin addition needs to update this file (vs before it wouldn't have to)
another issue here is that live-hub being owner of this added file, we will be pinged on each new coin addition
finally, i don't see anything that is LLD specifics

gre-ledger (commented):
the only parts that are live-hub owned actually shouldn't be in live-hub scope IMO (two files in apps/ledger-live-desktop)

i shared feedback in more details

apps/ledger-live-desktop/src/newArch/utils/getDescriptor.ts
dilaouid: 💯% agree on the generated file idea, didn't thought of it, I think I can create a generated file on LLC for the registry descriptors that could be used for both LWM/LWD

Basically this PR is the first on a two parts preparation for an architectural setup for the tx flow redesign
The global idea is to have the UI driven by the coin-modules (or in this case by live-common)

The second part will be the set-up of a wizard/orchestrator on LLD, and each feature (send, staking, receive etc etc) will receive the descriptors informations of this PR to display or not some controls (like memo for example, and also the kind of data expected e.g. for Stellar)
And also everything will also be drivable by the LiveConfig to allow or not some coin to use the new flow or the futur-legacy-one.

Actually this architectural idea is still subject of discussion, so i'm open to it 🙏

gre-ledger: 👍  ok

```
import type {
  CoinDescriptor,
  SendDescriptor,
  FeatureId,
} from "@ledgerhq/coin-framework/features/types";
import type { CryptoOrTokenCurrency } from "@ledgerhq/types-cryptoassets";
import { getCurrencyConfiguration } from "@ledgerhq/live-common/config/index";
import { descriptorRegistry } from "@ledgerhq/live-common/generated/descriptor";
```

but as there is no imports specifics to LLD so i think this file could still be in live-common no? will LLM eventually have to do the same?

dilaouid: Fixed ! The helpers are now also on LLC
If there will be different descriptors between LWC/LWD we can changes it directly on the descriptors themself

libs/coin-framework/src/config.ts
```diff
@@ -1,9 +1,16 @@
 import { CryptoCurrency, CryptoCurrencyId } from "@ledgerhq/types-cryptoassets";
 import { MissingCoinConfig } from "./errors";
+import type { FeatureConfig } from "./features/types";
 
 type ConfigStatus =
   | {
       type: "active";
+      features?: FeatureConfig[];
+      newFlows?: {
```
sprohaszka-ledger: What does this property means?

libs/ledger-live-common/scripts/sync-families-dispatch.mjs
```diff
@@ -260,6 +260,35 @@ ${exprtsStatusRaw}`;
   await fs.promises.writeFile(outpath, str, "utf8");
 }
 
+async function genDescriptorFile(families) {
```
sprohaszka-ledger: We should avoid adding things in this file as the target for moving the coin into a new repo is to remove this file...

libs/ledger-live-common/src/descriptor/getDescriptor.ts
```diff
@@ -0,0 +1,120 @@
+import type {
```
sprohaszka-ledger: Why creating a new directory for the descriptor?
Why not updating the bridge interface to "serve" this feature? It would ease the usage of CoinModule for the client if we can maintain a single entry point.

libs/coin-framework/src/features/types.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Feature status in liveconfig
+ */
+export type FeatureStatus = "active" | "inactive";
+
+/**
+ * Feature configuration in liveconfig
+ */
+export type FeatureConfig = {
+  id: FeatureId;
+  type: FeatureStatus;
```
sprohaszka-ledger: Instead of `type`, I would call it `status` or something like that. It is more self descriptive.

libs/coin-framework/src/features/types.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+export type FeatureIntentMap = {
+  blockchain_txs: BlockchainTxsIntent;
+  staking_txs: StakingTxsIntent;
+};
+
+/**
+ * A supported feature declaration in a coin module
+ */
+export type SupportedFeature<F extends FeatureId = FeatureId> = {
```
sprohaszka-ledger: **nit**: Unless it is easier for the client, I find it more readable if we can something like:
```ts
{
  "blockchain_txs": ["send"],
}
```

libs/coin-framework/src/config.ts
dilaouid: `features` is for all the existing intent for a specific coin, and their status (available/unavailable)
`newFlows` is to know if a specific new flow (send/receive/stake/...) should be used or not for a specific coin

libs/coin-framework/src/features/types.ts
dilaouid: Done !

libs/coin-framework/src/features/types.ts
dilaouid: Done !

libs/ledger-live-common/src/descriptor/getDescriptor.ts
dilaouid: Moved to `libs/ledger-live-common/src/bridge/descriptor.ts`

libs/coin-framework/src/config.ts
francois-guerin-ledger: Ok so if `false`, we keep using the current flow ? It sounds like a feature flag

libs/coin-framework/src/features/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export type InputFieldType = "text" | "number" | "tag" | "typed";
+
+/**
+ * Input field descriptor for a transaction flow
+ */
+export type InputDescriptor = {
+  type: InputFieldType;
+  maxLength?: number;
+  maxValue?: number;
+  options?: readonly string[];
```
francois-guerin-ledger: You are afraid values get pushed in `options` afterward ? In that case, `maxLength` and `maxValue` can be updated too

libs/coin-framework/src/features/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export type CoinDescriptor = {
+  send: SendDescriptor;
+  // Future: stake, swap, etc.
+};
+
+/**
+ * Helper function to check if a feature is supported
+ */
+export function hasFeature(supportedFeatures: SupportedFeatures, featureId: FeatureId): boolean {
+  return featureId in supportedFeatures && supportedFeatures[featureId] !== undefined;
```
francois-guerin-ledger: If `featureId` is a key in `supportedFeatures`, `supportedFeatures[featureId]` can not be `undefined`
```suggestion
  return featureId in supportedFeatures;
```

libs/ledger-live-common/src/bridge/descriptor.test.ts
```diff
@@ -0,0 +1,274 @@
...
+import * as configModule from "../config/index";
+
+jest.mock("../config/index");
+
+describe("getDescriptor", () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("should return default descriptor for undefined currency", () => {
```
francois-guerin-ledger: Probably use `it.each` to make these tests a bit more compact

libs/coin-framework/src/features/types.ts
dilaouid: 🤔 🤔 🤔 🤔 🤔 🤔 
Since descriptors are created once and never mutated in practice, the readonly might be overkill indeed... However, readonly on arrays prevents accidental mutations (like push as you say), when primitives are already immutable. 
The readonly keyword protects against array mutations, not primitive reassignment.
I think the readonly is useful to avoid array mutation, but we can remove it, wdyt?

libs/ledger-live-common/src/bridge/descriptor.test.ts
dilaouid: `it.each`-ed ✅

libs/coin-framework/src/features/types.ts
dilaouid: Done

libs/coin-framework/src/config.ts
dilaouid: Some coins will have the new send flow enabled (later receive, stake...), others won't. Using a feature flag would require creating 81 feature flags (3 per coin), which is overkill. Having it in the coin config allows us to control it per currency via LiveConfig/Firebase.

francois-guerin-ledger: Wallet XP team likes to use feature flags that contain object (see `lmModularDrawer` for example). Because this is their scope, you must ask them if they are confortable with coin config

libs/coin-framework/src/features/types.ts
francois-guerin-ledger: Indeed you are protected against in place mutation of `InputDescriptor.options` ... and that's it. I feel we are doing things half way, which seems weird. To me either remove `readonly` or go
```ts
export type InputDescriptor = Readonly<{
  type: InputFieldType;
  maxLength?: number;
  maxValue?: number;
  options?: readonly string[];
  supportsDomain?: boolean; // Whether the field supports domain names (ENS for EVM)
}>;
```

libs/coin-framework/src/config.ts
dilaouid: Still, it implies a way too big object, there will be a boolean for each flow and for each coin, even for an object it will stay way too big
I will still discuss it, but I think that the liveconfig is the best way for this, since a new flow won't be available for all coins at once, it will be easier to manage here

libs/coin-framework/src/features/types.ts
dilaouid: Also, `options` can take memos of `StellarMemoType`, which is also on readonly, ref:
```
src/families/stellar/descriptor.ts:15:9 - error TS4104: The type 'readonly ["NO_MEMO", "MEMO_TEXT", "MEMO_ID", "MEMO_HASH", "MEMO_RETURN"]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.
```

I can accept the `Readonly<{}>` solution though

libs/ledger-live-common/src/bridge/descriptor.test.ts
```diff
@@ -0,0 +1,218 @@
...
+    const currency = getCryptoCurrencyById(currencyId);
+    jest.spyOn(configModule, "getCurrencyConfiguration").mockReturnValue({
+      status: {
+        type: "active",
+        features: [{ id: "blockchain_txs", status: "active" }],
+      },
+    });
+
+    const descriptor = getDescriptor(currency);
+    expect(descriptor).toMatchObject(expected);
```
Moustafa-Koterba: **nit** wdyt of comparing strictly ? We want the config to not have unexpected attributes
```suggestion
    expect(descriptor).toEqual(expected);
```
same for others

Moustafa-Koterba (dismissed):

libs/ledger-live-common/src/bridge/descriptor.ts
```diff
@@ -0,0 +1,214 @@
...
+import { descriptor as tonDescriptor } from "../families/ton/descriptor";
+import { descriptor as tronDescriptor } from "../families/tron/descriptor";
+import { descriptor as vechainDescriptor } from "../families/vechain/descriptor";
+import { descriptor as xrpDescriptor } from "../families/xrp/descriptor";
+
+export const newFlowsConfig: Record<
+  string,
+  { send?: boolean; staking?: boolean; receive?: boolean }
+> = {
+  algorand: { send: true },
```
hedi-edelbloute: Shouldn't we start with false ?

libs/ledger-live-common/src/bridge/descriptor.ts
```diff
@@ -0,0 +1,214 @@
...
+function isFeatureActive(currency: CryptoCurrency, featureId: FeatureId): boolean {
+  try {
+    const currencyConfig = getCurrencyConfiguration(currency);
+
+    if (currencyConfig.status.type !== "active") {
+      return false;
+    }
+
+    if (!currencyConfig.status.features) {
+      return true;
```
hedi-edelbloute: everything is enabled if we don't define features ? shouldn't we want the opposite since it will mainly only affect new features

libs/ledger-live-common/src/bridge/descriptor.ts
dilaouid: True !
Set them as true first for testing purposes, forgot to changes them, nice catch !

libs/ledger-live-common/src/bridge/descriptor.ts
dilaouid: Fixed !

libs/ledger-live-common/src/bridge/descriptor.test.ts
dilaouid: Done !

qperrot (dismissed):

sprohaszka-ledger (commented):
If the `CoinDescriptor` is only implemented in LLC, I don't understand why we want to define it in the CoinFramework.
The CoinFramework should only define common interface for CoinModule. It needs to be unrelated with the UI.

PR #13166, opened 2025-12-05
[LIVE-23800] Create a ZCash lib for shielded transactions

alevito:
### ✅ Checklist

  - ...

### 📝 Description

Creates the scaffolding for a zcash-shielded library for shielded transactions.
The library is located in the `libs/coin-modules` path, as this turned out to be the most suitable.
This is not a new coin module, so it doesn't comply with the structure of the other coin modules.

### ❓ Context

- **JIRA**: https://ledgerhq.atlassian.net/browse/LIVE-23800

---

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1280,12 +1280,6 @@ importers:
       '@segment/sovran-react-native':
         specifier: 1.1.3
         version: 1.1.3(@react-native-async-storage/async-storage@2.1.2(react-native@0.77.3(patch_hash=2b53c1e5f1809d0b579eb158f0090709b837442527d3897d1f563069259d3db2)(@babel/core@7.27.1)(@babel/preset-env@7.27.1(@babel/core@7.27.1))(@react-native-community/cli@15.1.0(typescript@5.4.3))(@types/react@18.2.73)(react@18.3.1)))(react-native@0.77.3(patch_hash=2b53c1e5f1809d0b579eb158f0090709b837442527d3897d1f563069259d3db2)(@babel/core@7.27.1)(@babel/preset-env@7.27.1(@babel/core@7.27.1))(@react-native-community/cli@15.1.0(typescript@5.4.3))(@types/react@18.2.73)(react@18.3.1))(react@18.3.1)
-<<<<<<< HEAD
-=======
-      '@sentry/react-native':
-        specifier: 6.13.0
-        version: 6.13.0(expo@52.0.46(da95d96e6b14ca1108c91786bbc295f9))(react-native@0.77.3(patch_hash=2b53c1e5f1809d0b579eb158f0090709b837442527d3897d1f563069259d3db2)(@babel/core@7.27.1)(@babel/preset-env@7.27.1(@babel/core@7.27.1))(@react-native-community/cli@15.1.0(typescript@5.4.3))(@types/react@18.2.73)(react@18.3.1))(react@18.3.1)
->>>>>>> 862555bb85 (update dependencies)
       '@shopify/flash-list':
         specifier: 2.2.0
         version: 2.2.0(@babel/runtime@7.27.1)(react-native@0.77.3(patch_hash=2b53c1e5f1809d0b579eb158f0090709b837442527d3897d1f563069259d3db2)(@babel/core@7.27.1)(@babel/preset-env@7.27.1(@babel/core@7.27.1))(@react-native-community/cli@15.1.0(typescript@5.4.3))(@types/react@18.2.73)(react@18.3.1))(react@18.3.1)
@@ -15198,8 +15192,6 @@ packages:
   '@ledgerhq/logs@6.12.0':
     resolution: {integrity: sha512-ExDoj1QV5eC6TEbMdLUMMk9cfvNKhhv5gXol4SmULRVCx/3iyCPhJ74nsb3S0Vb+/f+XujBEj3vQn5+cwS0fNA==}
 
-<<<<<<< HEAD
-<<<<<<< HEAD
   '@ledgerhq/lumen-design-core@0.0.35':
     resolution: {integrity: sha512-qQ9j9y5S7XZ+W3YagTw1uQgWj5t825yHj3Ih4+/DWvpxT+ZQVgnszcQlpIdEbm4tiIJhl4Rwbanb7ecJDIVVtg==}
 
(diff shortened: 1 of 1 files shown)


libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -0,0 +1,38 @@
...
+  static AVERAGE_BLOCK_SYNC_TIME_SECONDS = 5;
+
+  /**
+   * Estimates sync time given a start and an end block.
+   *
+   * @param startBlock starting block
+   * @param endBlock end block, usually the latest confirmed block
+   * @return the estimated sync time
+   */
+  async estimateSyncTime(startBlock: number, endBlock: number) {
```
semeano, line 15: This is something I mentioned in a meeting. Here we're assuming an average time of block processing of 5 seconds, but IMO we should timestamp the beginning and end of a block processing event, and do the calculations based on that. Or just show the progress as a percentage (based on the current block that is being processed at the moment).

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -0,0 +1,38 @@
...
+    return (endBlock - startBlock) * ZCash.AVERAGE_BLOCK_SYNC_TIME_SECONDS;
+  }
+
+  /**
+   * Decrypts a ZCash shielded - i.e., encrypted - transaction.
+   *
+   * @param encryptedTransaction string raw string representing an encrypted transaction.
+   * @return a decrypted transaction
+   */
+  async decryptTransaction(encryptedTransaction: string): Promise<string> {
```
semeano, line 25: Maybe this should be a Buffer?

libs/coin-modules/zcash-shielded/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/LedgerHQ/ledger-live/issues"
+  },
+  "homepage": "https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/zcash-shielded",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "lib/ZCash.js",
+  "module": "lib-es/Cosmos.js",
```
may01: Looks like "Cosmos" should be replaced with zcash-shielded.

libs/coin-modules/zcash-shielded/src/ZCash.ts
may01: I suggest we will perform one block sync benchmark on client to provide relevant time estimation per block. It will be separate task.

libs/coin-modules/zcash-shielded/package.json
alevito: done in latest commits

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: I updated the average time per block constant to milliseconds rather than seconds, as ideally a block sync should not be in the seconds range!

alevito: resolving this one as the block sync benchmark will be estimated as part of a separate ticket.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: can you point me to an example please?

alevito: agreed to update the function prototype in a separate Jira ticket for implementing decryptTransaction.

libs/coin-modules/zcash-shielded/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@ledgerhq/zcash-shielded",
+  "version": "1.0.0",
```
sprohaszka-ledger: We could set the version as an early stage one :)

libs/coin-modules/zcash-shielded/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "ZCash"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/LedgerHQ/ledger-live.git"
+  },
+  "bugs": {
+    "url": "https://github.com/LedgerHQ/ledger-live/issues"
+  },
+  "homepage": "https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/zcash-shielded",
```
sprohaszka-ledger: This is not the correct url

libs/coin-modules/zcash-shielded/package.json
```diff
@@ -0,0 +1,77 @@
...
+      "lib-es/*": [
+        "lib-es/*"
+      ],
+      "*": [
+        "lib/*"
+      ]
+    }
+  },
+  "exports": {
+    "./lib/*": "./lib/*.js",
```
sprohaszka-ledger: I think we should remove all this file pattern from an export.
We need to control what is allow to be used outside of the library. We should only export what we allow as a public API.

libs/coin-modules/zcash-shielded/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "doc": "documentation readme src/** --section=API --pe ts --re ts --re d.ts",
+    "lint": "eslint ./src --no-error-on-unmatched-pattern --ext .ts,.tsx --cache",
+    "lint:fix": "pnpm lint --fix",
+    "test": "jest",
+    "test-integ": "jest --config=jest.integ.config.js",
+    "typecheck": "tsc --noEmit",
+    "unimported": "unimported"
+  },
+  "dependencies": {
+    "@ledgerhq/coin-framework": "workspace:^",
```
sprohaszka-ledger: I think that we should remove all libraries that are not used right now.
Ideally we shouldn't depend on `@ledgerhq/devices` for example.

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * ZCash API
+ */
+
+export default class ZCash {
+  static AVERAGE_BLOCK_SYNC_TIME_MS = 5;
```
sprohaszka-ledger: Try `readonly`

libs/coin-modules/zcash-shielded/package.json
alevito: changed to `0.1.0`

libs/coin-modules/zcash-shielded/package.json
alevito: updated to the correct URL

libs/coin-modules/zcash-shielded/package.json
alevito: I agree on this, although I added some of these dependencies - e.g., `@ledgerhq/coin-framework` - were required by the CI/CD pipeline.

I am going to remove any unused dependencies and devDependencies to check which one are strictly required.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: thanks, done in latest commit

libs/coin-modules/zcash-shielded/package.json
alevito: Changed the exports to strictly point to `lib/ZCash.js` and `lib-es/ZCash.js`

sprohaszka-ledger (approved):

PR #13203, opened 2025-12-08
[EXPLORATORY branch] Introduce identities library and unify IDs usage

gre-ledger:
### ✅ Checklist

- [ ] `npx changeset` was attached.
- [ ] **Covered by automatic tests.** 
- [ ] **Impact of the changes:** 
  - ...

### 📝 Description

General goal of this PR is to clarify all usecases of "IDs" in the apps. and converge them all to use a new library called `identities` that will be the central place to manage these and to secure the usages of ids.

This PR includes work from @paoun-ledger on https://ledgerhq.atlassian.net/browse/LIVE-23699
but also is exploring the general goal of https://ledgerhq.atlassian.net/browse/LIVE-23880 
as well as securing at least the `DeviceId` part for https://ledgerhq.atlassian.net/browse/LIVE-23878

when we say converging all the `identities` it also means we remove the env "USER_ID" as well as migrating from the old ways of storing ideas ( LLD : old old app.json location + old localStorage location )

We will not merge this PR this week but we'll extract some of the work to make this PR a general goal to target to.

### ❓ Context

- **JIRA or GitHub link**: 

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #13224, opened 2025-12-09
Introduce lib client-ids to unify Ledger Wallet IDs management and manage device IDs

gre-ledger:
### ✅ Checklist

  - [x] near 100% unit tests on the identities lib
  - e2e test is not possible as we need a real device
  - mobile
  - desktop

### 📝 Description

> FYI: the big picture PR is here : https://github.com/LedgerHQ/ledger-live/pull/13203 but we won't ship everything yet. For now we focus only on the DeviceId parts.

We introduce `client-ids`, a dedicated library that will eventually unify how Ledger Wallet manages IDs and isolate them from one another. We use ID classes that protect actual ID values by hiding them behind symbols, with strict rules that list all allowed usages.

```
const deviceId = new DeviceId("device-123");
console.log("" + deviceId); // [DeviceId:REDACTED]
JSON.stringify({ deviceId }); // {"deviceId":"[DeviceId:REDACTED]"}
```

we then have explicit export-rules.json that will declare where the actual extract of the ID are authorized so we prevent future overuse of ids. ( see https://ledgerhq.atlassian.net/browse/LIVE-23880 )

At the moment, the code effect is inactive, because there is no backend yet, the `PUSH_DEVICES_SERVICE_URL` is set to `""`.

### Testability

To be able to test it, 

you can run the app with 

```
export PUSH_DEVICES_SERVICE_URL=http://localhost:8888
./run/the/build
```

you can run a local server and set the `http://url` and you will be able to receive the data, as soon as you do a device Allow Manager action and if the analytics are opt in:

```ts
const express = require('express');
const app = express();

app.use(express.text({ type: "*/*" })); // accept any body

// Catch ALL routes and ALL methods
app.use((req, res) => {
  console.log("Method:", req.method);
  console.log("URL:", req.url);
  console.log("Headers:", req.headers);
  console.log("Body:", req.body);

  res.send({});
});

app.listen(8888, '0.0.0.0', () => {
  console.log("Listening on http://localhost:8888");
});
```

**as soon as you accept Allow Manager on the device, during any device flow, AND if the analytics are accepted in the settings, you will have this http call:**

```
node server.js
Listening on http://localhost:8888
Method: POST
URL: /pushdevices
Headers: {
  host: '192.168.1.50:8888',
  connection: 'keep-alive',
  'content-length': '134',
  'user-agent': ...,
  'content-type': 'application/json',
  'x-ledger-client-version': 'lld/2.133.0-dev',
  accept: '*/*',
  referer: 'http://localhost:8080/',
  'accept-encoding': 'gzip, deflate',
  'accept-language': 'en-US'
}
Body: {"equipment_id":"9e...............71","devices":["29.................1c"]} 
// (redacted)
```

you can then verify that the data is saved in app.json so we don't push back:

```
cat ~/Library/Application\ Support/Electron/app.json | jq '.data.identities'
{
  "deviceIds": [
    "29.......................1c"
  ],
  "pushDevicesSyncState": "synced",
  "pushDevicesServiceUrl": "http://192.168.1.50:8888"
}
```

### ❓ Context

- **JIRA or GitHub link**:
  - This PR includes work from @paoun-ledger on https://ledgerhq.atlassian.net/browse/LIVE-23699
  - https://ledgerhq.atlassian.net/browse/LIVE-23878

---

### DIFF
--- a/apps/ledger-live-mobile/src/context/store.ts
+++ b/apps/ledger-live-mobile/src/context/store.ts
@@ -6,6 +6,10 @@ import { rozeniteDevToolsEnhancer } from "@rozenite/redux-devtools-plugin";
 import { applyLlmRTKApiMiddlewares } from "./rtkQueryApi";
 import { setupCryptoAssetsStore } from "../config/bridge-setup";
 import { setupRecentAddressesStore } from "LLM/storage/recentAddresses";
+import { createIdentitiesSyncMiddleware } from "@ledgerhq/client-ids/store";
+import { State } from "~/reducers/types";
+import { trackingEnabledSelector } from "~/reducers/settings";
+import getOrCreateUser from "~/user";
 
 // === STORE CONFIGURATION ===
 export const store = configureStore({
@@ -14,7 +18,19 @@ export const store = configureStore({
   middleware: getDefaultMiddleware =>
     applyLlmRTKApiMiddlewares(
       getDefaultMiddleware({ serializableCheck: false, immutableCheck: false }),
-    ).concat(rebootMiddleware),
+    )
+      .concat(rebootMiddleware)
+      .concat(
+        createIdentitiesSyncMiddleware({
+          getIdentitiesState: (state: State) => state.identities,
+          getUserId: async (_state: State) => {
+            // FIXME LIVE-23880: Migrate to use userId from identities store or app-level user management
+            const { user } = await getOrCreateUser();
+            return user.id;
+          },
+          getAnalyticsConsent: (state: State) => trackingEnabledSelector(state),
+        }),
+      ),
 
   enhancers: getDefaultEnhancers => {
     const enhancers = getDefaultEnhancers();
--- a/libs/client-ids/src/api/api.ts
+++ b/libs/client-ids/src/api/api.ts
@@ -1,0 +1,72 @@
+import { createApi, fetchBaseQuery, retry } from "@reduxjs/toolkit/query/react";
+import { getEnv } from "@ledgerhq/live-env";
+import { DeviceId } from "../ids";
+
+/**
+ * Request payload for the PushDevices endpoint
+ */
+export interface PushDevicesRequest {
+  equipment_id: string;
+  devices: string[];
+}
+
+/**
+ * RTK Query API for Push Devices Service
+ */
+export const pushDevicesApi = createApi({
+  reducerPath: "pushDevicesApi",
+  baseQuery: retry(
+    fetchBaseQuery({
+      baseUrl: "",
+      prepareHeaders: (headers: Headers) => {
+        headers.set("Content-Type", "application/json");
+        headers.set("X-Ledger-Client-Version", getEnv("LEDGER_CLIENT_VERSION"));
+        return headers;
+      },
+    }),
+    {
+      maxRetries: 3,
+    },
+  ),
+  tagTypes: [],
+  endpoints: build => ({
+    /**
+     * Push devices to the backend service
+     * This endpoint updates the backend with the current list of device IDs for a user
+     * Success is determined by HTTP response code (2xx), no response body is expected
+     */
+    pushDevices: build.mutation<void, PushDevicesRequest>({
+      query: body => {
+        const baseUrl = getEnv("PUSH_DEVICES_SERVICE_URL");
+        return {
+          url: `${baseUrl}/pushdevices`,
+          method: "POST",
+          body,
+        };
+      },
+    }),
+  }),
+});
+
+/**
+ * Create a push devices request
+ * @param userId - User ID (equipment_id) as string (managed by apps)
+ * @param deviceIds - Array of DeviceId objects
+ */
+export function createPushDevicesRequest(
+  userId: string,
+  deviceIds: DeviceId[],
+): PushDevicesRequest {
+  const equipment_id = userId;
+  const devices = deviceIds.map(deviceId => deviceId.exportDeviceIdForPushDevicesService());
+
+  return {
+    equipment_id,
+    devices,
+  };
+}
+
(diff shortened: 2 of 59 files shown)


francois-guerin-ledger:
Question: will this PR re-open discussions around currency IDs, or is it unrelated ?

gre-ledger:
@francois-guerin-ledger unrelated I'm afraid. see https://ledgerhq.atlassian.net/browse/LIVE-23880

gre-ledger:
> Take the required action to fix the issue indicated by this comment.

actually no SonarQube, we'll do it incrementally.. the comment is here to locate the code that will be addressed by the next ticket.

---

also the lib has 98% coverage...

```

 PASS  src/__tests__/middleware.test.ts
 PASS  src/__tests__/slice.test.ts
 PASS  src/__tests__/api.test.ts
 PASS  src/__tests__/persistence.test.ts
 PASS  src/__tests__/ids.test.ts
-----------------|---------|----------|---------|---------|-------------------
File             | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-----------------|---------|----------|---------|---------|-------------------
All files        |   98.09 |     92.3 |     100 |      98 |                   
 api             |     100 |      100 |     100 |     100 |                   
  api.ts         |     100 |      100 |     100 |     100 |                   
  index.ts       |     100 |      100 |     100 |     100 |                   
 ids             |     100 |      100 |     100 |     100 |                   
  DeviceId.ts    |     100 |      100 |     100 |     100 |                   
  index.ts       |     100 |      100 |     100 |     100 |                   
 store           |   97.36 |     90.9 |     100 |   97.26 |                   
  index.ts       |     100 |      100 |     100 |     100 |                   
  middleware.ts  |   95.74 |    88.88 |     100 |   95.55 | 80,86             
  persistence.ts |     100 |      100 |     100 |     100 |                   
  slice.ts       |     100 |      100 |     100 |     100 |                   
  types.ts       |     100 |      100 |     100 |     100 |                   
-----------------|---------|----------|---------|---------|-------------------

Test Suites: 5 passed, 5 total
Tests:       34 passed, 34 total
Snapshots:   0 total
Time:        3.523 s
 ```

File: apps/ledger-live-mobile/src/context/store.ts [hunk 1]
```diff
@@ -14,7 +19,20 @@ export const store = configureStore({
...
-    ).concat(rebootMiddleware),
+    )
+      .concat(rebootMiddleware)
+      .concat(
+        createIdentitiesSyncMiddleware({
+          getIdentitiesState: (state: State) => state.identities,
+          getUserId: (_state: State) => {
+            // TEMPORARY: Direct MMKV access for synchronous user ID retrieval
+            // FIXME LIVE-23880: Migrate to use userId from identities store or app-level user management
+            const user = mmkvStorageWrapper.get<User>("user");
```
LucasWerey: You should use storage instead, in case some users haven't migrated yet

gre-ledger: the problem is Storage is async, which i can't afford easily in context of this function..
but it's a good point.

btw the global goal would also be to have the "userId" migrated to the store of that identities lib, but this will be a future iteration. so i was anticipating (on the exploratory work) to have a "migrate" action that will take the previous storage to import that id in.

but for this temporary code, i will have to figure out a way to use storage abstraction to get asynchronously I guess 👀

francois-guerin-ledger:
On this new library, test files are under a `__tests__`. From what I know, we prefer them close their production counterpart (although it is not very clear looking at the overall codebase). I think we should start to standardize this, what do you think ?

francois-guerin-ledger (dismissed):
Ok for coin integration scope

apps/ledger-live-mobile/src/context/store.ts
```diff
@@ -14,7 +18,19 @@ export const store = configureStore({
...
+      .concat(rebootMiddleware)
+      .concat(
+        createIdentitiesSyncMiddleware({
+          getIdentitiesState: (state: State) => state.identities,
+          getUserId: async (_state: State) => {
+            // FIXME LIVE-23880: Migrate to use userId from identities store or app-level user management
+            const { user } = await getOrCreateUser();
+            return user.id;
+          },
+          getAnalyticsConsent: (state: State) => analyticsEnabledSelector(state),
```
paoun-ledger: in LLD you used `trackingEnabledSelector` instead of `analyticsEnabledSelector`, and I think the same should be done here because it's also the selector used to track into segment (and will be set if either analytics or specialRecommendations is enabled by the user)

libs/client-ids/src/api/api.ts
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Push devices to the backend service
+     * This endpoint updates the backend with the current list of device IDs for a user
+     * Success is determined by HTTP response code (2xx), no response body is expected
+     */
+    pushDevices: build.mutation<void, PushDevicesRequest>({
+      query: body => {
+        const baseUrl = getEnv("PUSH_DEVICES_SERVICE_URL");
+        return {
+          url: `${baseUrl}/pushdevices`,
```
paoun-ledger, line 42: it looks like backend endpoint will be `/mut` (monthly user tracking) instead of `/pushdevices`, with a deployment matching the name of repository: https://github.com/LedgerHQ/mut

paoun-ledger: fixed in infra side

gre-ledger: seen with backend, it will stay pushdevices

LucasWerey (approved):

andrevmatos:
Looks like `@ledgerhq/client-ids@0.2.0` is not published to npmjs, and many packages which depend on packages dependending on it are broken, e.g. `@ledgerhq/types-live`;
Workaround for now is to pin the versions exactly

PR #13305, opened 2025-12-15
fix: valid stakedAmount in coin-hedera staking ops

mdomanski-ext-ledger:
### ✅ Checklist

  - Hedera operations in Alpaca should have correct `details.stakedAmount`
  - Hedera operations in Alpaca should have correct `value` with excluded fee

### 📝 Description

This PR fixes the `stakedAmount` calculation bug caused by Mirror Node balance snapshot timing window.
It also excludes `fees` from `value` for more operation types, not only `OUT`.

---

### DIFF
--- a/libs/coin-modules/coin-hedera/src/logic/utils.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/utils.ts
@@ -525,9 +525,9 @@ export const calculateUncommittedBalanceChange = async ({
   address: string;
   startTimestamp: string;
   endTimestamp: string;
-}): Promise<bigint> => {
+}): Promise<BigNumber> => {
   if (Number(startTimestamp) >= Number(endTimestamp)) {
-    return BigInt(0);
+    return new BigNumber(0);
   }
 
   const uncommittedTransactions = await apiClient.getTransactionsByTimestampRange({
@@ -540,9 +540,9 @@ export const calculateUncommittedBalanceChange = async ({
   const uncommittedBalanceChange = uncommittedTransactions.reduce((total, tx) => {
     const transfers = tx.transfers ?? [];
     const relevantTransfers = transfers.filter(t => t.account === address);
-    const netChange = relevantTransfers.reduce((sum, t) => sum + BigInt(t.amount), BigInt(0));
-    return total + netChange;
-  }, BigInt(0));
+    const netChange = relevantTransfers.reduce((sum, t) => sum.plus(t.amount), new BigNumber(0));
+    return total.plus(netChange);
+  }, new BigNumber(0));
 
   return uncommittedBalanceChange;
 };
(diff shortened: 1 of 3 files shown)


libs/coin-modules/coin-hedera/src/logic/utils.ts
```diff
@@ -504,6 +505,48 @@ export const calculateAPY = (rewardRateStart: number): number => {
   return annualRate;
 };
 
+/**
+ * Calculates the uncommitted balance change for an account between two timestamps.
+ *
+ * This function handles the timing mismatch between Mirror Node balance snapshots and actual transactions.
+ * Balance snapshots are taken at regular intervals, not at every transaction, so querying by exact timestamp
+ * may return a snapshot from before or after the moment you need.
```
jprudent: I'm unsure you can get the balance after queried timestamp 
```suggestion
 * may return a snapshot from before moment you need.
```

libs/coin-modules/coin-hedera/src/logic/utils.ts
```diff
@@ -552,10 +605,19 @@ export const analyzeStakingOperation = async (
     return null;
   }
 
+  // calculate uncommitted balance changes between the last snapshot and the staking tx
+  const uncommittedBalanceChange = await calculateUncommittedBalanceChange({
+    address,
+    startTimestamp: accountAfter.balance.timestamp,
```
jprudent, line 611: just a doubt about balance timestamp: do we know if it's in sync with timestamp of consensus ?

jprudent (approved):

jprudent (approved):
Great job!

- I also agree with the performance trade off. There is room for improvement if we realy need to.

- Computing a balance is always a hard job. The logic you implemented looks correct given what we know. Even if there is tiny mistakes on corner cases, that won't be much prejudice for end user, and we will fix the delta when that happens. (also, my account is already a corner case)

- Another refinement we could have: For current staking (when last operation is a delegate), I wonder if we could call a smart contract or some API that have the "real" current staked amount. (but your implem is good enough!)

libs/coin-modules/coin-hedera/src/logic/utils.ts
mdomanski-ext-ledger: Hmm, `account.balance.timestamp` indicates when balance was recorded, so basically when the mirror node snapshot happened (normally it happens every ~15 minutes). I think it should be in sync with the timestamp of last transaction in the snapshot

mdomanski-ext-ledger:
> Another refinement we could have: For current staking (when last operation is a delegate), I wonder if we could call a smart contract or some API that have the "real" current staked amount. (but your implem is good enough!)

Fetching real-time balance is possible with SDK, but it's hard to say how long this is going to stay free/supported. 
Based on https://hedera.com/blog/querying-data-on-hedera-sdk-vs-mirror-node-rest-api:

image

Same with mirror node docs:

image

libs/coin-modules/coin-hedera/src/logic/utils.ts
```diff
@@ -504,6 +505,48 @@ export const calculateAPY = (rewardRateStart: number): number => {
...
+    address,
+    startTimestamp: `gt:${startTimestamp}`,
+    endTimestamp: `lte:${endTimestamp}`,
+  });
+
+  // Sum all balance changes from transfers related to this account
+  const uncommittedBalanceChange = uncommittedTransactions.reduce((total, tx) => {
+    const transfers = tx.transfers ?? [];
+    const relevantTransfers = transfers.filter(t => t.account === address);
+    const netChange = relevantTransfers.reduce((sum, t) => sum + BigInt(t.amount), BigInt(0));
```
acewf: think more usual to sum values using BigNumber dependency.

mdomanski-ext-ledger: That's a good point, thanks - pushed as a separate commit for easier review: https://github.com/LedgerHQ/ledger-live/pull/13305/commits/fee064cf8ab11985ded45f65909f0013f1ec7f7e

francois-guerin-ledger (approved):

PR #13400, opened 2025-12-22
feat: save timestamp in the recentAddresses store

dilaouid:
### ✅ Checklist

  - ...

### 📝 Description

image

In order to be able to have this view, we need to store the timestamp of which we interacted with a specific address, so we can have the "x times ago" below the address tile element.

### ❓ Context

- **[JIRA or GitHub link](https://ledgerhq.atlassian.net/jira/software/c/projects/LIVE/boards/465?selectedIssue=LIVE-24369)**: 

---

### DIFF
--- a/libs/ledger-live-common/src/account/recentAddresses.ts
+++ b/libs/ledger-live-common/src/account/recentAddresses.ts
@@ -1,14 +1,14 @@
+import type { RecentAddressesState, RecentAddress } from "@ledgerhq/types-live";
+
 export const RECENT_ADDRESSES_COUNT_LIMIT = 12;
 
-export type RecentAddressesCache = Record<string, string[]>;
+export type RecentAddressesCache = RecentAddressesState;
 
 export interface RecentAddressesStore {
   addAddress(currency: string, address: string): void;
   syncAddresses(cache: RecentAddressesCache): void;
   getAddresses(currency: string): string[];
 }
-
-type CallbackMode = "triggerCallback" | "skipCallback";
 
 let recentAddressesStore: RecentAddressesStore | null = null;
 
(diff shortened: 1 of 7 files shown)


dilaouid:
Desktop e2e: https://github.com/LedgerHQ/ledger-live/actions/runs/20425160776
Mobile e2e: https://github.com/LedgerHQ/ledger-live/actions/runs/20425174857

libs/live-wallet/src/walletsync/modules/recentAddresses.ts
```diff
@@ -33,6 +33,7 @@ export type DistantRecentAddressesState = Record<
   {
     address: string;
     index: number;
+    lastUsed: number;
```
gre-ledger: the problem with this idea is that it's not backward compat with older model version. it would need to be `lastUsed?: number` , unless we clearly identify the possibly missing key in the object coming from an old version of the data in walletsync?

libs/live-wallet/src/walletsync/modules/recentAddresses.ts
```diff
@@ -84,6 +87,7 @@ function sameDistantState(
 const recentAddressesSchema = z.object({
   address: z.string(),
   index: z.number(),
+  lastUsed: z.number(),
```
gre-ledger: same here, if we want to guarantee parsing the old model (that reach prod?) we need to make it .optional()
because here it would fail parsing old model? correct me if i'm wrong but in that case, that makes the whole parsing fail (at the top level of the parser, not module level, as we merge all schema into one schema)

dilaouid: You were right actually, thanks for the catch
Now fixed, `lastUsed`is now optional in the distant state/schema with automatic migration fallback for backward compatibility

libs/ledger-live-common/src/account/recentAddresses.ts
```diff
@@ -30,26 +30,43 @@ export function setupRecentAddressesStore(
...
   syncAddresses(cache: RecentAddressesCache): void {
     const previousAddresses = { ...this.addressesByCurrency };
     this.addressesByCurrency = { ...cache };
     for (const currency in previousAddresses) {
-      for (const address of previousAddresses[currency]) {
-        this.addAddressToCache(currency, address, "skipCallback");
+      const entries = previousAddresses[currency] as (RecentAddress | string)[];
+      for (const entry of entries) {
+        const address = typeof entry === "string" ? entry : entry.address;
+        const timestamp = typeof entry === "string" ? undefined : entry.lastUsed;
```
Moustafa-Koterba, line 68: **nit**
```suggestion
        const timestamp = typeof entry === "string" ? Date.now() : entry.lastUsed;
        this.addAddressToCache(currency, address, timestamp, false);
```

PR #13415, opened 2025-12-22
feat: segment migration

RobinVncnt:
### ✅ Checklist

  - Staging and production environments are now separated in segment

### 📝 Description

The previous setup loaded `analytics.min.js` from `resources.live.ledger.app`, which contained a hardcoded writeKey. This made it impossible to separate staging and production analytics environments.

Additionally, the file was uploaded manually in 2020.

Solution: 

- Replace the external script with `@segment/analytics-next` npm package
- Pass `cdnSettings` directly to prevent any fetch from `cdn.segment.com`
- Add `getSegmentWriteKey()` to select staging/production keys based on environment
- Remove the Segment snippet and external script from `index.html`

### ❓ Context

- **JIRA or GitHub link**: [LIVE-24054]

---

[LIVE-24054]: https://ledgerhq.atlassian.net/browse/LIVE-24054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/segment-setup.ts
```diff
@@ -0,0 +1,8 @@
+const SEGMENT_KEYS = {
+  staging: "olBQc203GA3fXVa48rJB9c3826CY1axp",
+  production: "RduLzmQ1vSH5aeNrSDKWimFrivogTANI",
+};
+
+export function getSegmentWriteKey(): string {
+  return process.env.NODE_ENV === "production" ? SEGMENT_KEYS.production : SEGMENT_KEYS.staging;
```
gre-ledger: it would be preferable to be using `__DEV__` like we did in other places (e.g. sentry do `environment: __DEV__ ? "development" : "production",` )

that said it's still not perfect, i think develop is not equivalent to staging : to me staging is for nightly too, right?

so we may want to have another way to inject the envs, typically with the dot env files.

see how we manage these

https://github.com/LedgerHQ/ledger-live/blob/ac08e7e622b030dadd4e0f8847d6b964273913d6/apps/ledger-live-desktop/.env.production
and
https://github.com/LedgerHQ/ledger-live/blob/ac08e7e622b030dadd4e0f8847d6b964273913d6/apps/ledger-live-desktop/.env.staging

PR #13576, opened 2026-01-07
Live 24391 earn set up new custom wallet api get flag method to fetch ledger live flags

sergiubreban:
The work from this PR will enable any web app to fetch feature flags directly from Ledger Wallet. 
Each webapp should declare the permisions inside their manifest file. To fetch a specific flag from the ledger wallet, the webapp needs to add `custom.featureFlags.get` to the list of permissions and include the flags they need to fetch under the key `featureFlags`.
  
exmaple of manifest file:
```
{
  "id": "earn",
  "name": "Earn",
  "url": "http://localhost:3000",
  "homepageUrl": "http://localhost:3000",
  "icon": "",
  "platform": "all",
  "apiVersion": "^2.0.0",
  "manifestVersion": "1",
  "branch": "stable",
  "categories": [
    "earn"
  ],
  "currencies": "*",
  "content": {
    "shortDescription": {
      "en": "Earn dashboard"
    },
    "description": {
      "en": "Earn dashboard"
    }
  },
  "permissions": [
    "account.list",
    "account.receive",
    "account.request",
    "currency.list",
    "custom.getTransactionByHash",
    "device.close",
    "device.exchange",
    "device.transport",
    "message.sign",
    "transaction.sign",
    "transaction.signAndBroadcast",
    "storage.set",
    "storage.get",
    "bitcoin.getXPub",
    "wallet.capabilities",
    "wallet.userId",
    "wallet.info",
    "custom.isReady",
    "custom.earn.isReady",
    "custom.featureFlags.get",
    "custom.getFunds",
    "custom.navigate",
    "custom.getAccountIdFormats"
  ],
  "featureFlags": [
    "stakePrograms"
  ],
  "domains": [
    "http://localhost:3000",
    "http://*",
    "https://*"
  ]
}
```

example of usage can be found in this draft/poc PR https://github.com/LedgerHQ/earn-live-app/pull/968

### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

[LIVE-24391](https://ledgerhq.atlassian.net/browse/LIVE-24391)

---

[LIVE-24391]: https://ledgerhq.atlassian.net/browse/LIVE-24391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Valentin-Ledger (approved):
👏

liviuciulinaru (approved):

libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -313,11 +314,21 @@ export function useWalletAPIServer({
   // If we don't want the map to be empty when requesting an account
   useSetWalletAPIAccounts(accounts);
 
+  // Merge featureFlags handler with customHandlers
+  const mergedCustomHandlers = useMemo(() => {
+    const featureFlagsHandlersInstance = featureFlagsHandlers({ manifest });
```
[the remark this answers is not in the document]
Justkant: This comment seems correct especially if people start using other featureFlags

libs/ledger-live-common/src/wallet-api/FeatureFlags/resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+  // Fetch each feature flag
+  const features: Record<string, Feature<unknown> | null> = {};
+
+  for (const featureFlagId of featureFlagIdsToFetch) {
+    try {
+      const feature = getFeature({
+        key: featureFlagId as FeatureId,
+        appLanguage,
+        allowOverride: true,
+      });
```
Justkant: We should get the featureFlags instance from the hook and provider using `const featureFlags = useFeatureFlags();` inside `useWalletAPIServer`

sergiubreban: I checked all the references to getFeature in the codebase, and there are no uses of appLanguage. This is a supported feature, but not used anywhere in the project as far as I know. And I was afraid to add this support to avoid creating any inconsistency with other uses. I was thinking of a use case where the feature flag is used in both live app and ledger-live codebase, and wanted to have a similar approach. 

I can change the PR to use the appLanguage from ledger-live, but will this be the right move? since the other references did not care about the appLanguage?

Justkant: You can keep it as is for the appLanguage then but please use the getFeature from the provider instead of importing an implementation directly

beths-ledger (approved):

libs/ledger-live-common/src/wallet-api/FeatureFlags/README.md
```diff
@@ -0,0 +1,334 @@
...
+import { WalletAPIClient, WindowMessageTransport } from "@ledgerhq/wallet-api-client";
+
+// Setup transport and client (typically done once in your app)
+const transport = new WindowMessageTransport();
+transport.connect();
+const walletApiClient = new WalletAPIClient(transport);
+const featureFlagModule = new FeatureFlagModule(walletApiClient);
+
+// Fetch feature flags
+const features = await featureFlagModule.get(["stakePrograms", "earnLidoStaking"]);
```
Justkant: It's not exactly the way it's suppose to be used but should work I suppose

We usually pass the custom modules to the WalletAPIClient constructor `getCustomModule` param

libs/ledger-live-common/src/wallet-api/FeatureFlags/README.md
```diff
@@ -0,0 +1,334 @@
...
+import { WalletAPIClient, WindowMessageTransport } from "@ledgerhq/wallet-api-client";
+
+// Setup (typically done once in your app initialization)
+const transport = new WindowMessageTransport();
+transport.connect();
+const walletApiClient = new WalletAPIClient(transport);
+const featureFlagModule = new FeatureFlagModule(walletApiClient);
+
+// Fetch multiple feature flags
+const features = await featureFlagModule.get(["stakePrograms", "earnLidoStaking"]);
```
Justkant: Same as above

libs/ledger-live-common/src/wallet-api/react.ts
```diff
@@ -315,11 +318,21 @@ export function useWalletAPIServer({
...
+  // Merge featureFlags handler with customHandlers
+  const mergedCustomHandlers = useMemo(() => {
+    const featureFlagsHandlersInstance = featureFlagsHandlers({ manifest, getFeature });
+
+    return {
+      ...featureFlagsHandlersInstance,
+      ...customHandlers,
+    };
+  }, [manifest, customHandlers, getFeature]);
+
```
Justkant, line 330: This is the only part where I'm not totally sure we want this, in one way it's simpler than adding it every where we declare the customHandlers but in another way we already declare outside for the rest even if shared so it's more explicit what a webview is using/exposing

wdyt ?

Justkant (commented):
LGTM just a last comment on this

libs/ledger-live-common/src/wallet-api/react.ts
sergiubreban: IMO declaring every shared handler where we define customHandlers for a specific live app would be redundant
- it's easy to forget adding it somewhere, leading to inconsistent behavior across apps.                                    

I looked at useDeeplinkCustomHandlers as an example - it's explicitly added in every place we declare custom handlers, and I don't see the clear benefit of that approach. It creates boilerplate and risks inconsistency if someone forgets to include it in a new integration.
                                                                                                                               
The trade-off is:                                                                                                            
- Current approach (merge here): Single point of truth, less boilerplate, guaranteed consistency across all live apps        
- Explicit everywhere: More visible what each webview exposes, but requires updating multiple places when adding new shared handlers                                                                                                                     
                                                                                                                               
Since feature flags should be available to all live apps uniformly, I'd lean toward keeping it centralized here. But happy to discuss if you feel strongly about the explicit approach.

Justkant: > Since feature flags should be available to all live apps uniformly

If it was the case it wouldn't be a custom handler and would be directly done in the wallet-api itself

We also don't have a lot of places and we don't create new webview pages every day but with the current change we are also forcing it on webviews that wouldn't need it here

In the end it's not a massive deal and you are correct on the trade-off, it's just more work here spreading existing custom handlers instead of adding them at the source where the rest is declared

Your solution is fine, let's go with it

PR #13721, opened 2026-01-15
LWD - feat/LIVE-23998 - upgrade styled-components to v6

iqbalibrahim-ledger:
### ✅ Checklist

  - the UI E2E and smoke tests capture screenshots, and these proved very helpful during development, showing the visual diff as a result of this upgrade, allowing me to resolve any discrepancies. These are passing as expected, and so this is a solid indicator the changes are correct and have not caused a visual regression.
  - unused dependencies `react-is` and `stylis` removed from LWD as `stylis` is now internal in `styled-components` v6
  - some components refactored from class to functional, takes advantage of modern hooks instead of HOC wrappers
  - resolved many TypeScript issues
  - minimal, if not 0, visual changes - UI tests do not require screenshots updating

### 📝 Description
This PR upgrades the `styled-components` package for LWD to v6. Changes include:
- refactoring some class components into functional components to remove HOC wrappers, and use modern hooks instead (`useTheme` and `useTranslation` instead of exporting as `withTheme()(Component)` or `withTranslation()(Component)` - HOC such as `withTheme` that was removed in v6, same for `withTranslation` and hooks was the replacement requiring refactoring class components
- fixed all TypeScript errors as a result of the ugprade
- added typing to `attrs` method on style components
- removed unused props such as `secondary` on the `Button` components which never existed
- updated theme typings to reflect actual theme definition

### ❓ Context

- **JIRA or GitHub link**: 
- [LIVE-23998](https://ledgerhq.atlassian.net/browse/LIVE-23998)

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepFinish.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepFinish.tsx
@@ -36,16 +36,7 @@ const StepFinish = ({
   );
 };
 
-export const StepFinishFooter = ({
-  t,
-  onAddAccounts,
-  onAddMore,
-  isReonboarding,
-  importableAccounts,
-  creatableAccount,
-}: StepProps) => {
-  const accounts = [...importableAccounts, creatableAccount];
-
+export const StepFinishFooter = ({ t, onAddAccounts, onAddMore, isReonboarding }: StepProps) => {
   return (
     <Box
       horizontal
@@ -68,7 +59,7 @@ export const StepFinishFooter = ({
         event="Page AddAccounts Step 4 Close"
         data-testid="add-accounts-finish-close-button"
         primary
-        onClick={() => onAddAccounts(accounts)}
+        onClick={onAddAccounts}
         aria-label="Complete Canton account setup"
       >
         {isReonboarding ? t("common.continue") : t("common.done")}
--- a/apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/types.ts
+++ b/apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/types.ts
@@ -29,7 +29,7 @@ export type StepProps = {
   authorizeStatus: AuthorizeStatus;
   error: Error | null;
   isReonboarding?: boolean;
-  onAddAccounts: (accounts: Account[]) => void;
+  onAddAccounts: () => void;
   onAddMore: () => void;
   onAuthorizePreapproval: () => void;
   onOnboardAccount: () => void;


gre-ledger:
unless this is flakiness, the screenshot test may experience some diff:

```
    Expected: /home/runner/work/ledger-live/ledger-live/apps/ledger-live-desktop/tests/specs/manager/manager.spec.ts-snapshots/manager-install-tron-linux.png
    Received: /home/runner/work/ledger-live/ledger-live/apps/ledger-live-desktop/tests/artifacts/test-results/manager-manager-Manager-smoke-mocked-tests/manager-install-tron-actual.png
```

you may want to download the github action artifacts to dig

(side note: we can't trust these screenshot test much because we have some % diff allowed which may less pass some slight diff changes. so manual test is still the way)

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/LiveAppDisclaimer.tsx
```diff
@@ -60,7 +60,6 @@ const InfoText = styled(Text).attrs(p => ({
   ml: "6px",
   ff: "Inter|Medium",
   fontSize: 12,
-  lineHeight: 18,
```
mcayuelas-ledger, line 63: do we need to remove this ?

apps/ledger-live-desktop/src/renderer/styles/styled/fontFamily.ts
```diff
@@ -1,5 +1,12 @@
+import { DefaultTheme } from "styled-components";
 import { ff } from "./../helpers";
-export default (props: { ff?: string }) => {
+
+type FontFamilyProps = {
+  ff?: string;
+  theme?: DefaultTheme;
```
mcayuelas-ledger: why do we need this?

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepRecipient.tsx
```diff
@@ -94,6 +94,7 @@ const StepRecipient = ({
           <Box flow={1}>
             <Label>{t("send.steps.details.selectAccountDebit")}</Label>
             <SelectAccount
+              accounts={[]}
```
mcayuelas-ledger: I think you should make accounts optional inside `SelectAccount` instead because it can overwrite it

apps/ledger-live-desktop/src/renderer/families/polkadot/RebondFlowModal/steps/StepConfirmation.tsx
```diff
@@ -93,7 +93,7 @@ export function StepConfirmationFooter({
           <Trans i18nKey="polkadot.rebond.steps.confirmation.success.cta" />
         </Button>
       ) : error ? (
-        <RetryButton primary ml={2} onClick={onRetry} />
+        <RetryButton primary ml={2} onClick={() => onRetry(undefined)} />
```
mcayuelas-ledger: Don't like too much to have to pass `undefined`

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/LiveAppDisclaimer.tsx
iqbalibrahim-ledger: I have restored this, it was accidental.

apps/ledger-live-desktop/src/renderer/families/polkadot/RebondFlowModal/steps/StepConfirmation.tsx
iqbalibrahim-ledger: I have updated the `RetryButton` component's `onClick` prop so it respects the typing of the `Button` component used internally, but does not require any arguments for the function passed to it. Updated the usage across all instances of `RetryButton` so it's clean.

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepRecipient.tsx
iqbalibrahim-ledger: I have made the `accounts` prop optional and defaults to `[]`, so it does not cause a regression, and removed `accounts={[]}` for any instance of `<SelectAccount>`

apps/ledger-live-desktop/src/renderer/styles/styled/fontFamily.ts
iqbalibrahim-ledger: We don't, I have removed it

apps/ledger-live-desktop/src/renderer/components/PlaceholderChart.tsx
```diff
@@ -1,47 +1,43 @@
...
 };
 
-class PlaceholderChart extends Component<Props> {
-  shouldComponentUpdate(next: Props) {
-    return next.tickXScale !== this.props.tickXScale;
-  }
+const PlaceholderChart = ({ data, tickXScale, magnitude }: Props) => {
+  const theme = useTheme();
+  const themeType = theme.theme;
+  const mappedData: Data = data.map((i: Item) => {
```
gre-ledger: put this is in a `useMemo` for perf 🙏

iqbalibrahim-ledger: I have done this :)

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/LiveAppDisclaimer.tsx
Justkant: Looks like it's still removed, no ?

apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepFinish.tsx
```diff
@@ -59,7 +68,7 @@ export const StepFinishFooter = ({ t, onAddAccounts, onAddMore, isReonboarding }
         event="Page AddAccounts Step 4 Close"
         data-testid="add-accounts-finish-close-button"
         primary
-        onClick={onAddAccounts}
+        onClick={() => onAddAccounts(accounts)}
```
Justkant: onAddAccounts in `apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/index.tsx` doesn't seem to receive any params, why did you need to change this ?

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/LiveAppDisclaimer.tsx
iqbalibrahim-ledger: I've updated it so it matches the correct type `string` rather than number. Only this file and `AnnouncementPanel.tsx` appear to use this property on the styled `Text` component; reverted to ensure compatibility.

apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepFinish.tsx
iqbalibrahim-ledger: the `onAddAccounts` function receives one param which is `accounts: Account[]`, but the `onClick` of the button expects a `MouseEvent` type of event, so this causes a mismatch when passing the function like before. By doing it this way, we ignore the click event, and invoke the `onAddAccounts` function.

image

apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/LiveAppDisclaimer.tsx
Justkant: Oh okay my bad

apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/steps/StepFinish.tsx
Justkant: I see strange because the signature of `handleAddAccounts` is not correct and typescript doesn't say anything (no idea how this could be broken)
We should update the type to be correct IMO, it might fix this issue with the onClick at the same time
Screenshot 2026-01-26 at 10 42 01

iqbalibrahim-ledger: the prop `onAddAccounts` has the param `accounts: Account[]` in the file `apps/ledger-live-desktop/src/renderer/families/canton/OnboardModal/index.tsx`, which was last modified 4 months ago

image

So the type is correct, it just means that `this.handleAddAccounts` is not using the `accounts` param data that gets passed through. I have checked the diff with `develop` and I cannot see any modifications to it in my branch, this may be a separate issue and not directly related to this PR? 

I'd say the `onClick` issue is not an issue, because the `onClick` is passing `React.MouseEventHandler` but we want to invoke the `onAddAccounts(accounts)` function that does not accept this event type, so the type mismatch can be overridden by calling the function without the implicit click handler event being passed in.

iqbalibrahim-ledger:
https://github.com/user-attachments/assets/f518d931-b5aa-463e-ba2b-15db6ea42402

tested onboarding flow @gbecerra-ledger 


https://github.com/user-attachments/assets/649c5c91-5268-4457-b0f8-73da06d9f2ba


https://github.com/user-attachments/assets/5b3cf7b1-733b-4e57-b8e9-d719b1652cc8

many thanks for your help :)

PR #13750, opened 2026-01-19
fix(live-25022): add DrawerClosedError handling in error parsing

liviuciulinaru:
### ✅ Checklist

  - ...

### 📝 Description

https://github.com/user-attachments/assets/b99b4ffe-0429-40fc-a470-bda5396d1e1e

- Introduced DrawerClosedError type to manage user-initiated cancellations.
- Updated parseError and createStepError functions to return DrawerClosedError without wrapping.
- Added tests to verify the correct behavior of DrawerClosedError handling.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/ledger-live-common/src/wallet-api/Exchange/parser.test.ts
+++ b/libs/ledger-live-common/src/wallet-api/Exchange/parser.test.ts
@@ -95,6 +95,19 @@ describe("createStepError", () => {
     expect(wrapped).toBe(drawerClosedError);
   });
 
+  it("returns error with DrawerClosedError cause without wrapping", () => {
+    const errorWithCause = Object.assign(new Error("Wrapper error"), {
+      cause: drawerClosedError,
+    });
+
+    const wrapped = createStepError({
+      error: errorWithCause,
+      step: StepError.NONCE,
+    });
+
+    expect(wrapped).toBe(errorWithCause);
+  });
+
   it.each(STEP_EXPECTATIONS)(
     "$step - wraps downstream error with appropriate swap error",
     ({ step, ExpectedError, swapCode }) => {


libs/ledger-live-common/src/wallet-api/Exchange/parser.test.ts
```diff
@@ -35,6 +35,9 @@ const STEP_EXPECTATIONS: StepExpectation[] = [
 ];
 
 const mockDownstreamError = new Error("error message");
+const drawerClosedError = Object.assign(new Error("User closed the drawer"), {
+  name: "DrawerClosedError",
+});
```
Justkant, line 40: You could export the error from common in `wallet-api/Exchange/` and share it with desktop and mobile instead of having it created in the apps, then you could just import it in the test

No need to change it for this PR, better to do this in another specific PR

libs/ledger-live-common/src/wallet-api/Exchange/parser.test.ts
```diff
@@ -66,13 +69,32 @@ describe("parseError", () => {
...
+
+  it("returns DrawerClosedError when customErrorType is 'swap'", () => {
+    const error = parseError({
+      error: drawerClosedError,
+      step: StepError.NONCE,
+      customErrorType: CustomErrorType.SWAP,
+    });
+
+    expect(error).toBe(drawerClosedError);
+  });
```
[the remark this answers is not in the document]
Justkant: It might be interesting to also cover this case

libs/ledger-live-common/src/wallet-api/Exchange/parser.ts
```diff
@@ -43,6 +43,17 @@ export type ParseError = {
...
 
+const DRAWER_CLOSED_ERROR_NAME = "DrawerClosedError";
+
+const isDrawerClosedError = (error: Error): boolean => {
+  if (error.name === DRAWER_CLOSED_ERROR_NAME) {
+    return true;
+  }
+
+  const nestedName = (error as { cause?: { name?: string } }).cause?.name;
+  return nestedName === DRAWER_CLOSED_ERROR_NAME;
```
Justkant, line 54: You could remove the `as` like this but it feels a lot heavier so you can keep it as is if you want
```suggestion
  if (
    "cause" in error &&
    typeof error.cause === "object" &&
    error.cause &&
    "name" in error.cause &&
    typeof error.cause.name === "string"
  ) {
    const nestedName = error.cause.name;
    return nestedName === DRAWER_CLOSED_ERROR_NAME;
  }
  return false;
```

Justkant (approved):

PR #13794, opened 2026-01-19
chore: make ethereum-provider use rspack instead of rollup

gre-ledger:
### ✅ Checklist

  - ethereum-provider library build

### 📝 Description

- [ ] manually confirm that the .umd.js correctly works

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/ethereum-provider/package.json
+++ b/libs/ethereum-provider/package.json
@@ -30,7 +30,6 @@
   "scripts": {
     "clean": "rimraf lib",
     "build": "pnpm run clean && rspack build -c rspack.config.mjs && tsc --emitDeclarationOnly && node scripts/build-umd-json.mjs",
-    "watch": "rspack build --watch -c rspack.config.mjs",
     "lint": "eslint ./src --no-error-on-unmatched-pattern --ext .ts,.tsx --cache",
     "lint:fix": "pnpm lint --fix",
     "typecheck": "tsc --noEmit",


libs/ethereum-provider/package.json
```diff
@@ -23,18 +23,14 @@
...
-    "rollup-plugin-esbuild": "^6.1.1"
+    "@rspack/cli": "catalog:",
+    "@rspack/core": "catalog:"
   },
   "scripts": {
     "clean": "rimraf lib",
-    "build": "rollup -c",
-    "watch": "rollup -c -w",
+    "build": "pnpm run clean && rspack build -c rspack.config.mjs && tsc --emitDeclarationOnly && node scripts/build-umd-json.mjs",
+    "watch": "rspack build --watch -c rspack.config.mjs",
```
Justkant: How does it work for the watch mode if we need to emit ts declarations in build mode ?
Do the typescript declarations not update using the watch script ?

libs/ethereum-provider/rspack.config.mjs
```diff
@@ -0,0 +1,79 @@
...
+  },
+  resolve: {
+    extensions: [".ts", ".js"],
+  },
+  module: {
+    rules: [swcRule],
+  },
+  target: "web",
+  optimization: {
+    minimize: false,
```
Justkant, line 36: We might want this to be true as it reduces the size of what we send to the webview but not 100% sure

gre-ledger:
(i only rebased the PR. will need to address comment later)

libs/ethereum-provider/rspack.config.mjs
Justkant: I tested and it seems fine like that for now, I've diff develop and your branch on the umd and js files, it's the same appart from the webpack specific stuff

Justkant (approved):

PR #13798, opened 2026-01-19
chore: dummy-wallet-app to use rspack

gre-ledger:
### ✅ Checklist

  - dummy wallet used in playwright mock tests

### 📝 Description

in continuation of unifying our stack to rspack for the preferred build system, we can make dummy-wallet-app uses it (instead of having to update vite)

### ❓ Context

- **JIRA or GitHub link**: 

---


tests/dummy-wallet-app/public/index.html
```diff
@@ -1,17 +1,14 @@
...
     <meta name="theme-color" content="#000000" />
-    <meta name="description" content="Ledger Live App created using create-react-app" />
+    <meta name="description" content="Dummy Wallet API App" />
     <title>Dummy Wallet API App</title>
   </head>
-
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
-    <script type="module" src="/src/index.tsx"></script>
```
Justkant, line 15: Do we control where rspack places the script tags ?
Not needed in that case, just a curiosity

tests/dummy-wallet-app/package.json
```diff
@@ -17,10 +17,9 @@
     "react-dom": "catalog:"
   },
   "scripts": {
-    "start": "vite",
-    "build": "tsc && vite build",
-    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
-    "preview": "vite preview"
+    "start": "rspack serve",
```
Justkant, line 20: I see that `serve` is an alias for the main command `dev`, should we use `dev` instead ?
It's a clearer that we are on a dev build
There is also the `preview` command that can be used with a prod build, not sure we really need that but it was present before

tests/dummy-wallet-app/tsconfig.json
```diff
@@ -6,6 +6,5 @@
     "noEmit": true,
```
Justkant, line 6: I think we used the `noEmit` for the `tsc` call in the build script, not sure we still need it

Justkant (approved):

tests/dummy-wallet-app/package.json
gre-ledger: start would keep it non breaking with before, but we could also rename it to dev. but i see that live-app also uses start, probably from the react-scripts convention era^^

tests/dummy-wallet-app/public/index.html
gre-ledger: as far as i understand it's not needed anymore (because of https://rspack.rs/plugins/rspack/html-rspack-plugin ?)

at least from my test, if keeping it, it produces a `Failed to load resource: the server responded with a status of 404 (Not Found)` on the index.tsx

gre-ledger:
@Justkant quick question:
- is dummy wallet app covered by running the ubuntu mock tests?
- is it expected to see this when running the dummy app in Chrome directly? 
```
Uncaught TypeError: crypto.randomUUID is not a function
    at secureBrowsing.bundle.js:1:506
    at secureBrowsing.bundle.js:31:5746
    at secureBrowsing.bundle.js:31:5750
```

tests/dummy-wallet-app/public/index.html
Justkant: got it, I see in the doc we can control it with the `inject` field

tests/dummy-wallet-app/package.json
Justkant: I mean `rspack dev` vs `rspack serve`

Justkant: But you can keep it like this

Justkant:
@gre-ledger 

Yes it's covered with ubuntu mock tests

But I'm not sure about the second one, we don't really run it in chrome directly and we probably had a crypto polyfill with the `vite-plugin-node-polyfills`

tests/dummy-wallet-app/package.json
gre-ledger: ooooh ok

gre-ledger:
thanks. i'll iterate tomorrow

gre-ledger:
`Uncaught TypeError: crypto.randomUUID is not a function` was actually caused by Torii extension on Chrome ^^

Justkant (approved):

PR #13809, opened 2026-01-20
feat(BACK-10485): add support for sending Hedera ERC20 tokens to EVM addresses

jnicoulaud-ledger:
### ✅ Checklist

  - hedera

### 📝 Description

Resolve the recipient address using EVM alias or derived long-zero address when crafting ERC-20 transfers.

### ❓ Context

- https://ledgerhq.atlassian.net/browse/BACK-10485

---

### DIFF
--- a/libs/ledger-live-common/src/families/hedera/__snapshots__/bridge.integration.test.ts.snap
+++ b/libs/ledger-live-common/src/families/hedera/__snapshots__/bridge.integration.test.ts.snap
@@ -171,7 +171,7 @@ exports[`hedera currency bridge scanAccounts hedera seed 1 1`] = `
       "delegation": {
         "delegated": "333985159",
         "nodeId": 18,
-        "pendingReward": "143829",
+        "pendingReward": "164370",
       },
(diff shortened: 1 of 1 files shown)


acewf (dismissed):

mdomanski-ext-ledger (dismissed):
Looks good to me, but I'm wondering if we should support the `AccountId.fromString(accountId).toEvmAddress()` fallback if mirror fetch fails. 

[Docs](https://docs.hedera.com/hedera/core-concepts/accounts/account-properties#value-transfer-behavior-with-account-identifiers) says:
>EVM Transactions: The address used for a value transfer recipient in EVM transactions (ContractCreate, ContractCall, and EthereumTransaction) must be the main one for the account. This will be the ECDSA-derived EVM address alias if set on the account. If not, the long zero address (account number alias) will be used. This ensures compatibility with the EVM’s expected behavior for account addresses.

So I think that transaction could later fail anyway if we get long-zero address for account having proper non-long-zero alias. It also makes this a little more challenging to unit test as we have the dependency from the same module (toEVMAddress)

PR #13818, opened 2026-01-20
ci: add workflow to auto-update PR titles from labels

desirendr:
### ✅ Checklist

## What
Adds a GitHub Action that automatically prefixes PR titles based on applied labels.
## Why
Improves PR readability and helps quickly identify which areas of the codebase are affected (Desktop, Mobile, Common).
## How
When labels are added/removed or a PR is opened, the workflow updates the title with the corresponding prefix:

| Label | Prefix |
|-------|--------|
| `desktop` | `[LWD]` |
| `mobile` | `[LWM]` |
| `common` | `[LWDM]` |
| `desktop & mobile` | `[LWDM]` |

**Example:**
- Labels: `desktop`, `common`
- Title: `Fix sync issue` → `[LWDM] Fix sync issue`

## Notes

- Works alongside the existing `labeler.yml` workflow
- Uses the same GitHub App token for authentication
- Only updates title if a change is needed

### DIFF
--- a/.github/workflows/pr-title-from-labels.yml
+++ b/.github/workflows/pr-title-from-labels.yml
@@ -1,0 +1,78 @@
+name: "[CI] - PR Title - Auto Update from Labels"
+run-name: "@PR • Title update triggered by ${{ github.actor }} ${{ format('on branch {0}', github.head_ref) }}"
+
+on:
+  pull_request_target:
+    types: [labeled, unlabeled]
+
+jobs:
+  update-title:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-22.04
+    steps:
+      - name: generate token
+        id: generate-token
+        uses: tibdex/github-app-token@v1
+        with:
+          app_id: ${{ secrets.GH_BOT_APP_ID }}
+          private_key: ${{ secrets.GH_BOT_PRIVATE_KEY }}
+
+      - name: Update PR Title from Labels
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.generate-token.outputs.token }}
+          script: |
+            const pr = context.payload.pull_request;
+            const labels = pr.labels.map(l => l.name);
+
+            // Build prefix from labels
+            // If 'common' OR both 'desktop' + 'mobile' are present, use only LWDM
+            let prefixes = [];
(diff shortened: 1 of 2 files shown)


Justkant:
Hey thx for the PR, I think it's fine as a first step but it won't cover some cases

The tags do not guarantee that any changes that would impact common, desktop or mobile are taken into account
For example:
 - A PR on a coin module: https://github.com/LedgerHQ/ledger-live/pull/13873/changes
 - A PR for any lib from the mono-repo used by common, desktop or mobile: https://github.com/LedgerHQ/ledger-live/pull/13863/changes

desirendr:
Hey @Justkant thanks for the feedback, would it be worth adding labels for these blind spots ?
I've noticed most of them come from the libs folder like here with the DMK https://github.com/LedgerHQ/ledger-live/pull/13793/files.

New rule could be that I consider coin-modules as common as they affect both platforms.
Libs seem to be more granular.

Justkant:
coin-modules seems fine to add, for the rest it's trickier and we don't have a label for everything, it's going to be some burden to maintain if we add or remove libs

Common, mobile, desktop and coin-modules don't really change so it's fine

But keep in mind that we might move the coin-modules outside the mono-repo soon too (in that case we would also update common when update the lib)

.github/workflows/pr-title-from-labels.yml
```diff
@@ -0,0 +1,77 @@
+name: "[CI] - PR Title - Auto Update from Labels"
+run-name: "@PR • Title update triggered by ${{ github.actor }} ${{ format('on branch {0}', github.head_ref) }}"
+
+on:
+  pull_request_target:
+    types: [labeled, unlabeled, opened]
```
lewisd5: I don't think we need the "opened" type as you're parsing the labels on the PR and we only want to run this when the labels have been generated

gre-ledger:
i'm not sure i really understand the Why, the labels are even more visible than the title when it comes to listing PRs 🤔  is it maybe more for when we see it from Jira side?
does it mean than most PRs are going to be [LWDM] prefix as soon as it touches one library?

desirendr:
@gre-ledger the labels can't be read by the AI agent, this was my obvious first choice too but it's not possible as of now.
And I guess changes in coin modules and common will, not the others even if it would be best for 100% coverage.

PR #13840, opened 2026-01-20
Feat(LIVE-24184): Update Back to quotes' logic after successful tx

ooke-ledger:
### ✅ Checklist

  - ...

### 📝 Description

This PR updates Back to quotes' logic after successful tx

| Before        
Screenshot 2026-01-20 at 16 36 59

| After         
Screenshot_1768930253

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-24184

---

### DIFF
--- a/apps/ledger-live-mobile/src/screens/Swap/LiveApp/WebView.tsx
+++ b/apps/ledger-live-mobile/src/screens/Swap/LiveApp/WebView.tsx
@@ -71,14 +71,13 @@ export const WebView = forwardRef<WebviewAPI, Props>(
     const initialSource = useMemo(() => currentRouteNameRef.current || "", []);
 
     return (
-
       <SafeAreaView edges={["bottom"]} isFlex>
         <Web3AppWebview
           ref={ref}
           manifest={manifest}
           customHandlers={customHandlers}
           onStateChange={setWebviewState}
-                      allowsBackForwardNavigationGestures={false}
+          allowsBackForwardNavigationGestures={false}
           inputs={{
             source: initialSource,
             swapApiBase: SWAP_API_BASE,


apps/ledger-live-mobile/src/screens/Swap/LiveApp/navigationHandlers/useSwapHeaderNavigation.tsx
```diff
@@ -68,6 +75,21 @@ export function useSwapHeaderNavigation(webviewRef: React.RefObject<WebviewAPI |
     navigation.goBack();
   }, [navigation, track]);
 
+  const navigateToSwapForm = useCallback(() => {
```
philipptpunkt, line 75: Just some thoughts - If we don't replace the navigation history this means if the user uses "swipe back" we will still navigate back to the quotes list?!
So how about resetting the history and fill only the swap-form route/url in? Then the "x" button would just be a "back" navigation and the "swipe back" would still work

apps/ledger-live-mobile/src/components/Web3AppWebview/helpers.ts
```diff
@@ -294,6 +294,10 @@ export function useWebviewState(
         loadURL: (url: string): void => {
           setURI(url);
         },
+        resetToInitialURL: (): void => {
+          const webview = safeGetRefValue(webviewRef);
+          webview.injectJavaScript(`window.location.replace('${initialURL}'); true;`);
```
Justkant: It feels a bit wrong to inject js in the webview to control the navigation like that, is the idea to keep the page in the webview loaded and only do a js navigation to avoid reloading the app ?

ooke-ledger: We need `location.replace()` to clear the history entry so swipe-back exits the swap flow instead of going to the quotes list. react-native-webview doesn't have a native "replace" method, only "load" (which pushes to history).
Open to alternatives if there's a cleaner native approach.

Justkant: So the issue you're trying to solve is that you want the swipe back to not move the user back to the quotes in the webview but to actually leave the webview

Do you need/want the `allowsBackForwardNavigationGestures` on the webview ?
You could simply disable it and it would always leave the webview (recover disables it)

I'm not sure there is another solution where you could intercept the gesture to manually choose to handle the swipe back in the webview or in the native view depending on the current URL of the webview or another state letting you know you are done
This could also leave directly and only on the webview used for swap

ooke-ledger: Good point - I've disabled `allowsBackForwardNavigationGestures` on the swap webview. Users can still navigate back within the webview using the header back button, and since recover uses this approach, it's consistent. Simpler than the JS injection workaround.

Justkant: Do you still need the `resetToInitialURL` now then ?

ooke-ledger: Yes, we still need it to navigate back to the swap form with all the original query params (theme, lang, swapApiBase, etc.). The X button after two-step-approval completion calls `resetToInitialURL()` to return to the form.

apps/ledger-live-mobile/src/screens/Swap/LiveApp/navigationHandlers/useSwapNavigationHelper.tsx
```diff
@@ -28,8 +28,22 @@ export function useSwapNavigationHelper({ navigation }: { navigation: Navigation
 
         let canGoBack = canGoBackFromEvent;
 
-        if (urlFromEvent.includes("two-step-approval")) {
+        const isTwoStepFlow =
```
philipptpunkt, line 31: we should probably already include the new "/multi-step-transaction" route

apps/ledger-live-mobile/src/screens/Swap/LiveApp/WebView.tsx
```diff
@@ -81,6 +81,7 @@ export const WebView = forwardRef<WebviewAPI, Props>(
             manifest={manifest}
             customHandlers={customHandlers}
             onStateChange={setWebviewState}
+            allowsBackForwardNavigationGestures={false}
```
philipptpunkt: 🤨

philipptpunkt (dismissed):

Justkant (dismissed):

PR #13864, opened 2026-01-21
fix(ci): test affected to the wrong variable

henri-ly:
### ✅ Checklist

  - ...

### 📝 Description

Fix wrong affected variable

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/.changeset/sour-goats-teach.md
+++ b/.changeset/sour-goats-teach.md
@@ -1,0 +1,5 @@
+---
+"@actions/live-common-affected": minor
+---
+
+Fix coin module integration test on schedule


.github/workflows/test-coin-modules-integ.yml
```diff
@@ -23,7 +23,7 @@ permissions:
 jobs:
   is-affected:
     outputs:
-      is-affected: ${{ steps.affected.outputs.is-affected }}
```
Justkant, line 26: Should we update the other workflows using `is-affected` ?
also why is it the only output to use `steps.set-affected` ?

henri-ly: From what I'm comparing it's seems that only this workflow use the `set-affected` steps, so we don't need to change `is-affected`.

and the `steps.set-affected` seems to force value to `true` if workflow is running through `schedule` otherwise it's keep value of `steps.affected`.

But couldn't explain why it was done this way and not like other flow. Maybe @qperrot can answered you

Justkant: The `set-affected` is only used for scheduled workflow so to keep it working for the others we need to still use `steps.affected`

Justkant: Forget my last comment I see that we also add the base affect in set-affected so it should be the correct output as you said

Justkant (approved):

PR #13873, opened 2026-01-21
[LWDM] feat: add AlpacaApi getBlock and getBlockInfo to coin-xrp module

pvoliveira:
### ✅ Checklist

  - AlpacaApi - getBlock

### 📝 Description

_Implements the AlpacaApi getBlock method for the coin-xrp module._

### ❓ Context

- **LIVE-24974**: https://ledgerhq.atlassian.net/browse/LIVE-24974 

---


libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,71 @@
...
+function mapBlockTransactions(txs: XrplOperation[]): BlockTransaction[] {
+  return txs.map(blk => {
+    const operations: TransferBlockOperation[] = [];
+
+    blk.meta.AffectedNodes?.forEach((op: Record<string, any>) => {
+      const prop = Object.keys(op)[0];
+      const op_data = op[prop];
+
+      // Handle XRP balance changes (AccountRoot)
+      if (
```
Wozacosta: CreatedNode` has `NewFields`, not `PreviousFields`/`FinalFields`. (see: https://github.com/LedgerHQ/ledger-live/pull/13873/changes#diff-021037f3c69a2789b9b642223ad86434be299070bb5f0402fc82bfda87321481R29-R32)


 So the condition passes but the logic always fails silently as `before` and `after` are both `undefined`.

pvoliveira: Changed, thanks! Still, I didn't find in the documentation if the CreatedNode can have the LedgerEntryType == "AccountRoot"...

libs/coin-modules/coin-xrp/src/logic/getBlock.test.ts
```diff
@@ -0,0 +1,479 @@
...
+          },
+        ],
+      },
+    });
+
+    // When
+    const result = await getBlock(blockIndex);
+
+    // Then
+    expect(result.transactions).toHaveLength(1);
```
acewf: Francois-guerin, in some other PRs was pushing to have explicit objects defined, instead of testing single properties.

I feel it can become too verbose. (Just leaving it has a note here)

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,141 @@
...
+import type {
+  AffectedNode,
+  LedgerTxResponse,
+  XrplOperation,
+  ModifiedNodeData,
+  CreatedNodeData,
+} from "../network/types";
+
+function pushTransferOperation(
+  operations: BlockOperation[],
```
acewf: bit odd, passing a mutable array, just then push some data into it.

Would make more sense this function, just processes the data, and returns the operation schema or not.

acewf: odd in the sense that becomes harder to track what is mutating data, since is scattered around the code.

pvoliveira: Changed, thanks!

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,148 @@
...
+  return [{ type: "other", ...data }];
+}
+
+/**
+ * Maps a CreatedNode to operations.
+ * Returns transfer operation for new AccountRoot entries, or "other" operation for other entry types.
+ */
+function mapCreatedNode(data: CreatedNodeData, fee: bigint, feesPayer: string): BlockOperation[] {
+  if (data.LedgerEntryType === "AccountRoot") {
+    const balance = data.NewFields?.Balance as string;
```
acewf: better to have a method to create a string, then doing an assertion that doesn't guarantee a string 
```
string(data.NewFields?.Balance)
```

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,148 @@
...
+    const account = data.NewFields?.Account as string;
+
+    if (typeof balance === "string" && account) {
```
acewf: kind of checking the typeof balance where, where you should be casting to a string in previous lines.

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Maps a DeletedNode to operations.
+ * AccountRoot deletions don't create operations; other entry types create "other" operations.
+ */
+function mapDeletedNode(data: DeletedNodeData): BlockOperation[] {
+  if (data.LedgerEntryType === "AccountRoot") {
+    return [];
+  }
+  return [{ type: "other", ...data }];
```
acewf: should be this type declared as constant somewhere in the code? also odd to have a type as "other"

pvoliveira: BlockOperation can be one of these types: transfer or other. I can move the string to a constant to make it clear.

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
pvoliveira: if Balance exists, when it's a string, it will contain the amount, but as it depends on the type of the node, NewFields may have different fields, for that reason, I'm not properly enforcing a type for the property.

I'll keep the type test in the if condition in this case.

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
pvoliveira: I kept this check and removed the casting.

libs/coin-modules/coin-xrp/src/logic/getBlock.integ.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        type: "transfer" as const,
+        address: "refV71jPdVAK3vQQb77SQrCJ2ibbov2H8",
+        amount: 999000000n,
+        asset: {
+          type: "native",
+          name: "XRP",
+        },
+      },
+    ];
+    expect(tx0.operations).toEqual(
```
Moustafa-Koterba: **nit** Good assertion, but I think we can change the expected variable to an object to avoid mapping (not necessary)
```suggestion
    expect(tx0.operations).toEqual(
        expect.arrayContaining(op)),
    );
```

Plus, does this block contains this contains only one operation ? Then we can make an assertion on size to ensure that you have only this element

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (diff === BigInt(0)) return null;
+
+  // Adjust the fee payer's balance diff to exclude fees
+  // (fees are reported separately in BlockTransaction.fees)
+  let adjustedDiff = diff;
+  if (account === feesPayer) {
+    adjustedDiff = diff + fee;
+  }
+
+  if (adjustedDiff === BigInt(0)) return null;
```
Moustafa-Koterba: I am sorry, but I dont see how this case can happen ? Cann `fee` or `diff` be negative ?

libs/coin-modules/coin-xrp/src/network/getLedgerByIndex.test.ts
Moustafa-Koterba, line 1: Can you name this file `index.test.ts` ? Since it tests the function `getLedgerByIndex` from `index.ts`

libs/coin-modules/coin-xrp/src/network/getLedgerByIndex.test.ts
```diff
@@ -0,0 +1,81 @@
...
+          },
+        },
+      },
+    });
+
+    // When
+    const result = await getLedgerByIndex(blockIndex);
+
+    // Then
+    expect(result.ledger_index).toBe(blockIndex);
```
Moustafa-Koterba: Can you use `toEqual` on the whole object instead of testing each attribute ? It will ensure that the returned object is exactly what we expect, without other attributes
```ts
expect(result).toEqual({
```

Otherwise, If you dont want to test all the attributes, you can use `toMatchObject` (dont know the name perfectly)

libs/coin-modules/coin-xrp/src/api/index.ts
```diff
@@ -45,9 +45,7 @@ export function createApi(config: XrpConfig): Api<XrpMapMemo> {
     lastBlock,
     listOperations: operations,
     validateIntent,
-    getBlock(_height): Promise<Block> {
-      throw new Error("getBlock is not supported");
-    },
+    getBlock,
     getBlockInfo(_height: number): Promise<BlockInfo> {
```
jnicoulaud-ledger: `getBlockInfo` needs to be implemented as well 🙏

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  return [{ type: BLOCK_OPERATION_OTHER, ...data }];
+}
+
+function mapBlockInfo(result: LedgerTxResponse): BlockInfo {
+  return {
+    height: result.ledger_index,
+    hash: result.ledger.ledger_hash,
+    time: new Date(result.ledger.close_time_iso),
```
jnicoulaud-ledger, line 98: Please provide the parent block hash, it should be `ledger.parent_hash`

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
pvoliveira: The idea was not to generate operations for fee-only transactions, but after some tests, fetching data from the network, it seems that we will not have fee-only transactions with an AffectedNode with the LedgerEntryType as AccountRoot, so I removed this test.

libs/coin-modules/coin-xrp/src/logic/getBlock.integ.test.ts
pvoliveira: Changed.

libs/coin-modules/coin-xrp/src/network/getLedgerByIndex.test.ts
pvoliveira: Done.

libs/coin-modules/coin-xrp/src/network/getLedgerByIndex.test.ts
pvoliveira: Done.

libs/coin-modules/coin-xrp/src/api/index.ts
pvoliveira: Added.

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
pvoliveira: Added.

libs/coin-modules/coin-xrp/src/logic/getBlock.ts
```diff
@@ -0,0 +1,148 @@
...
+  if (account === feesPayer) {
+    adjustedDiff = diff + fee;
+  }
+
+  return {
+    type: BLOCK_OPERATION_TRANSFER,
+    address: account,
+    asset: ASSETINFO_NATIVE,
+    amount: adjustedDiff,
+  };
```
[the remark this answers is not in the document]
pvoliveira: This method is being called for the ModifiedNode node, and LedgerEntryType === "AccountRoot" items, and the API doesn't return any meta.AffectedNode with items matching those properties for fee-only transactions.

PR #13887, opened 2026-01-22
[LWDM] zcash-shielded lib: decrypting shielded transactions

alevito:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  - the zcash-shielded library, and the changes in this PR, are not currently used anywhere in Ledger-Live. It will be enabled via feature flag for the purpose of development.

### 📝 Description

The method `ZCash.decryptTransaction` adds the functionality of decrypting the shielded part of a ZCash transaction.
It returns the `transfer_type`, `memo` and the `amount` of the transaction's `orchard_outputs` and/or `sapling_outputs`.
**Note:** depends on [`@ledgerhq/zcash-decrypt`](https://github.com/LedgerHQ/zcash-decrypt).

### ❓ Context

- **JIRA**: [LIVE-24171](https://ledgerhq.atlassian.net/browse/LIVE-24171)

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #13927, opened 2026-01-23
[LWDM] LIVE-24193: zcash lib parse block shielded part

alevito:
### ✅ Checklist

  - the zcash-shielded library, and the changes in this PR, are not currently used anywhere in Ledger-Live. It will be enabled via feature flag for the purpose of development.

### 📝 Description

The method `ZCash.processBlock` adds the functionality of scanning a block for shielded transactions that are matching a given UFVK viewing key.
It returns a list of transactions for the given viewing key.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/libs/coin-modules/zcash-shielded/src/ZCash.ts
+++ b/libs/coin-modules/zcash-shielded/src/ZCash.ts
@@ -1,8 +1,7 @@
 import { log } from "@ledgerhq/logs";
 import { decrypt_tx, DecryptedTransaction } from "@ledgerhq/zcash-decrypt";
-import type { ShieldedTransaction } from "./shieldedTransaction";
 import { Block, JsonRpcClient } from "./jsonRpcClient";
-import { toShieldedTransaction } from "./shieldedTransaction";
+import { toShieldedTransaction, ShieldedTransaction } from "./shieldedTransaction";
 import { LOG_TYPE } from "./constants";
 
 /**
@@ -71,6 +70,10 @@ export default class ZCash {
     // 2. retrieve each tx hash
     if (transactions) {
       for (const txId of transactions) {
+        if (!txId) {
+          continue;
+        }
+
         const tx = await this.jsonRpcClient.getRawTransaction(txId);
 
         // 3. call decryptTransaction for each tx hash containing orchard actions
--- a/libs/coin-modules/zcash-shielded/src/shieldedTransaction.ts
+++ b/libs/coin-modules/zcash-shielded/src/shieldedTransaction.ts
@@ -7,7 +7,7 @@ export type ShieldedTransaction = {
   height: number;
   blockHash: string;
   time: number;
-  fee: number; // BigNumber as string
+  fee: number;
   decryptedData?: DecryptedTransaction; // Decrypted transaction data
 };
 
--- a/libs/coin-modules/zcash-shielded/tests/ZCash.test.ts
+++ b/libs/coin-modules/zcash-shielded/tests/ZCash.test.ts
@@ -131,4 +131,13 @@ describe("findShieldedTxsInBlock", () => {
     });
     expect(transactions).toEqual([]);
   });
+
+  test("edge case: retrieves an empty list when the block has emtpy tx", async () => {
+    const zcash = new ZCash({ nodeUrl: JSON_RPC_SERVER });
+    const transactions = await zcash.findShieldedTxsInBlock({
+      block: { ...blockWithoutAnyTx, tx: ["", ""] },
+      viewingKey: testAccount1.viewingKey,
+    });
+    expect(transactions).toEqual([]);
+  });
 });


libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -18,6 +20,36 @@ export default class ZCash {
     return (endBlock - startBlock) * ZCash.AVERAGE_BLOCK_SYNC_TIME_MS;
   }
 
+  async processBlock(blockHex: string, viewingKey: string): Promise<string[]> {
```
alevito: rename to something like processShieldedBlock

alevito: done in latest commit

alevito: @may01 , the function `processBlock` has been renamed `findShieldedTxInBlock`

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
```diff
@@ -0,0 +1,158 @@
...
+    monitored: true;
+  };
+  valuePools: object[];
+  trees: { sapling: object; orchard: object };
+  previousblockhash: string;
+  nextblockhash: string;
+};
+
+export type RawTransaction = {
+  in_active_chain: boolean;
```
semeano, line 54: Any reason for this to be in snake_case?

```suggestion
  inActiveChain: boolean;
```

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
```diff
@@ -0,0 +1,158 @@
...
+  };
+  valueBalance: number;
+  valueBalanceZat: number;
+  size: number;
+  time: number;
+  txid: string;
+  authdigest: string;
+  overwintered: true;
+  version: number;
+  versiongroupid: string;
```
semeano, line 106: Same issue here. Does this need to be like that or can we keep to one convention like camelCase?

libs/coin-modules/zcash-shielded/tests/jsonRcpClient.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { server } from "../mocks/node";
+import { HttpResponse, http } from "msw";
+
+beforeAll(() => server.listen({
+  onUnhandledRequest: 'error',
+}));
+afterEach(() => server.resetHandlers());
+afterAll(() => server.close());
+
+const JSON_RPC_SERVER = "http://localhost:18232";
```
semeano: Shouldn't this be pointing to the backend? It will fail in CI.

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
```diff
@@ -0,0 +1,158 @@
...
+    chainValueZat: number;
+    monitored: true;
+  };
+  valuePools: object[];
+  trees: { sapling: object; orchard: object };
+  previousblockhash: string;
+  nextblockhash: string;
+};
+
+export type RawTransaction = {
```
semeano, line 53: Do we need all these props?

libs/coin-modules/zcash-shielded/mocks/handlers.ts
```diff
@@ -0,0 +1,58 @@
...
+              result: {
+                ...block,
+              },
+            });
+          }
+        }
+
+        return HttpResponse.json({
+          error: {
+            code: -5,
```
may01: Lets replace error code with const.

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -18,6 +22,39 @@ export default class ZCash {
...
+    const transactions = block?.tx;
+    const decryptedTransactions = [];
+
+    // 3. retrieve each tx hash
+    if (transactions) {
+      for (const txId of transactions) {
+        const tx = await jsonRpcClient.getRawTransaction(txId);
+
+        // 4. call decryptTransaction for each tx hash containing orchard actions
+        if (tx?.orchard.actions.length) {
```
may01, line 80: We should check for sapling as well.

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -18,6 +22,39 @@ export default class ZCash {
...
+          try {
+            const decryptedTx = await this.decryptTransaction(tx?.hex, viewingKey);
+            decryptedTransactions.push(toShieldedTransaction(tx, decryptedTx));
+          } catch (_e) {
+            console.warn("could not decrypt memo");
```
may01: Will this be logged for each unsuccessful tx?

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -1,11 +1,15 @@
...
+import { JsonRpcClient } from "./jsonRpcClient";
+import { toShieldedTransaction } from "./shieldedTransaction";
 
 /**
  * ZCash API
  */
-
 export default class ZCash {
   static readonly AVERAGE_BLOCK_SYNC_TIME_MS = 5;
+  static readonly JSON_RPC_SERVER = "http://localhost:18232";
```
may01: We should use currency config url here

libs/coin-modules/zcash-shielded/tests/jsonRcpClient.test.ts
may01: does this points to the `libs/coin-modules/zcash-shielded/mocks/node.ts`?
Suggest to have only one variable that will set the expected server url
https://github.com/LedgerHQ/ledger-live/pull/13927/changes#diff-6e5c77ad08c5dd6654790d16c21346ec4448eb47f3a858e005cf1ac9d8a21eb5R4

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: These types have been derived from the HTTP responses served by the JSON RPC server. They are used in the JSON RPC client only.
We aren't currently using all the props in the response, but I thought it would be useful to have the full structure of the responses.

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: The reason why this is not in camel case is that this is exactly the response received from the server. The JsonRpcClient module is not restructuring the response retrieved from the server, but we could this functionality could be added.

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: See https://github.com/LedgerHQ/ledger-live/pull/13927/changes#r2728550296

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: I'm not 100% sure whether cipherscan's zcash-decoder can actually decrypt sapling transactions.
Are actions used also in sapling, or are they specific for orchard only?

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: Yes, it would. I can remove this and replace it with a unique log message emitted when the block is completely parsed.

libs/coin-modules/zcash-shielded/tests/jsonRcpClient.test.ts
alevito: The `JSON_RPC_SERVER` constant should be removed completely. The tests use a msw mock server which interceps requests to the backend and responds with mocked responses.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: This issue is now resolved.
We added support for sapling transactions in [LIVE-24171](https://github.com/LedgerHQ/ledger-live/pull/13887)

libs/coin-modules/zcash-shielded/tests/jsonRcpClient.test.ts
alevito: For the time being JSON_RPC_SERVER is defined in consts.
The `nodeUrl` string is currently passed as a parameter when the instance of ZCash is created by the consumer. Is this what we want?

libs/coin-modules/zcash-shielded/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const LOG_TYPE = "zcash-shielded";
```
semeano, line 1: Can you rename this file to `constants.ts` to be consistent with other modules. I also created a `constants.ts` file in my PR and this way we can merge them.

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
```diff
@@ -0,0 +1,153 @@
...
```
semeano, line 53: Are we going to use this type somewhere else? If so, it might make sense to move it into a `types.ts` file.

libs/coin-modules/zcash-shielded/src/shieldedTransaction.ts
```diff
@@ -0,0 +1,25 @@
+import { DecryptedTransaction } from "@ledgerhq/zcash-decrypt";
+import { RawTransaction } from "./jsonRpcClient";
+
+export type ShieldedTransaction = {
```
semeano, line 4: Same here.

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -39,6 +49,48 @@ export default class ZCash {
...
   }
 
+  /**
+   * Scans a block for shielded transactions matching the given viewing key.
+   *
+   * @param {string} blockHex Block hex
+   * @param {string} viewingKey the UFVK - unified full viewing key.
+   * @returns {ShieldedTransaction[]} list of shielded transactions
+   */
+  async findShieldedTxsInBlock(
```
may01: It will be better to pass block's data as an parameter. It will allow us to process batch of blocks in one take.

libs/coin-modules/zcash-shielded/src/constants.ts
alevito: Thanks! Done in latest commit

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: @semeano 
Good idea, although I'd rather keep the type colocated with the context they belong to, in this case `jsonRpcClient.ts`, as they are specific of this module.
I'd rather re-export the types from `types.ts` in case they are ever needed by the consumer of the `zcash-shielded` module.

e.g.,
```typescrypt
// file: type.ts
export { RawTransaction } from "./src/jsonRpcClient";
```

Sounds good?

libs/coin-modules/zcash-shielded/src/constants.ts
```diff
@@ -0,0 +1,2 @@
...
+export const JSON_RPC_SERVER = "http://localhost:18232";
```
may01: We should target proper ledger service.

libs/coin-modules/zcash-shielded/tests/mocks/handlers.ts
may01, line 1: It's better to move mocks into `test` folder

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    if (data.error || !data.result) {
+      log(LOG_TYPE, `error: Empty dataonse result from server - ${data?.error?.message}`);
+    } else {
+      return data.result;
+    }
+  }
+
+  getBlock(blockHash: string) {
```
may01, line 140: We also should be able to get range of blocks.

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
semeano: Sure!

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: agreed: Might not be possible for the time being. Lets leave it aside for the time being.

libs/coin-modules/zcash-shielded/src/constants.ts
alevito: Agreed: update the const value to the ledger service if it works. Otherwise leave localhost as example.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: Done: Updated findShieldedTxsInBlock to accept a Block argument, instead of a blockHash argument,.

libs/coin-modules/zcash-shielded/src/constants.ts
alevito: Done: update the RPC server URL to ledger service.

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: Verified that this is not currently possible.
ZCash 6.2.0 RPC getBlock call doesn't accept a block range  as argument.
See https://zcash.github.io/rpc/getblock.html

libs/coin-modules/zcash-shielded/tests/jsonRcpClient.test.ts
alevito: Resolved: Discussed over slack and agreed to keep `JSON_RPC_SERVER` in consts, and its value is the url to the legder node RPC service.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: Resolved: moved this to constants.ts

libs/coin-modules/zcash-shielded/src/jsonRpcClient.ts
alevito: Done in latest commit.
* `src/types.ts` exports all the module types.
* the compiled `lib/types.d.ts` is exposed by the package to the consumers - i.e., types field in package.json.

libs/coin-modules/zcash-shielded/mocks/handlers.ts
alevito: Done! Replaced with a const

libs/coin-modules/zcash-shielded/tests/mocks/handlers.ts
alevito: done

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: this `try..catch` block was removed as the exceptions are already handled by `decryptTransaction`, which doesn't throw.

may01 (dismissed):

libs/coin-modules/zcash-shielded/jest.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   coverageDirectory: "coverage",
   testEnvironment: "node",
   testPathIgnorePatterns: ["lib/", "lib-es/", ".*\\.integration\\.test\\.[tj]s"],
+  transformIgnorePatterns: ["node_modules/(?!cipherscan)/"],
```
alevito, line 7: remove this transformIgnorePatterns as it's not required anymore

```suggestion
```

libs/coin-modules/zcash-shielded/src/shieldedTransaction.ts
```diff
@@ -0,0 +1,25 @@
...
+  id: string;
+  hex: string;
+  height: number;
+  blockHash: string;
+  time: number;
+  fee: number; // BigNumber as string
```
qperrot: I don't understand this comment

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -1,5 +1,9 @@
-import { decrypt_tx, DecryptedTransaction } from "@ledgerhq/zcash-decrypt";
 import { log } from "@ledgerhq/logs";
+import { decrypt_tx, DecryptedTransaction } from "@ledgerhq/zcash-decrypt";
+import type { ShieldedTransaction } from "./shieldedTransaction";
+import { Block, JsonRpcClient } from "./jsonRpcClient";
+import { toShieldedTransaction } from "./shieldedTransaction";
```
qperrot: import from ./shieldedTransaction line 3 and 5

libs/coin-modules/zcash-shielded/src/ZCash.ts
```diff
@@ -39,6 +49,45 @@ export default class ZCash {
...
+    const { block, viewingKey } = args;
+
+    // 1. get list of tx
+    const transactions = block?.tx;
+    const decryptedTransactions = [];
+
+    // 2. retrieve each tx hash
+    if (transactions) {
+      for (const txId of transactions) {
+        const tx = await this.jsonRpcClient.getRawTransaction(txId);
```
qperrot, line 77: what if `txId` is "" ? is it possible ?

alevito: Thanks, I think it's an edge case, but we need to handle it.
It should skip to the next iteration is `txId` is empty.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: Will fix this.

libs/coin-modules/zcash-shielded/src/shieldedTransaction.ts
alevito: Will fix this.

libs/coin-modules/zcash-shielded/src/ZCash.ts
alevito: I fixed this in commit 0fa3577116913e84e9e1b05e95cfe82c506231d7, and added a unit test for this edge case.

PR #14001, opened 2026-01-27
[LWDM] PERF: rework operations retention to 20 ops for Account and sub accounts

gre-ledger:
### ✅ Checklist

  - LLD and LLM store way less historical operations in the storage

#### Facts

- given that Ledger Wallet don’t show more than 20 operations at first page
- given that a full sync will restore at the first full sync all operations
- given that the balanceHistoryCache have stored the datapoint calculating the graph, so we can still show the graph if operations is []

#### Proposal

- we rework opRetentionFilter to only keep 20 operations per account – regardless of op date. we also start applying the logic inside subAccounts#operations
- if we want to be safe, we can keep 40, we have one page in advance and enough time for full sync to run.
- we are therefore safe, at boot to load Portfolio or Account pages
- https://github.com/LedgerHQ/ledger-live/pull/14127 must be going with this evolution

#### metrics

on our dataset, with a conservative value of 40, we reduced the relative size from **2.64 MB to 0.7 MB**, which means for this specific dataset we had, it's 3-4x smaller.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-25362

---

### DIFF
--- a/.changeset/proud-ravens-brush.md
+++ b/.changeset/proud-ravens-brush.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
+---
+
+reduced operations retention to strictly max 40 ops per account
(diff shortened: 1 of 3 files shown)


gre-ledger:
@jnicoulaud-ledger it will redo a sync, like today's. to be confirmed but today it's already doing this, it's capped at 500 ops max today and we are reducing to 20. would be worth having a call to understand what you mean by rpc calls because i'm not familiar with all implementations.

gre-ledger:
We can't do this yet unfortunately. coin implementations don't pull the older operations history. this pagination will be implemented in future by https://ledgerhq.atlassian.net/browse/LIVE-22350 and at that time we can include this optimisation

PR #14015, opened 2026-01-27
[LWDM] PERF: optimize countervalues export size dropping non observed datapoints

gre-ledger:
### ✅ Checklist

  - graph
  - historical countervalues

### 📝 Description

Optimize the countervalues exported in app.json to only select what user observes.

in our 16MB app.json dataset, we can drastically optimize; with all the other optimize:
in a **1 MONTH** time window, that optim on Daily alone moves me **from 3.4 MB to 0.8 MB**.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-25408

---

### DIFF
--- a/libs/live-countervalues/src/helpers.ts
+++ b/libs/live-countervalues/src/helpers.ts
@@ -1,5 +1,6 @@
 import { Currency } from "@ledgerhq/types-cryptoassets";
 import type { RateGranularity } from "./types";
+import type { PortfolioRange } from "@ledgerhq/types-live";
 
 export const inferCurrencyAPIID = (currency: Currency): string => {
   switch (currency.type) {
@@ -28,9 +29,25 @@ export const datapointLimits: Record<RateGranularity, number> = {
   hourly: 7 * DAY, // we fetch at MOST a week of hourly. after that there are too much data...
 };
 
-export const datapointRetention: Record<Extract<RateGranularity, "hourly">, number> = {
+export const datapointRetention: Record<RateGranularity, number> = {
   hourly: 7 * DAY, // we keep hourly data for 7 days
+  daily: 9999 * DAY, // we keep daily data for a very long time by default, can be overridden by user settings
 };
+
+export function portfolioRangeToDays(range: PortfolioRange): number | undefined {
+  switch (range) {
+    case "day":
+      return 1;
+    case "week":
+      return 7;
+    case "month":
+      return 30;
+    case "year":
+      return 365;
+    case "all":
+      return undefined;
+  }
+}
 
 /**
  * efficient implementation of YYYY-MM-DD formatter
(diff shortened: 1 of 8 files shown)


libs/live-countervalues/src/helpers.ts
```diff
@@ -28,10 +29,32 @@ export const datapointLimits: Record<RateGranularity, number> = {
...
+  switch (range) {
+    case "day":
+      return 1;
+    case "week":
+      return 7;
+    case "month":
+      return 30;
+    case "year":
+      return 365;
+    case "all":
```
mcayuelas-ledger, line 47: we have a lot of helpers for this no? or maybe a bit different ? could be great to harmonize them no?

libs/live-countervalues/src/logic.ts
```diff
@@ -27,19 +27,31 @@ import {
...
   { data, status }: CounterValuesState,
   trackingPair: TrackingPair[],
+  selectedTimeRange?: PortfolioRange,
 ): CounterValuesStateRaw {
   const out = { status } as CounterValuesStateRaw;
   const hourlyLimit = formatCounterValueDay(new Date(Date.now() - datapointRetention.hourly));
   const pairIds = new Set(trackingPairIds(trackingPair));
 
+  // Only filter daily data if selectedTimeRange is provided and not "all"
+  // If undefined or "all", we keep all daily data
```
LL782: Personally I feel the information from the comment is well provided by the tests and the implementation code itself. I generally see comments as a code-smell and in my opinion we should skip them when we can


```suggestion
```

libs/live-countervalues/src/types.ts
```diff
@@ -16,6 +17,10 @@ export type CountervaluesSettings = {
   disableAutoRecoverErrors?: boolean;
 
   granularitiesRates?: Record<RateGranularity, number>;
+  // PortfolioRange selected by the user for graph view
+  // Used to determine how much daily countervalue data to keep
+  // If undefined or "all", no filtering is applied (keeps all daily data)
+  selectedTimeRange?: PortfolioRange;
```
LL782: Will this comment be kept up to date if logic changes?
Will the comment be trusted or will the reader want to look at where it's used anyway?
Is the user is a AI is it better to get understanding from the code or from comments?

Personally I would remove this. WDYT?


```suggestion
  selectedTimeRange?: PortfolioRange;
```

LL782 (dismissed):
All looks good to me. I've left the same comment about comments in a couple of places. It would be good to sync opinions on this one but they're not blocking comments at all :)

libs/live-countervalues/src/helpers.ts
gre-ledger: this one is different in that it defines the default data retention for a given range. do we have such data somewhere else? i'm not sure we even do but if we do, we could indeed infer it.

libs/live-countervalues/src/types.ts
gre-ledger: agreed that the input option is kinda self explanatory. i can remove

LL782 (approved):

PR #14025, opened 2026-01-27
chore(lwd): update react-transition-group usage add nodeRef for react…

iqbalibrahim-ledger:
… 19 upgrade prep

### ✅ Checklist

  - affects only the components that use `react-transition-group`
  - enables a smoother upgrade for React 19 on LWD

### 📝 Description

When upgrading LWD to React v19, the `react-transition-group` package causes unit tests to fail. `This is because react-transition-group@4.4.5` uses `ReactDOM.findDOMNode` internally, which React 19 removes. This blocks the React 19 upgrade for ledger-live-desktop - there is a temporary solution by mocking `findDOMNode` in Jest setup, however this is not a long term stable solution.

[Github issue on `react-transition-group` related to this](https://github.com/reactjs/react-transition-group/issues/893) - one of the comments suggests the same solution as the changes in this PR.

**Without nodeRef**: `react-transition-group` would try to use `findDOMNode` on the child. In React 19 that API doesn’t exist, causing runtime errors, type errors, and/or broken transitions.
**With nodeRef**: You tell the library exactly which DOM node to use. It never calls `findDOMNode`, so it’s compatible with React 19 and follows current React patterns (avoid refs to components, ref the DOM node explicitly).

Simply updated their usage pattern like so
 
**Before:**
```
<Transition in={isOpen}>
  <div>Content</div>
</Transition>
```
**After:**
```
const nodeRef = useRef(null);
<Transition in={isOpen} nodeRef={nodeRef}>
  <div ref={nodeRef}>Content</div>
</Transition>
```

### ❓ Context

- **JIRA or GitHub link**: 

[LIVE-23594](https://ledgerhq.atlassian.net/browse/LIVE-25394)

---

### DIFF
--- a/apps/ledger-live-desktop/tests/jestSetup.js
+++ b/apps/ledger-live-desktop/tests/jestSetup.js
@@ -140,7 +140,7 @@ const originalWarn = console.warn;
 // eslint-disable-next-line no-console
 const originalLog = console.log;
 
-const EXCLUDED_ERRORS = ["act(...)"];
+const EXCLUDED_ERRORS = ["act(...)", "ReactDOMTestUtils.act", "Warning: findDOMNode is deprecated"];
 
 const EXCLUDED_WARNINGS = ["@polkadot"];
 


iqbalibrahim-ledger:
formatting is happening when i run `pnpm desktop lint:fix` - I did find that there is incorrect usage of the `!` override for Tailwind, and the linting fixed it (so now it's `px-16!` rather than `!px-16 `which is incorrect for Tailwind v4. I've reverted these changes and will open a separate PR for them.

iqbalibrahim-ledger:
I have tested LWD portfolio, accounts, buy/sell screens which show the drawer and I'm happy there are no functional changes or issues. Awaiting support from the Wallet XP team for someone to assist in testing the My Ledger/Device Management screens as I do not have a Ledger device to test with.

@ysitbon FYI

apps/ledger-live-desktop/tests/jestSetup.js
```diff
@@ -140,7 +140,7 @@ const originalWarn = console.warn;
 // eslint-disable-next-line no-console
 const originalLog = console.log;
 
-const EXCLUDED_ERRORS = ["act(...)", "ReactDOMTestUtils.act", "Warning: findDOMNode is deprecated"];
```
mcayuelas-ledger, line 143: why remove this ` "ReactDOMTestUtils.act", `?

iqbalibrahim-ledger: accidental, I have added this back in. apologies!

iqbalibrahim-ledger: `findDOMNode` error exclusion is removed purposely, because of the `nodeRef` updates it no longer uses `findDOMNode` internally in `react-transition-group`

iqbalibrahim-ledger: maybe I will revert this and remove it in the React 19 upgrade PR, to be safe.

PR #14086, opened 2026-01-29
PERF: optimize accounts persistence to only save on significant changes

gre-ledger:
### ✅ Checklist

  - mobile and desktop

### 📝 Description

Don’t persist `lastSyncDate` and only re-save when account data actually changes in the user's eye.

- `Account#lastSyncDate` is no longer stored in app.json / MMKV. A predicate decides when an account is “worth re-saving” (balance, ops count, subAccounts, etc.; not `lastSyncDate` or `blockHeight`).
- `AccountRaw.lastSyncDate` optional; `toAccountRaw` omits it; `fromAccountRaw` still reads it (default `new Date(0)`). New helpers in live-common: `accountPersistedStateChanged`, `accountsPersistedStateChanged` (compare by index).
- **Mobile:** Only accounts that pass the predicate are written (unchanged: still per-account writes).
- **Desktop:** Accounts are written only when `accountsPersistedStateChanged(old, new)` is true.
- **Compatibility:** Old payloads with `lastSyncDate` still load; new ones don’t include it. No migration.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-25545

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/middlewares/db.ts
+++ b/apps/ledger-live-desktop/src/renderer/middlewares/db.ts
@@ -22,6 +22,7 @@ import { extractPersistedCALFromState } from "@ledgerhq/cryptoassets/cal-client/
 
 import { marketStoreSelector } from "../reducers/market";
 import { exportIdentitiesForPersistence } from "@ledgerhq/client-ids/store";
+import { accountsPersistedStateChanged } from "@ledgerhq/live-common/account/index";
 
 let DB_MIDDLEWARE_ENABLED = true;
 
@@ -60,14 +61,16 @@ const DBMiddleware: Middleware<object, State> = store => next => action => {
   }
 
   if (DB_MIDDLEWARE_ENABLED && action.type.startsWith("DB:")) {
+    const oldState = store.getState();
     const [, type] = action.type.split(":");
     store.dispatch({
       type,
       payload: action.payload,
     });
-    const state = store.getState();
-    setKey("app", "accounts", accountsExportSelector(state));
-    // ^ TODO ultimately we'll do same for accounts to drop DB: pattern
+    const newState = store.getState();
+    if (accountsPersistedStateChanged(oldState.accounts, newState.accounts)) {
+      setKey("app", "accounts", accountsExportSelector(newState));
+    }
   } else if (DB_MIDDLEWARE_ENABLED && action.type.startsWith(postOnboardingActionTypePrefix)) {
     next(action);
     const state = store.getState();
--- a/libs/ledger-live-common/src/account/persistence.ts
+++ b/libs/ledger-live-common/src/account/persistence.ts
@@ -1,0 +1,66 @@
+import type { Account, TokenAccount } from "@ledgerhq/types-live";
+
+/**
+ * Returns true when something that is persisted and user-visible (or structurally
+ * relevant) changed between prev and next. Used to decide if an account is worth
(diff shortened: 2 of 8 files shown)


libs/ledger-live-common/src/account/persistence.ts
```diff
@@ -0,0 +1,52 @@
...
+export function accountPersistedStateChanged(prev: Account, next: Account): boolean {
+  if (prev.id !== next.id) return true;
+  if (!prev.balance.eq(next.balance)) return true;
+  if (!prev.spendableBalance.eq(next.spendableBalance)) return true;
+  if (prev.operationsCount !== next.operationsCount) return true;
+  if (prev.operations.length !== next.operations.length) return true;
+  if (prev.pendingOperations.length !== next.pendingOperations.length) return true;
+  if ((prev.subAccounts?.length ?? 0) !== (next.subAccounts?.length ?? 0)) return true;
+  if (prev.syncHash !== next.syncHash) return true;
+  if ((prev.swapHistory?.length ?? 0) !== (next.swapHistory?.length ?? 0)) return true;
```
qperrot, line 28: Do you think we can improve it for more readability ?
Something like 
```if (prev.id !== next.id) return true;

if (
  !prev.balance.eq(next.balance) ||
  !prev.spendableBalance.eq(next.spendableBalance)
) {
  return true;
}

if (
  prev.operationsCount !== next.operationsCount ||
  prev.operations.length !== next.operations.length ||
  prev.pendingOperations.length !== next.pendingOperations.length
) {
  return true;
}

if (
  (prev.subAccounts?.length ?? 0) !== (next.subAccounts?.length ?? 0) ||
  (prev.swapHistory?.length ?? 0) !== (next.swapHistory?.length ?? 0)
) {
  return true;
}

if (prev.syncHash !== next.syncHash) return true;
```

gre-ledger: @qperrot Ok, I just reorganized a bit and added a js comment

libs/ledger-live-common/src/account/persistence.ts
```diff
@@ -0,0 +1,66 @@
...
+  const nextSubs = next.subAccounts ?? [];
+  for (let i = 0; i < prevSubs.length; i++) {
+    const p = prevSubs[i];
+    const n = nextSubs[i];
+    if (!n || n.type !== "TokenAccount") return true;
+    if (tokenAccountPersistedStateChanged(p, n)) return true;
+  }
+
+  return false;
+}
```
LL782, line 42: Should sub accounts be covered in the tests?

LL782: Sorry, I see it now

apps/ledger-live-desktop/src/renderer/middlewares/db.ts
LL782, line 1: **curiousity: is there no coverage of this change in unit tests**

I loved the tests around persistence and I'm not requesting a change or, just curious... I was expecting to see some coverage around these db.ts changes, is it missing or unnecessary?

gre-ledger: indeed you touch something very true, the persistence layers of desktop and mobile aren't unit tested at all.
hopefully we're going to rework them in a near future & embrace more the new architecture, on top of these they are not enough uniformed between both platforms, we have different paradigm and ways to debounce, detect changes,..

PR #14256, opened 2026-02-05
✨ (signer-evm) [LIVE-25119]: Disable auto trusted name lookup when no domain provided

aussedatlo:
### ✅ Checklist

  - Prevents unnecessary trusted name lookups when no domain is provided in the transaction resolution config
  - Affects `@ledgerhq/live-signer-evm` package

### 📝 Description

This PR disables automatic trusted name lookups in the DMK Ethereum signer when no domain information is provided in the resolution configuration.

**Problem:**
The context module was performing trusted name lookups even when no domain was available, resulting in unnecessary API calls and potential errors.

**Solution:**
- Created a `CustomHttpTrustedNamesDataSource` wrapper class that can be dynamically enabled/disabled
- Modified `DmkSignerEth.signTransaction()` to:
  - Enable trusted names only when `resolution?.domains?.[0]?.domain` is provided
  - Disable trusted names otherwise

**Usage:**
```typescript
// Trusted names enabled when domain is provided
signer.signTransaction(path, rawTx, { domains: [{ domain: "example.eth" }] });

// Trusted names disabled when no domain
signer.signTransaction(path, rawTx, {});
```

### ❓ Context

- **JIRA or GitHub link**: [LIVE-25119](https://ledgerhq.atlassian.net/browse/LIVE-25119)

---

### DIFF
--- a/libs/live-signer-evm/src/TogglableTrustedNameDataSource.ts
+++ b/libs/live-signer-evm/src/TogglableTrustedNameDataSource.ts
@@ -1,0 +1,36 @@
+import {
+  DEFAULT_CONFIG,
+  GetDomainNameInfosParams,
+  GetTrustedNameInfosParams,
+  HttpTrustedNameDataSource,
+  TrustedNameDataSource,
+  TrustedNamePayload,
+} from "@ledgerhq/context-module";
+import { Either, Left } from "purify-ts";
+
+export class TogglableTrustedNameDataSource implements TrustedNameDataSource {
+  private _defaultDataSource = new HttpTrustedNameDataSource(DEFAULT_CONFIG);
+  private _enableTrustedNames = false;
+
+  public enableTrustedNames() {
+    this._enableTrustedNames = true;
+  }
+
+  public disableTrustedNames() {
+    this._enableTrustedNames = false;
+  }
+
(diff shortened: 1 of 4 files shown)


libs/live-signer-evm/src/TogglableTrustedNameDataSource.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  public disableTrustedNames() {
+    this._enableTrustedNames = false;
+  }
+
+  getDomainNamePayload(
+    params: GetDomainNameInfosParams,
+  ): Promise<Either<Error, TrustedNamePayload>> {
+    return this._defaultDataSource.getDomainNamePayload(params);
```
paoun-ledger, line 30: I think the previous behavior was rather:
* `getDomainNamePayload` should resolve a trusted name only if the param `domain` comes explicitly from LW (`resolution?.domains?.[0]?.domain`)
* `getTrustedNamePayload` is always resolved (that's not consistent but we could not do better before your change. Now we may want to not resolve the trusted name if the source is `ens`. Other sources such as CAL should be resolved)

aussedatlo:
Not needed anymore

PR #14314, opened 2026-02-09
fix(mobile): prevent double-firing of close animation callbacks [LIVE-25842]

Justkant:
### ✅ Checklist

- [ ] **Covered by automatic tests.** Tested manually
  - All drawers close animation callbacks
  - Recover device select flow

### 📝 Description

Use refs instead of local variables in `QueuedDrawer`'s `closeAnim` to deduplicate callbacks across rapid successive calls.
Add a guard ref in `DeviceConnect` to prevent `handleSuccess` from executing twice.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-25842]

---

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawer/index.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawer/index.tsx
@@ -84,7 +84,6 @@ const QueuedDrawerNative = ({
   const translateY = useSharedValue(1000);
   const backdropOpacity = useSharedValue(0);
   const closeAnimTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
-  const closeAnimFiredRef = useRef(false);
 
   const openAnim = useCallback(() => {
     translateY.value = withTiming(0, {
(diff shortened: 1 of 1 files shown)


apps/ledger-live-mobile/src/mvvm/components/QueuedDrawer/index.tsx
```diff
@@ -94,25 +95,34 @@ const QueuedDrawerNative = ({
...
-              callbackFired = true;
-              after();
-            }
-          }, ANIMATION_DURATION + 50)
-        : null;
+      // Cancel any pending callback from a previous closeAnim call
+      if (closeAnimTimeoutRef.current) {
+        clearTimeout(closeAnimTimeoutRef.current);
+        closeAnimTimeoutRef.current = null;
+      }
```
Justkant, line 102: This one could stop a call that was passing before, not sure of the real impact but it's the biggest change here I think

mcayuelas-ledger: Maybe we need to test when a drawer follows one another?

ysitbon (approved):

apps/ledger-live-mobile/src/screens/DeviceConnect/index.tsx
```diff
@@ -50,6 +51,8 @@ export default function DeviceConnect({ navigation, route }: NavigationProps) {
 
   const handleSuccess = useCallback(
     (result: AppResult) => {
+      if (hasHandledSuccessRef.current) return;
+      hasHandledSuccessRef.current = true;
```
Justkant, line 55: I added that later on and tested every time without it too in order to make sure the fix in the QueuedDrawer was correct

This one is just another guard to avoid seeing the issue pop up again later for another reason

The only drawback is that we don't see incorrect behaviour surface as easily

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawer/index.tsx
ysitbon: As far as I remember it's tested in the integration tests

ysitbon: Here [openSuccessiveDrawers](https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawer/__integrations__/openSuccessiveDrawers.test.tsx)

PR #14354, opened 2026-02-10
feat: zcash account (LIVE-25304)

semeano:
### ✅ Checklist

  - @ledgerhq/zcash-shielded
  - @ledgerhq/coin-bitcoin
  - ledger-live-desktop
  - @ledgerhq/live-common

### 📝 Description

Update ZCash account with sync status data.

The sync state is saved in the account data. I changed the states of the sync a little bit by removing the "paused" state. This is because if you pause the sync and come back later, the progress is no longer valid because meanwhile there are more blocks to process (a new block is generated every ~75 seconds).
So I replaced "paused" with "stopped" and this way when we stop, it will show how many blocks were processed (since the last sync) and how many are missing to reach a complete sync. And every time the sync starts it will start from 0%.

https://github.com/user-attachments/assets/41879964-799e-4f1a-8d50-76e34e986e1c

Once the sync is finished it will show the timestamp of the last sync and after 2 minutes the "Start sync" button will show up because of the appearance of new blocks in the blockchain.

image

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-25304

---


libs/ledger-live-common/src/account/persistence.ts
```diff
@@ -27,6 +27,32 @@ export function accountPersistedStateChanged(prev: Account, next: Account): bool
   // swap history
   if ((prev.swapHistory?.length ?? 0) !== (next.swapHistory?.length ?? 0)) return true;
 
+  // zcash private info (or any account with privateInfo)
```
gre-ledger, line 30: short term feedback:

I would prefer that this function does not contain any coin-specific code. It's aim to be a UI heuristic that will makes the wallet not very often resaving accounts (this approach was recently added to limit the number of re-updates to app.json)

So I’m trying to better understand the underlying need here: why is it relevant to resave account when there is actually no UI impacts when it comes to balance or operations?


---

long term feedback:

I’m not very familiar with `privateInfo` in the account object, is that a new thing? because I’m not convinced we should be adding more fields into `Account`. It feels like a pattern we’re trying to move away from. In the future, we want to reduce the number of fields in Account and split responsibilities into smaller slices. While working on reducing the size of app.json, we noticed similar patterns — for example, large coin-specific caches such as the one in `bitcoinResources` caused significant performance issues.
If a specific coin requires its own storage, I think we should consider providing a dedicated database interface rather than adding large objects directly into the `Account` object.

now, i put this in long term feedback, because without a rework of model, there is nothing that can easily be addressed for now, but this year we plan to split the model into slices and it could be relevant to have this in mind.

semeano: This was done to track the progress of the private information (balance and transactions) of ZCash.

Basically, we need to sweep the blockchain looking for private transactions that belong to this account, and that means downloading blocks from the blockchain, decrypt them using the Unified Full Viewing Key from this account, and save it in app.json. Given that this is a process that can take a while, the progress is saved in app.json to persist the information, so we don't have to repeat the process (download and decrypt the blocks) every time we open the app.

I'm ok with finding a different option to store this information, **as long as** this DB is local. I'm saying this because the initial idea was to run the process (download + decrypt) in the backend but it was declined by ZCash Foundation. Everything needs to happen locally in the app, including the storage of the key, balance and txs.

libs/ledger-live-common/src/account/persistence.ts
```diff
@@ -27,6 +27,32 @@ export function accountPersistedStateChanged(prev: Account, next: Account): bool
...
+  const prevPrivateInfo =
+    "privateInfo" in prev ? (prev as { privateInfo?: any }).privateInfo : null;
+  const nextPrivateInfo =
+    "privateInfo" in next ? (next as { privateInfo?: any }).privateInfo : null;
+  if (prevPrivateInfo || nextPrivateInfo) {
+    if (!prevPrivateInfo || !nextPrivateInfo) return true;
+    if (prevPrivateInfo.ufvk !== nextPrivateInfo.ufvk) return true;
+    if (!prevPrivateInfo.balance?.eq?.(nextPrivateInfo.balance)) return true;
+    if (prevPrivateInfo.lastSyncTimestamp !== nextPrivateInfo.lastSyncTimestamp) return true;
```
gre-ledger, line 39: quick question on this: doesn't `lastSyncTimestamp` always change? => if so, i think we can consider the code to always returns true => which possibly mean it's equivalent to always return true in case of new zcash implementation 🤔  a tradeoff we could accept rather than having to maintain too much logic here

semeano: `lastSyncTimestamp` changes depending on the last time the user made a full sync. It could have been 20 minutes ago or 2 months ago, but we need to keep it in order to know if the sync is up to date or not, and to show when the last sync was done.

libs/coin-modules/coin-bitcoin/src/types.ts
```diff
@@ -174,16 +174,23 @@ export type TransactionStatusRaw = TransactionStatusCommonRaw & {
...
-  key: string | null;
-  balance: BigNumber;
-  syncState: "disabled" | "ready" | "running" | "paused" | "complete" | "outdated";
-  progress: number;
+export type ZcashPrivateInfo = {
+  saplingBalance: BigNumber;
+  orchardBalance: BigNumber;
+  ufvk: string | null;
+  syncState: ZcashSyncState;
   lastSyncTimestamp: number | null;
```
may01, line 188: It will be good to have here the `startSyncBlock` that user inputs in the modal. For cases when we will clean cache so that we can sync without prompting start date each time.

semeano: As far as I know we only have a date input not a block height input. We will show the starting block for the selected date once we have that implemented.

image

qperrot (approved):

libs/ledger-live-common/src/account/persistence.ts
gre-ledger: I see. 
and the rest of the privateInfo fields can change outside of a full sync ? how ? because in the Wallet i think the only time we update accounts are only via a sync . 
which means possibly that `lastSyncTimestamp` changes may be enough to check and return true as a predicate for a privateInfo changes?🤔 

and btw, at the Account level, we already have `.lastSyncDate` , so i'm actually wondering if we could have a more generic way to check it for a specific list of account currency ids.

PR #14368, opened 2026-02-11
🐛 fix(desktop,mobile): default time range to 1D when Wallet 40 FF is on

LucasWerey:
### ✅ Checklist

  - default timerange

### 📝 Description

When Wallet 4.0 is enabled, the portfolio time range is reset to **1D on every app relaunch**. During a session the user can still switch to other ranges (1W, 1M, etc.); the reset only happens on cold start.

**Why desktop and mobile implement this differently:**

- **Mobile — logic in the settings reducer (`SETTINGS_IMPORT`):** Settings are loaded by dispatching `importSettings(settingsData)`. By then LiveConfig/remote config is already available, so we can call `getFeature({ key: "lwmWallet40", localOverrides })` inside the reducer and force `selectedTimeRange: "day"` when the flag is on.

- **Desktop — logic in a `useEffect` in `FirebaseFeatureFlagsProvider`:** `FETCH_SETTINGS` runs during `init()` before the React tree mounts. Firebase remote config is loaded later when `FirebaseRemoteConfigProvider` mounts, so when the reducer runs `LiveConfig.instance?.provider` is not set and `getFeature()` returns `null`. The fix is to run the check after config is ready: a one-time `useEffect` in `FirebaseFeatureFlagsProvider` (which only mounts once Firebase has loaded), where we call `getFeature` and dispatch `setSelectedTimeRange("day")` when the lwdWallet40 flag is enabled.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-25889](https://ledgerhq.atlassian.net/browse/LIVE-25889)

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
@@ -6,8 +6,13 @@ import type { FirebaseFeatureFlagsProviderProps as Props } from "@ledgerhq/live-
 import { Feature, FeatureId } from "@ledgerhq/types-live";
 import { useFirebaseRemoteConfig } from "./FirebaseRemoteConfig";
 import { overriddenFeatureFlagsSelector } from "../reducers/settings";
-import { setOverriddenFeatureFlag, setOverriddenFeatureFlags } from "../actions/settings";
+import {
+  setOverriddenFeatureFlag,
+  setOverriddenFeatureFlags,
+  setSelectedTimeRange,
+} from "../actions/settings";
 import { setAnalyticsFeatureFlagMethod } from "../analytics/segment";
+import { useWalletFeaturesConfig } from "@ledgerhq/live-common/featureFlags/walletFeaturesConfig/useWalletFeaturesConfig";
 
 export const FirebaseFeatureFlagsProvider = ({
   children,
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -55,6 +59,19 @@ export const FirebaseFeatureFlagsProvider = ({
...
   }, [remoteConfig, wrappedGetFeature]);
 
+  // That's temporary until wallet 4.0 is 100% enabled
+  // We need to set the selected time range to day at each app launch for the wallet 4.0 feature flag
+  useEffect(() => {
+    const wallet40FF = getFeature({ key: "lwdWallet40", localOverrides });
+    const isWallet40Enabled = wallet40FF?.enabled === true;
+    const isWallet40GraphReworkEnabled =
+      wallet40FF?.params?.graphRework === true && isWallet40Enabled;
+    if (isWallet40GraphReworkEnabled) {
```
mcayuelas-ledger: can be simplified with `useWalletFeaturesConfig`

LucasWerey: indeed my bad I'll change

LucasWerey: done

apps/ledger-live-desktop/src/renderer/components/FirebaseFeatureFlags.tsx
```diff
@@ -6,17 +6,23 @@ import type { FirebaseFeatureFlagsProviderProps as Props } from "@ledgerhq/live-
...
 export const FirebaseFeatureFlagsProvider = ({
   children,
   getFeature,
 }: Props): React.JSX.Element => {
   const { config: remoteConfig } = useFirebaseRemoteConfig();
 
-  const localOverrides = useSelector(overriddenFeatureFlagsSelector);
   const dispatch = useDispatch();
+  const localOverrides = useSelector(overriddenFeatureFlagsSelector);
+  const wallet40FF = useWalletFeaturesConfig("desktop");
```
mcayuelas-ledger:
```suggestion
  const { shouldDisplayGraphRework: isWallet40GraphReworkEnabled } = useWalletFeaturesConfig("desktop");
```

PR #14370, opened 2026-02-11
Feat/live 25795 buy device botton sheet

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Add bottom sheet implementation for global drawer for the reborn flow (lazy onboarding).

https://github.com/user-attachments/assets/a5913595-7562-4a93-99f7-0f4623a3d244

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-25795

---

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/features/Reborn/components/BuyDeviceView/useBuyDeviceViewModel.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/features/Reborn/components/BuyDeviceView/useBuyDeviceViewModel.tsx
@@ -1,0 +1,92 @@
+import { useFeature } from "@ledgerhq/live-common/featureFlags/index";
+import { useNavigation } from "@react-navigation/native";
+import { useCallback } from "react";
+import { useTranslation } from "~/context/Locale";
+import { Linking } from "react-native";
+import { useSelector, useDispatch } from "~/context/hooks";
+import { useTheme } from "styled-components/native";
+import { setOnboardingHasDevice } from "~/actions/settings";
+import { track } from "~/analytics";
+import { BuyDeviceNavigatorParamList } from "~/components/RootNavigator/types/BuyDeviceNavigator";
+import {
+  BaseNavigationComposite,
+  StackNavigatorNavigation,
+} from "~/components/RootNavigator/types/helpers";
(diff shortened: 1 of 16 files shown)


apps/ledger-live-mobile/src/mvvm/features/Reborn/components/BuyDeviceView/useBuyDeviceViewModel.tsx
```diff
@@ -0,0 +1,86 @@
...
+>;
+
+const useBuyDeviceViewModel = () => {
+  const { t } = useTranslation();
+  const navigation = useNavigation<NavigationProp>();
+  const { colors } = useTheme();
+  const buyDeviceFromLive = useFeature("buyDeviceFromLive");
+  const readOnlyModeEnabled = useSelector(readOnlyModeEnabledSelector);
+  const dispatch = useDispatch();
+  const currentNavigation = navigation.getParent()?.getParent()?.getState().routes[0].name;
```
gre-ledger: having to do this feels weird, conceptually a feature component shouldn't be self aware of where it's mounted. 
if we rework the navigation tree independently of the feature, it can easily break.
that said i don't exactly know the best pattern to solve this: 
if keeping the idea to figure out from navigation context: a first idea would be to more generically iterate on climbing up on parent to see if one of the parent is BaseOnboarding without assuming we're currently mounted on level -2 🤔

gre-ledger: or maybe another way:

```
navigation.navigate(NavigatorName.BuyDevice, {
  screen: ScreenName.PurchaseDevice,
  params: { source: "onboarding" }
});

// Then in the component, extract from route params instead
const route = useRoute();
const isInOnboarding = route.params?.source === "onboarding";
```

mitchellv-ledger: Completely agree with this comment, this logic was already existing so hadn't looked into updating it as was out of scope. I moved the logic to a recursive function out and made it agnostic of location, was more complicated to figure out all the locations where the screen was being navigated from.

PR #14375, opened 2026-02-11
💄style(desktop): [LIVE-25279] Less intrusive OS update banner

benruseau:
### ✅ Checklist

  - **Firmware / My Ledger:** OS update entry point is now a compact transparent button instead of a large banner card. QA should check visibility and tap behavior on Portfolio (TopBar) and on Manager/Apps screen.
  - **Analytics:** If `banner_impression` for "OS update" on portfolio page is added, confirm it fires when the banner is shown.

### 📝 Description

**Problem:** The OS update banner was a large card (background, padding, two text lines including device name and version). It took a lot of space and felt intrusive on Portfolio and Manager.

**Solution:** Replace it with a **less intrusive** entry point:

- **UI:** A single transparent `LumenButton` (small, with device icon) showing only the title "OS update available". No background card, no description line, no version in the banner.
- **Layout:** On Portfolio the button sits in the TopBar; on Manager/Apps screen the same component is used where the firmware banner is rendered.
- **Data:** The `version` prop is removed from the banner; the unsupported-update drawer is unchanged and still explains the context when opened.
- **Behavior:** Clicking the button dispatches `osUpdateRequested(true)` and navigates to `/manager`. The Manager screen reads the flag and opens the firmware update drawer. End-to-end behavior (disclaimer flow, unsupported drawer, etc.) is unchanged; only the banner’s size and content are reduced.

**Why we did it this way:** The firmware update flow (drawer, steps, device checks, installed apps, reinstall logic) is fully implemented and owned by the Manager screen (`FirmwareUpdate`, `DeviceDashboard`). We had two options: (1) Move the whole flow out of Manager so the drawer could be opened from the TopBar without leaving the page—which would mean either duplicating the flow or lifting a lot of Manager-specific state and components to a global layer, a large refactor with more regression risk; or (2) Keep the flow in Manager and only trigger it from the TopBar by dispatching an event and redirecting. As a quick win, we chose (2): we **still redirect to the Manager page** and **do not move anything into `FirmwareUpdateBanner.tsx`**. The banner only dispatches `osUpdateRequested(true)` and navigates to `/manager`; when the user lands on Manager, the existing `useEffect` there sees the flag and opens the drawer. So the compromise is: **redirect + one dispatched event** to open the drawer on the Manager page, instead of moving or duplicating the whole update flow. This keeps the change small and the flow in one place.

**Before:**

https://github.com/user-attachments/assets/86374c50-9d0e-4b4b-839f-9dcf089b1693

**After:**

https://github.com/user-attachments/assets/be310546-99cf-4634-a9c4-2a21bad57154

### ❓ Context

- **JIRA or GitHub link**: [LIVE-25279](https://ledgerhq.atlassian.net/browse/LIVE-25279)

---

[LIVE-25279]: https://ledgerhq.atlassian.net/browse/LIVE-25279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/mvvm/components/TopBar/__tests__/FirmwareUpdateBannerEntry.test.tsx
```diff
@@ -0,0 +1,123 @@
+import React from "react";
+import { render, screen, userEvent } from "tests/testSetup";
+import FirmwareUpdateBannerEntry from "~/renderer/components/FirmwareUpdateBanner";
+import type { FirmwareUpdateContext } from "@ledgerhq/types-live";
+
+jest.mock("@ledgerhq/live-common/featureFlags/index", () => {
```
mcayuelas-ledger: you don't need to mock this one :) 

you can override the store like in `apps/ledger-live-desktop/src/mvvm/features/NftEntryPoint/__tests__/useNftsEntryPointViewModel.test.tsx` in getDefaultInitialState

apps/ledger-live-desktop/src/mvvm/components/TopBar/__tests__/FirmwareUpdateBannerEntry.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const React = require("react");
+  return {
+    useWalletFeaturesConfig: jest.fn(),
+    FeatureFlagsProvider: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+    isFeature: jest.fn(),
+  };
+});
+const useWalletFeaturesConfig = jest.requireMock(
+  "@ledgerhq/live-common/featureFlags/index",
+).useWalletFeaturesConfig;
```
mcayuelas-ledger: same for this

apps/ledger-live-desktop/src/mvvm/components/TopBar/__tests__/FirmwareUpdateBannerEntry.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+      initialState: getDefaultInitialState({
+        settings: { latestFirmware: mockFirmware },
+      }),
+    });
+
+    expect(screen.getByTestId("topbar-os-update-button")).toBeVisible();
+  });
+
+  it("dispatches osUpdateRequested when OS update button is clicked", async () => {
+    const user = userEvent.setup();
```
mcayuelas-ledger: you can get user from {} = render also :)

apps/ledger-live-desktop/src/mvvm/components/TopBar/TopBarView.tsx
```diff
@@ -41,7 +47,15 @@ const TopBarView = ({ actionsList }: TopBarViewProps) => {
...
       <NavBarTrailing className="h-48 gap-12">
+        {hasAccounts && (
+          <>
+            <ActivityIndicator />
+            <TopBarDivider />
+          </>
+        )}
         <LiveAppDrawer />
+
+        {isWallet40Enabled ? <FirmwareUpdateBannerEntry /> : null}
```
mcayuelas-ledger: this view is rendered only if we are in wallet4.0 mode, so you can remove the check

apps/ledger-live-desktop/src/renderer/__tests__/Default.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  default: () => <div data-testid="firmware-update-banner-entry">FirmwareUpdateBannerEntry</div>,
+}));
+
+jest.mock(
+  "@ledgerhq/live-common/featureFlags/walletFeaturesConfig/useWalletFeaturesConfig",
+  () => ({
+    useWalletFeaturesConfig: jest.fn(),
+  }),
+);
+const useWalletFeaturesConfig = jest.requireMock(
```
mcayuelas-ledger: no need to mock, just by overriding state you can avoid it

apps/ledger-live-desktop/src/renderer/reducers/manager.ts
mcayuelas-ledger, line 1: Could you Slice form RTKQuery please? https://redux-toolkit.js.org/api/createSlice
you have example like modularDrawer

apps/ledger-live-desktop/src/renderer/screens/manager/__tests__/Dashboard.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  default: () => <div data-testid="firmware-update">FirmwareUpdate</div>,
+}));
+jest.mock("~/renderer/analytics/TrackPage", () => ({
+  __esModule: true,
+  default: () => null,
+}));
+jest.mock("@ledgerhq/live-common/device/use-cases/getLatestFirmwareForDeviceUseCase", () => ({
+  getLatestFirmwareForDeviceUseCase: jest.fn(() => Promise.resolve(null)),
+}));
+jest.mock("@ledgerhq/live-common/featureFlags/index", () => {
```
mcayuelas-ledger: same as above

apps/ledger-live-desktop/src/renderer/screens/manager/FirmwareUpdate/__tests__/index.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    firmwareUnsupported: jest.fn(() => false),
+    firmwareUpdateWillUninstallApps: jest.fn(() => false),
+    firmwareUpdateRequiresUserToUninstallApps: jest.fn(() => false),
+  },
+}));
+jest.mock("@ledgerhq/devices", () => ({
+  ...jest.requireActual("@ledgerhq/devices"),
+  getDeviceModel: jest.fn(() => ({ id: "nanoX" })),
+}));
+jest.mock("~/renderer/hooks/useLocalizedUrls", () => ({
```
mcayuelas-ledger: why do you need to mock this one?

apps/ledger-live-desktop/src/renderer/screens/manager/FirmwareUpdate/__tests__/index.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+}));
+jest.mock("~/renderer/hooks/useKeepScreenAwake", () => ({
+  useKeepScreenAwake: () => {},
+}));
+jest.mock("~/renderer/analytics/segment", () => ({
+  track: jest.fn(),
+  setAnalyticsFeatureFlagMethod: jest.fn(),
+}));
+
+const useWalletFeaturesConfig = jest.requireMock(
```
mcayuelas-ledger: same here

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
```diff
@@ -1,45 +1,93 @@
...
+import TopBanner from "./TopBanner";
 import Box from "./Box";
-import StyleProvider from "../styles/StyleProvider";
+import getCleanVersion from "../screens/manager/FirmwareUpdate/getCleanVersion";
+import { useTheme } from "styled-components";
+import { getEnv } from "@ledgerhq/live-env";
+import { radii } from "~/renderer/styles/theme";
+import { Button, Text } from "@ledgerhq/react-ui";
+
+function getDeviceIcon(modelId: DeviceModelId | undefined) {
```
gre-ledger: not super future proof (when new device model comes), i wonder if there is a way to at least move it into its dedicated file because it almost feel like it's something for device team to configure & collocate the maintenance of adding devices 🤔 
wdyt @OlivierFreyssinet ?

benruseau: Totally agree. I noticed that there are several places in the repo that use their own switch on device model for icons (and similar logic), which is easy to get out of sync when a new device is added.

Maybe we should:

**1. Single source of truth for device types**
The list of device models (e.g. an enum or union type) should live in one place: either exposed from an existing package (I don’t have enough context to say which one), or even in a dedicated package. I’d avoid defining it in `ledger-live` itself, since this repo is mainly UI and the list of devices feels like a shared/domain concern.

**2. Utils with an exhaustive switch**
Using that type, we could have dedicated utils in desktop and mobile apps that maps device model → icon (or other UI) via an exhaustive `switch`. For example:

```typescript
   function getDeviceIcon(modelId: DeviceModelId) {
     switch (modelId) {
       case DeviceModelId.stax:
         return Stax;
       case DeviceModelId.europa:
         return Flex;
       case DeviceModelId.apex:
         return Apex;
       case DeviceModelId.nanoS:
       case DeviceModelId.nanoSP:
       //...
         return Nano;
       default: 
         exhaustive(modelId)
     }
   }

  function exhaustive(modelId: never): never {
    throw new Error('Should never get there.');
  }
```
All current call sites (`FirmwareUpdateBanner`, `DeviceItem`, `BleDeviceItem`, etc.) would use this helper instead of their own switches.

**This way, we make sure that:**
When a new device model is added in the source-of-truth package and we bump the dependency, **typecheck will fail** until we add a corresponding case (or explicitly handle it) in this util. So we get a single place to update and a build-time guarantee that no model is forgotten.

mcayuelas-ledger: Indeed a single source of truth at least for W4.0 stuff, because we already have it in TopBar for the new My ledger entry

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
```diff
@@ -1,74 +1,143 @@
...
       }}
       content={{
         message: (
           <Box>
-            <Text fontFamily="Inter|Bold" fontSize={5} color="neutral.c100">
-              {t(old ? "manager.firmware.banner.old.warning" : "manager.firmware.banner.warning")}
+            <Text
+              fontFamily="Inter|Bold"
+              fontSize={5}
+              color={isWallet40Enabled ? "--text-on-accent" : "neutral.c100"}
```
mcayuelas-ledger: I'm not sure you should use design tokens like that, maybe better to just extract the component and create a new & old one no?

benruseau:
Small changes after UI uptades: 
- https://www.figma.com/design/QZv5fm4oJ1GUS1iIUU8lJt/Portfolio-Production-%E2%80%A2--Wallet-4.0?node-id=12624-32868&p=f&m=dev
- https://www.figma.com/design/fGS8z9oNs5lxVIFU6rIH16/My-Wallet-Production-%E2%80%A2-Wallet-4.0?node-id=10711-32337&p=f&m=dev

**Before:**

https://github.com/user-attachments/assets/53c5620a-3c9f-46a0-8574-6c304d578965


**After:**

https://github.com/user-attachments/assets/87d82f02-6663-4041-8690-2db91736fefa

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
benruseau: Moved [getDeviceIcon.ts](https://github.com/LedgerHQ/ledger-live/blob/aaf0143a55705ca9457e1dcda573a576ca2c513d/apps/ledger-live-desktop/src/mvvm/utils/getDeviceIcon.ts) in `src/mvvm/utils` for now...

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
```diff
@@ -74,6 +117,91 @@ const FirmwareUpdateBanner = ({ old, right }: { old?: boolean; right?: React.Rea
...
+
+  const visibleFirmwareVersion =
+    process.env.DEBUG_FW_VERSION ??
+    (latestFirmware ? getCleanVersion(latestFirmware.final.name) : "");
+
+  // The 2.1.0-rc2 release caused issues with localization e2e tests because it
+  // displayed a banner for updating the FW, which would continue to show with
+  // future releases. To fix this, the banner is currently being manually removed.
+  // A more stable solution would be to mock all API calls.
+  const hideBannerForMocks = getEnv("MOCK") && visibleFirmwareVersion.startsWith("2.1");
```
gre-ledger, line 183: we should avoid having test related code that land in the components side 🙏 
there might be way to jest.mock something to be able to achieve the same.

benruseau: I don't know, this code has 2 yo. 

Maybe we should investigate in a dedicated PR?

apps/ledger-live-desktop/src/mvvm/components/TopBar/TopBarView.tsx
```diff
@@ -5,16 +5,19 @@ import { LiveAppDrawer } from "~/renderer/components/LiveAppDrawer";
...
   return (
     <NavBar className="items-center px-32 pt-32 pb-24">
       <NavBarTitle className="h-48">
         <Breadcrumb />
       </NavBarTitle>
       <NavBarTrailing className="h-48 gap-12">
         <LiveAppDrawer />
+
+        {shouldShowFirmwareUpdateBanner && <FirmwareUpdateBanner />}
         <Updater />
```
mcayuelas-ledger, line 20: we need to set priority between those 2, like in Default

apps/ledger-live-desktop/src/mvvm/components/TopBar/__tests__/TopBarView.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  __esModule: true,
+  default: () => null,
+}));
+jest.mock("../components/ActionsList", () => ({
+  TopBarActionsList: () => null,
+}));
+jest.mock("~/renderer/components/FirmwareUpdateBanner", () => ({
+  __esModule: true,
+  default: () => <div data-testid="firmware-update-banner" />,
+}));
```
mcayuelas-ledger: why do you need to mock everything? :)

apps/ledger-live-desktop/src/renderer/__tests__/Default.test.tsx
```diff
@@ -0,0 +1,99 @@
+import React from "react";
+import { render, screen, waitFor } from "tests/testSetup";
+import Default from "../Default";
+
```
mcayuelas-ledger, line 4: you are mocking too much I think

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
```diff
@@ -1,49 +1,92 @@
...
-const FirmwareUpdateBanner = ({ old, right }: { old?: boolean; right?: React.ReactNode }) => {
-  const navigate = useNavigate();
+type BannerContentProps = {
+  old?: boolean;
+  right?: React.ReactNode;
+  visibleFirmwareVersion: string;
+  onClick: () => void;
+};
+
+const FirmwareUpdateBannerNew = ({
```
mcayuelas-ledger, line 29: why in the new component we have styled-component stuffs? :)

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
```diff
@@ -74,6 +117,87 @@ const FirmwareUpdateBanner = ({ old, right }: { old?: boolean; right?: React.Rea
...
+  const location = useLocation();
+  const { t } = useTranslation();
+  const latestFirmware = useSelector(latestFirmwareSelector);
+  const currentDevice = useSelector(getCurrentDevice);
+  const { shouldDisplayWallet40MainNav } = useWalletFeaturesConfig("desktop");
+  const hasTrackedImpressionRef = useRef(false);
+  const inManager = location.pathname === "/manager";
+
+  useEffect(() => {
+    if (!shouldDisplayWallet40MainNav) return;
```
mcayuelas-ledger: you already check that in `Default` no?

apps/ledger-live-desktop/src/mvvm/components/TopBar/__tests__/TopBarView.test.tsx
benruseau: The tests don't pass if I don't mock those three ones

apps/ledger-live-desktop/src/renderer/__tests__/Default.test.tsx
benruseau: Updated mocks to the strict minimum for the tests to pass

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
benruseau: Well spotted!

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
benruseau: Will leave it like this for now (waiting for dedicated Card component ; cf `#lumen-support` discussion)

apps/ledger-live-desktop/src/mvvm/components/TopBar/TopBarView.tsx
benruseau: done!

apps/ledger-live-desktop/src/renderer/screens/manager/Dashboard.tsx
```diff
@@ -36,18 +37,22 @@ const Dashboard = ({
...
   const [preventResetOnDeviceChange, setPreventResetOnDeviceChange] = useState(false);
   const deviceChangedWhenResetPrevented = useRef(false);
   const [firmware, setFirmware] = useState<FirmwareUpdateContext | null>(null);
   const [firmwareError, setFirmwareError] = useState(null);
+  const osUpdateRequested = useSelector(osUpdateRequestedSelector);
   const params = new URLSearchParams(search || "");
-  const openFirmwareUpdate = params.get("firmwareUpdate") === "true";
+  const openFirmwareUpdate = shouldDisplayWallet40MainNav
+    ? osUpdateRequested
+    : params.get("firmwareUpdate") === "true";
```
OlivierFreyssinet: Why don't we also rely on navigation parameter for the new flow ? It would keep things simple, avoid having to introduce a new store reducer etc.

apps/ledger-live-desktop/src/renderer/components/FirmwareUpdateBanner.tsx
```diff
@@ -74,6 +117,87 @@ const FirmwareUpdateBanner = ({ old, right }: { old?: boolean; right?: React.Rea
...
+
+  const onClick = () => {
+    if (shouldDisplayWallet40MainNav) {
+      track("button_clicked", {
+        page: "portfolio",
+        banner: "OS update",
+        button: "click(update)",
+      });
+      dispatch(osUpdateRequested(true));
+      navigate("/manager");
```
OlivierFreyssinet: Here you could add `?firmwareUpdate` and you wouldn't need to dispatch `osUpdateRequested`

benruseau: Good point, done!

PR #14384, opened 2026-02-11
[LWMD] remove counter values market cap provider

LucasWerey:
### ✅ Checklist

    - "New" files in live-countervalues-react are renames of old files. Their behaviour is liable to change with redesign state management in a future ticket, so investment in new coverage at this point is not sensible.
    - Existing test setup has been updated.
    - Accessing counter value IDs sorted by market cap without going through React Context Provider
    - Removing React Context Provider

### 📝 Description

- Context provider removed
- API in `live-coutervalues-react` renamed and updated
- ...

### ❓ Context

**JIRA or GitHub link**:
- https://ledgerhq.atlassian.net/browse/LIVE-25658
- https://ledgerhq.atlassian.net/browse/LIVE-25659
- https://ledgerhq.atlassian.net/browse/LIVE-25660

---

### DIFF
--- a/libs/live-countervalues-react/package.json
+++ b/libs/live-countervalues-react/package.json
@@ -28,11 +28,9 @@
     "@ledgerhq/types-cryptoassets": "workspace:*",
     "@ledgerhq/cryptoassets": "workspace:*",
     "@ledgerhq/logs": "workspace:*",
-    "@ledgerhq/live-env": "workspace:*",
     "@ledgerhq/live-countervalues": "workspace:*",
     "@ledgerhq/coin-framework": "workspace:*",
     "@ledgerhq/live-hooks": "workspace:*",
-    "@reduxjs/toolkit": "catalog:",
     "bignumber.js": "9"
   },
   "devDependencies": {


libs/live-countervalues-react/src/index.tsx
```diff
@@ -25,7 +25,7 @@ import React, {
 } from "react";
 import { useMarketcapIds } from "./CountervaluesMarketcapProvider";
 
-export { CountervaluesMarketcapProvider, useMarketcapIds } from "./CountervaluesMarketcapProvider";
+export { useMarketcapIds } from "./CountervaluesMarketcapProvider";
```
LL782: This import export was here before our changes but do we want to keep it like this? I'm leaving a note to look into it longer because it looks weird now

```ts
import { useMarketcapIds } from "./CountervaluesMarketcapProvider";

export { useMarketcapIds } from "./CountervaluesMarketcapProvider";
```

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
LL782, line 1: **Note for changes:** let's rename and maybe relocate this file. Perhaps to `useMarketcapIds`

apps/ledger-live-mobile/src/api/countervaluesMarketcapApi.ts
```diff
@@ -0,0 +1,25 @@
...
+      keepUnusedDataFor: (30 * 60 * 1000) / 1000,
+    }),
+  }),
+});
+
+export const selectCountervaluesMarketcapIdsResult =
+  countervaluesMarketcapApi.endpoints.getMarketcapIds.select();
+export const selectCountervaluesMarketcapIds = (
+  state: Parameters<typeof selectCountervaluesMarketcapIdsResult>[0],
+): string[] => selectCountervaluesMarketcapIdsResult(state).data ?? [];
```
LL782: It looks to me like these can be removed

```suggestion
```

libs/live-countervalues-react/src/api.ts
LL782, line 1: We noticed a better name for this, `counterValuesByMarketCap`

Let's rename everywhere, we have the name `CounterValuesMarketcap` in a few places

apps/ledger-live-mobile/src/reducers/countervalues.ts
LL782, line 1: **To confirm my understanding**

In this file (counterValues) we remove things related to counterValuesByMarketCap, since those are handled elsewhere. I'm not really sure what I'm looking at, so I'm leaving this note to come back and get a better understanding later.

apps/ledger-live-mobile/src/context/rtkQueryApi.ts
```diff
@@ -6,6 +6,7 @@ import { cryptoAssetsApi } from "@ledgerhq/cryptoassets/cal-client/state-manager
 import { firebaseRemoteConfigApi } from "LLM/api/firebaseRemoteConfigApi";
 import { pushDevicesApi } from "@ledgerhq/client-ids/api";
 import { cmcApi } from "@ledgerhq/live-common/cmc-client/state-manager/api";
+import { countervaluesMarketcapApi } from "~/api/countervaluesMarketcapApi";
```
LL782: **Note:** The API has ended up inside the `ledger-live-mobile` workspace. We probably want to move that

apps/ledger-live-mobile/src/reducers/countervalues.ts
LL782, line 1: **Additional question to confirm understanding**

Where should this file live? It's in mobile and there is a lot of code duplicated between here and desktop. I would expect them to share a common file. Is that right?

Also, if they are meant to be separate it seems we've only updated mobile.

libs/live-countervalues-react/src/api.ts
```diff
@@ -0,0 +1,25 @@
...
+  reducerPath: "countervaluesMarketcapApi",
+  baseQuery: fetchBaseQuery({
+    baseUrl: getEnv("LEDGER_COUNTERVALUES_API"),
+  }),
+  endpoints: build => ({
+    getMarketcapIds: build.query<string[], void>({
+      query: () => ({
+        url: "/v3/supported/crypto",
+      }),
+      keepUnusedDataFor: (30 * 60 * 1000) / 1000,
```
LL782: **Idea:**

This value is in seconds, 
- we could express it more simply as `30 * 60` 
- or an explainer variable like `THIRTY_MINUTES`

apps/ledger-live-mobile/src/screens/RequestAccount/01-SelectCrypto.tsx
```diff
@@ -62,7 +62,10 @@ export default function RequestAccountsSelectCrypto({ navigation, route }: Props
 
   // Pagination is a bit strange with this because we order by market cap
   // but it works fine and keeps the old/expected order as much as possible
-  const ids = useMarketcapIds();
+  const { data: ids = [] } = useGetMarketcapIdsQuery(undefined, {
+    pollingInterval: 30 * 60 * 1000,
```
LL782: **Little idea:**

This is 30 minutes, let's use an explainer variable like `THIRTY_MINUTES`

_Note: the [`pollingInterval`](https://redux-toolkit.js.org/rtk-query/usage/polling) is in ms and [`keepUnusedDataFor`](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#reducing-subscription-time-with-keepunuseddatafor) is in seconds_

libs/live-countervalues-react/src/index.tsx
LL782: There's actually a lot more in this file that looks like it might only have been used by the bridge. We might be able to reduce this quite a lot

LL782: The context bridge that's still in place relates to CounterValuesContext (rather than CounterValuesMarketcapProvider) so let's leave this until https://ledgerhq.atlassian.net/browse/LIVE-26212

apps/ledger-live-mobile/src/api/countervaluesMarketcapApi.ts
LL782: Done in https://github.com/LedgerHQ/ledger-live/pull/14384/commits/3c98b2a98125448ede09509c635f8ed4debeca42

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
LL782: Done in d97b26b277f35bb359e6fb0d4af43d103af3f1a1

apps/ledger-live-mobile/src/context/rtkQueryApi.ts
LL782: We've moved it as part of f6b59a62a9e801511b0ee7ff96f1e81f48d6e2f9

libs/live-countervalues-react/src/api.ts
LL782: We performed a partial rename in f6b59a62a9e801511b0ee7ff96f1e81f48d6e2f9 but the more accurate name `counterValuesApi` is already in use.  This can be addressed in https://ledgerhq.atlassian.net/browse/LIVE-26226

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
LL782, line 1: There's a lot of polling logic, etc. that I don't think we need to handle on this level. 

We have RTK Query for that surely? WDYT @mcayuelas-ledger ?

mcayuelas-ledger: You can enhance it of course

apps/ledger-live-mobile/src/context/rtkQueryApi.ts
```diff
@@ -6,6 +6,7 @@ import { cryptoAssetsApi } from "@ledgerhq/cryptoassets/cal-client/state-manager
 import { firebaseRemoteConfigApi } from "LLM/api/firebaseRemoteConfigApi";
 import { pushDevicesApi } from "@ledgerhq/client-ids/api";
 import { cmcApi } from "@ledgerhq/live-common/cmc-client/state-manager/api";
+import { countervaluesByMarketcapApi } from "@ledgerhq/live-countervalues-react/api";
```
mcayuelas-ledger: should be added in `apps/ledger-live-desktop/src/renderer/reducers/rtkQueryApi.ts` also

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
LL782: My point is that I don't think we need `useSyncExternalStore`. It was introduced during the mob by the agent.

apps/ledger-live-mobile/src/context/rtkQueryApi.ts
LL782: Is it needed in desktop? 🤔

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
gre-ledger: indeed, it seems we expose a `useGetIdsByMarketcapQuery` from the other file that ain't used & could be used instead ?

LL782: So, in hindsight I think moving fully over to RTK Query as part of these changes would have been simpler. 

The current situation is a half-way-house using `useSyncExternalStore` and not using the new RTK [counterValuesByMarketCapApi](https://github.com/LedgerHQ/ledger-live/blob/37e8887de3ca6046f6efc4c601eac4f369911a22/libs/live-countervalues-react/src/api.ts) at all.

To keep things moving and merge the good work done so far, this is my plan:
- removing the unfinished RTK work from this changeset
- leave the `useSyncExternalStore` implementation as a temporary solution that's been manaully verified
- come back immediately after merge to remove `useSyncExternalStore` and fully implement RTK in a [future ticket](https://ledgerhq.atlassian.net/browse/LIVE-25578)

The obvious alternative is to push forward with the RTK implementation on this branch. 

@gre-ledger @mcayuelas-ledger @LucasWerey  what do you think?

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
```diff
@@ -0,0 +1,81 @@
...
+  };
+}
+
+function getMarketcapSnapshot() {
+  return marketcapStore.ids;
+}
+
+export function useIdsByMarketCap(): string[] {
+  return useSyncExternalStore(subscribeMarketcap, getMarketcapSnapshot, getMarketcapSnapshot);
+}
```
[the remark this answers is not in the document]
LL782: This is a good point but the use of `useSyncExternalStore` is very temporary. We fully intend to remove it and reply on RTK Query instead. I'd rather leave adding test coverage until then

libs/live-countervalues-react/package.json
```diff
@@ -28,9 +28,11 @@
     "@ledgerhq/types-cryptoassets": "workspace:*",
     "@ledgerhq/cryptoassets": "workspace:*",
     "@ledgerhq/logs": "workspace:*",
+    "@ledgerhq/live-env": "workspace:*",
     "@ledgerhq/live-countervalues": "workspace:*",
     "@ledgerhq/coin-framework": "workspace:*",
     "@ledgerhq/live-hooks": "workspace:*",
+    "@reduxjs/toolkit": "catalog:",
```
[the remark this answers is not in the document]
LL782: Agree. Not longer needs until we implement the RTK api

libs/live-countervalues-react/src/useIdsByMarketCap.tsx
LL782: Change of plan, let's push forward with the full introduction of RTK before merging.

Why: The comments below from Copilot highlight real flaws in the current `useSyncExternalStore` solution. Rather than addressing them let's just fix forwards

PR #14458, opened 2026-02-13
[LWDM] feat(LIVE-25649): Display APY in green for non-UK users in MAD

cseille:
## Summary

- Display APY indicator in green (success appearance) for non-UK users in the Modular Asset Drawer
- Keep APY indicator in grey for UK users based on locale detection
- Uses lumen Tag component with conditional `appearance` prop instead of the pre-ldls ApyIndicator

## Changes

**Desktop (`ledger-live-desktop`):**
- Replace `ApyIndicator` from `@ledgerhq/react-ui/pre-ldls` with `Tag` from `@ledgerhq/lumen-ui-react`
- Use `getParsedSystemDeviceLocale().region` for UK detection

**Mobile (`live-mobile`):**
- Replace `ApyIndicator` from `@ledgerhq/native-ui/pre-ldls` with `Tag` from `@ledgerhq/lumen-ui-rnative`
- Use `getCountryLocale()` from existing helper for UK detection

## Acceptance Criteria

- [x] Non-UK users see APY in green color
- [x] UK users see APY in grey color (unchanged behavior)
- [x] Only impacts Modular Asset Drawer in LLD and LLM
- [x] APY value and format preserved (`~ X.XX% APY`)

## Test Plan

- [x] Verify APY displays in green for non-UK locale
- [x] Verify APY displays in grey for UK locale (GB)
- [x] Verify APY value is correctly displayed
- [x] Test on both desktop and mobile

## Jira

https://ledgerhq.atlassian.net/browse/LIVE-25649

--------

# Beth updates: 

## Demo 

Desktop. Mac System locale is set to en-FR: 
image

Mobile from UK: 
Screenshot (5 Mar 2026 17_49_25)

- Adjusted to follow the "dumb view + smart model" architecture
- APY Indicator logic differs between platforms with respect to how we get the system locale from the device. So the view model hook could not be centralised in common. Instead, the logic lives next to the APY indicator component that uses it. 
- Tests added
- Note: we had to move APY Indicator to the modular _drawer_ component to avoid circular imports because the core drawer opening logic is still imported from _drawer_ into the _dialog_. if the drawer is eventually fully replaced by the dialog then the indicator should be moved, too.


apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/components/AccountApy.tsx
```diff
@@ -1,7 +1,14 @@
...
+import { Tag } from "@ledgerhq/lumen-ui-rnative";
 import { ApyType } from "@ledgerhq/live-common/dada-client/types/trend";
+import { getApyAppearance } from "@ledgerhq/live-common/modularDrawer/utils/getApyAppearance";
+import { getCountryLocale } from "~/helpers/getStakeLabelLocaleBased";
 
 export const accountsApy = ({ value, type }: { value?: number; type?: ApyType }) => {
-  return value && type ? <ApyIndicator value={value} type={type} /> : undefined;
+  if (!value || !type) return undefined;
+
+  const region = getCountryLocale();
```
gre-ledger: looking at the code of the underlying module `react-native-localize` , i'm seeing that the getCountry code will call in sync on the native module https://github.com/zoontek/react-native-localize/blob/master/src/module.native.ts#L15

therefore, calling this code can have some CPU impact,
i would prefer we product it with a `useMemo` so when we re-render this component each time, it doesn't have to be called again 🤔  wdyt?

cseille: Make sense, it won't change mid-session anyway

gre-ledger (approved):
ok for me (afaik no platform code ownership on these files)

apps/ledger-live-desktop/src/mvvm/features/ModularDialog/components/Account/AccountCountApy.tsx
```diff
@@ -14,6 +14,7 @@ export const accountsCountAndApy = ({
 }) => (
   <>
     {label && accountsCount({ label })}
-    {value && type ? <ApyIndicator value={value} type={type} /> : undefined}
+    {/* Allow valid 0% APY to display */}
```
mcayuelas-ledger: Hmm it's a requirement?

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/components/AccountApy.tsx
```diff
@@ -1,7 +0,0 @@
-import React from "react";
-import { ApyIndicator } from "@ledgerhq/native-ui/pre-ldls/index";
```
mcayuelas-ledger, line 2: if not used you can remove it from the lib also :) 

For Mobile & Desktop 🙏🏼

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/__integrations__/modules.test.tsx
```diff
@@ -231,4 +237,66 @@ describe("ModularDrawer modules integration", () => {
...
+      advanceTimers();
+
+      const [firstApyTag] = UNSAFE_getAllByProps({ testID: "apy-indicator" });
+      expect(firstApyTag.props.appearance).toBe("gray");
+    });
+
+    it("should render the APY tag with success appearance for non-GB users on asset list", async () => {
+      jest.mocked(getStakeLabelHelpers.getCountryLocale).mockReturnValue("US");
+
+      const { getByText, UNSAFE_getAllByProps, user } = render(
```
mcayuelas-ledger: `getByTestId` is better to use IMO

apps/ledger-live-desktop/src/mvvm/features/ModularDrawer/components/ApyIndicator/ApyView.tsx
```diff
@@ -0,0 +1,20 @@
+import { Tag } from "@ledgerhq/lumen-ui-react";
+import React from "react";
+import type { ApyProps } from ".";
+
+export const ApyView = ({
+  value,
+  type,
+  appearance,
+}: ApyProps & {
+  appearance: "success" | "gray";
```
mcayuelas-ledger, line 10: As you have exactly the same of both LWM & LWD, maybe you can put it in `common`?

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/__integrations__/modules.test.tsx
beths-ledger: I agree in general, but in this instance `getAllByTestId` returns the native element (the underlying View), which only has native props. appearance is a React component prop on `Tag` that never reaches the native layer.

Hence in order to test the React prop that we pass to Lumen, the fix is `UNSAFE_getAllByProps({ testID: "apy-indicator" })` because it queries React component instances and has all their props.

apps/ledger-live-desktop/src/mvvm/features/ModularDrawer/components/ApyIndicator/ApyView.tsx
beths-ledger: One uses `Tag` for React and the other `Tag` for React Native. I could add a conditional `container` that checks, but that might be more complex? LMK if you had another idea in mind

apps/ledger-live-desktop/src/mvvm/features/ModularDialog/components/Account/AccountCountApy.tsx
beths-ledger: just a reasonable assumption given  `0 !== undefined` but I'll revert to avoid any unintended behavioural changes 👍

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/__integrations__/modules.test.tsx
beths-ledger: i've changed the implementation instead to include the appearance in the test ID, then we can just use the ordinary selector fns 👍

apps/ledger-live-desktop/src/mvvm/features/ModularDrawer/components/ApyIndicator/ApyView.tsx
mcayuelas-ledger: I had in mind `ApyProps` type to be shared in `libs/ledger-live-common/src/modularDrawer/types`

libs/ui/packages/native/src/pre-ldls/components/index.ts
```diff
@@ -1,7 +1,6 @@
 export * from "./Input/Input";
 export * from "./AssetItem/AssetItem";
 export * from "./AssetTypeList/AssetList";
-export * from "./ApyIndicator/ApyIndicator";
 export * from "./MarketPriceIndicator/MarketPriceIndicator";
 export * from "./MarketPercentIndicator/MarketPercentIndicator";
 export * from "./NetworkItem/NetworkItem";
```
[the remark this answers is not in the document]
beths-ledger: bumped

libs/ui/packages/react/src/pre-ldls/components/index.ts
```diff
@@ -4,7 +4,6 @@ export * from "./Address/Address";
 export * from "./AssetItem/AssetItem";
 export * from "./AssetList/AssetList";
 export * from "./CardButton/CardButton";
-export * from "./ApyIndicator/ApyIndicator";
 export * from "./MarketPriceIndicator/MarketPriceIndicator";
 export * from "./MarketPercentIndicator/MarketPercentIndicator";
```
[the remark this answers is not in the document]
beths-ledger: bumped

beths-ledger (approved):

PR #14488, opened 2026-02-16
Fix visibility check for last operations in subAccount test

gre-ledger:
### ✅ Checklist

  - desktop e2e tests

### 📝 Description

This was peer programmed and investigated with @iqbalibrahim-ledger 

If the test is faster than the screen to actually navigate, we fall into a flaky case where the "last-operations" of the previous screen is still found in the DOM while the new screen didn't had time to remount and have the "last-operations" of the token account.

therefore, it makes more sense to first await for the TokenAccount to be visible, before scrolling to last operations.

See for yourself the error we constantly had in context of React 19 (which makes screen mounting / navigation more async with the Suspense behavior of React 19)

Screenshot 2026-02-16 at 14 21 00

---

NB: i suspect there may more case like this we may want to find: basically we must not expect that a `click()` that produce a navigation have straightaway made the app navigated to the other screen. i'm attempting a general study of this in another PR.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/e2e/desktop/tests/specs/subAccount.spec.ts
+++ b/e2e/desktop/tests/specs/subAccount.spec.ts
@@ -97,8 +97,8 @@ for (const token of subAccounts) {
           await app.portfolio.navigateToAsset(token.account.currency.name);
         }
         await app.account.navigateToToken(token.account);
-        await app.account.expectLastOperationsVisibility();
         await app.account.expectTokenAccount(token.account);
+        await app.account.expectLastOperationsVisibility();
       },
     );
   });


gre-ledger:
the PR #14492 correct reveals the problem to exist by slowing down react router navigation

    tests/specs/subAccount.spec.ts:57:9 › Add subAccount without parent › Add Sub Account without parent (Stellar) - USDC @NanoSP @LNS @NanoX @Stax @Flex @NanoGen5 @stellar @family-stellar @smoke 


we already confirmed with @iqbalibrahim-ledger in React 19 context that this test change should be fixed, so we'll merge the Pr and i'll then rebase on the study PR

PR #14523, opened 2026-02-17
test(lwd): playwright for the new send flow

dilaouid:
### ✅ Checklist

  - ...

### 📝 Description

Apply e2e tests for the new send flow on LWD, list of tested scenarios:

#### Recipient Step
- Open dialog with address input visible
- Show address matched after typing valid address
- Show sanctioned banner for sanctioned address
- Proceed to Amount step when clicking valid address
- Show recent history warning for address with no prior OUT operations
- Resolve ENS domain and allow proceed (EVM)
- Allow self-transfer when policy is free (EVM)
- Block self-transfer when policy is impossible (Tezos)

#### Memo skip
- Skip memo and need to confirm
- No confirm when "never ask again" enabled from send modal
- No confirm when "never ask again" enabled from settings
- Need to confirm when "never ask again" disabled from settings

#### Memo on Stellar
- Memo Text type as default option
- Show all Stellar memo types in selector
- Switch to NO_MEMO / MEMO_TEXT / MEMO_ID / MEMO_HASH / MEMO_RETURN, fill and go to amount step (×5)

#### Amount Step
- Show amount input and quick actions
- Allow entering valid amount
- Allow entering fiat and crypto amounts
- Show error for insufficient funds and Get Funds button
- Disable review button for zero amount
- Set correct amount when clicking quick actions (skipped)
- Navigate back to Recipient
- Proceed to Signature step when clicking Review
- Allow changing fee presets (EVM)
- Allow changing fee presets (BTC)
- Show warning when fees > 10% of amount (BTC)
- Change max amount when changing fee presets (Bitcoin, Ethereum)
- Show custom + coin control in fees menu (BTC)
- Show custom fees in fees menu (EVM)
- No fee menu for Tezos / Algorand / XRP
- Show custom fees only for Stellar

#### Generic Flow
- Complete full flow: Algorand, Bitcoin, Ethereum, Arbitrum, Tezos, Stellar, XRP (×7)

#### Validation & Errors
- Validate address format (Ethereum invalid address)
- Transition from invalid to valid address

#### Navigation
- Close from recipient / amount / signature step and at end of workflow (×4)
- Navigate back and forth (Recipient ↔ Amount)
- Prefill recipient when editing from Amount header

#### Signature & Confirmation Errors
- Show confirmation info when user refuses on device
- Show confirmation error when device disconnected

#### Edge Cases
- Show security intro card in initial state (accordion collapsed by default, expand on click)

#### Feature Flag
- New flow when flag enabled for EVM
- Old flow when flag disabled
- Family-specific flag respected for EVM

### ❓ Context

- **[JIRA or GitHub link](https://ledgerhq.atlassian.net/browse/LIVE-22192)**: 

---

### DIFF
--- a/apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
+++ b/apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
@@ -1,30 +1,35 @@
 import { expect } from "@playwright/test";
-import { FeePreset } from "tests/common/newSendFlow/types";
 import { step } from "tests/misc/reporters/step";
-import { Component } from "tests/page/abstractClasses";
+import { Modal } from "../../component/modal.component";
+import { looksLikeFiatValue } from "../../utils/looksLikeFiatValue";
 
 export type StellarMemoType = "NO_MEMO" | "MEMO_TEXT" | "MEMO_ID" | "MEMO_HASH" | "MEMO_RETURN";
 
-export class NewSendFlowPage extends Component {
+type FeePreset = "slow" | "medium" | "fast";
+
+export class NewSendModal extends Modal {
   // ========== Dialog container ==========
   readonly dialog = this.page.getByRole("dialog");
 
   // ========== Header elements ==========
   readonly headerTitle = this.dialog.getByTestId("send-dialog-header");
   readonly backButton = this.headerTitle.getByRole("button", { name: "Go back" });
-  readonly closeButton = this.headerTitle.getByRole("button", { name: "Close" });
+  override readonly closeButton = this.headerTitle.getByRole("button", { name: "Close" });
 
   readonly recipientInput = this.dialog.getByTestId("send-recipient-input");
   readonly editRecipientButton = this.dialog.getByTestId("send-edit-recipient-button");
 
   // ========== RECIPIENT STEP ==========
-  readonly myAccountsSection = this.dialog.getByTestId("send-my-accounts-section");
   readonly addressMatchedTitle = this.dialog.getByTestId("send-address-matched-title");
-  // Recent addresses are rendered as tiles with a "More actions" secondary button.
-  readonly recentTileMoreActionsButtons = this.dialog.getByTestId("send-recent-tile-action");
+  readonly introCard = this.dialog.getByTestId("send-recipient-intro-card");
+  readonly securityToggle = this.dialog.getByTestId("send-recipient-security-toggle");
 
   // Primary selectable button in "Address matched" section (ListItem is a button with "Send to X" text)
   readonly sendToButton = this.dialog.getByTestId("send-matched-address-button");
+  readonly visibleSendToButton = this.dialog
+    .locator('[data-testid="send-matched-address-button"]')
+    .filter({ visible: true })
+    .first();
   readonly addressNotFoundText = this.dialog.getByText(/address not found/i).first();
   readonly validationStatusMessage = this.dialog.getByTestId("address-validation-status");
 
@@ -32,11 +37,12 @@ export class NewSendFlowPage extends Component {
   readonly sanctionedBanner = this.dialog.getByTestId("sanctioned-address-banner");
   readonly recipientErrorBanner = this.dialog.getByTestId("recipient-error-banner");
   readonly senderErrorBanner = this.dialog.getByTestId("sender-error-banner");
-  readonly newAddressBanner = this.dialog.getByText(/sending to a new address/i).first();
+  readonly recentHistoryWarningCard = this.dialog.getByTestId("send-recent-history-warning");
 
   // Memo
   readonly memoInput = this.dialog.getByTestId("send-memo-input");
   readonly memoOptionsSelect = this.dialog.getByTestId("send-memo-options-select");
+  // Select content is portalled outside the dialog subtree.
   readonly memoOptionNoMemo = this.page.getByTestId("send-memo-select-option-NO_MEMO");
   readonly memoOptionText = this.page.getByTestId("send-memo-select-option-MEMO_TEXT");
   readonly memoOptionId = this.page.getByTestId("send-memo-select-option-MEMO_ID");
@@ -57,17 +63,51 @@ export class NewSendFlowPage extends Component {
   readonly quickAction75 = this.dialog.getByTestId("send-quick-actions-threeQuarters");
   readonly quickActionMax = this.dialog.getByTestId("send-quick-actions-max");
   readonly feesMenuTrigger = this.dialog.getByTestId("send-network-fees-menu-trigger");
+  // Menu items are portalled outside the dialog subtree.
   readonly customFeesMenuItem = this.page.getByTestId("send-custom-fees-menu-item");
   readonly coinControlFeesMenuItem = this.page.getByTestId("send-coin-control-fees-menu-item");
   readonly reviewButton = this.dialog.getByTestId("send-review-button");
   readonly getFundsButton = this.dialog.getByTestId("send-get-funds-button");
   // Toggle button to switch between FIAT and CRYPTO input modes
   readonly toggleInputModeButton = this.dialog.getByTestId("send-toggle-input-mode-button");
+  readonly amountSecondaryValue = this.dialog.getByTestId("send-amount-secondary-value");
+
+  getFeePreset(preset: FeePreset) {
+    return this.page.getByTestId(`send-fees-preset-${preset}`);
+  }
+
+  // ========== CUSTOM FEES STEP ==========
+  readonly customFeesConfirmButton = this.dialog.getByRole("button", { name: /confirm/i });
+  readonly customFeesFeePerByteInput = this.dialog.getByLabel(/fees amount.*sat\/vbyte/i);
+  readonly customFeesMaxFeeInput = this.dialog.getByLabel(/max fee.*gwei/i);
+  readonly customFeesMaxPriorityFeeInput = this.dialog.getByLabel(/max priority fee.*gwei/i);
+
+  // ========== COIN CONTROL STEP ==========
+  readonly coinControlFooter = this.dialog.getByTestId("send-coin-control-footer");
+  readonly coinControlChangeToReturn = this.dialog.getByTestId("send-change-to-return-row");
+  readonly coinControlAmountInput = this.dialog.getByLabel(/amount to send/i);
 
   // ========== CONFIRMATION STEP ==========
-  readonly confirmationStatusContent = this.dialog.locator(
-    '[data-testid="send-confirmation-success-content"], [data-testid="send-confirmation-error-content"], [data-testid="send-confirmation-info-content"]',
+  readonly confirmationSuccessContent = this.dialog.getByTestId(
+    "send-confirmation-success-content",
   );
+  readonly confirmationErrorContent = this.dialog.getByTestId("send-confirmation-error-content");
+  readonly confirmationInfoContent = this.dialog.getByTestId("send-confirmation-info-content");
+  /**
+   * Any terminal confirmation screen (success, error, or info).
+   * Use `waitForSuccessConfirmation()` when asserting success specifically.
+   */
+  readonly confirmationTerminalContent = this.confirmationSuccessContent
+    .or(this.confirmationErrorContent)
+    .or(this.confirmationInfoContent);
+  readonly successConfirmationTitle = this.dialog.getByTestId("send-confirmation-success-title");
+
+  private async isCryptoInputMode(): Promise<boolean | null> {
+    const secondaryValue = (await this.amountSecondaryValue.textContent())?.trim() ?? "";
+    // Return null when the element isn't populated yet so callers can retry.
+    if (!secondaryValue) return null;
+    return looksLikeFiatValue(secondaryValue);
+  }
 
   // ========== DIALOG LIFECYCLE ==========
 
@@ -76,11 +116,6 @@ export class NewSendFlowPage extends Component {
     await this.dialog.waitFor({ state: "visible", timeout: 15000 });
     // Wait for animations
     await this.page.waitForTimeout(500);
-  }
-
-  @step("Close dialog")
-  async close() {
-    await this.closeButton.click();
   }
 
   @step("Click back button")
@@ -95,71 +130,50 @@ export class NewSendFlowPage extends Component {
   @step("Type address in search input: $0")
   async typeAddress(address: string) {
     await this.recipientInput.waitFor({ state: "visible" });
-    await this.recipientInput.clear();
     await this.recipientInput.fill(address);
   }
 
-  @step("Select address from list (index: $0)")
+  @step("Select address from list")
   async clickOnSendToButton() {
-    const visibleSendToButton = this.dialog
-      .locator('[data-testid="send-matched-address-button"]:visible')
-      .first();
-    await visibleSendToButton.waitFor({ state: "visible", timeout: 30000 });
+    await this.visibleSendToButton.waitFor({ state: "visible", timeout: 30000 });
     // Bridge async validation may cause re-renders; wait for DOM to stabilize
     await this.page.waitForTimeout(500);
-    await visibleSendToButton.scrollIntoViewIfNeeded();
-    await visibleSendToButton.click();
-  }
-
-  @step("Remove recent address tile (index: $0)")
-  async removeRecentAddressTile(index: number = 0) {
-    const moreActions = this.recentTileMoreActionsButtons.nth(index);
-    await moreActions.waitFor({ state: "visible", timeout: 10000 });
-    await moreActions.click();
-
-    const removeItem = this.page.getByRole("menuitem", { name: /remove/i });
-    await removeItem.waitFor({ state: "visible", timeout: 10000 });
-    await removeItem.dispatchEvent("click");
-
-    // Menu closes asynchronously.
-    await this.page.waitForTimeout(300);
+    await this.visibleSendToButton.scrollIntoViewIfNeeded();
+    await this.visibleSendToButton.click();
   }
 
   @step("Wait for address validation")
   async waitForRecipientValidation() {
     const timeoutMs = 20000;
-    const waiters = [
-      this.sendToButton.waitFor({ state: "visible", timeout: timeoutMs }),
-      this.addressNotFoundText.waitFor({ state: "visible", timeout: timeoutMs }),
-      this.validationStatusMessage.waitFor({ state: "visible", timeout: timeoutMs }),
-      this.sanctionedBanner.waitFor({ state: "visible", timeout: timeoutMs }),
-      this.recipientErrorBanner.waitFor({ state: "visible", timeout: timeoutMs }),
-      this.senderErrorBanner.waitFor({ state: "visible", timeout: timeoutMs }),
-      this.skipMemoProposal.waitFor({ state: "visible", timeout: timeoutMs }),
-    ];
-    try {
-      await Promise.any(waiters);
-    } catch {
-      throw new Error(
-        `Recipient validation did not complete within ${timeoutMs}ms: none of the expected states (Send to button, address not found, validation status, sanctioned/recipient/sender error banners, skip memo proposal) became visible.`,
-      );
-    }
-    await this.page.waitForTimeout(250);
+    await expect
+      .poll(
+        async () => {
+          if (await this.sendToButton.isVisible().catch(() => false)) return "sendToButton-visible";
+          if (await this.addressNotFoundText.isVisible().catch(() => false))
+            return "addressNotFound-visible";
+          if (await this.validationStatusMessage.isVisible().catch(() => false))
+            return "validationStatus-visible";
+          if (await this.sanctionedBanner.isVisible().catch(() => false))
+            return "sanctionedBanner-visible";
+          if (await this.recipientErrorBanner.isVisible().catch(() => false))
+            return "recipientErrorBanner-visible";
+          if (await this.senderErrorBanner.isVisible().catch(() => false))
+            return "senderErrorBanner-visible";
+          if (await this.skipMemoProposal.isVisible().catch(() => false))
+            return "skipMemoProposal-visible";
+          return null;
+        },
+        {
+          timeout: timeoutMs,
+          message: `Recipient validation did not complete within ${timeoutMs}ms: none of the expected states (Send to button, address not found, validation status, sanctioned/recipient/sender error banners, skip memo proposal) became visible.`,
+        },
+      )
+      .not.toBeNull();
   }
 
   @step("Verify sanctioned banner visible")
   async expectSanctionedBanner() {
     await expect(this.sanctionedBanner).toBeVisible({ timeout: 5000 });
-  }
-
-  @step("Verify recipient error banner visible")
-  async expectRecipientError() {
-    await expect(this.recipientErrorBanner).toBeVisible({ timeout: 5000 });
-  }
-
-  @step("Verify sender error banner visible")
-  async expectSenderError() {
-    await expect(this.senderErrorBanner).toBeVisible({ timeout: 5000 });
   }
 
   @step("Verify address matched section visible")
@@ -174,31 +188,8 @@ export class NewSendFlowPage extends Component {
     await this.page.waitForTimeout(300);
   }
 
-  @step("Verify validation message contains: $0")
-  async expectValidationMessage(text: RegExp | string) {
-    const hasValidationMessage = await this.validationStatusMessage.isVisible().catch(() => false);
-    if (hasValidationMessage) {
-      await expect(this.validationStatusMessage).toContainText(text, { timeout: 10000 });
-      return;
-    }
-
-    const hasAddressNotFound = await this.addressNotFoundText.isVisible().catch(() => false);
-    if (hasAddressNotFound) {
-      await expect(this.addressNotFoundText).toBeVisible({ timeout: 5000 });
-      return;
-    }
-
-    const hasRecipientBanner = await this.recipientErrorBanner.isVisible().catch(() => false);
-    if (hasRecipientBanner) {
-      await expect(this.recipientErrorBanner).toBeVisible({ timeout: 5000 });
-      return;
-    }
-
-    throw new Error("No validation message or error banner appeared");
-  }
-
   @step("Skip memo")
-  async skipMemo(confirm: boolean = true) {
+  async skipMemo({ confirm = true }: { confirm?: boolean } = {}) {
     await this.skipMemoLink.click();
     if (confirm) {
       await this.confirmSkipMemo();
@@ -253,36 +244,25 @@ export class NewSendFlowPage extends Component {
 
   @step("Switch to crypto input mode")
   async switchToCryptoMode() {
-    // By default, the amount input is in FIAT mode. Click toggle to switch to CRYPTO.
-    const toggleButton = this.toggleInputModeButton;
-    const isVisible = await toggleButton.isVisible().catch(() => false);
-    if (isVisible) {
-      await toggleButton.click();
-      await this.page.waitForTimeout(300);
+    // Poll until secondary value is populated (null = not ready yet)
+    await expect.poll(() => this.isCryptoInputMode()).not.toBeNull();
+    const isCryptoMode = await this.isCryptoInputMode();
+    if (!isCryptoMode) {
+      await this.toggleInputModeButton.click();
+      // null won't match true, so the poll keeps retrying until the switch completes
+      await expect.poll(() => this.isCryptoInputMode()).toBe(true);
     }
   }
 
-  @step("Wait for bridge to be ready (not loading)")
-  async waitForBridgeReady(timeout: number = 30000) {
-    // Wait for the review button to stop showing loading state
-    // The button has loading={reviewLoading} which is true when bridgePending && shouldPrepare
-    const start = Date.now();
-    while (Date.now() - start < timeout) {
-      const isLoading = await this.reviewButton.getAttribute("data-loading").catch(() => null);
-      const isDisabled = await this.reviewButton.isDisabled().catch(() => true);
-
-      // If button is visible and not in a loading state, bridge is ready
-      if (isLoading !== "true" && !isDisabled) {
-        return;
-      }
-
-      // Also check if button becomes enabled
-      const isEnabled = await this.reviewButton.isEnabled().catch(() => false);
-      if (isEnabled) {
-        return;
-      }
-
-      await this.page.waitForTimeout(250);
+  @step("Switch to fiat input mode")
+  async switchToFiatMode() {
+    // Poll until secondary value is populated (null = not ready yet)
+    await expect.poll(() => this.isCryptoInputMode()).not.toBeNull();
+    const isCryptoMode = await this.isCryptoInputMode();
+    if (isCryptoMode) {
+      await this.toggleInputModeButton.click();
+      // null won't match false, so the poll keeps retrying until the switch completes
+      await expect.poll(() => this.isCryptoInputMode()).toBe(false);
     }
   }
 
@@ -302,7 +282,7 @@ export class NewSendFlowPage extends Component {
   @step("Fill amount: $0")
   async fillAmount(amount: string) {
     await this.amountInput.waitFor({ state: "visible", timeout: 10000 });
-    await this.amountInput.clear();
+    await this.switchToFiatMode();
     await this.amountInput.fill(amount);
     await this.page.waitForTimeout(800); // Wait for validation
   }
@@ -312,7 +292,6 @@ export class NewSendFlowPage extends Component {
     await this.amountInput.waitFor({ state: "visible", timeout: 10000 });
 
     await this.switchToCryptoMode();
-    await this.amountInput.clear();
     await this.amountInput.fill(amount);
   }
 
@@ -350,12 +329,6 @@ export class NewSendFlowPage extends Component {
     await this.page.waitForTimeout(500);
   }
 
-  @step("Click review to proceed to signature")
-  async clickReview2() {
-    await this.reviewButton.waitFor({ state: "visible" });
-    await this.reviewButton.click();
-  }
-
   @step("Verify review button is disabled")
   async expectReviewDisabled() {
     await expect(this.reviewButton).toBeDisabled({ timeout: 10000 });
@@ -382,12 +355,6 @@ export class NewSendFlowPage extends Component {
     await expect(this.getFundsButton).toBeVisible({ timeout: 5000 });
   }
 
-  @step("Click Get Funds button")
-  async clickGetFunds() {
-    await this.getFundsButton.waitFor({ state: "visible", timeout: 5000 });
-    await this.getFundsButton.click();
-  }
-
   @step("Open fees menu")
   async openFeesMenu() {
     await this.feesMenuTrigger.waitFor({ state: "visible" });
@@ -398,70 +365,60 @@ export class NewSendFlowPage extends Component {
   async selectFeePreset(preset: FeePreset) {
     await this.openFeesMenu();
 
-    const item = this.page.getByTestId(`send-fees-preset-${preset}`);
+    const item = this.getFeePreset(preset);
     await item.waitFor({ state: "visible", timeout: 10000 });
     await item.click();
 
     await this.page.waitForTimeout(300);
   }
 
-  async getFeePreset(preset: FeePreset) {
-    return this.page.getByTestId(`send-fees-preset-${preset}`);
+  // ========== CUSTOM FEES STEP METHODS ==========
+
+  @step("Open custom fees screen")
+  async openCustomFees() {
+    await this.openFeesMenu();
+    await this.customFeesMenuItem.waitFor({ state: "visible" });
+    await this.customFeesMenuItem.click();
+    await this.page.waitForTimeout(500);
+  }
+
+  @step("Open coin control screen")
+  async openCoinControl() {
+    await this.openFeesMenu();
+    await this.coinControlFeesMenuItem.waitFor({ state: "visible" });
+    await this.coinControlFeesMenuItem.click();
+    await this.page.waitForTimeout(500);
+  }
+
+  @step("Confirm custom fees")
+  async confirmCustomFees() {
+    await this.customFeesConfirmButton.waitFor({ state: "visible" });
+    await this.customFeesConfirmButton.click();
+    await this.page.waitForTimeout(500);
   }
 
   // ========== SIGNATURE STEP METHODS ==========
 
+  // The device loader is rendered globally, outside the dialog subtree.
   readonly deviceActionLoader = this.page.getByTestId("device-action-loader");
+  readonly signaturePrompt = this.dialog.getByTestId("send-signature-prompt");
 
-  @step("Wait for signature screen and device action loader")
+  @step("Wait for signature screen or device action loader")
   async waitForSignature() {
-    // Wait for device action loader to appear - this means we're in signature step
-    await this.deviceActionLoader.waitFor({ state: "visible" });
+    await this.deviceActionLoader.or(this.signaturePrompt).waitFor({ state: "visible" });
   }
 
   // ========== CONFIRMATION STEP METHODS ==========
 
-  @step("Wait for confirmation screen")
+  @step("Wait for terminal confirmation screen")
   async waitForConfirmation() {
-    await this.confirmationStatusContent.waitFor({ state: "visible" });
+    await this.confirmationTerminalContent.waitFor({ state: "visible" });
   }
 
-  @step("Wait for signature or confirmation step")
-  async waitForSignatureOrConfirmation() {
-    const timeout = 20000;
-    const start = Date.now();
-
-    while (Date.now() - start < timeout) {
-      const isSignatureLoaderVisible = await this.deviceActionLoader.isVisible().catch(() => false);
-      if (isSignatureLoaderVisible) {
-        return;
-      }
-
-      const isConfirmationVisible = await this.confirmationStatusContent
-        .isVisible()
-        .catch(() => false);
-      if (isConfirmationVisible) {
-        return;
-      }
-
-      await this.page.waitForTimeout(250);
-    }
-
-    throw new Error("Timed out waiting for signature or confirmation step");
-  }
-
-  @step("Verify success confirmation")
-  async expectSuccess() {
-    await expect(this.dialog.getByTestId("send-confirmation-success-title")).toBeVisible({
-      timeout: 10000,
-    });
-  }
-
-  @step("Verify error confirmation")
-  async expectError() {
-    await expect(this.dialog.getByTestId("send-confirmation-error-title")).toBeVisible({
-      timeout: 10000,
-    });
+  @step("Wait success confirmation screen")
+  async waitForSuccessConfirmation() {
+    await this.waitForConfirmation();
+    await expect(this.successConfirmationTitle).toBeVisible();
   }
 
   // ========== UTILITY METHODS ==========
@@ -478,10 +435,6 @@ export class NewSendFlowPage extends Component {
 
   async getAmountValue(): Promise<string> {
     return (await this.amountInput.inputValue()) || "";
-  }
-
-  async getAddressValue(): Promise<string> {
-    return (await this.recipientInput.inputValue()) || "";
   }
 
   @step("Going to the amount step")
--- a/apps/ledger-live-desktop/tests/page/settings.page.ts
+++ b/apps/ledger-live-desktop/tests/page/settings.page.ts
@@ -80,7 +80,6 @@ export class SettingsPage extends AppPage {
 
   @step("Switch never ask again skip memo")
   async switchNeverAskAgainSkipMemo() {
-    await this.neverAskAgainSkipMemoSwitch.waitFor({ state: "visible" });
     await this.neverAskAgainSkipMemoSwitch.click();
   }
 
--- a/apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
+++ b/apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
@@ -5,54 +5,39 @@ import { AccountsPage } from "../../page/accounts.page";
 import { Layout } from "../../component/layout.component";
 import { DeviceAction } from "../../models/DeviceAction";
 import { Addresses } from "@ledgerhq/live-common/e2e/enum/Addresses";
-import { updateRecentAddresses } from "@ledgerhq/live-wallet/store";
-import type { RecentAddressesState } from "@ledgerhq/types-live";
 import { BigNumber } from "bignumber.js";
 import { Application } from "tests/page";
-import { FEE_PRESETS } from "tests/common/newSendFlow/types";
+import { StellarMemoType } from "tests/page/modal/new.send.modal";
 
-import { StellarMemoType } from "tests/page/modal/newSendFlow.page";
+const FEE_PRESETS = ["slow", "medium", "fast"] as const;
 
 // Test addresses per family (using abandon seed addresses for safety)
 const TEST_ADDRESSES = {
   ethereum: "0x000000000000000000000000000000000000dEaD",
-  polygon: "0x000000000000000000000000000000000000dEaD",
   arbitrum: "0x000000000000000000000000000000000000dEaD",
-  solana: Addresses.SOL_OTHER_SEED,
   bitcoin: "1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa",
   tezos: "tz1aWXP237BLwNHJcCD4b3DutCevhqq2T1Z9",
-  tron: "TQDyHP3QX24zSDn5418DHm2hvYs2wzuYYY",
-  near: "test-pristine.testnet",
-  algorand: "PSHLIWQKDEETIIBQEOTLGCT5IF7BTTOKCUULONOGVGF2HYDT2IHW3H4CCI", // abandonSeed address
+  /** Algorand 1 freshAddress in userdata — self-transfer is impossible on Algorand. */
+  algorand: "PSHLIWQKDEETIIBQEOTLGCT5IF7BTTOKCUULONOGVGF2HYDT2IHW3H4CCI",
+  /** Valid Algorand recipient, distinct from Algorand 1 account (coin-algorand bridge test dataset). */
+  algorandRecipient: "MVE6C3XB4JBKXKORC3NLAWFW4M7EY3MADU6L72DADFP4NZBJIAYXGSLN3Y",
   stellar: "GDYPMQMYW2JTLPWAUAHIDY3E4VHP5SGTFC5SMA45L7ZPOTHWQ2PHEW3E", // abandonSeed address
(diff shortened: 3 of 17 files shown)


apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
```diff
@@ -0,0 +1,1105 @@
...
+    });
+
+    test("should show custom fees only for Stellar", async ({ app, page }) => {
+      await reachAmountStep(app, page, ACCOUNT_NAMES.stellar, TEST_ADDRESSES.stellar, true);
+      await app.newSendFlow.fillCryptoAmount("1");
+      await app.newSendFlow.openFeesMenu();
+      await expect(app.newSendFlow.customFeesMenuItem).toBeVisible();
+
+      for (const preset of FEE_PRESETS) {
+        expect(await app.newSendFlow.getFeePreset(preset)).toBeHidden();
```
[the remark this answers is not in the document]
dilaouid:
```
Property 'toBeHidden' does not exist on type 'MakeMatchers<void, Promise<Locator> 
```

apps/ledger-live-desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,504 @@
...
+      throw new Error("Review button did not become enabled within timeout");
+    }
+    await this.reviewButton.click();
+    await this.page.waitForTimeout(500);
+  }
+
+  @step("Click review to proceed to signature")
+  async clickReview2() {
+    await this.reviewButton.waitFor({ state: "visible" });
+    await this.reviewButton.click();
```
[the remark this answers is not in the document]
dilaouid: fixed

apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
```diff
@@ -0,0 +1,1105 @@
...
+          await test.step("Verify old flow modal is used", async () => {
+            // With FF disabled, should use old modal
+            const hasOldModal = await page
+              .getByTestId("modal-container")
+              .isVisible()
+              .catch(() => false);
+            const hasNewDialog = await app.newSendFlow.dialog.isVisible().catch(() => false);
+
+            // One should be visible
+            expect(hasOldModal || hasNewDialog).toBe(true);
```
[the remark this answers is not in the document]
dilaouid: fixed

apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
```diff
@@ -0,0 +1,1105 @@
...
+    await page.waitForTimeout(500);
+
+    // Check if account exists
+    const accountExists = await page
+      .getByTestId(`account-component-${accountName}`)
+      .isVisible()
+      .catch(() => false);
+    if (!accountExists && shouldSkipIfMissing) {
+      test.skip();
+      return;
```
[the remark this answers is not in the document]
dilaouid: fixed

apps/ledger-live-desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,504 @@
...
+    const toggleButton = this.toggleInputModeButton;
+    const isVisible = await toggleButton.isVisible().catch(() => false);
+    if (isVisible) {
+      await toggleButton.click();
+      await this.page.waitForTimeout(300);
+    }
+  }
+
+  @step("Wait for bridge to be ready (not loading)")
+  async waitForBridgeReady(timeout: number = 30000) {
```
Moustafa-Koterba: can be removed

apps/ledger-live-desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,504 @@
...
+    await button.waitFor({ state: "visible", timeout: 5000 });
+    await button.click();
+    await this.page.waitForTimeout(500);
+  }
+
+  @step("Click review to proceed to signature")
+  async clickReview() {
+    await this.reviewButton.waitFor({ state: "visible", timeout: 10000 });
+    // Wait for button to be enabled (no errors, has amount, not bridgePending)
+    const isEnabled = await this.waitForReviewButtonEnabled(30000);
```
Moustafa-Koterba: I think there is no need for this, we expect the review button to be visible, the test should fail before

sorry I edited this one, and replaced the function to clickReview2 function, this function can be deleted

apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
```diff
@@ -0,0 +1,1063 @@
...
+import { Application } from "tests/page";
+import { Layout } from "../../component/layout.component";
+import test from "../../fixtures/common";
+import { DeviceAction } from "../../models/DeviceAction";
+import { AccountPage } from "../../page/account.page";
+import { AccountsPage } from "../../page/accounts.page";
+
+import { StellarMemoType } from "tests/page/modal/newSendFlow.page";
+
+const TEST_ADDRESSES = {
```
mcayuelas-ledger, line 15: you can extract all the data used in another file no?

dilaouid: Some are not used / exported anywhere, but I could extract them to example: `tests/common/newSendFlow/fixtures.ts`, but these addresses are only used in this spec and several are tied to this test’s userdata (e.g. `selfEthereum`, `selfTezos`, `ACCOUNT_NAMES`....)

wdyt? 🤔

dilaouid:
E2E tests:
- NanoS : https://github.com/LedgerHQ/ledger-live/actions/runs/22394034122
- NanoSP: https://github.com/LedgerHQ/ledger-live/actions/runs/22394761403
- NanoX: https://github.com/LedgerHQ/ledger-live/actions/runs/22394773662
- Stax: https://github.com/LedgerHQ/ledger-live/actions/runs/22394787080
- Flex: https://github.com/LedgerHQ/ledger-live/actions/runs/22394787080
- Gen5: https://github.com/LedgerHQ/ledger-live/actions/runs/22394794029

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+  // ========== Header elements ==========
+  readonly headerTitle = this.dialog.getByTestId("send-dialog-header");
+  readonly backButton = this.headerTitle.getByRole("button", { name: "Go back" });
+  readonly closeButton = this.headerTitle.getByRole("button", { name: "Close" });
+
+  readonly recipientInput = this.dialog.getByTestId("send-recipient-input");
+  readonly editRecipientButton = this.dialog.getByTestId("send-edit-recipient-button");
+
+  // ========== RECIPIENT STEP ==========
+  readonly myAccountsSection = this.dialog.getByTestId("send-my-accounts-section");
```
ypolishchuk-ledger: this element is not in use

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+
+  readonly sendToButton = this.dialog.getByTestId("send-matched-address-button");
+  readonly addressNotFoundText = this.dialog.getByText(/address not found/i).first();
+  readonly validationStatusMessage = this.dialog.getByTestId("address-validation-status");
+
+  // Banners
+  readonly sanctionedBanner = this.dialog.getByTestId("sanctioned-address-banner");
+  readonly recipientErrorBanner = this.dialog.getByTestId("recipient-error-banner");
+  readonly senderErrorBanner = this.dialog.getByTestId("sender-error-banner");
+  readonly newAddressBanner = this.dialog.getByText(/sending to a new address/i).first();
```
ypolishchuk-ledger: this element is not in use

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+
+  // Memo
+  readonly memoInput = this.dialog.getByTestId("send-memo-input");
+  readonly memoOptionsSelect = this.dialog.getByTestId("send-memo-options-select");
```
ypolishchuk-ledger, line 44: this element is not in use

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+
+  // ========== AMOUNT STEP ==========
+  readonly amountInput = this.dialog.getByTestId("send-amount-input");
+  readonly amountErrorMessage = this.dialog.getByTestId("send-amount-message");
+  readonly quickAction25 = this.dialog.getByTestId("send-quick-actions-quarter");
+  readonly quickAction50 = this.dialog.getByTestId("send-quick-actions-half");
+  readonly quickAction75 = this.dialog.getByTestId("send-quick-actions-threeQuarters");
+  readonly quickActionMax = this.dialog.getByTestId("send-quick-actions-max");
+  readonly feesMenuTrigger = this.dialog.getByTestId("send-network-fees-menu-trigger");
+  readonly customFeesMenuItem = this.page.getByTestId("send-custom-fees-menu-item");
```
ypolishchuk-ledger, line 67: this element is not in use

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+  readonly coinControlFeesMenuItem = this.page.getByTestId("send-coin-control-fees-menu-item");
```
ypolishchuk-ledger, line 68: this element is not in use

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+  // ========== CONFIRMATION STEP ==========
+  readonly confirmationStatusContent = this.dialog.locator(
+    '[data-testid="send-confirmation-success-content"], [data-testid="send-confirmation-error-content"], [data-testid="send-confirmation-info-content"]',
+  );
+
+  // ========== DIALOG LIFECYCLE ==========
+
+  @step("Wait for send flow dialog to open")
+  async waitForDialog() {
+    await this.dialog.waitFor({ state: "visible", timeout: 15000 });
```
ypolishchuk-ledger, line 116: we already have a default timeout in playwright, no need to specify timeouts in the method here and everywhere else in this file

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+    await this.page.waitForTimeout(500);
```
ypolishchuk-ledger, line 118: could you please avoid using waitForTimeout and instead find a more reliable way to wait for a specific event or element visibility? Applies to several places in this class

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
+import { expect } from "@playwright/test";
+import { step } from "tests/misc/reporters/step";
+import { Component } from "tests/page/abstractClasses";
+
+export type StellarMemoType = "NO_MEMO" | "MEMO_TEXT" | "MEMO_ID" | "MEMO_HASH" | "MEMO_RETURN";
+
+type FeePreset = "slow" | "medium" | "fast";
+
+export class NewSendFlowPage extends Component {
```
ypolishchuk-ledger: if this is a modal then this file should have the name "new.send.modal.ts" and class should have name "NewSendModal". Additionally, you should extend Modal call instead of Component

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+
+  @step("Wait for send flow dialog to open")
+  async waitForDialog() {
+    await this.dialog.waitFor({ state: "visible", timeout: 15000 });
+    await this.page.waitForTimeout(500);
+  }
+
+  @step("Close dialog")
+  async close() {
+    await this.closeButton.click();
```
ypolishchuk-ledger: close button click mathod already exists in the parent class Modal which you should extend :)

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+  }
+
+  @step("Click back button")
+  async goBack() {
+    await this.backButton.waitFor({ state: "visible" });
```
ypolishchuk-ledger, line 123: Playwright's actionable methods like .click(), .fill(), and .clear() automatically wait for the element to be visible and actionable before performing the action (with the default timeout).

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+  @step("Type address in search input: $0")
+  async typeAddress(address: string) {
+    await this.recipientInput.waitFor({ state: "visible" });
+    await this.recipientInput.clear();
+    await this.recipientInput.fill(address);
+  }
+
+  @step("Select address from list (index: $0)")
+  async clickOnSendToButton() {
+    const visibleSendToButton = this.dialog
```
ypolishchuk-ledger: move this element to the top of the class

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+    await visibleSendToButton.click();
+  }
+
+  @step("Remove recent address tile (index: $0)")
+  async removeRecentAddressTile(index: number = 0) {
+    const moreActions = this.recentTileMoreActionsButtons.nth(index);
+    await moreActions.waitFor({ state: "visible", timeout: 10000 });
+    await moreActions.click();
+
+    const removeItem = this.page.getByRole("menuitem", { name: /remove/i });
```
ypolishchuk-ledger: move this element to the top of the class and check all other occurrences please

e2e/desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,507 @@
...
+    await this.page.waitForTimeout(500);
+  }
+
+  @step("Close dialog")
+  async close() {
+    await this.closeButton.click();
+  }
+
+  @step("Click back button")
+  async goBack() {
```
ypolishchuk-ledger, line 122: this method is not in use anywhere, could you please remove it along with other methods in this class that are not in use? Thanks

apps/ledger-live-desktop/tests/page/settings.page.ts
```diff
@@ -75,6 +78,12 @@ export class SettingsPage extends AppPage {
     await this.counterValueropdownChoiceEuro.click();
   }
 
+  @step("Switch never ask again skip memo")
+  async switchNeverAskAgainSkipMemo() {
+    await this.neverAskAgainSkipMemoSwitch.waitFor({ state: "visible" });
```
martijnhjk: Do we have to wait for visible? Most frameworks have built in wait for visible and enabled for click methods...

martijnhjk: https://playwright.dev/docs/actionability#introduction

dilaouid: Looking at it !

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,440 @@
...
+  @step("Select fee preset: $0")
+  async selectFeePreset(preset: FeePreset) {
+    await this.openFeesMenu();
+
+    const item = this.page.getByTestId(`send-fees-preset-${preset}`);
+    await item.click();
+    await item.waitFor({ state: "hidden" });
+  }
+
+  async getFeePreset(preset: FeePreset) {
```
martijnhjk: this is not actually used, do you really need it?

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+  "xrp",
+  "bitcoin",
+  "kaspa",
+];
+
+const MEMO_FAMILIES = new Set(["algorand", "stellar", "xrp", "solana"]);
+
+const feeToPreset = (fee?: Fee): "slow" | "medium" | "fast" | undefined => {
+  if (!fee) return undefined;
+  return fee.toLowerCase() as "slow" | "medium" | "fast";
```
martijnhjk: Image
eslint is warning against inconsistent type checking.

a few observations:
- you could export and reuse `FeePreset` for consistency and reduce duplication.
- Since `Fee` is currently a string enum you could also remove `feeToPreset()` altogether and expect the `Fee` enum as a parameter in the modal page object, eg:

```
async selectFeePreset(preset: Fee) {
  ...
  const item = this.page.getByTestId(`send-fees-preset-${preset.toLowerCase()}`);
  ...
```

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: Yes (will be on the second part of the playwright)
But I remove it and I'll implement it on the other part with another PR 🙏

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
dilaouid: Fixed !

dilaouid:
E2E tests:
- NanoS : https://github.com/LedgerHQ/ledger-live/actions/runs/22430431385
- NanoSP: (soon)
- NanoX: (soon)
- Stax: (soon)
- Flex: (soon)
- Gen5: (soon)

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,520 @@
...
+    const removeItem = this.page.getByRole("menuitem", { name: /remove/i });
+    await removeItem.waitFor({ state: "visible", timeout: 10000 });
+    await removeItem.dispatchEvent("click");
+
+    // Menu closes asynchronously.
+    await this.page.waitForTimeout(300);
+  }
+
+  @step("Wait for address validation")
+  async waitForRecipientValidation() {
```
cunhabruno, line 146: I can see this step is failing in Allure report is it known / expected?

dilaouid: I think those are some Tron or Speculos related flakiness 🤔 (not happening in SP)
Relaunching it for Nano S

EDIT: Yes, those were flakiness (speculos)

apps/ledger-live-desktop/tests/page/modal/newSendFlow.page.ts
```diff
@@ -0,0 +1,520 @@
+import { expect } from "@playwright/test";
+import { step } from "tests/misc/reporters/step";
+import { Component } from "tests/page/abstractClasses";
+
+export type StellarMemoType = "NO_MEMO" | "MEMO_TEXT" | "MEMO_ID" | "MEMO_HASH" | "MEMO_RETURN";
+
+type FeePreset = "slow" | "medium" | "fast";
+
+export class NewSendFlowPage extends Component {
```
martijnhjk: Other files in this directory follow a `*.modal.ts` convention. And Pages seem to extend `Page` rather than `Component`. Is this exception intentional?

libs/ledger-live-common/src/families/evm/bridge/mock.ts
```diff
@@ -106,11 +116,21 @@ const getTransactionStatus = (account, transaction) => {
 };
 
 const prepareTransaction = async (_a, t) => {
+  const feesStrategy = (t as unknown as { feesStrategy?: string }).feesStrategy;
```
martijnhjk: What is the benefit of this in line type assertion? In effect we are telling the compiler not to infer but to treat the parameter as a specific type to then read an optional property. Since `t` is `any` there is no guarantee the property exists even with this type assertion.

You could consider optional chain on `t`, eg:
```
const strategy = t?.feesStrategy;
```
Notice the const is named different from the property for readability.

martijnhjk: EDIT: if you want to keep the const as a _typed_ variable of type `string | undefined` you could set it simply like so:
```
const strategy: string | undefined = t?.feesStrategy;
```
This has the same result as the type check but reads as "'strategy' is a constant with a string or undefined value derived from an optional property on t"

libs/ledger-live-common/src/families/evm/bridge/mock.ts
```diff
@@ -106,11 +116,21 @@ const getTransactionStatus = (account, transaction) => {
...
+  const gasPriceByStrategy: Record<string, BigNumber> = {
+    slow: new BigNumber(20000000000), // 20 gwei
+    medium: new BigNumber(30000000000), // 30 gwei
+    fast: new BigNumber(50000000000), // 50 gwei
+  };
+
+  const gasPrice = feesStrategy
+    ? gasPriceByStrategy[feesStrategy] ?? new BigNumber(30000000000)
```
martijnhjk: please be mindful adding nested ternaries and defaults as they add to cognitive load.

martijnhjk: a very different example that produces more lines and looks very verbose is using a switch case with default fall back:
```
const prepareTransaction = async (_a, t) => {
  let gasPrice: BigNumber;

  const strategy = t?.feesStrategy?.toLowerCase();
  switch (true) {
    case strategy === "slow":
      gasPrice = new BigNumber(20000000000); // 20 gwei
      break;
    case strategy === "medium":
      gasPrice = new BigNumber(30000000000); // 30 gwei
      break;
    case strategy === "fast":
      gasPrice = new BigNumber(50000000000); // 50 gwei
      break;
    default:
      gasPrice = new BigNumber(30000000000);
  }
  ...
```
its just an example of a 'simple' and verbose implementation as a way to demonstrate readability as an important consideration in our code. There are of course multiple ways to do this and some personal preference. However, code should always be as readable as possible.

apps/ledger-live-desktop/tests/page/modal/newSendFlow.page.ts
dilaouid: File renamed

libs/ledger-live-common/src/families/evm/bridge/mock.ts
dilaouid: Your choice seems easier to read I agree !
Applied !

libs/ledger-live-common/src/families/evm/bridge/mock.ts
dilaouid: Applied among the below comment
Thank you so much !!

dilaouid:
E2E tests:

- NanoS : https://github.com/LedgerHQ/ledger-live/actions/runs/23232046561
- NanoSP: https://github.com/LedgerHQ/ledger-live/actions/runs/23232062659
- NanoX: https://github.com/LedgerHQ/ledger-live/actions/runs/23232074722
- Stax: https://github.com/LedgerHQ/ledger-live/actions/runs/23232084167
- Flex: https://github.com/LedgerHQ/ledger-live/actions/runs/23232092103
- Gen5: https://github.com/LedgerHQ/ledger-live/actions/runs/23232100353

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  @step("Select address from list")
+  async clickOnSendToButton() {
+    await this.visibleSendToButton.scrollIntoViewIfNeeded();
+    await this.visibleSendToButton.click();
+  }
+
+  @step("Remove recent address tile (index: $0)")
+  async removeRecentAddressTile(index: number = 0) {
```
ypolishchuk-ledger: This method is not in use anywhere. Could you please remove it and check any other code that's not in use? Thank you!

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,435 @@
...
+    const moreActions = this.recentTileMoreActionsButtons.nth(index);
+    await moreActions.click();
+
+    await this.removeRecentMenuItem.click();
+    await this.removeRecentMenuItem.waitFor({ state: "hidden" });
+  }
+
+  @step("Wait for address validation")
+  async waitForRecipientValidation() {
```
ypolishchuk-ledger: I see two issues with this method:
1. Dangling promises. `Promise.any` resolves when the first `waitFor()` succeeds, but the remaining 6 promises keep running in the background. They will eventually time out and produce unhandled rejections: this can cause flaky failures, noisy logs, or `"Target closed"` errors if the page navigates before they settle. The same issue applies to `waitForSignature()`, `waitForSignatureOrConfirmation()`, and the `Promise.any` inside `clickReview()`.
2. Non-deterministic assertion. This method accepts any of 7 outcomes as success (valid address, sanctioned banner, error banner, etc.). That means if you enter a valid address but the sanctioned banner appears due to a regression, the test still passes. A test should always assert the specific expected outcome.

Suggested approach: Remove `waitForRecipientValidation()` and have each call site assert the expected result directly. Playwright's `expect().toBeVisible()` already auto-retries, so no separate "wait" step is needed

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,435 @@
...
+      return;
+    }
+
+    throw new Error("No validation message or error banner appeared");
+  }
+
+  @step("Skip memo")
+  async skipMemo(confirm: boolean = true) {
+    await this.skipMemoLink.click();
+    if (confirm) {
```
ypolishchuk-ledger, line 69: Why do you have `if` here? Does it mean that the confirmation dialog is not always visible? Is it expected?

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,435 @@
...
+      if (isEnabled) {
+        return;
+      }
+
+      await this.page.waitForTimeout(250);
+    }
+  }
+
+  @step("Wait for review button to be enabled")
+  async waitForReviewButtonEnabled(timeout: number = 30000) {
```
ypolishchuk-ledger: This reimplements Playwright's built-in retry with `while` + `waitForTimeout(250)` (discouraged by Playwright docs). `expect(locator).toBeEnabled({ timeout })` does the same thing natively - with better error reporting and no hard-coded sleep. Same issue with `waitForBridgeReady()` (which also appears unused).

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,435 @@
...
+        break;
+      case "Max":
+        button = this.quickActionMax;
+        break;
+    }
+    await button.click();
+  }
+
+  @step("Click review to proceed to signature")
+  async clickReview() {
```
ypolishchuk-ledger, line 91: There are several issues with this method:
1. Hand-rolled polling loop via `waitForReviewButtonEnabled`
2. Manual diagnostics that duplicate Playwright's error reporting
3. Dangling promises from `Promise.any`
4. Mixed responsibilities (it does three things: waits for a precondition, performs an action, and waits for the outcome.)
5. Return-boolean pattern propagates bad error handling

Could you please address these issues? :)

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: Yes this is expected, it lets tests call `skipMemo(false)` so they can assert the confirmation UI is shown without dismissing it yet
Essentially useful for ALGO

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: Resolved !

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: Indeed ! Some remaining artifacts from the old version of the new flow, mb !
Deleted !

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: Rresolved !

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+  "near",
+  "solana",
+  "algorand",
+  "stellar",
+  "xrp",
+  "bitcoin",
+  "kaspa",
+];
+
+const MEMO_FAMILIES = new Set(["algorand", "stellar", "xrp", "solana"]);
```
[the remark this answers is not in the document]
dilaouid: Wrong: this is xrp not ripple for this case

apps/ledger-live-desktop/tests/specs/send/newSendFlow.spec.ts
```diff
@@ -0,0 +1,1224 @@
...
+        // Include bitcoin + kaspa to cover fee presets across families when supported.
+        families: [
+          "tezos",
+          "tron",
+          "evm",
+          "near",
+          "solana",
+          "algorand",
+          "stellar",
+          "xrp",
```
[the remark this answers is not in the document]
dilaouid: Wrong: this is xrp not ripple for this case

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,110 @@
...
+  async clickOnSendToButton() {
+    await this.visibleSendToButton.scrollIntoViewIfNeeded();
+    await this.visibleSendToButton.click();
+  }
+
+  @step("Skip memo")
+  async skipMemo(confirm: boolean = true) {
+    await this.skipMemoLink.click();
+    if (confirm) {
+      await this.page.getByTestId("send-skip-memo-confirm-button").click();
```
ypolishchuk-ledger: could you please move this locator to the top of the page, and also could you please double check that across this class you are calling `dialog.` instead of `page.`, because using `page`  will find the first match in the whole DOM, not necessarily the one within the send flow dialog.

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,110 @@
...
+  @step("Select fee preset: $0")
+  async selectFeePreset(preset: Fee) {
+    await this.openFeesMenu();
+
+    const item = this.page.getByTestId(`send-fees-preset-${preset.toLowerCase()}`);
+    await item.click();
+    await item.waitFor({ state: "hidden" });
+  }
+
+  @step("Wait for signature screen and device action loader")
```
ypolishchuk-ledger: The description says "and" but .or() means it resolves as soon as either element is visible - they are mutually exclusive outcomes. This is misleading in Allure/step reports. Could you please update the step description to reflect that?

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,110 @@
...
+  readonly confirmationStatusContent = this.dialog.locator(
+    '[data-testid="send-confirmation-success-content"], [data-testid="send-confirmation-error-content"], [data-testid="send-confirmation-info-content"]',
+  );
+  readonly successConfirmationTitle = this.dialog.getByTestId("send-confirmation-success-title");
+  readonly deviceActionLoader = this.page.getByTestId("device-action-loader");
+  readonly signaturePrompt = this.page.getByTestId("send-signature-prompt");
+
+  private async isCryptoInputMode() {
+    const secondaryValue = (await this.amountSecondaryValue.textContent())?.trim() ?? "";
+    return /[$€£¥₩₹]|\b(?:USD|EUR|GBP|JPY|CHF|CAD|AUD)\b/.test(secondaryValue);
```
ypolishchuk-ledger: is there a way to avoid hardcoding currencies here? it might be very fragile in the future

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,110 @@
...
+  @step("Open fees menu")
+  async openFeesMenu() {
+    await this.feesMenuTrigger.click();
+  }
+
+  @step("Select fee preset: $0")
+  async selectFeePreset(preset: Fee) {
+    await this.openFeesMenu();
+
+    const item = this.page.getByTestId(`send-fees-preset-${preset.toLowerCase()}`);
```
ypolishchuk-ledger: please move locator to the top of the class

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Fee } from "@ledgerhq/live-common/e2e/enum/Fee";
+import { step } from "tests/misc/reporters/step";
+import { Modal } from "tests/component/modal.component";
+
+export class NewSendModal extends Modal {
+  readonly dialog = this.page.getByRole("dialog");
+
+  readonly recipientInput = this.dialog.getByTestId("send-recipient-input");
+  readonly visibleSendToButton = this.dialog
+    .locator('[data-testid="send-matched-address-button"]:visible')
```
ypolishchuk-ledger: I would suggest making visible check more explicit ` .locator('[data-testid="send-matched-address-button"]').filter({ visible: true }).first()`

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    transaction: new Transaction(Account.KASPA_1, Account.KASPA_2, "0.2"),
+    xrayTicket: "B2CQA-3840",
+  },
+  {
+    transaction: new Transaction(Account.BASE_1, Account.BASE_2, "0.000001"),
+    xrayTicket: "B2CQA-4225",
+  },
+];
+
+const LNS_UNSUPPORTED_CURRENCIES = new Set([Currency.SUI.id, Currency.VET.id, Currency.HBAR.id]);
```
ypolishchuk-ledger: LNS_UNSUPPORTED_CURRENCIES / shouldSkipLNSTag is dead logic in this file. None of SUI, VET, or HBAR appear in transactionsNewSendFlow.

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+          if (!recipientAddress) {
+            throw new Error(
+              `Missing recipient address for ${tx.accountToCredit.accountName}. ` +
+                `Ensure the CLI setup populates the address.`,
+            );
+          }
+          await app.newSendFlow.typeAddress(recipientAddress);
+
+          if (hasMemo) {
+            if (tx.memoTag && tx.memoTag !== "noTag") {
```
ypolishchuk-ledger: This branch is never hit in the reality, could you please check

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+function shouldSkipLNSTag(currencyId: string): boolean {
+  return LNS_UNSUPPORTED_CURRENCIES.has(currencyId);
+}
+
+test.describe("New Send Flow", () => {
+  for (const entry of transactionsNewSendFlow) {
+    const tx = entry.transaction;
+    const family = getFamilyByCurrencyId(tx.accountToDebit.currency.id);
+    const hasMemo = family ? MEMO_FAMILIES.has(family) : false;
```
ypolishchuk-ledger: `hasMemo` variable name is misleading could you please rename it to `requiresMemoStep` or `isMemoFamily`, also could you please move `hasMemo` inside the `async ({ app }) => { ... }` callback.

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+import { Account } from "@ledgerhq/live-common/e2e/enum/Account";
+import { Fee } from "@ledgerhq/live-common/e2e/enum/Fee";
+import { Transaction } from "@ledgerhq/live-common/e2e/models/Transaction";
+import { addTmsLink } from "tests/utils/allureUtils";
+import { getDescription } from "tests/utils/customJsonReporter";
+import { getFamilyByCurrencyId } from "@ledgerhq/live-common/currencies/helpers";
+import { liveDataWithRecipientAddressCommand } from "tests/utils/cliCommandsUtils";
+import { Currency } from "@ledgerhq/live-common/e2e/enum/Currency";
+
+const NEW_SEND_FLOW_FAMILIES = [
```
ypolishchuk-ledger: can you please use Currencies enum to avoid hardcoding strings here?

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
ypolishchuk-ledger: same here

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+  return LNS_UNSUPPORTED_CURRENCIES.has(currencyId);
+}
+
+test.describe("New Send Flow", () => {
+  for (const entry of transactionsNewSendFlow) {
+    const tx = entry.transaction;
+    const family = getFamilyByCurrencyId(tx.accountToDebit.currency.id);
+    const hasMemo = family ? MEMO_FAMILIES.has(family) : false;
+
+    test.describe(`Send via new flow: ${tx.accountToDebit.accountName} → ${tx.accountToCredit.accountName}`, () => {
```
ypolishchuk-ledger: I think the describe level can be simplified to `${tx.accountToDebit.currency.ticker}` or `${tx.accountToDebit.accountName}`, letting the test name carry the full detail.

e2e/desktop/tests/page/modal/new.send.modal.ts
martijnhjk: I think we might be able to use playwright filter here too, eg:
```
readonly visibleSendToButton = this.dialog
    .getByTestId("send-matched-address-button").filter({ visible: true });
```
https://playwright.dev/docs/locators#matching-only-visible-elements

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,110 @@
...
+      .waitFor({ state: "visible", timeout: 120000 });
+  }
+
+  @step("Wait for terminal confirmation screen")
+  async waitForConfirmation() {
+    await this.confirmationStatusContent.waitFor({ state: "visible" });
+  }
+
+  @step("Wait for success confirmation screen")
+  async waitForSuccessConfirmation() {
```
martijnhjk: if the step contains an assertion please consider updating the title and name to contain 'check' or 'expect'.

e2e/desktop/tests/page/modal/new.send.modal.ts
martijnhjk: ah jinx I didn't notice @ypolishchuk-ledger commented the same here https://github.com/LedgerHQ/ledger-live/pull/14523#discussion_r2980577960 😄

apps/ledger-live-desktop/tests/userdata/1AccountBTC1AccountETH1AccountARB1AccountSOL1AccountXTZ1AccountXLM1AccountALGO1AccountXRP.json
```diff
@@ -0,0 +1,530 @@
...
+      {
+        "data": {
+          "id": "mock:1:ripple:rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH:",
+          "seedIdentifier": "mock",
+          "name": "XRP 1",
+          "starred": false,
+          "derivationMode": "",
+          "index": 0,
+          "freshAddress": "rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH",
+          "freshAddressPath": "44'/501'/0'",
```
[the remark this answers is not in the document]
dilaouid: not necessary, this is just mock data

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
+import { expect } from "@playwright/test";
+import { Fee } from "@ledgerhq/live-common/e2e/enum/Fee";
+import { step } from "tests/misc/reporters/step";
+import { Modal } from "tests/component/modal.component";
+
+const FIAT_ISO_CODES = new Set<string>(Intl.supportedValuesOf("currency"));
+
+function looksLikeFiatValue(value: string): boolean {
```
martijnhjk: If used only once, does it make sense to abstract into a function? And if it isn't using the class instance it might be useful as a utils module.

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
...
+      .waitFor({ state: "visible", timeout: 120000 });
+  }
+
+  @step("Wait for terminal confirmation screen")
+  async waitForConfirmation() {
+    await this.confirmationStatusContent.waitFor({ state: "visible" });
+  }
+
+  @step("Check success confirmation screen")
+  async waitForSuccessConfirmation() {
```
martijnhjk, line 121: please consider updating the function name to reflect an assertion rather than a fluent wait.

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,460 @@
...
+    await this.backButton.click();
+    await this.page.waitForTimeout(300);
+  }
+
+  // ========== RECIPIENT STEP METHODS ==========
+
+  @step("Type address in search input: $0")
+  async typeAddress(address: string) {
+    await this.recipientInput.waitFor({ state: "visible" });
+    await this.recipientInput.clear();
```
ypolishchuk-ledger: clear() before fill() is redundant

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,460 @@
...
+      await this.page.waitForTimeout(250);
+    }
+    return false;
+  }
+
+  @step("Fill amount: $0")
+  async fillAmount(amount: string) {
+    await this.amountInput.waitFor({ state: "visible", timeout: 10000 });
+    await this.switchToFiatMode();
+    await this.amountInput.clear();
```
ypolishchuk-ledger: clear() before fill() is redundant

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,460 @@
...
+    await this.amountInput.fill(amount);
+    await this.page.waitForTimeout(800); // Wait for validation
+  }
+
+  @step("Fill crypto amount: $0 (switches to crypto mode first)")
+  async fillCryptoAmount(amount: string) {
+    await this.amountInput.waitFor({ state: "visible", timeout: 10000 });
+
+    await this.switchToCryptoMode();
+    await this.amountInput.clear();
```
ypolishchuk-ledger: clear() before fill() is redundant

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @step("Wait for send flow dialog to open")
+  async waitForDialog() {
+    await this.dialog.waitFor({ state: "visible" });
+  }
+
+  @step("Type address in search input: $0")
+  async typeAddress(address: string) {
+    await this.recipientInput.clear();
```
ypolishchuk-ledger: clear() before fill() is redundant

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @step("Fill crypto amount: $0 (switches to crypto mode first)")
+  async fillCryptoAmount(amount: string) {
+    const isCryptoMode = await this.isCryptoInputMode();
+    if (!isCryptoMode) {
+      await this.toggleInputModeButton.click();
+      await expect.poll(() => this.isCryptoInputMode()).toBe(true);
+    }
+    await this.amountInput.clear();
```
ypolishchuk-ledger: clear() before fill() is redundant

e2e/desktop/tests/page/modal/new.send.modal.ts
ypolishchuk-ledger: I agree

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
...
+  );
+  readonly successConfirmationTitle = this.dialog.getByTestId("send-confirmation-success-title");
+  readonly deviceActionLoader = this.page.getByTestId("device-action-loader");
+  readonly signaturePrompt = this.dialog.getByTestId("send-signature-prompt");
+
+  getFeePreset(preset: Fee) {
+    return this.page.getByTestId(`send-fees-preset-${preset.toLowerCase()}`);
+  }
+
+  private async isCryptoInputMode() {
```
ypolishchuk-ledger: `textContent()` does not wait for the element to contain a non-empty value. If the component hasn't rendered the secondary value yet, this returns "", `looksLikeFiatValue("")` returns `false`, and the toggle is incorrectly triggered. The UI can then briefly be in an intermediate state before `expect.poll` catches up. 

Recommendation: Wait for the element to have non-empty content before reading it, e.g. await `expect(this.amountSecondaryValue).not.toBeEmpty()` before the `textContent()` call.

e2e/desktop/tests/page/modal/new.send.modal.ts
ypolishchuk-ledger: Ok, in this case, I would recommend using an options object: `skipMemo({ confirm = true } = {})`, or split into `skipMemo()` and `skipMemoWithoutConfirm()`.

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
...
+    const item = this.getFeePreset(preset);
+    await item.click();
+    await item.waitFor({ state: "hidden" });
+  }
+
+  @step("Wait for signature screen or device action loader")
+  async waitForSignature() {
+    await this.deviceActionLoader
+      .or(this.signaturePrompt)
+      .waitFor({ state: "visible", timeout: 120000 });
```
ypolishchuk-ledger: I would suggest removing hardcoded timeouts in this file and rely on default timeout we set in playwright.config

e2e/desktop/tests/page/modal/new.send.modal.ts
```diff
@@ -0,0 +1,127 @@
...
+  readonly memoInput = this.dialog.getByTestId("send-memo-input");
+  readonly skipMemoLink = this.dialog.getByTestId("send-skip-memo-link");
+  readonly skipMemoConfirmButton = this.dialog.getByTestId("send-skip-memo-confirm-button");
+  readonly amountInput = this.dialog.getByTestId("send-amount-input");
+  readonly feesMenuTrigger = this.dialog.getByTestId("send-network-fees-menu-trigger");
+  readonly reviewButton = this.dialog.getByTestId("send-review-button");
+  readonly toggleInputModeButton = this.dialog.getByTestId("send-toggle-input-mode-button");
+  readonly amountSecondaryValue = this.dialog.getByTestId("send-amount-secondary-value");
+  readonly confirmationStatusContent = this.dialog.locator(
+    '[data-testid="send-confirmation-success-content"], [data-testid="send-confirmation-error-content"], [data-testid="send-confirmation-info-content"]',
```
ypolishchuk-ledger: The raw CSS multi-selector string is not type-safe and breaks IDE navigation/refactoring. Additionally, this locator matches any of the three states, but the name `confirmationStatusContent` is generic - `waitForConfirmation` doesn't distinguish success from error.                                                   
                                                                                                                                                                                                                                                                                                                       
Recommendation: Either expose three typed locators and combine with `.or()`, or at minimum document that this locator intentionally covers all terminal states (and that `waitForSuccessConfirmation` is the way to assert success specifically).

dilaouid:
e2e nanox:https://github.com/LedgerHQ/ledger-live/actions/runs/23578528268
e2e flex: https://github.com/LedgerHQ/ledger-live/actions/runs/23578534683

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: done

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: done

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
dilaouid: done

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
dilaouid: done

apps/ledger-live-desktop/tests/page/modal/new.send.modal.ts
dilaouid: done

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: done

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: fixed

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: done used the default timeout

e2e/desktop/tests/page/modal/new.send.modal.ts
dilaouid: Done 👍

e2e/desktop/tests/specs/newSendFlow.tx.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+            ...(family ? [`@family-${family}`] : []),
+          ],
+          annotation: { type: "TMS", description: entry.xrayTicket },
+        },
+        async ({ app }) => {
+          const requiresMemoStep = family ? MEMO_STEP_FAMILIES.has(family) : false;
+
+          await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+
+          await app.layout.goToAccounts();
```
martijnhjk: please consider using `await app.mainNavigation.openTargetFromMainNavigation("accounts");` to make sure it works for legacy wallet and wallet 4.0

dilaouid: Changed !

PR #14537, opened 2026-02-17
[QAA-976] Use remote speculos for LWD tests to eliminate Network issue

ypolishchuk-ledger:
## Summary

This PR improves the reliability of E2E tests using remote Speculos by fixing CLI transport configuration, adding retry logic for transient failures, and splitting large test files for better CI distribution.

Before:
Screenshot 2026-02-17 at 15 50 48

After:
Screenshot 2026-02-17 at 15 51 20

### 1. Fixed CLI Speculos Transport Configuration
**File:** `apps/cli/src/live-common-setup.ts`

- Fixed `registerSpeculosTransport` to explicitly read `process.env.SPECULOS_ADDRESS` and pass it as `baseURL`
- Changed type from `Record<string, number>` to `SpeculosHttpTransportOpts` for proper typing
- Converted `apiPort` to string to match expected type

### 2. Added Retry Logic for CLI Commands
**File:** `e2e/desktop/tests/utils/runCli.ts`

- Added `runCliCommandWithRetry` function with configurable retries (default: 3) and delay (default: 3 seconds)
- Added `isRetryableError` helper to detect transient errors (503, 502, 504, GeneralDmkError, ECONNREFUSED, ETIMEDOUT, etc.)
- Added `sleep` and `extractFlagValue` helper functions

### 3. Updated CLI Utilities to Use Retry Logic
**File:** `e2e/desktop/tests/utils/cliUtils.ts`

- Updated `CLI.liveData()` to use `runCliCommandWithRetry`
- Updated `CLI.getAddress()` to use `runCliCommandWithRetry`

### 4. Split Large Swap Test File for Better CI Distribution

**Deleted:** `e2e/desktop/tests/specs/swap.other.spec.ts` (1498 lines)

**Created:**

| File | Description |
|------|-------------|
| `accounts.swap.spec.ts` | Account selection, default currency, rejected on device, different seed, no account tests |
| `entrypoint.swap.spec.ts` | Entry points (Portfolio, Market, Account page), Send Max, History, Blacklist tests |
| `provider.swap.spec.ts` | Provider redirection, 1inch flow, Best Offer, Landing page tests |
| `validation.swap.spec.ts` | Error messages, too low amounts, network fees validation tests |

**Renamed:** `swap.send.spec.ts` → `send.swap.spec.ts`

## Expected Improvements

- **Fixed:** `GeneralDmkError` when running CLI commands with remote Speculos
- **Fixed:** Transient 503 errors causing test failures (now retried automatically)
- **Improved:** CI run time reduced from ~20 minutes to ~10-12 minutes by distributing swap tests across multiple shards


e2e/desktop/tests/fixtures/common.ts
```diff
@@ -157,7 +159,8 @@ export const test = base.extend<TestFixtures>({
         speculos = await launchSpeculos(speculosApp.name, testInfo.title);
 
         if (cliCommands?.length) {
-          CLI.registerSpeculosTransport(speculos.port.toString());
+          const speculosAddress = process.env.SPECULOS_ADDRESS || "http://localhost";
```
martijnhjk: seems like we are setting this twice, would it be more maintainable to set it once as a const higher up and reuse?

e2e/desktop/tests/specs/provider.swap.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+
+test.describe("Swap - Landing page", () => {
+  const fromAccount = Account.ETH_1;
+  const toAccount = TokenAccount.ETH_USDC_1;
+
+  setupEnv(true);
+
+  test.beforeEach(async () => {
+    const accountPair: string[] = [fromAccount, toAccount].map(acc =>
+      acc.currency.speculosApp.name.replace(/ /g, "_"),
```
martijnhjk, line 212: Sonar suggests using replaceAll() for global replace.

e2e/desktop/tests/utils/allureUtils.ts
```diff
@@ -9,6 +9,20 @@ import { isLastRetry } from "tests/utils/testInfoUtils";
...
+  testInfo: TestInfo,
+  name: string,
+  filePath: string,
+  contentType: string,
+) {
+  try {
+    const data = await readFileAsync(filePath);
+    await testInfo.attach(name, { body: data, contentType });
+  } catch {
+    // File does not exist → silently ignore
```
martijnhjk, line 22: should we at least log something for traceability / visibility?

ypolishchuk-ledger: what would be the purpose of it? Because if there are no network errors or other logs, there is nothing to log, hence no file to attach, so I'm not sure why we would need a log here.

e2e/desktop/tests/specs/provider.swap.spec.ts
ypolishchuk-ledger: I didn't do new changes in this file, I just split the existing big swap spec file into several smaller ones, so if you have ideas for improvements, please create a dedicated Jira ticket :)

e2e/desktop/tests/fixtures/common.ts
ypolishchuk-ledger: done

hedi-edelbloute (approved):

e2e/desktop/tests/specs/provider.swap.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+          app: toAccount.currency.speculosApp,
+          cmd: liveDataWithAddressCommand(toAccount),
+        },
+      ],
+      { scope: "test" },
+    ],
+  });
+
+  test(
+    `Swap landing page`,
```
francois-guerin-ledger, line 237:
```suggestion
    "Swap landing page",
```
I see multiple of these patterns. There is no prettier / lint rule to polish / warn on them ?

ypolishchuk-ledger: Hey, I didn't change those files in my PR I just split the big swap spec into smaller ones. If you want, we can address your comment in a separate PR, would it be acceptable?

VicAlbr (approved):

e2e/desktop/tests/utils/allureUtils.ts
martijnhjk: Could there be an error for other reasons, eg failing to attach? Anyway, if you think a log would not add value here thats fair enough!

e2e/desktop/tests/specs/provider.swap.spec.ts
martijnhjk: @francois-guerin-ledger good point about the prettier / linter. I will add a ticket int he backlog to investigate.

PR #14583, opened 2026-02-18
feat(LIVE-26235): add onboarding has device state to desktop settings

cfloume:
### ✅ Checklist

  - Desktop settings now include `onboardingHasDevice` as a strict `boolean` (default `false`).
  - Added Redux action `SET_ONBOARDING_HAS_DEVICE` and action creator `setOnboardingHasDevice`.
  - Added selector `onboardingHasDeviceSelector` for lazy onboarding UI flow decisions.
  - Value persists through existing desktop settings persistence (`SAVE_SETTINGS` / `FETCH_SETTINGS`) without DB middleware changes.

### 📝 Description

This PR implements LIVE-26235 by adding desktop support for lazy onboarding device state.

A new `onboardingHasDevice` field is added to desktop `SettingsState` with a default value of `false`, plus:
- a dedicated action creator to update
- a reducer case to handle updates,
- and a selector to read it from the store.

The implementation follows existing desktop settings patterns and uses the current settings persistence pipeline so the value survives app restarts.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-26235](https://ledgerhq.atlassian.net/browse/LIVE-26235)

---

- **Performance** considerations have been taken into account. (s have been profiled or benchmarked if necessary)

[LIVE-26235]: https://ledgerhq.atlassian.net/browse/LIVE-26235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/reducers/settings.ts
+++ b/apps/ledger-live-desktop/src/renderer/reducers/settings.ts
@@ -127,6 +127,7 @@ export type SettingsState = {
   alwaysShowMemoTagInfo: boolean;
   anonymousUserNotifications: { LNSUpsell?: number } & Record<string, number>;
   hasSeenWalletV4Tour: boolean;
+  onboardingHasDevice: boolean;
   doNotAskAgainSkipMemo: boolean;
 };
 
@@ -227,6 +228,7 @@ export const INITIAL_STATE: SettingsState = {
   alwaysShowMemoTagInfo: true,
   anonymousUserNotifications: {},
   hasSeenWalletV4Tour: false,
+  onboardingHasDevice: false,
   doNotAskAgainSkipMemo: false,
 };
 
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/renderer/reducers/settings.ts
```diff
@@ -227,6 +228,7 @@ export const INITIAL_STATE: SettingsState = {
   alwaysShowMemoTagInfo: true,
   anonymousUserNotifications: {},
   hasSeenWalletV4Tour: false,
+  onboardingHasDevice: false,
```
gre-ledger, line 231: so it will be false for existing users? is that expected?
what makes us need to persist it for users then, if this doesn't reflect the past reality 🤔 how can we trust this value

also I do think we already have this information: as soon as we see a device, we will save it, so i would advocate for not introducing new state but just trust

`lastSeenDeviceSelector` for this, wdyt?

PR #14592, opened 2026-02-19
ci: stop making SonarQube failing PRs

gre-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Problem: most PRs (80%+) appears as ❌  due to SonarQube failing and our too strict conditions we have on SonarQube rules.

The current status quo is that we don't make SonarQube PRs blocking the PR merge, which means it's only use in an informational way. So **on the short term** we should also make the PR displayed as so and keeping SonarQube informal via a comment.

**on the long term**, we should rediscuss and redefine which part of sonarqube is the mandatory bit, but at the moment it has way too unrealistic goals.

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/.github/workflows/scan-sonar-reusable.yml
+++ b/.github/workflows/scan-sonar-reusable.yml
@@ -24,6 +24,7 @@ jobs:
   scan-sonar:
     name: "Sonar Cloud"
     runs-on: [ubuntu-22.04]
+    continue-on-error: true # keep PRs green until we make SonarQube failing case more constructive
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:


.github/workflows/scan-sonar-reusable.yml
```diff
@@ -24,6 +24,7 @@ jobs:
   scan-sonar:
     name: "Sonar Cloud"
     runs-on: [ubuntu-22.04]
+    continue-on-error: true # keep PRs green until we make SonarQube failing case more constructive
```
martijnhjk, line 27: I dont think this is what you necessarily want since the build will always pass. It is the reported 'quality gate' that passes or fails and is reported asynchronously separate from the build job. CC @lewisd5 @LL782

PR #14633, opened 2026-02-20
doc: introduce coin families ui specifics vs generic AI rules

gre-ledger:
### ✅ Checklist

  - ...

### 📝 Description

I see myself having this recurrent feedback on many ongoing PR and I notice there is no rules yet defined that clarify how the coin specifics are supposed to land in families/ and that generic components MUST NOT have coin specifics. 🙏 

### ❓ Context

- **JIRA or GitHub link**: 

---

### DIFF
--- a/.cursor/agents/code-reviewer.md
+++ b/.cursor/agents/code-reviewer.md
@@ -12,6 +12,7 @@ Follow all project rules in `.cursor/rules/`. Pay special attention to:
 - `react-mvvm.mdc` — MVVM compliance is a key review concern for code in `src/mvvm/`
 - `ldls-web.mdc` — Lumen UI for desktop (`src/mvvm/`)
 - `ldls-native.mdc` — Lumen UI for mobile (`src/mvvm/`)
+- `coin-families-contract.mdc` — No coin-specific branches (`if (family === "evm")` etc.) in generic UI; extend the families contract and implement in `families/<family>/` instead
 
 ## Review Scope
 
@@ -31,6 +32,7 @@ By default, review unstaged changes from `git diff`. The user may specify differ
 - Sonar issues: complexity, duplication, security hotspots
 - Missing integration tests for new features under `src/mvvm/` (required by `react-mvvm.mdc`)
 - Lumen UI compliance: verify new UI in `src/mvvm/` uses design-system components
+- **Coin-families contract:** In generic code (outside `families/<family>/`), flag new `if (family === "…")` or coin-specific hooks; suggest extending the families contract and implementing in the family folder instead (see `coin-families-contract.mdc`)
 
 ## Confidence Scoring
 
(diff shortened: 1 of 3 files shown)


dilaouid:
Nice rules !
Just to precision, wdyt of this [descriptor file](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/bridge/descriptor.ts)?
(just noticed it's a bit of a mess atm)

Actually only used for the new send flow atm but is made to be extended for all the incoming flows (receive, swap, staking...)
Defines some properties / fn that is generic and can be used in LWD / LWM by simply passing the family name in parameters

ADR ref: https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/6540492802/Coin+modules+-+ADR-012+-+Feature+descriptor

Moustafa-Koterba (approved):

gre-ledger:
@dilaouid to me it's legit, always better than hardcoding ifs in generic parts :) 

It's a good idea even tho on the long term, I think we still haven't figured out a way to really decouple coin libs from end apps.

The day, I can take a coin impl individually and glue all together "à la carte" to make a "Desktop app", it will be great.

we'll have to see how to reach this goal, and reduce as much friction as possible. There were a few ideas discussed with archi over the past months, but i would dream like having some sort of "Coin Modules Manager" that is instanciate at the app level (on desktop, on mobile, etc..) that would be used to `.add` coin implementations together & maybe dynamically too (lazy loaded).

(one manager of multiple manager maybe, because we could distinct coin implementations logic and coin ui implementations,.., but the essence of this idea is that we should only have one location where we glue the things together and eventually get rid of any entry point that do the multiplexing or generated things)

```typescript
## coin-setups.ts

export const manager = new CoinGlueManager(); // exported because from external, we may need to use manager to access the implems (basically do the `.getAccountBridge` equivalent)

manager.add({
  handles: ["bitcoin", "litecoin",.. ], // idk if we even explicitly need. (today it's more like a "family" actually, we need a way to be able to implement the dispatch)
  logic: await require("coin-bitcoin")({ ...any dependencies it needs... }), // provide the basic coin implem
  ui: await require("coin-bitcoin-ui")(...), // the families/ code, basically the UI specific components for desktop
  signer: ...take from the DMK?... // or maybe it's a dependencies of the logic ? idk
  // ... any other usecase of needed to have a coin resolution, we can extend here. that way, we never have to maintain a dispatch anywhere else
})
```

it's just a random idea, don't take the code literally, it's pseudo code and pseudo spec, but you get the idea: essentially we could only have one location where we "setup" this on each app, and that replaces all the "set supported currencies" too, it could be a one in replacement to both defining the supported coins as well as defining the actual implementation of them. + an open door to allowing dynamic loading (i intentionally made them require() async, actually maybe it's even more `logic: () => await require("...)` on the idea that we could only call it when it's needed, so the bundle becomes super light by design)

PR #14662, opened 2026-02-20
feat(LIVE-23006): add buy navigation hook, support fromCurrency/toCurrency swap deeplinks, and fix cold-start deep-linking

ooke-ledger:
### ✅ Checklist

  - ...

### 📝 Description

- Buy navigation hook: Adds a reusable `useBuyNavigation` hook on desktop and `useOpenBuy` on mobile that resolves accounts for a given currency before navigating to the buy flow - handling zero, single, and multi-account scenarios with an account selection drawer. Replaces the previous static route-based approach in quick actions.
 
- Swap deeplink `fromCurrency`/`toCurrency` support: Extends swap deep-linking on both desktop and mobile to accept `fromCurrency` and `toCurrency` query params (currency IDs) alongside the existing `fromToken`/`toToken` (token IDs). Updates the swap webview to conditionally pass only defined params, avoiding empty values.
 
- Cold-start deep-link reliability: Extracts a reusable `sendDeepLink` helper in the main process that defers `webContents.send("deep-linking", url)` until `did-finish-load` when the renderer is still loading. Applied to all 5 deep-link entry points.

-- 

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-23006

---

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

### DIFF
--- a/apps/ledger-live-desktop/src/main/index.ts
+++ b/apps/ledger-live-desktop/src/main/index.ts
@@ -65,9 +65,7 @@ if (!gotLock) {
           SUPPORTED_SCHEMES.some(scheme => arg.startsWith(`${scheme}://`)),
         );
         if (uri.length) {
-          if ("send" in w.webContents) {
-            w.webContents.send("deep-linking", uri[0]);
-          }
+          sendDeepLink(w, uri[0]);
         }
       }
     }
@@ -87,16 +85,7 @@ app.on("will-finish-launching", () => {
       .then(w => {
         if (w) {
           show(w);
-          const sendDeepLink = () => {
-            if ("send" in w.webContents) {
-              w.webContents.send("deep-linking", url);
-            }
-          };
-          if (w.webContents.isLoading()) {
-            w.webContents.once("did-finish-load", sendDeepLink);
-          } else {
-            sendDeepLink();
-          }
+          sendDeepLink(w, url);
         }
       })
       .catch((err: unknown) => console.log(err));
@@ -241,9 +230,7 @@ app.on("ready", async () => {
           .then(w => {
             if (w) {
               show(w);
-              if ("send" in w.webContents) {
-                w.webContents.send("deep-linking", url);
-              }
+              sendDeepLink(w, url);
             }
           })
           .catch((err: unknown) => console.log(err));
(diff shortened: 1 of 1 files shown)


apps/ledger-live-desktop/src/main/index.ts
```diff
@@ -87,8 +87,15 @@ app.on("will-finish-launching", () => {
       .then(w => {
         if (w) {
           show(w);
-          if ("send" in w.webContents) {
-            w.webContents.send("deep-linking", url);
+          const sendDeepLink = () => {
```
mcayuelas-ledger: why do we need this change?

ooke-ledger: On cold start from a web deep link, we were sending the IPC before the renderer had loaded, so the message was dropped and the swap never opened. We now wait for `did-finish-load` so the renderer can receive it and the deep link works on first launch

Justkant: I see that you only updated it here but we do call `deep-linking` on multiple places in this file, could you check if it makes sense to make a reusable fonction to use on all places we send `deep-linking` ?

gre-ledger: if possible, this should be send into a dedicated PR as it seems to fix a bug? (on top of possibly addressing @Justkant 's feedback)
thanks

the PR description isn't filled also, so please either add more detail OR put the PR in draft 🙏

gre-ledger (approved):

apps/ledger-live-mobile/src/hooks/useQuickActions.ts
```diff
@@ -136,13 +138,7 @@ function useQuickActions({ currency, accounts }: QuickActionProps = {}) {
     if (canBeBought) {
       list.BUY = {
         disabled: isPtxServiceCtaExchangeDrawerDisabled || readOnlyModeEnabled,
-        route: [
-          NavigatorName.Exchange,
-          {
-            screen: ScreenName.ExchangeBuy,
```
mcayuelas-ledger, line 142: if not used anymore can delete right?

ooke-ledger: Yes, this has already been replaced with the custom handler.

PR #14681, opened 2026-02-23
fix: stop race condition on skip onboarding

cfloume:
### ✅ Checklist

### 📝 Description

* There existed a race condition between an effect hook supposed to safeguard the user from seeing the onboarding if they had already completed it AND navigation to the settings page after manually setting `hasCompletedOnboarding`.

---

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Onboarding/screens/WelcomeNew/hooks/useWelcomeNewViewModel.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/Onboarding/screens/WelcomeNew/hooks/useWelcomeNewViewModel.ts
@@ -35,13 +35,14 @@ export function useWelcomeNewViewModel() {
   // Redux selectors
   const hasCompletedOnboarding = useSelector(hasCompletedOnboardingSelector);
   const trustchain = useSelector(trustchainSelector);
+  const shouldInitiallySelectDevice = useRef(hasCompletedOnboarding && !trustchain);
 
   // Navigation effect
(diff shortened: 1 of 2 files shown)


apps/ledger-live-desktop/src/mvvm/features/Onboarding/screens/WelcomeNew/hooks/useWelcomeNewViewModel.ts
```diff
@@ -41,7 +41,8 @@ export function useWelcomeNewViewModel() {
     if (hasCompletedOnboarding && !trustchain) {
       navigate("/onboarding/select-device");
     }
-  }, [hasCompletedOnboarding, navigate, trustchain]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
gre-ledger: - navigate is still legit as a dep
- the condition `hasCompletedOnboarding && !trustchain` should probably just goes into a static ref and that ref is used in the effect

that way you don't need to escape the lint rule 🙏 

```ts
const shouldInitiallySelectDevice = useRef(hasCompletedOnboarding && !trustchain);

useEffect(() => {
    if (shouldInitiallySelectDevice.current) {
      navigate("/onboarding/select-device");
    }
  }, [navigate]);
```

cfloume: Solid suggestion, will change 👍

PR #14795, opened 2026-02-25
refactor(desktop,mobile): rename store files to consistent state-manager/configureStore

LL782:
### ✅ Checklist

- [x] **Covered by automatic tests.** _This is a pure file rename/import path refactor with no behavioral changes. Existing tests validate the store still initializes correctly._
      - Rename to Desktop app Redux store configureation
      - Rename of Mobile app Redux store configuration
      - All import paths referencing the renamed files

### 📝 Description

Both `ledger-live-desktop` and `ledger-live-mobile` had inconsistent naming for their Redux store configuration files — desktop used `renderer/createStore.ts` while mobile used `context/store.ts`.

This PR renames both to a consistent `state-manager/configureStore.ts` convention, aligning the project structure and making it easier to navigate between the two apps. All import paths across 14 files have been updated accordingly. No runtime behavior has changed.

https://ledgerhq.atlassian.net/browse/LIVE-27032

### ❓ Context

- No associated ticket — this is a housekeeping refactor to improve codebase consistency.
- Changes based on [[Guidelines] Data Layer Structure](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6111232117/Guideline+Data+Layer+Data+Layer+Structure+Flow)

---

### DIFF
--- a/apps/ledger-live-mobile/.eslintrc.js
+++ b/apps/ledger-live-mobile/.eslintrc.js
@@ -139,19 +139,6 @@ module.exports = {
   },
   overrides: [
     {
-      // Allow direct react-redux imports in hooks.ts/store.ts where typed hooks are defined
-      files: ["src/context/hooks.ts", "src/context/store.ts", "src/context/selectors.ts"],
-      rules: {
-        "no-restricted-imports": [
-          "error",
-          {
-            patterns: commonImportRestrictions,
-            paths: lodashImportRestriction,
-          },
-        ],
-      },
-    },
-    {
       files: [
         "src/**/*.test.{ts,tsx}",
         "src/screens/Settings/Debug/**/*",
--- a/apps/ledger-live-mobile/src/context/selectors.ts
+++ b/apps/ledger-live-mobile/src/context/selectors.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-restricted-imports */
 import {
   createStructuredSelector as createStructuredSelectorBase,
   createSelector as createSelectorBase,
--- a/.changeset/consistent-configure-store-files.md
+++ b/.changeset/consistent-configure-store-files.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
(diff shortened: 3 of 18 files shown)


deepyjr:
Hey ! I think the comments of copilot are good. Could you do it please? Also i think the pipeline is red for desktop and mobile. 

Otherwise, it's good for me! 🙏

apps/ledger-live-mobile/src/context/selectors.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-restricted-imports */
```
mcayuelas-ledger, line 1: why do we need that now?

apps/ledger-live-mobile/.eslintrc.js
```diff
@@ -138,19 +138,6 @@ module.exports = {
     },
   },
   overrides: [
-    {
```
mcayuelas-ledger, line 141: not sure we want to remove the rule for everything no? we should adapt it I think

apps/ledger-live-mobile/src/context/selectors.ts
LL782: Because it's removed from the `eslintrc`

apps/ledger-live-mobile/.eslintrc.js
LL782: we're not removing the rule but removing a rule-override. 

previously we applied the override to three files but actually it's only needed in two files. This PR drew my attention to that via a Copilot comment.

I had to go into the files to understand what the override was for and decided it would be easier reading in future to read the override as a line comment in the two files where it's needed ("src/context/hooks.ts" and "src/context/selectors.ts")

WDYT?

apps/ledger-live-mobile/src/context/selectors.ts
LL782: I've explained better in https://github.com/LedgerHQ/ledger-live/pull/14795#discussion_r2859833331

PR #14809, opened 2026-02-26
test: add e2e desktop setup for earn 2.0

sergiubreban:
### ✅ Checklist

  - ...

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: 

---


e2e/desktop/tests/page/earn.dashboard.page.ts
```diff
@@ -20,6 +20,15 @@ export class EarnPage extends WebViewAppPage {
   private tabAssetsButton = "tab-assets";
   private learnMoreButton = (currency: string) => `get-${currency}-button`;
 
+  // V2 selectors
```
VicAlbr: Is the V2 completely different from the V1? if yes maybe better to create a new file eg: `earn.v2.dashboard.page.ts`
This will allow us to just delete the current file when V1 is dropped

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+  test.describe("Ice cold start", () => {
+    const account = Account.ETH_1;
+
+    test.use({
+      userdata: "skip-onboarding",
+      speculosApp: account.currency.speculosApp,
+      featureFlags: EARN_V2_DESKTOP_FLAGS,
+    });
+
+    const xrayTicket: string | undefined = undefined; // replace with real xrayTicket
```
VicAlbr: can you add a ticket here 🙏

VicAlbr: same for all the file

e2e/desktop/tests/utils/earnMockData.ts
```diff
@@ -0,0 +1,182 @@
+import { ElectronApplication, Page } from "@playwright/test";
```
VicAlbr: Why do we need mocks ?

e2e/desktop/tests/page/earn.dashboard.page.ts
```diff
@@ -151,4 +160,93 @@ export class EarnPage extends WebViewAppPage {
...
+  async verifyIceColdStartPage() {
+    await this.verifyElementIsVisible(this.footerDisclaimer);
+    await this.verifyElementIsNotVisible(this.maxPotentialRewards);
+    await this.verifyElementIsNotVisible(this.walletHeaderAmount);
+  }
+
+  @step("Click ice cold start earn CTA")
+  async clickIceColdStartEarnCTA() {
+    const webview = await this.getWebView();
+    await webview.getByRole("button", { name: /earn/i }).first().click();
```
ypolishchuk-ledger: could you please move the element locator to the top of the class? and in all other places as well across this file

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+  test.afterAll(async () => {
+    if (originalBroadcastValue === undefined) {
+      delete process.env.DISABLE_TRANSACTION_BROADCAST;
+    } else {
+      process.env.DISABLE_TRANSACTION_BROADCAST = originalBroadcastValue;
+    }
+  });
+}
+
+async function selectAccountInModularSelector(app: any, page: any, account: Account) {
```
ypolishchuk-ledger: I think this method should belong to corresponding page class

martijnhjk:
@sergiubreban please resolve conflicts 🤞

e2e/desktop/tests/page/earn.dashboard.page.ts
martijnhjk: This will click the first button matching, is this stable? Can we make it more unique / specific?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+
+test.describe("Earn [v2]", () => {
+  setupEnv(true);
+  test.use({
+    localManifestOverride: useLocalEarnManifest ? [EARN_LOCAL_MANIFEST] : undefined,
+  });
+
+  // --- User States ---
+
+  test.describe("Ice cold start", () => {
```
ypolishchuk-ledger, line 47: Why do we need this test if "Inline Add Account" already checks everything this test does?

e2e/desktop/tests/page/earn.dashboard.page.ts
```diff
@@ -151,4 +160,93 @@ export class EarnPage extends WebViewAppPage {
...
+
+  @step("Verify rewards summary boxes")
+  async verifyRewardsSummaryBoxes() {
+    await this.verifyElementIsVisible(this.rewardsSummary);
+  }
+
+  @step("Verify position row present: $0")
+  async verifyPositionRowPresent(identifier: string) {
+    const webview = await this.getWebView();
+    const row = webview.getByTestId(/^deposit-row-/).filter({ hasText: identifier });
```
martijnhjk: can we be specific with the locator, why can we only use the prefix "deposit-row-"?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+        ...(xrayTicket ? { annotation: { type: "TMS", description: xrayTicket } } : {}),
+      },
+      async ({ app }) => {
+        if (xrayTicket)
+          await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+        await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+        await app.earnDashboard.verifyIceColdStartPage();
+        await app.earnDashboard.clickIceColdStartEarnCTA();
+        const assetSelector = await getModularSelector(app, "ASSET");
+        if (assetSelector) {
```
ypolishchuk-ledger: is there a specific reason to have so many "if" and "else" here if the flow is pretty deterministic for ETH currency?

e2e/desktop/tests/page/earn.dashboard.page.ts
```diff
@@ -151,4 +160,93 @@ export class EarnPage extends WebViewAppPage {
...
+    await expect(row.first()).toBeVisible();
+  }
+
+  @step("Click position row: $0")
+  async clickPositionRow(identifier: string) {
+    const webview = await this.getWebView();
+    const row = webview.getByTestId(/^deposit-row-/).filter({ hasText: identifier });
```
martijnhjk: This pattern is repeated `/^deposit-row-/`. Is there any way we can make it more specific and reusable (eg declare as a property higher up)?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+        await app.addAccount.close();
+        await app.layout.goToAccounts();
+        await app.accounts.expectAccountsCountToBeNotNull();
+      },
+    );
+  });
+
+  // --- Navigation: CTA Flows ---
+
+  test.describe("CTA → Native staking (SOL)", () => {
```
ypolishchuk-ledger, line 185: what is the difference between this test and the one on line 101? because from what I can see it does exactly the same

e2e/desktop/tests/page/earn.dashboard.page.ts
```diff
@@ -151,4 +160,93 @@ export class EarnPage extends WebViewAppPage {
...
+    await row.first().click();
+  }
+
+  @step("Verify navigated away from earn dashboard")
+  async verifyNavigatedFromDashboard() {
+    await this.verifyElementIsNotVisible(this.tokensToEarnBanner);
+    await this.verifyElementIsNotVisible(this.maxPotentialRewards);
```
martijnhjk: It is generally recommended to have at least 1 positive assertion in the step. For example, if the mentioned elements are no longer visible but the application would show an error or the wrong page the test might not pick it up giving a false pass.

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+        await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+        await app.earnDashboard.clickAssetEarnCta(account.currency.ticker);
+        await selectAccountInModularSelector(app, page, account);
+        // SOL uses native staking: LLD opens the staking modal
+        await expect(page.getByTestId("modal-container")).toBeVisible();
+      },
+    );
+  });
+
+  test.describe("CTA → Partner dapp (ETH)", () => {
```
ypolishchuk-ledger: same question as for previous test

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+        ...(xrayTicket ? { annotation: { type: "TMS", description: xrayTicket } } : {}),
+      },
+      async ({ app, page }) => {
+        if (xrayTicket)
+          await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+        await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+        await app.earnDashboard.clickAssetEarnCta("USDT");
+        await selectAccountInModularSelector(app, page, account);
+        // USDT uses the earn native deposit flow
+        await app.earnDashboard.verifyDepositFlowVisible();
```
ypolishchuk-ledger: would it make sense to check more things than just the URL?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+
+async function selectAccountInModularSelector(app: any, page: any, account: Account) {
+  const selector = await getModularSelector(app, "ACCOUNT");
+  if (selector) {
+    await selector.selectAccount(account);
+    await expect(
+      page
+        .getByTestId("modular-drawer-screen-ACCOUNT_SELECTION")
+        .or(page.getByTestId("modular-dialog-screen-ACCOUNT_SELECTION")),
+    ).toBeHidden();
```
martijnhjk: Can we move this to a central module or page object? Storing functions and logic on spec files can make them harder to maintain.

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+          await verifyProviderUrlPromise;
+        },
+      );
+    });
+  }
+
+  // --- Navigation: Position Row Flows ---
+
+  const positionAccountCurrencies = [
+    { account: Account.NEAR_1, xrayTicket: undefined as string | undefined }, // replace with real xrayTicket
```
ypolishchuk-ledger: Does it make sense to check it for 2 currencies? Could we say that navigation works fine if we check only with one currency?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+      async ({ app }) => {
+        if (xrayTicket)
+          await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+        await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+        await app.earnDashboard.verifyIceColdStartPage();
+        await app.earnDashboard.clickIceColdStartEarnCTA();
+        const selector = await getModularSelector(app, "ASSET");
+        if (selector) {
+          await selector.validateItems();
+        }
```
martijnhjk: If selector is falsy we will not finish the test with an assertion. Can we make an alternate validation here?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+          ...(xrayTicket ? { annotation: { type: "TMS", description: xrayTicket } } : {}),
+        },
+        async ({ app }) => {
+          if (xrayTicket)
+            await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+          await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+          await app.earnDashboard.verifyHotStartPage();
+          await app.earnDashboard.clickPositionRow(account.currency.ticker);
+          await app.account.waitForAccountHeaderName(account.accountName);
+        },
```
martijnhjk, line 270: please consider finishing the test with a validation

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+        await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+        await interceptReady;
+
+        await app.earnDashboard.verifyHotStartPage();
+        await app.earnDashboard.verifyPositionRowPresent(account.currency.ticker);
+        await app.earnDashboard.clickPositionRow(account.currency.ticker);
+        // After clicking, LLD navigates to the partner dapp via custom.navigate → redirect-provider.
+        // The earn webview closes as LLD navigates to the dapp platform page.
+        // Verify the main page URL changed to the platform route.
+        await expect(page).toHaveURL(/\/platform\//, { timeout: 10_000 });
```
ypolishchuk-ledger: I would suggest to stick to default timeout to avoid specifying it here

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+          [mockUsdtMorphoStake],
+          "ethereum",
+          account.address ?? "",
+        );
+        const interceptReady = interceptEarnStakes(electronApp, mockResponse);
+        await app.earnDashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
+        await interceptReady;
+
+        await app.earnDashboard.verifyHotStartPage();
+        await app.earnDashboard.verifyPositionRowPresent("USDT");
```
ypolishchuk-ledger: Please use Currency value from enum here and all other places in this file where you use plain string

e2e/desktop/tests/utils/earnMockData.ts
```diff
@@ -0,0 +1,182 @@
...
+    execution_rewards: "923513202584148972",
+    claimable_consensus_rewards: "0",
+    claimable_execution_rewards: "0",
+    type: "EthNativeDetailsView",
+    state: "active_ongoing",
+    is_kiln: true,
+  },
+};
+
+export const mockStEthLidoStake = {
```
ypolishchuk-ledger: this is not in use anywhere

e2e/desktop/tests/utils/earnV2Flags.ts
```diff
@@ -0,0 +1,66 @@
...
+import { LiveAppManifest } from "@ledgerhq/live-common/platform/types";
+
+export const useLocalEarnManifest = process.env.USE_LOCAL_EARN_MANIFEST === "1";
+
+export const EARN_V2_DESKTOP_FLAGS: OptionalFeatureMap = {
+  ...(useLocalEarnManifest && {
+    ptxEarnLiveApp: { enabled: true, params: { manifest_id: "earn-local-manifest" } },
+  }),
+  ptxEarnUi: { enabled: true, params: { value: "v2" } },
+  lwdWallet40: {
```
martijnhjk: The 4.0 feature flag is declared in `featureFlagUtils.ts`. Maybe we can move the earn v2 feature flag to that file and have a specific file for the local earn manifest to keep things clear and separate?

e2e/desktop/tests/specs/earn.v2.spec.ts
martijnhjk: Is it OK to use expect directly here or should we wrap in a step? @VicAlbr @cunhabruno @abdurrahman-ledger please advise.

VicAlbr: imo should be wrapped in a step

e2e/desktop/tests/page/earn.dashboard.page.ts
sergiubreban: v2 it is different from v1, eventually v2 will replace the v1 tests after the rollout. Since we add tests for v2, we want to enhance the tests add more use case (which requires more selectors). The initial plan was to create a new file for tests, but to reuse this file

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: I don't know too much about xray, but I saw it's used for v1 tests.I added these placeholders, and I assume someone from the QA team needs to generate these tickets, right? Or can I just put a random string?

e2e/desktop/tests/utils/earnMockData.ts
sergiubreban: I added this interceptor for a specific use-case where I need a different response from the backend. I am not sure if this is the right way, or if we should create some accounts in Speculos. Looking for your guidance here

e2e/desktop/tests/page/earn.dashboard.page.ts
sergiubreban: we can't construct it from just the ticker. The regex prefix + hasText filter should be enough imo

e2e/desktop/tests/page/earn.dashboard.page.ts
sergiubreban: In this case we need to check if the user navigated away from earn. the only positive check that I could think of could be the route to contain reedem|platform|....all ledger supported routes. And that could break in the future if we add new apps and we forget to update the list

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: it's pretty much the same test from earn.spec.ts, just that this is using the main cta instead of cta next to a currency

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,119 @@
...
+  async verifyIceColdStartPage() {
+    await this.verifyElementIsVisible(this.footerDisclaimer);
+    await this.verifyElementIsNotVisible(this.maxPotentialRewards);
+    await this.verifyElementIsNotVisible(this.walletHeaderAmount);
+  }
+
+  @step("Click ice cold start earn CTA")
+  async clickIceColdStartEarnCTA() {
+    const webview = await this.getWebView();
+    await webview.getByTestId("ice-cold-start-earn-cta").click();
```
martijnhjk: please consider adding the locator as a property on the class for reusability.

martijnhjk: and other instances across this PR

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+    const xrayTicket: string | undefined = "B2CQA-4639";
+    test(
+      "Earn v2 ice cold start page displays correctly",
+      {
+        tag: getTags(account),
+        ...xrayAnnotation(xrayTicket),
+      },
+      async ({ app }) => {
+        await registerXrayLink(test.info());
+        await app.earnV2Dashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
```
martijnhjk: please use `app.mainNavigation.openTargetFromMainNavigation("earn")` instead which is compatible with both legacy wallet and wallet 4.0

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+
+      test(
+        `Earn v2 cold start page shows ${account.currency.ticker} ready to earn`,
+        {
+          tag: getTags(account),
+          ...xrayAnnotation(xrayTicket),
+        },
+        async ({ app, page }) => {
+          await registerXrayLink(test.info());
+          await app.earnV2Dashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
```
martijnhjk: same here, please use `app.mainNavigation.openTargetFromMainNavigation("earn")`

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+          await app.earnV2Dashboard.verifyColdStartPage();
+          await app.earnV2Dashboard.verifyAssetReadyToEarn(account.currency.ticker);
+          await app.earnV2Dashboard.clickAssetEarnCta(account.currency.ticker);
+          await expect(page.getByTestId("modal-container")).toBeVisible();
```
martijnhjk: please consider wrapping in a step and storing the locator as a property on the page object.

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+
+    test(
+      "Earn v2 ice cold start allows inline account addition",
+      {
+        tag: getTags(account),
+        ...xrayAnnotation(xrayTicket),
+      },
+      async ({ app }) => {
+        await registerXrayLink(test.info());
+        await app.earnV2Dashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
```
martijnhjk: see other comments, please use `app.mainNavigation.openTargetFromMainNavigation("earn")`

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+    xrayTicket: string | undefined;
+    verify: (app: any, page: any) => Promise<void>;
+  }[] = [
+    {
+      name: "CTA → Native staking (SOL)",
+      account: Account.SOL_2,
+      ticker: Account.SOL_2.currency.ticker,
+      xrayTicket: "B2CQA-4643",
+      verify: async (app, page) => {
+        await expect(page.getByTestId("modal-container")).toBeVisible();
```
martijnhjk: Are we using this pattern anywhere else? Please write the assertion step in the test and store locators as properties on the page object.

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+
+      test(
+        `Earn v2 ETH staking flow - ${provider.name}`,
+        {
+          tag: getTags(account),
+          ...xrayAnnotation(xrayTicket),
+        },
+        async ({ app, page }) => {
+          await registerXrayLink(test.info());
+          await app.earnV2Dashboard.goAndWaitForEarnToBeReady(() => app.layout.goToEarn());
```
martijnhjk: please use `app.mainNavigation.openTargetFromMainNavigation("earn")` instead.

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+      speculosApp: account.currency.speculosApp,
+      featureFlags: EARN_V2_DESKTOP_FLAGS,
+    });
+
+    const xrayTicket: string | undefined = "B2CQA-4639";
+    test(
+      "Earn v2 ice cold start page displays correctly",
+      {
+        tag: getTags(account),
+        ...xrayAnnotation(xrayTicket),
```
martijnhjk: please consider using the established pattern calling the `annotation` property directly:
`annotation: { type: "TMS", description: xrayTicket },`

e2e/desktop/tests/utils/allureUtils.ts
```diff
@@ -30,6 +31,25 @@ export async function addTmsLink(ids: string[]) {
...
+/**
+ * If an xray ticket is present, register its TMS link in allure.
+ * Call at the top of every test body that has an xray ticket.
+ */
+export async function registerXrayLink(testInfo: TestInfo) {
+  const tmsDescription = getDescription(testInfo.annotations, "TMS");
+  if (tmsDescription) {
+    await addTmsLink(tmsDescription.split(", "));
+  }
+}
```
martijnhjk: Is there a particular value for this abstraction? Can we follow the pattern that is used across the existing specs and avoid this extra layer?

Please note these also allow empty or undefined parameters meaning that the function would be called but in effect annotation would not happen.

e2e/desktop/tests/utils/featureFlagUtils.ts
```diff
@@ -30,3 +30,21 @@ export const LWD_WALLET_40_FF_ENABLED: OptionalFeatureMap = {
 export const LWD_WALLET_40_FF_DISABLED: OptionalFeatureMap = {
   lwdWallet40: { enabled: false },
 };
+
+export const useLocalEarnManifest = process.env.USE_LOCAL_EARN_MANIFEST === "1";
+
+export const EARN_V1_DESKTOP_FLAGS: OptionalFeatureMap = {
+  ptxEarnUi: { enabled: false, params: { value: "v1" } },
+  lwdWallet40: { enabled: false },
```
martijnhjk: this is already stored in a const, please reuse eg:
```suggestion
  ...LWD_WALLET_40_FF_DISABLED,
```

e2e/desktop/tests/utils/featureFlagUtils.ts
```diff
@@ -30,3 +30,21 @@ export const LWD_WALLET_40_FF_ENABLED: OptionalFeatureMap = {
...
+
+export const EARN_V2_DESKTOP_FLAGS: OptionalFeatureMap = {
+  ...(useLocalEarnManifest && {
+    ptxEarnLiveApp: { enabled: true, params: { manifest_id: "earn-local-manifest" } },
+  }),
+  ptxEarnUi: { enabled: true, params: { value: "v2" } },
+  lwdWallet40: {
+    enabled: true,
+    params: { marketBanner: true, graphRework: true, quickActionCtas: true },
+  },
```
martijnhjk: same here, please reuse:
```suggestion
  ...LWD_WALLET_40_FF_ENABLED,
```

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: I'll move the locator to the page object. 

> Are we using this pattern anywhere else?

Are you referring to the verify function that is defined here? Since the tests are very similar, only the verification differs. I added this pattern to avoid code duplication raised by SonarQube CI job

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: That's precisely what `xrayAnnotation` fn does, I just extracted it to a shared utility to avoid repetitive work. Or do you think having it explicitly in each test gives us more value?

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,133 @@
...
+    `asset-item-ticker-${ticker.toLowerCase()}`;
+  private readonly iceColdStartEarnCta = "ice-cold-start-earn-cta";
+  private readonly modalContainer = this.page.getByTestId("modal-container");
+
+  // Ice Cold Start
+
+  @step("Verify ice cold start page")
+  async verifyIceColdStartPage() {
+    await this.verifyElementIsVisible(this.footerDisclaimer);
+    await this.verifyElementIsNotVisible(this.maxPotentialRewards);
```
gbecerra-ledger, line 24: This element is present on the Ice cold start? I see it only on the cold start

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,133 @@
...
+    await this.verifyElementIsNotVisible(this.walletHeaderAmount);
```
gbecerra-ledger, line 25: This element is present on the Ice cold start? I see it only on the cold start

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
sergiubreban: maxPotentialRewards is not visible on iceColdStart, that's why I check verifyElementIsNotVisible. Ice cold start is the only variant that does not contain this element

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
sergiubreban: wallet-header-amount is only visible on hot-start variant

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,133 @@
...
+  async verifyModalOrProviderVisible() {
+    await expect(
+      this.modalContainer.or(this.page.locator("[data-test-id*='provider']")),
+    ).toBeVisible();
+  }
+
+  // Navigation
+
+  @step("Verify navigated away from earn dashboard")
+  async verifyNavigatedFromDashboard() {
```
ypolishchuk-ledger: this is not in use anywhere

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,133 @@
...
+      await expect(
+        this.page
+          .getByTestId("modular-drawer-screen-ACCOUNT_SELECTION")
+          .or(this.page.getByTestId("modular-dialog-screen-ACCOUNT_SELECTION")),
+      ).toBeHidden();
+    }
+  }
+
+  @step("Get stake platform for currency: $0")
+  async getStakePlatform(currencyId: string): Promise<string | null> {
```
ypolishchuk-ledger: this is not in use anywhere

e2e/desktop/tests/specs/earn.v2.spec.ts
martijnhjk: Yeah in my opinion this is over engineered, stating it explicitly in each test is readable and we are not saving any lines or improving functionality by simply putting it into a function. I say let's stick to stating the `annotation` in the spec explicitly.

e2e/desktop/tests/specs/earn.v2.spec.ts
martijnhjk: Thanks for moving the locator!

I think there is always a balance when it comes to readability and duplication and sometimes it is OK to ignore Sonar if justified. At first glance it may seem we are simplifying the test block contents but we are also introducing a pattern different from other tests adding to cognitive load. Please consider keeping the validation scope explicitly in the test so that parameters are simple and readable and the test contains all the information regarding flow and validation. If the verification is really too different we should consider separate tests and using the Page Object to store any shared logic.

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+      async ({ app }) => {
+        await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+        await app.earnV2Dashboard.goAndWaitForEarnToBeReady(() =>
+          app.mainNavigation.openTargetFromMainNavigation("earn"),
+        );
+        await app.earnV2Dashboard.verifyIceColdStartPage();
+        await app.earnV2Dashboard.clickIceColdStartEarnCTA();
+        const selector = await getModularSelector(app, "ASSET");
+        expect(selector).toBeTruthy();
+        await selector!.validateItems();
```
martijnhjk: The assertion error for this will not be very helpful, I suggest throwing an explicit error, eg:
```
if (!selector) {
  throw new Error("Helpful error message here")
}
await selector.validateItems();
```
Or you can add this if condition with error to the validate function itself for the wider benefit and keeping the spec simple.

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  @step("Select account in modular selector: $0")
+  async selectAccountInModularSelector(app: Application, account: Account) {
+    const selector = await getModularSelector(app, "ACCOUNT");
+    if (selector) {
+      await selector.selectAccount(account);
+      await expect(
+        this.page
+          .getByTestId("modular-drawer-screen-ACCOUNT_SELECTION")
+          .or(this.page.getByTestId("modular-dialog-screen-ACCOUNT_SELECTION")),
```
martijnhjk: this looks like the `dialogLocator` and `drawerLocator` in the `getModularSelector()` function. 

Can we add a fluent wait or fluent assert on the `ModularDrawer` and `ModularDialog`? Or a generic function that accepts a predicate to make it easier for other use cases to reuse?

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
martijnhjk: is this used anywhere?

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
```diff
@@ -0,0 +1,133 @@
...
+  // Navigation
+
+  @step("Verify navigated away from earn dashboard")
+  async verifyNavigatedFromDashboard() {
+    await this.verifyElementIsNotVisible(this.tokensToEarnBanner);
+    await this.verifyElementIsNotVisible(this.maxPotentialRewards);
+  }
+
+  @step("Verify navigated to deposit flow")
+  async verifyDepositFlowVisible() {
```
martijnhjk, line 100: same here, is this used, do we need it?

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
martijnhjk: not currently used

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
sergiubreban: this is used for usdt stake flow

e2e/desktop/tests/page/earn.base.page.ts
```diff
@@ -0,0 +1,54 @@
+import { Account } from "@ledgerhq/live-common/e2e/enum/Account";
+import { step } from "tests/misc/reporters/step";
+import { WebViewAppPage } from "./webViewApp.page";
+
+export abstract class EarnBasePage extends WebViewAppPage {
+  protected readonly webviewIdentifier = "earn";
+
+  @step("Go and wait for Earn app to be ready")
+  async goAndWaitForEarnToBeReady(earnFunction: () => Promise<void>) {
```
ypolishchuk-ledger, line 10: This function has no timeout guard; `appReadyPromise` resolves only when "Earn Live App Loaded" appears in the console. If the app never emits that message (crash, env issue, message text changed), the test will hang silently until Playwright's global test timeout, with no diagnostics. Add a `Promise.race` with a timeout that throws a descriptive error

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+  });
+
+  // --- Navigation: CTA Flows ---
+
+  const ctaFlows: {
+    name: string;
+    account: Account;
+    ticker: string;
+    xrayTicket: string;
+    verify: (app: any, page: any) => Promise<void>;
```
ypolishchuk-ledger: I think app should be of type Application and page as Page from Playwright

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+        annotation: { type: "TMS", description: xrayTicket },
+      },
+      async ({ app }) => {
+        await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+        await app.earnV2Dashboard.goAndWaitForEarnToBeReady(() =>
+          app.mainNavigation.openTargetFromMainNavigation("earn"),
+        );
+        await app.earnV2Dashboard.verifyIceColdStartPage();
+        await app.earnV2Dashboard.clickIceColdStartEarnCTA();
+        const selector = await getModularSelector(app, "ASSET");
```
ypolishchuk-ledger: I would suggest creating a validation function for this, like expectModalToBeVisible and move there 3 lines below, also please use standard PW method for visibility verification, such as: `await expect(this.fearAndGreedCard).toBeVisible();`

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+        expect(selector).toBeTruthy();
+        await selector!.validateItems();
+      },
+    );
+  });
+
+  const coldStartCurrencies = [
+    { account: Account.ETH_2, xrayTicket: "B2CQA-4640" },
+    { account: Account.ATOM_2, xrayTicket: "B2CQA-4719" },
```
ypolishchuk-ledger, line 74: Does it make sense to check the cold start for 2 currencies? Can we draw a conclusion that it works as expected, testing with only 1 currency?

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+  for (const { account, hotStartXrayTicket, positionXrayTicket } of activePositionCurrencies) {
+    test.describe(`Hot start & Position - ${account.currency.ticker}`, () => {
+      test.use({
+        userdata: "skip-onboarding",
+        speculosApp: account.currency.speculosApp,
+        featureFlags: EARN_V2_DESKTOP_FLAGS,
+        cliCommands: [liveDataCommand(account)],
+      });
+
+      test(
```
ypolishchuk-ledger, line 132: I would suggest combining these two tests into one

e2e/desktop/tests/specs/earn.v2.spec.ts
```diff
@@ -0,0 +1,391 @@
...
+      async ({ app }) => {
+        await navigateToEarn(app);
+        await app.earnV2Dashboard.clickAssetEarnCta(account.currency.ticker);
+        await app.earnV2Dashboard.selectAccountInModularSelector(app, account);
+        await app.earnV2Dashboard.verifyModalContainerVisible();
+      },
+    );
+  });
+
+  test.describe("CTA → Partner dapp (ETH)", () => {
```
ypolishchuk-ledger: I think this test duplicates "ETH staking flow - ${provider.name}", so I suggest removing this one and keep the one with provider check

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: I added since was listed in jira, but indeed add inline account test contains the same logic and beyond that

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: https://github.com/LedgerHQ/ledger-live/pull/14809#discussion_r3000424117

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: https://github.com/LedgerHQ/ledger-live/pull/14809#discussion_r3000424117

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: You're the best person to advise here. if you want, I can remove one

e2e/desktop/tests/specs/earn.v2.spec.ts
sergiubreban: What X-ray ID should I keep in this case?

ypolishchuk-ledger: you can have 2 X-ray ID or more for a given automated test

e2e/desktop/tests/page/earn.v2.dashboard.page.ts
martijnhjk: sorry my bad!

PR #14894, opened 2026-02-27
fix(mobile): add readonly attribute to any track

sarneijim:
### ✅ Checklist

  - New tests in `src/analytics/__tests__/segment.readOnlyMode.test.ts`.
  - **Analytics (Segment):** readOnlyMode is sent in all events.

---

### 📝 Description

**Problem:** We need to track read-only mode (Lazy Onboarding) in analytics without relying on feature flags—including `readOnlyMode` in any track.

**Solution:**

- **Segment (`segment.ts`):**
  - Add `readOnlyMode` to mandatory properties (`getMandatoryProperties`) via `readOnlyModeEnabledSelector`, so it appears in all events via `extraProperties`.

---

### ❓ Context

- **JIRA or GitHub link:** [LIVE-26959](https://ledgerhq.atlassian.net/browse/LIVE-26959)

---

[LIVE-26959]: https://ledgerhq.atlassian.net/browse/LIVE-26959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/analytics/segment.ts
+++ b/apps/ledger-live-mobile/src/analytics/segment.ts
@@ -478,6 +478,8 @@ export type LoggableEvent = {
 };
 export const trackSubject = new ReplaySubject<LoggableEvent>(30);
 
+type EventType = (string & {}) | "button_clicked" | "error_message";
+
 export function getIsTracking(
   state: State | null | undefined,
   mandatory?: boolean | null | undefined,
@@ -495,7 +497,7 @@ export function getIsTracking(
 }
 
 export const track = async (
-  event: string,
+  event: EventType,
   eventProperties?: Error | Record<string, unknown> | null,
   mandatory?: boolean | null,
 ) => {
@@ -534,7 +536,7 @@ export const getPageNameFromRoute = (route: RouteProp<ParamListBase>) => {
   return snakeCase(routeName);
 };
 export const trackWithRoute = (
-  event: string,
+  event: EventType,
   route: RouteProp<ParamListBase>,
   properties?: Record<string, unknown> | null,
   mandatory?: boolean | null,
@@ -552,7 +554,7 @@ export const flush = () => {
 export const useTrack = () => {
   const route = useRoute();
   const track = useCallback(
-    (event: string, properties?: Record<string, unknown> | null, mandatory?: boolean | null) =>
+    (event: EventType, properties?: Record<string, unknown> | null, mandatory?: boolean | null) =>
       trackWithRoute(event, route, properties, mandatory),
     [route],
   );


apps/ledger-live-mobile/src/analytics/segment.ts
```diff
@@ -435,6 +438,14 @@ export const start = async (store: AppStore): Promise<SegmentClient | undefined>
...
     }
     await updateIdentify();
+    let lastReadOnlyMode: boolean = readOnlyModeEnabledSelector(store.getState());
+    store.subscribe(() => {
+      const current = readOnlyModeEnabledSelector(store.getState());
+      if (lastReadOnlyMode !== current) {
+        lastReadOnlyMode = current;
+        updateIdentify();
+      }
+    });
```
tonykhaov: I think we should instead check where readOnlyMode redux state is updated and then there call `updateIdentify`

gre-ledger:
```suggestion
```

correct me if i'm wrong, but i don't think it's necessary to actually do all these additions

you don't have to recall updateIdentify() again 

the `extraProperties` will be used in runtime and since you did use `  const readOnlyMode = readOnlyModeEnabledSelector(state);` you directly infer at runtime from the state,

i'm not sure why we even need to do updateIdentify() in other places, but to me here it's not needed 🤔 especially that you're talking about track() part , which is the one inferred in runtime.

gre-ledger: (same feedback from @tonykhaov  i think )

gre-ledger:
Could you also review the SonarQube feedbacks ( you can open the SonarQube UI and sort out the topics, not fix them if irrelevant to your PR but at least could categorize them ) , thanks 👍 
screenshot_2026-02-26_at_12 50 58

sarneijim:
> Could you also review the SonarQube feedbacks ( you can open the SonarQube UI and sort out the topics, not fix them if irrelevant to your PR but at least could categorize them ) , thanks 👍 screenshot_2026-02-26_at_12 50 58

Small refactors. Change it!

apps/ledger-live-mobile/src/analytics/segment.ts
```diff
@@ -468,14 +478,12 @@ export type LoggableEvent = {
 };
 export const trackSubject = new ReplaySubject<LoggableEvent>(30);
 
-type EventType = string | "button_clicked" | "error_message";
```
tonykhaov, line 471: hey, I don't think it is a good idea to remove it, just seems to me that this isn't properly typed:

we should have:
`type EventType = (string & {}) | "button_clicked" | "error_message";`

so "button_clicked" | "error_message" are hinted in the autocomplete since they are often repeated

BUT we also allow any other event name

apps/ledger-live-mobile/src/analytics/segment.ts
```diff
@@ -287,30 +318,16 @@ const extraProperties = async (store: AppStore) => {
...
     .map(([key]) => key);
 
-  const accountsWithFunds = accounts
-    ? [
-        ...new Set(
-          accounts
-            .filter(account => account?.balance.isGreaterThan(0))
-            .map(account => account?.currency?.ticker),
-        ),
-      ]
```
tonykhaov, line 297: seems to me that the scope of the PR is too broad which isn't ideal, 

it should only focus on adding readOnly attribute

apps/ledger-live-mobile/src/analytics/segment.ts
sarneijim: I don't think there's a strong consensus on this. We're not using that pattern elsewhere in the project (we use either closed literal unions or plain string), so I'd keep string for consistency. Sonarqube was complaining about this.

Happy to open this debae in live-engineering channel :)

apps/ledger-live-mobile/src/analytics/segment.ts
sarneijim: The refactors are the SonarQube fixes we got on this file. I understand the approach is to fix what Sonar reports in each PR as we touch the files, as long as the refactors stay small, otherwise the process wouldn’t have been set up this way.

The main scope is still adding the readOnly attribute; the rest are the small cleanups that came out of the last Sonar report on this file.

apps/ledger-live-mobile/src/analytics/segment.ts
tonykhaov: there is actually, talking with data team, they have a convention with events named specifically `button_clicked`. This thus makes devs more aware of it

apps/ledger-live-mobile/src/analytics/segment.ts
tonykhaov: Yeah but it requires more efforts for us reviewers to understand the context

this ticket now not only becomes a simple "add readonly attribute" but make really sure you break nothing during this clean up. This adds complexity to the ticket and that during QA we need to throughly test them

we have bandwidth in our sprint to tackle tech debts so I believe we can create Continuous Improvement tickets where

apps/ledger-live-mobile/src/analytics/segment.ts
sarneijim: In the track() project, it’s used with many more values: "Start", "TestEvent", "Page...", etc. I do not think there is any convention. Anyway, I’ve made the change.

apps/ledger-live-mobile/src/analytics/segment.ts
sarneijim: If the intention was to make the changes in separate PRs, I understand SonarQube wouldn’t have been implemented this way. If you disagree with how it’s set up, please talk to @gre or discuss it in live-engineering.

tonykhaov: for this PR it's fine and since we're in a hurry no problem


but in the future that'd be very helpful to split them pls!

tonykhaov (approved):

PR #14953, opened 2026-03-03
feat(LIVE-26864): hedera last block and get block v2

mdomanski-ext-ledger:
### ✅ Checklist

  - existing logic shouldn't be affected

### 📝 Description

This PR adds `getBlockV2` and `lastBlockV2` methods that are based on new ERC20 data source. They will be used when `useHgraphForErc20` feature flag is true. 

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-26864
- source branch: https://github.com/LedgerHQ/ledger-live/pull/13683

---

### DIFF
--- a/libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
+++ b/libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
@@ -4,7 +4,6 @@ import type {
   BlockOperation,
   BlockTransaction,
 } from "@ledgerhq/coin-framework/api/types";
-import BigNumber from "bignumber.js";
 import { FINALITY_MS, HEDERA_TRANSACTION_NAMES } from "../constants";
 import { apiClient } from "../network/api";
 import { hgraphClient } from "../network/hgraph";
@@ -23,7 +22,20 @@ import {
   getDateRangeFromBlockHeight,
   mergeTransactionsFromDifferentSources,
   toEntityId,
+  extractFeesPayer,
+  millisToSeconds,
+  secondsToNanos,
 } from "./utils";
+
+function isStakingTransactionType(
+  item: MergedTransaction,
+): item is Extract<MergedTransaction, { type: "mirror" }> {
+  return item.type === "mirror" && item.data.name === HEDERA_TRANSACTION_NAMES.UpdateAccount;
+}
+
+function getMirrorTransaction(item: MergedTransaction): HederaMirrorTransaction {
+  return item.type === "mirror" ? item.data : item.data.mirrorTransaction;
+}
 
 function createBlockOperationFromCoinTransfer({
   payerAccount,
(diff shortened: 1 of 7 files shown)


libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
```diff
@@ -0,0 +1,250 @@
...
+  chargedFee,
+  transfer,
+  rewardTransfers,
+}: {
+  payerAccount: string;
+  chargedFee: number;
+  transfer: HederaMirrorCoinTransfer;
+  rewardTransfers: HederaMirrorTransaction["staking_reward_transfers"];
+}): BlockOperation {
+  let amount = BigInt(transfer.amount);
```
henri-ly: [nitpick] I'm not a bigfan of using let but I'll let you choose which version you like the most, because I find mine not that great also : 

```
const amount =
  BigInt(transfer.amount) +
  (payerAccount === address ? BigInt(chargedFee) : 0n) -
  BigInt(rewardTransfers.find(r => r.account === transfer.account)?.amount ?? 0);
```

libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  // block data should be immutable: do not allow querying blocks on non-finalized time range
+  if (end.getTime() > Date.now() - FINALITY_MS) {
+    throw new Error(`Block ${height} is not available yet`);
+  }
+
+  const latestHgraphIndexedTimestampNs = await hgraphClient.getLatestIndexedConsensusTimestamp();
+  const startSeconds = new BigNumber(start.getTime()).dividedBy(1000);
+  const endSeconds = new BigNumber(end.getTime()).dividedBy(1000);
+  const endNanos = endSeconds.multipliedBy(10 ** 9);
```
henri-ly: [nitpick] maybe if it's something that you are using a lot, have a small utils function like 

`timeToSeconds`
or 
`secondsToNano` to help readability

libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
```diff
@@ -0,0 +1,250 @@
...
+    fetchAllPages: true,
+  });
+
+  // analyze CRYPTOUPDATEACCOUNT transactions to distinguish staking operations from regular account updates.
+  // this creates a map of transaction_hash -> StakingAnalysis to avoid repeated lookups.
+  const stakingAnalyses = await Promise.all(
+    mergeResult.merged
+      .filter((item): item is Extract<MergedTransaction, { type: "mirror" }> => {
+        return item.type === "mirror" && item.data.name === HEDERA_TRANSACTION_NAMES.UpdateAccount;
+      })
```
henri-ly: [nitpick] I prefer to have a guard seperated from function on top of it for readability but it's up to you.

libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
```diff
@@ -0,0 +1,250 @@
...
+      .map(async item => {
+        const payerAccount = item.data.transaction_id.split("-")[0];
+        const analysis = await analyzeStakingOperation(payerAccount, item.data);
```
henri-ly: [question] I think since we are in a promise.All it should have no problem with async right ?

libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        return [item.data.transaction_hash, analysis] as const;
+      }),
+  );
+  const stakingAnalysisMap = new Map(stakingAnalyses);
+
+  const blockTransactions: BlockTransaction[] = mergeResult.merged.map(item => {
```
henri-ly, line 194: [comment] This one is hard to read, I have a feeling that we can upgrade it but right now I don't want it to be a blocker, maybe i'll come with something later, we can leave it as this if you find it easy to maintain and debug

henri-ly (dismissed):

libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
mdomanski-ext-ledger: Yup, we need an array of promises for Promise.all and this is what this `.map` produces, async is supported as expected

libs/coin-modules/coin-hedera/src/logic/getBlock.v2.ts
mdomanski-ext-ledger: Unfortunately it's a bit complex because of the two data sources + possibility to mix things in a single tx with Hedera. For example, you can broadcast one transaction that:
- sends 1 HBAR to 0.0.1234
- sends 1 HBAR to 0.0.4321
- sends 1 PACK (token) to 0.0.1111
- triggers auto staking rewards claim on your account

In the API, this is represented as a single object with different transfer arrays. I tried to decompose it as much as possible and right now it feels acceptable when it comes to the maintenance and debugging. However, I noticed that two small utils could be used there + I'm open to any suggestions, we can iterate on this later

henri-ly (approved):

PR #14961, opened 2026-03-03
fix(live-countervalues): align status.oldestDateRequested with data cut by selectedTimeRange

gre-ledger:
### ✅ Checklist

  - ability to refresh countervalues history when going wider time window range. (eg 1M -> 1Y)

### 📝 Description

- **Bug:** When exporting countervalues with `selectedTimeRange`, history was trimmed but `status.oldestDateRequested` was left unchanged. After restore, `needOlderReload` was wrong and the missing range was not refetched.
- **Fix:** In `exportCountervalues`, when filtering by `selectedTimeRange`, set each exported pair’s `oldestDateRequested` to `max(current, dailyLimit)` so the persisted status matches the data we keep.
- **Tests:** Added 4 tests for status/`oldestDateRequested` when data is cut or not.

### ❓ Context

- **JIRA or GitHub link**: ctx: @KVNLS on slack
https://ledgerhq.atlassian.net/browse/LIVE-27579
---

### DIFF
--- a/libs/live-countervalues/src/logic.ts
+++ b/libs/live-countervalues/src/logic.ts
@@ -60,12 +60,13 @@ export function hasNewCountervaluesToExport(
 }
 
 // Raw state for db: history + status; no "latest" (so persistence only changes when history changes).
+// When data is cut by selectedTimeRange, status.oldestDateRequested must be aligned so it never
+// claims we have data older than what we actually persist (otherwise needOlderReload would be wrong).
 export function exportCountervalues(
   { data, status }: CounterValuesState,
   trackingPair: TrackingPair[],
   selectedTimeRange?: PortfolioRange,
 ): CounterValuesStateRaw {
-  const out = { status } as CounterValuesStateRaw;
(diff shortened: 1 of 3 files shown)


libs/live-countervalues/src/logic.ts
```diff
@@ -77,6 +78,9 @@ export function exportCountervalues(
     ? formatCounterValueDay(new Date(Date.now() - dailyRetentionDays * 24 * 60 * 60 * 1000))
     : null;
 
+  const exportedPairIds: string[] = [];
+  const out = { status: { ...status } } as CounterValuesStateRaw;
```
mcayuelas-ledger, line 82: Why do we neee to spread this one ?

gre-ledger: indeed. no need as it gets recreated

gre-ledger: actually we need it

```
out.status[path] = { ...s, oldestDateRequested: dailyLimit };
```

we modify out.status in place here. and to not overrides the one from input, we copy it.

PR #15000, opened 2026-03-03
fix(desktop): use lastOnboardedDevice in hasOnboardedDeviceSelector

cfloume:
### ✅ Checklist

  - `hasOnboardedDeviceSelector` now returns `true` when `lastOnboardedDevice` is set, even if `lastSeenDevice` is still null. This fixes false negatives for users who completed onboarding but haven't yet performed a device action (pairing, firmware update, etc.).

### 📝 Description

`hasOnboardedDeviceSelector` previously relied solely on `lastSeenDevice`, which is only populated after a device action (e.g. pairing or firmware update). Users who completed onboarding without performing such an action were incorrectly treated as not having an onboarded device.

The selector now checks `lastOnboardedDevice` (set at onboarding completion) first, falling back to `lastSeenDevice` for backward compatibility with existing users.

**Changes:**
- Rewrote `hasOnboardedDevicector` to use `lastOnboardedDeviceSelector(state) !== null || lastSeenDeviceSelector(state) !== null`
- Removed leftover debug `console.log` statements
- Removed unused `setLastSeenDevice` import from `Tutorial/index.tsx`
- Rewrote unit tests to cover both `lastOnboardedDevice` and `lastSeenDevice` paths

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-27157

### DIFF
--- a/.changeset/poor-falcons-explode.md
+++ b/.changeset/poor-falcons-explode.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+fix: correct hasOnboardedDevice selector by using lastOnboardedDevice in logic check
--- a/apps/ledger-live-desktop/src/mvvm/features/QuickActions/__tests__/useQuickActions.test.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/QuickActions/__tests__/useQuickActions.test.ts
@@ -405,7 +405,11 @@ describe("useQuickActions", () => {
       const { result } = renderHook(() => useQuickActions(trackingPageName), {
         initialState: {
           accounts: [createAccountWithFunds()],
-          settings: { hasCompletedOnboarding: true, lastSeenDevice: null },
+          settings: {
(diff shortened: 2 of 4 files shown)


cfloume:
@sarneijim –
Good point, it is weird that `lastSeenDevice` stays null after onboarding.

That said, I think adding a `getDeviceInfo` call to the completion screen would be a bigger change with more risk than it's worth here:
* `lastSeenDevice` is a `DeviceModelInfo`, so populating it requires actual device communication via `getDeviceInfo`.
* Adding a device call during onboarding could fail or delay things if the device is disconnected or busy after its own setup.
* The selector is really answering "has the user onboarded a device?". `lastOnboardedDevice` is the most direct signal for that. The `lastSeenDevice` fallback is just backward compat.

I'd rather keep this change small and scoped. If we want to populate `lastSeenDevice` earlier as a follow-up, that could make sense, but it's a separate concern from fixing this selector.

PR #15062, opened 2026-03-05
[LWDM] feat(validate transaction evm): :sparkles: add validate transaction before broadcast

qperrot:
### ✅ Checklist

  - ...

### 📝 Description

Create a new hook that happens just before the broadcast, which allows the coin-module to check if a tx is still valid or not.
if tx is invalid, we ask the user to start the flow again

| desktop | mobile |
|-----|----|
| Screenshot 2026-03-05 at 14 15 49 | Screenshot 2026-03-05 at 13 49 51 |

### ❓ Context

- **JIRA or GitHub link**:  [LIVE-26161](https://ledgerhq.atlassian.net/browse/LIVE-26161)

---

[LIVE-26161]: https://ledgerhq.atlassian.net/browse/LIVE-26161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-framework/src/api/types.ts
+++ b/libs/coin-framework/src/api/types.ts
@@ -675,14 +675,6 @@ export type AlpacaApi<
    * @throws if the transaction is rejected by the network (e.g., invalid signature, insufficient funds)
    */
   broadcast: (tx: string, broadcastConfig?: BroadcastConfig) => Promise<string>;
-
-  /**
-   * Validate a transaction signature.
-   *
-   * @param signature the transaction signature
-   * @returns { error: Error | undefined } object with the error if the transaction is invalid, undefined if the transaction is valid
-   */
-  validateTransaction: (signature: string) => Promise<{ error: Error | undefined }>;
 };
 
 export type ChainSpecificRules = {
@@ -707,7 +699,7 @@ export type BridgeApi<
   getAssetFromToken?: (token: TokenCurrency, owner: string) => AssetInfo;
   computeIntentType?: (transaction: Record<string, unknown>) => string;
   refreshOperations?: (operations: LiveOperation[]) => Promise<LiveOperation[]>;
-  // validateTransaction?: (signature: string) => Promise<{ error: Error | undefined }>;
+  validateTransaction?: (signature: string) => Promise<{ error: Error | undefined }>;
 };
 
 export type Api<
--- a/.changeset/silent-laws-glow.md
+++ b/.changeset/silent-laws-glow.md
@@ -1,18 +1,5 @@
 ---
-"@ledgerhq/coin-module-boilerplate": minor
-"@ledgerhq/coin-concordium": minor
-"@ledgerhq/coin-algorand": minor
-"@ledgerhq/coin-polkadot": minor
-"@ledgerhq/coin-stellar": minor
-"@ledgerhq/coin-canton": minor
-"@ledgerhq/coin-hedera": minor
-"@ledgerhq/coin-aptos": minor
-"@ledgerhq/coin-tezos": minor
-"@ledgerhq/coin-aleo": minor
-"@ledgerhq/coin-tron": minor
 "@ledgerhq/coin-evm": minor
-"@ledgerhq/coin-sui": minor
-"@ledgerhq/coin-xrp": minor
 "ledger-live-desktop": minor
 "live-mobile": minor
 "@ledgerhq/live-common": minor
--- a/libs/coin-modules/coin-aleo/src/api/index.ts
+++ b/libs/coin-modules/coin-aleo/src/api/index.ts
@@ -101,8 +101,5 @@ export function createApi(
     getSequence: async (_address: string) => {
       throw new Error("getSequence is not supported");
     },
-    validateTransaction: (_signature: string): Promise<{ error: Error | undefined }> => {
-      throw new Error("validateTransaction is not supported");
-    },
   };
 }
--- a/libs/coin-modules/coin-algorand/src/api/index.ts
+++ b/libs/coin-modules/coin-algorand/src/api/index.ts
@@ -59,8 +59,5 @@ export function createApi(config: AlgorandCoinConfig): Api<AlgorandMemo> {
     ): Promise<CraftedTransaction> => {
       throw new Error("craftRawTransaction is not supported for Algorand");
     },
-    validateTransaction: (_signature: string): Promise<{ error: Error | undefined }> => {
-      throw new Error("validateTransaction is not supported");
-    },
   };
 }
--- a/libs/coin-modules/coin-aptos/src/api/index.ts
+++ b/libs/coin-modules/coin-aptos/src/api/index.ts
@@ -54,8 +54,5 @@ export function createApi(config: AptosConfigApi): AlpacaApi {
     getValidators(_cursor?: Cursor): Promise<Page<Validator>> {
       throw new Error("getValidators is not supported");
     },
-    validateTransaction: (_signature: string): Promise<{ error: Error | undefined }> => {
-      throw new Error("validateTransaction is not supported");
-    },
   };
 }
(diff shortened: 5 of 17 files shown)


apps/ledger-live-desktop/src/renderer/modals/Platform/Exchange/CompleteExchange/Body.tsx
```diff
@@ -26,6 +25,8 @@ export enum ExchangeModeEnum {
 
 export type ExchangeMode = "sell" | "swap" | "fund" | "legacy";
 
+const shouldRestartFlow = (error: Error) => error.name === "InvalidTransactionError";
```
francois-guerin-ledger, line 28: There is 4 instances of this helper, can we mutualise ?

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
```diff
@@ -20,6 +20,13 @@ import { AccountLike } from "@ledgerhq/types-live";
 import { createTransactionBroadcastError } from "@ledgerhq/live-common/errors/transactionBroadcastErrors";
 import AbortButton from "~/renderer/components/AbortButton";
 
+const isInvalidTransactionError = (error: Error) =>
+  error.name === "InvalidTransactionError" ||
+  error.message === "InvalidTransactionError" ||
+  ((error as Error & { cause?: unknown }).cause instanceof Error &&
+    ((error as Error & { cause: Error }).cause.name === "InvalidTransactionError" ||
+      (error as Error & { cause: Error }).cause.message === "InvalidTransactionError"));
```
francois-guerin-ledger, line 28: Is there a use case for this nested check ?

libs/coin-modules/coin-evm/src/logic/validateTransaction.ts
```diff
@@ -0,0 +1,43 @@
...
+      const { hash, blockHeight = null } = await nodeApi.getTransaction(currency, transaction.hash);
+      if (blockHeight || hash) {
+        if (blockHeight) {
+          return { error: new InvalidTransactionError() };
+        }
+        if (hash) {
+          return { error: new InvalidTransactionError() };
+        }
+      }
+    } catch {
```
francois-guerin-ledger, line 20:
```suggestion
        if (blockHeight) {
          return { error: new InvalidTransactionError() };
        }
        if (hash) {
          return { error: new InvalidTransactionError() };
        }
```

libs/coin-modules/coin-evm/src/errors.ts
```diff
@@ -34,3 +34,6 @@ export const InsufficientFunds = createCustomErrorClass("InsufficientFunds");
 // Nfts
 export const NotOwnedNft = createCustomErrorClass("NotOwnedNft");
 export const NotEnoughNftOwned = createCustomErrorClass("NotEnoughNftOwned");
+
+// Broadcast
+export const InvalidTransactionError = createCustomErrorClass("InvalidTransactionError");
```
francois-guerin-ledger, line 39: This error should be in `@ledgerhq/errors`, so other coin-modules can use it

libs/coin-modules/coin-evm/src/logic/validateTransaction.ts
qperrot: await nodeApi.getTransaction(currency, transaction.hash); will fail if the transaction doesn't exist (which should be the case) since we haven't broadcast it yet 
So the catch shouldn't do anything

apps/ledger-live-desktop/src/renderer/modals/Send/steps/StepConfirmation.tsx
qperrot: StepConfirmation decides between NodeError vs ErrorDisplay based on whether the error is an InvalidTransactionError. In some flows, the top-level error can be a generic broadcast error, while the root cause is InvalidTransactionError (i.e. wrapped in error.cause). Without the nested check, that case would be misclassified and rendered as a node/broadcast error.

libs/coin-modules/coin-evm/src/logic/validateTransaction.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (transaction.hash) {
+    try {
+      const { hash, blockHeight = null } = await nodeApi.getTransaction(currency, transaction.hash);
+      if (hash || blockHeight) {
+        return { error: new InvalidTransactionError() };
+      }
+    } catch {
+      // eslint-disable-next-line no-empty
+    }
+  }
```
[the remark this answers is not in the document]
qperrot: yes that is what we want

libs/coin-framework/src/api/types.ts
```diff
@@ -699,6 +699,7 @@ export type BridgeApi<
   getAssetFromToken?: (token: TokenCurrency, owner: string) => AssetInfo;
   computeIntentType?: (transaction: Record<string, unknown>) => string;
   refreshOperations?: (operations: LiveOperation[]) => Promise<LiveOperation[]>;
+  validateTransaction?: (signature: string) => Promise<{ error: Error | undefined }>;
```
francois-guerin-ledger: We have this line in the JIRA ticket

> in alpaca api, implement validateTransaction for EVMs

So we should move this line under `AlpacaApi`

libs/coin-modules/coin-evm/src/logic/validateTransaction.ts
```diff
@@ -0,0 +1,43 @@
...
+    try {
+      const { hash, blockHeight = null } = await nodeApi.getTransaction(currency, transaction.hash);
+      if (blockHeight || hash) {
+        if (blockHeight) {
+          return { error: new InvalidTransactionError() };
+        }
+        if (hash) {
+          return { error: new InvalidTransactionError() };
+        }
+      }
```
francois-guerin-ledger, line 19:
```suggestion
      if (blockHeight || hash) {
        return { error: new InvalidTransactionError() };
      }
```

libs/coin-framework/src/api/types.ts
```diff
@@ -699,6 +707,7 @@ export type BridgeApi<
   getAssetFromToken?: (token: TokenCurrency, owner: string) => AssetInfo;
   computeIntentType?: (transaction: Record<string, unknown>) => string;
   refreshOperations?: (operations: LiveOperation[]) => Promise<LiveOperation[]>;
+  // validateTransaction?: (signature: string) => Promise<{ error: Error | undefined }>;
```
jprudent:
```suggestion
```

libs/coin-framework/src/api/types.ts
```diff
@@ -675,6 +675,14 @@ export type AlpacaApi<
...
    */
   broadcast: (tx: string, broadcastConfig?: BroadcastConfig) => Promise<string>;
+
+  /**
+   * Validate a transaction signature.
+   *
+   * @param signature the transaction signature
+   * @returns { error: Error | undefined } object with the error if the transaction is invalid, undefined if the transaction is valid
+   */
+  validateTransaction: (signature: string) => Promise<{ error: Error | undefined }>;
```
jprudent: since:
- it's only used in bridge (unless it's a common need I'm not aware of ?)
- it has no implementation in most coin-module

=> I suggest to not introduce it in the Alpaca api and keep that in bridge only

libs/coin-framework/src/api/types.ts
jprudent: Hello @francois-guerin-ledger I just found your comment. I don't understand why it should be in Alpaca API ?
(I made this comment https://github.com/LedgerHQ/ledger-live/pull/15062#discussion_r2912870813)

libs/coin-framework/src/api/types.ts
jprudent: maybe I don't understand the purpose, but I would suggest to change the name 

```suggestion
  validateSignature: (signature: string) => Promise<{ error: Error | undefined }>;
```

qperrot: the purpose is to check that the transaction is still valide before broadcast to avoid nonce error for exemple when trying to speedup or cancel a tx
But to me it's more usable for ledger live than alpaca (perhaps I am wrong)

jprudent: yes that's what I think, at least with current api: it only takes the signature as parameter, this is hardly implementable with other coins

PR #14990, opened 2026-03-03
feat(client-ids): add UserId, DatadogId and integrate in Desktop

gre-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  - desktop app

### 📝 Description

This is a reborn of https://github.com/LedgerHQ/ledger-live/pull/13325

This work prepare what is necessary to allow for a smooth integration of Datadog on Desktop ( part of https://github.com/LedgerHQ/ledger-live/pull/14755 )

**Summary**  
Introduces `UserId` and `DatadogId` in `@ledgerhq/client-ids`, extends the identities Redux store with `userId`/`datadogId`, and migrates Desktop from legacy user storage to this unified identities flow. All sensitive IDs stay behind explicit export methods and `export-rules.json` allowlists.

**@ledgerhq/client-ids**
- **New ID classes**: `UserId` and `DatadogId` (Symbol-backed, redacted `toString`/`toJSON`, dedicated `export*` methods).
- **Store**: `IdentitiesState` gains `userId` and `datadogId` (with dummy placeholders). Slice actions: `initFromPersisted`, `importFromLegacy`, `initFromScratch`. Persistence and sync middleware use identities; optional `getUserId` for mobile.
- **Allowlists**: Every export method allowlisted in `export-rules.json`; `DeviceId.exportDeviceIdForPersistence` restricted to persistence only.

**Desktop**
- **Migration**: Removes `~/helpers/user.ts`. Boot runs `initIdentities(store)`: prefers `app.identities` → `initFromPersisted`, else `app.user` / localStorage `userId` → `importFromLegacy`, else `initFromScratch`. DB middleware persists `identities` (userId, datadogId, deviceIds) to app.json.
- **Consumers**: Segment, Braze, Recover, Wallet API, Export Logs, Developer settings, and live-common-setup use `userIdSelector` + the right `export*`. Sentry renderer uses `datadogIdSelector` and `exportDatadogIdForSentry()` (store passed into init).

**E2E**
- New suite: “Identities migration from legacy user” (skip-onboarding userdata). Asserts app.json gets `data.identities` with legacy userId preserved, new datadogId, and empty deviceIds. Adds `waitForIdentitiesInAppJson` and `portfolio.expectIdentitiesPersistedInAppJson`.

**Docs / tooling**
[...]

PR #15091, opened 2026-03-05
fix(desktop): await startAnalytics to ensure user preferences are loaded before tracking

tonykhaov:
### ✅ Checklist

- [x] \`npx changeset\` was attached.
- [ ] **Covered by automatic tests.** _Not covered — the race condition is timing-dependent and hard to reproduce in unit tests; manually verified._
  - `Page Analytics Opt In Prompt Main` tracking event was never received
  - All early `track` / `trackPage` calls during app startup were silently dropped
  - No change in behavior for events fired after startup

### 📝 Description

**Problem**: `startAnalytics(store)` was called without `await` in `init.tsx`. Inside `startAnalytics`, `storeInstance = store` is assigned only **after** `await user()` resolves. Because the call was fire-and-forget, the React tree was rendered and `trackPage("Analytics Opt In Prompt", "Main")` fired before `storeInstance` was ever set.

Every call to `track()` and `trackPage()` guards on `storeInstance`:

\`\`\`ts
// segment.ts
if (!storeInstance || ...) return; // ← storeInstance is null → event dropped
\`\`\`

So the analytics opt-in prompt page event (and any other early tracking calls) were silently discarded.

**Fix**:

1. **`init.tsx`** — `startAnalytics(store)` → `await startAnalytics(store)` wrapped in `try/catch` with `logger.error`, ensuring `storeInstance` is set before React renders **and** any failure is surfaced in logs instead of silently swallowed.

```ts
// before
startAnalytics(store); // fire-and-forget, no error handling

// after
try {
  await startAnalytics(store);
} catch (error) {
  logger.error("Failed to start analytics:", error);
}
```

2. **`segment.ts`** — `initializeSegment()` moved before the `canBeTracked` guard so the Segment instance is always initialized regardless of opt-in state. Previously, opted-out users would skip `initializeSegment()` entirely, breaking `updateIdentify` (called via `runOnceWhen`) for users who later opt in.

```ts
// before: early return if not tracked → segment never initialized for opted-out users
const canBeTracked = trackingEnabledSelector(store.getState());
if (!canBeTracked) return;
initializeSegment();
// ...identify always called if we get here

// after: segment always initialized, identify only fires if opted in
initializeSegment();
const analytics = getAnalytics();
if (!analytics) return;
const canBeTracked = trackingEnabledSelector(store.getState());
if (canBeTracked) {
  // ...identify
}
```

**Before**: app starts → `startAnalytics` fires without await → React renders → `trackPage("Analytics Opt In Prompt", "Main")` runs → `storeInstance` is null → **event dropped**.

**After**: app starts → `await startAnalytics` completes → `storeInstance` set → React renders → `trackPage` runs → **event sent**.

| refuse all | accept all |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/65c9d919-272a-4188-b810-3ec44d11e065"/> | <video src="https://github.com/user-attachments/assets/7160f6b5-5773-4060-83aa-d26d4bfa9a02"/> |

### ❓ Context

- **JIRA or GitHub link**: [LIVE-26998](https://ledgerhq.atlassian.net/browse/LIVE-26998)

---

[LIVE-26998]: https://ledgerhq.atlassian.net/browse/LIVE-26998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/mean-tomatoes-look.md
+++ b/.changeset/mean-tomatoes-look.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+fix: analytics consent page tracking not sent
--- a/apps/ledger-live-desktop/src/renderer/analytics/segment.ts
+++ b/apps/ledger-live-desktop/src/renderer/analytics/segment.ts
@@ -339,9 +339,9 @@ export const startAnalytics = async (store: ReduxStore) => {
   storeInstance = store;
 
   const canBeTracked = trackingEnabledSelector(store.getState());
-  if (!canBeTracked) return;
-
-  const id = userIdSelector(store.getState()).exportUserIdForAnalytics();
+  if (!canBeTracked) {
+    return;
+  }
 
   // Initialize Segment with the write key from config
   initializeSegment();
@@ -349,6 +349,7 @@ export const startAnalytics = async (store: ReduxStore) => {
   const analytics = getAnalytics();
   if (!analytics) return;
 
+  const id = userIdSelector(store.getState()).exportUserIdForAnalytics();
   const allProperties = {
     ...extraProperties(store),
     userId: id,
--- a/apps/ledger-live-desktop/src/renderer/init.tsx
+++ b/apps/ledger-live-desktop/src/renderer/init.tsx
@@ -170,7 +170,6 @@ async function init() {
     deepLinkUrl = url;
   });
   const initialSettings = (await getKey("app", "settings")) || {};
-  startAnalytics(store);
 
   // Build settings to load, ensuring hasCompletedOnboarding is false after a hard reset
   const settingsToLoad = { ...initialSettings };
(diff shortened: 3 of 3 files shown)


gre-ledger:
Are you sure this is stil an issue?

I wonder if i actually fixed it yesterday by effectively making startAnalytics synchronous, i think we can actually remove the `async` function now and no longer need to await it. because I think in https://github.com/LedgerHQ/ledger-live/pull/14990, i've removed the last reason it needed to be async before

Screenshot 2026-03-06 at 10 23 28

I suspect therefore that we could let it be sync again and that maybe there is not even the issue you are fixing on develop anymore?

PR #15130, opened 2026-03-06
chore: (desktop, mobile) get counter value IDs via RTK Query

LL782:
### ✅ Checklist

  - Introduction of new `counterValuesApi` (RTK)
  - Replace `useMarketcapIds` from `live-countervalues-react` with new RTK Query
  - This impacts fetching counter value IDs sorted by market cap in Desktop and Mobile
  - The legecy `useMarketcapIds` is still used in `live-countervalues-react` within `CountervaluesProvider`
     This is be migrated in future.

#### Migrate `useMarketcapIds` from Context to RTK Query

Replace `useMarketcapIds` from `live-countervalues-react` with new RTK Query endpoint in `countervalues-client`, for fetching counter value IDs sorted by market cap.

#### Draft a pattern for createApi with RTK

Files structure around the api, schema and uses the new arch guidelines [Guideline+Data+Layer+Types+Initial+State+Mocks](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6102941793/Guideline+Data+Layer+Types+Initial+State+Mocks) with some modifications:

- Tests include more complete coverage, including error scenarios, integration with schema validation, etc.
- `onSchemaFailure` util is created with the intention of sharing it later. Note the loose coupling in tests and implementation

### ❓ Context

- **JIRA**: https://ledgerhq.atlassian.net/browse/LIVE-27203

---


libs/ledger-live-common/src/counterValues/state-manager/api.ts
```diff
@@ -0,0 +1,23 @@
...
+export const counterValuesApi = createApi({
+  reducerPath: "counterValuesApi",
+  baseQuery: fetchBaseQuery({ baseUrl: getEnv("LEDGER_COUNTERVALUES_API") }),
+  tagTypes: [Tags.CounterValueIdsSortedByMarketCap],
+  endpoints: build => ({
+    getCounterValueIdsSortedByMarketCap: build.query<CounterValueIdsSortedByMarketCap, void>({
+      query: () => "/v3/supported/crypto",
+      providesTags: [Tags.CounterValueIdsSortedByMarketCap],
+      keepUnusedDataFor: THIRTY_MINUTES,
+      transformResponse: transformIdsSortedByMarketCap,
```
LL782: **note:** 

If we get the Zod upgrade (#15150) merged then we can use the schema in `responseSchema` directly and can get rid of `transformResponse`

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * @jest-environment jsdom
+ */
+import { http, HttpResponse } from "msw";
+import { createTestStore } from "@tests/test-helpers/testUtils";
+import { counterValuesApi as api } from "./api";
+import { idsMock } from "./schema";
+import { setupServer } from "msw/node";
```
mcayuelas-ledger, line 8: why don't use `libs/ledger-live-common/src/__tests__/server.ts` ?

libs/ledger-live-common/src/counterValues/state-manager/schema.ts
mcayuelas-ledger, line 1: If we look at https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6102941793/Guideline+Data+Layer+Types+Initial+State+Mocks then `schema.ts` should be under `entities` folder like we have in `cal-client` or `dada-client`

libs/ledger-live-common/src/counterValues/state-manager/api.ts
```diff
@@ -0,0 +1,28 @@
...
+} from "./schema";
+import { onSchemaFailure } from "./onSchemaFailure";
+
+const THIRTY_MINUTES = 30 * 60;
+
+export const counterValuesApi = createApi({
+  reducerPath: "counterValuesApi",
+  baseQuery: fetchBaseQuery({ baseUrl: getEnv("LEDGER_COUNTERVALUES_API") }),
+  tagTypes: [Tags.CounterValueIdsSortedByMarketCap],
+  onSchemaFailure,
```
mcayuelas-ledger: I think Copilot is technically correct

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
LL782: It's a good question. I can see advantages to using it, I also have reasons for not using it. Let's think it through:

**In favour of using shared `setupServer`:**
- **Shared handlers:** Handlers for other APIs are automatically in place
- **Singleton:** Only one instance of `setupServer` _Not sure this is an advantage_
- **Obvious convention:** It's easier to find examples of other handlers, fixture factories, etc.

**For using msw directly:**
- **Simple:** Immediate understanding of where `setupServer` comes from and what it is
- **Clean-up not needed:** Each unit does it's own setup, reset and clean-up code isn't needed
- **Separation of concerns:** Only the API relevant to the test needs to be in context
- **High cohesion:** Fixture, mock handler, test and implementation files all located together
- **Low coupling:** Different entities are not coupled unnecessarily via a test util
- **Obvious relationships:** If APIs other than the one under test need to be mocked we notice the smell

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+jest.mock("./schema", () => ({
+  ...jest.requireActual("./schema"),
+  counterValueIdsSortedByMarketCapSchema: { "~standard": { validate: () => mockedValidate() } },
+}));
```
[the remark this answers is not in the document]
LL782: This is not an issue since `["~standard"].validate` will not run before `mockedValidate` is initialized. Static checks are able to prevent these errors:

```diff
- { "~standard": { validate:  mockedValidate } } // this would not pass static checks
+ { "~standard": { validate: () => mockedValidate() } } // this is okay
```

However, I only separated the order for readability purpose. If it's triggering comments when reading I might as well move it back

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const getIdsEndpoint = `https://countervalues.live.ledger.com/v3/supported/crypto`;
```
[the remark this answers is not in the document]
LL782: What a good idea

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  server.listen();
+  console.error = jest.fn();
+});
+
+afterAll(() => {
+  server.close();
+  console.error = originalConsoleError;
+});
+
+afterEach(() => {
```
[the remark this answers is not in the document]
LL782: Since the `server` in this test is not used in any other files this isn't an issue

libs/ledger-live-common/src/counterValues/state-manager/api.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import {
+  Tags,
+  counterValueIdsSortedByMarketCapSchema,
+  type CounterValueIdsSortedByMarketCap,
+} from "./schema";
+import { onSchemaFailure } from "./onSchemaFailure";
+
+const THIRTY_MINUTES = 30 * 60;
+
```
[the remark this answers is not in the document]
LL782: I disagree. The explainer variable is great for readability. This line and the line where it's used are easy to read and understand. The pattern used elsewhere is important because the variables are not so local and some are used with seconds interfaces and others with milliseconds.

libs/ledger-live-common/src/counterValues/state-manager/api.ts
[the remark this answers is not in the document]
LL782: I see. Let's see if we can update RTK in this case

LL782: ~True. I prefer moving to the latest version of RTK. I'll see if this is doable.~

LL782: I'm curious about this though. Static checks would have detected an issue...

I suspect Copilot may have incomplete information. The node_module itself has the interface:

Image

libs/ledger-live-common/src/counterValues/state-manager/api.ts
```diff
@@ -0,0 +1,28 @@
...
+  reducerPath: "counterValuesApi",
+  baseQuery: fetchBaseQuery({ baseUrl: getEnv("LEDGER_COUNTERVALUES_API") }),
+  tagTypes: [Tags.CounterValueIdsSortedByMarketCap],
+  onSchemaFailure,
+  endpoints: build => ({
+    getCounterValueIdsSortedByMarketCap: build.query<CounterValueIdsSortedByMarketCap, void>({
+      query: () => "/v3/supported/crypto",
+      providesTags: [Tags.CounterValueIdsSortedByMarketCap],
+      keepUnusedDataFor: THIRTY_MINUTES,
+      responseSchema: counterValueIdsSortedByMarketCapSchema,
```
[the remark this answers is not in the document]
LL782: I think your information is incorrect

Image

libs/ledger-live-common/src/counterValues/state-manager/onSchemaFailure.test.ts
```diff
@@ -0,0 +1,30 @@
+import { log } from "@ledgerhq/logs";
+import { NamedSchemaError } from "@reduxjs/toolkit/query";
+import { onSchemaFailure } from "./onSchemaFailure";
```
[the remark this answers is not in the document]
LL782: Once again, your training data seems out of date

Image

libs/ledger-live-common/src/counterValues/state-manager/schema.ts
LL782: So I see it differently. In the guidelines I see a directory called `entities` populated by domain entities. The example is a `user` entity with its schema, selectors, actions and slices inside.

Image

What I've done here in `counterValues` and what has been done before in `cal-client`, `cg-client` and `dada-client` is temporary. When Platform team have created `/domains/` and `/shared/`* we can align and refine the guidelines but that will probably be after the move to Nx.

In the immediate term separating the contents of `/counterValues/` into different directories doesn't make sense to me since all the files belong to the same entity. Re. `onSchemaFailure` I intent that to move to `/shared/` when possible. 

*BTW I think Lucid chart is more up to date than the Confluence page now:

Image

https://lucid.app/lucidchart/592f495f-4d2b-48ef-857c-b9ec6aeec814

libs/ledger-live-common/src/counterValues/state-manager/onSchemaFailure.ts
```diff
@@ -0,0 +1,8 @@
+import { log } from "@ledgerhq/logs";
+import type { SchemaFailureHandler } from "@reduxjs/toolkit/query/react";
+
+export const onSchemaFailure: SchemaFailureHandler = (error, { type, endpoint }) => {
```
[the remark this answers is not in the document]
LL782: This is incorrect. I see `SchemaFailureHandler` and `onSchemaFailure` inside the node module for our version of Zod

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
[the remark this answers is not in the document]
LL782: I've updated things and effectively applied the suggestion

libs/ledger-live-common/src/counterValues/state-manager/api.ts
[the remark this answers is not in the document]
LL782: As per other comments, I see `onSchemaFailure` in the node module for our version of Zod

libs/ledger-live-common/src/counterValues/state-manager/onSchemaFailure.ts
```diff
@@ -0,0 +1,8 @@
...
+  log(`Invalid ${type} response for ${endpoint}`, `Invalid ${endpoint} response schema:`, {
+    issues: error.issues,
+  });
```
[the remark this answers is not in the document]
LL782: This was a good point. I've updated `onSchemaFailure` to accept an apiName param

libs/ledger-live-common/src/counterValues/state-manager/api.test.ts
```diff
@@ -0,0 +1,62 @@
...
+jest.mock("./schema", () => ({
+  ...jest.requireActual("./schema"),
+  counterValueIdsSortedByMarketCapSchema: { "~standard": { validate: () => mockedValidate() } },
+}));
+
+const getIdsEndpoint = `https://countervalues.live.ledger.com/v3/supported/crypto`;
+
+const server = setupServer();
+const store = createTestStore([api]);
+const originalConsoleError = console.error;
```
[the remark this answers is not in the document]
LL782: Address in latest push

libs/ledger-live-common/src/counterValues/state-manager/api.ts
LL782: After looking into the Copilot comments about the methods available on RTK's `createApi`  I have to conclude that Copilot is wrong

libs/ledger-live-common/src/counterValues/state-manager/schema.ts
gre-ledger: @ysitbon currently is working on integrating into our project the @domain systems,
It's not too bad if it's just a matter of folder because in any case, we'll have to move these files when the project will be ready for it and there will be possibly adaptation / changes to do anyway (as the theory and the practice may differ!)
and as for this topic, i don't think there is even an "entity" here, it's just an api to me. there is not even a slice 🤔 
so in future, your work may just move to the data-layer/api

gre-ledger: Complementary notes: the actual entity we may need here is something like `Currency` 🤔  because it seems it's the actual entity you manipulate here, as i look at the selector is about `useCurrenciesByMarketcap`

LL782: Interesting notes, thank you. So, for my better understanding `counterValues` are more a pairing of `currencies`?

gre-ledger: the part related to marketcap sorting feels more related to currencies (the list of assets and their ordering)
than the "countervalues" itself, so even if we actually consume the countervalues api, it's even weird to me that we name it countervalues here, which refers more to the api used to get market prices of assets against other assets (typically on USD)

mcayuelas-ledger: I stand with you but for me schema should not be in api folder 
We will to migrate everything in the end.

But ok :)

PR #15203, opened 2026-03-09
[LWDM] feat: remove account auto-selection on dapp browser v3

Canestin:
### ✅ Checklist

- [x] **Impact of the changes:** Web3Hub

### 📝 Description

https://github.com/user-attachments/assets/a262d361-2679-41d5-85ab-c505abc2158b

### ❓ Context

- **JIRA**: https://ledgerhq.atlassian.net/browse/LIVE-27155

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
@@ -31,9 +31,9 @@ import { openModal } from "~/renderer/actions/modals";
 import { flattenAccountsSelector } from "~/renderer/reducers/accounts";
 import BigSpinner from "../BigSpinner";
 import { NetworkErrorScreen } from "./NetworkError";
-import { NoAccountOverlay } from "./NoAccountOverlay";
+import { DappAccountGate } from "./DappAccountGate";
 import { useWebviewState } from "./helpers";
-import { Loader } from "./styled";
+import { Loader as StyledLoader } from "./styled";
 import { WebviewAPI, WebviewProps, WebviewTag } from "./types";
 import { HOOKS_TRACKING_LOCATIONS } from "~/renderer/analytics/hooks/variables";
 import { useModularDrawerVisibility } from "@ledgerhq/live-common/modularDrawer/useModularDrawerVisibility";
@@ -317,8 +317,12 @@ function useWebView(
     manifest,
     customHandlers,
     currentAccountHistDb,
+    setCurrentAccountHistDb,
     inputs,
-  }: Pick<WebviewProps, "manifest" | "customHandlers" | "currentAccountHistDb" | "inputs">,
+  }: Pick<
+    WebviewProps,
+    "manifest" | "customHandlers" | "currentAccountHistDb" | "setCurrentAccountHistDb" | "inputs"
+  >,
   webviewRef: RefObject<WebviewTag | null>,
   tracking: TrackingAPI,
   serverRef: RefObject<WalletAPIServer | undefined>,
@@ -373,12 +377,13 @@ function useWebView(
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [server]);
 
-  const { onDappMessage, noAccounts } = useDappLogic({
+  const { onDappMessage, noAccounts, isLoadingAccounts } = useDappLogic({
     manifest,
     accounts,
     uiHook,
     postMessage: webviewHook.postMessage,
-    currentAccountHistDb,
+    currentAccountHistDb: currentAccountHistDb,
+    setCurrentAccountHistDb: setCurrentAccountHistDb,
     tracking,
     initialAccountId: inputs?.accountId?.toString(),
     referrer: inputs?.referrer?.toString(),
@@ -443,7 +448,7 @@ function useWebView(
     };
   }, [customWebviewStyle, widgetLoaded]);
 
-  return { webviewRef, widgetLoaded, onReload, webviewStyle, noAccounts };
+  return { webviewRef, widgetLoaded, onReload, webviewStyle, noAccounts, isLoadingAccounts };
 }
 
 export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
@@ -452,6 +457,8 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
       manifest,
       inputs = {},
       currentAccountHistDb,
+      setCurrentAccountHistDb,
+      currentAccountHistDbLoaded,
       customHandlers,
       onStateChange,
       hideLoader,
@@ -503,11 +510,12 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
       }
     }, [webviewState, onStateChange]);
 
-    const { webviewStyle, widgetLoaded, noAccounts } = useWebView(
+    const { webviewStyle, widgetLoaded, noAccounts, isLoadingAccounts } = useWebView(
       {
         manifest,
         customHandlers,
         currentAccountHistDb,
+        setCurrentAccountHistDb,
         inputs,
       },
       webviewRef,
@@ -520,8 +528,16 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
     const isDapp = !!manifest.dapp;
     const preloader = isDapp ? "webviewDappPreloader" : "webviewPreloader";
 
-    if (isDapp && noAccounts) {
-      return <NoAccountOverlay manifest={manifest} currentAccountHistDb={currentAccountHistDb} />;
+    if (isDapp && noAccounts && setCurrentAccountHistDb) {
+      return (
+        <DappAccountGate
+          manifest={manifest}
+          isLoadingAccounts={isLoadingAccounts}
+          currentAccountHistDbLoaded={currentAccountHistDbLoaded}
+          setCurrentAccountHistDb={setCurrentAccountHistDb}
+          Loader={Loader}
+        />
+      );
     }
 
     return (
@@ -564,9 +580,9 @@ function DefaultLoader({ isLoading }: { isLoading: boolean }) {
   if (!isLoading) return null;
 
   return (
-    <Loader>
+    <StyledLoader>
       <BigSpinner size={50} />
-    </Loader>
+    </StyledLoader>
   );
 }
 
--- a/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
+++ b/apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
@@ -12,7 +12,7 @@ import { LiveAppManifest } from "@ledgerhq/live-common/platform/types";
 import { getInitialURL } from "@ledgerhq/live-common/wallet-api/helpers";
 import { isUrlAllowedByManifestDomains } from "@ledgerhq/live-common/wallet-api/manifestDomainUtils";
 import {
-  CurrentAccountHistDB,
+  SetCurrentAccountHistDb,
   safeGetRefValue,
   useDAppManifestCurrencyIds,
 } from "@ledgerhq/live-common/wallet-api/react";
@@ -316,10 +316,10 @@ export function useWebviewState(
 
 export function useSelectAccount({
   manifest,
-  currentAccountHistDb,
+  setCurrentAccountHistDb,
 }: {
   manifest: LiveAppManifest;
-  currentAccountHistDb?: CurrentAccountHistDB;
+  setCurrentAccountHistDb?: SetCurrentAccountHistDb;
 }) {
   const { isModularDrawerVisible } = useModularDrawerVisibility({
     modularDrawerFeatureFlagKey: "lldModularDrawer",
@@ -333,7 +333,7 @@ export function useSelectAccount({
   const currencyIds = useDAppManifestCurrencyIds(manifest);
   const { setCurrentAccountHist, setCurrentAccount, currentAccount } = useDappCurrentAccount(
     manifest.id,
-    currentAccountHistDb,
+    setCurrentAccountHistDb,
   );
 
   const onSuccess = useCallback(
@@ -349,11 +349,6 @@ export function useSelectAccount({
     setDrawer();
   }, []);
 
-  const source =
-    currentRouteNameRef.current === "Platform Catalog"
-      ? "Discover"
-      : (currentRouteNameRef.current ?? "Unknown");
-
   const flow = manifest.name;
 
   const dispatch = useDispatch();
@@ -361,6 +356,11 @@ export function useSelectAccount({
   const { openAssetAndAccount } = useOpenAssetAndAccount();
 
   const onSelectAccount = useCallback(() => {
+    const source =
+      currentRouteNameRef.current === "Platform Catalog"
+        ? "Discover"
+        : (currentRouteNameRef.current ?? "Unknown");
+
     if (modularDrawerVisible) {
       dispatch(setFlowValue(flow));
       dispatch(setSourceValue(source));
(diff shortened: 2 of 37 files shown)


apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
```diff
@@ -16,6 +17,24 @@ export function NoAccountScreen({ manifest, currentAccountHistDb }: NoAccountScr
...
     currentAccountHistDb,
   });
+  const [displayAddAccountButton, setDisplayAddAccountButton] = useState(false);
+
+  useEffect(() => {
+    handleAddAccountPress();
+
+    const timeoutDisplayAddAccountButton = setTimeout(() => {
+      setDisplayAddAccountButton(true);
+    }, 1000);
```
Canestin: This is to avoid displaying the button before the modal appears automatically.
We display it after 1 second so that if the user closes the modal, they remain on the screen and can press the button.
We could also keep the previous behavior and simply display the button immediately and not open the modal automatically.

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -46,12 +41,23 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
       ref,
       onStateChange,
     );
     const [error, setError] = useState(false);
+    const [waitingForAccounts, setWaitingForAccounts] = useState(true);
+
+    useEffect(() => {
+      const timeout = setTimeout(() => {
+        setWaitingForAccounts(false);
+      }, 1000);
```
Canestin: This is because `currentAccountHistDb` takes a bit of time to load. As a result, even if there are accounts, it is initially empty (so `noAccounts` is `true`)
Since we cannot know in advance whether `currentAccountHistDb` is actually empty or simply hasn’t finished loading yet, I wait for that short delay

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -379,7 +378,6 @@ function useWebView(
     postMessage: webviewHook.postMessage,
     currentAccountHistDb,
     tracking,
-    initialAccountId: inputs?.accountId?.toString(),
```
Justkant, line 383: It's important to keep the input from a deeplink or other navigation, we still need to have account selected on first visit from swap live-app or earn live-app for example
Can you work out how to keep it please ?

Canestin: Done here https://github.com/LedgerHQ/ledger-live/pull/15203/commits/92a75bbfcf2f6f89a7c8f06d41ae2cf2a0fcc6fa

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
Justkant: Why can't we just show it directly ?
It shouldn't really be an issue no, or is it because the modal takes some time to open ?

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
```diff
@@ -16,6 +17,24 @@ export function NoAccountScreen({ manifest, currentAccountHistDb }: NoAccountScr
...
+
+    return () => {
+      clearTimeout(timeoutDisplayAddAccountButton);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
Justkant: It would be good to memoize the useSelectAccount and not disable the exhaustive-deps check

Image

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
```diff
@@ -16,6 +17,24 @@ export function NoAccountScreen({ manifest, currentAccountHistDb }: NoAccountScr
...
+
+  if (!displayAddAccountButton) {
```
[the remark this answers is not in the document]
Justkant: We could also rename this if we keep it

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
```diff
@@ -16,6 +17,24 @@ export function NoAccountScreen({ manifest, currentAccountHistDb }: NoAccountScr
...
+    return null;
+  }
```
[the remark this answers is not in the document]
Justkant: I think this is why we get the bottom bar issue on the current webview, we should consider remove the timeout and always showing the button, we can disable it until we open the drawer automatically

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -46,6 +46,18 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
+
+      return () => {
+        clearTimeout(timeout);
+      };
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
Justkant: Yes please don't use exhaustive-deps if not needed and in this case because the `setWaitingForAccounts` is from a `useState` it shouldn't give any warning or error

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
Justkant: This doesn't feel reliable at all, we should have a better way to look at `currentAccountHistDb` and it was not a problem until now apparently

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
Canestin: Yes, it because the modal takes some time to open

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -57,6 +69,10 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
     const javaScriptCanOpenWindowsAutomatically =
       internalAppIds.includes(manifest.id) || manifest.id === WC_ID;
 
+    if (waitingForAccounts && noAccounts) {
+      return <DefaultLoader />;
```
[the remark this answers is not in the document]
Justkant: We should use Loader from the props

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -52,6 +52,18 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
+    useEffect(() => {
+      const timeout = setTimeout(() => {
+        setWaitingForAccounts(false);
+      }, 1000);
+
+      return () => {
+        clearTimeout(timeout);
+      };
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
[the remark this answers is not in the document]
Justkant: I think we should rework this, trying to remove the timeout usage

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
```diff
@@ -121,44 +119,17 @@ function useDappAccountLogic({
...
-
-  const storedCurrentAccountIsPermitted = useCallback(() => {
-    if (!currentAccount) return false;
-    return accounts.some(
-      account =>
-        account.type === "Account" &&
-        currencyIds.includes(account.currency.id) &&
-        account.id === currentAccount.id,
-    );
-  }, [currentAccount, accounts, currencyIds]);
```
Justkant, line 161: What happens now if we stored an account that is not allowed anymore ?

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
```diff
@@ -191,22 +162,7 @@ function useDappAccountLogic({
       setCurrentAccount(currentAccountFromHist);
       return;
```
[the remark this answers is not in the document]
Justkant: This is the same as my comment above

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
```diff
@@ -170,43 +138,12 @@ function useDappAccountLogic({
...
-    currentAccountFromHist,
-    firstAccountAvailable,
-    initialAccount,
-    initialAccountSelected,
-    manifest.id,
-    setCurrentAccount,
-    setCurrentAccountHist,
-    storedCurrentAccountIsPermitted,
-  ]);
+  }, [currentAccountFromHist, setCurrentAccount]);
```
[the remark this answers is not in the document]
Justkant: This is actually correct and duplicated from the above comment

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
```diff
@@ -191,22 +162,7 @@ function useDappAccountLogic({
...
```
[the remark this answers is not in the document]
Justkant: Yes I think we're missing some deps because the linter was giving a warning IIRC

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
```diff
@@ -16,6 +17,24 @@ export function NoAccountScreen({ manifest, currentAccountHistDb }: NoAccountScr
...
+    handleAddAccountPress();
+
+    const timeoutDisplayAddAccountButton = setTimeout(() => {
+      setDisplayAddAccountButton(true);
+    }, 1000);
+
+    return () => {
+      clearTimeout(timeoutDisplayAddAccountButton);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
qperrot: Do we really need this // eslint-disable-next-line react-hooks/exhaustive-deps ?

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
Canestin: I removed the time out

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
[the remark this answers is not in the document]
Canestin: I removed this time out

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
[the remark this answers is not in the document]
Canestin: I removed the timeout. Even when I memoize the function, it causes a lot of re-renders of the No Account screen, and the modal does not appear. Is it really necessary to disable `react-hooks/exhaustive-deps` here, i kept it.

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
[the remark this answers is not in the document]
Canestin: I removed it

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
Canestin: I removed it

apps/ledger-live-mobile/src/components/Web3AppWebview/NoAccountScreen.tsx
Canestin: Yes, we really need it.

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
```diff
@@ -349,18 +349,18 @@ export function useSelectAccount({
...
   const flow = manifest.name;
 
   const dispatch = useDispatch();
 
   const { openAssetAndAccount } = useOpenAssetAndAccount();
 
   const onSelectAccount = useCallback(() => {
+    const source =
+      currentRouteNameRef.current === "Platform Catalog"
+        ? "Discover"
```
henri-ly, line 361: [nitpick] Is that fine to have string based instead of enum or const ? I guess it's fine since it was like that but just wanted to mention

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -65,8 +69,17 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
+      if (!currentAccountHistDbLoaded || isLoadingAccounts) {
+        return (
+          <View style={styles.root}>
+            <Loader />
+          </View>
+        );
+      }
+      return (
+        <NoAccountScreen manifest={manifest} setCurrentAccountHistDb={setCurrentAccountHistDb} />
+      );
```
henri-ly, line 81: [nitpick] I'm puzzle having loading not inside NoAccountScreen, but that's nitpick

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -129,7 +140,7 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
 
 WalletAPIWebview.displayName = "WalletAPIWebview";
 
-function DefaultLoader() {
+export function DefaultLoader() {
```
henri-ly: [question] do you know why we do export this function, I think I miss where this loader is needed

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/helpers.ts
Justkant: I'm not sure we had route in a const somewhere like we do on mobile and it's only used for analytics IIRC

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
Justkant: We don't want to render the component below because it will open the select account automatically, I could move all the logic inside the NoAccountScreen as you suggest but I would need to pass the Loader and the other missing props so I thing it's a nice middle ground even if not the prettiest

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
Justkant: I think it was done at first by Canestin, I can remove it

Justkant: I updated it when rebasing

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
Justkant: I updated the component on both mobile and desktop to introduce a DappAccountGate that does the check before showing the NoAccountScreen

henri-ly (approved):

libs/ledger-live-common/src/wallet-api/useDappLogic.ts
```diff
@@ -739,5 +721,7 @@ export function useDappLogic({
     ],
   );
 
-  return { onDappMessage, noAccounts: !currentAccount };
+  const isLoadingAccounts = !currentAccount && !!currentAccountFromHist;
+
+  return { onDappMessage, noAccounts: !currentAccount, isLoadingAccounts };
```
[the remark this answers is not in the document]
Justkant: if you base it on currentAccountIdFromHist it will wait forever in some cases, that's the best solution for now

deepyjr (approved):

PR #15237, opened 2026-03-10
[LWDM] refactor(coin-fmk): promote imports from Framework

francois-guerin-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Types `BroadcastConfig` and `AddressValidationCurrencyParameters` are exported from `types-live`, but this package won't be moved to `alpaca-coin-modules`. This is changing import paths, with a duplication in the case of `BroadcastConfig`.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-27457

---

### DIFF
--- a/libs/coin-framework/src/api/types.ts
+++ b/libs/coin-framework/src/api/types.ts
@@ -1,5 +1,22 @@
 import type { TokenCurrency } from "@ledgerhq/types-cryptoassets";
-import { BroadcastConfig, Operation as LiveOperation } from "@ledgerhq/types-live";
+import type { Operation as LiveOperation } from "@ledgerhq/types-live";
+
+// NOTE: from types-live
+export type BroadcastConfig = {
+  mevProtected: boolean;
+  sponsored?: boolean;
+  source?: TransactionSource;
+};
+
+/**
+ * TransactionSource identifies the origin of a transaction
+ */
+export type TransactionSource = {
+  // Type of the transaction source
+  type: "dApp" | "live-app" | "coin-module" | "swap";
+  // Name/identifier of the source (e.g., manifestId, provider name)
+  name: string;
+};
 
 export type BlockInfo = {
   height: number;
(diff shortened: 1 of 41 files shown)


libs/coin-framework/src/api/types.ts
```diff
@@ -467,7 +484,7 @@ export type AccountInfo = {
 };
 
 export type AddressValidationCurrencyParameters = {
-  currency: CryptoCurrency;
+  currency: { id: string; name: string };
```
Justkant: thought: Do we plan to do the same for the TokenCurrency ?

libs/coin-modules/coin-bitcoin/src/logic.ts
```diff
@@ -40,7 +40,7 @@ export const inferFeePerByte = (t: Transaction, networkInfo: NetworkInfo): BigNu
   return t.feePerByte || networkInfo.feeItems.defaultFeePerByte;
 };
 export const isValidRecipient = async (params: {
-  currency: CryptoCurrency;
+  currency: { id: string; name: string };
```
Justkant: suggestion(non-blocking): maybe we could create a type in the types.ts of the coin-framework, wdyt ?

libs/coin-framework/src/api/types.ts
francois-guerin-ledger: The plan (my opinion, can be discussed / challenged) is to have methods relying on `TokenCurrency` moved into `generic-alpaca` (the same way we already exported signers for instance).

libs/coin-modules/coin-bitcoin/src/logic.ts
francois-guerin-ledger: The end goal it to only pass a currency id, so a new type would be overkill in my opinion. I am adding a comment regarding the needed refactor though

francois-guerin-ledger: Refactor done in https://github.com/LedgerHQ/ledger-live/pull/15266

PR #15242, opened 2026-03-10
[LWDM] refactor(coin-framework): move functions to `AlpacaApi`

francois-guerin-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Move following methods to `AlpacaApi`
- `validateIntent`
- `validateAddress`
- `getNextSequence`

### ❓ Context

- **JIRA or GitHub link**: 

---


francois-guerin-ledger:
E2E desktop: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/52c42c72-1606-4b80-8ac2-fa0c2f1fada0/

libs/coin-modules/coin-aleo/src/api/index.ts
```diff
@@ -98,8 +105,9 @@ export function createApi(
     ): Promise<TransactionValidation> => {
       throw new Error("validateIntent is not supported");
     },
-    getSequence: async (_address: string) => {
-      throw new Error("getSequence is not supported");
+    getNextSequence: async (_address: string) => {
```
jprudent, line 108: There is a design choice here: to throw or to return a "null value/object"

If we return "null" upstream user of the lib can implement a generic flow whatever the coin.

When throwing, lib user must conditionally decide if the function can be called per coin basis.

Do you have an opinion ?

jprudent (approved):

libs/coin-modules/coin-aleo/src/api/index.ts
francois-guerin-ledger: This topic in under ongoing discussion at [Not applicable vs not supported](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/5855281381/Coin+Framework+Weekly+Refinement), but a better error management is needed.

I am fine with throwing when methods are not implemented, but we may want a way to reflect this to users before they even try to call them.

jprudent: Thanks.

Just for the discussion: I'm more favorable to avoid the complexity on user. Either make it uncallable by design (with some TS typing for instance, in last resort make the method an option `?`) or "let it flow" without consequences for caller.

francois-guerin-ledger: I like the idea of having it "uncallable", that is not exposed (and this gets reflected in the returned Typescript interface).

The only drawback is clients like the generic adapter having to constantly cast to fit their own type needs.

PR #15427, opened 2026-03-17
[LWM] Add more logs for LWM E2E test debug

abdurrahman-ledger:
### ✅ Checklist

- [ ] `npx changeset` was attached. (no need)
  - Mobile E2E

### 📝 Description

E2E Mobile improvements: webview/speculos logging.
- **Detox:** Bump to 20.48.0 and patch update (`patches/detox@20.48.0.patch`).

### ❓ Context

- Mobile [Run](https://github.com/LedgerHQ/ledger-live/actions/runs/23641865542)

---

### DIFF
--- a/apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
+++ b/apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
@@ -478,9 +478,9 @@ export const PlatformAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
               return;
             }
           }
-          receive(msg);
+          void receive(msg);
         } catch {
-          // Invalid JSON or receive() rejected — same as historical Platform WebView handling
+          void receive(data);           // let JSON-RPC handle / error-respond
         }
       },
       [receive],
--- a/apps/ledger-live-mobile/src/e2e/webviewNetworkLogCapture.ts
+++ b/apps/ledger-live-mobile/src/e2e/webviewNetworkLogCapture.ts
@@ -9,119 +9,122 @@ export const E2E_WEBVIEW_CONSOLE_LOG_TYPE = "__E2E_WEBVIEW_CONSOLE_LOG__";
 
 export const E2E_WEBVIEW_NETWORK_CAPTURE_SCRIPT = `
 try {
-  const E2E_NETWORK_TYPE = "${E2E_WEBVIEW_NETWORK_LOG_TYPE}";
-  const E2E_CONSOLE_TYPE = "${E2E_WEBVIEW_CONSOLE_LOG_TYPE}";
+  if (!window.__LEDGER_E2E_WEBVIEW_CAPTURE__) {
+    window.__LEDGER_E2E_WEBVIEW_CAPTURE__ = true;
+    const E2E_NETWORK_TYPE = "${E2E_WEBVIEW_NETWORK_LOG_TYPE}";
+    const E2E_CONSOLE_TYPE = "${E2E_WEBVIEW_CONSOLE_LOG_TYPE}";
 
-  const sendNetworkLog = function (entry) {
-    try {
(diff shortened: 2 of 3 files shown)


gre-ledger (approved):

apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
```diff
@@ -456,18 +463,46 @@ export const PlatformAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
+              return;
+            }
+            receive(msg);
+          } catch {
+            try {
+              receive(JSON.parse(e.nativeEvent.data));
+            } catch {
+              // ignore malformed messages
+            }
+          }
```
Justkant: This part is changing a bit too much the logic IMO

I'm not sure we should ignore malformed messages and the double parse and receive seems redundant

apps/ledger-live-mobile/src/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -104,7 +116,14 @@ export const WalletAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
...
         allowsUnsecureHttps={__DEV__ && !!Config.IGNORE_CERTIFICATE_ERRORS}
         javaScriptCanOpenWindowsAutomatically={javaScriptCanOpenWindowsAutomatically}
-        injectedJavaScriptBeforeContentLoaded={manifest.dapp ? INJECTED_JAVASCRIPT : undefined}
+        injectedJavaScriptBeforeContentLoaded={
+          Config.DETOX
+            ? E2E_WEBVIEW_NETWORK_CAPTURE_SCRIPT +
+              (manifest.dapp ? ";\n" + INJECTED_JAVASCRIPT : "")
+            : manifest.dapp
+              ? INJECTED_JAVASCRIPT
+              : undefined
```
Justkant: I'm not a big fan of this change because it doesn't load the injectedScript in the same way in tests and in prod

We could try to push the `E2E_WEBVIEW_NETWORK_CAPTURE_SCRIPT` in `injectedJavaScript`, I'm not sure it works with the `injectedJavaScriptBeforeContentLoaded` also defined but it would be a bit cleaner IMO

apps/ledger-live-mobile/src/e2e/webviewNetworkLogCapture.ts
```diff
@@ -0,0 +1,124 @@
+/**
+ * Injected script to capture fetch/XHR network traffic inside WebViews during e2e tests.
+ * Posts log entries to React Native via postMessage for attachment to Allure (like desktop WebviewLogCollector).
+ */
+
+export const E2E_WEBVIEW_NETWORK_LOG_TYPE = "__E2E_WEBVIEW_NETWORK_LOG__";
+export const E2E_WEBVIEW_CONSOLE_LOG_TYPE = "__E2E_WEBVIEW_CONSOLE_LOG__";
+
+export const E2E_WEBVIEW_NETWORK_CAPTURE_SCRIPT = `
+(function() {
```
Justkant: nitpick: The code is already wrapped in an IIFE in the lib, so I don't think you even need it here

Image

apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
abdurrahman-ledger: If you want non-JSON messages to be handled instead of dropped, that would be a new behavior and needs a defined contract (e.g. pass raw string somewhere); today this WebView only goes through JSON-RPC receive.

Justkant: I know my point was that we do the same thing twice here

`receive(msg);` is literally the same as the fallback `receive(JSON.parse(e.nativeEvent.data));`

PR #15455, opened 2026-03-17
[LWDM] perf(desktop): load bridge and families only when the app shell is used

gre-ledger:
We avoid loading the heavy “bridge + all coins” code until a test (or the app) actually needs the main app shell.

- **Before:** As soon as any test touched code that imported the main app, we loaded the bridge and all coin libs for that worker, even when the test didn’t need them.
- **After:** That heavy code lives in a lazy chunk and loads only when we render the main app shell. Tests that don’t render it never load it → faster runs for most tests.

We also add a small test that checks the boot path doesn’t load the bridge or families by mistake.


gre-ledger:
no perf gain..
this must not be properly working.

PR #15460, opened 2026-03-17
[LWD] PERF STUDY: perf(desktop): run jest with NODE_ENV=production for test speed

gre-ledger:
Disable React and reselect dev-mode checks in unit tests to reduce overhead (e.g. reselect input-stability warnings, React dev warnings).


gre-ledger:
act(...) is not supported in production builds of React.

PR #15465, opened 2026-03-17
[LWDM] feat(sui): add support for testnet config

ishaba:
### ✅ Checklist

  - ...

### 📝 Description

Add support for testnet config for SUI

### ❓ Context

- **JIRA or GitHub link**: 
- [LIVE-27478](https://ledgerhq.atlassian.net/browse/LIVE-27478
)

---

[LIVE-27478]: https://ledgerhq.atlassian.net/browse/LIVE-27478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/tests/fixtures/wallet-api.ts
+++ b/apps/ledger-live-desktop/tests/fixtures/wallet-api.ts
@@ -1151,6 +1151,15 @@ export const expectedCurrencyList = [
   },
   {
     type: "CryptoCurrency",
+    id: "sui_testnet",
+    ticker: "SUI",
+    name: "Sui (Testnet)",
+    family: "sui",
+    color: "#000",
+    decimals: 9,
+  },
+  {
+    type: "CryptoCurrency",
     id: "zenrock",
     ticker: "ROCK",
     name: "Zenrock",
--- a/libs/coin-modules/coin-sui/src/test/config.test.ts
+++ b/libs/coin-modules/coin-sui/src/test/config.test.ts
@@ -35,7 +35,7 @@ describe("Configuration", () => {
     });
 
     it("should return config for specific currency", () => {
-      const config = coinConfig.getCoinConfig(mockCurrency);
+      const config = coinConfig.getCoinConfig(mockCurrency.id);
       expect(config).toEqual({ node: { url: "123" }, status: { type: "active" } });
     });
   });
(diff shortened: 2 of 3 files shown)


libs/coin-modules/coin-sui/src/config.ts
```diff
@@ -9,6 +10,9 @@ export type SuiConfig = {
 
 export type SuiCoinConfig = CurrencyConfig & SuiConfig;
 
-const coinConfig = buildCoinConfig<SuiCoinConfig>();
+const { setCoinConfig, getCoinConfig: _getCoinConfig } = buildCoinConfig<SuiCoinConfig>();
 
-export default coinConfig;
+const getCoinConfig = (currencyId?: string): SuiCoinConfig =>
+  _getCoinConfig(currencyId ? ({ id: currencyId } as CryptoCurrency) : undefined);
```
francois-guerin-ledger, line 16: question: why not passing the `CryptoCurrency` here ?

lysyi3m: @francois-guerin-ledger this is the same approach we used for Concordium  https://github.com/LedgerHQ/ledger-live/pull/15055

in short, "currency" (not CryptoCurrency, nor currencyId - speaking of the concept of "knowing the currency we're operate with" in general) is needed across entire coin-module, starting from bridge/API instance (where config is set) and ending up the lowest-level ones on the network layer. 

in order to make getCoinConfig/setCoinConfig work out of the box, `cryptoCurrency: CryptoCurrency` needs to be passed across the board, which raises concerns of usage types from types-live in the API level, for example

but the thing is: coinConfig resolution is happenning, essentially, based on the `currency.id` - so with this coin-module-specific override we can now use `currencyId: string` in the coin-module, without `cryptoCurrency` dependency

does that make sense? 😄

francois-guerin-ledger: @lysyi3m I was expecting (hopping for) this type of answer, and it completely makes sense!

Two ways it can be improved
- pass the coin config to low layers, so we don't have to recompute it each time
- challenge the need of `CryptoCurrency` in `getCurrencyConfiguration`, when only an id is needed

This is way beyond the scope of this PR and might even be discussed during coin framework meeting

PR #15506, opened 2026-03-18
[LWD] feat(LIVE-26461): send flow record picker

mateuszpalosz-ext:
### ✅ Checklist

  - aleo send flow record picker for private records

### 📝 Description

Record picker for aleo send flow. 

Records view: 

Screenshot from 2026-03-18 10-54-58

Empty records list view: 

Screenshot from 2026-03-17 09-33-49

### ❓ Context

- [https://ledgerhq.atlassian.net/browse/LIVE-26461](https://ledgerhq.atlassian.net/browse/LIVE-26461)

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/aleo/shared/StepRecordPicker.test.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/aleo/shared/StepRecordPicker.test.tsx
@@ -330,7 +330,7 @@ describe("StepRecordPicker", () => {
     expect(screen.getByTestId("aleo-pick-records-alert")).toBeInTheDocument();
     expect(
       screen.getByText(
-        "Maximum spendable amount depends on maximum value your selected Record. Fees are sponsored by Provable for Ledger users.",
+        "The maximum spendable amount depends on the maximum value of your selected record. Fees are sponsored by Provable for Ledger users.",
       ),
     ).toBeInTheDocument();
   });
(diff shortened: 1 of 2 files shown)


mateuszpalosz-ext:
E2E send flow tests to check if it didn't break other coins

https://github.com/LedgerHQ/ledger-live/actions/runs/23242341578

apps/ledger-live-desktop/src/renderer/families/aleo/shared/StepRecordPicker.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const RecordsLimitDescription = styled.p`
+  padding: 10px;
+  font-size: 12px;
+  color: ${p => p.theme.colors.neutral.c70};
+  font-weight: 400;
+  text-align: center;
+`;
+
+const MAX_RECORDS_DISPLAYED = 10;
```
Justkant, line 94: question(non-blocking): Doesn't it block some users if you don't show all the records ?

Also in the ticket it says that it should be 5 records, is it expected to not match ?

mdomanski-ext-ledger: This is some sort of a "dust protection" that we agreed on for MVP. There is no record management yet, so users won't be able to split/join manually. This limitation is added to avoid breaking UI by rendering 10 000 small records when somebody pranks us.

Regarding the ticket, I've just updated it o latest Epic changes that happened yesterday, sorry!

mateuszpalosz-ext: @Justkant the epic changed yesterday and it's still need to be updated, but 10 is the target number, so the current numbers shown on the branch are the final values (more or less :D)

As part of the MVP, it was decided that we won’t display all records. This is a safeguard in case someone maliciously submits, for example, 10k records that would need to be displayed on this screen. At the moment, we also don’t have record joining at the wallet level, nor will there be support for multi record transactions, so we only show the X largest records.

PR #15585, opened 2026-03-19
[LWM] bug/LIVE-27816 - fix console warnings related to Redux useSelector usage

iqbalibrahim-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached. _(Optional for app-only internal refactor; add if your team requires it.)_
- [x] **Covered by automatic tests.** _(No new tests: subscription wiring only; behaviour unchanged.)_
  - **Persistence / DB save:** Same data should still be written when relevant slices change; no user-facing UI change.
  - **QA focus:** Smoke test app launch, change settings/accounts, background/foreground, and confirm persistence still works where exercised.

### 📝 Description

**Problem**

- `ConfigureDBSaveEffects` used `useSelector(identity)` in each `useDBSaveEffect`, subscribing to the **entire** Redux tree. That caused React-Redux dev warnings (“selector identity returned the root state”) and unnecessary rerenders on any state update.
- Using `extractPersistedCALFromState` as the crypto-assets `stateSelector` also warned because it returns a **new object** every call (unstable selector reference).

**Solution**

- Each effect passes a **`stateSelector`** that returns only the slice it cares about (e.g. `state.settings`, `state.accounts`, `countervaluesStateSelector`, `state.cryptoAssetsApi`, etc.).
- The hook uses `useSelector(stateSelector)` for subscriptions and **`store.getState()`** inside the throttled save path when full state is needed for `getChangesStats` / `lense`.
- Crypto assets: subscribe to **`state.cryptoAssetsApi`**; keep **`extractPersistedCALFromState`** as the `lense` for serialization.

**Before → after**

- Before: root-state subscription + unstable CAL-derived selector → warnings and extra rerenders.
- After: slice-level subscriptions + stable RTK slice for CAL → same save semantics, fewer rerenders.

### ❓ Context

- **JIRA or GitHub link:** [LIVE-27816](https://ledgerhq.atlassian.net/browse/LIVE-27816)
- **ADR link (if any):** _N/A_

---

_(Do not edit if you are the PR author.)_

[LIVE-27816]: https://ledgerhq.atlassian.net/browse/LIVE-27816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/components/DBSave.ts
+++ b/apps/ledger-live-mobile/src/components/DBSave.ts
@@ -62,6 +62,8 @@ function useDBSaveEffect<D, S>({
   saveAtStart = false,
 }: Props<D, S>) {
   const store = useStore();
+  const storeRef = useRef(store);
+  storeRef.current = store;
   const selectedSlice = useSelector(stateSelector);
   const forceSave = useRef(saveAtStart);
   const lastSavedState = useRef<MaybeState>(undefined);
@@ -78,7 +80,7 @@ function useDBSaveEffect<D, S>({
       throttleFn(async (): Promise<void> => {
         if (isSaving.current) return checkForSave(); // if we are already saving, we re-schedule
         const { lense, save, getChangesStats } = latestProps.current;
-        const state = store.getState() as State;
+        const state = storeRef.current.getState() as State;
         const prev = lastSavedState.current;
         if (prev !== undefined && prev !== null) {
           const changedStats = getChangesStats(prev, state); // we compare last saved with latest state
@@ -93,7 +95,7 @@ function useDBSaveEffect<D, S>({
         lastSavedState.current = state; // for the next round, we will be able to compare with latest successful state
         forceSave.current = false;
       }, throttle),
-    [throttle, store],
+    [throttle],
   );
   useFlushMechanism(checkForSave);
   // each time the selected slice or props change, we will checkForSave


apps/ledger-live-mobile/src/components/DBSave.ts
```diff
@@ -177,6 +186,7 @@ export const ConfigureDBSaveEffects = () => {
   const trackingPairs = useTrackingPairs();
 
   useDBSaveEffect({
+    stateSelector: countervaluesStateSelector,
```
thesan, line 203: Here the lense depends on the tracking pairs not just the countervalues state. Did you check whether the countervalues updates correctly once you change the tracking pairs?

I think countervalues is  the main reason the logic is that hacky.

apps/ledger-live-mobile/src/reducers/accounts.ts
```diff
@@ -166,7 +166,10 @@ const shallowAccountsSelectorCreator = createSelectorCreator(lruMemoize, (a, b):
     flattenAccounts(b as AccountLikeArray).map(accountHash),
   ),
 );
-export const shallowAccountsSelector = shallowAccountsSelectorCreator(accountsSelector, a => a);
+export const shallowAccountsSelector = shallowAccountsSelectorCreator(
+  accountsSelector,
+  (accounts: Account[]) => accounts.slice(),
```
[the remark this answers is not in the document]
iqbalibrahim-ledger: The combiner isn’t invoked on every selector call—only when lruMemoize’s custom equality says accounts changed in a meaningful way. When it says equal, the previous result (including the prior slice() output) is reused, so we don’t get a new array per dispatch. slice() only runs on cache miss, same as a => a would run on cache miss; the difference is satisfying Reselect’s identity-function dev check without dropping custom equality.

LucasWerey (dismissed):

tonykhaov (dismissed):

Justkant (dismissed):
LGTM for the changes in coin scope

apps/ledger-live-mobile/src/components/DBSave.ts
iqbalibrahim-ledger: @thesan hey I've simplified the code, could you do a check one more time please?

apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
```diff
@@ -480,7 +480,7 @@ export const PlatformAPIWebview = forwardRef<WebviewAPI, WebviewProps>(
           }
           void receive(msg);
         } catch {
-          void receive(data);           // let JSON-RPC handle / error-respond
+          void receive(data); // let JSON-RPC handle / error-respond
```
KVNLS, line 483: Why do we use void here, please? Can we remove it?

apps/ledger-live-mobile/src/components/DBSave.ts
```diff
@@ -75,33 +77,38 @@ function useDBSaveEffect<D, S>({
...
             isSaving.current = false;
           }
-          lastSavedState.current = state; // for the next round, we will be able to compare with latest successful state
-          forceSave.current = false;
         }
+        lastSavedState.current = state; // for the next round, we will be able to compare with latest successful state
+        forceSave.current = false;
       }, throttle),
-    [throttle],
+    [throttle, store],
```
thesan: Sorry I missed this earlier but doesn't this break the throttle if we redeclare it on every `store` changes. FYI [`options.leading=true`](https://lodash.com/docs/4.17.23#throttle) by default so it could potentially unthrottle the calls

Justkant: I was thinking about that but didn't comment because I think the store ref shouldn't change only the state when you call getState should change, it would correctly reset if we actually changed the store from the provider

apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
Justkant: IIRC the void was used to show that it was a promise not awaited but I could be wrong on this one

apps/ledger-live-mobile/src/components/DBSave.ts
iqbalibrahim-ledger: yep that make sense to me, I have remove `store` as a dependency so as to retain throttle as it was previously.

apps/ledger-live-mobile/src/components/Web3AppWebview/PlatformAPIWebview.tsx
iqbalibrahim-ledger: in this PR it's only a lint change; I'll resolve this comment for this PR.

apps/ledger-live-mobile/src/components/DBSave.ts
thesan: ✔️

PR #15622, opened 2026-03-20
[LWM] fix(walletConnect): add ScriptManager resolver for dynamic imports

amaslakov:
### ✅ Checklist

  - no change in production behavior

### 📝 Description

- change dynamic wallet connect import to a static one to avoid issues in RN app

### ❓ Context

- **JIRA or GitHub link**: 
- **ADR link (if any)**: 

---

### DIFF
--- a/apps/ledger-live-mobile/rspack.config.mjs
+++ b/apps/ledger-live-mobile/rspack.config.mjs
@@ -215,6 +215,12 @@ export default withRozeniteUrlFix(
         plugins: [
           new Repack.RepackPlugin({
             logger: false,
+            extraChunks: [
+              {
+                include: /.*/,
+                type: "local",
+              },
+            ],
           }),
           new ReanimatedPlugin({
             unstable_disableTransform: true,
--- a/.changeset/quiet-squids-sort.md
+++ b/.changeset/quiet-squids-sort.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Configure Re.Pack ScriptManager resolver for dynamic imports
--- a/apps/ledger-live-mobile/index.js
+++ b/apps/ledger-live-mobile/index.js
@@ -15,6 +15,21 @@ import "react-native-url-polyfill/auto";
 // cosmjs use TextEncoder that's not available in React Native but on Node
 import "text-encoding-polyfill";
 
+import { ScriptManager, Script } from "@callstack/repack/client";
+
+ScriptManager.shared.addResolver(async (scriptId) => {
+  if (__DEV__) {
+    return {
+      url: Script.getDevServerURL(scriptId),
+      cache: false,
+    };
+  }
+
+  return {
+    url: Script.getFileSystemURL(scriptId),
+  };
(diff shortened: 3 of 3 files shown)


lysyi3m (dismissed):

apps/ledger-live-mobile/rspack.config.mjs
```diff
@@ -174,6 +174,14 @@ export default withRozeniteUrlFix(
           },
         },
         module: {
+          parser: {
```
KVNLS: We want to leverage code splitting in the app and more in the future, so please don't change this behavior.

KVNLS (changes requested):

Justkant (dismissed):
LGTM but I'll let @KVNLS confirm too if possible

amaslakov:
> Note on extraChunks configuration

This adds an extraChunks config to the Re.Pack plugin that forces all dynamically imported chunks to be bundled locally (inline) rather than treated as remote/async chunks.

> Why:

Concordium coin module uses a dynamic import() for WalletConnect, which Re.Pack interprets as a [code-splitting boundary](https://re-pack.dev/docs/features/code-splitting#remote-chunks), producing a separate chunk. Without a configured ScriptManager resolver, we get a runtime error because the app doesn't know how to load these chunks.

These changes are a necessary stepping stone for proper code splitting support in the app. Re.Pack requires explicit chunk configuration via extraChunks and a ScriptManager resolver to handle dynamically loaded modules — both are added in this PR. Currently everything is set to "local" (bundled inline), but this can be updated later to support remote chunks when the team is ready to fully leverage code splitting. As @KVNLS  noted: "We want to leverage code splitting in the app and more in the future" — this PR lays the groundwork for that.

  The alternative would be to replace the dynamic import with a static one, but that increases the main bundle size and moves away from the code-splitting direction.

Justkant (approved):
LGTM

Also from the doc:
> By default all extra chunks will be saved under \<projectRoot>/build/outputs/\<platform>/remotes directory.
Specifying custom value for this option, will disable default setting - you will need to configure outputPath for type: 'remote' yourself.

Just a note because I suppose we don't want to use remote chunks for now but maybe in the future

ysitbon (changes requested):
Please, can you provide metrics on app startup time before/after `extraChunks`:

- without accounts
- with accounts

You can get the metrics in `Settings / Debug / Performance`.

amaslakov:
@ysitbon `before` is with `extraChunks` config commented out, otherwise unchanged. It's android, emulator

image

KVNLS (changes requested):
Android startup time is degraded with this solution as described in the documentation of Repack.
https://re-pack.dev/docs/features/code-splitting#local-chunks
Please consider another option.

> If you're using Hermes and you compile your code into bytecode bundles, it's better not to use local chunks and instead make the code a part of the main bundle.
>Using local chunks with Hermes and bytecode bundles, will likely result in worse performance.

Justkant:
Is this PR still required ?

PR #15661, opened 2026-03-23
chore: init Claude with importing prompt

gre-ledger:
`claude`

initialized with

`/init`

and a prompt

`use the knowledge available in existing .cursor rules and .github rules in order to initialize Claude completely, create appropriate rules and agents.`


gre-ledger:
(probably not to take in. I think it might be best to minimize a very concise version of all these files)

PR #15733, opened 2026-03-25
[LWDM] feat: [LIVE-26450] include tezos fa2 tokens on balance

acewf:
### ✅ Checklist

  - Nothing changes on the UI
  - Alpaca method getBalance now includes information about FA2 tokens

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-26450
- **ADR link (if any)**: 

---

### DIFF
--- a/libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
@@ -1,9 +1,10 @@
 import { HttpResponse, http } from "msw";
 import { setupServer } from "msw/node";
 import coinConfig, { TezosCoinConfig } from "../config";
+import type { APITokenBalance } from "../network/types";
 import { getBalance } from "./getBalance";
 
-const mockTokensBalancesWithAllInfo = [
+const mockTokensBalancesWithAllInfo: APITokenBalance[] = [
   {
     id: 2398642347442177,
     account: { address: "tz1TzrmTBSuiVHV2VfMnGRMYvTEPCP42oSM8" },
@@ -12,13 +13,10 @@ const mockTokensBalancesWithAllInfo = [
       contract: { alias: "BTCtz", address: "KT1T87QbpXEVgkwsNPzz8iRoah3SS3D1MDmh" },
       tokenId: "0",
       standard: "fa2",
-      totalSupply: "100262584",
       metadata: {
         name: "BTCtez",
         symbol: "BTCtz",
         decimals: "8",
-        thumbnailUri: "ipfs://QmfAJX4A534cpp2XjDGSM8b8PrqnSt8b7ytuteY8MXkBPD",
-        shouldPreferSymbol: true,
       },
     },
     balance: "2000",
@@ -30,7 +28,7 @@ const mockTokensBalancesWithAllInfo = [
   },
 ];
 
-const mockTokensBalancesWithMissingInfo = [
+const mockTokensBalancesWithMissingInfo: APITokenBalance[] = [
   {
     id: 2398642347442177,
     account: { address: "tz1TzrmTBSuiVHV2VfMnGRMYvTEPCP42oMMM" },
@@ -39,7 +37,6 @@ const mockTokensBalancesWithMissingInfo = [
       contract: { alias: "BTCtz", address: "KT1T87QbpXEVgkwsNPzz8iRoah3SS3D1MDmh" },
       tokenId: "0",
       standard: "fa2",
-      totalSupply: "100262584",
     },
     balance: "2000",
     transfersCount: 2,
@@ -143,6 +140,11 @@ describe("getBalance", () => {
           assetReference: "KT1T87QbpXEVgkwsNPzz8iRoah3SS3D1MDmh",
           assetOwner: "tz1TzrmTBSuiVHV2VfMnGRMYvTEPCP42oMMM",
           name: "BTCtz",
+          unit: {
+            code: "BTCtz",
+            magnitude: 0,
+            name: "BTCtz",
+          },
         },
       },
     ]);
--- a/libs/coin-modules/coin-tezos/src/logic/getBalance.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/getBalance.ts
@@ -7,11 +7,18 @@ import api from "../network/tzkt";
  * followed by any token balances associated with the address.
  */
 export async function getBalance(address: string): Promise<Balance[]> {
-  const [apiAccount, tokensBalancesRaw] = await Promise.all([
+  const [apiAccountResult, tokensBalancesResult] = await Promise.allSettled([
     api.getAccountByAddress(address),
     api.getTokensBalances(address),
   ]);
 
+  if (apiAccountResult.status !== "fulfilled") {
+    throw apiAccountResult.reason;
+  }
+
+  const apiAccount = apiAccountResult.value;
+  const tokensBalancesRaw =
+    tokensBalancesResult.status === "fulfilled" ? tokensBalancesResult.value : [];
   const normalized = apiAccount.type === "user" ? BigInt(apiAccount.balance) : 0n;
 
   const stake: Stake | undefined =
@@ -31,7 +38,7 @@ export async function getBalance(address: string): Promise<Balance[]> {
     const name = token.metadata?.name ?? token.contract.alias ?? "";
     const symbol = token.metadata?.symbol ?? token.contract.alias ?? "";
     const unit =
-      magnitude && name && symbol
+      Number.isFinite(magnitude) && name && symbol
         ? {
             magnitude,
             name,
--- a/libs/coin-modules/coin-tezos/src/network/types.ts
+++ b/libs/coin-modules/coin-tezos/src/network/types.ts
@@ -208,7 +208,6 @@ export type APITokenBalance = {
     };
   };
   balance: string;
-  balanceValue: string | null;
   transfersCount: number;
   firstLevel: number;
   firstTime: string;
--- a/libs/coin-modules/coin-tezos/src/network/tzkt.ts
+++ b/libs/coin-modules/coin-tezos/src/network/tzkt.ts
@@ -122,7 +122,8 @@ const api = {
   },
 
   /**
-   * Fetches FA token balances for a given account.
+   * Fetches FA2 token balances (tokenId = 0 only) for a given account.
+   * This is limited to `token.standard=fa2` and `token.tokenId=0` on the TzKT API.
    * https://api.tzkt.io/#operation/Tokens_GetTokenBalances
    */
   async getTokensBalances(address: string): Promise<APITokenBalance[]> {


libs/coin-modules/coin-tezos/src/network/types.ts
```diff
@@ -200,3 +200,33 @@ export type APITokenTransfer = {
...
+  };
+  token: {
+    id: number;
+    contract: { address: string; alias?: string };
+    tokenId: string;
+    standard: "fa1.2" | "fa2";
+    metadata?: {
+      name?: string;
+      symbol: string;
+      decimals: string;
```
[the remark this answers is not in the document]
acewf: on documentation here https://api.tzkt.io/#operation/Tokens_GetTokenBalances only specifies metadata is optional, but doesn't mentioned their properties are optional also.

libs/coin-modules/coin-tezos/src/logic/getBalance.ts
```diff
@@ -1,10 +1,49 @@
...
+      : undefined;
+
+  const tokensBalancesRaw = await api.getTokensBalances(address);
+  const tokensBalance = tokensBalancesRaw.map(({ balance, token }) => ({
+      value: BigInt(balance),
+      asset: {
+        type: token.standard,
+        assetReference: token.contract.address,
+        assetOwner: address,
+        name: token.contract.alias,
```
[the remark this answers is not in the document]
acewf: on requests we are filtering by the tokenId=0.

libs/coin-modules/coin-tezos/src/network/tzkt.ts
```diff
@@ -103,6 +104,25 @@ const api = {
...
+   * Fetches FA token balances for a given account.
+   * https://api.tzkt.io/#operation/Tokens_GetTokenBalances
+   */
+  async getTokensBalances(address: string): Promise<APITokenBalance[]> {
+    // Same rationale as getBlockTransactionsPage: explicit ascending sort keeps the
+    // offset.cr cursor advancing forward regardless of the API's default ordering.
+    const params: Record<string, unknown> = {
+      account: address,
+      "token.standard": "fa2",
+      "token.tokenId": "0",
```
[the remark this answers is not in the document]
acewf: the scope of the task is just to include FA2 tokens where the tokenId is 0

libs/coin-modules/coin-tezos/src/network/types.ts
```diff
@@ -200,3 +186,32 @@ export type APITokenTransfer = {
...
+    };
+  };
+  balance: string;
+  balanceValue: string | null;
```
may01: It seems odd that `balanceValue` can be null

libs/coin-modules/coin-tezos/src/logic/getBalance.ts
```diff
@@ -1,10 +1,52 @@
...
+
+  const tokensBalance = tokensBalancesRaw.map(({ balance, token }) => ({
+    value: BigInt(balance),
+    asset: {
+      type: token.standard,
+      assetReference: token.contract.address,
+      assetOwner: address,
+      name: token.contract.alias,
+      unit: {
+        magnitude: Number.parseInt(token.metadata?.decimals || "0", 10),
```
may01: Is it safe to calculate balance if we failed to get decimals?

libs/coin-modules/coin-tezos/src/network/types.ts
acewf: is what they have on their docs https://api.tzkt.io/#operation/Tokens_GetTokenBalances:~:text=).%0A%5Bsortable%5D-,balanceValue,-string%20Nullable

libs/coin-modules/coin-tezos/src/logic/getBalance.ts
acewf: you recon is better to filter out this entry in case the endpoint didn't provide the decimals?

acewf: the problem is the schema in api docs says balance is a string (so exists for each entry) but the metadata is nullable. 
From the endpoints i queried i it always add info in the metadata object.

libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
```diff
@@ -21,9 +68,76 @@ describe("getBalance", () => {
...
+        HttpResponse.json({
+          type: "user",
+          balance: 15,
+          delegate: { address: "tz1TzrmTBSuiVHV2VfMnGRMYvTEPCP42oMMM" },
+        }),
+      ),
+      http.get("http://tezos.explorer.com/v1/tokens/balances", ({ request }) => {
+        const url = new URL(request.url);
+        const account = url.searchParams.get("account");
+
```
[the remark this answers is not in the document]
acewf: the mocked data is already taking into account this parameters, so checking for extra parameters isn't needed.

libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
```diff
@@ -21,9 +53,43 @@ describe("getBalance", () => {
...
+      },
+      {
+        value: BigInt("2000"),
+        asset: {
+          type: "fa2",
+          assetReference: "KT1T87QbpXEVgkwsNPzz8iRoah3SS3D1MDmh",
+          assetOwner: "tz1TzrmTBSuiVHV2VfMnGRMYvTEPCP42oSM8",
+          name: "BTCtz",
+          unit: { magnitude: 8, name: "BTCtez", code: "BTCtz" },
+        },
```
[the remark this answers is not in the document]
acewf: currently code is `...(stake && { stake }),` which means no stake property will be defined.

libs/coin-modules/coin-tezos/src/logic/getBalance.ts
acewf: not including unit object if metadata properties aren't fully available.

libs/coin-modules/coin-tezos/src/network/types.ts
```diff
@@ -200,3 +186,32 @@ export type APITokenTransfer = {
...
+  transfersCount: number;
+  firstLevel: number;
+  firstTime: string;
+  lastLevel: number;
+  lastTime: string;
```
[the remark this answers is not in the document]
acewf: for the FA2 transactions created on our accounts we always have values for this properties, having the properties as optional is something we can change, but from the transactions i have seen for FA2 when tokenId=0 they are always present.

libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
```diff
@@ -3,8 +3,64 @@ import { setupServer } from "msw/node";
 import coinConfig, { TezosCoinConfig } from "../config";
 import { getBalance } from "./getBalance";
 
+const mockTokensBalancesWithAllInfo = [
```
cted-ledger: all constants must be typed. Here it should be `APITokenBalance[]` I guess...

if you don't want to provide all the information, you can use:
```ts
import type { DeepPartial, DeepPartialReturn } from "@ledgerhq/coin-framework/test/utils";
```

acewf: What is the rationality for making all constants typed? 
I generally prefer a more inference type of approach, in the specific case of constants if we want the object literal type we can use the `as const` https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-inference.


On a side note, and maybe this is just a personal observation in my opinion the LW repo is already filled with too many type declarations that most of the times not only had verbosity to files, but also make contributors opt-in deceiving approaches with the wrong assertions as like ` txRaw as EvmTransactionRaw`    https://github.com/LedgerHQ/ledger-live/blob/develop/libs/coin-modules/coin-evm/src/transaction.ts#L198

This assertions is just telling typescript the object is a given schema, when probably isn't. So why using typescript at all if we can't trust the properties on objects are really there or not.

acewf: updated it, although think is redundant in this case.

libs/coin-modules/coin-tezos/src/network/tzkt.ts
```diff
@@ -103,6 +104,23 @@ const api = {
...
+  async getTokensBalances(address: string): Promise<APITokenBalance[]> {
+    const params: Record<string, unknown> = {
+      account: address,
+      "token.standard": "fa2",
+      "token.tokenId": "0",
+    };
+    const { data } = await network<APITokenBalance[]>({
+      url: `${getExplorerUrl()}/v1/tokens/balances`,
+      params,
+    });
```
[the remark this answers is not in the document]
acewf: We don't support more then 100 tokens on the CAL for tezos.

libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
cted-ledger: Defining strong typings is very important, and I fully agree with you that inference must be preferred...
Do not type correctly the constants here implies we will miss any breaking structure changes in the future...

Anyway in this very specific case, with correct typings, you would have seen that the object you are defining is missing `balanceValue` property which is mandatory in `APITokenBalance`, and you didn't, because there is no inference here (using `as const` does not resolve anything here) :

Screenshot 2026-03-30 at 13 09 50

Then, using `DeepPartial<>` is a way to bypass the need to fulfil all mandatory properties which is sometimes very convenient in tests files where we do not want to pass or define them all. This prevent also the need to cast things with `as unknown as Type`. And finally it also alert you if you defined some unexpected property: 

Screenshot 2026-03-30 at 13 09 59

**So here my recommendation is to remove the `DeepPartial<>` and add `balanceValue` missing property as it is the only one props missing...**

For me the rules are simple: 
- Everything MUST be typed by inference if possible or explicitly if not... 
- Nothing MUST be `any`
- Avoid casts with `unknown` when possible
- Bypass those rules **in tests files exclusively** with `DeepPartial<>` to mitigate the type losing and prevent pollution with unnecessary code

acewf: I checked again, can't see the reason why i have there balanceValue, removing it and using a typed constant.

Thanks for checking this, think we are aligned.

acewf: remembered why we had the balanceValue property, was on the API https://api.tzkt.io/#operation/Tokens_GetTokenBalances schema, but doesn't come on the API response itself.

cted-ledger: yes sometime external types are wrong, and that's why we must always separate types layers. types that are coming from the API can be used internally into the coin module but must never leave it. If necessary, we can override the external type here the API) if we observe it is wrong like you describe. But my concern here was to point the need to type everything strongly :) glad we are aligned

libs/coin-modules/coin-tezos/src/logic/getBalance.ts
```diff
@@ -1,10 +1,68 @@
...
+    api.getTokensBalances(address),
+  ]);
+
+  if (apiAccountResult.status !== "fulfilled") {
+    throw apiAccountResult.reason;
+  }
+
+  const apiAccount = apiAccountResult.value;
+  const tokensBalancesRaw =
+    tokensBalancesResult.status === "fulfilled" ? tokensBalancesResult.value : [];
```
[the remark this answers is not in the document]
acewf: follows same pattern as coin-evm

libs/coin-modules/coin-tezos/src/logic/getBalance.test.ts
```diff
@@ -1,10 +1,63 @@
...
   const mockServer = setupServer();
+
+  beforeAll(() => {
+    mockServer.listen({ onUnhandledRequest: "error" });
+  });
+
+  afterAll(() => {
+    mockServer.close();
+  });
+
```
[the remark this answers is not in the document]
acewf: solving this in another PR.

libs/coin-modules/coin-tezos/src/logic/getBalance.ts
```diff
@@ -1,10 +1,68 @@
...
+        : undefined;
+    return {
+      value: BigInt(balance),
+      asset: {
+        type: token.standard,
+        assetReference: token.contract.address,
+        assetOwner: address,
+        name: token.contract.alias,
+        ...(unit && { unit }),
+      },
```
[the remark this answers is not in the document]
acewf: will correct this in in another PR if necessary.

PR #15855, opened 2026-03-27
Add first draft CLAUDE.md

LL782:
### ✅ Checklist

- [x] `npx changeset` n/a
- [x] **Covered by automatic tests.** n/a
  - Adding CLAUDE.md to give repo level context to Claude Code
  - Moved some Cursor rules into a shared /docs location
  - Pointed CLAUDE.md and Cursor rules to the share location

#### Keeping the file minimal

Following advice from [Claude docs](https://claude.com/blog/using-claude-md-files) and a [Human Layer blog post](https://www.humanlayer.dev/blog/writing-a-good-claude-md) this first `CLAUDE.md` is minimal. 

> **Treat customization as an ongoing practice rather than a one-time setup task.** Projects change, teams learn better patterns, and new tools enter your workflow. A well-maintained CLAUDE.md evolves with your codebase, continuously reducing the friction of working with AI assistance on complex software.

> Each addition **should solve a real problem you have encountered**, not theoretical concerns about what Claude might need.

We should start small and make small, carefully considered changes.

#### Principle of Progressive Disclosure

Aside from keeping this document minimal, we are guided to include only what is applicable across the entire repo. Guidelines that are specific to a particular app, lib or practice (like testing, state-management or how to validate changes) can be located in a separate file and in many cases closer to where it's used. An ADR to further discuss this topic will be created.

#### Single source of truth

Rather than duplicating existing rules and guidelines we take the opportunity in this PR to move some existing Cursor rules to a common location. This introduces the idea that `CLAUDE.md` and `.cursor/*.mdc` are entry points to a single source of documentation. Further work will be done to apply this approach more widely. That work will start with **[LIVE-28131: [AI] - Audit all our docs to list all relevant 1) entry points and 2) content](https://ledgerhq.atlassian.net/browse/LIVE-28131)**
 

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-28130
- **ADR link (if any)**: https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6957301977 (very early draft at time of writing)

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,32 @@
+# Project Context
+
+## "Ledger Wallet" (formerly "ledger-live") overview
+
+- Ledger Wallet is a crypto wallet with desktop and mobile applications
+- This monorepo (pnpm + turbo) provides frontend application code
+- Backend code (REST APIs) is outside this repo
+
+### Key Directories
+
+- `apps/` – Frontend applications for mobile, desktop, etc.
+- `domain/` NEW – Domain entities shared by apps (React, TypeScript)
+- `features/` NEW – Features shared by apps (React, TypeScript)
+- `shared/` NEW – Cross-cutting packages used by more than one feature or domain entity (React, TypeScript)
+- `libs/` LEGACY — Shared libraries
+- `tools/` — Build and deploy related tooling (GitHub actions, scripts)
+
+## General Principles
+
+- Prioritize cleanliness over cleverness
+- Note ideas for regular incremental improvement
+- Prefer community best-practice over legacy code
+
+## Common Commands
+
+- Use pnpm commands for build, dev, linting and testing.
+- See [/docs/common-commands.md](/docs/common-commands.md)
+
+## Validate Before Finishing
+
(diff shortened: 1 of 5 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,32 @@
+# Project Context
+
+## "Ledger Wallet" (formerly "ledger-live") overview
+
+- Ledger Wallet is a crypto wallet with desktop and mobile applications
+- This monorepo (pnpm + turbo) provides frontend application code
+- Backend code (Rest APIs) is outside this repo but can be changed when necessary
```
thesan: Do we really want to insist to the agents that changing the backend API is an option. IMO this type of decision should come from the human developer. I feel that this instruction would lead to the wrong result most of the time.

LL782: I agree that including this would be an experiment. It was an experiment I was happy to try but I'm also happy to leave it out and add something like this only if necessary

```suggestion
- Backend code (REST APIs) is outside this repo
```

PR #15936, opened 2026-03-31
feat(tooling): sync claude skills with cursor and extract shared docs

VicAlbr:
### ✅ Checklist

- [x] **Covered by automatic tests.** N/A — tooling/config changes only
  - Claude Code skills and Cursor commands now share a single source of truth in `docs/dev/`
  - No functional code changed

### 📝 Description

Syncs `.claude/skills/` with `.cursor/commands|skills/` so both AI tools use the same workflow definitions. Previously, content was duplicated across both systems and getting out of sync.

**What changed:**
- Extract stable reference content (PR template, changeset tables, e2e setup checks, Slack message format) to `docs/dev/` files shared by both systems
- Add missing `create-changeset` and `slack-pr-message` sub-skills to `.claude/skills/` to match cursor's architecture
- Slim down all skill/command files to reference `docs/dev/` instead of embedding duplicate content

### ❓ Context

- **JIRA or GitHub link**: N/A — internal tooling improvement

---


.claude/skills/create-pr/SKILL.md
LL782, line 1: **suggestion: extract body to a file shared with Cursor**

Doesn't this duplicate a lot of what is in Cursor?

How about we aim for vendor specific skill and rule files to be less than 20 lines

.claude/skills/e2e-desktop-onboard/SKILL.md
LL782, line 1: Same comment here about keeping the length of vendor specific files (`.claude/**/*`, `.cursor/**/*`, etc) down to around 20 lines. We should just be providing front-matter and a link to the shared doc.

LL782, line 1: **question: can we keep things specific to e2e-desktop inside `e2e/desktop/.claude/skills`?**

LL782 (commented):
Some quick feedback:
 
- I like some of what has been done to share docs between claude and cursor

- I don't like the length of some of the claude specific files

- I do like the locating of docs files that apply to specific directories, e.g. `e2e/desktop/CLAUDE.md`

.claude/skills/e2e-desktop-onboard/SKILL.md
VicAlbr: hmmm 🤔 idk, I think it's better to have the skills on .claude/skills on the root so it can be used from anywhere in the repo. All e2e commands are used from the root so imo it's better to have the skills there also.

gre-ledger (commented):
i'm ok for the skills parts , and why not for the docs/* even tho I think we should just reinforce/refactor/improve the existing README.md / to not hesitate to introduce more README.md (e.g. for e2e related doc, it must be README in e2e/desktop/*)

for the CLAUDE.md localized to projects, what is the difference with doing this vs having a README.md ?

the CLAUDE.md must try to be way way more minimal, because it will be injected in context of all prompts

VicAlbr:
> the CLAUDE.md must try to be way way more minimal, because it will be injected in context of all prompts

Ok, I will remove the claude.md in the e2e folder

.claude/skills/create-pr/SKILL.md
thesan, line 1: These are implemented as commands on Cursor is it to keep them from getting wrongly invocated by an agent? If so Claude has this [`disable-model-invocation: true`](https://code.claude.com/docs/en/skills#frontmatter-reference). I think we should use this whenever possible to avoid the risk of getting our prompt running stuff we don't want because of a skill we may not even no exist.

WDYT?

.claude/skills/create-changeset/SKILL.md
```diff
@@ -0,0 +1,14 @@
...
+description: Create changesets for the ledger-live monorepo using @changesets/cli. Use when the user asks to add a changeset, prepare a release, or document package changes for a PR.
+---
+
+# Create Changeset
+
+Create a `.changeset/<adjective-noun-verb>.md` file with the correct package names and impact level.
+
+See [/docs/dev/changesets.md](/docs/dev/changesets.md) for:
+- Package name reference table
+- Impact level definitions (minor / major / patch)
```
thesan, line 12: With the current release flow IMO we should just tell it to always use minor

VicAlbr:
I will close this PR. I will do the work in a different way: separate in small PR etc...

PR #15850, opened 2026-03-27
[LWDM] fix: add delegates in tezos getBlock

estrauser-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [x] **Impact of the changes:** 
  1. `src/network/tzkt.ts`:
    • Added getBlockDelegationsPage() - fetches a page of delegation operations at a given block level from TzKT API (/v1/operations/delegations)
    • Added fetchBlockDelegations() - handles pagination to fetch all delegation operations for a block
  2. `src/network/index.ts`:
    • Exported fetchBlockDelegations
  3. `src/logic/getBlock.ts`:
    • Added delegation helper functions:
      • computeDelegationFees() - calculates fees for delegation operations
      • buildDelegationOperations() - creates BlockOperation entries with type "other" and details containing operationType: "DELEGATE" or "UNDELEGATE" plus the delegate address
      • buildBlockTransactionFromDelegation() - builds a BlockTransaction from a delegation operation
    • Updated groupAndMapTransactions() to accept and process delegations
    • Updated getBlock() to fetch delegations in parallel with other data

  The delegation operations are now included in getBlock results with:
  • type: "other" (since they're not value transfers)
  • amount: 0n (delegations don't transfer XTZ)
  • details containing operationType ("DELEGATE" or "UNDELEGATE") and delegate address

### 📝 Description

_Replace this text by a clear and concise description of what this pull request is about and why it is needed. Be sure to explain the problem you're addressing and the solution you're proposing._
_For libraries, you can add a code sample of how to use it._
_For bug fixes, you can explain the previous behaviour and how it was fixed._
_In case of visual features, please attach screenshots or video recordings to demonstrate the changes._

### ❓ Context

- **JIRA or GitHub link**: 
- **ADR link (if any)**: 

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #15983, opened 2026-04-01
[LWDM] feat: [LIVE-26466] add FA2 tokens to list of operations

acewf:
### ✅ Checklist

  - Should be able to list operations tokens with fa2 standard where the tokenId is 0

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-26466

---

### DIFF
--- a/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/listOperations.ts
@@ -19,11 +19,11 @@ function parseCursor(token?: string): { nativeLastId?: number; tokenLastId?: num
     const parsed: unknown = JSON.parse(token);
     if (typeof parsed === "number") return { nativeLastId: parsed };
     if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
-      const output = parsed as { n?: unknown; t?: unknown };
-      return {
-        nativeLastId: typeof output.n === "number" ? output.n : undefined,
-        tokenLastId: typeof output.t === "number" ? output.t : undefined,
-      };
+      const output = parsed as { nativeLastId?: unknown; tokenLastId?: unknown };
+      const nativeLastId =
+        typeof output.nativeLastId === "number" ? output.nativeLastId : undefined;
+      const tokenLastId = typeof output.tokenLastId === "number" ? output.tokenLastId : undefined;
+      return { nativeLastId, tokenLastId };
     }
   } catch {
     // ignore invalid cursor
@@ -54,16 +54,18 @@ export async function listOperations(
 ): Promise<[Operation[], string]> {
   const { nativeLastId, tokenLastId } = parseCursor(token);
 
-  let nativeOptions: AccountsGetOperationsOptions = { limit, sort, "level.ge": minHeight };
-  if (nativeLastId !== undefined) {
-    nativeOptions = { ...nativeOptions, lastId: nativeLastId };
-  }
+  const nativeOptions: AccountsGetOperationsOptions = {
+    limit,
+    sort,
+    "level.ge": minHeight,
+    ...(nativeLastId ? { lastId: nativeLastId } : {}),
+  };
 
   const tokenOptions = {
     limit,
     sort,
     "level.ge": minHeight,
-    ...(tokenLastId !== undefined ? { lastId: tokenLastId } : {}),
+    ...(tokenLastId ? { lastId: tokenLastId } : {}),
   };
 
   const [nativeOps, tokenTransfers] = await Promise.all([
@@ -85,7 +87,12 @@ export async function listOperations(
   const lastNativeOp = limitedNativeOps.at(-1);
   const lastTokenOp = limitedTokenTransfers.at(-1);
   const nextToken =
-    lastNativeOp || lastTokenOp ? JSON.stringify({ n: lastNativeOp?.id, t: lastTokenOp?.id }) : "";
+    lastNativeOp || lastTokenOp
+      ? JSON.stringify({
+          nativeLastId: lastNativeOp?.id,
+          tokenLastId: lastTokenOp?.id,
+        })
+      : "";
 
   const filteredNativeOps = limitedNativeOps
     .filter(op => isAPITransactionType(op) || isAPIDelegationType(op) || isAPIRevealType(op))
@@ -102,11 +109,11 @@ export async function listOperations(
     .map(op => convertOperation(address, op))
     .flat();
 
-  const tokenConverted = limitedTokenTransfers.map(token =>
+  const tokenConverted = limitedTokenTransfers.map(transfer =>
     convertTokenOperation(
       address,
-      token,
-      token.transactionId !== null ? parentMap.get(token.transactionId || 0) : undefined,
+      transfer,
+      transfer.transactionId ? parentMap.get(transfer.transactionId)! : undefined,
     ),
   );
 
@@ -242,8 +249,8 @@ function convertTokenOperation(
 
   const fee = parent
     ? BigInt(parent.storageFee ?? 0) +
-    BigInt(parent.bakerFee ?? 0) +
-    BigInt(parent.allocationFee ?? 0)
+      BigInt(parent.bakerFee ?? 0) +
+      BigInt(parent.allocationFee ?? 0)
     : 0n;
   const feesPayer = parent?.initiator?.address ?? parent?.sender?.address;
 
--- a/libs/coin-modules/coin-tezos/src/network/tzkt.ts
+++ b/libs/coin-modules/coin-tezos/src/network/tzkt.ts
@@ -19,6 +19,13 @@ const BLOCK_PAGE_SIZE = 1000;
 
 const getExplorerUrl = () => coinConfig.getCoinConfig().explorer.url;
 
+const clearUndefined = (obj: Record<string, unknown>) => {
+  const newObj = { ...obj };
+  Object.entries(newObj).forEach(([key, value]) => value === undefined && delete newObj[key]);
+
+  return newObj;
+};
+
 const api = {
   async getBlockCount(): Promise<number> {
     const { data } = await network<number>({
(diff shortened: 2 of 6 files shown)


libs/coin-modules/coin-tezos/src/network/tzkt.ts
```diff
@@ -122,8 +176,37 @@ const api = {
...
+
+    const data = await api.getTokenTransfers(
+      query["level.ge"] as number,
+      query["lastId"],
+      params,
+    );
+    const transactions = await api.getOperationsTransactions(query["level.ge"] || 0, undefined, {
+      "id.in": data.map(t => t.transactionId ?? 0).join(","),
+    });
+
```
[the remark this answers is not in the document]
acewf: this is already dealt with.

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -175,3 +225,63 @@ function convertOperation(
...
+  const type: Operation["type"] = isOut && !isIn ? "OUT" : isIn && !isOut ? "IN" : "FEES";
+
+  const fee = parent
+    ? BigInt(parent.storageFee ?? 0) +
+      BigInt(parent.bakerFee ?? 0) +
+      BigInt(parent.allocationFee ?? 0)
+    : 0n;
+  const feesPayer = parent?.initiator?.address ?? parent?.sender?.address;
+
+  const assetReference = transfer.token.contract.address;
```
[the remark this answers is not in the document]
acewf: Think this potentially might be wrongly implemented on getBlock, or just have another logic.
The logic in generic-alpaca is creates the sub-accounts when it matches the asset reference of the CAL https://github.com/LedgerHQ/ledger-live/blob/LIVE-26466/libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts#L346

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -60,7 +101,16 @@ export async function listOperations(
...
+      address,
+      token,
+      token.transactionId !== null ? parentMap.get(token.transactionId || 0) : undefined,
+    ),
+  );
+
+  const sortedOperations = [...filteredNativeOps, ...tokenConverted].sort(
     (a, b) => b.tx.date.getTime() - a.tx.date.getTime(),
   );
   return [sortedOperations, nextToken];
```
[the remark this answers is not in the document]
acewf: we already have code to limit for the native and tokens, not sure is desired to have the same limit on the object that combines both.

libs/coin-modules/coin-tezos/src/network/tzkt.ts
```diff
@@ -106,6 +134,32 @@ const api = {
...
+      limit: BLOCK_PAGE_SIZE,
+      "sort.asc": "id",
+      ...args,
+    };
+    if (cursor !== undefined) params["offset.cr"] = cursor;
+    const { data } = await network<APITokenTransfer[]>({
+      url: `${getExplorerUrl()}/v1/tokens/transfers`,
+      params,
+    });
+    return data;
```
semeano, line 165: Can you refactor this into one single function, so we don't repeat the code?

libs/ledger-wallet-framework/src/account/support.ts
```diff
@@ -33,12 +29,10 @@ export const shouldShowNewAccount = (
 };
 
 export const getReceiveFlowError = (
```
semeano, line 31: Where is this function being used? Does it make sense to keep it given that it only returns `undefined`?

acewf: This function is being called in the the desktop and mobile apps, it was set some years ago. Don't think removing it should be part of this PR.
There is no clear indication this should be used in the future or not, and maybe would be better just coin-integration create a story to remove it if it makes sense.

libs/coin-modules/coin-tezos/src/network/tzkt.ts
acewf: i was going to do initially, but there is some parallel work with the vault team, where we are also repeating the code. 
Think is better to coordinate with this parallel work, to ensure we could refactor all functions to use helper functions instead of having the repetition.

semeano: I'm sorry, I'm not able to understand. The vault team is working on the same feature or similar feature? Is the code on their side new, or it already existed?
I don't understand why that could be a blocker into refactoring this.

libs/ledger-wallet-framework/src/account/support.ts
semeano: Ok, I'm tagging @hedi-edelbloute so this isn't forgotten.

hedi-edelbloute: Hi, indeed, let's not edit this method if it's not required in the scope of this PR please

libs/coin-modules/coin-tezos/src/network/tzkt.ts
acewf: there were changes even on last week on this PR https://github.com/LedgerHQ/ledger-live/pull/15850.
was just trying to avoid more conflicts for now, since we aren't tracking each others work.

We already have the `network` methods as an abstraction layer.
To refactor to have less repetition has you said, only if we move the arguments defined each method to its initial signature call, which in the case of for example the method `getBlockTokenTransfersPage` not even aware were is being used.

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -32,23 +52,49 @@ export async function listOperations(
     minHeight,
   }: { limit?: number; token?: string; sort: "Ascending" | "Descending"; minHeight: number },
 ): Promise<[Operation[], string]> {
-  let options: AccountsGetOperationsOptions = { limit, sort, "level.ge": minHeight };
-  if (token) {
-    options = { ...options, lastId: JSON.parse(token) };
+  const { nativeLastId, tokenLastId } = parseCursor(token);
+
+  let nativeOptions: AccountsGetOperationsOptions = { limit, sort, "level.ge": minHeight };
```
henri-ly: [nitpick] I would like to have an immutable object at this level if not needed, wdyt of something like : 

```
const nativeOptions: AccountsGetOperationsOptions = {
  limit,
  sort,
  "level.ge": minHeight,
  ...(nativeLastId !== undefined && { lastId: nativeLastId }),
};
```

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
```diff
@@ -32,23 +52,49 @@ export async function listOperations(
...
+  if (nativeLastId !== undefined) {
+    nativeOptions = { ...nativeOptions, lastId: nativeLastId };
   }
-  const operations = await tzkt.getAccountOperations(address, options);
+
+  const tokenOptions = {
+    limit,
+    sort,
+    "level.ge": minHeight,
+    ...(tokenLastId ? { lastId: tokenLastId } : {}),
```
henri-ly, line 68: which you did here. haha, nice

libs/coin-modules/coin-tezos/src/network/tzkt.ts
```diff
@@ -89,6 +90,33 @@ const api = {
...
+   */
+  async getOperationsTransactions(
+    level: number,
+    cursor?: number,
+    args: Record<string, unknown> = {},
+  ): Promise<APITransactionType[]> {
+    // "sort.asc": "id" guarantees forward progress for cursor-based paging (offset.cr).
+    // Without an explicit sort the API default may be descending, which would cause the
+    // cursor to go backwards and produce duplicates or an infinite loop.
+    Object.keys(args).forEach(key => args[key] === undefined && delete args[key]);
```
henri-ly: [question] I'm kind of confuse about what args is doing ? and what is this code is about ? is it possible to have more implicit name than args ?

libs/coin-modules/coin-tezos/src/network/tzkt.ts
```diff
@@ -106,6 +134,32 @@ const api = {
...
+   * https://api.tzkt.io/#operation/Tokens_GetTokenTransfers
+   */
+  async getTokenTransfers(
+    level: number,
+    cursor?: number,
+    args: Record<string, unknown> = {},
+  ): Promise<APITokenTransfer[]> {
+    // Same rationale as getBlockTransactionsPage: explicit ascending sort keeps the
+    // offset.cr cursor advancing forward regardless of the API's default ordering.
+    Object.keys(args).forEach(key => args[key] === undefined && delete args[key]);
```
henri-ly: Same here

libs/coin-modules/coin-tezos/src/network/tzkt.ts
acewf: the idea initially was to have this on a more generic method, but because of some work parallel work with other teams, i didn't want to refactor this file too much.

the args here are just query parameters, they differ depending on the endpoint, but a example of them you can check here https://api.tzkt.io/#operation/Tokens_GetTokenBalances

ok i will change the name for tzktApiParams, maybe it becomes more clear.

About what is going on specifically in this line, is we might pass down variables with undefined values, but we should clear this "empty" values so we don't end up with a query parameter like `?token.id=undefined&account=tz1kasda`

libs/coin-modules/coin-tezos/src/network/tzkt.ts
acewf: hopes this one https://github.com/LedgerHQ/ledger-live/pull/15983/changes#r3052985518 explains enough.

libs/coin-modules/coin-tezos/src/logic/listOperations.ts
acewf: thanks for spotting this.

PR #16002, opened 2026-04-02
[LWDM] FULL EXPLO: feat(live-common): defer load of coin bridges

gre-ledger:
### ✅ Checklist

  - `getAccountBridge`, `getCurrencyBridge` and all related helpers in `bridge/index.ts` are now `async`.
  - All call sites across desktop, mobile, CLI, platform/wallet-api layers updated to `await` these calls.
  - No changes to coin module internals — individual `coin-*` packages are untouched.

### 📝 Description

**Problem:** [LIVE-28411](https://ledgerhq.atlassian.net/browse/LIVE-28411)

The coin module registry (`coin-modules/loaders.ts`) already existed on `develop` with lazy loading — but using synchronous `require()` calls inside arrow functions. This has two compounding issues:

1. **CJS vs ESM duplication**: `require()` forces the CJS code path. When the rest of the app goes through ESM (rspack prefers `lib-es/`), the same coin module ends up bundled twice — once as CJS, once as ESM — inflating the bundle.
2. **No real code splitting**: bundlers cannot split on `require()` calls the way they can on `import()`, so all coin families still land in the same chunk.

**Solution:**

Replace every `require()` in `loaders.ts` with a dynamic `import()`:

```diff
- loadSetup: () => require("../families/bitcoin/setup"),
- loadTransaction: () => require("@ledgerhq/coin-bitcoin/transaction").default,
+ loadSetup: () => import("../families/bitcoin/setup.js"),
+ loadTransaction: () => import("@ledgerhq/coin-bitcoin/transaction").then(m => m.default),
```

Because `import()` is truly async, the registry's `load*ForFamily()` functions now return real Promises, which propagates upward: `getAccountBridgeByFamily` → `getAccountBridge` / `getCurrencyBridge` in `bridge/index.ts` all become `async`. All call sites are updated accordingly (`await` / `.then()`).

The in-memory cache in `registry.ts` means the `import()` only fires once per family — subsequent calls resolve immediately.

**Why this matters:**

- Fixes CJS/ESM duplication: each coin family is now loaded once, through the ESM path, reducing bundle size.
- Boot time relief: JS parsing and module initialisation for unused coin families is deferred to first use, not startup.
- Enables building lightweight consumers (e.g. a single-family CLI) that only pay for what they register.

### ❓ Context

- **JIRA**: [LIVE-28411](https://ledgerhq.atlassian.net/browse/LIVE-28411)
- **Previous step** (landed on `develop`): introduced the registry + `loaders.ts` with `require()` — PR #16031
- **Next step**: apply the same `import()` treatment to `renderer/families/` on desktop

---

[LIVE-28411]: https://ledgerhq.atlassian.net/browse/LIVE-28411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-28411]: https://ledgerhq.atlassian.net/browse/LIVE-28411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/bridge/useBridgeTransaction.ts
```diff
@@ -238,12 +261,20 @@ const useBridgeTransaction = <T extends Transaction = Transaction>(
     dispatch,
   ] = useReducer(reducer as Reducer<T>, undefined, makeInit<T>(optionalInit));
   const setAccount = useCallback(
-    (account, parentAccount) =>
+    async (account, parentAccount) => {
+      let bridge: AccountBridge<T> | undefined;
+      try {
+        bridge = await getAccountBridge(account, parentAccount);
```
gre-ledger, line 280: unsatisfied area of improvement: I don't know how better could do this, but we need to minimize the impact of changing this hook and all usages of it.
that said, it still need to load the bridge somehow. so i've not yet found a satisfied solution on this.

PR #16031, opened 2026-04-02
[LWDM] feat(live-common): defer load of coins: sync require() registry (replace generated/ barrel files)

gre-ledger:
### ✅ Checklist

  - mobile e2e smoke pass https://github.com/LedgerHQ/ledger-live/actions/runs/24184417897
  - `@ledgerhq/live-common`: internal refactor only, all public APIs unchanged
  - Apps (`cli`, `desktop`, `mobile`): must call `registerAllCoins()` at startup (already done in this PR)
  - No functional change for end users

### 📝 Description

Replaces the 12 generated barrel files under `libs/ledger-live-common/src/generated/` with a synchronous `require()`-based coin module registry.

**Problem:** The generated barrel files eagerly imported every coin family at startup, pulling the entire coin ecosystem into the initial bundle regardless of which currencies were actually used.

**Solution:** A lazy `require()` registry in `src/coin-modules/`:
- `registry.ts` — `Map<family, CoinModuleLoader>` with sync `load*ForFamily()` accessors
- `loaders.ts` — 29 families registered via `require()` (lazy in Node.js, deferred by bundlers)
- `load-all-coins.ts` — `registerAllCoins()` entry point called once at app startup
- `bot/allSpecs.ts` — hand-maintained bot specs manifest (kept separate to avoid pulling spec deps into CLI/web builds)

All public APIs (`getCurrencyBridge`, `getAccountBridge`, `fromTransactionRaw`, etc.) remain synchronous — **non-breaking** relative to `develop`.

## Better boot performance

- Desktop Unit Tests on CI: 18min -> 12min
- Desktop boot time: 1.4s -> 1.0s

### ⚠️ Known trade-off: desktop renderer bundle size (+2.7 MB)

Using `require()` inside arrow functions defers module evaluation (coin family `createBridges()`, `createResolver()`, etc. no longer run at startup — only when a family is first accessed). This is the **intended startup performance benefit** of this PR: on Mac M1, **TS-load improves from 1.4 to 1.0 seconds.**

However, `require()` inside functions prevents rspack's scope hoisting (module concatenation), causing the renderer bundle to grow by ~2.7 MB (26.6 MB → 29.3 MB). This is a **known and accepted trade-off** for this PR.

**main root causes:**

Screenshot 2026-04-07 at 12 49 07

**Future fix:** Replacing `require()` with async `import()` (the next planned step per `FUTURE.md` in loaders.ts) will use true rspack code splitting — coin family modules will be absent from the main bundle entirely and loaded on demand, which will restore and likely improve bundle size.

### ❓ Context

- **JIRA or GitHub link**: LIVE-28411


libs/ledger-live-common/src/transaction/deviceTransactionConfig.ts
```diff
@@ -1,19 +1,32 @@
...
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_stacks } from "@ledgerhq/coin-stacks/bridge/deviceTransactionConfig";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_polkadot } from "@ledgerhq/coin-polkadot/bridge/deviceTransactionConfig";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_tron } from "@ledgerhq/coin-tron/bridge/deviceTransactionConfig";
+
+export type ExtraDeviceTransactionField =
+  | ExtraDeviceTransactionField_casper
+  | ExtraDeviceTransactionField_filecoin
+  | ExtraDeviceTransactionField_stacks
+  | ExtraDeviceTransactionField_polkadot
+  | ExtraDeviceTransactionField_tron;
```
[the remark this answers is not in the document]
gre-ledger: believe me Copilot, we don't want to go back to generated approach.

mcayuelas-ledger (dismissed):

apps/ledger-live-mobile/src/families/canton/Onboard/__integrations__/Onboard.integration.test.tsx
```diff
@@ -128,15 +122,6 @@ describe("Canton onboarding integration", () => {
...
-
-    coinConfig.setCoinConfig(() => ({
-      status: { type: "active" },
-      networkType: "devnet",
-      gatewayUrl: CANTON_DEVNET_GATEWAY,
-      nodeId: CANTON_DEVNET_NODE_ID,
-      useGateway: true,
-      nativeInstrumentId: "Amulet",
-    }));
   });
```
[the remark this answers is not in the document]
gre-ledger: code is redondant with now doing

```
import { liveConfig } from "@ledgerhq/live-common/config/sharedConfig";
LiveConfig.setConfig(liveConfig);
```

which set the defaults

libs/ledger-live-common/src/flows/send/amount/utils/errors.ts
```diff
@@ -1,4 +1,4 @@
-import type { TransactionStatus } from "@ledgerhq/live-common/generated/types";
+import type { TransactionStatus } from "@ledgerhq/live-common/coin-modules/transaction-types";
```
LL782: **nit: since we're updating all this import paths anyway lets introduce some consistency**

In many of the updated files we see relative paths to import `TransactionStatus`, above we a self-referencing import path. Let's pick one convention and stick this it?



```suggestion
import type { TransactionStatus } from "../../../../coin-modules/transaction-types";
```

libs/ledger-live-common/scripts/build-ts.mjs
Justkant, line 1: When we go full import and make sure we don't need tsc we could also drop the script and only build once
Coupled with the ts project references it should become pretty fast to build

libs/ledger-live-common/src/transaction/deviceTransactionConfig.ts
```diff
@@ -1,19 +1,32 @@
...
+import type { Transaction, TransactionStatus } from "../coin-modules/transaction-types";
 import { getMainAccount } from "../account";
 import type { Account, AccountLike } from "@ledgerhq/types-live";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_casper } from "@ledgerhq/coin-casper/bridge/deviceTransactionConfig";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_filecoin } from "@ledgerhq/coin-filecoin/bridge/deviceTransactionConfig";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_stacks } from "@ledgerhq/coin-stacks/bridge/deviceTransactionConfig";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_polkadot } from "@ledgerhq/coin-polkadot/bridge/deviceTransactionConfig";
+import { ExtraDeviceTransactionField as ExtraDeviceTransactionField_tron } from "@ledgerhq/coin-tron/bridge/deviceTransactionConfig";
+
+export type ExtraDeviceTransactionField =
```
LL782: Does this need to be exported? It's making me harder to understand where it's used


```suggestion
type ExtraDeviceTransactionField =
```

libs/ledger-live-common/src/coin-modules/types.ts
```diff
@@ -0,0 +1,89 @@
...
+// Parameters are `any` because each family uses its own transaction/raw types at the loader boundary.
+// Method syntax enables bivariant param checking so family-specific implementations are assignable.
+export type TransactionModule = {
+  fromTransactionRaw(raw: any): TransactionCommon;
+  toTransactionRaw(tx: any): Record<string, unknown>;
+  formatTransaction(tx: any, account: Account): string;
+  fromTransactionStatusRaw?(raw: any): TransactionStatusCommon;
+  toTransactionStatusRaw?(status: any): Record<string, unknown>;
+  formatTransactionStatus?(tx: any, status: any, mainAccount?: Account): string;
+};
```
LL782, line 31: **note: I don't yet understand why we have to resort to any here**

I intend to look closer at this file and at `libs/ledger-live-common/src/coin-modules/transaction-types.ts`

libs/ledger-live-common/src/coin-modules/registry.ts
```diff
@@ -0,0 +1,63 @@
...
+  family: string,
+): DeviceTransactionConfigFn => getLoader(family).loadDeviceTxConfig!();
+
+export const loadWalletApiAdapterForFamily = (
+  family: string,
+): WalletApiAdapterModule => getLoader(family).loadWalletApiAdapter!();
+
+export const loadPlatformAdapterForFamily = (
+  family: string,
+): PlatformAdapterModule => getLoader(family).loadPlatformAdapter!();
```
Justkant: De we really want to keep `!()` instead of `?.()` ?
I'm wondering if it could be called on a family that doesn't have the function and fail

The only drawback is that you need to handle the potential undefined on the caller side

philipptpunkt (dismissed):

libs/ledger-live-common/src/exchange/swap/hooks/useSwapTransaction.test.ts
```diff
@@ -7,7 +7,7 @@ import { AmountRequired, NotEnoughGas, NotEnoughGasSwap } from "@ledgerhq/errors
 
 import { useFromAmountStatusMessage } from "./useSwapTransaction";
 import { Result } from "../../../bridge/useBridgeTransaction";
-import { Transaction } from "../../../generated/types";
+import { Transaction } from "../../../coin-modules/transaction-types";
```
Justkant: nitpick: We could make sure they are all using `import type`

LL782 (dismissed):
Overall it looks sane to me. I'm happy to approve. Note that I haven't been able to go right to the bottom of everything but I trust your decision making on a the overall approach and the QA and checks process for everything else.

I've left two minor comments which are not at all blocking.

I'm also curious about the `coin-modules/types` and will continue to look into this, but that doesn't need to hold up this PR

libs/ledger-live-common/src/hw/actions/app.ts
```diff
@@ -425,7 +425,12 @@ function inferCommandParams(appRequest: AppRequest): ConnectAppRequest {
   if (account) {
     derivationMode = account.derivationMode;
     derivationPath = account.freshAddressPath;
-    const m = perFamilyAccount[account.currency.family];
+    let m;
+    try {
+      m = loadAccountModuleForFamily(account.currency.family);
+    } catch {
+      m = null;
+    }
```
Justkant: Here we have a try catch because of the `!()` in the function but we could already return undefined or null with a `?.()` in the load function

libs/ledger-live-common/src/platform/converters.ts
```diff
@@ -95,7 +95,12 @@ export const getPlatformTransactionSignFlowInfos = (
 } => {
   const liveFamily = FAMILIES_MAPPING_PLATFORM_TO_LL[platformTx.family] ?? platformTx.family;
 
-  const familyModule = byFamily[liveFamily];
+  let familyModule;
+  try {
+    familyModule = loadPlatformAdapterForFamily(liveFamily);
+  } catch {
+    familyModule = null;
+  }
```
Justkant: Same as the others here

libs/ledger-live-common/src/transaction/deviceTransactionConfig.ts
```diff
@@ -1,19 +1,32 @@
...
   parentAccount: Account | null | undefined;
   transaction: Transaction;
   status: TransactionStatus;
 }): Promise<Array<DeviceTransactionField>> {
   const mainAccount = getMainAccount(arg.account, arg.parentAccount);
-  const f = perFamily[mainAccount.currency.family];
-  if (!f) return [];
+  const family = mainAccount.currency.family;
+  if (!hasDeviceTxConfigForFamily(family)) return [];
+  const f = loadDeviceTxConfigForFamily(family);
```
Justkant, line 29: We could keep it as before if loadDevice would return undefined if not present and we could remove the hasDeviceTxConfigForFamily

libs/ledger-live-common/src/wallet-api/converters.ts
```diff
@@ -107,7 +107,12 @@ export const getWalletAPITransactionSignFlowInfos: GetWalletAPITransactionSignFl
   const liveFamily =
     FAMILIES_MAPPING_WAPI_TO_LL[walletApiTransaction.family] ?? walletApiTransaction.family;
 
-  const familyModule = byFamily[liveFamily];
+  let familyModule;
+  try {
+    familyModule = loadWalletApiAdapterForFamily(liveFamily);
+  } catch {
+    familyModule = null;
+  }
```
Justkant: Same as the others here

libs/ledger-live-common/.unimportedrc.json
```diff
@@ -338,6 +338,56 @@
     "**/fixtures/**"
   ],
   "ignoreUnimported": [
+    "src/generated/types.ts",
```
Justkant: Do we still keep generated here ?

Justkant (commented):
I think we should update the loaders to return undefined and we might want to make sure we use import type for the types import we updated from generated

libs/ledger-live-common/src/coin-modules/types.ts
Justkant: Happy to see if you can find a good solution, the transaction type has always been hard to work with because it mixes all the transaction types from all coin modules IIRC

LL782: Perhaps I'm being naive but it seems like we could do this: https://github.com/LedgerHQ/ledger-live/pull/16195/changes

I only ran `typecheck` within `live-common`. Perhaps the issue are with consumers that I haven't looked into.

Let me know your thoughts

philipptpunkt (approved):

mcayuelas-ledger (approved):

Justkant (approved):

PR #16068, opened 2026-04-06
[LIVE-26484] add craft transaction support for FA2 tokens

acewf:
### ✅ Checklist

  - Able to send FA2 tokens on the tezos network

### 📝 Description

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-26484

---

### DIFF
--- a/libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
@@ -98,7 +98,6 @@ export async function craftTransaction(
   }
 
   let type: TransactionType;
-  console.log("transaction.type", transaction.type);
   switch (transaction.type) {
     case "send": {
       type = "OUT";
--- a/libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
@@ -125,7 +125,7 @@ export async function craftTransaction(
               {
                 to_: transaction.recipient,
                 token_id: transaction.tokenId,
-                amount: Number(transaction.amount),
+                amount: transaction.amount,
               },
             ],
           },
--- a/libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
@@ -19,10 +19,12 @@ export async function craftTransaction(
     counter?: number;
   },
   transaction: {
-    type: "send" | "delegate" | "undelegate";
+    type: "send" | "delegate" | "undelegate" | "send_token";
     recipient: string;
     amount: bigint;
     fee: TransactionFee;
+    contractAddress?: string;
+    tokenId?: number;
   },
   publicKey?: {
     publicKey: string;
@@ -96,6 +98,7 @@ export async function craftTransaction(
   }
 
   let type: TransactionType;
+  console.log("transaction.type", transaction.type);
   switch (transaction.type) {
     case "send": {
       type = "OUT";
@@ -105,6 +108,37 @@ export async function craftTransaction(
         destination: transaction.recipient,
         source: address,
         counter: (counter + 1 + contents.length).toString(),
+        ...transactionFees,
+      });
+      break;
+    }
+    case "send_token": {
+      if (!transaction.contractAddress || transaction.tokenId === undefined) {
+        throw new Error("FA2 transfer requires contractAddress and tokenId");
+      }
+      type = "OUT";
+      const tokenContract = await tezosToolkit.contract.at(transaction.contractAddress);
+      const transferParams = tokenContract.methods
+        .transfer([
+          {
+            from_: address,
+            txs: [
+              {
+                to_: transaction.recipient,
+                token_id: transaction.tokenId,
+                amount: Number(transaction.amount),
+              },
+            ],
+          },
+        ])
+        .toTransferParams({ mutez: true });
+      contents.push({
+        kind: OpKind.TRANSACTION,
+        source: address,
+        destination: transaction.contractAddress,
+        amount: "0",
+        counter: (counter + 1 + contents.length).toString(),
+        parameters: transferParams.parameter,
         ...transactionFees,
       });
       break;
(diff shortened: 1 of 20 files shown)


libs/coin-modules/coin-tezos/src/logic/validateIntent.ts
```diff
@@ -49,13 +52,17 @@ function validateBasicSendParams(intent: TransactionIntent): Record<string, Erro
...
-  // send max not allowed on delegated accounts (must undelegate acc first)
-  if (intent.type === "send" && intent.useAllAmount) {
-    if (senderInfo.type === "user" && senderInfo.delegate?.address) {
+  // send max not allowed on delegated accounts (must undelegate acc first); native XTZ only
+  if (
+    intent.type === "send" &&
+    intent.useAllAmount &&
+    resolveTezosOperationMode(intent.type, intent.asset) === "send"
+  ) {
+    if (senderInfo.delegate?.address) {
```
semeano: Why not add this `if` to the parent `if`?

libs/coin-modules/coin-tezos/src/types/bridge.fixture.ts
```diff
@@ -65,6 +65,8 @@ export function createFixtureTransaction(tx?: Partial<Transaction>): Transaction
     storageLimit: tx?.storageLimit || undefined,
     estimatedFees: tx?.estimatedFees || undefined,
     taquitoError: tx?.taquitoError || undefined,
+    contractAddress: tx?.contractAddress,
+    tokenId: tx?.tokenId,
```
semeano, line 69: I'm not sure if this makes sense but maybe it should be to prevent values to be `null`.

```suggestion
    contractAddress: tx?.contractAddress || undefined,
    tokenId: tx?.tokenId || undefined,
```

semeano (dismissed):

libs/coin-modules/coin-tezos/src/logic/validateIntent.ts
acewf: was already 2 if statements, but yep don't see a reason to have 2 if statements.

libs/coin-modules/coin-tezos/src/types/bridge.fixture.ts
acewf: the compiler should give you an error, because the type of this properties aren't nullable.

semeano (dismissed):

libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
```diff
@@ -109,6 +111,37 @@ export async function craftTransaction(
...
+      const tokenContract = await tezosToolkit.contract.at(transaction.contractAddress);
+      const transferParams = tokenContract.methods
+        .transfer([
+          {
+            from_: address,
+            txs: [
+              {
+                to_: transaction.recipient,
+                token_id: transaction.tokenId,
+                amount: Number(transaction.amount),
```
cted-ledger: could we lose precision here for amounts above 2^53? With FA2 tokens at 18 decimals that seems reachable in practice. Does Taquito accept a string instead ?

libs/coin-modules/coin-tezos/src/logic/estimateFees.ts
```diff
@@ -108,15 +118,40 @@ export async function estimateFees({
...
+        const tokenContract = await tezosToolkit.contract.at(transaction.contractAddress);
+        const transferParams = tokenContract.methods
+          .transfer([
+            {
+              from_: account.address,
+              txs: [
+                {
+                  to_: transaction.recipient,
+                  token_id: transaction.tokenId,
+                  amount: Number(amount),
```
cted-ledger: Same question about Number(amount) as in craftTransaction.ts:129

libs/coin-modules/coin-tezos/src/network/tzkt.ts
cted-ledger, line 194: "token.tokenId": "0"

Is it intentional to limit to tokenId 0 transfers only? Since the rest of the PR handles arbitrary token IDs in craft/estimate/validate, I'm wondering if tokens with tokenId > 0 would be missing from the operation history. If it's temporary, maybe worth a TODO ?

acewf: the epic https://ledgerhq.atlassian.net/browse/LIVE-25364
says `Limited Scope: Support is restricted to single-asset transfers (token_id: 0).`

on here this QueryParams are specifically to fetch data related to intended FA2 tokens that are of this tokenId type.

acewf: Btw also this already included in the develop branch at this point, so not part of this PR.

libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
acewf: docs here https://taquito.io/docs/24.2.0/making_transfers/#transfer-from-an-implicit-tz1-address-to-a-tz1-address

about the types accepted isn't clear on the documentation, and looking into the type definition we have the transfer args types as 
```
Record<string, (args?: any) 
```

the property `transaction.amount` type is bigint, and then is casted to number.
Let me run some tests if it works using the bigint.

acewf: for small amounts, UI only supports this values 0.000001, for big amounts is hard to tests. 

But seems we can get rid of casting at least, i will update this.

libs/coin-modules/coin-tezos/src/logic/estimateFees.ts
acewf: explanation is the same, removed the casting and the type is bigint for the amount.

semeano (approved):

libs/coin-modules/coin-tezos/src/logic/craftTransaction.ts
cted-ledger: Great, the fix is correct.
A semantic note: what you removed wasn't just a "cast" but a real number conversion. The difference matters here because Number(bigint) silently rounds values above 2^53 - 1 to the nearest float, so you'd get wrong amounts without any error. A cast would just reinterpret the type. Worth keeping in mind for future reviews!

cted-ledger (approved):

PR #16080, opened 2026-04-07
[LWDM] fix(coin-cosmos): make cryptoFactory resilient to uninitialized coinConfig

gre-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** The existing `Delegations.test.tsx` desktop unit tests cover the fixed code path.
  - `coin-cosmos` only — `cryptoFactory` no longer throws when called before bridge setup is complete
  - No behaviour change when coinConfig is initialized normally (production hot path unchanged)

### 📝 Description

**Context:** this fix is a prerequisite for #16031 (lazy require() registry — deferred bridge loading). That PR moves family `setup.ts` loading from eager (barrel-file import time) to lazy (first `getCurrencyBridge()` call). This change is needed to unblock it and more broadly to prepare for deferred bridge loading.

**The bug:** `cryptoFactory()` in `chain/chain.ts` is imported directly by the `Delegation` UI component and called on every render. It calls `cosmosCoinConfig.getCoinConfig()`, which throws `MissingCoinConfig` if `setCoinConfig()` has not been called yet. `setCoinConfig()` is only called inside `createBridges()`, which runs when `setup.ts` loads lazily — i.e. on first `getCurrencyBridge("cosmos")`. If the Delegation component renders before any bridge operation (e.g. from persisted store at app start), it crashes.

There is also a secondary caching bug: the previous code wrote to the cache *before* calling `getCoinConfig()`, so after a throw the cache entry was permanently stale and coinConfig enrichment was silently skipped on all subsequent calls.

**Fix:** use a local variable in the `switch` block, wrap `getCoinConfig()` in a try/catch. On success: cache the enriched result as before. On `MissingCoinConfig`: return base chain defaults without caching, so the next call retries enrichment once bridges are loaded.

### ❓ Context

- **JIRA or GitHub link**: #16031

### DIFF
--- a/libs/coin-modules/coin-cosmos/src/chain/chain.ts
+++ b/libs/coin-modules/coin-cosmos/src/chain/chain.ts
@@ -25,74 +25,79 @@ const cosmosChainParams: { [key: string]: CosmosBase } = {};
 export default function cryptoFactory(currencyId: string): CosmosBase {
   currencyId = currencyId === "osmosis" ? "osmo" : currencyId;
   if (!cosmosChainParams[currencyId]) {
+    let chain: CosmosBase;
     switch (currencyId) {
       case "osmo":
-        cosmosChainParams[currencyId] = new Osmosis();
+        chain = new Osmosis();
         break;
       case "cosmos":
-        cosmosChainParams[currencyId] = new Cosmos();
+        chain = new Cosmos();
         break;
       case "axelar":
-        cosmosChainParams[currencyId] = new Axelar();
+        chain = new Axelar();
         break;
       case "binance_beacon_chain":
-        cosmosChainParams[currencyId] = new BinanceBeaconChain();
+        chain = new BinanceBeaconChain();
         break;
       case "desmos":
-        cosmosChainParams[currencyId] = new Desmos();
+        chain = new Desmos();
         break;
       case "dydx":
-        cosmosChainParams[currencyId] = new Dydx();
+        chain = new Dydx();
         break;
       case "nyx":
-        cosmosChainParams[currencyId] = new Nyx();
+        chain = new Nyx();
         break;
       case "persistence":
-        cosmosChainParams[currencyId] = new Persistence();
+        chain = new Persistence();
         break;
       case "quicksilver":
-        cosmosChainParams[currencyId] = new Quicksilver();
+        chain = new Quicksilver();
         break;
       case "secret_network":
-        cosmosChainParams[currencyId] = new SecretNetwork();
+        chain = new SecretNetwork();
         break;
       case "stargaze":
-        cosmosChainParams[currencyId] = new Stargaze();
+        chain = new Stargaze();
         break;
       case "stride":
-        cosmosChainParams[currencyId] = new Stride();
+        chain = new Stride();
         break;
       case "umee":
-        cosmosChainParams[currencyId] = new Umee();
+        chain = new Umee();
         break;
       case "coreum":
-        cosmosChainParams[currencyId] = new Coreum();
+        chain = new Coreum();
         break;
       case "injective":
-        cosmosChainParams[currencyId] = new Injective();
+        chain = new Injective();
         break;
       case "mantra":
-        cosmosChainParams[currencyId] = new Mantra();
+        chain = new Mantra();
         break;
       case "crypto_org":
-        cosmosChainParams[currencyId] = new CryptoOrg();
+        chain = new CryptoOrg();
         break;
       case "xion":
-        cosmosChainParams[currencyId] = new Xion();
+        chain = new Xion();
         break;
       case "zenrock":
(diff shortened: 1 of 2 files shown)


libs/coin-modules/coin-cosmos/src/chain/chain.ts
```diff
@@ -25,74 +25,79 @@ const cosmosChainParams: { [key: string]: CosmosBase } = {};
...
-    const coinConfig = cosmosCoinConfig.getCoinConfig(currencyId);
-    if (coinConfig) {
-      cosmosChainParams[currencyId] = { ...cosmosChainParams[currencyId], ...coinConfig };
+    try {
+      const coinConfig = cosmosCoinConfig.getCoinConfig(currencyId);
+      cosmosChainParams[currencyId] = coinConfig ? ({ ...chain, ...coinConfig } as CosmosBase) : chain;
+    } catch {
+      // coinConfig not yet initialized (bridges not loaded) — return defaults without caching
+      // so enrichment is retried on the next call once bridges are set up
+      return chain;
```
Justkant, line 100: We might want to log here if the fallback happens because we should almost never end up in this situation unless we don't have internet or firebase is down

libs/coin-modules/coin-cosmos/src/chain/chain.ts
```diff
@@ -25,74 +25,79 @@ const cosmosChainParams: { [key: string]: CosmosBase } = {};
...
       default:
         throw new Error(`${currencyId} is not supported`);
     }
 
-    const coinConfig = cosmosCoinConfig.getCoinConfig(currencyId);
-    if (coinConfig) {
-      cosmosChainParams[currencyId] = { ...cosmosChainParams[currencyId], ...coinConfig };
+    try {
+      const coinConfig = cosmosCoinConfig.getCoinConfig(currencyId);
+      cosmosChainParams[currencyId] = coinConfig ? ({ ...chain, ...coinConfig } as CosmosBase) : chain;
```
Justkant, line 96: I'm really wondering if spreading a class instance is safe here, looks like it worked so far but we might want to rework this at some point, more for coin-inte to look at

Justkant (approved):
Left some comments but I think we can start with this change

libs/coin-modules/coin-cosmos/src/chain/chain.ts
gre-ledger: it would happen if something imports this logic before loading the bridge first. on the other branch (where we started defer loading the bridges on demand) apparently it was happening in a test due to it rendering a component without first loading the bridge.

Here is brainstorm I got with help of Claude Code when I investigated into this:



> "Is it a race condition?"
>                            
>   No — it's a deterministic load-order dependency. cryptoFactory is called synchronously during React render (Delegation/index.tsx:127). If no bridge
>   operation has triggered loadSetupForFamily("cosmos") yet, setCoinConfig was never called, so getCoinConfig throws.
>                                                                                                                                                           
>   "Coin module initialization setting coinConfig — even if it was settled before? A bug to fix?"                                                            
>                                                                                                                                                           
>   Yes, real bug. In bridge/index.ts:92: cosmosCoinConfig.setCoinConfig(coinConfig) runs unconditionally every time setup.ts loads. The buildCoinConfig      
>   singleton has no guard — it always overwrites. If a test set setCoinConfig(mockFn) in beforeAll, and then a component render triggers lazy              
>   loadSetupForFamily, the mock is overwritten by the real getCurrencyConfiguration-based function. That's what happened with the Canton mobile test (which  
>   we fixed by removing the redundant setCoinConfig from the test, and relying on LiveConfig.setConfig(liveConfig) instead).                               
>                                                                                                                                                             
>   "Can this happen in production, besides tests?"                                                                                                         
>                                                                                                                                                           
>   Yes. If a user opens the app with cosmos accounts in persisted storage:                                                                                   
>   1. Store hydrates from localStorage                                                                                                                     
>   2. Account page renders immediately                                                                                                                       
>   3. Delegation renders → cryptoFactory("cosmos") → getCoinConfig throws (because bridges are still lazy-unloaded)                                        
>                                                                                                                                                             
>   The old barrel files prevented this by loading setup.ts at import time as a side effect, setting setCoinConfig before anything could render.            
>                                                                                                                                                             
>   The fix (chain.ts):                                                                                                                                       
>                                                                                                                                                             
>   Use a local chain variable in the switch (not assigned to the cache yet), then try to apply coinConfig enrichment. On failure: return base defaults       
>   without caching — so the next call retries enrichment once bridges are set up. Once coinConfig is available, cache the enriched result normally.          
>                                                                                                                                                             
>   This is:                        
>   - Safe defaults when bridges aren't loaded yet (hardcoded chain params are already correct)                                                               
>   - Correct enrichment once bridges load (retried on next call)                                                                                           
>   - No throw in production or tests                                                                                                                       
>   - No eagerly loading all setups

PR #16089, opened 2026-04-07
[LWDM] refactor(live-common): lazy-load alpaca coin API and signer modules via require()

gre-ledger:
### ✅ Checklist

  - `live-common` only: no change to public API or runtime behaviour.
  - Each coin module (xrp, stellar, canton, tron, evm, tezos) is now loaded only when the matching family is actually requested, instead of at module evaluation time.
  - No impact on LLD / LLM runtime — bundlers tree-shake these imports already. The fix primarily benefits headless CJS consumers (e.g. `wallet-cli`) where all `require()` calls are evaluated eagerly at startup.

### 📝 Description

`generic-alpaca/alpaca/index.ts` and `generic-alpaca/signer.ts` used to statically import every coin's API factory and hardware signer at the top of the file. This means that simply importing **any** alpaca bridge pulled in the full dependency trees of xrp, stellar, canton, tron, evm and tezos coin packages — including heavy optional deps like `tronweb` / protobuf — even if only one family is ever used at runtime.

This PR replaces those static imports with inline `require()` calls inside each `switch` branch, so a coin module is only evaluated when that family is actually matched. Each factory is extracted to a thin `local/<family>.ts` shim to keep the callsite readable.

```
// before
import { createApi as createTronApi } from "@ledgerhq/coin-tron/api/index";
...
case "tron":
  return createTronApi(getCurrencyConfiguration<TronCoinConfig>(currency.id));

// after
case "tron":
  return require("./local/tron").createLocalTronApi(currency.id);
```

A `TODO` comment is left on both files noting that `dynamic import` would be the cleaner long-term solution once the callers are made async.

### ❓ Context

- **JIRA or GitHub link**: Part of the `feat/wallet-cli` work — prerequisite to land before the wallet-cli bridge PR.
- **ADR link (if any)**: N/A

---

### DIFF
--- a/.changeset/alpaca-lazy-require.md
+++ b/.changeset/alpaca-lazy-require.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Lazy-load Alpaca coin API and signer modules on demand via `require()` instead of static imports to avoid evaluating unrelated coin stacks at startup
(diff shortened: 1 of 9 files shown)


Justkant (approved):

Justkant:
Not sure if we'll need to update the doc for coin-integration with those changes, maybe @hedi-edelbloute

hedi-edelbloute (approved):

PR #16187, opened 2026-04-09
[LWDM] Support/dmk singleton shared

fAnselmi-Ledger:
### ✅ Checklist

  - ...

### Summary
- Introduces a shared DMK singleton (`initDmk` / `getDeviceManagementKit`) in `@ledgerhq/live-dmk-shared`, replacing the duplicated lazy-init logic that existed independently in both `live-dmk-desktop` and `live-dmk-mobile`.
- Each app now owns its DMK initialisation: desktop calls `initialiseDmk()` inside `registerTransportModules()` (after `USER_ID` is set), mobile calls it at module-level in `AppProviders.tsx`.

### Motivation
The DMK initialisation logic (builder + transport factories + config) was copy-pasted across desktop and mobile with only the transport factories differing. This made it easy for the two to drift and made it unclear who "owns" the singleton. Now:
- **One source of truth** for the singleton lifecycle (`live-dmk-shared`)
- **Apps decide** which transports and config to use (`initDmk.ts` per app)
- **Platform libs adapt** (feature flags, platform-specific hooks) without owning the instance

### ❓ Context

- **JIRA or GitHub link**: 
- [LIVE-28652](https://ledgerhq.atlassian.net/browse/LIVE-28652)
- [LIVE-28653](https://ledgerhq.atlassian.net/browse/LIVE-28653)

---

[LIVE-28652]: https://ledgerhq.atlassian.net/browse/LIVE-28652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Justkant:
Nice this might also be useful for the wallet-cli branch where we use the DMK now, I'll check that

Justkant (dismissed):

benruseau (dismissed):
LGTM 👍

apps/ledger-live-desktop/src/renderer/App.tsx
```diff
@@ -48,6 +48,11 @@ type Props = {
 
 const queryClient = new QueryClient();
 
+const DmkProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+  const ldmkTransportFlag = !!useFeature("ldmkTransport")?.enabled;
+  return <SharedDmkProvider disabled={!ldmkTransportFlag}>{children}</SharedDmkProvider>;
```
OlivierFreyssinet: The feature flag is enabled on all environments, reintroducing the "disabled" mechanism on the provider goes in the wrong direction. This PR would be a good opportunity to clean up the feature flag actually.

apps/ledger-live-mobile/src/AppProviders.tsx
```diff
@@ -23,6 +24,8 @@ type AppProvidersProps = {
 
 const queryClient = new QueryClient();
 
+initialiseDmk();
```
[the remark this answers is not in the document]
OlivierFreyssinet: very important comment

apps/ledger-live-mobile/src/initDmk.ts
```diff
@@ -0,0 +1,16 @@
+import { initDmk, LedgerLiveLogger, UserHashService } from "@ledgerhq/live-dmk-shared";
+import { RNBleTransportFactory } from "@ledgerhq/device-transport-kit-react-native-ble";
+import { RNHidTransportFactory } from "@ledgerhq/device-transport-kit-react-native-hid";
+import { LogLevel } from "@ledgerhq/device-management-kit";
+import { getEnv } from "@ledgerhq/live-env";
+
+export function initialiseDmk(): void {
+  const userId = getEnv("USER_ID");
```
OlivierFreyssinet: I think we need to make the `USER_ID` initialization contract explicit here and guarantee it before DMK initialization, especially on mobile.

## Why this matters

`firmwareDistributionSalt` is derived from `getEnv("USER_ID")` and is used for progressive OS update rollout. The backend uses it to decide whether a given user is in the rollout bucket.

Because of that, the value used to derive the salt must be:
- stable for a given user / app instance
- non-empty
- available before DMK is initialized

If DMK is initialized with `USER_ID === ""`, then the derived `firmwareDistributionSalt` is identical for everyone. In that case the rollout is no longer progressive: either everyone gets the update, or no one does.

## Current concern

On desktop, we explicitly set `USER_ID` before calling `initialiseDmk()`, so the ordering is clear.

I do not see the same guarantee on mobile. In this PR, mobile even moves DMK initialization earlier, at module evaluation time in `AppProviders.tsx`, which makes the contract even more implicit.

So I think we should guarantee that `USER_ID` is set before `initialiseDmk()` is called on mobile as well, and ideally enforce it in code rather than relying on implicit startup ordering.

## Possible improvement

More broadly, I wonder whether DMK should stop relying on `getEnv("USER_ID")` and instead use the actual app source of truth, i.e. `userIdSelector(store.getState()).exportUserIdForAnalytics()`.

That said, this is only safe if DMK is initialized after identities hydration, because `userIdSelector` may still return `DUMMY_USER_ID` before the store is initialized.

So to me the key point is not only *which source* we use, but also *when* DMK is initialized.

Could we make DMK initialization happen only once the app has a real user ID available, and explicitly fail / guard if the value is still unset or dummy?

OlivierFreyssinet (changes requested):

OlivierFreyssinet (commented):
More general comment:
It's going in the right direction, but the DMK initialization is still not centralized:
With `DeviceManagementKitTransportSpeculos` still instantiating its own DMK, DMK initialization is only centralized for the “regular” transports. In practice, this leaves us with two DMK ownership models:
- app-owned singleton for desktop/mobile runtime transports
- transport-owned DMK for Speculos

It means the app cannot interact with Speculos through the same DMK instance/API surface as the rest of the platform.
It would be difficult to use the same singleton for Speculos as its config requires a URL, so maybe it would require a rework of the Speculos transport to make it configurable at after instantiation ...

PR #16326, opened 2026-04-14
[LWDM] feat(lwdm): extractBalance adapt for generic adapter staking evm

dilaouid:
### ✅ Checklist

  - ...

### 📝 Description

This updates the generic adapter so send-related balance checks use the actual spendable balance instead of the account’s total balance. That fixes staking EVM cases like SEI where maxSpendable and send max were incorrectly including delegated funds. In short, delegated/locked assets are no longer treated as sendable by the generic balance extraction used in fee estimation and validation.

### ❓ Context

- **[JIRA or GitHub link](https://ledgerhq.atlassian.net/browse/LIVE-29170)** 
- **ADR link (if any)**: 

---

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
+++ b/libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
@@ -68,9 +68,11 @@ export function extractBalances(
 ): Balance[] {
   const balances: Balance[] = [
     {
-      value: BigInt(account.balance.toFixed()),
+      // Send flows expect spendable balances here. Using the account total balance
+      // makes `useAllAmount` include staked/locked funds for staking-enabled EVM accounts.
+      value: BigInt(account.spendableBalance.toFixed()),
       asset: { type: "native" },
-      locked: BigInt(account.balance.minus(account.spendableBalance).toFixed()),
+      locked: 0n,
     },
   ];
 
@@ -81,9 +83,9 @@ export function extractBalances(
   for (const subAccount of account.subAccounts) {
     const asset = getAssetFromToken(subAccount.token, account.freshAddress);
     balances.push({
-      value: BigInt(subAccount.balance.toFixed()),
+      value: BigInt(subAccount.spendableBalance.toFixed()),
       asset,
-      locked: BigInt(subAccount.balance.minus(subAccount.spendableBalance).toFixed()),
+      locked: 0n,
     });
   }
 
@@ -202,10 +204,10 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
   }
 
   if (op.details?.stake && typeof op.details.stake === "object") {
-    const s = op.details.stake as { address?: string; amount?: bigint | number | string };
+    const s = op.details.stake as { address?: string; amount?: bigint };
     extra.stake = {
       address: s.address ?? "",
-      amount: new BigNumber(s.amount?.toString() ?? "0"),
+      amount: new BigNumber(s.amount !== undefined ? String(s.amount) : "0"),
     };
   }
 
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -203,9 +205,17 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
 
   if (op.details?.stake && typeof op.details.stake === "object") {
     const s = op.details.stake as { address?: string; amount?: bigint | number | string };
+    let normalizedStakeAmount = "0";
```
Moustafa-Koterba: **nit** Can we simplify this ?

```suggestion
    const normalizedStakeAmount = typeof s.amount === "number" && !Number.isFinite(s.amount) ? "0" : String(s.amount);
```

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -483,6 +493,8 @@ export const buildOptimisticOperation = (
   const { subAccounts } = account;
   const parentType = subAccountId ? "FEES" : type;
   const tokenAccount = subAccountId ? subAccounts?.find(ta => ta.id === subAccountId) : null;
+  const normalizedSequenceNumber = sequenceNumber === undefined ? "0" : String(sequenceNumber);
```
Moustafa-Koterba: **nit**
```suggestion
  const normalizedSequenceNumber = String(sequenceNumber ?? 0);
```
same for nonce

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
dilaouid: `s.amount ` can be `undefined`, so it will in this case returns String(undefined) = "undefined"

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
dilaouid: nice catch!
(but cannot for nonce, it will returns `new BigNumber("0")` instead of `undefined`

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
```diff
@@ -203,9 +205,17 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
...
   if (op.details?.stake && typeof op.details.stake === "object") {
     const s = op.details.stake as { address?: string; amount?: bigint | number | string };
+    let normalizedStakeAmount = "0";
+    if (typeof s.amount === "bigint") {
+      normalizedStakeAmount = String(s.amount);
+    } else if (typeof s.amount === "number") {
+      normalizedStakeAmount = Number.isFinite(s.amount) ? s.amount.toString() : "0";
+    } else if (typeof s.amount === "string") {
+      normalizedStakeAmount = s.amount;
+    }
```
francois-guerin-ledger: *question*: Why is `op.details.stake.amount` one of `bigint | number | string` in the first place ? It creates complex code like this one.

francois-guerin-ledger (dismissed):

libs/ledger-live-common/src/bridge/generic-alpaca/utils.ts
dilaouid: those were defensive guards for the record type (abusive a little i agree)
simplified it

francois-guerin-ledger: Way better, thanks.

francois-guerin-ledger (approved):

PR #16333, opened 2026-04-14
[LWDM] feat(live-common): migrate getAlpacaApi to async dynamic import() [LIVE-29178]

gre-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** Unit tests in `alpaca/index.unit.test.ts` updated to `async/await`.
  - Internal refactor within `libs/ledger-live-common` — no public API change.
  - All callers of `getAlpacaApi` were already in async contexts; adding `await` is the only consumer change.
  - No behavioral change expected; lazy-loading semantics are preserved (now truly async).

### 📝 Description

Replaces the synchronous `require()` lazy-loading in `generic-alpaca/alpaca/index.ts` with proper async `import()`, making `getAlpacaApi` an `async` function that returns `Promise<AlpacaApi & BridgeApi>`.

All 7 call sites (`getAccountShape`, `prepareTransaction`, `getTransactionStatus`, `estimateMaxSpendable`, `signOperation`, `signRawOperation`, `broadcast`) were already inside `async` functions — the migration is purely mechanical (`await getAlpacaApi(...)`).

A `moduleNameMapper` entry (`^(\.{1,2}/.+)\.js$`) was added to `jest.config.ts` to fix module resolution when using `moduleResolution: NodeNext` with Jest (which doesn't understand `.js` imports pointing to `.ts` source files).

### ❓ Context

- **JIRA**: [LIVE-29178](https://ledgerhq.atlassian.net/browse/LIVE-29178) — part of [LIVE-29176](https://ledgerhq.atlassian.net/browse/LIVE-29176) (broader require → import() migration)
- `getSigner` in `signer.ts` is **not** in scope here: its callers are synchronous and the propagation is wider (~100 call sites). Tracked separately.

---

[LIVE-29178]: https://ledgerhq.atlassian.net/browse/LIVE-29178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-alpaca/alpaca/index.ts
+++ b/libs/ledger-live-common/src/bridge/generic-alpaca/alpaca/index.ts
@@ -7,28 +7,29 @@ import { getNetworkAlpacaApi } from "./network/network-alpaca";
  * Lazy-load coin Alpaca API modules so consumers (e.g. wallet-cli EVM-only) do not evaluate
  * unrelated coin stacks (Tron pulls tronweb/protobuf, which breaks under some Bun runtimes).
  *
- * Uses `require()` so the first `lib/` (CJS) build only loads the module for the matched family.
- * The `lib-es/` build reuses the same source; bundlers / Bun resolve these subpaths on demand.
+ * Uses dynamic `import()` so each build only loads the module for the matched family on demand.
  */
-// TODO: we could also use dynamic import here but we need to update everything to async/await where getAlpacaApi is used
-export function getAlpacaApi(network: string, kind: string): AlpacaApi<any> & BridgeApi {
+export async function getAlpacaApi(
+  network: string,
+  kind: string,
+): Promise<AlpacaApi<any> & BridgeApi> {
   if (kind === "local") {
     const currency = findCryptoCurrencyByNetwork(network);
     switch (currency?.family) {
       case "xrp":
-        return require("./local/xrp").createLocalXrpApi(currency.id);
+        return (await import("./local/xrp.js")).createLocalXrpApi(currency.id);
       case "stellar":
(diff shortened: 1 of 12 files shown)


francois-guerin-ledger (approved):

libs/ledger-live-common/src/bridge/generic-alpaca/alpaca/index.ts
```diff
@@ -7,28 +7,29 @@ import { getNetworkAlpacaApi } from "./network/network-alpaca";
...
+export async function getAlpacaApi(
+  network: string,
+  kind: string,
+): Promise<AlpacaApi<any> & BridgeApi> {
   if (kind === "local") {
     const currency = findCryptoCurrencyByNetwork(network);
     switch (currency?.family) {
       case "xrp":
-        return require("./local/xrp").createLocalXrpApi(currency.id);
+        return (await import("./local/xrp.js")).createLocalXrpApi(currency.id);
```
Justkant, line 20: It would be good if we could avoid the `.js` extension because it will probably won't find the types when we haven't built common I think

Justkant: Also
> To be able to drop .js extensions, the project would need to migrate to "moduleResolution": "bundler" (which is designed for bundler-processed projects and doesn't require explicit extensions).

gre-ledger: let's study this more in near future, i'll add it in the EPIC because it will be the same topic for the bridge

PR #16386, opened 2026-04-16
feat(wallet-cli): support arbitrary EVM calldata in TransactionIntent

gre-ledger:
### ✅ Checklist

  - `EvmTransactionIntentSchema` gains an optional `data` field (0x-prefixed hex calldata) — fully backward-compatible, existing sends unaffected
  - `BridgeAdapter.buildTxExtras` now propagates `data` to the live-common EVM transaction as `Buffer`
  - `buildTxExtras` refactored from flat spread-ternary + `"key" in intent` guards to a `switch (intent.family)` block — no behaviour change, maintainability improvement

### 📝 Description

Adds support for arbitrary EVM calldata in the `wallet-cli` `TransactionIntent` layer.

**Before:** `EvmTransactionIntentSchema` only had `recipient` and `amount`. No way to express contract calls, approvals, or any transaction with custom calldata.

**After:** An optional `data` field (0x-prefixed hex string, even-digit enforced by regex) can be passed. The `BridgeAdapter` converts it to a `Buffer` and forwards it to the live-common EVM transaction, where `coin-evm` already gives it priority over auto-generated ERC-20 calldata.

For pure contract calls with no ETH value, callers pass `amount: "0 ETH"` (already supported by `parseAmountWithTicker`).

Also refactors `buildTxExtras`:
- Before: flat list of `...("key" in intent && intent.key ? { key: val } : {})` — no TypeScript narrowing, fragile as new families/fields were added
- After: `switch (intent.family)` — TypeScript narrows each case fully, per-family logic is isolated, adding a field touches only one case

### ❓ Context

- **JIRA or GitHub link**: N/A (internal `wallet-cli` improvement)
- **ADR link (if any)**: N/A

---

### DIFF
--- a/apps/wallet-cli/src/wallet/compatibility/bridge.ts
+++ b/apps/wallet-cli/src/wallet/compatibility/bridge.ts
@@ -183,19 +183,27 @@ export class BridgeAdapter {
   }
 
   private buildTxExtras(intent: TransactionIntent, tokenAccount: TokenAccount | undefined) {
-    return {
-      ...(tokenAccount ? { subAccountId: tokenAccount.id } : {}),
-      ...("feePerByte" in intent && intent.feePerByte
-        ? { feePerByte: new BigNumber(intent.feePerByte) }
-        : {}),
-      ...("rbf" in intent && intent.rbf != null ? { rbf: intent.rbf } : {}),
-      ...("mode" in intent && intent.mode ? { mode: intent.mode } : {}),
-      ...("validator" in intent && intent.validator ? { validator: intent.validator } : {}),
-      ...("stakeAccount" in intent && intent.stakeAccount
-        ? { stakeAccountId: intent.stakeAccount }
-        : {}),
-      ...("memo" in intent && intent.memo ? { memo: intent.memo } : {}),
-    };
+    const patch: Record<string, unknown> = {};
+    if (tokenAccount) patch.subAccountId = tokenAccount.id;
+    switch (intent.family) {
+      case "bitcoin":
+        if (intent.feePerByte) patch.feePerByte = new BigNumber(intent.feePerByte);
+        if (intent.rbf != null) patch.rbf = intent.rbf;
+        break;
+      case "evm":
+        if (intent.data) {
+          const hexData = intent.data.slice(2);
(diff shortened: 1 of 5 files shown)


apps/wallet-cli/src/wallet/compatibility/bridge.ts
```diff
@@ -183,19 +183,24 @@ export class BridgeAdapter {
...
+        break;
+      case "evm":
+        if (intent.data) patch.data = Buffer.from(intent.data.slice(2), "hex");
+        break;
+      case "solana":
+        if (intent.mode) patch.mode = intent.mode;
+        if (intent.validator) patch.validator = intent.validator;
+        if (intent.stakeAccount) patch.stakeAccountId = intent.stakeAccount;
+        if (intent.memo) patch.memo = intent.memo;
+        break;
```
gre-ledger, line 204: note on the necessary to maintain this layer: tbd if we can converge to the idea of "cli-transaction" that we had in the past, for now we inline it here. TBD if alpaca tx model have something to offer

apps/wallet-cli/src/wallet/compatibility/bridge.ts
```diff
@@ -183,19 +183,27 @@ export class BridgeAdapter {
...
+      case "evm":
+        if (intent.data) {
+          const hexData = intent.data.slice(2);
+          if (hexData.length > 0) patch.data = Buffer.from(hexData, "hex");
+        }
+        break;
+      case "solana":
+        if (intent.mode) patch.mode = intent.mode;
+        if (intent.validator) patch.validator = intent.validator;
+        if (intent.stakeAccount) patch.stakeAccountId = intent.stakeAccount;
```
Justkant, line 202: Do we transform somewhere from descriptor to accountId, I'm not sure if stakeAccount could be a descriptor here

Justkant (approved):

apps/wallet-cli/src/wallet/compatibility/bridge.ts
gre-ledger: definitely can't. good catch. that makes it basically blind from the cli usage agent perspective, as it doesn't know about accountId (abstracted out from the agent side). so let's track it somewhere for future fix.

PR #16403, opened 2026-04-16
[LWDM] Revert "feat(live-common): migrate getAlpacaApi to async dynamic impo…

gre-ledger:
possible PR to merge if we don't find a solution:

there is a runtime complication in mobile app by using `import()`

7d8cf19e-9055-49c0-a083-471a68b45f0d

I hope we manage to fix it different but let's confirm it's related to this code

### DIFF
--- a/.changeset/async-getalpacaapi-import.md
+++ b/.changeset/async-getalpacaapi-import.md
@@ -1,5 +1,0 @@
----
-"@ledgerhq/live-common": patch
----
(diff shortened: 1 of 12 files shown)


Justkant (approved):

gre-ledger:
possibly better fix will be https://github.com/LedgerHQ/ledger-live/pull/16404

PR #16446, opened 2026-04-17
[LWD] fix: e2e test reliability

claudiiafg:
### ✅ Checklist

  - ...

### 📝 Description

Add stable data-testids for W40 Assets rows and Cryptos account rows
Page objects: portfolio / accounts / market / navigation aligned with W40 routes and selectors.

### ❓ Context

[LIVE-29286](https://ledgerhq.atlassian.net/browse/LIVE-29286)
---

[LIVE-29286]: https://ledgerhq.atlassian.net/browse/LIVE-29286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/desktop/tests/page/account.page.ts
+++ b/e2e/desktop/tests/page/account.page.ts
@@ -39,20 +39,9 @@ export class AccountPage extends AppPage {
   private selectSpecificOperation = (operationType: string) =>
     this.page.locator("[data-testid^='operation-row-']").filter({ hasText: operationType });
 
-  private async isExpectedAccountHeaderVisible(expectedNames: string[]): Promise<boolean> {
-    if (!(await this.accountName.isVisible().catch(() => false))) {
-      return false;
-    }
-
-    const expectedSet = new Set(expectedNames.map(name => name.trim()));
-    const inputValue = await this.accountName.inputValue().catch(() => "");
-
-    if (inputValue && expectedSet.has(inputValue.trim())) {
-      return true;
-    }
-
-    const textValue = (await this.accountName.textContent())?.trim() ?? "";
-    return textValue !== "" && expectedSet.has(textValue);
+  private accountHeaderNamePattern(...names: string[]) {
+    const escaped = names.map(n => n.trim().replaceAll(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`));
+    return new RegExp(`^(${escaped.join("|")})$`);
   }
 
   @step("Navigate to token")
@@ -60,7 +49,7 @@ export class AccountPage extends AppPage {
     const tokenRow = this.tokenValue(account.currency.name).or(
       this.tokenRowByTicker(account.currency.ticker),
     );
-    await tokenRow.waitFor({ state: "visible" });
+    await expect(tokenRow).toBeVisible();
     await tokenRow.click();
   }
 
@@ -71,9 +60,8 @@ export class AccountPage extends AppPage {
 
   @step("Click `Show balance` button")
   async clickShowBalance() {
-    if (await this.activatePrivateBalanceButton.isVisible().catch(() => false)) {
+    if (await this.activatePrivateBalanceButton.isVisible()) {
       await this.activatePrivateBalanceButton.click();
-      return;
     }
   }
 
@@ -143,17 +131,15 @@ export class AccountPage extends AppPage {
 
   @step("Wait for account $0 to be visible")
   async expectAccountVisibility(firstAccountName: string) {
-    await expect
-      .poll(() => this.isExpectedAccountHeaderVisible([firstAccountName]), {
-        timeout: 30000,
-      })
-      .toBeTruthy();
-    await this.settingsButton.waitFor({ state: "visible" });
+    await expect(this.accountName).toHaveValue(this.accountHeaderNamePattern(firstAccountName), {
+      timeout: 30000,
+    });
+    await expect(this.settingsButton).toBeVisible();
   }
 
   @step("Expect account to be not null")
   async expectAccountBalance() {
-    expect(this.accountBalance).toBeTruthy();
+    await expect(this.accountBalance).toBeVisible();
   }
 
   @step("Expect `Last operations` to be visible")
@@ -194,7 +180,7 @@ export class AccountPage extends AppPage {
   @step("Expect token to be present")
   async expectTokenToBePresent(tokenAccount: AccountType) {
     const row = this.tokenRow(tokenAccount.currency.ticker);
-    await this.showAllTokensButton.or(row).first().waitFor({ state: "visible" });
+    await expect(this.showAllTokensButton.or(row).first()).toBeVisible();
     if (await this.showAllTokensButton.isVisible()) {
       await this.showAllTokensButton.click();
     }
@@ -204,7 +190,7 @@ export class AccountPage extends AppPage {
   @step("Navigate to token in account")
   async navigateToTokenInAccount(tokenAccount: AccountType) {
     const row = this.tokenRow(tokenAccount.currency.ticker);
-    await this.showAllTokensButton.or(row).first().waitFor({ state: "visible" });
+    await expect(this.showAllTokensButton.or(row).first()).toBeVisible();
     if (await this.showAllTokensButton.isVisible()) {
       await this.showAllTokensButton.click();
     }
@@ -214,26 +200,22 @@ export class AccountPage extends AppPage {
 
   @step("Wait for account header name $0 to be visible")
   async waitForAccountHeaderName(headerName: string, fallbackName?: string) {
-    const expectedNames = fallbackName ? [headerName, fallbackName] : [headerName];
-    await expect
-      .poll(() => this.isExpectedAccountHeaderVisible(expectedNames), {
-        timeout: 30000,
-      })
-      .toBeTruthy();
+    const pattern = fallbackName
+      ? this.accountHeaderNamePattern(headerName, fallbackName)
+      : this.accountHeaderNamePattern(headerName);
+    await expect(this.accountName).toHaveValue(pattern, { timeout: 30000 });
   }
 
   @step("Verify account with header name $0 is visible")
   async verifyAccountHeaderNameIsVisible(headerName: string) {
-    await expect
-      .poll(() => this.isExpectedAccountHeaderVisible([headerName]), {
-        timeout: 30000,
-      })
-      .toBeTruthy();
+    await expect(this.accountName).toHaveValue(this.accountHeaderNamePattern(headerName), {
+      timeout: 30000,
+    });
   }
 
   @step("Check account chart is visible")
   async checkAccountChart() {
-    await this.accountChart.waitFor({ state: "visible" });
+    await expect(this.accountChart).toBeVisible();
   }
 
   @step("Click on selected ($0) last operation")
--- a/e2e/desktop/tests/page/account.page.ts
+++ b/e2e/desktop/tests/page/account.page.ts
@@ -39,6 +39,22 @@ export class AccountPage extends AppPage {
   private selectSpecificOperation = (operationType: string) =>
     this.page.locator("[data-testid^='operation-row-']").filter({ hasText: operationType });
 
+  private async isExpectedAccountHeaderVisible(expectedNames: string[]): Promise<boolean> {
+    if (!(await this.accountName.isVisible().catch(() => false))) {
+      return false;
+    }
+
+    const expectedSet = new Set(expectedNames.map(name => name.trim()));
+    const inputValue = await this.accountName.inputValue().catch(() => "");
+
+    if (inputValue && expectedSet.has(inputValue.trim())) {
+      return true;
+    }
+
+    const textValue = (await this.accountName.textContent())?.trim() ?? "";
+    return textValue !== "" && expectedSet.has(textValue);
+  }
+
   @step("Navigate to token")
   async navigateToToken(account: AccountType) {
     const tokenRow = this.tokenValue(account.currency.name).or(
@@ -127,7 +143,11 @@ export class AccountPage extends AppPage {
 
   @step("Wait for account $0 to be visible")
   async expectAccountVisibility(firstAccountName: string) {
-    await expect(this.accountName).toHaveValue(firstAccountName);
+    await expect
+      .poll(() => this.isExpectedAccountHeaderVisible([firstAccountName]), {
+        timeout: 30000,
+      })
+      .toBeTruthy();
     await this.settingsButton.waitFor({ state: "visible" });
   }
 
@@ -195,12 +215,20 @@ export class AccountPage extends AppPage {
   @step("Wait for account header name $0 to be visible")
   async waitForAccountHeaderName(headerName: string, fallbackName?: string) {
     const expectedNames = fallbackName ? [headerName, fallbackName] : [headerName];
-    await expect(this.accountName).toHaveValue(new RegExp(`^(${expectedNames.join("|")})$`));
+    await expect
+      .poll(() => this.isExpectedAccountHeaderVisible(expectedNames), {
+        timeout: 30000,
+      })
+      .toBeTruthy();
   }
 
   @step("Verify account with header name $0 is visible")
   async verifyAccountHeaderNameIsVisible(headerName: string) {
-    await expect(this.accountName).toHaveValue(headerName);
+    await expect
+      .poll(() => this.isExpectedAccountHeaderVisible([headerName]), {
+        timeout: 30000,
+      })
+      .toBeTruthy();
   }
 
   @step("Check account chart is visible")
--- a/e2e/desktop/tests/page/accounts.page.ts
+++ b/e2e/desktop/tests/page/accounts.page.ts
@@ -5,14 +5,14 @@ import { Currency } from "@ledgerhq/live-common/e2e/enum/Currency";
 
 export class AccountsPage extends AppPage {
   private accountsTitle = this.page.getByRole("heading", { name: "Accounts" });
-  private firstAccount = this.page.locator(`[data-testid^="crypto-account-row-"]`).first();
+  private readonly visibleAccountsList = this.page
+    .locator(`[data-testid^="crypto-account-row-"]`)
+    .filter({ visible: true });
 
-  private getSanitizedAccountName = (accountName: string) => accountName.replaceAll(/\s+/g, "-");
-
-  // Accounts context menu
-  private async getAccountListLocator() {
-    return this.page.locator(`[data-testid^="crypto-account-row-"]:visible`);
-  }
+  private readonly getSanitizedAccountName = (accountName: string) =>
+    accountName.replaceAll(/\s+/g, "-");
+  private readonly tokenRow = (childCurrency: Currency) =>
+    this.page.getByTestId(`token-row-${childCurrency.ticker}`);
 
   private syncAccountButton = (accountName: string) =>
     this.page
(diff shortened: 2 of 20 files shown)


claudiiafg:
latest run: https://github.com/LedgerHQ/ledger-live/actions/runs/24561778818

e2e/desktop/tests/page/drawer/ledger.sync.drawer.ts
```diff
@@ -22,8 +25,12 @@ export class LedgerSyncDrawer extends Drawer {
 
   @step("Synchronize accounts")
   async syncAccounts() {
-    await expect(this.continueButton).toBeVisible();
-    await this.continueButton.click();
+    if (await this.continueButton.isVisible().catch(() => false)) {
```
martijnhjk: Do we need the catch statement? It should just return true or false, see https://playwright.dev/docs/api/class-locator#locator-is-visible

e2e/desktop/tests/page/drawer/ledger.sync.drawer.ts
```diff
@@ -22,8 +25,12 @@ export class LedgerSyncDrawer extends Drawer {
...
+      await this.continueButton.click();
+    } else if (await this.alreadyTurnedOnButton.isVisible().catch(() => false)) {
```
martijnhjk: same here, I think we can remove the catch.

e2e/desktop/tests/page/drawer/ledger.sync.drawer.ts
```diff
@@ -58,7 +65,21 @@ export class LedgerSyncDrawer extends Drawer {
...
+            .isVisible()
+            .catch(() => false);
+          const hasConnectDevice = await this.walletSyncConnectDeviceButton
+            .isVisible()
+            .catch(() => false);
+          return hasContinue || hasAlreadyTurnedOn || hasConnectDevice;
+        },
+        { timeout: 30000 },
+      )
+      .toBeTruthy();
```
martijnhjk: It might be safer to explicitly expect `true`, eg `.toBe(true)`. There is also expect to [pass](https://playwright.dev/docs/test-assertions#expecttopass) which expects the test within the block to pass. That said, we will not know from the report which of the conditions passed so it is always best to have a specific single state assertion if possible.

e2e/desktop/tests/page/account.page.ts
```diff
@@ -39,6 +39,22 @@ export class AccountPage extends AppPage {
   private selectSpecificOperation = (operationType: string) =>
     this.page.locator("[data-testid^='operation-row-']").filter({ hasText: operationType });
 
+  private async isExpectedAccountHeaderVisible(expectedNames: string[]): Promise<boolean> {
```
martijnhjk: Is there no way we can achieve what you want using native playwright matchers? They can allow partial string matching, regex patterns, etc. Playwright matchers will auto wait and auto retry and are recommended for optimal stability.

e2e/desktop/tests/page/account.page.ts
```diff
@@ -127,7 +143,11 @@ export class AccountPage extends AppPage {
 
   @step("Wait for account $0 to be visible")
   async expectAccountVisibility(firstAccountName: string) {
-    await expect(this.accountName).toHaveValue(firstAccountName);
+    await expect
+      .poll(() => this.isExpectedAccountHeaderVisible([firstAccountName]), {
+        timeout: 30000,
+      })
+      .toBeTruthy();
```
martijnhjk: Same here, we've had tests showing false passes before because truthy will pass for anything that is defined, for example `"false"` or an object, etc. Please consider explicit `.toBe(true)`.

e2e/desktop/tests/page/accounts.page.ts
```diff
@@ -50,36 +41,30 @@ export class AccountsPage extends AppPage {
...
-      await this.verifyTokenNotVisible(parentName, childCurrency);
-    }
+    await this.navigateToAccountByName(parentName);
+    await this.verifyTokenNotVisible(parentName, childCurrency);
   }
 
   @step("Verify token visibility having parent $0")
-  async verifyTokenVisibility(parentName: string, childCurrency: Currency) {
-    await expect(this.tokenRow(parentName, childCurrency)).toBeVisible();
+  async verifyTokenVisibility(_parentName: string, childCurrency: Currency) {
```
martijnhjk: please consider removing the parameter `_parentName` if no longer used.

e2e/desktop/tests/page/mainNavigation.page.ts
```diff
@@ -38,63 +35,54 @@ export class MainNavigationPage extends AppPage {
     });
   }
 
-  private get sidebarTargets(): Readonly<Record<TargetName, NavigationTarget>> {
+  private getSidebarTargets(): Readonly<Record<TargetName, NavigationTarget>> {
```
martijnhjk: why does this need to be a function instead of a getter?

e2e/desktop/tests/page/mainNavigation.page.ts
```diff
@@ -38,63 +35,54 @@ export class MainNavigationPage extends AppPage {
...
     };
   }
   @step("Open $0 from main navigation")
   async openTargetFromMainNavigation(target: TargetName) {
-    const targetSelector = (await isWallet40Enabled(this.page))
-      ? this.sidebarTargets[target].selector
-      : this.sidebarTargets[target].legacySelector;
-
-    await targetSelector.click();
+    const sidebarTargets = this.getSidebarTargets();
```
martijnhjk: this change is not required if we keep the getter syntax.

e2e/desktop/tests/page/mainNavigation.page.ts
```diff
@@ -38,63 +35,54 @@ export class MainNavigationPage extends AppPage {
...
+    await sidebarTargets[target].selector.click();
   }
 
   @step("Validate $0 target from main navigation is selected and redirect to the expected path")
   async validateTargetFromMainNavigation(target: TargetName) {
-    const targetConfig = this.sidebarTargets[target];
+    const sidebarTargets = this.getSidebarTargets();
```
martijnhjk: same here, we can keep as is if we keep the getter further up.

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -2,60 +2,106 @@ import { step } from "tests/misc/reporters/step";
...
+  private readonly assetRowElements = this.page.locator("[data-testid^='asset-row-']");
+  private readonly showAllButton = this.page.getByText("Show all");
+  private readonly showMoreButton = this.page.getByText("Show more");
+  private readonly assetRow = (asset: string) =>
+    this.page.getByTestId(`asset-row-${asset.toLowerCase()}`);
+  private readonly sanitizeAssetNameForTestId = (asset: string) =>
+    asset
+      .toLowerCase()
+      .replaceAll(/\s+/g, "-")
+      .replaceAll(/[^a-z0-9-]/g, "");
```
martijnhjk: This feels like adding complexity to our test code rather than making the testid in the source code fit for purpose. When the testid is created the usage in the tests (component and e2e) should be considered so that the test code can stay simple and effective.

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -98,7 +144,9 @@ export class PortfolioPage extends AppPage {
 
   @step("Click on asset row $0")
   async clickOnSelectedAssetRow(asset: string) {
-    await this.assetRow(asset).click();
+    const { row } = await this.resolveAssetRow(asset);
+    await expect(row).toBeVisible();
```
martijnhjk: The [click()](https://playwright.dev/docs/api/class-locator#locator-click) function should have a built-in wait for visible. Do we need this explicit expect?

claudiiafg:
lateste run: https://github.com/LedgerHQ/ledger-live/actions/runs/24655859333

mcayuelas-ledger (dismissed):

e2e/desktop/tests/page/accounts.page.ts
```diff
@@ -2,45 +2,37 @@ import { expect } from "@playwright/test";
...
-
-  private showTokensButton = (parentName: string) =>
-    this.accountComponent(parentName).locator("xpath=following-sibling::button").first();
+  private readonly getSanitizedAccountName = (accountName: string) =>
+    accountName.replaceAll(/\s+/g, "-");
 
   // Accounts context menu
-  private accountListNumber = this.page.locator(`[data-testid^="account-component-"]`);
+  private async getAccountListLocator() {
+    return this.page.locator(`[data-testid^="crypto-account-row-"]:visible`);
```
martijnhjk: it might be more readable and reliable using playwright [filter](https://playwright.dev/docs/locators#matching-only-visible-elements) for this, for example:

```
this.page.locator(`[data-testid^="crypto-account-row-"]`).filter({ visible: true });
```

and if this just returns a locator can we set this as a property instead of a function? For example, like `firstAccount`?

e2e/desktop/tests/page/accounts.page.ts
```diff
@@ -2,45 +2,37 @@ import { expect } from "@playwright/test";
...
+  }
 
   private syncAccountButton = (accountName: string) =>
-    this.accountComponent(accountName).getByTestId("sync-button").locator("div").first();
+    this.page
+      .getByTestId(`crypto-account-row-${this.getSanitizedAccountName(accountName)}`)
```
martijnhjk, line 19: this pattern is used elsewhere too, maybe we can put it into a reusable function or getter?

e2e/desktop/tests/page/accounts.page.ts
```diff
@@ -50,36 +42,32 @@ export class AccountsPage extends AppPage {
...
+    await this.navigateToAccountByName(parentName);
+    await this.verifyTokenNotVisible(childCurrency);
   }
 
-  @step("Verify token visibility having parent $0")
-  async verifyTokenVisibility(parentName: string, childCurrency: Currency) {
-    await expect(this.tokenRow(parentName, childCurrency)).toBeVisible();
+  @step("Verify token visibility")
+  async verifyTokenVisibility(childCurrency: Currency) {
+    await expect(this.page.getByTestId(`token-row-${childCurrency.ticker}`)).toBeVisible();
```
martijnhjk: this is also used in the negative assertion, maybe we can move it to a reusable function or getter too?

e2e/desktop/tests/page/accounts.page.ts
```diff
@@ -90,12 +78,14 @@ export class AccountsPage extends AppPage {
 
   @step("Get number of accounts in the list")
   async countAccounts(): Promise<number> {
-    return await this.accountListNumber.count();
+    const accountList = await this.getAccountListLocator();
```
martijnhjk: see other comment, if the locator is stored as a property this could be simplified to:

```
return await this.visibleAccountsList.count();
```

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -1,61 +1,95 @@
...
+  private readonly cryptoBannerAddAccountButton = this.page.getByTestId(
+    "crypto-addresses-banner-add-account-cta",
+  );
+  /** Prefer banner CTA when both exist in DOM; Playwright picks the actionable (visible) match. */
+  private readonly addAccountCta = this.cryptoBannerAddAccountButton.or(
+    this.portfolioAddAccountButton,
+  );
+  private readonly noDeviceTitle = this.page.getByTestId("no-device-title");
+
+  private async isVisible(locator: Locator): Promise<boolean> {
```
martijnhjk: Do we need this extra layer of abstraction?

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -1,61 +1,95 @@
...
+  private async resolveAssetRow(asset: string): Promise<{ row: Locator; isLegacy: boolean }> {
+    const legacyRow = this.assetRow(asset);
+    if (await this.isVisible(legacyRow)) {
+      return { row: legacyRow, isLegacy: true };
+    }
+
+    return { row: this.w40AssetRow(asset), isLegacy: false };
+  }
+
+  private async waitForBalanceDiffWidget(): Promise<Locator> {
```
martijnhjk: This is only used once, do we need it as a function? Can we just do the expect assertion directly in the step?

The returned locator value can just be called directly in the step for its required purpose. I am not sure this waitFor function is adding extra value.

e2e/desktop/tests/page/drawer/ledger.sync.drawer.ts
```diff
@@ -22,8 +25,17 @@ export class LedgerSyncDrawer extends Drawer {
 
   @step("Synchronize accounts")
   async syncAccounts() {
-    await expect(this.continueButton).toBeVisible();
-    await this.continueButton.click();
+    await this.continueButton
+      .or(this.alreadyTurnedOnButton)
+      .or(this.walletSyncConnectDeviceButton)
+      .waitFor({ state: "visible", timeout: 30000 });
+
```
VicAlbr: Im' not a super fan of this or method. Is there a way to be sure to always have the right scénario ?

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -136,30 +161,43 @@ export class PortfolioPage extends AppPage {
...
   async expectBalanceDiffCounterValue(counterValue: string) {
-    const selector = (await isWallet40Enabled(this.page)) ? this.portfolioTrend : this.balanceDiff;
-    await expect(selector).toBeVisible();
-    await expect(selector).toContainText(counterValue);
+    const locator = await this.waitForBalanceDiffWidget();
+
+    // W40 trend percentage can be hidden in discreet mode and displayed as "***".
+    if (counterValue === "%") {
+      await expect(locator).toContainText(/%|\*\*\*/);
+      return;
```
martijnhjk: please consider using if ... else instead of a return here for readability and consistency.

e2e/desktop/tests/page/account.page.ts
```diff
@@ -127,13 +131,15 @@ export class AccountPage extends AppPage {
 
   @step("Wait for account $0 to be visible")
   async expectAccountVisibility(firstAccountName: string) {
-    await expect(this.accountName).toHaveValue(firstAccountName);
+    await expect(this.accountName).toHaveValue(this.accountHeaderNamePattern(firstAccountName), {
+      timeout: 30000,
```
VicAlbr, line 135: Do we need to hardcode the timeout ? is it possible to use the global one instead ?

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -1,61 +1,95 @@
...
+  );
+  private readonly noDeviceTitle = this.page.getByTestId("no-device-title");
+
+  private async isVisible(locator: Locator): Promise<boolean> {
+    return locator.isVisible();
+  }
+
+  private async resolveAssetRow(asset: string): Promise<{ row: Locator; isLegacy: boolean }> {
+    const legacyRow = this.assetRow(asset);
+    if (await this.isVisible(legacyRow)) {
```
martijnhjk: personally I think testing the wallet 4.0 toggle state is faster and more reliable for the test instead of relying on the UI under test.

e2e/desktop/tests/page/portfolio.page.ts
```diff
@@ -1,61 +1,95 @@
...
+
+    return { row: this.w40AssetRow(asset), isLegacy: false };
+  }
+
+  private async waitForBalanceDiffWidget(): Promise<Locator> {
+    await expect(this.portfolioTrendPercentage).toBeVisible({ timeout: 30000 });
+    return this.portfolioTrendPercentage;
+  }
+
+  private getExpectedCounterValuePattern(counterValue: string): RegExp {
```
martijnhjk, line 55: This is only used once, do we need it in a function? Please consider writing this in the step for better context and readability.

e2e/desktop/tests/specs/subAccount.spec.ts
```diff
@@ -20,7 +20,20 @@ import {
 } from "@ledgerhq/live-common/e2e/cliCommandsUtils";
 import { Addresses } from "@ledgerhq/live-common/e2e/enum/Addresses";
 import { Currency } from "@ledgerhq/live-common/e2e/enum/Currency";
-import { isWallet40Enabled } from "tests/utils/featureFlagUtils";
+import type { Application } from "tests/page";
+
+async function navigateToSubAccountAfterAdd(app: Application, tokenAccount: Account) {
```
VicAlbr: Is it possible to remove this function from the spec file ?

e2e/desktop/tests/page/settings.page.ts
```diff
@@ -103,6 +104,12 @@ export class SettingsPage extends AppPage {
 
   @step("Enable Wallet Sync")
   async enableWalletSync() {
+    if (await this.turnOnLedgerSyncButton.isVisible()) {
+      await this.turnOnLedgerSyncButton.click();
+      return;
```
martijnhjk: I can see how using return here leads to less indentation, however, it is easy to miss when reading the code. Even though it leads to more verbose and nested code I think using if ... else will be explicit and will correctly highlight the complexity of this step.

e2e/desktop/tests/specs/subAccount.spec.ts
martijnhjk: adding functions to spec files is generally not recommended for maintainability and complexity reasons.

please write the lines in this function directly in the step or if there is a reusable benefit add them to the respective page objects.

however, in this case the function is only used once so I think the contents can go straight into the step.

e2e/desktop/tests/specs/subAccount.spec.ts
```diff
@@ -20,7 +20,20 @@ import {
...
+  const parentAccountName = getParentAccountName(tokenAccount);
+
+  await app.mainNavigation.openTargetFromMainNavigation("accounts");
+  await app.accounts.navigateToAccountByName(parentAccountName);
+  await app.account.expectAccountVisibility(parentAccountName);
+  await app.account.navigateToTokenInAccount(tokenAccount);
+}
+
+async function openSendFlowFromPortfolio(app: Application) {
+  await app.portfolio.clickSendButton();
```
martijnhjk: see other comment about functions in specs.

please simply state the click action directly in the relevant tests.

e2e/desktop/tests/page/portfolio.page.ts
martijnhjk: or if we are abandoning legacy compatibility and always expect 4.0 we do not need this 'resolve' function and we can just always call the 4.0 row instead.

e2e/desktop/tests/page/accounts.page.ts
```diff
@@ -50,52 +41,52 @@ export class AccountsPage extends AppPage {
...
+        .getByTestId(`crypto-account-row-${this.getSanitizedAccountName(accountName)}`)
+        .filter({ visible: true }),
+    ).toHaveCount(0);
     expect(await this.getAccountsName()).not.toContain(accountName);
   }
 
   @step("Get number of accounts in the list")
   async countAccounts(): Promise<number> {
-    return await this.accountListNumber.count();
+    return await this.visibleAccountsList.count();
```
ypolishchuk-ledger, line 84: `.count()` returns 0 for an empty match, never `null` the assertion it's used in always passes regardless of actual state:

```
  // current                                                                                                                                                                                 
  expect(await this.countAccounts()).not.toBeNull();                                                                                                                                                       
                                                                                                                                                                                                         
  // should be                                                                                                                                                                                             
  expect(await this.countAccounts()).toBeGreaterThan(0);   
```

e2e/desktop/tests/page/drawer/ledger.sync.drawer.ts
```diff
@@ -58,7 +74,17 @@ export class LedgerSyncDrawer extends Drawer {
 
   @step("Check if sync entry point exists")
   async expectSyncAccountsButtonExist() {
-    await expect(this.continueButton).toBeVisible();
+    await expect
+      .poll(
+        async () => {
+          const hasContinue = await this.continueButton.isVisible();
+          const hasAlreadyTurnedOn = await this.alreadyTurnedOnButton.isVisible();
+          const hasConnectDevice = await this.walletSyncConnectDeviceButton.isVisible();
```
ypolishchuk-ledger: `walletSyncConnectDeviceButton` shows up after the user has already clicked the entry point - it's not the entry point itself. So if `expectSyncAccountsButtonExist()` passes only because `hasConnectDevice` is visible, we haven't actually confirmed the entry point is there; we just know the drawer is somewhere in the sync flow.

Also, the fast-path in `syncAccounts()` makes sense, and I get why `hasConnectDevice` is handled there. The awkward part is `expectSyncAccountsButtonExist` - the name implies we're checking for the entry point, but it also covers a state that's past it. Either rename it (e.g. "Wait for Ledger Sync drawer to be ready") or drop `hasConnectDevice` from the predicate here.

e2e/desktop/tests/specs/add.account.spec.ts
```diff
@@ -70,6 +70,8 @@ for (const currency of currencies) {
         await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
         const firstAccountName = `${currency.currency.name} 1`;
 
+        // Wait for portfolio empty state to load before clicking add account
+        await app.portfolio.checkNoBalanceTitleVisibility();
```
ypolishchuk-ledger: The inline comment correctly describes the intent (loading gate), but the step that appears in the Allure report will read as "Check no balance title is visible" - a scenario verification - making a failure here look like the test expected an empty portfolio rather than that the UI wasn't ready. Consider a dedicated `waitForPortfolioEmptyState()` method on `PortfolioPage`, or keep the comment at minimum so future readers understand this is synchronisation, not an assertion.

claudiiafg:
latest run: https://github.com/LedgerHQ/ledger-live/actions/runs/24672811678/job/72148162438

mcayuelas-ledger (approved):

ypolishchuk-ledger (approved):

PR #16449, opened 2026-04-17
[LWD] fix(desktop): prevent Live Apps from launching external OS apps via custom URL schemes [LIVE-29182]

Justkant:
### ✅ Checklist

  - desktop webviews

### 📝 Description

A malicious Live App could launch external OS applications (e.g. the App Store via `itms-apps://`) from inside its webview via several navigation vectors: embedding an iframe with a custom scheme, assigning such a URL to `window.location.href`, opening it through `window.open`, submitting a form, or — subtler — serving a trusted `https://` URL that returns an HTTP 3xx redirect to a custom scheme. Electron's JS navigation events (`will-navigate`, `will-frame-navigate`, `setWindowOpenHandler`, permission handlers, `shell.openExternal`, `webRequest`) are bypassed by Chromium's external-protocol handoff, so the existing guards never fired.

All the new logic lives in the dedicated [`apps/ledger-live-desktop/src/main/webviewHandlers.ts`](apps/ledger-live-desktop/src/main/webviewHandlers.ts) module, wired in from `main/index.ts` via a single `setupWebviewHandlers(SUPPORTED_SCHEMES)` call. Pure helpers (`createLiveAppSchemeChecker`, `mergeCspHeaders`, `WEBVIEW_GUEST_CSP`) are split out into [`webviewHandlers.helpers.ts`](apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts) with no `electron` imports so they can be unit-tested without mocking the runtime.

Hardening layers:

- **CSP injection on every webview session response.** Attached on `session-created` (plus `defaultSession`) via `webRequest.onHeadersReceived`, adding `frame-src` / `child-src` / `form-action 'self' http: https:`. Chromium enforces `frame-src` *before* the external-protocol handoff, which is the only layer capable of blocking `<iframe src="itms-apps://…">`. `form-action` covers `<form action="itms-apps://…">` submissions. Existing CSP headers from the Live App are preserved (browsers intersect multiple CSP headers, so stacking is strictly additive). `Content-Security-Policy-Report-Only` is left untouched.

- **Early navigation scheme guard, attached on `web-contents-created`** (not on `webview-dom-ready`, which is too late: `dom-ready` never fires for a page that immediately assigns `window.location.href = "itms-apps://…"`). A single guard function is registered on three events to cover every navigation vector that can trigger external-protocol handoff:
  - `will-navigate` — top-level navigations from link clicks / JS.
  - `will-redirect` — HTTP 3xx redirects, which don't re-trigger `will-navigate`.
  - `will-frame-navigate` — subframe navigations that `will-navigate` skips, belt-and-braces on top of the CSP `frame-src` policy.

- **`setWindowOpenHandler` also attached on `web-contents-created`** (previously registered on `webview-dom-ready`, which left an inline `<script>window.open("itms-apps://…")</script>` running during initial parse unguarded). The `new URL(url)` call is now wrapped in `try/catch` so malformed input is denied rather than throwing. http/https URLs are routed to the user's default browser via `openURL`; everything else is denied.

The scheme allow-list covers `http:`, `https:`, `about:`, and Ledger Live's own deep-link schemes (`ledgerlive:`, `ledgerwallet:`, sourced from `SUPPORTED_SCHEMES` so they cannot drift). The Ledger schemes are safe to allow because they are handled internally by Ledger Live itself — via `protocol.handle` in dev and `setAsDefaultProtocolClient` in production — and never result in a handoff to a third-party OS app.

Session hardening is deduplicated via a `WeakSet` so shared partitions don't register the header listener more than once, and the per-`WebContents` manifest-domain `will-navigate` handler is cleaned up on `destroyed` to avoid leaks. `setupWebviewHandlers` is documented as must-call-exactly-once.

### 🧪 Tests

31 unit tests in [`webviewHandlers.helpers.test.ts`](apps/ledger-live-desktop/src/main/webviewHandlers.helpers.test.ts) cover:

- Scheme checker: allowed (`https:`, `http:`, `about:blank`, `ledgerlive:`, `ledgerwallet:`), blocked (`itms-apps:`, `ms-word:`, `ms-notepad:`, `ms-paint:`, `file:`, `javascript:`, `mailto:`, `ftp:`), malformed input (empty string, free text, missing scheme), and the `supportedSchemes` argument.
- CSP header merge: no existing CSP, canonical / lowercase / mixed-case `Content-Security-Policy` key normalisation, multi-value preservation, input immutability, `Content-Security-Policy-Report-Only` untouched, `undefined` responseHeaders.
- CSP string invariants (`frame-src`, `child-src`, `form-action` present; no `itms-apps` / `ms-word` / `file:` tokens).

Manual smoke tests to run during QA: see the checklist shared in the thread (covers Live App `window.open("https://...")` still opening the system browser, `ledgerlive://` / `ledgerwallet://` deep links unchanged, and every attack vector above verified to be blocked).

### ❓ Context

- **JIRA or GitHub link**: [LIVE-29182]
- **ADR link (if any)**: 

---

[LIVE-29182]: https://ledgerhq.atlassian.net/browse/LIVE-29182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts
+++ b/apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts
@@ -1,0 +1,77 @@
+/**
+ * Pure helpers for the webview hardening layer. Kept free of any `electron`
+ * imports so they can be unit-tested without mocking the Electron runtime.
+ */
+
+/**
+ * Builds a scheme-allow checker for Live App <webview> navigations.
+ *
+ * Everything not in the allow list (itms-apps:, ms-word:, ms-notepad:,
+ * ms-paint:, file:, javascript:, ...) is rejected so the OS's external
+ * protocol handler cannot be triggered from within the webview via
+ * <iframe src>, window.location, <a href target=_blank>, HTTP 3xx redirects,
+ * etc.
+ *
+ * `ledgerlive:` / `ledgerwallet:` are allowed via `supportedSchemes`: these
+ * are Ledger Live's own deep-link schemes (registered via `protocol.handle`
+ * in dev mode and via `setAsDefaultProtocolClient` in production).
+ * Navigations to them are handled internally by Ledger Live itself, never by
+ * a third-party OS app, so they don't carry the external-handoff risk that
+ * motivated this guard.
+ */
+export function createLiveAppSchemeChecker(supportedSchemes: string[]): (url: string) => boolean {
+  const allowed = new Set([
+    "http:",
+    "https:",
+    "about:",
+    ...supportedSchemes.map(scheme => `${scheme}:`),
+  ]);
+
+  return (url: string): boolean => {
+    try {
+      return allowed.has(new URL(url).protocol);
+    } catch {
+      return false;
+    }
+  };
+}
+
+/**
+ * Merges an extra Content-Security-Policy value into an existing set of
+ * response headers without dropping any CSP already returned by the Live App.
+ *
+ * Browsers intersect multiple CSP headers, so stacking is safe and strictly
+ * additive - our injected directives can only further restrict, not relax,
+ * the effective policy.
+ *
+ * HTTP header names are case-insensitive; we normalise to the canonical
+ * `Content-Security-Policy` key on write and leave any
+ * `Content-Security-Policy-Report-Only` header untouched.
+ */
+export function mergeCspHeaders(
+  responseHeaders: Record<string, string[]> | undefined,
+  cspValue: string,
+): Record<string, string[]> {
+  const headers: Record<string, string[]> = { ...responseHeaders };
+
+  const existingCspKey = Object.keys(headers).find(
+    k => k.toLowerCase() === "content-security-policy",
+  );
+  const existingCsp = existingCspKey ? (headers[existingCspKey] ?? []) : [];
+  if (existingCspKey) delete headers[existingCspKey];
+  headers["Content-Security-Policy"] = [...existingCsp, cspValue];
+
+  return headers;
+}
+
+/**
+ * The CSP value injected on every guest <webview> document response.
+ *
+ * Scoped narrowly to the attack surface we care about (external-protocol
+ * handoff from framed content and form submissions). We intentionally do
+ * NOT set `default-src` / `script-src` here so we don't break Live App JS.
+ */
+export const WEBVIEW_GUEST_CSP =
+  "frame-src 'self' http: https:; " +
+  "child-src 'self' http: https:; " +
+  "form-action 'self' http: https:;";
--- a/apps/ledger-live-desktop/src/main/webviewHandlers.test.ts
+++ b/apps/ledger-live-desktop/src/main/webviewHandlers.test.ts
@@ -1,0 +1,187 @@
+import { app, ipcMain, session, webContents } from "electron";
+import { setupWebviewHandlers } from "./webviewHandlers";
+
+jest.mock("electron", () => ({
+  app: {
+    on: jest.fn(),
+  },
+  ipcMain: {
+    on: jest.fn(),
+  },
+  session: {
+    defaultSession: {
+      webRequest: {
+        onHeadersReceived: jest.fn(),
+      },
+    },
+  },
+  webContents: {
+    fromId: jest.fn(),
+  },
+}));
+
+type DomReadyHandler = (
+  event: Electron.IpcMainEvent,
+  id: number,
+  domains?: string[],
+) => void;
+type WebContentsCreatedHandler = (
+  event: Electron.Event,
+  contents: Electron.WebContents,
+) => void;
+type SchemeGuard = (event: Electron.Event<{ url: string }>) => void;
+
+describe("setupWebviewHandlers", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    delete (globalThis as typeof globalThis & { __ledgerLiveWebviewHandlersSetup__?: boolean })
+      .__ledgerLiveWebviewHandlersSetup__;
+  });
+
+  const getDomReadyHandler = () =>
+    (jest.mocked(ipcMain.on) as jest.Mock).mock.calls.find(
+      ([eventName]) => eventName === "webview-dom-ready",
+    )?.[1] as DomReadyHandler | undefined;
+
+  const getWebContentsCreatedHandler = () =>
+    (jest.mocked(app.on) as jest.Mock).mock.calls.find(
+      ([eventName]) => eventName === "web-contents-created",
+    )?.[1] as WebContentsCreatedHandler | undefined;
+
+  it("ignores webview-dom-ready events for non-webview contents", () => {
+    setupWebviewHandlers(["ledgerlive"]);
+
+    const domReadyHandler = getDomReadyHandler();
+
+    const nonWebviewContents = {
+      getType: jest.fn(() => "window"),
+      off: jest.fn(),
+      on: jest.fn(),
+    };
+    jest.mocked(webContents.fromId).mockReturnValue(nonWebviewContents as never);
+
+    domReadyHandler?.({} as Electron.IpcMainEvent, 42, ["https://app.example.com"]);
+
+    expect(nonWebviewContents.getType).toHaveBeenCalled();
+    expect(nonWebviewContents.off).not.toHaveBeenCalled();
+    expect(nonWebviewContents.on).not.toHaveBeenCalled();
+  });
+
+  it("attaches a will-navigate handler for webview contents", () => {
+    setupWebviewHandlers(["ledgerlive"]);
+
+    const domReadyHandler = getDomReadyHandler();
+
+    const webviewContents = {
+      getType: jest.fn(() => "webview"),
+      off: jest.fn(),
+      on: jest.fn(),
+    };
+    jest.mocked(webContents.fromId).mockReturnValue(webviewContents as never);
+
+    domReadyHandler?.({} as Electron.IpcMainEvent, 7, ["https://app.example.com"]);
+
+    expect(webviewContents.on).toHaveBeenCalledWith("will-navigate", expect.any(Function));
+    expect(webviewContents.off).not.toHaveBeenCalled();
+  });
+
+  it("registers its one-time listeners during setup", () => {
+    setupWebviewHandlers(["ledgerlive"]);
+
+    expect(jest.mocked(app.on)).toHaveBeenCalledWith("session-created", expect.any(Function));
+    expect(jest.mocked(app.on)).toHaveBeenCalledWith("web-contents-created", expect.any(Function));
+    expect(jest.mocked(session.defaultSession.webRequest.onHeadersReceived)).toHaveBeenCalledWith(
+      expect.any(Function),
+    );
+  });
+
+  it("attaches will-navigate, will-redirect and will-frame-navigate handlers for webview contents", () => {
+    setupWebviewHandlers(["ledgerlive"]);
+
+    const webContentsCreatedHandler = getWebContentsCreatedHandler();
+    const guestContents = {
+      id: 7,
+      getType: jest.fn(() => "webview"),
+      on: jest.fn(),
+      once: jest.fn(),
+      setWindowOpenHandler: jest.fn(),
+    };
+
+    webContentsCreatedHandler?.(
+      {} as Electron.Event,
+      guestContents as unknown as Electron.WebContents,
+    );
+
+    expect(guestContents.on).toHaveBeenCalledWith("will-navigate", expect.any(Function));
+    expect(guestContents.on).toHaveBeenCalledWith("will-redirect", expect.any(Function));
+    expect(guestContents.on).toHaveBeenCalledWith("will-frame-navigate", expect.any(Function));
+  });
+
+  it.each(["will-navigate", "will-redirect", "will-frame-navigate"])(
+    "prevents %s for disallowed schemes",
+    eventName => {
+      setupWebviewHandlers(["ledgerlive"]);
+
+      const webContentsCreatedHandler = getWebContentsCreatedHandler();
+      const guestContents = {
+        id: 7,
+        getType: jest.fn(() => "webview"),
+        on: jest.fn(),
+        once: jest.fn(),
+        setWindowOpenHandler: jest.fn(),
+      };
+
+      webContentsCreatedHandler?.(
+        {} as Electron.Event,
+        guestContents as unknown as Electron.WebContents,
+      );
+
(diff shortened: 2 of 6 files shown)


Justkant:
There is still some improvements to do IMO, after a review by GPT-5.4:
> Medium: wc.once("destroyed", ...) is now registered on every webview-dom-ready, so reloads stack one-shot listeners until the webview dies. That is a small listener leak pattern and easy to avoid by attaching destruction cleanup once in web-contents-created or by tracking whether cleanup was already installed for that id.

> Low: the CSP injection currently runs on every response in every guest webview session, even though the policy only needs to be added to document responses. Narrowing it to at least mainFrame would reduce blast radius and make the intent clearer.

Justkant:
Here's a QA smoke-test checklist. Each row is runnable end-to-end against a dev build (`pnpm dev:lld`) unless marked otherwise.

## Prerequisites

- Dev build running via `pnpm dev:lld`.
- DevTools open on the Live App webview (right-click inside the app → Inspect, or use the host window's Discover menu to open the webview inspector).
- A simple test Live App you can point at a controllable host. The quickest way is to run a local static server (`npx serve .`) that returns HTML you can edit, and load it as a custom dev manifest. If that's not possible, run the checks from the console of any existing Live App by pasting the snippets.

## Smoke tests

| # | Area | Scenario | Steps | Expected | Regression risk |
|---|------|----------|-------|----------|-----------------|
| 1 | Baseline | Live App loads normally | Open any Live App (e.g. Discover → pick one). | App renders, no console errors about CSP violations for normal content, no infinite spinner. | CSP injection breaking legit framed content. |
| 2 | Deep link | `ledgerlive://` still works from inside a Live App | In the webview DevTools console run: `window.location.href = "ledgerlive://discover"`. | Ledger Live main window focuses and routes to Discover. No "blocked" error. | Scheme guard over-blocking our own deep-link schemes. |
| 3 | Deep link | `ledgerwallet://` still works from inside a Live App | Console: `window.location.href = "ledgerwallet://discover"`. | Same as #2. | Same. |
| 4 | `window.open` http/https | External links open in system browser | Console: `window.open("https://example.com")`. | Default OS browser opens `example.com`. The Live App webview does NOT navigate. No new Electron window. | `setWindowOpenHandler` moved to `web-contents-created` must still route http(s) to `openURL`. |
| 5 | `window.open` disallowed scheme | App Store handoff is blocked | Console: `window.open("itms-apps://itunes.apple.com/app/id1234")`. | Nothing happens. App Store does NOT open. No new Electron window. No crash. | `setWindowOpenHandler` deny branch. |
| 6 | `window.open` malformed URL | Malformed input does not crash | Console: `window.open("not a url")`. | Nothing happens. Live App still responsive. No unhandled exception in the main process log. | New `try/catch` around `new URL(url)`. |
| 7 | Top-level navigation to disallowed scheme | `window.location` assignment blocked | Console: `window.location.href = "itms-apps://itunes.apple.com/app/id1234"`. | Webview stays on current page. App Store does NOT open. | `will-navigate` scheme guard. |
| 8 | Top-level navigation, early script | Inline script before `dom-ready` is blocked | Serve a page whose `<head>` contains `<script>window.location.href="itms-apps://id1234"</script>`. Navigate the Live App to it. | App Store does NOT open. The page either stays blank or shows the loaded content (depending on where the script runs). This is the specific case `dom-ready`-based handlers used to miss. | `setWindowOpenHandler` + scheme guard attached on `web-contents-created`. |
| 9 | HTTP 3xx redirect bypass (main fix) | Redirect chain ending in `itms-apps://` blocked | Serve a URL that returns `HTTP/1.1 302` with `Location: itms-apps://itunes.apple.com/app/id1234`. Navigate the Live App to it. | App Store does NOT open. Webview stays on previous page (or errors cleanly). | New `will-redirect` handler. |
| 10 | Iframe navigation | `<iframe src="itms-apps://...">` blocked | In the Live App, inject: `const f=document.createElement("iframe"); f.src="itms-apps://id1234"; document.body.appendChild(f);`. | App Store does NOT open. DevTools Console shows a CSP `frame-src` violation. | CSP `frame-src` + `will-frame-navigate` belt-and-braces. |
| 11 | Iframe redirect | Iframe navigating to `itms-apps://` via `location` blocked | In the Live App: `const f=document.createElement("iframe"); f.src="https://example.com"; document.body.appendChild(f);` — then once loaded: `f.contentWindow.location.href="itms-apps://id1234"`. | App Store does NOT open. | `will-frame-navigate` handler. |
| 12 | Form action bypass | `<form action="itms-apps://...">` submission blocked | In the Live App: `const f=document.createElement("form"); f.action="itms-apps://id1234"; f.method="POST"; document.body.appendChild(f); f.submit();`. | App Store does NOT open. DevTools Console shows CSP `form-action` violation. | New `form-action` CSP directive + `will-navigate`. |
| 13 | `javascript:` URL | `javascript:` blocked | Console: `window.open("javascript:alert(1)")`. Also: `window.location.href="javascript:alert(1)"` should NOT trigger the alert from a cross-origin link, but a same-origin call is expected to run (this is Chromium behaviour, not our guard) — only verify `window.open` is denied. | `window.open("javascript:...")` returns null / no new window. | `window.open` deny path. |
| 14 | `file:` URL | Local file navigation blocked | Console: `window.open("file:///etc/passwd")` and `window.location.href="file:///etc/passwd"`. | No new window, webview stays put. | Scheme guard. |
| 15 | Live App with own CSP | Existing CSP is preserved | Pick a Live App you know ships its own `Content-Security-Policy` header. Open DevTools → Network → click the document response. | Response shows at least two `Content-Security-Policy` headers (the Live App's original + ours). Existing functionality unchanged. | `mergeCspHeaders` preserves existing values. |
| 16 | Manifest domain whitelist | Feature flag unchanged | Enable feature flag `lldWebviewManifestDomainCheck`. Open a Live App whose manifest has a `domains` whitelist. Try to navigate to an off-whitelist URL via `window.location.href = "https://other.com"`. | Navigation is blocked, same as before this branch's work. | `will-navigate` handler registered via `webview-dom-ready` IPC. |
| 17 | Webview reload | Manifest-domain handler does not leak on reload | With feature flag on, reload the webview 3–5 times (Cmd-R while the webview has focus, or click the app's internal reload button). Inspect `webContents` listener counts in the main process (or just verify behaviour). | Behaviour unchanged across reloads. No duplicate blocking, no memory growth warning, no "MaxListenersExceededWarning" in the main process log. | `willNavigateHandlers` lifecycle (add/remove/delete-on-destroyed). |
| 18 | Multiple Live Apps | Partitioned sessions all get CSP | Open two different Live Apps in sequence (they use partitioned sessions). Check the document response headers for each in DevTools Network. | Both show our injected CSP. Neither breaks. | `session-created` listener + `hardenedSessions` WeakSet. |
| 19 | Host window unaffected | Main renderer not touched | Open the main Ledger Live DevTools (the host, not a Live App). Check Network headers on the renderer's own document. | No extra `Content-Security-Policy` from our handler. All normal app features work. | `details.webContents?.getType() === "webview"` gate. |
| 20 | Platform parity | Same checks on Windows and Linux | Re-run #2, #4, #5, #9, #12 on Windows and on Linux dev builds. | Same behaviour as macOS. | External-protocol handoff differs per OS (e.g. `ms-word:` is the Windows analogue of `itms-apps:`). On Windows, swap #5/#9 target to `ms-word://open`. |

## Console snippets for quick copy-paste

Drop these into the Live App DevTools console for tests #2, #5, #7, #10, #12, #14:

```js
window.location.href = "ledgerlive://discover";

window.open("itms-apps://itunes.apple.com/app/id1234");

window.location.href = "itms-apps://itunes.apple.com/app/id1234";

const f = document.createElement("iframe");
f.src = "itms-apps://id1234";
document.body.appendChild(f);

const form = document.createElement("form");
form.action = "itms-apps://id1234";
form.method = "POST";
document.body.appendChild(form);
form.submit();

window.open("file:///etc/passwd");
```

## What a pass looks like overall

- No Live App is broken compared to `main`.
- No console error mentioning `setWindowOpenHandler`, `onHeadersReceived`, or "Maximum call stack" in the main-process log.
- Every `itms-apps:` / `ms-word:` / `file:` / `javascript:` attempt ends with nothing happening — no OS-level app launch, no toast, no redirect to a blank screen.
- Our own `ledgerlive://` / `ledgerwallet://` deep links still work.

apps/ledger-live-desktop/src/main/webviewHandlers.ts
```diff
@@ -0,0 +1,161 @@
...
+    // All three events expose the target URL directly on the event object
+    // in modern Electron, so a single guard function covers every signature.
+    const schemeGuard = (event: Electron.Event<{ url: string }>) => {
+      if (!isLiveAppSchemeAllowed(event.url)) {
+        event.preventDefault();
+      }
+    };
+    contents.on("will-navigate", schemeGuard);
+    contents.on("will-redirect", schemeGuard);
+    contents.on("will-frame-navigate", schemeGuard);
```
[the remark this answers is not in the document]
Justkant: This is not correct for our Electron version

Electron’s typings show these listeners receive a first `Event<...>` object that already includes the navigation params, including `url`, and `will-navigate` / `will-redirect` only keep the old positional `url` argument as a deprecated extra.

So `event.url` in `schemeGuard` is valid, and the current shared handler does not block all navigation.

Justkant:
Desktop E2E run: https://github.com/LedgerHQ/ledger-live/actions/runs/24784939084

Justkant:
Desktop e2e run: https://github.com/LedgerHQ/ledger-live/actions/runs/24788529987

apps/ledger-live-desktop/src/main/webviewHandlers.ts
```diff
@@ -0,0 +1,173 @@
...
+  app.on("web-contents-created", (_appEvent, contents) => {
+    if (contents.getType() !== "webview") return;
+
+    // Route same-tab `window.open` attempts: http(s) goes to the user's
+    // default browser via `openURL`; everything else (including
+    // `itms-apps:`, `javascript:`, malformed URLs) is denied.
+    contents.setWindowOpenHandler(({ url }) => {
+      try {
+        const protocol = new URL(url).protocol;
+        if (protocol === "https:" || protocol === "http:") {
```
ysitbon, line 136: Do we have a case where we could have `ledgerlive:` here?

apps/ledger-live-desktop/src/main/webviewHandlers.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    const webviewContents = {
+      getType: jest.fn(() => "webview"),
+      off: jest.fn(),
+      on: jest.fn(),
+    };
+    jest.mocked(webContents.fromId).mockReturnValue(webviewContents as never);
+
+    domReadyHandler?.({} as Electron.IpcMainEvent, 7, ["https://app.example.com"]);
+
+    expect(webviewContents.on).toHaveBeenCalledWith("will-navigate", expect.any(Function));
```
ysitbon, line 84: Maybe we can add the missing ones `"will-redirect"` and `"will-frame-navigate"`?

apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * The CSP value injected on every guest <webview> document response.
+ *
+ * Scoped narrowly to the attack surface we care about (external-protocol
+ * handoff from framed content and form submissions). We intentionally do
+ * NOT set `default-src` / `script-src` here so we don't break Live App JS.
+ */
+export const WEBVIEW_GUEST_CSP =
+  "frame-src 'self' http: https: data: blob:; " +
+  "child-src 'self' http: https: data: blob:; " +
```
ysitbon: Is `data:` required, since this allow `data:text/html,<script>...</script>`? If yes it worth to document it too no?

apps/ledger-live-desktop/src/main/webviewHandlers.ts
Justkant: Yes some live-apps are not using the custom handler we provide to go to other live-apps and use the deeplink instead

Justkant: And they don't use the window.open for the deeplink but a simple `location.href = "ledgerlive://"`

apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts
Justkant: data is used for icons sometimes so I think it can be a legitimate use but I can see if removing it is fine, probably same for blob

Justkant: I've removed that

apps/ledger-live-desktop/src/main/webviewHandlers.ts
Justkant: So no we don't really expect any ledgerlive: here and it's the same logic as before the refactor in this file so no changes in this PR

apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts
```diff
@@ -0,0 +1,77 @@
...
+ * a third-party OS app, so they don't carry the external-handoff risk that
+ * motivated this guard.
+ */
+export function createLiveAppSchemeChecker(supportedSchemes: string[]): (url: string) => boolean {
+  const allowed = new Set([
+    "http:",
+    "https:",
+    "about:",
+    ...supportedSchemes.map(scheme => `${scheme}:`),
+  ]);
```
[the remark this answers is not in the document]
Justkant: Updated the PR description

apps/ledger-live-desktop/src/main/webviewHandlers.helpers.ts
```diff
@@ -0,0 +1,77 @@
...
+ * The CSP value injected on every guest <webview> document response.
+ *
+ * Scoped narrowly to the attack surface we care about (external-protocol
+ * handoff from framed content and form submissions). We intentionally do
+ * NOT set `default-src` / `script-src` here so we don't break Live App JS.
+ */
+export const WEBVIEW_GUEST_CSP =
+  "frame-src 'self' http: https:; " +
+  "child-src 'self' http: https:; " +
+  "form-action 'self' http: https:;";
```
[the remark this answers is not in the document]
Justkant: Updated the PR description

LL782 (approved):
Looks sane to me

PR #16465, opened 2026-04-17
[LWDM] fix(coin-evm): drop non-value-transferring internals and dedup root traces in getBlock

adussarps:
## Context

`getBlock` in `@ledgerhq/coin-evm` was surfacing every EVM frame reported by
Blockscout (and Etherscan-like explorers) as an independent native operation.
Downstream indexers like A4 add operations to compute balances, so this
double-counting drove balances negative — in particular on Somnia
(**BACK-11037**, where reference accounts ended up as `BalanceValue.Invalid`).

Two orthogonal issues were at play:

1. **Root-trace duplicate.** The first internal transaction returned by the
   explorer for each tx is the top-level call, with `from` / `to` / `value`
   identical to the coin tx itself. Merging it naively emitted the same native
   transfer twice.
2. **Non-value-transferring frames.** `delegatecall`, `staticcall`, and
   `callcode` inherit `msg.value` from their enclosing frame but cannot move
   native ETH. Blockscout still reports a non-zero `value` on those frames,
   which meant the proxy → implementation `delegatecall` was converted into a
   third ghost OUT for every proxy interaction.

A single proxy-contract call (e.g. `accountId2`'s swaps through the Somnia
router) could therefore emit up to **3 identical OUTs** for the same native
transfer.

## Why filtering these call types is safe

The EVM has four `CALL`-family opcodes. Only **one** of them can ever move
native ETH:

| Opcode         | Storage context        | `msg.sender` seen by callee | Moves ETH? | Purpose                                                                  |
|----------------|------------------------|-----------------------------|------------|--------------------------------------------------------------------------|
| `CALL`         | callee's               | the caller                  | **YES**    | Regular call; the only opcode whose `value` field actually moves ether   |
| `DELEGATECALL` | **caller's**           | preserved (caller's sender) | NO         | Proxy pattern: borrow implementation's code, execute in proxy's storage  |
| `STATICCALL`   | callee's (read-only)   | the caller                  | NO         | View calls; any state change (SSTORE, LOG, CREATE, value-bearing CALL…) reverts |
| `CALLCODE`     | **caller's**           | the caller                  | NO         | Deprecated precursor to `DELEGATECALL` (EIP-7); same storage semantics   |

By protocol, `DELEGATECALL` / `STATICCALL` / `CALLCODE` **cannot** transfer
native value — the VM never touches balances in those frames. The non-zero
`value` the explorers report on them is the `msg.value` **inherited** from the
enclosing call, copied into the frame's `value` field for display purposes.
No balance movement actually happened.

A nested real transfer inside a proxy call still appears separately as a
`CALL` frame in the trace — we keep those. We only drop the wrapper frames
that, by EVM semantics, are guaranteed not to represent a transfer.

Concretely for the Somnia proxy example:

- `CALL` user → proxy, `value = X`   ← root trace (dedupped as the coin tx itself)
- `DELEGATECALL` proxy → implementation, `value = X` inherited   ← **dropped: code-only frame**
- `CALL` proxy → recipient, `value = X`   ← **kept: actual native transfer**

## Fix

Layered defence, so a future explorer variant that leaks the root call in an
unexpected shape still gets caught:

- **Semantic filter** in the adapters, via the shared
  `NON_VALUE_TRANSFER_CALL_TYPES = {delegatecall, staticcall, callcode}`:
  - `internalTxsToOperationsByHash` (Etherscan / Blockscout, `getBlock`
    path) skips entries whose `callType` (Blockscout) or `type` (Etherscan)
    is in the set.
  - `traceBlockItemsToOperationsByHash` (RPC `trace_block`, `getBlock`
    fallback) applies the same filter to `action.callType`.
  - `etherscanInternalTransactionToOperations` (`listOperations` path)
    applies the same filter at its entrypoint, so Blockscout-backed Ledger
    Live account history stops surfacing phantom IN/OUT rows.
  - `EtherscanInternalTransaction` gains an optional `callType?` field since
    Blockscout exposes the discriminator separately from `type`.
- **Structural dedup** in `mergeInternalTransactions`: a new
  `dropRootTraceDuplicates` helper drops any internal native op whose
  `(address, peer, amount)` tuple matches one of the coin tx's own native
  ops. ERC20/721/1155 ops are never touched — they come from receipt logs,
  not from trace data, so the two sources never overlap.

Both layers are needed: the `delegatecall` frame on the Somnia proxy example
has a different `to` (implementation) than the coin tx's `to` (proxy), so the
structural dedup alone doesn't catch it; conversely the semantic filter alone
doesn't catch root-trace duplicates on explorers that don't expose `callType`.

## Tests

New tests, all in `libs/coin-modules/coin-evm/src/`:

- `logic/getBlock.test.ts`
  - 2 integration tests: root-trace duplicate dropped; nested subcall kept
    (with equal amounts but different peers, exercising the `(address, peer,
    amount)` key).
  - 7 unit tests on `dropRootTraceDuplicates` (exact match, peer/amount
    differs, case-insensitive addresses, ERC20 untouched, empty coin ops,
    undefined-peer handling).
- `adapters/etherscan.test.ts` — 6 parametric tests (`delegatecall` /
  `staticcall` / `callcode` via `callType` and via `type`) + case-insensitive
  discriminator.
- `adapters/blockOperations.test.ts` — 3 parametric tests (same call types
  via `action.callType`).

Local run: `pnpm test` → 2542 passing, `pnpm typecheck` clean, `pnpm lint` 0
errors (existing develop warnings untouched). Line coverage on touched files
is ≥ 96 %.

## Verification

Validated end-to-end against a locally-running `alpaca-coin-module` linked to
this branch (recipe in A4's
[`fast-coin-evm-dev-loop` skill](https://github.com/LedgerHQ/a4/blob/back-11037-alpaca-local-dev-fixture/.agents/skills/fast-coin-evm-dev-loop/SKILL.md))
on Somnia block `103014295`: the proxy interaction that previously produced
6 native ops now correctly produces 4.

Residual Somnia negative balances on `accountId2` are traceable to a separate
explorer data-completeness issue (IN/OUT mismatch on the Blockscout side)
unrelated to this fix; details in the A4-side investigation at
`docs/investigations/2026-04-16-somnia-native-balance.md` on the same branch.

### DIFF
--- a/libs/coin-modules/coin-evm/src/adapters/etherscan.test.ts
+++ b/libs/coin-modules/coin-evm/src/adapters/etherscan.test.ts
@@ -1705,74 +1705,57 @@ describe("EVM Family", () => {
         // Blockscout reports `delegatecall`/`staticcall`/`callcode` internal frames with an
         // inherited `msg.value`, but those opcodes cannot move native ETH. Emitting operations
         // for them would surface phantom IN/OUT ops in the user's history.
+        const accountIdForCallType = encodeAccountId({
+          type: "js",
+          version: "2",
+          currencyId: "ethereum",
+          xpubOrAddress: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
+          derivationMode: "",
+        });
+        const makeInternalTx = (overrides: {
+          type?: string;
+          callType?: string;
+        }): EtherscanInternalTransaction => ({
+          blockNumber: "14878012",
+          timeStamp: "1653990239",
+          hash: "0xb3effb3b6c52c719507f8219fe0dd2147a9f7ba366261ab43532efb0b9b01885",
+          from: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
+          to: "0xdef171fe48cf0115b1d80b88dc8eab59176fee57",
+          value: "66616263350003",
+          contractAddress: "",
+          input: "",
+          type: overrides.type ?? "call",
+          gas: "129878",
+          gasUsed: "0",
+          traceId: "0_1",
+          isError: "0",
+          errCode: "",
+          ...(overrides.callType !== undefined ? { callType: overrides.callType } : {}),
+        });
+
         it.each([
-          { field: "callType", value: "delegatecall" },
-          { field: "callType", value: "staticcall" },
-          { field: "callType", value: "callcode" },
-          { field: "type", value: "delegatecall" },
-          { field: "type", value: "staticcall" },
-          { field: "type", value: "callcode" },
-        ])(
-          "returns no operations when $field is $value (non-value-transferring call type)",
-          ({ field, value }) => {
-            const accountId = encodeAccountId({
-              type: "js",
-              version: "2",
-              currencyId: "ethereum",
-              xpubOrAddress: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
-              derivationMode: "",
-            });
-
-            const etherscanOp: EtherscanInternalTransaction = {
-              blockNumber: "14878012",
-              timeStamp: "1653990239",
-              hash: "0xb3effb3b6c52c719507f8219fe0dd2147a9f7ba366261ab43532efb0b9b01885",
-              from: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
-              to: "0xdef171fe48cf0115b1d80b88dc8eab59176fee57",
-              value: "66616263350003",
-              contractAddress: "",
-              input: "",
-              type: field === "type" ? value : "call",
-              gas: "129878",
-              gasUsed: "0",
-              traceId: "0_1",
-              isError: "0",
-              errCode: "",
-              ...(field === "callType" ? { callType: value } : {}),
-            };
-
-            expect(etherscanInternalTransactionToOperations(accountId, etherscanOp)).toEqual([]);
-          },
-        );
+          { callType: "delegatecall" },
+          { callType: "staticcall" },
+          { callType: "callcode" },
+          { type: "delegatecall" },
+          { type: "staticcall" },
+          { type: "callcode" },
+        ])("returns no operations for non-value-transferring call type %o", overrides => {
+          expect(
+            etherscanInternalTransactionToOperations(
+              accountIdForCallType,
+              makeInternalTx(overrides),
+            ),
+          ).toEqual([]);
+        });
 
         it("is case-insensitive on the call type discriminator", () => {
-          const accountId = encodeAccountId({
-            type: "js",
-            version: "2",
-            currencyId: "ethereum",
-            xpubOrAddress: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
-            derivationMode: "",
-          });
-
-          const etherscanOp: EtherscanInternalTransaction = {
-            blockNumber: "14878012",
-            timeStamp: "1653990239",
-            hash: "0xb3effb3b6c52c719507f8219fe0dd2147a9f7ba366261ab43532efb0b9b01885",
-            from: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
-            to: "0xdef171fe48cf0115b1d80b88dc8eab59176fee57",
-            value: "66616263350003",
-            contractAddress: "",
-            input: "",
-            type: "call",
-            callType: "DelegateCall",
-            gas: "129878",
-            gasUsed: "0",
-            traceId: "0_1",
-            isError: "0",
-            errCode: "",
-          };
-
-          expect(etherscanInternalTransactionToOperations(accountId, etherscanOp)).toEqual([]);
+          expect(
+            etherscanInternalTransactionToOperations(
+              accountIdForCallType,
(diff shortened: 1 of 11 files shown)


adussarps:
Thanks for the review! Addressed all 5 comments in 9d082db (plus a bit of JSDoc polish from a self-review pass):

1. `dropRootTraceDuplicates` is no longer re-exported. Moved it (and `signatureOf`) to a new `logic/rootTraceDedup.ts` which isn't listed in `logic/index.ts`'s barrel. Tests import it directly from the new module.
2. Restored the stricter `BigInt(it.value) <= 0n` check in `isInternalTransactionValid` — well spotted, that was an accidental loosening.
3. Extracted `NON_VALUE_TRANSFER_CALL_TYPES` to `adapters/nonValueTransferCallTypes.ts`, imported from both `etherscan.ts` and `blockOperations.ts`. Not re-exported from `adapters/index.ts` either.
4. Reordered imports in `getBlock.test.ts` so the external type-only import (`BlockOperation`) comes first.
5. Fixed the `mergeInternalTransactions` docstring to match the actual `(address, peer, amount)` match key — also expanded the `dropRootTraceDuplicates` JSDoc to explain why excluding `asset` is sound and why the structural dedup can't over-drop (only the root frame has `from == EOA`, and the EVM opcodes that would break that invariant are the ones filtered upstream).

Tests and typecheck green locally; ready for another pass.

adussarps:
Follow-up after a second self-review pass — extended the fix + tests:

**New in d2deedf** — `listOperations` had the same gap: `etherscanInternalTransactionToOperations` (the mapper used by `getInternalOperations` → `listOperations`) had no `callType` filter, so on Blockscout-backed chains a `delegatecall`/`staticcall`/`callcode` internal frame would surface as a phantom IN/OUT in Ledger Live Desktop/Mobile account history. Applied the same `NON_VALUE_TRANSFER_CALL_TYPES` guard at the top of the mapper (defense-in-depth for any future caller of the mapper too).

**Extra `dropRootTraceDuplicates` edge cases** covered:
- Mixed native + ERC20 internal ops where both tuple-match a coin native — only the native entry is dropped, ERC20 passes through.
- Undefined-peer symmetry: two ops both without `peer` still dedup; one with vs one without do NOT collapse.

**Tests added**: 7 for the new `listOperations` path filter (6 parametric `callType`/`type` × 3 opcodes, + 1 case-insensitivity), 3 new `dropRootTraceDuplicates` edge cases. `pnpm test` → 251 passing across the four relevant files, typecheck clean, 0 lint errors.

**Changeset** updated to call out the `listOperations` fix explicitly.

Happy to re-resolve or iterate further.

libs/coin-modules/coin-evm/src/adapters/etherscan.test.ts
```diff
@@ -1701,6 +1701,79 @@ describe("EVM Family", () => {
...
+        it("is case-insensitive on the call type discriminator", () => {
+          const accountId = encodeAccountId({
+            type: "js",
+            version: "2",
+            currencyId: "ethereum",
+            xpubOrAddress: "0x6cbcd73cd8e8a42844662f0a0e76d7f79afd933d",
+            derivationMode: "",
+          });
+
+          const etherscanOp: EtherscanInternalTransaction = {
```
jprudent: could have a builder utility for EtherscanInternalTransaction with param {type?, callType?} => it visually helps to see what is under test

libs/coin-modules/coin-evm/src/logic/rootTraceDedup.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Why the structural dedup cannot over-drop: in an EVM trace, every nested call frame has
+ * `from = <currently-executing contract>`, so a nested native transfer cannot satisfy both
+ * `address == tx.originator` and `amount == tx.value` unless it is literally the root-trace
+ * duplicate of the coin tx itself. The only EVM opcodes that would violate this invariant
+ * (`delegatecall` / `staticcall` / `callcode`, which preserve the caller's `from`) are
+ * filtered semantically upstream in the adapters, so they never reach this point.
+ *
+ * Intentionally not re-exported from `logic/index.ts` — implementation detail of `getBlock`.
+ */
+export function dropRootTraceDuplicates(
```
jprudent, line 22: edge case: what if there are several internal transfer of same amount/sender/recipient that matches native ops ?

libs/coin-modules/coin-evm/src/adapters/etherscan.test.ts
adussarps: done in a60967d

libs/coin-modules/coin-evm/src/logic/rootTraceDedup.ts
adussarps: In theory; it can't happen within one tx: signature uses the *signed* amount, and only the root frame has an EOA as `from`. any nested op touching the EOA can only be an IN (`+amount`), while the coin-side op is an OUT (`-amount`) — no collision possible. the only thing that ever shares a full signature with a coin op is the root-trace duplicate itself, which is unique by construction.

jprudent (approved):

francois-guerin-ledger (approved):

PR #16504, opened 2026-04-20
docs: migrate to commands to skills

LL782:
### 📝 Description

These changes are the result of applying Cursors `/migrate-to-skills` skill:
- All **commands** are migrated to **agent-skills** with `disable-model-invocation: true`
- All **rules** that were "applied intelligently" migrated to **agent-skills**

### ❓ Context

Jira: https://ledgerhq.atlassian.net/browse/LIVE-29279
ADR: https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7035977860

### DIFF
--- a/.cursor/agents/code-reviewer.md
+++ b/.cursor/agents/code-reviewer.md
@@ -10,8 +10,8 @@ You are a senior code reviewer for Ledger Wallet applications, specializing in m
 Follow all project rules in `.cursor/rules/`. Pay special attention to:
 
 - `react-mvvm.mdc` — MVVM compliance is a key review concern for code in `src/mvvm/`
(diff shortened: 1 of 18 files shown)


mise.lock
LucasWerey, line 1: You shouldn't have that. I think it's because you have a different bun version as the one specified by mise

deepyjr (dismissed):

couto (dismissed):

mise.lock
LL782: thanks @LucasWerey! I've removed that commit

mcayuelas-ledger (approved):

couto (approved):

PR #16508, opened 2026-04-20
[LWDM] test(live-common): add coin-module eager-import non-regression tests

gre-ledger:
### ✅ Checklist

  - New tests only — no runtime behaviour changed.

### 📝 Description

Adds a bundle-level non-regression test (`no-coin-eager-imports.test.ts`) that
uses esbuild's metafile to assert that `@ledgerhq/coin-*` packages (other than
`coin-module-framework`) are **not** eagerly imported outside the designated zones:
`families/`, `coin-modules/`, `bridge/generic-alpaca/`, `wallet-api/`.

A dedicated esbuild entry-point (`non-regression-entry.ts`) re-exports all shared
files that currently use `load*ForFamily` helpers, so the bundle graph is stable.

Known violations are listed with ticket references and will be removed as each
file is migrated (tracked under LIVE-29326).

Also adds `loaders.test.ts` — a smoke test that calls every method on every
`CoinModuleLoader` entry and asserts it does not throw.

Fixes `registry.test.ts` typecheck by importing the three existing but
unimported helpers and removing seven loader entries whose backing
`CoinModuleLoader` keys do not exist yet in the type.

### ❓ Context

- **JIRA or GitHub link**: preparing code for https://github.com/LedgerHQ/ledger-live/pull/16470

### DIFF
--- a/libs/ledger-live-common/src/coin-modules/loaders.test.ts
+++ b/libs/ledger-live-common/src/coin-modules/loaders.test.ts
@@ -1,0 +1,14 @@
+import { coinModuleLoaders } from "./loaders";
+
+describe("coinModuleLoaders smoke test", () => {
+  for (const loader of coinModuleLoaders) {
+    const entries = Object.entries(loader).flatMap(([k, fn]) =>
+      k !== "family" && typeof fn === "function" ? [[k, fn] as const] : [],
+    );
+    for (const [method, fn] of entries) {
+      it(`${loader.family}.${method}`, () => {
+        expect(() => fn()).not.toThrow();
+      });
+    }
+  }
+});
--- a/libs/ledger-live-common/src/coin-modules/no-coin-eager-imports.test.ts
+++ b/libs/ledger-live-common/src/coin-modules/no-coin-eager-imports.test.ts
@@ -1,0 +1,101 @@
+import * as esbuild from "esbuild";
+import * as path from "path";
+
+// Sources allowed to import coin-* packages or families/* directly.
+const ALLOWED_SOURCES = [
+  "ledger-live-common/src/families/",
+  "ledger-live-common/src/coin-modules/",
+  "ledger-live-common/src/bridge/generic-alpaca/", // LIVE-29339
+  "ledger-live-common/src/wallet-api/", // LIVE-29338
+];
+
+// Import paths with minimal bundle impact — types and error definitions only.
+const LOW_IMPACT_PATH = /\/(types|errors)(\/|$)/;
+
+let metafileInputs: esbuild.Metafile["inputs"];
+const root = path.resolve(__dirname, "../../..");
+const rel = (p: string) => path.relative(root, p).replace(/\\/g, "/");
+
+beforeAll(async () => {
+  const result = await esbuild.build({
+    entryPoints: [path.resolve(__dirname, "non-regression-entry.ts")],
+    bundle: true,
+    write: false,
+    packages: "external",
+    platform: "node",
+    format: "cjs",
+    metafile: true,
+  });
+  metafileInputs = result.metafile.inputs;
+}, 15000);
+
+function buildCausesMap(predicate: (pkg: string) => boolean): Record<string, string[]> {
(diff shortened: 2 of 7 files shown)


libs/ledger-live-common/src/coin-modules/no-coin-eager-imports.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  const violations: Record<string, string[]> = {};
+  for (const [file, info] of Object.entries(metafileInputs)) {
+    const src = path.relative(root, file);
+    if (ALLOWED_SOURCES.some(p => src.includes(p))) continue;
+    for (const imp of info.imports) {
+      const impRel = path.relative(root, imp.path);
+      if (!impRel.includes("ledger-live-common/src/families/")) continue;
+      if (KNOWN[impRel]) continue;
+      (violations[impRel] ??= []).push(src);
+    }
```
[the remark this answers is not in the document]
gre-ledger: we don't run tests on windows.

libs/ledger-live-common/src/coin-modules/registry.test.ts
```diff
@@ -31,65 +34,38 @@ describe("registerCoinModules / getRegisteredFamilies", () => {
...
+    });
+    if (required) {
+      it("throws CurrencyNotSupported for unknown family", () => {
+        expect(() => fn("__none__")).toThrow(CurrencyNotSupported);
+      });
+    } else {
+      it("returns undefined for unknown family", () => {
+        expect(fn("__none__")).toBeUndefined();
+      });
+    }
```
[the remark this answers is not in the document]
gre-ledger: it's ok. we just simplify a bit the code of this test because way more loaders are coming in.

libs/ledger-live-common/src/coin-modules/no-coin-eager-imports.test.ts
```diff
@@ -0,0 +1,101 @@
+import * as esbuild from "esbuild";
+import * as path from "path";
+
+const BANNED_LIBS = ["tronweb", "web3"];
```
LL782: Can we move this const close to where it's used. 
It's hard to understand out of context but in the context of the test at the bottom it is well explained

libs/ledger-live-common/src/coin-modules/registry.test.ts
```diff
@@ -26,70 +29,47 @@ const makeLoader = (family: string, overrides: Partial<CoinModuleLoader> = {}):
...
-    registerCoinModules([
-      makeLoader("evm", {
-        loadMockBridge: () => mockBridge,
-        loadMockAccount: () => mockAccount,
-      }),
-    ]);
-    expect(loadMockBridgeForFamily("evm")).toBe(mockBridge);
-    expect(loadMockAccountForFamily("evm")).toBe(mockAccount);
+for (const { loaderKey, fn, required } of allLoaders) {
+  describe(fn.name, () => {
```
LL782: **nit: prefer describe.each**

I would recommend the [describe.each](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) rather than a loop but that's mostly personal taste

LL782 (dismissed):
All looks sane!

I don't find it the most readable and self-explanatory code, which is a real concern, but it's a net improvement so LGTM

libs/ledger-live-common/src/coin-modules/loaders.test.ts
```diff
@@ -0,0 +1,15 @@
+import { coinModuleLoaders } from "./loaders";
+
+describe("coinModuleLoaders smoke test", () => {
+  for (const loader of coinModuleLoaders) {
+    const entries = Object.entries(loader).filter(([k]) => k !== "family") as [
+      string,
+      () => unknown,
+    ][];
```
ysitbon: No need to cast here, can be inferred like

```
    const entries = Object.entries(loader).flatMap(([k, fn]) =>
      k !== "family" && typeof fn === "function" ? [[k, fn] as const] : [],
    );
```

PR #16523, opened 2026-04-20
[LWDM] feat(live-common): remove direct coin-* imports from wallet-api logic (LIVE-29338)

gre-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** `no-coin-eager-imports.test.ts` now covers wallet-api code; `logic.test.ts` updated.
  - Heavy coin SDKs (`bitcoinjs-lib` via `coin-bitcoin/wallet-btc`, `ethers/viem` via `coin-evm/utils`) are no longer eagerly loaded for all wallet-api users.
  - No behavioral change — lazy loading is transparent at runtime.

### 📝 Description

`wallet-api/logic.ts` imported `getWalletAccount` from `@ledgerhq/coin-bitcoin/wallet-btc` (pulls in bitcoinjs-lib) and `wallet-api/useDappLogic.ts` imported `safeEncodeEIP55` from `@ledgerhq/coin-evm/utils` (pulls in ethers/viem) as top-level static imports. These eagerly loaded heavy coin SDKs for all wallet-api consumers regardless of which coin they use.

**Solution:** Move the require calls into the `coin-modules/` registry following the existing loader pattern.

Two new registry entries added:
- `loadGetWalletAccountForFamily` — bitcoin loader calls `require("@ledgerhq/coin-bitcoin/wallet-btc/index")` lazily
- `loadEvmUtilsForFamily` — evm loader calls `require("@ledgerhq/coin-evm/utils")` lazily

The `no-coin-eager-imports.test.ts` regression test now includes wallet-api entry points and the LIVE-29338 exception is removed.

### ❓ Context

- **JIRA**: [LIVE-29338](https://ledgerhq.atlassian.net/browse/LIVE-29338)
- Part of the coin-modules eager-import cleanup series (LIVE-29176).

[LIVE-29338]: https://ledgerhq.atlassian.net/browse/LIVE-29338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/wallet-api/logic.ts
```diff
@@ -279,7 +280,7 @@ export const bitcoinFamilyAccountGetAddressLogic = (
       return Promise.reject(new Error("Invalid derivationPath"));
     }
 
-    const walletAccount = getWalletAccount(account);
+    const walletAccount = loadGetWalletAccountForFamily("bitcoin")!.getWalletAccount(account) as BitcoinWalletAccount;
```
gre-ledger, line 283: i prefer we use `invariant` here instead of force casting with `!` and casting.

gre-ledger (commented):
NB: i'm not a super fan of introducing new loader that is coin specific, but at the same time it's better than directly importing the coin.
it seems wallet-api currently have strong dependency on coin-bitcoin and coin-evm and it's a way to decouple the parts at least, to be discussed if acceptible.

PR #16568, opened 2026-04-21
[LWDM] fix(dynamic-content): append deeplink location to content card links

tonykhaov:
### ✅ Checklist

- [ ] **Covered by automatic tests.** _No automated tests were added for this mapping-only change; validation was limited to linting the touched files._
  - Mobile dynamic content wallet, asset, category, and notification cards now append `deeplinkLocation` to deeplink URLs.
  - Desktop dynamic content action, portfolio, bottom portfolio, and notification cards now append `deeplinkLocation` to deeplink URLs.
  - Invalid or non-parseable URLs keep their original value instead of failing.

### 📝 Description

This PR fixes missing placement attribution on dynamic content deeplinks across mobile and desktop.

**Problem**: dynamic content card links were opened without preserving the card location, which meant downstream deeplink routing and tracking could lose the original placement context.

**Solution**: both platforms now rewrite dynamic content links during card mapping with `new URL(...)` and add a `deeplinkLocation` query param for the relevant placements, while safely falling back to the original link when parsing fails.

desktop: deeplinkLocation is appended to the url
image

### ❓ Context

- **JIRA or GitHub link**: [LIVE-27306](https://ledgerhq.atlassian.net/browse/LIVE-27306)

---

[LIVE-27306]: https://ledgerhq.atlassian.net/browse/LIVE-27306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/deeplinks/appendDeeplinkLocation.ts
```diff
@@ -0,0 +1,22 @@
...
+  try {
+    const url = new URL(urlValue);
+
+    if (!LEDGER_DEEPLINK_PROTOCOLS.has(url.protocol)) {
+      return urlValue;
+    }
+
+    url.searchParams.set("deeplinkLocation", location);
+
+    return url.toString();
```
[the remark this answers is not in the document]
tonykhaov: Image

LucasWerey (dismissed):
Ok for WXP

libs/ledger-live-common/src/deeplinks/appendDeeplinkLocation.ts
```diff
@@ -0,0 +1,27 @@
+const LEDGER_DEEPLINK_PROTOCOLS = new Set(["ledgerlive:", "ledgerwallet:"]);
+
+export function appendDeeplinkLocation(urlValue: string, location: string | undefined): string;
+export function appendDeeplinkLocation(
+  urlValue: undefined,
+  location: string | undefined,
+): undefined;
+export function appendDeeplinkLocation(
+  urlValue: string | undefined,
```
ysitbon: Do you feel comfortable with the first argument accepting `undefined` here? I would drop the overloads, drop `| undefined` on `urlValue`, keep `location` optional (the asymmetric case in practice is the link; the hardcoded-location mappers always have one):

```ts
export function appendDeeplinkLocation(
  urlValue: string,
  location: string | undefined,
): string {
  if (!location) return urlValue;
  try {
    const url = new URL(urlValue);
    if (!LEDGER_DEEPLINK_PROTOCOLS.has(url.protocol)) return urlValue;
    url.searchParams.set("deeplinkLocation", location);
    return url.toString();
  } catch {
    return urlValue;
  }
}
```

Call sites guard the nullish link at the point where it actually lives:

```ts
link: card.extras.link
  ? appendDeeplinkLocation(card.extras.link, ContentCardLocation.Wallet)
  : undefined,
```

This removes the overload machinery, makes the function honest about its inputs, and
keeps the nullish handling co-located with the data (mapper knows best).

tonykhaov: yeah you're right it makes sense, appendDeeplinkLocation should only expects a defined urlValue!

LL782: I have to say I much preferred it how it was before. Sorry @ysitbon 

Keeping the lib slightly more complex to keep all of the places where it's used much simpler seems worth it to me.

Not a massive deal either way. I'm happy to approve before or now

LL782 (dismissed):
Looks good to me

libs/ledger-live-common/src/deeplinks/appendDeeplinkLocation.ts
ysitbon: Yeah fair, I see the appeal of keeping callers simple 😄. One thing that nags me: the `return` is still `string | undefined` either way, so the union shows up downstream. And since the name doesn't hint at that, callers only find out when the compiler forces them to narrow it.

What I'd land on instead is two helpers with honest shapes:

```ts
  export function appendDeeplinkLocation(
    urlValue: string,
    location: string | undefined,
  ): string { /* ... */ }

  export function appendDeeplinkLocationOrNone(
    urlValue: string | undefined,
    location: string | undefined,
  ): string | undefined {
    return urlValue ? appendDeeplinkLocation(urlValue, location) : undefined;
  }
```

Each call site picks the one that matches its data: `appendDeeplinkLocation` in the hardcoded-location mappers, where the link is guaranteed, `appendDeeplinkLocationOrNone` where it isn't. Call sites stay one-liners (no ternaries repeated 10×), the lib handles its own nullish case once, and each signature tells the truth about what it returns.

Best of both worlds IMHO, keeps the ergonomic win you were asking for without leaking the union back into every mapper.

ysitbon (dismissed):

mitchellv-ledger (dismissed):

ysitbon (approved):

mitchellv-ledger (approved):

LucasWerey (approved):

PR #16646, opened 2026-04-22
[LWDM] fix(coin-modules): sui network fees

hhumphrey-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Currently we swallow any errors when preparing a sui transaction and default to 0 fees. This then allows a user to attempt a swap which then errors before they can perform the swap with a `UNKNOWN_ERROR (0x6e05)` because the parent account doesn't have enough coin to cover the fees.

Fix:
In order to prevent a user from performing the swap and give them a more user friendly message throw the error instead of adding a default. This will require a change on the swap side [WIP PR here](https://github.com/LedgerHQ/swap-live-app/pull/1569)

Before (no errors):
Screenshot 2026-04-22 at 14 53 31

After (errors)
Screenshot 2026-04-22 at 14 47 33

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-29319

---

### DIFF
--- a/libs/coin-modules/coin-sui/src/bridge/prepareTransaction.ts
+++ b/libs/coin-modules/coin-sui/src/bridge/prepareTransaction.ts
@@ -1,7 +1,6 @@
 import { findSubAccountById } from "@ledgerhq/ledger-wallet-framework/account/index";
 import { updateTransaction } from "@ledgerhq/ledger-wallet-framework/bridge/jsHelpers";
 import { AccountBridge } from "@ledgerhq/types-live";
-import BigNumber from "bignumber.js";
 import { DEFAULT_COIN_TYPE } from "../network/sdk";
 import type { SuiTransactionMode, SuiAccount, Transaction } from "../types";
 import getFeesForTransaction from "./getFeesForTransaction";
@@ -25,15 +24,10 @@ export const prepareTransaction: AccountBridge<
     amount = calculateAmount({ account, transaction });
   }
 
-  let fees: BigNumber;
-  try {
-    fees = await getFeesForTransaction({
-      account,
-      transaction,
-    });
-  } catch {
-    fees = BigNumber(0);
-  }
+  const fees = await getFeesForTransaction({
+    account,
+    transaction,
+  });
 
   let mode: SuiTransactionMode = transaction.mode ?? "send";
   let coinType = DEFAULT_COIN_TYPE;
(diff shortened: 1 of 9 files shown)


ishaba (dismissed):

hedi-edelbloute:
Hi @hhumphrey-ledger, what happens if you throw an error instead of defaulting to 0 / default value ?

hhumphrey-ledger:
@hedi-edelbloute I've updated the logic so we never return a default value if we error during `getFeesForTransaction` and then have added a specific error flow for when we know we have error'd because they have insufficient balance which allows us to surface a more actionable error on the UI.

I've tried to match the pattern found in [aptos](https://github.com/LedgerHQ/ledger-live/blob/d6e83b0b08092a83317883b3767139006e75a14b/libs/coin-modules/coin-aptos/src/bridge/getTransactionStatus.ts#L57) for surfacing errors via the transaction

libs/coin-modules/coin-sui/src/bridge/prepareTransaction.ts
```diff
@@ -25,14 +25,19 @@ export const prepareTransaction: AccountBridge<
...
   try {
     fees = await getFeesForTransaction({
       account,
       transaction,
     });
-  } catch {
-    fees = BigNumber(0);
+  } catch (error) {
+    if (error instanceof Error) {
+      errors = {
```
hedi-edelbloute: Majority of coin-modules throw here, only aptos applies this behavior, let's throw!

libs/coin-modules/coin-sui/src/logic/mapDryRunError.ts
```diff
@@ -0,0 +1,49 @@
...
+const balanceErrorMatchers: ErrorMatcher[] = [
+  {
+    // "Balance of gas object 10 is lower than the needed amount: 100"
+    pattern: /Balance of gas object \d+ is lower than the needed amount:\s*\d+/i,
+    createError: (_match, cause) => new NotEnoughBalanceFees(undefined, undefined, { cause }),
+  },
+  {
+    // "Insufficient balance of 0x2::sui::SUI for owner 0xabc...
+    //  Required: 1000, Available: 500"
+    pattern: /Insufficient balance of .+ for owner/i,
```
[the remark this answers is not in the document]
hhumphrey-ledger: we don't want to be that specific, comment is just an example message

libs/coin-modules/coin-sui/src/test/bridge.dataset.ts
```diff
@@ -87,22 +86,6 @@ const sui: CurrenciesData<Transaction> = {
             amount: BigNumber("2000000"),
           },
         },
-        {
-          name: "not enough balance",
```
hhumphrey-ledger, line 91: This is no longer relevant. The previous code path meant the fees were set to 0 and `getTransactionStatus` set this error. But now we always throw the error this won't be the case.

If this is too risky we should have a backwards compatible fallback like this one:

https://github.com/LedgerHQ/ledger-live/pull/16646/changes/8480195096e29054ac9c7c53c4efac3eaf11fffa#diff-b79ac1e098ed62605a09f68aba2c6010df684a529ca4506bf20445f5c3099386

hedi-edelbloute (approved):

liviuciulinaru (approved):

PR #16772, opened 2026-04-24
[LWM] fix: reload Recover webview when same deeplink is opened

claudiiafg:
### ✅ Checklist

  - ...

### 📝 Description

This pull request addresses an issue where the Recover webview on mobile was not reloading when the same deeplink was opened while already on the Recover screen. The solution ensures that opening a Recover deeplink with identical parameters triggers a remount of the webview by appending a unique query parameter. The implementation includes robust comparison logic for deeplink identity and comprehensive tests.

### ❓ Context

[LIVE-29768](https://ledgerhq.atlassian.net/browse/LIVE-29768)

---

[LIVE-29768]: https://ledgerhq.atlassian.net/browse/LIVE-29768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Justkant (commented):
I think that you might have a simpler solution by just using the query params of the deeplink and to pass it in the url of the webview (it should already be the case for some IIRC)
So you can just make sure in your deeplink that it's unique when sending it and we don't need any custom logic in the webview or the deeplink handler

claudiiafg:
first build: https://github.com/LedgerHQ/ledger-live-build/actions/runs/24879716872

apps/ledger-live-mobile/src/navigation/deeplinks/appendRecoverDeeplinkReloadIfNeeded.ts
Justkant, line 1: Do you still need this file ?

apps/ledger-live-mobile/src/screens/Protect/Player.tsx
```diff
@@ -35,7 +35,13 @@ const pollingPeriodMs = 1000;
...
 export function RecoverPlayer({ navigation, route }: Props) {
   const { theme } = useTheme();
-  const { platform: appId, device, fromOnboarding, ...params } = route.params || {};
+  const {
+    platform: appId,
+    device,
+    fromOnboarding,
+    date: recoverReloadDate,
+    ...params
+  } = route.params || {};
```
Justkant, line 44: This change will work, but IMO it would work the exact same if you don't use key and just let the date end up in the `inputs` below so it's appended in the url and means it will change on all new deeplink

apps/ledger-live-mobile/src/navigation/DeeplinksProvider.tsx
```diff
@@ -107,6 +107,15 @@ function getProxyURL(url: string, customBuySellUiAppId?: string) {
...
   }
 
+  // Ensure Recover deeplinks are unique so opening the same link remounts the webview.
+  if (
+    (hostname === "recover" || (hostname === "discover" && platform?.startsWith("protect"))) &&
+    platform
+  ) {
+    uri.searchParams.set("date", String(Date.now()));
+    return uri.toString();
+  }
```
Justkant, line 117: That's fine for me to keep this here for recover only but I feel like the date could be added in the deeplink sent by the live-app directly unless you don't control the original deeplink

Justkant (changes requested):
It's simpler thx for the changes, I think we can still clean things up a bit more for your issue

PR #16795, opened 2026-04-24
[LWDM] fix(live-common): replace require() with static imports in coin-modules loaders

gre-ledger:
### ✅ Checklist

  - `libs/ledger-live-common/src/coin-modules/loaders.ts` only
  - Bundle size and boot-up time will return to their state prior to the `require()` approach (i.e. identical to last release, slightly slower than current `develop`)
  - Desktop unit test execution time also returns to its pre-`require()` baseline

### 📝 Description

Dynamic `require()` calls in `coinModuleLoaders` were introduced to work around ESM/CJS module boundaries, but they introduce their own interop issues when both CJS and ESM modules are present in the dependency graph.

This replaces all `require()` calls with top-level static `import` statements so each `load*` function returns the already-resolved value synchronously. This is also a safer foundation for the next step: migrating the loaders to `async () => import(...)` (dynamic import) to achieve true lazy loading, which we want to land carefully.

**Exceptions kept as dynamic imports:**
- `casper` — `casper-js-sdk` runs `typedjson` decorators at module-evaluation time (side effects) that break live-common coverage tests when loaded eagerly
- `celo`, `kaspa` — Buffer/polyfill initialization side effects require late loading

**Before:**
```ts
loadTransaction: () => require("@ledgerhq/coin-aleo/transaction").default,
loadSetup: () => require("../families/aleo/setup"),
loadMockBridge: () => require("../families/algorand/bridge/mock").default,
```

**After:**
```ts
import aleoTransaction from "@ledgerhq/coin-aleo/transaction";
import * as aleoSetup from "../families/aleo/setup";
import algorandMockBridge from "../families/algorand/bridge/mock";
// ...
loadTransaction: () => aleoTransaction,
loadSetup: () => aleoSetup,
loadMockBridge: () => algorandMockBridge,
```

No public API change — `CoinModuleLoader` signatures are identical.

### 🔧 Integration test fixes

The static imports changed module load order, breaking 3 LLD integration tests (canton AddAccountDrawer, canton OnboardModal, concordium OnboardModal). Root cause: `jest.mock()` creates a fresh module object and does not patch already-loaded references, while `jest.spyOn` mutates the shared cached module object (live binding). Since `setupFilesAfterEnv` now loads all coin modules eagerly before tests run, `jest.mock()` was too late.

Fixes applied:
- **canton & concordium setup files** — changed named import to namespace import (`import * as liveSignerX from "..."`) so `jest.spyOn` can intercept the constructor at call time
- **`live-signer-concordium/src/index.ts`** — changed `export * from` (wildcard) to named `export { DmkSignerConcordium }`: the wildcard compiles to `Object.defineProperty` without a pre-declaration, producing `configurable: false` which blocks `jest.spyOn`; named export pre-declares `exports.DmkSignerConcordium = void 0` keeping `configurable: true`
- **canton & concordium integration tests** — switched from `jest.mock()` to `jest.spyOn(...).mockImplementation()` in `beforeAll`; `@ledgerhq/live-signer-canton` and `@ledgerhq/live-signer-concordium` added as devDependencies to `ledger-live-desktop`
- **`loaders.ts` — restored `.default`** on all 15 `loadMockBridge` `require()` calls (accidentally dropped): without it, Detox/mock-mode tests crash because the mock bridge is unwrapped as `{ __esModule: true, default: {...} }` instead of `{ currencyBridge, accountBridge }`

### ❓ Context

- **ADR link (if any)**: N/A

### DIFF
--- a/.changeset/clean-modules-import.md
+++ b/.changeset/clean-modules-import.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Replace dynamic require() with static imports in coin-modules loaders
--- a/libs/ledger-live-common/src/coin-modules/loaders.ts
+++ b/libs/ledger-live-common/src/coin-modules/loaders.ts
@@ -1,242 +1,358 @@
-/* eslint-disable @typescript-eslint/no-require-imports */
-import type { CoinModuleLoader } from "./types";
+import type { CoinModuleLoader, FamilySetup } from "./types";
 
-// Hand-maintained: one entry per coin family. See FUTURE.md for the async evolution plan.
+import * as aleoSetup from "../families/aleo/setup";
+import aleoTransaction from "@ledgerhq/coin-aleo/transaction";
+
+import * as algorandSetup from "../families/algorand/setup";
+import algorandTransaction from "@ledgerhq/coin-algorand/transaction";
+import algorandMockAccount from "@ledgerhq/coin-algorand/mock";
+
+import * as aptosSetup from "../families/aptos/setup";
+import aptosTransaction from "@ledgerhq/coin-aptos/transaction";
+
+import * as bitcoinSetup from "../families/bitcoin/setup";
+import bitcoinTransaction from "@ledgerhq/coin-bitcoin/transaction";
+import * as bitcoinWalletApiAdapter from "../families/bitcoin/walletApiAdapter";
+import * as bitcoinPlatformAdapter from "../families/bitcoin/platformAdapter";
+import { clearAccount as bitcoinClearAccount } from "../families/bitcoin/clearAccount";
+import {
+  isEditableOperation as bitcoinIsEditableOperation,
+  isStuckOperation as bitcoinIsStuckOperation,
+  getStuckAccountAndOperation as bitcoinGetStuckAccountAndOperation,
+} from "../families/bitcoin/operations";
+
+import * as cantonSetup from "../families/canton/setup";
+import cantonTransaction from "@ledgerhq/coin-canton/transaction";
+import { isAccountEmpty as cantonIsAccountEmpty } from "@ledgerhq/coin-canton";
+
+import * as cardanoSetup from "../families/cardano/setup";
+import cardanoTransaction from "@ledgerhq/coin-cardano/transaction";
+
+import casperTransaction from "@ledgerhq/coin-casper/transaction";
+
+import celoTransaction from "@ledgerhq/coin-celo/transaction";
+
+import * as concordiumSetup from "../families/concordium/setup";
+import concordiumTransaction from "@ledgerhq/coin-concordium/transaction";
+
+import * as cosmosSetup from "../families/cosmos/setup";
+import cosmosTransaction from "@ledgerhq/coin-cosmos/transaction";
+import * as cosmosWalletApiAdapter from "../families/cosmos/walletApiAdapter";
+import cosmosMockAccount from "@ledgerhq/coin-cosmos/mock";
+import { isAccountEmpty as cosmosIsAccountEmpty } from "@ledgerhq/coin-cosmos/helpers";
+import { getVotesCount as cosmosGetVotesCount } from "../families/cosmos/getVotesCount";
+
+import * as evmSetup from "../families/evm/setup";
+import evmTransaction from "@ledgerhq/coin-evm/transaction";
+import * as evmWalletApiAdapter from "../families/evm/walletApiAdapter";
+import * as evmPlatformAdapter from "../families/evm/platformAdapter";
+import { validateAddress as evmValidateAddress } from "@ledgerhq/coin-evm/logic/validateAddress";
+import evmAlpacaSigner from "../bridge/generic-alpaca/families/evm/signer";
+import {
+  isEditableOperation as evmIsEditableOperation,
+  isStuckOperation as evmIsStuckOperation,
+  getStuckAccountAndOperation as evmGetStuckAccountAndOperation,
+} from "../families/evm/operations";
+
+import * as filecoinSetup from "../families/filecoin/setup";
+import filecoinTransaction from "@ledgerhq/coin-filecoin/transaction";
+
+import * as hederaSetup from "../families/hedera/setup";
+import hederaTransaction from "@ledgerhq/coin-hedera/transaction";
+
+import * as iconSetup from "../families/icon/setup";
+import iconTransaction from "@ledgerhq/coin-icon/transaction";
+
+import * as icSetup from "../families/internet_computer/setup";
+import icTransaction from "@ledgerhq/coin-internet_computer/transaction";
+
+import * as minaSetup from "../families/mina/setup";
+import minaTransaction from "@ledgerhq/coin-mina/transaction";
+
+import * as multiversxSetup from "../families/multiversx/setup";
+import multiversxTransaction from "@ledgerhq/coin-multiversx/transaction";
+
+import * as nearSetup from "../families/near/setup";
+import nearTransaction from "@ledgerhq/coin-near/transaction";
+
+import * as polkadotSetup from "../families/polkadot/setup";
+import polkadotTransaction from "@ledgerhq/coin-polkadot/transaction";
+import * as polkadotWalletApiAdapter from "../families/polkadot/walletApiAdapter";
+import * as polkadotPlatformAdapter from "../families/polkadot/platformAdapter";
+
+import solanaAlpacaSigner from "../bridge/generic-alpaca/families/solana/signer";
+import * as solanaSetup from "../families/solana/setup";
+import solanaTransaction from "@ledgerhq/coin-solana/transaction";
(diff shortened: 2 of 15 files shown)


libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
```diff
@@ -1,27 +1,12 @@
...
-    }
-    case "solana": {
-      return require("./families/solana/signer").default;
-    }
-  }
-  throw new Error(`signer for ${network} not implemented`);
+  const family = networkToFamily[network] ?? network;
+  const signer = loadSignerForFamily(family);
+  if (!signer) throw new Error(`No signer registered for network ${network}`);
+  return signer;
```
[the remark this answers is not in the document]
gre-ledger: @Justkant do you know about this "ripple" and "xrp" co-existing in the network parameter here? could there be a bug on this? 🤔

.changeset/clean-modules-import.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Replace dynamic require() with static imports in coin-modules loaders
```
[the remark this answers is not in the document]
gre-ledger: well, in a way, for signer.ts, it's also about require() moving to import, so i think we can stick with this minimal changeset sentence..

libs/ledger-live-common/src/coin-modules/loaders.ts
```diff
@@ -1,237 +1,420 @@
...
+    loadSetup: () => xrpSetup,
+    loadTransaction: () => xrpTransaction,
+    loadDeviceTxConfig: () => xrpDeviceTxConfig,
+    loadWalletApiAdapter: () => xrpWalletApiAdapter,
+    loadPlatformAdapter: () => xrpPlatformAdapter,
+    loadMockBridge: () => xrpMockBridge,
+    loadValidateAddress: () => xrpValidateAddress,
+    loadSigner: () => xrpAlpacaSigner,
+  },
+] as unknown as CoinModuleLoader[];
```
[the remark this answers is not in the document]
gre-ledger: this is necessary for the moment.

libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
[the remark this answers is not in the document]
gre-ledger: ok it's normal.

libs/ledger-live-common/src/coin-modules/registry.ts
```diff
@@ -101,6 +102,9 @@ export const loadClearAccountForFamily = (
 export const loadValidateAddressForFamily = (family: string): ValidateAddressFn | undefined =>
   loaders.get(family)?.loadValidateAddress?.();
 
+export const loadSignerForFamily = (family: string): AlpacaSigner | undefined =>
+  loaders.get(family)?.loadSigner?.();
```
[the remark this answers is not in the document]
gre-ledger: added

Justkant (dismissed):

apps/ledger-live-desktop/jest.polyfills.js
```diff
@@ -94,3 +94,13 @@ Object.defineProperties(globalThis, {
...
+
+// Must run before setupFilesAfterEnv (jestSetup.js) so coin setup modules that call
+// newKit/ecc library at import time (celo, kaspa…) get a proper Buffer from the start.
+// 'buffer' bare import gets translated to 'node:buffer' by the bundler, so use relative path.
+if (!globalThis.Buffer) {
+  const { Buffer } = require("./node_modules/buffer");
+  Object.defineProperty(globalThis, "Buffer", { value: Buffer });
+} else if (!(globalThis.Buffer.prototype instanceof Uint8Array)) {
+  Object.setPrototypeOf(globalThis.Buffer.prototype, Uint8Array.prototype);
+}
```
LL782: Can we go for this

```suggestion
const { Buffer } = require("buffer");
globalThis.Buffer ??= Buffer;
```

Not sure what case we're trying to cover with the check against `instanceof Uint8Array` Maybe include it with a console.warn to make it observable. Is it really needed? 

```suggestion
const { Buffer } = require("buffer");
globalThis.Buffer ??= Buffer;

if (!(globalThis.Buffer.prototype instanceof Uint8Array)) {
  console.warn("Patching Buffer prototype");
  Object.setPrototypeOf(globalThis.Buffer.prototype, Uint8Array.prototype);
}
```

libs/ledger-live-common/src/coin-modules/loaders.ts
```diff
@@ -1,237 +1,405 @@
...
+
+import * as xrpSetup from "../families/xrp/setup";
+import xrpTransaction from "../families/xrp/transaction";
+import xrpDeviceTxConfig from "../families/xrp/deviceTransactionConfig";
+import xrpWalletApiAdapter from "../families/xrp/walletApiAdapter";
+import xrpPlatformAdapter from "../families/xrp/platformAdapter";
+import { validateAddress as xrpValidateAddress } from "@ledgerhq/coin-xrp/logic/validateAddress";
+
+// Mock bridges/accounts are lazy: their mockHelpers → bridge/impl chain would pre-load generic-alpaca, breaking jest.mock() isolation.
+/* eslint-disable @typescript-eslint/no-require-imports */
```
LL782:
```suggestion
```

Is this needed?

libs/ledger-live-common/src/coin-modules/loaders.ts
```diff
@@ -1,237 +1,405 @@
...
+
 // Hand-maintained: one entry per coin family. See FUTURE.md for the async evolution plan.
-export const coinModuleLoaders: CoinModuleLoader[] = [
+// Cast needed: coin-specific types (AccountBridge<AleoTransaction> etc.) are structurally
+// narrower than the generic boundary types in CoinModuleLoader. This is the same type-erasure
+// that require() provided implicitly; type safety is preserved at the consuming end (registry).
```
LL782: Not sure what value these comments are providing


```suggestion
```

libs/ledger-live-common/src/coin-modules/loaders.ts
```diff
@@ -1,237 +1,405 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
LL782: I don't actually see the explicit any


```suggestion
```

libs/ledger-live-common/src/coin-modules/loaders.ts
LL782: Ah I see it now

apps/ledger-live-desktop/jest.polyfills.js
gre-ledger: ideally i would prefer to not even need to polyfill this. i'm trying to understand why the import ordering have created this new problem 😢

libs/ledger-live-common/src/coin-modules/loaders.ts
LL782: I think the comments are getting a bit out of hand and hard to keep inline with the latest code

How about this instead?


```suggestion
/* eslint-disable @typescript-eslint/no-require-imports */
// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```

libs/ledger-live-common/src/coin-modules/loaders.ts
LL782, line 1: The type-assertion in this file worries me but I understand it's a temporary state. If the problem comes up again when we're working on a longterm solution I'd be happy to share some ideas

libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
LL782, line 1: **suggestion: can we try to make the code more self-explanitory?**

```ts
const networkToFamily = (network: string) => {
  if (network === "xrp") return "ripple";
  return network;
};

export function getSigner(network: string): AlpacaSigner {
  const family = networkToFamily(network);
  const signer = loadSignerFor(family);
  if (!signer) throw new Error(`No signer registered for family ${family}`);
  return signer;
}
```

For me the comment increases cognitive load rather than reducing it. How's this instead? Really just making `networkToFamily` as simple to read as possible without any extra noise.

gre-ledger: @LL782  👍 good point. can you review https://github.com/LedgerHQ/ledger-live/pull/16835 ? i did update there, i want to ship this bit first.

libs/ledger-live-common/src/coin-modules/loaders.ts
gre-ledger: i've managed to fix it. with some `any` in the types of the modules, it's still better than full blind. i'll continue on this PR as Draft until i can manage to pass CI.

libs/ledger-live-common/src/coin-modules/loaders.ts
gre-ledger: fixed

PR #16832, opened 2026-04-27
[LWM] feat(live-dmk-mobile): add DMK HTTP proxy transport for dev mode

lysyi3m:
### ✅ Checklist

    - Mobile dev mode `httpdebug` transport (iOS simulator + `ledger-live proxy` CLI workflow)
    - `@ledgerhq/live-dmk-mobile`: new `DeviceManagementKitHTTPProxyTransport` export
    - `ledger-live proxy` CLI: HTTP exchange log line now prints the APDU hex

### 📝 Description

DMK-only coin families were broken on mobile in dev mode when forwarding APDUs through the `ledger-live proxy` CLI — the connection was rejected because the existing `httpdebug` transport was built on `withStaticURLs` from `@ledgerhq/hw-transport-http`, a legacy `Transport` with no `dmk`/`sessionId` properties, so `isDmkTransport()` returned false.

This PR wraps the proxy CLI's HTTP protocol into a proper DMK transport. The shape mirrors the speculos integration:

```
DeviceManagementKitHTTPProxyTransport ← outer hw-transport wrapper (exposes dmk + sessionId)
          │ shares the singleton DMK + caches the session for the active URL
          ▼
HttpProxyDmkTransport                 ← DMK Transport plugin
          │ implements DMK's Transport interface
          ▼
fetch() → ledger-live proxy           ← POST / with { apduHex }
```

Key behaviors:
  - URL normalization (`ws://` → `http://`, `wss://` → `https://`) since the proxy CLI emits `ws://` URLs by default while sharing the HTTP endpoint on the same port.
  - Session reuse: the most recently established `(url, sessionId)` pair is cached and returned by subsequent `open()` calls for the same URL. Concurrent `open()` calls are de-duplicated via a URL-keyed in-flight promise so the shared `httpProxyUrlSubject` and DMK transport state aren't clobbered mid-connect. Switching to a different URL reconnects (only one URL can be active at a time because the underlying `BehaviorSubject` holds a single value).
  - `sendApdu` returns `Left(UnknownDeviceError)` on fetch failure but **does not** call `onDisconnect` — that callback signals physical disconnection (USB unplugged, BLE dropped) and would wipe the DMK session registry. A failed HTTP request is not a physical disconnect.
  - Session refresher disabled (`isRefresherDisabled: true`) — the proxy is a dev-only tool with no need for periodic keepalives.
  - Disconnect detection via `getDeviceSessionState`: on `NOT_CONNECTED`, the cached `sessionId` is cleared so the next `open()` call establishes a fresh session.

The legacy `withStaticURLs` is retained inside the `__DEV__ && Config.DEVICE_PROXY_URL` block — it still drives URL-polling discovery; only the `open` path was swapped.

### ❓ Context

- **JIRA or GitHub link**: 
- **ADR link (if any)**: 

---

### DIFF
--- a/apps/ledger-live-mobile/src/services/registerTransports.ts
+++ b/apps/ledger-live-mobile/src/services/registerTransports.ts
@@ -7,7 +7,10 @@ import { getDeviceModel } from "@ledgerhq/devices";
 import { DescriptorEvent } from "@ledgerhq/hw-transport";
 import { DeviceModelId } from "@ledgerhq/types-devices";
 import getBLETransport from "~/transport/bleTransport";
-import { DeviceManagementKitHIDTransport } from "@ledgerhq/live-dmk-mobile";
+import {
+  DeviceManagementKitHIDTransport,
+  DeviceManagementKitHTTPProxyTransport,
+} from "@ledgerhq/live-dmk-mobile";
 import { DeviceManagementKitTransportSpeculos } from "@ledgerhq/live-dmk-speculos";
 import { retry } from "@ledgerhq/live-common/promise";
 
@@ -46,37 +49,40 @@ export const registerTransports = () => {
     });
   }
 
-  // Add dev mode support of an http proxy
-  let DebugHttpProxy: ReturnType<typeof withStaticURLs>;
-  const httpdebug: TransportModule = {
-    id: "httpdebug",
-    open: id => (id.startsWith("httpdebug|") ? DebugHttpProxy.open(id.slice(10)) : null),
-    disconnect: id =>
-      id.startsWith("httpdebug|")
-        ? Promise.resolve() // nothing to do
-        : null,
-  };
-
-  if (__DEV__ && Config.DEVICE_PROXY_URL) {
-    DebugHttpProxy = withStaticURLs(Config.DEVICE_PROXY_URL.split("|"));
-    httpdebug.discovery = new Observable<DescriptorEvent<string>>(o =>
-      DebugHttpProxy.listen(o),
-    ).pipe(
-      map(({ type, descriptor }) => ({
-        type,
-        id: `httpdebug|${descriptor}`,
-        deviceModel: getDeviceModel(
-          (Config?.FALLBACK_DEVICE_MODEL_ID as DeviceModelId) || DeviceModelId.nanoX,
-        ),
-        wired: Config?.FALLBACK_DEVICE_WIRED === "YES",
-        name: descriptor,
-      })),
-    );
-  } else {
-    DebugHttpProxy = withStaticURLs([]);
+  // Dev-only HTTP proxy transport. Gated on __DEV__ so production builds don't expose
+  // an `open` path that would accept arbitrary URLs. The module is registered for the
+  // whole __DEV__ build so the runtime "Settings > Debug > Connectivity > HTTP transport"
+  // flow can supply a URL at runtime; build-time URL polling is opt-in via DEVICE_PROXY_URL.
+  if (__DEV__) {
+    const httpdebug: TransportModule = {
+      id: "httpdebug",
+      open: id =>
+        id.startsWith("httpdebug|")
+          ? DeviceManagementKitHTTPProxyTransport.open(id.slice("httpdebug|".length))
+          : null,
+      disconnect: id =>
+        id.startsWith("httpdebug|")
+          ? Promise.resolve() // nothing to do
+          : null,
+    };
+    if (Config.DEVICE_PROXY_URL) {
+      const DebugHttpProxy = withStaticURLs(Config.DEVICE_PROXY_URL.split("|"));
+      httpdebug.discovery = new Observable<DescriptorEvent<string>>(o =>
+        DebugHttpProxy.listen(o),
+      ).pipe(
+        map(({ type, descriptor }) => ({
+          type,
+          id: `httpdebug|${descriptor}`,
+          deviceModel: getDeviceModel(
+            (Config?.FALLBACK_DEVICE_MODEL_ID as DeviceModelId) || DeviceModelId.nanoX,
+          ),
+          wired: Config?.FALLBACK_DEVICE_WIRED === "YES",
+          name: descriptor,
+        })),
+      );
+    }
+    registerTransportModule(httpdebug);
   }
-
-  registerTransportModule(httpdebug);
 
   // BLE is always the fallback choice because we always keep raw id in it
   registerTransportModule({
--- a/libs/live-dmk-mobile/src/transport/DeviceManagementKitHTTPProxyTransport.test.ts
+++ b/libs/live-dmk-mobile/src/transport/DeviceManagementKitHTTPProxyTransport.test.ts
@@ -1,0 +1,433 @@
+import { BehaviorSubject, Observable, Subject } from "rxjs";
+import {
+  DeviceManagementKit,
+  DeviceModelId as DMKDeviceModelId,
+  DeviceSessionState,
+  DeviceSessionStateType,
+  DeviceStatus,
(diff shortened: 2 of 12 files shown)


libs/live-dmk-mobile/src/transport/DeviceManagementKitHTTPProxyTransport.test.ts
```diff
@@ -0,0 +1,322 @@
...
+  } as unknown as DeviceManagementKit;
+}
+
+describe("DeviceManagementKitHTTPProxyTransport", () => {
+  let mockDmk: DeviceManagementKit;
+
+  beforeEach(() => {
+    jest.restoreAllMocks();
+    (DeviceManagementKitHTTPProxyTransport as unknown as { byUrl: Map<string, unknown> }).byUrl.clear();
+
```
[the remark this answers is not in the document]
lysyi3m: we're mirroring  DeviceManagementKitHIDTransport.test.ts pattern in the same package

hedi-edelbloute (dismissed):
lgtm, what do you think @mbertin-ledger @KVNLS ? were you aware ?

LucasWerey (dismissed):

libs/live-dmk-mobile/src/transport/HttpProxyDmkTransport.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  isSupported(): boolean {
+    return true;
+  }
+
+  listenToAvailableDevices(): Observable<TransportDiscoveredDevice[]> {
+    const device = this.syntheticDevice();
+    return new Observable(subscriber => {
+      subscriber.next([device]);
+    });
```
OlivierFreyssinet: simpler:

```suggestion
import { from } from "rxjs"

// ...
    return from([device])
```

libs/live-dmk-mobile/src/transport/HttpProxyDmkTransport.ts
```diff
@@ -0,0 +1,142 @@
...
+    deviceId,
+  }: {
+    deviceId: string;
+    onDisconnect: (deviceId: string) => void;
+  }): Promise<Either<ConnectError, TransportConnectedDevice>> {
+    const url = this.url;
+
+    const sendApdu = async (apdu: Uint8Array): Promise<Either<DmkError, ApduResponse>> => {
+      try {
+        const apduHex = Buffer.from(apdu).toString("hex");
```
OlivierFreyssinet: We prefer not to use Buffer in the DMK as it's not a native JS object, in LW it's just polyfilled.
Prefer using Uint8Array directly, or use those helpers from the DMK:

```suggestion
import {
  bufferToHexaString,
} from "@ledgerhq/device-management-kit";

const apduHex = bufferToHexaString(apdu, false);
```

libs/live-dmk-mobile/src/transport/HttpProxyDmkTransport.ts
```diff
@@ -0,0 +1,142 @@
...
+              ? body.error
+              : body.error instanceof Error
+                ? body.error.message
+                : JSON.stringify(body.error);
+          return Left(new UnknownDeviceError(message));
+        }
+        if (!body.data) {
+          return Left(new UnknownDeviceError("empty response from proxy"));
+        }
+        const bytes = Uint8Array.from(Buffer.from(body.data, "hex"));
```
OlivierFreyssinet:
```suggestion
import {
  hexaStringToBuffer,
} from "@ledgerhq/device-management-kit";

const bytes = hexaStringToBuffer(body.data);
if (!bytes) {
  return Left(new UnknownDeviceError(`Invalid hex data`));
}
```

OlivierFreyssinet (commented):
Question: how much testing did you do? Is it as stable as the previous proxy transport?

I have a work in progress for a full rework of the proxy logic (https://github.com/LedgerHQ/ledger-live/pull/14865) but didn't merge yet as I didn't have time to test it enough.
Your work should be enough for full parity with the current transport, adding DMK support and unlocking some devs in the meantime, provided that it's as stable as what was already in place.

lysyi3m:
@OlivierFreyssinet on the testing: 

honestly, I've tested the following:

* Concordium (DMK-only) - app install, create account, send/receive
* Solana - app install, create account, send/receive
* Celo - (only LegacySignerCelo is implemented) - app install, create account, send/receive
* Ethereum - app install, create account, send/receive

I understand that this might be not a fullest QA, but it is definitely unblocks local development for DMK-only signers, and from what I see so far - doesn't break existing functionality

libs/live-dmk-mobile/src/transport/DeviceManagementKitHTTPProxyTransport.ts
```diff
@@ -0,0 +1,143 @@
...
+    return raw.replace(/^ws(s?):\/\//, "http$1://");
+  }
+
+  private static ensureEntry(url: string): DmkEntry {
+    let entry = this.byUrl.get(url);
+    if (!entry) {
+      const dmk = new DeviceManagementKitBuilder()
+        .addTransport(httpProxyTransportFactory(url))
+        .build();
+      entry = { dmk };
```
OlivierFreyssinet: We should no create DMK instances on the fly, there should only be one instance in Ledger Wallet. It will cause issues down the line for future integrations that are not going through ledgerjs `Transport` register modules.
The constraint here is that we have to instantiate the transport with a fixed URL. What if we could change the url even after DMK instantiation?    
I think that an easy solution to do this, a bit quick and dirty, but better I think (and ok as it's only for dev workflows), is to pass a subject to the transport factory, instead of a static url. You can emit a new url in the subject here when it needs to change, and the HttpProxyDmkTransport receives this subject and uses it current value.
I think it would work.

lysyi3m: @OlivierFreyssinet addressed

libs/live-dmk-mobile/src/transport/HttpProxyDmkTransport.test.ts
```diff
@@ -0,0 +1,326 @@
...
+describe("HttpProxyDmkTransport", () => {
+  const url = "http://localhost:8435";
+  let originalFetch: typeof global.fetch;
+  let urlSubject: BehaviorSubject<string | null>;
+
+  beforeEach(() => {
+    originalFetch = global.fetch;
+    jest.restoreAllMocks();
+    global.fetch = jest.fn();
+    urlSubject = new BehaviorSubject<string | null>(url);
```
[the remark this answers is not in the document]
lysyi3m: kept because the rest of the package (DeviceManagementKitHIDTransport.test.ts) uses the same pattern

apps/ledger-live-mobile/src/services/registerTransports.ts
```diff
@@ -47,18 +50,18 @@ export const registerTransports = () => {
   }
 
   // Add dev mode support of an http proxy
-  let DebugHttpProxy: ReturnType<typeof withStaticURLs>;
   const httpdebug: TransportModule = {
     id: "httpdebug",
-    open: id => (id.startsWith("httpdebug|") ? DebugHttpProxy.open(id.slice(10)) : null),
+    open: id =>
+      id.startsWith("httpdebug|") ? DeviceManagementKitHTTPProxyTransport.open(id.slice(10)) : null,
```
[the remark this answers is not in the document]
lysyi3m: > "httpdebug|" is 9 chars

no. `"httpdebug|".length === 10`

OlivierFreyssinet (approved):

Sebastien-Dav1d (approved):

LucasWerey (approved):

PR #16835, opened 2026-04-27
[LWDM] feat(live-common): add loadSigner to CoinModuleLoader registry

gre-ledger:
### ✅ Checklist

  - `libs/ledger-live-common/src/coin-modules/` and `bridge/generic-alpaca/signer.ts` only
  - No runtime behaviour change — `getSigner` resolves identically, just via the registry

### 📝 Description

Subset of #16795: extracts only the `loadSigner` / registry wiring, leaving the `require()→static import` rework for a separate PR.

**What changed:**
- `CoinModuleLoader` gains `loadSigner?: () => AlpacaSigner`
- `loadSignerForFamily` exported from `coin-modules/registry`
- `loaders.ts`: `loadSigner` entries added for `evm`, `solana`, `stellar`, `tezos`, `xrp`
- `generic-alpaca/signer.ts`: `getSigner` now delegates to the registry instead of a hardcoded switch/require block
- `wallet-cli`: `loadSigner` entries for `evm` and `solana`

**Consumer side (after merge):**
```ts
// instead of maintaining the switch in signer.ts, new families just add:
loadSigner: () => require("../bridge/generic-alpaca/families/myfamily/signer").default,
// in their loaders.ts entry
```

### ❓ Context

- **JIRA or GitHub link**: subset of #16795
- **ADR link (if any)**: N/A

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
+++ b/libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
@@ -1,27 +1,14 @@
 import type { AlpacaSigner } from "./types";
+import { loadSignerForFamily } from "../../coin-modules/registry";
 
-/**
- * Lazy-load Alpaca signer modules so wallet-cli (EVM-only flows) does not bundle XRP/Stellar/Tezos HW stacks.
- */
-// TODO: we could also use dynamic import here but we need to update everything to async/await where getSigner is used
+const networkToFamily = (network: string) => {
+  if (network === "ripple") return "xrp";
+  return network;
+};
+
 export function getSigner(network: string): AlpacaSigner {
-  switch (network) {
-    case "ripple":
-    case "xrp": {
-      return require("./families/xrp/signer").default;
-    }
-    case "stellar": {
-      return require("./families/stellar/signer").default;
-    }
-    case "tezos": {
-      return require("./families/tezos/signer").default;
-    }
-    case "evm": {
-      return require("./families/evm/signer").default;
-    }
-    case "solana": {
-      return require("./families/solana/signer").default;
-    }
-  }
-  throw new Error(`signer for ${network} not implemented`);
+  const family = networkToFamily(network);
+  const signer = loadSignerForFamily(family);
+  if (!signer) throw new Error(`No signer registered for network ${network}`);
+  return signer;
 }
--- a/libs/ledger-live-common/src/coin-modules/registry.ts
+++ b/libs/ledger-live-common/src/coin-modules/registry.ts
@@ -1,6 +1,7 @@
 import { CurrencyNotSupported } from "@ledgerhq/errors";
 import type {
   AccountModule,
+  AlpacaSigner,
   CoinModuleLoader,
   DeviceTransactionConfigFn,
   FamilySetup,
@@ -100,6 +101,9 @@ export const loadClearAccountForFamily = (
 
 export const loadValidateAddressForFamily = (family: string): ValidateAddressFn | undefined =>
   loaders.get(family)?.loadValidateAddress?.();
+
+export const loadSignerForFamily = (family: string): AlpacaSigner | undefined =>
+  loaders.get(family)?.loadSigner?.();
 
 export const loadIsEditableOperationForFamily = (
   family: string,
--- a/libs/ledger-live-common/src/coin-modules/types.ts
+++ b/libs/ledger-live-common/src/coin-modules/types.ts
@@ -15,6 +15,8 @@ import type { Transaction as WalletAPITransaction } from "@ledgerhq/wallet-api-c
 import type { Resolver } from "../hw/getAddress/types";
 import type { SignMessage } from "../hw/signMessage/types";
 import type { GetWalletAPITransactionSignFlowInfos } from "../wallet-api/types";
+import type { AlpacaSigner } from "../bridge/generic-alpaca/types";
+export type { AlpacaSigner };
 
 export type MessageSignerModule = {
   signMessage: SignMessage;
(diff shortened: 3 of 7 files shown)


libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
```diff
@@ -1,27 +1,12 @@
...
-    }
-    case "solana": {
-      return require("./families/solana/signer").default;
-    }
-  }
-  throw new Error(`signer for ${network} not implemented`);
+  const family = networkToFamily[network] ?? network;
+  const signer = loadSignerForFamily(family);
+  if (!signer) throw new Error(`No signer registered for network ${network}`);
+  return signer;
```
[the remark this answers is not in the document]
gre-ledger: accepted tradeoff to not have such test.

libs/ledger-live-common/src/bridge/generic-alpaca/signer.ts
```diff
@@ -1,27 +1,12 @@
 import type { AlpacaSigner } from "./types";
+import { loadSignerForFamily } from "../../coin-modules/registry";
+
+// Alpaca network names differ from coin family names in a few cases.
+const networkToFamily: Record<string, string> = { ripple: "xrp" };
```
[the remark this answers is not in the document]
gre-ledger: ok to keep it inline for now.

Justkant (approved):

gre-ledger:
`@smoke` e2e tests on
desktop https://github.com/LedgerHQ/ledger-live/actions/runs/24988147778
mobile https://github.com/LedgerHQ/ledger-live/actions/runs/24988152444

libs/ledger-live-common/src/coin-modules/registry.ts
```diff
@@ -101,6 +102,9 @@ export const loadClearAccountForFamily = (
 export const loadValidateAddressForFamily = (family: string): ValidateAddressFn | undefined =>
   loaders.get(family)?.loadValidateAddress?.();
 
+export const loadSignerForFamily = (family: string): AlpacaSigner | undefined =>
+  loaders.get(family)?.loadSigner?.();
+
```
LL782, line 107: **nit: prefer a short name that doesn't repeat the arguement**


```suggestion
export const loadSignerFor = (family: string): AlpacaSigner | undefined =>
  loaders.get(family)?.loadSigner?.();

```

Then you can have implementation like

```ts
const signer = loadSignerFor(family);
```

gre-ledger: they are all named `*ForFamily` in this file. so i would prefer consistency for now.

libs/ledger-live-common/src/coin-modules/types.ts
```diff
@@ -117,4 +119,5 @@ export type CoinModuleLoader = {
   loadGetVotesCount?: () => (account: Account) => number;
   loadClearAccount?: () => (account: Account) => void;
   loadValidateAddress?: () => ValidateAddressFn;
+  loadSigner?: () => AlpacaSigner;
```
LL782, line 122: **query: should we include the word Alpaca throughout?**

As I've reviewed the code changes, the inconsistent use of "signer" and "alpacaSigner" has really drawn my attention. This is a bit distracting.

How about we avoid the generic "signer" everywhere?

```ts
loadAlpacaSigner?:: () => AlpacaSigner`
```

libs/ledger-live-common/src/coin-modules/registry.ts
LL782: In relation to the other comment about consistent use of "Alpaca", this shorter name lets us include the missing "alpaca" detail:

```ts
export const alpacaSignerFor = (family: string): AlpacaSigner | undefined => 
  loaders.get(family)?.loadAlpacaSigner?.();
```

WDYT?

gre-ledger: I asked Claude and i'm a bit balanced, Alpaca is a bit a technical implementation detail here. I think with alpaca initiative they want to make this the generic signer.

> ⏺ AlpacaSigner — leaks an internal codename into a public type. The shape (getAddress, signMessage, context) is fully generic. Better:          
>  BridgeSigner. But renaming it touches all of generic-alpaca/ — out of scope for this PR.                                                      
                                                                                                                                                
>  loadSigner in CoinModuleLoader — ambiguous, because setup.ts already has a different kind of signer (CreateSigner, the HW transport one). A   
>  reader of CoinModuleLoader can't tell which one this is. loadBridgeSigner is unambiguous. Want to rename it now?

LL782 (approved):
My comments are to try to improve readability but not blocking for this PR. I'd love to know your thoughts and if we agree see the changes in a future PR. Looking forward to hearing back

As for these changes, I'm happy to approve 👍

libs/ledger-live-common/src/coin-modules/registry.ts
gre-ledger: @hedi-edelbloute what's your opinion on this?

Justkant: In that case I feel like you couldn't have another `loadSignerFor(SomethingElse)` because you probably don't want generic params on a single function `loadSignerFor`

If we always load by family we could just have a `loadSigner` function and we can omit the `ForFamily` as it always the case

libs/ledger-live-common/src/coin-modules/types.ts
Justkant: Alpaca is going to be removed as it's confusing for a lot of people so IMO we should keep it as is

libs/ledger-live-common/src/coin-modules/registry.ts
hedi-edelbloute: It's what you said @gre-ledger we want this to be the "generic signer" that ultimately will be an interface for all signers, we should consider having an abstracted interface at the signer level but for now this is this glue that does it

I agree with the `loadSigner` option, today it's family as input but it can evolve, as we could have one family having different signers

PR #16874, opened 2026-04-28
[LWDM] test(live-common): silence and assert console.error in wallet-api helpers tests

gre-ledger:
### ✅ Checklist

  - Test-only change, no production code modified.

### 📝 Description

`isWhitelistedDomain` intentionally logs `console.error` when rejecting URLs (wrong scheme, wrong domain, invalid format). The existing tests were hitting these paths but not asserting the logs, causing noisy output on every test run.

This PR mocks `console.error` in `beforeEach`/`afterEach` to suppress the noise, and adds explicit assertions:
- rejection tests: `expect(consoleSpy).toHaveBeenCalledWith(<exact message>)`
- allow tests: `expect(consoleSpy).not.toHaveBeenCalled()`

No source code changed.

### ❓ Context

### DIFF
--- a/libs/ledger-live-common/src/wallet-api/helpers.test.ts
+++ b/libs/ledger-live-common/src/wallet-api/helpers.test.ts
@@ -1,11 +1,27 @@
 import { getInitialURL } from "./helpers";
 
 describe("wallet-api helpers", () => {
+  let consoleSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleSpy = jest.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    consoleSpy.mockRestore();
+  });
+
   describe("isWhitelistedDomain (via getInitialURL)", () => {
     const mockManifest = {
       url: "https://default.example.com/",
       domains: ["ledger.com", "*.subdomain.ledger.com", "approved.io"],
     };
+
+    const DOMAIN_ERROR =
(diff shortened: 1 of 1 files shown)


francois-guerin-ledger (approved):
A bit cumbersome I think. Can't we just call `@ledgerhq/logs` instead ?

gre-ledger:
it's the src code that uses console.error, i didn't want to change that behavior, i'm only fixing the code to silent the error from Jest logs.

Justkant (commented):
I'm not really sure I like this change because then the test is not focusing only on input and output but depends on implementation details now

gre:
The way the console.error was prefixed with `#isWhitelistedDomain:: ` made me think it was a strict feature to have in this code, and it's win-win to assert it is still logged, if we want we can remove the assert part and exclusively silent the error, but silenting an error without esserting why we do silent it would feel uncomplete and possibly can be dangerous too.
feel free to reconsider this (weither we want to keep the console error, or better way we want to assert it)

Justkant:
Yeah it shouldn't really be a concern for the test but we should probably use `@ledgerhq/logs` as suggested by @francois-guerin-ledger instead

If we use `@ledgerhq/logs` I suppose we don't need any special mocking, not sure if we should even assert the logs, the main thing is that I don't want this test to fail because I refactor the method used to log something or changed the message a bit, it won't impact if the function works or not

PR #17113, opened 2026-05-04
[LWDM] fix(llc): prevent top level internal operation

francois-guerin-ledger:
### ✅ Checklist

  - ...

### 📝 Description

In `parentOpsForTxWithOnlyInternalOperations`, after building the synthetic NONE parent, the bridge was re emitting each internal operation as an additional standalone top-level entry. This caused internal operations (e.g. an ETH `IN` triggered by a smart contract) to appear twice in the account's operation list — once nested under the NONE parent and once at the top level.

The fix removes the loop responsible for that re-emission. Internal operations are now only reachable via `internalOperations` on the NONE parent, consistent with how internal operations are handled in all other cases.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-29873
- **ADR link (if any)**: 

---

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
+++ b/libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
@@ -157,7 +157,7 @@ function syntheticParentForTokenOnlyTx(
   // In the case of smart contract interaction, the contract must be the recipient of the parent operation => this
   // is why we need to extract this information from the operation details.
   const contract = getTokenContract(referenceOp);
-  const parentRecipients = contract !== undefined ? [contract] : referenceOp.recipients ?? [];
+  const parentRecipients = contract !== undefined ? [contract] : (referenceOp.recipients ?? []);
   const parentSenders = referenceOp.senders ?? [];
   return cleanedOperation({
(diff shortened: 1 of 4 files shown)


libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
```diff
@@ -221,45 +222,33 @@ function parentOpsForTxWithOnlyInternalOperations(
...
-      cleanedOperation({
-        ...internalOp,
-        subOperations,
-        internalOperations,
-      }),
-    );
-  }
-  return out;
+  return [
+    cleanedOperation({
```
qperrot, line 227: did you test the case when feePayer is null (behaviours of the old if statement) ?
and also special case like referenceOp.recipients === [] ?

qperrot (approved):
LGTM!
One question about test but all good

libs/ledger-live-common/src/bridge/generic-alpaca/getAccountShape.ts
francois-guerin-ledger: The parent is `NONE`, so the wallet won't even print it. Internal operations are displayed as is.

PR #17038, opened 2026-04-30
[LWDM] refactor(mobile): prepare Web3AppWebview for async getAccountBridge

gre-ledger:
### ✅ Checklist

- [ ] `npx changeset` was attached.
- [x] **Covered by automatic tests.** `liveSDKLogic.test.ts` updated for the async signature.
- [ ] **Impact of the changes:** Pure forward-compatibility refactor, no user-facing change.

### 📝 Description

Prepares mobile Web3AppWebview and WebPlatformPlayer signing flows for async `getAccountBridge` by making `prepareSignTransaction` async. Call sites that rely on `onSuccess`/`onError` callbacks wrap the `await` in `try/catch` to route failures to `onError`.

```diff
- function prepareSignTransaction(...): Transaction {
-   const bridge = getAccountBridge(account, parentAccount);
+ async function prepareSignTransaction(...): Promise<Transaction> {
+   const bridge = await getAccountBridge(account, parentAccount);
```

```diff
  uiHook["transaction.sign"] = async ({ account, parentAccount, ... }) => {
+   try {
      const tx = await prepareSignTransaction(...);
      onSuccess(tx);
+   } catch (err) {
+     onError(err);
+   }
  };
```

### ❓ Context

- **JIRA or GitHub link**: [LIVE-29176](https://ledgerhq.atlassian.net/browse/LIVE-29176) / #16733
- **ADR link (if any)**: N/A

[LIVE-29176]: https://ledgerhq.atlassian.net/browse/LIVE-29176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #17155, opened 2026-05-05
[LWDM] LIVE-26176 remaining work

gre-ledger:
Showcase PR. not for actual merge.

## What was done in April 2025

Since merge of #16031 (tech enabler for wallet-cli), coin modules are loaded on demand, improving the performance of our apps: coin-* can sometimes be heavy on the fact they themselves will import a lot of code and a lot of libraries and sdks, doing it at boot had a cost. **Typical gain: 1.4s->1.0s** of boot time.

This also partially addresses [ADR-005](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/5822054449/Coin+modules+-+ADR-005+-+Remove+generated+files) of coin-integration team: we removed live-common/src/generated/ folder.

But this works with a tradeoff: we dynamically use `require()` synchronously, those organised under `libs/live-common/src/coin-modules/registry`: this still make the current implementation relatively fragile: bundles could easily end up with CJS and ESM concurrent modules that both share their own state. *Side note: a good module should be pure and not have a state, if a state is needed, `global.__myState` is a workaround to prefer over local variable.*

The goal is now to head to a better architecture: decouple the coins implementation from the generic parts. A few iterative improvements has been shipped toward this goal without problem so far. See [ADR-033](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7041482926/Coin+modules+-+ADR-033+-+Reverse+CoinModule+Wallet+dependency) for a more long term plans.

## What remains to make live-common ESM-only (loaders using import())

This PR reflects what remains to be done to fully migrate to `import()` in `libs/live-common/src/coin-modules/loaders.ts`. Even tho most of this work will be delivered separately, it's still good to have the big picture by this PR. Each commit is atomic and could be shipped into individual PRs (they are not all independent).

The 15 commits on this branch, in apply order, map to the following PRs:

**AccountBridgeExtensions migration** (unlocks the deprecation step below)
- #17484
- #17486
- #17487
- #17491
- #17492
- #17493
- #17494

**Drop top-level helpers** (depends on the migrations above)
- #17338

**Prepare call sites for async getAccountBridge**
- #17038
- #17752
- #17402

**Async final activation** 
- `feat(live-common): move loaders to import() and make bridge async`
- `test: adapt tests for async getAccountBridge`

### Risk to eliminate

- [x] With the Suspense introduction we will need to confirm there is no issue. If #16216 or #16221 could land before, it may help secure this but is currently not yet a focus (team-platform).
- [x] according to e2e mobile, something fails in `  ● Cosmos delegate flow › goes through the delegate flow`

## Future

- applying the same on LLD/LLM ui, shall also be possible and first estimation show we would go **1.0s -> 0.7s**


gre-ledger:
[26342fe](https://github.com/LedgerHQ/ledger-live/pull/17155/commits/26342fef866524a275ffb5f9fd6f6dc404e13621) only fails due to a bad conflict resolve i did, but otherwise the tree is pretty green. 
WIP commit pushed but unfinished parts that i'm still iterating on. using it to double check things with the CI.
Desktop app runs fine but i need to give more manual test.
e2e not yet run, need to finalize.

apps/ledger-live-desktop/src/renderer/families/aleo/hooks/useAleoPrivateSync.ts
```diff
@@ -98,15 +98,14 @@ export const useAleoPrivateSync = ({
...
+  const runSync = useCallback(async () => {
     const acc = accountRef.current;
     if (!isSyncingRef.current || acc?.type !== "Account" || !isAleoAccount(acc)) return;
 
     const currentAccountId = acc.id;
     let receivedFinalResult = false;
-    const sub = from(Promise.resolve(getAccountBridge(acc)))
-      .pipe(mergeMap(bridge => bridge.sync(acc, { paginationConfig: {}, syncType: SYNC_TYPE_SHIELDED })))
-      .subscribe({
+    const bridge = await getAccountBridge(acc);
```
gre-ledger: undo, the previous approach was better because promise was pipelined to rxjs.

apps/ledger-live-desktop/src/renderer/families/aleo/hooks/useAleoPrivateSync.ts
```diff
@@ -98,15 +98,14 @@ export const useAleoPrivateSync = ({
   const [progress, setProgress] = useState(keepAliveEntry?.progress ?? 0);
   const [error, setError] = useState<Error | null>(null);
 
-  const runSync = useCallback(() => {
+  const runSync = useCallback(async () => {
```
gre-ledger: undo

libs/ledger-live-common/src/account/formatters.ts
```diff
@@ -86,9 +86,9 @@ const cliFormat = (account, level?: string) => {
     getAccountCurrency(account).units[0],
   );
 
-  const formatAccountSpecifics = getAccountBridge(account).formatAccountSpecifics;
-  if (formatAccountSpecifics) {
-    str += formatAccountSpecifics(account);
+  const bridge = (getAccountBridge(account) as any).value;
```
gre-ledger: this can't work. it's async now.

libs/ledger-live-common/src/account/formatters.ts
```diff
@@ -124,7 +124,8 @@ const cliFormat = (account, level?: string) => {
           console.error("unexpected missing token account " + id);
         },
         (operation, unit) => {
-          return getAccountBridge(account).formatOperationSpecifics?.(operation, unit) ?? "";
+          const bridge = (getAccountBridge(account) as any).value;
```
gre-ledger: this can't work it's async now.

libs/ledger-live-common/src/account/formatters.ts
```diff
@@ -213,8 +214,8 @@ export function formatOperation(account: Account | null | undefined): (arg0: Ope
 
   const familyExtra = (operation, unit) => {
     if (!account) return "";
-
-    return getAccountBridge(account).formatOperationSpecifics?.(operation, unit) ?? "";
+    const bridge = (getAccountBridge(account) as any).value;
```
gre-ledger: this can't work it's async now.

apps/ledger-live-mobile/android/app/build.gradle
```diff
@@ -133,7 +133,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 36176158
-        versionName "3.100.1"
+        versionName "4.3.0"
```
gre-ledger: undo

libs/ledger-live-common/src/flows/send/hooks/useSendFlowAmountReviewCore.ts
```diff
@@ -56,14 +56,22 @@ export function useSendFlowAmountReviewCore({
   );
   const accountCurrency = useMemo(() => getAccountCurrency(mainAccount), [mainAccount]);
 
+  const bridgeRef = useRef<AccountBridge<Transaction> | null>(null);
+  useEffect(() => {
+    getAccountBridge(account, parentAccount ?? undefined).then(bridge => {
+      bridgeRef.current = bridge;
+    });
+  }, [account, parentAccount]);
```
gre-ledger: this is not the good approach for this. we must prefer `useAccountBridgeOrNull` because it's going to properly cache, maybe the bridge shall be sent to the hook. alternatively and maybe preferably: i think we can accept a lil async behavior in the `updateTransactionWithPatch` hook: just keep doing the getAccountBridge there, just await it.

gre-ledger:
closed in favor of https://github.com/LedgerHQ/ledger-live/pull/17760

PR #17222, opened 2026-05-06
[LWM] fix(lwm): fix Ledger logo overlay persisting on iOS after app resume

iqbalibrahim-ledger:
### ✅ Checklist

  - iOS only. QA should verify the Ledger logo overlay disappears correctly after single and rapid background/foreground cycles, and that the privacy blur screen still appears when the app is sent to background.

### 📝 Description

On iOS, the Ledger "L" logo privacy overlay (shown when the app is backgrounded to hide sensitive data from the app switcher) was persisting on screen after the app was brought back to the foreground.

**Root cause — `Overlay.swift`:**
`showOverlay()` created new `UIVisualEffectView` + `UIImageView` objects on every call and identified them only by integer tags (`12345`, `12346`). `hideOverlay()` used `viewWithTag()` to find them, which returns only the **first** match in the view hierarchy. When the user rapidly cycled between background and foreground during the 0.5s fade-out animation, a second overlay was added on top of the first. `hideOverlay()` always targeted the first view, leaving the second permanently stuck on screen.

**Fix:**
- Added `overlayBlurView: UIVisualEffectView?` and `overlayLogoView: UIImageView?` as stored properties on `AppDelegate` to track the views by direct reference instead of tag lookup.
- `showOverlay()` now guards with `guard overlayBlurView == nil` to prevent stacking.
- `hideOverlay()` nils out the stored references immediately before starting the animation, so any concurrent `showOverlay()` call can safely create a fresh overlay while the old one fades out.

**Secondary fixes:**
- `useIsAppInBackground.tsx`: added missing `[]` dependency array to `useEffect` to prevent the AppState listener from being re-registered on every render.
- `auth.hooks.ts`: replaced `useState` with `useRef` for tracking `appState` so `handleAppStateChange` has a stable identity, preventing the AppState listener in `AuthPass` from being torn down and re-added on every state transition.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-30318](https://ledgerhq.atlassian.net/browse/LIVE-30318)
- **ADR link (if any)**: N/A

---

[LIVE-30318]: https://ledgerhq.atlassian.net/browse/LIVE-30318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iqbalibrahim-ledger:
Tested successfully on iPhone 12 Mini with [iOS Build](https://github.com/LedgerHQ/ledger-live-build/actions/runs/25446603293)


https://github.com/user-attachments/assets/6574cbad-4a4e-4392-99c6-3ac52414be04

PR #17245, opened 2026-05-07
docs(agents): add ongoing rework rule for async bridge migration

gre-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** N/A — docs only
  - Agents and contributors are now guided to avoid synchronous calls to `getAccountBridge` / `getCurrencyBridge`

### 📝 Description

Adds an "Ongoing Rework" section to `AGENTS.md` to document that `getAccountBridge` and `getCurrencyBridge` are being migrated to async (tracked in PR #17155). Agents and contributors must use the appropriate pattern depending on context:

- React: `useAccountBridge` / `useCurrencyBridge` hooks
- RxJS: `from(Promise.resolve(getAccountBridge(...))).pipe(mergeMap(bridge => ...))`
- async: `const bridge = await getAccountBridge(...)`

### ❓ Context

- **JIRA or GitHub link**: #17155
- **ADR link (if any)**: N/A

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,3 +14,10 @@
 
 - Before finishing any agentic code change, run static checks for the affected scope
 - See [/docs/dev/validate-before-finishing.md](/docs/dev/validate-before-finishing.md)
+
+## Ongoing Rework
+
+- `getAccountBridge` and `getCurrencyBridge` must not be called synchronously (becoming async, PR #17155).
+  - React: use `useAccountBridge` / `useCurrencyBridge` hooks
+  - RxJS: `from(Promise.resolve(getAccountBridge(...))).pipe(mergeMap(bridge => ...))`
+  - async: `const bridge = await getAccountBridge(...)`


LL782 (commented):
Hey @gre-ledger – a few thoughts from me:

### 1. Do we need it?

Is this a real issue or an assumed one? Perhaps we should wait until we see these rules being violated before we add guardrails to prevent that.

### 2. Does it need to be in global context?

I think this should be in a skill rather than `AGENTS.md`. I've drafted guidance around that here: 
- [/docs/about-docs/docs-locations.md](https://github.com/LedgerHQ/ledger-live/blob/1d8d16e6b1cebd6c8419a56e815c29242160e412/docs/about-docs/docs-locations.md) and here 
- [/docs/about-docs/progressive-disclosure.md](https://github.com/LedgerHQ/ledger-live/blob/1d8d16e6b1cebd6c8419a56e815c29242160e412/docs/about-docs/progressive-disclosure.md)

### 3. Can we make it easier to understand?

I think the writing is not very clear. Would moving it into a skill, with frontmatter and a body provide more space for clearer writing?

gre-ledger:
ok let's not change the agent rules at all, here is what we'll more simply do:
- in the `getAccountBridge` and `getCurrencyBridge` code implementation location, we will introduce a JS doc comment above the function to explicitly inform of the proper usage. if LLM see the function and want to use it, it shall be able to infer it.
- if a LLM don't see the code implementation locally, the other way it will do is copy pattern from existing usage, good news is all usage is now properly migrated, so it should be able to yield correct code.

PR #17353, opened 2026-05-12
[LWDM] feat(dmk): add shared session registry

OlivierFreyssinet:
### ✅ Checklist

  - DMK session reuse in legacy BLE, HID, and desktop transport adapters
  - Cleanup of disconnected DMK sessions
  - `useDeviceSessionState` consumers
  - BLE pairing session registration

### 📝 Description

This PR introduces a shared DMK session registry used by the DMK transport packages to track active device sessions. The registry stores only the minimal session ownership data, `{ sessionId, dmk }`, and is responsible for subscribing to each session state. When DMK reports `DeviceStatus.NOT_CONNECTED`, or when the session state stream ends, the registry removes the session and releases its subscription.

Previously, active session state was split across incompatible implementations. BLE and desktop transports relied on `@ledgerhq/live-dmk-shared`'s `activeDeviceSessionSubject`, while the mobile HID transport had its own local subject with the same name. Those subjects stored transport instances, which made session reuse depend on a specific adapter object instead of the DMK session itself. That made it easy for different flows to miss each other's sessions, and it also mixed two responsibilities: tracking an active DMK session and returning a legacy transport implementation.

The new registry separates those concerns. It centralizes session lifecycle tracking, but it does not store transport instances. Legacy adapters now query the registry for a reusable DMK session and then create their own transport instance around the matching `sessionId`, preserving each adapter's own `exchange` implementation. This keeps BLE, HID, and desktop behavior compatible while making the session source of truth shared and extensible for multiple simultaneous sessions.

Concretely, this PR:

- Replaces the shared active-session subject with `ActiveDeviceSessionRegistry`.
- Removes the duplicate HID-local active-session subject.
- Moves active session cleanup to the registry through `dmk.getDeviceSessionState` subscriptions.
- Adds `dispose()`/`clear()` to release all registry subscriptions when the DMK provider unmounts.
- Updates BLE, HID, and desktop legacy transports to reuse registry sessions without reusing stored transport objects.
- Updates `useBleDevicePairing` to register sessions without creating a BLE transport only for subject compatibility.
- Updates shared hook and transport tests, and adds direct registry unit coverage.

### ❓ Context

- **ADR link (if any)**: N/A

---

### DIFF
--- a/.changeset/neat-sessions-register.md
+++ b/.changeset/neat-sessions-register.md
@@ -1,0 +1,7 @@
+---
+"@ledgerhq/live-dmk-shared": minor
+"@ledgerhq/live-dmk-mobile": minor
+"@ledgerhq/live-dmk-desktop": minor
+---
+
+Add a shared DMK session registry for active device sessions
--- a/libs/live-dmk-desktop/src/hooks/useDeviceManagementKit.tsx
+++ b/libs/live-dmk-desktop/src/hooks/useDeviceManagementKit.tsx
@@ -1,11 +1,15 @@
-import React, { createContext, useContext, useMemo } from "react";
+import React, { createContext, useContext, useEffect, useMemo } from "react";
 import {
   DeviceManagementKitBuilder,
   DeviceManagementKit,
   LogLevel,
 } from "@ledgerhq/device-management-kit";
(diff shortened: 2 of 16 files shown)


OlivierFreyssinet:
actually this is not required, we can use `dmk.listConnectedDevices` for new use cases

PR #17450, opened 2026-05-13
[LWDM] feat(zcash): live-signer-zcash boilerplate [LIVE-23288]

semeano:
### ✅ Checklist

  - @ledgerhq/coin-bitcoin
  - @ledgerhq/live-common
  - @ledgerhq/live-signer-zcash

### 📝 Description

Zcash live signer boilerplate; it includes the `getAddress` method.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-23288

---


libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts
```diff
@@ -16,12 +17,30 @@ const resolver = (signerContext: SignerContext): GetAddressFn => {
...
+            signer as unknown as {
+              getAddress: (path: string, display?: boolean) => Promise<ZcashAddress>;
+            }
+          ).getAddress(path, verify || false);
+          return {
+            bitcoinAddress: address,
+            publicKey: "",
+            chainCode: "",
+          };
+        })) as BitcoinAddress;
```
cted-ledger: I think we should use chain-adapter pattern here. We should never have any zcash specific code at first level

libs/ledger-live-common/src/families/bitcoin/setup.ts
```diff
@@ -17,12 +17,20 @@ import { GetAddressFn } from "@ledgerhq/ledger-wallet-framework/bridge/getAddres
...
+import { isDmkTransport } from "../../hw/dmkUtils";
 
-const createSigner = (transport: Transport, currency: CryptoCurrency) => {
+const createSigner = (transport: Transport, currency: CryptoCurrency): Btc => {
+  if (currency.id === "zcash") {
+    if (!isDmkTransport(transport)) {
+      throw new Error("Zcash requires DMK transport");
+    }
+    return new DmkSignerZcash(transport.dmk, transport.sessionId) as unknown as Btc;
+  }
```
cted-ledger: same here. please use chain adapter pattern

cted-ledger (changes requested):

cted-ledger (approved):

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -60,6 +73,31 @@ const zcashChainAdapter: ChainAdapter = {
...
+  getAddress(deviceId, { currency, path, verify }, signerContext: SignerContext) {
+    return signerContext(deviceId, currency, async signer => {
+      if (!("getAddress" in signer) || typeof signer.getAddress !== "function") {
+        throw new Error("Zcash signer must implement getAddress(path, display?)");
+      }
+      const { address } = await (
+        signer as unknown as {
+          getAddress: (path: string, display?: boolean) => Promise<ZcashAddress>;
+        }
+      ).getAddress(path, verify || false);
```
cted-ledger:
```suggestion
      if (!isZcashSigner(signer)) {
        throw new Error("Zcash signer must implement getAddress(path, display?)");
      }
      const { address } = await signer.getAddress(path, verify || false);
```

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -1,12 +1,25 @@
...
+  sessionId: string;
+};
+
+const isDmkTransport = (transport: unknown): transport is DmkTransport =>
+  !!transport &&
+  typeof transport === "object" &&
+  "dmk" in transport &&
+  "sessionId" in transport &&
+  typeof (transport as { sessionId: unknown }).sessionId === "string";
+
```
cted-ledger, line 24:
```suggestion
const isZcashSigner = (
  signer: unknown,
): signer is { getAddress: (path: string, display?: boolean) => Promise<ZcashAddress> } =>
  !!signer &&
  typeof signer === "object" &&
  "getAddress" in signer &&
  typeof signer.getAddress === "function";
```

libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts
```diff
@@ -34,7 +39,7 @@ const resolver = (signerContext: SignerContext): GetAddressFn => {
       throw e;
     }
 
-    const { bitcoinAddress, publicKey, chainCode } = result;
+    const { bitcoinAddress, publicKey, chainCode } = result as BitcoinAddress;
```
cted-ledger:
```ts
let result: BitcoinAddress | undefined;
    try {
      result = await adapter.getAddress?.(deviceId, options, signerContext);
      if (!result) {
        result = await signerContext(deviceId, currency, signer =>
          signer.getWalletPublicKey(path, {
            verify: verify || false,
            format,
          }),
        );
      }
    } catch (e: any) {
      // TODO Should normalize error returned from ledgerjs
      if (
        e &&
        e.message &&
        (e.message.includes("invalid format") || e.message.includes("Unsupported address format"))
      ) {
        throw new UnsupportedDerivation();
      }
      throw e;
    }

    const { bitcoinAddress, publicKey, chainCode } = result;
```

cted-ledger (changes requested):

libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts
semeano: Apart from avoiding a cast I don't see any advantages of these changes. Could you please clarify?
Meanwhile I removed the cast by making a small change.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -60,6 +85,62 @@ const zcashChainAdapter: ChainAdapter = {
     // TODO: implement PCZT transaction preparation (ZIP-317 fee info)
     return undefined;
   },
+
+  getAddress(deviceId, { currency, path, verify }, signerContext: SignerContext) {
+    return signerContext(deviceId, currency, async signer => {
+      if (!isZcashSigner(signer)) {
+        throw new Error("Zcash signer must implement getAddress(path, display?)");
```
may01, line 92: Should we fall back to the `hw-app` in this case?

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/xpub.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Compose a BIP32 extended public key (xpub) string from raw key material.
+ *
+ * Mirrors the algorithm used by `hw-app-btc` `BtcOld.getWalletXpub`, kept local
+ * to the zcash adapter because the new `DmkSignerZcash` (DMK signer kit) only
+ * exposes `getAddress` and not a native xpub command.
+ *
+ * Both `publicKey` arguments accept either uncompressed (65 bytes, 0x04 prefix)
+ * or already-compressed (33 bytes) SECP256K1 keys. Inputs are hex strings.
+ */
+export function composeXpub(args: {
```
may01, line 22: Do you think it will be right to move this logic to DMK?

libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts
```diff
@@ -44,7 +49,7 @@ const resolver = (signerContext: SignerContext): GetAddressFn => {
       address: bitcoinAddress,
       path,
       publicKey,
-      chainCode,
+      ...(chainCode ? { chainCode } : {}),
```
may01, line 50: Why `chainCode` became optional?

semeano:
🖥️ Desktop E2E test report: https://github.com/LedgerHQ/ledger-live/actions/runs/25923123681

libs/ledger-live-common/src/families/bitcoin/setup.ts
```diff
@@ -11,18 +11,22 @@ import makeCliTools from "@ledgerhq/coin-bitcoin/cli-transaction";
...
 import { GetAddressFn } from "@ledgerhq/ledger-wallet-framework/bridge/getAddressWrapper";
 import { getCurrencyConfiguration } from "../../config";
 import { BitcoinConfigInfo } from "@ledgerhq/coin-bitcoin/config";
 import { SignMessage } from "../../hw/signMessage/types";
 
-const createSigner = (transport: Transport, currency: CryptoCurrency) => {
+const createSigner = (transport: Transport, currency: CryptoCurrency): Btc => {
+  const adapter = getChainAdapter(currency.id);
+  const custom = adapter.createSigner?.(transport, currency);
+  if (custom) return custom as unknown as Btc;
```
may01, line 25: I believe that signer we get from `adapter.createSigner` is the DMK struct. Do we verify it's compatible with BTC  type?

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/xpub.ts
semeano: It does make sense yes. Maybe we can address that in another ticket?

libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts
semeano: `chainCode` was already optional. Check `libs/ledger-wallet-framework/src/derivation.ts`

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
semeano: That is already being done inside `libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts`

libs/ledger-live-common/src/families/bitcoin/setup.ts
semeano: Yes It will return a new `DmkSignerZcash` but `createSigner` is intentionally loose.
If you check `libs/coin-modules/coin-bitcoin/src/chain-adapters/types.ts:102` the comment says “non-BTC signer implementations,” so it's by design that the returned object may not implement the full BitcoinSigner (a.k.a. Btc) interface.

Let me know if that answered your question.

libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts
cted-ledger: 1) avoiding a cast is enough :)

strict typing is always better:
```ts
let result: BitcoinAddress | undefined; 
```
is far better than
```ts
let result;
```

2) awaiting `getAddress` is really more elegant than doing
```ts
result = await custom;
```
this also prevent the need to have `else` branch

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
cted-ledger: Are you sure ? It seems the catch in `libs/coin-modules/coin-bitcoin/src/hw-getAddress.ts` re-throw the error no ?

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/xpub.ts
cted-ledger: Since hw-app-btc is being deprecated, it makes sense to have this BIP32 serialization logic in coin-bitcoin. But it shouldn't live in the Zcash adapter specifically, nor in the DMK Zcash signer...
This is standard BIP32, not Zcash-specific. Any future chain adapter whose signer only returns raw keys will need the same composeXpub. Could you move it to a shared location (e.g. coin-bitcoin/src/utils/) ?

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
semeano: To be honest maybe this check doesn't make sense, given the only way for it to be thrown is if the DMK signer doesn't have `getAddress` implemented, which it has, so...

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/xpub.ts
semeano: This BIP32 serialization already exists inside `hw-app-btc` but it has device-specific APDU's. It would require a refactor to be able to share this code with a chain adapter, which is something we should avoid at all costs IMO.

may01: For me it will be good to have a separate task to track this topic. And resolve it when we have resources.

may01 (dismissed):

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
cted-ledger: We decided together to return undefined instead of rejecting error

gre-ledger (approved):

cted-ledger (approved):

may01 (approved):

PR #17504, opened 2026-05-15
[LWM] fix(live-mobile): duplicated known devices Redux slice

benruseau:
### ✅ Checklist

- [x] **Covered by automatic tests.** Non-regression tests cover HID updates when DMK reports a new id for the same device model, and when multiple HID known devices share the same model but only one exact id should be updated.
  - Mobile known devices deduplication now handles HID devices by first matching the exact id, then falling back to model-based matching only when no id match exists.
  - This avoids updating the wrong known HID device when multiple devices of the same model are stored.
  - QA should focus on the device selection flow on mobile, especially USB/HID discovery and reconnection for Nano S Plus / Flex-style devices.
  - BLE known device matching remains unchanged and continues to use the existing `findMatchingOldDevice` logic.

### 📝 Description

This PR fixes duplicated or incorrectly updated known devices in Ledger Live Mobile for HID devices.

Previously, HID matching relied only on the device model. That handled cases where DMK reports a different runtime id for the same USB device, but it could update the wrong entry when multiple known HID devices of the same model were stored.

The reducer now matches HID known devices by transport and exact id first. If no exact id match exists, it falls back to model-based matching to preserve the id-format migration behavior. If neither match exists, the device is inserted as a new known device.

The PR also adds reducer non-regression tests for both HID update paths and includes a patch changeset for `live-mobile`.

### ❓ Context

- **JIRA**: [LIVE-30665](https://ledgerhq.atlassian.net/browse/LIVE-30665)
---

[LIVE-30665]: https://ledgerhq.atlassian.net/browse/LIVE-30665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/reducers/knownDevices.ts
+++ b/apps/ledger-live-mobile/src/reducers/knownDevices.ts
@@ -123,6 +123,14 @@ function findMatchingKnownDevice(
   const oldDevicesForTransport = knownDevices.filter(
     device => device.transport === newDevice.transport,
   );
+
+  if (newDevice.transport === rnHidTransportIdentifier) {
+    return (
+      oldDevicesForTransport.find(device => device.deviceModelId === newDevice.deviceModelId) ??
+      null
+    );
+  }
+
   const matchingOldDevice = findMatchingOldDevice(
     mapKnownDeviceToDeviceBaseInfo(newDevice),
     oldDevicesForTransport.map(mapKnownDeviceToDeviceBaseInfo),
--- a/.changeset/cyan-dryers-develop.md
+++ b/.changeset/cyan-dryers-develop.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Fix duplicated known devices Redux slice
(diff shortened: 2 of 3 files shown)


apps/ledger-live-mobile/src/reducers/knownDevices.ts
```diff
@@ -123,6 +123,20 @@ function findMatchingKnownDevice(
   const oldDevicesForTransport = knownDevices.filter(
     device => device.transport === newDevice.transport,
   );
+
+  if (newDevice.transport === rnHidTransportIdentifier) {
+    const matchingDeviceById = oldDevicesForTransport.find(device => device.id === newDevice.id);
```
OlivierFreyssinet: I'm not sure we should do that as on RN HID all devices have the same id, or an id generated at runtime, the id is not reliable across connections.
Maybe we should just match with the device model id so it gives the user the opportunity to have 1 device per model, and if a new model appears, it is rightfully added to the slice.

OlivierFreyssinet (approved):

PR #17539, opened 2026-05-18
feat(wallet-cli): add assets token resolver commands

gre-ledger:
### ✅ Checklist

  - Adds two read-only commands under a new `assets` group in wallet-cli. No behavior change to existing commands.

### 📝 Description

Adds `wallet-cli assets token <network> <address>` and `wallet-cli assets token-by-id <id>` — thin wrappers over `CryptoAssetsStore.findTokenByAddressInCurrency` and `findTokenById`. Lets agents and scripts resolve the canonical Ledger token id (e.g. for `swap quote --from` / `--to`) from a contract address instead of from a ticker.

Consumer side:

```bash
wallet-cli assets token ethereum 0xdac17f958d2ee523a2206206994597c13d831ec7
wallet-cli assets token-by-id ethereum/erc20/usd_tether__erc20_
```

Screenshot 2026-05-18 at 14 32 03

### ❓ Context

- **JIRA or GitHub link**: [LIVE-30697](https://ledgerhq.atlassian.net/browse/LIVE-30697)
- **ADR link (if any)**: —

[LIVE-30697]: https://ledgerhq.atlassian.net/browse/LIVE-30697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-cli/src/commands/assets/index.ts
+++ b/apps/wallet-cli/src/commands/assets/index.ts
@@ -1,0 +1,9 @@
+import { defineGroup } from "@bunli/core";
+import TokenCommand from "./token";
+import TokenByIdCommand from "./token-by-id";
+
+export default defineGroup({
+  name: "assets",
+  description: "Crypto-assets store queries (resolve tokens by address or id)",
+  commands: [TokenCommand, TokenByIdCommand],
+});
--- a/apps/wallet-cli/src/test/commands/assets-token-by-id.test.ts
+++ b/apps/wallet-cli/src/test/commands/assets-token-by-id.test.ts
@@ -1,0 +1,79 @@
+import { describe, it, expect, beforeAll, afterAll } from "bun:test";
+import { MockServer } from "../helpers/mock-server";
+import { runCli } from "../helpers/cli-runner";
+import { USDT_CONTRACT, USDT_API_RESPONSE } from "../helpers/cal-fixtures";
+
+describe("assets token-by-id command", () => {
+  const server = new MockServer([
+    {
+      method: "GET",
+      match: /\/v1\/tokens.*id=/i,
+      response: USDT_API_RESPONSE,
+    },
+  ]);
+
+  beforeAll(() => server.start());
+  afterAll(() => server.stop());
+
+  it("human output: prints token details for a valid id", async () => {
+    const { stdout, exitCode, stderr } = await runCli(
+      ["assets", "token-by-id", "ethereum/erc20/usd_tether__erc20_"],
+      { WALLET_CLI_MOCK_PORT: String(server.port) },
+    );
+    expect(exitCode, `stderr: ${stderr}`).toBe(0);
+    expect(stdout).toContain("ethereum/erc20/usd_tether__erc20_");
+    expect(stdout).toContain("USDT");
+    expect(stdout).toContain(USDT_CONTRACT);
+  });
+
+  it("json output: returns a valid envelope", async () => {
+    const { stdout, exitCode, stderr } = await runCli(
+      ["assets", "token-by-id", "ethereum/erc20/usd_tether__erc20_", "--output", "json"],
+      { WALLET_CLI_MOCK_PORT: String(server.port) },
+    );
+    expect(exitCode, `stderr: ${stderr}`).toBe(0);
+    const data = JSON.parse(stdout);
+    expect(data.status).toBe("success");
+    expect(data.command).toBe("assets token-by-id");
+    expect(data.network).toBe("ethereum");
+    expect(data.token).toMatchObject({
+      id: "ethereum/erc20/usd_tether__erc20_",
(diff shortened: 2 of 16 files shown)


apps/wallet-cli/src/test/commands/assets-token-by-id.test.ts
LL782, line 1: **suggestion: reflect the same file structure as the implementation files**

Why are the tests in a `test` directory? Why not:
- apps/wallet-cli/src/commands/assets/token-by-id.test.ts
- apps/wallet-cli/src/commands/assets/token-by-id.ts

### Alternatively 

And if there is a good reason to use the test directory why fully reflect the implementation path:
```diff
- apps/wallet-cli/src/test/commands/assets-token-by-id.test.ts
+ apps/wallet-cli/src/test/commands/assets/token-by-id.test.ts
```

apps/wallet-cli/src/test/commands/assets-token-by-id.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    expect(data.command).toBe("assets token-by-id");
+    expect(data.network).toBe("ethereum");
+    expect(data.token).toMatchObject({
+      id: "ethereum/erc20/usd_tether__erc20_",
+      ticker: "USDT",
+      decimals: 6,
+    });
+  });
+
+  it("exits non-zero when the id arg is missing", async () => {
```
LL782, line 46: **nit: prefer more natural business language for test names**

I think `expect(exitCode).toBe(1);` is perfect for the body of the test but the name should be natural language, e.g.

```suggestion
  it("errors when no ID is provided", async () => {
```

The same comment applies below as well:

```diff
- it("exits non-zero when the store returns no token", async () => {
+ it("errors when no token is found", async () => {
```

For me the test names stay high level and the details can be found in the body of the test

LL782: These same comments apply to the next test file too.

apps/wallet-cli/src/commands/assets/index.ts
```diff
@@ -0,0 +1,9 @@
+import { defineGroup } from "@bunli/core";
+import TokenCommand from "./token";
+import TokenByIdCommand from "./token-by-id";
+
+export default defineGroup({
+  name: "assets",
+  description: "Crypto-assets store queries (resolve tokens by address or id)",
```
LL782, line 7: **nit: use uppercase for "ID"?**

```suggestion
  description: "Crypto-assets store queries (resolve tokens by address or ID)",
```

apps/wallet-cli/src/wallet/formatter/human.ts
```diff
@@ -59,6 +65,22 @@ export class HumanFormatter {
...
+      `ID:        ${colors.bold(t.id)}`,
+      `Ticker:    ${t.ticker}`,
+      `Name:      ${t.name}`,
+      `Contract:  ${t.contractAddress}`,
+      `Parent:    ${t.parentCurrencyId}`,
+      `Type:      ${t.tokenType}`,
+      `Decimals:  ${t.decimals}`,
+    ];
+    if (t.delisted) {
+      lines.push(colors.yellow("Warning: this token is delisted"));
```
LL782, line 79: Is this readable on a light themed terminal? Can you set the background the black as well?

LL782 (approved):
**Idea: move customer facing strings into locale files**

There are quite a lot of places where I would suggest little changes to the messages coming back from the CLI. It makes me consider that perhaps we should move all the strings into locale files to make it easy to iterate on the _language_ being used now or in the future.

gre-ledger:
there is no locales for this CLI atm. (unlike desktop and mobile)

apps/wallet-cli/src/commands/assets/index.ts
gre-ledger: good point. that said i think overall the concept of "id" will need refinement, like we need more precision of the concept (e.g. maybe it should be called AssetID or CurrencyID everywhere)

apps/wallet-cli/src/wallet/formatter/human.ts
gre-ledger: Image

i think it's fine. people can map yellow to any theme too. it's an alias.

apps/wallet-cli/src/test/commands/assets-token-by-id.test.ts
gre-ledger: that current choice is half consistent with the existing group, half non consistent 😆 (both session and account commands are done like this, only swap commands have its own folder like you asked) => refactoring can be done but need not only to address `assets` group to unify to a same convention)

PR #17597, opened 2026-05-19
[LWDM] fix(coin-bitcoin): Zcash transparent send crash

cted-ledger:
### ✅ Checklist

  - Zcash transparent send flow (was crashing, now works)
  - Zcash shielded sync / UFVK export (unchanged, still uses DMK overlay)
  - Other Bitcoin-family coins (no impact, chain adapter falls through to default)

### 📝 Description

**Bug:** Sending a transparent ZCash transaction crashes with `btc.splitTransaction is not a function`.

**Root cause:** The Zcash chain adapter's `createSigner` returned a raw `DmkSignerZcash` instance which does not implement the `BitcoinSigner` interface (`splitTransaction`, `createPaymentTransaction`). When the `signOperation` falls back to the standard Bitcoin legacy path (PCZT not yet implemented), it calls `signer.splitTransaction()` on the DMK signer and crashes.

**Fix:** `createSigner` now receives the default hw-app-btc signer as a `defaultSigner` parameter and augments it with DMK-specific methods (`getAddress`, `getFullViewingKey`) via `Object.assign`. This preserves all `BitcoinSigner` methods needed by the legacy signing path while exposing DMK capabilities for address resolution, xpub derivation, and viewing key export.

Additionally, `setup.ts` now uses `BitcoinSigner` throughout instead of casting to `Btc`, and the `ChainAdapter.createSigner` interface is properly typed.

### ❓ Context

- **JIRA or GitHub link**: LIVE-30736

### DIFF
--- a/libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
+++ b/libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
@@ -152,11 +152,17 @@ const zcashChainAdapter: ChainAdapter = {
     });
   },
 
-  createSigner(transport, _currency) {
-    if (!isDmkTransport(transport)) {
-      throw new Error("Zcash requires DMK transport");
-    }
-    return new DmkSignerZcash(transport.dmk, transport.sessionId);
+  createSigner(transport, _currency, defaultSigner) {
+    if (!isDmkTransport(transport)) return undefined;
+
+    // Augment the default BitcoinSigner with DmkSignerZcash methods.
+    // This gives chain adapter overrides (getAddress, getWalletXpub, getFullViewingKey)
+    // access to the DMK signer, while signOperation keeps using BitcoinSigner methods from Btc.
+    const dmk = new DmkSignerZcash(transport.dmk, transport.sessionId);
+    return Object.assign(defaultSigner, {
+      getAddress: dmk.getAddress.bind(dmk),
+      getFullViewingKey: dmk.getFullViewingKey.bind(dmk),
+    });
   },
 };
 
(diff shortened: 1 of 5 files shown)


libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -152,11 +152,17 @@ const zcashChainAdapter: ChainAdapter = {
...
+  createSigner(transport, _currency, defaultSigner) {
+    if (!isDmkTransport(transport)) return undefined;
+
+    // Augment the default BitcoinSigner with DmkSignerZcash methods.
+    // This gives chain adapter overrides (getAddress, getWalletXpub, getFullViewingKey)
+    // access to the DMK signer, while signOperation keeps using BitcoinSigner methods from Btc.
+    const dmk = new DmkSignerZcash(transport.dmk, transport.sessionId);
+    return Object.assign(defaultSigner, {
+      getAddress: dmk.getAddress.bind(dmk),
+      getFullViewingKey: dmk.getFullViewingKey.bind(dmk),
```
semeano, line 164: Do we need to add `getWalletXpub`?

cted-ledger: No, getWalletXpub is an adapter-level method, not a signer method. It already delegates to signer.getAddress() internally, which is overlaid. DmkSignerZcash doesn't have a getWalletXpub to overlay.

semeano (approved):

hedi-edelbloute (approved):

PR #17760, opened 2026-05-25
[LWDM] feat(live-common)!: switch coin-module loaders to import() and make bridge async

gre-ledger:
> [!IMPORTANT]
> This activates **async coin-module loading** across the bridge and leverages **React Suspense** everywhere in Ledger Live Desktop and Mobile. It finalizes ~2 months of iteration on the migration. Full write-up: [Notes on Shipping Async Bridge in Ledger Live monorepo](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7168229588/Notes+on+Shipping+Async+Bridge+in+Ledger+Live+monorepo).

### ✅ Checklist

  - see the full async-bridge / Suspense QA checklist listed in [LIVE-29187](https://ledgerhq.atlassian.net/browse/LIVE-29187) (Send / Receive / Stake / Add-account / Swap, Buy, Sell across all families, plus cold start, throttled network and coin-module load-failure scenarios on both desktop and mobile).

### 📝 Description

Final step of the async coin-module migration: replace synchronous `require()` lazy-loading with dynamic `import()` in the coin-module registry, and propagate `Promise<...>` through the bridge layer. Earlier PRs prepared all call sites; this PR flips the actual signatures and removes the temporary sync shims.

Highlights:
- `coin-modules/loaders.ts` — every entry is now `() => import(...).then(m => m.default)`.
- `coin-modules/registry.ts` + `types.ts` — `loadXxxForFamily` are async with a per-family Promise cache (`makeLoaderCache`). Added `getLoadedMockAccountForFamily` (sync, deprecated) backed by a Map filled by `loadMockAccountForFamily` on resolve.
- `bridge/impl.ts` — `getCurrencyBridge` / `getAccountBridge` / `getAccountBridgeByFamily` now return `Promise<...>`. Promise caches are annotated with React's `use()` hint fields (`status`/`value`/`reason`), so subsequent reads return synchronously.
- `bridge/generic-coin-framework/{signer,accountBridge,currencyBridge,validateAddress}` are async.
- Hooks/consumers (`useAccountBridge`, `useCurrencyBridge`, `useAccountBridgeMany`, `useBridgeTransaction`, `BridgeSync`, `mockHelpers`) drop the temporary sync shims and `await` / `.then(...)` the Promise.
- `mock/account.ts` — added async `genMockAccount(id, { currency, ... })` that pre-loads the family's mock module so `genAccountEnhanceOperations` is applied (fixes the cosmos delegations / algorand opt-ins drop introduced by the previous attempt on this branch). The sync `genAccount` is renamed `genAccountLegacy` and kept as a deprecated alias.

Code-breaking change for direct (non-React) bridge callers:

```diff
- const bridge = getAccountBridge(account);
+ const bridge = await getAccountBridge(account);
```

```diff
- const bridge = getCurrencyBridge(currency);
+ const bridge = await getCurrencyBridge(currency);
```

**In React, prefer `useAccountBridge(account)` / `useCurrencyBridge(currency)` — both require a `<Suspense>` boundary in the parent tree (see prior PR `bridge-suspense-boundaries`).**

### ❓ Context

- **JIRA or GitHub link**: [LIVE-29187](https://ledgerhq.atlassian.net/browse/LIVE-29187)
- **ADR link (if any)**: [ADR — Coin Module Loading Strategy](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7049511127/ADR+Coin+Module+Loading+Strategy)

[LIVE-29187]: https://ledgerhq.atlassian.net/browse/LIVE-29187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre-ledger:
ah finally Copilot is able to emit some comment. i was blind on the other PR 😆

francois-guerin-ledger:
I am launching Coin Modules Monitoring: https://github.com/LedgerHQ/ledger-live/actions/runs/26391291427

libs/ledger-live-common/src/mock/account.ts
```diff
@@ -190,25 +200,45 @@ export function genAccount(id: number | string, opts: GenAccountOptions = {}): A
...
+ * currency would re-introduce the silent-skip bug this function is meant
+ * to fix.
+ */
+export async function genMockAccount(
+  id: number | string,
+  opts: GenAccountOptions & { currency: CryptoCurrency },
+): Promise<Account> {
+  await loadMockAccountForFamily(opts.currency.family);
+  return genAccountLegacy(id, opts);
+}
```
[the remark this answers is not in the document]
gre-ledger: this is a false positive. this works as intended: 
- some code in this file is deliberately left deprecated, we're not ready to update the 500+ callsites of the **synchronous genAccount**, this **asynchronous** `genMockAccount` is however the way to go in future, it loads only the coin relevant to what you ask, making unit tests faster.
- the `genMockAccount` will only preload the family code, in order for `genAccountLegacy` to access it in sync. we leverage the deprecated path for code sharing here, but in future, we can imagine the loading will be inlined in the generator.

libs/ledger-live-common/src/mock/account.ts
```diff
@@ -190,25 +200,45 @@ export function genAccount(id: number | string, opts: GenAccountOptions = {}): A
...
-          } catch {
-            // to fix /src/__tests__/cross.ts, skip bridge error if there is no bridge in such currency
-          }
+          getAccountBridge(account)
+            .then(bridge => {
+              bridge.initAccount?.(account);
+            })
+            .catch(() => {
+              // to fix /src/__tests__/cross.ts, skip bridge error if there is no bridge in such currency
+            });
```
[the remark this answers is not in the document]
gre-ledger: this is not a problem: all the **mock** tests currently still pass. **the mock tests that become problematic need to switch to the non legacy way over the next months, using the new async genMockAccount way**, this is not realistic however to do this migration now, when I tried it recently, I end up with 500+ files to change.

libs/ledger-live-common/src/bridge/useBridgeTransaction.ts
```diff
@@ -300,19 +300,24 @@ const useBridgeTransaction = <T extends Transaction = Transaction>(
...
+      sub = bridge.sync(mainAccount, { paginationConfig: {} }).subscribe({
+        error: (_: Error) => {
+          // we do not block the user in case of error for now but it should be the case
+          dispatch({ type: "onSync" });
+        },
+        complete: () => {
+          dispatch({ type: "onSync" });
+        },
+      });
     });
```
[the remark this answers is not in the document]
gre-ledger: this is legit concern. I'll update the code to make the promise going into the RXJS piping.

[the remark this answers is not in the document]
gre-ledger: BTW, practically speaking, this is not an actual user case : we do not use useBridgeTransaction on an account that isn't supported. in such case, you would be stuck in the send flow before with an error displayed. in fact, in the previous code, it would have thrown an error from getAccountBridge, that error was unhandled as well, so we're consolidating code for the sake of good code, but not actual user path.

gre-ledger:
E2E
- Desktop E2E https://github.com/LedgerHQ/ledger-live/actions/runs/26390658019
- Mobile E2E https://github.com/LedgerHQ/ledger-live/actions/runs/26390662995

Builds
- [Android](https://github.com/LedgerHQ/ledger-live-build/actions/runs/26390742378)
- [iOS](https://github.com/LedgerHQ/ledger-live-build/actions/runs/26390750429)
- [Desktop](https://github.com/LedgerHQ/ledger-live-build/actions/runs/26390758947)

francois-guerin-ledger:
Result here: https://github.com/LedgerHQ/ledger-live/actions/runs/26391291427/job/77681376773
Mina not working is no blocker, issue with the nodes.

gre-ledger:
The feedback from Copilot created the opportunity to fix it directly against develop, here is a side PR preparing the async of useBridgeTransaction https://github.com/LedgerHQ/ledger-live/pull/17768

gre-ledger:
Rebased (post RN upgrade merge). New Builds.

Android https://github.com/LedgerHQ/ledger-live-build/actions/runs/26404649927
iOS https://github.com/LedgerHQ/ledger-live-build/actions/runs/26404652065
Desktop https://github.com/LedgerHQ/ledger-live-build/actions/runs/26404657141

libs/ledger-live-common/src/coin-modules/registry.ts
```diff
@@ -22,6 +10,19 @@ function getLoader(family: string): CoinModuleLoader {
   return loader;
 }
 
+export function makeLoaderCache<T>(fn: (family: string) => Promise<T>): (family: string) => Promise<T>;
+export function makeLoaderCache<T>(fn: (family: string) => Promise<T> | undefined): (family: string) => Promise<T> | undefined;
+export function makeLoaderCache<T>(fn: (family: string) => Promise<T> | undefined) {
```
ysitbon, line 15: If import("@ledgerhq/coin-xxx/...") rejects (race during HMR, or later if we put them on a CDN, we could experience a network blip), the rejection is cached forever and becomes permanently broken for the session. It could be worth implementing a retry strategy, wdyt?

libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -57,95 +56,105 @@ export const getCurrencyBridge = (currency: CryptoCurrency): CurrencyBridge => {
...
+  const mainAccount = getMainAccount(account, parentAccount);
+  const { currency } = mainAccount;
+  const supportedError = checkAccountSupported(mainAccount);
+
+  if (supportedError) {
+    return Promise.reject(supportedError);
   }
-  return accountBridgeCache[family];
+
+  return getAccountBridgeByFamily(currency.family, mainAccount.id);
```
ysitbon: We might have an invisible issue here. The function works fine for the success path, but not for the failure path.  `useAccountBridge` calls `getAccountBridge` directly in render with no `useMemo`:
- The success path returns a cached Promise, and `use()` reads the annotated `status='fulfilled'` and returns synchronously.
- The failure path every render:
  - → new Promise.reject(...) 
  - → no status annotation initially 
  - → `use()` suspends 
  - → React waits for the rejection
  - → microtask sets `status='rejected'` on that Promise
  - → React retries 
  - → hook calls `getAccountBridge` again 
  - → new Promise, no status 
  - → suspend again. Loop.

Maybe cache the rejection too, so identity is stable in both branches?

gre-ledger:
- [ ] to fix: wallet-cli loaders must be updated to import() too. they still use require() which returns any and hides possible problem that the loaders are supposed to return a promise.

libs/ledger-live-common/src/bridge/impl.ts
gre-ledger: mmh I think getAccountBridge is not supposed to fail. If it does, some screens will not be functional because the bridge is needed to perform many things (one example: to know if the account is empty we ask `bridge.isAccountEmpty`). 
This is why we want to filter out unsupported accounts (which is the only handled case of this function to fail) ( PR https://github.com/LedgerHQ/ledger-live/pull/17576 )

We could still try to fix the infinite loop possibility.

apps/ledger-live-mobile/src/families/canton/StepRecipientCustomAlert.test.tsx
```diff
@@ -95,6 +97,8 @@ describe("StepRecipientCustomAlert", () => {
     status: createMockTransactionStatus(),
   };
 
+  beforeAll(() => getCurrencyBridge(defaultProps.account.currency as CryptoCurrency));
```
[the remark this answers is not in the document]
gre-ledger: both code are semantically the same. I don't get why this is getting suggested.
`beforeAll(()=>foo())` will make the foo promise returned to the beforeAll callback. same as with the async await syntax (modulo it's a Promise<void> and not a Promise a currency)

apps/ledger-live-desktop/src/renderer/families/aleo/__tests__/operationsList.test.tsx
```diff
@@ -4,9 +4,12 @@ import { render, screen } from "tests/testSetup";
 import { ALEO_ACCOUNT_1 } from "../__mocks__/account.mock";
 import { OperationsList } from "~/renderer/components/OperationsList";
 import type { AleoOperation } from "@ledgerhq/live-common/families/aleo/types";
+import { getAccountBridgeByFamily } from "@ledgerhq/live-common/bridge/impl";
 
 const mockT = jest.fn() as unknown as TFunction;
 
+beforeAll(() => getAccountBridgeByFamily("aleo"));
```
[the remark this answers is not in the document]
gre-ledger: both code are semantically the same. I don't get why this is getting suggested.
beforeAll(()=>foo()) will make the foo promise returned to the beforeAll callback. same as with the async await syntax (modulo it's a Promise and not a Promise a currency)

apps/ledger-live-mobile/src/families/canton/PendingTransferProposals/__tests__/index.test.tsx
```diff
@@ -10,6 +11,8 @@ import { ACCOUNT_XPUB, createCantonAccount, createRawProposal } from "./test-uti
 const unit: Unit = { code: "CANTON", magnitude: 38, name: "Canton" };
 const mockAccount = createCantonAccount();
 
+beforeAll(() => getCurrencyBridge(mockAccount.currency as CryptoCurrency));
```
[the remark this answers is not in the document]
gre-ledger: both code are semantically the same. I don't get why this is getting suggested.
beforeAll(()=>foo()) will make the foo promise returned to the beforeAll callback. same as with the async await syntax (modulo it's a Promise and not a Promise a currency)

libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -57,95 +56,105 @@ export const getCurrencyBridge = (currency: CryptoCurrency): CurrencyBridge => {
...
-): ResolvedAccountBridge<any> => {
-  const mainAccount = getMainAccount(account, parentAccount);
-  const { currency } = mainAccount;
-  const supportedError = checkAccountSupported(mainAccount);
+export const getCurrencyBridge = (currency: CryptoCurrency): Promise<CurrencyBridge> => {
+  if (!currencyBridgePromiseCache[currency.family]) {
+    currencyBridgePromiseCache[currency.family] = settleAnnotate(buildCurrencyBridge(currency));
+  }
+  return currencyBridgePromiseCache[currency.family];
+};
```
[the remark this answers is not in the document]
gre-ledger: connexe to Yoann feedback. on it.

libs/ledger-live-common/src/bridge/impl.ts
gre-ledger: I've fixed it in latest commit.

libs/ledger-live-common/src/coin-modules/registry.ts
gre-ledger: fixed.

gre-ledger:
rebased + history of 10 commits polishes squashed into one.

libs/ledger-live-common/src/coin-modules/types.ts
```diff
@@ -22,27 +23,41 @@ export type MessageSignerModule = {
   prepareMessageToSign?: (opts: { account: Account; message: string }) => AnyMessage;
 };
 
-/* eslint-disable @typescript-eslint/no-explicit-any */
-// Parameters are `any` because each family uses its own transaction/raw types at the loader boundary.
-// Method syntax enables bivariant param checking so family-specific implementations are assignable.
 export type TransactionModule = {
   fromTransactionRaw(raw: any): TransactionCommon;
```
gre-ledger: self note: i'm looking at making these `any` go away too!

OlivierFreyssinet (dismissed):

libs/ledger-live-common/src/coin-modules/loaders.ts
```diff
@@ -1,235 +1,291 @@
...
   },
   {
     family: "casper",
-    loadSetup: () => require("../families/casper/setup"),
-    loadTransaction: () => require("@ledgerhq/coin-casper/transaction").default,
-    loadDeviceTxConfig: () => require("@ledgerhq/coin-casper/deviceTransactionConfig").default,
-    loadMockBridge: () => require("../families/casper/bridge/mock").default,
+    loadSetup: () => import("../families/casper/setup.js") as object as Promise<FamilySetup>,
+    loadTransaction: () => import("@ledgerhq/coin-casper/transaction").then(m => m.default),
+    loadDeviceTxConfig: () =>
```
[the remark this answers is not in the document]
gre-ledger: https://ledgerhq.atlassian.net/browse/LIVE-31427

gre-ledger:
Copilot is full of mystery. always have more to say. 😆  BTW CI very flaky today. this morning, everything was passing...

gre-ledger:
E2E Desktop https://github.com/LedgerHQ/ledger-live/actions/runs/26458289661
E2E Mobile https://github.com/LedgerHQ/ledger-live/actions/runs/26458294835
Build Android https://github.com/LedgerHQ/ledger-live-build/actions/runs/26458264357
Build iOS https://github.com/LedgerHQ/ledger-live-build/actions/runs/26458267212
Build Desktop https://github.com/LedgerHQ/ledger-live-build/actions/runs/26458270596

libs/ledger-live-common/src/transaction/deviceTransactionConfig.ts
```diff
@@ -28,5 +28,5 @@ export async function getDeviceTransactionConfig(arg: {
   const family = mainAccount.currency.family;
   const f = await loadDeviceTxConfigForFamily(family);
   if (!f) return [];
-  return await f(arg);
+  return f(arg) as Promise<Array<DeviceTransactionField>>;
```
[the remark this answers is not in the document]
gre-ledger: it is necessary.

gre-ledger:
rebuild post rebase

https://github.com/LedgerHQ/ledger-live-build/actions/runs/26559056153
https://github.com/LedgerHQ/ledger-live-build/actions/runs/26559058331
https://github.com/LedgerHQ/ledger-live-build/actions/runs/26559061312

gre-ledger:
> [33.3% Duplication on New Code](https://sonarcloud.io/component_measures?id=LedgerHQ_ledger-live&pullRequest=17760&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)

Totally expected. unfortunately can't be flagged on sonarcloud side but this is finding problem on the config declaration of live-common-setup, yes it's duplicated by design. each app is master of its config setup.

gre-ledger:
retriggering E2E

https://github.com/LedgerHQ/ledger-live/actions/runs/26578170757
https://github.com/LedgerHQ/ledger-live/actions/runs/26578177118

mcayuelas-ledger (dismissed):
OK for WXP scope

ysitbon (dismissed):

OlivierFreyssinet (dismissed):

gre-ledger:
- minor flaky case on desktop e2e: `Swap - OKX flow` fails
- minor flaky case on mobile e2e (pass on 1/2, so likely unrelated as our changes are not platform specific) and a broken case of `Vote on CELO with moonli.me` due to delegator list change, but unrelated to this PR.

libs/ledger-live-common/src/coin-modules/registry.ts
```diff
@@ -30,65 +39,75 @@ export function getRegisteredFamilies(): string[] {
...
+export const loadTransactionForFamily = makeLoaderCache((family) =>
+  getLoader(family).loadTransaction()
+);
 
-export const loadDeviceTxConfigForFamily = (
-  family: string,
-): DeviceTransactionConfigFn | undefined => loaders.get(family)?.loadDeviceTxConfig?.();
+export const loadDeviceTxConfigForFamily = makeLoaderCache((family) =>
+  loaders.get(family)?.loadDeviceTxConfig?.()
+);
```
Justkant, line 52: Should we align the code ?
Sometimes we don't use getLoader and other times we do

libs/ledger-live-common/src/tsconfig.build.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "customConditions": [],
+    "module": "CommonJS",
```
Justkant, line 5: Do we really need that because I thought we passed the module type in the command lines when we build for esm vs cjs, no ?

apps/wallet-cli/src/live-common-setup.ts
```diff
@@ -93,9 +99,10 @@ setSupportedCurrencies([...WALLET_CLI_SUPPORTED_CRYPTO_CURRENCY_IDS]);
 // bridge code is reached through ESM imports).
 LiveConfig.setConfig(walletCliConfig);
 // Also set on the CJS singleton — Bun's bundler resolves ESM imports to lib-es/
-// and require() to lib/, creating separate LiveConfig.instance singletons.
-// Non-coin-framework families (solana, bitcoin) load their bridge via require() in
-// the lazy loaders above, so they read from the CJS instance.
+// and require() to lib/, creating separate LiveConfig.instance singletons. Any
+// downstream code path that still hits a CJS-resolved require() reads from this
+// instance.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
```
gre-ledger: undo code diff. we will solve in future PR that we may be able to remove the require hack.

libs/ledger-live-common/src/coin-modules/registry.ts
gre-ledger: for all the load that are `T | undefined` we prefer to propagate the falseness and not have `throw` that happens too often. this pattern is used when the load returns undefined possibly, loadSetup* is on all coins. i wouldn't reconsider this choice for now, it's more part of what is already on develop btw (than this iteration)

libs/ledger-live-common/src/tsconfig.build.json
gre-ledger: ok, so here is the situation : the main `tsconfig.json` had to force the esm mode for typechecker to use it, because without it, it would typescript error

example
```ts
  src/coin-modules/loaders.ts:15:28 - error TS2322:
  Type 'Promise<TransactionModule<…> | typeof import(".../coin-algorand/lib-es/transaction")>'
  is not assignable to type 'Promise<TransactionModule<…>>'.

  15  loadTransaction: () => import("@ledgerhq/coin-algorand/transaction").then(m => m.default),
  ```

but then `tsconfig.build.json` which inherits from `tsconfig.json` would start building for ESM, because currently in our build script we have this:

```
  await $`pnpm tsc --project src/tsconfig.build.json`;                                          // → lib/   (CJS)
  await $`pnpm tsc --project src/tsconfig.build.json -m esnext --moduleResolution bundler --outDir lib-es`; // → lib-es/ (ESM)
  ```

so for now, we make it explicit that the build by default is still on CJS.

as for the study weither or not we can go ESM-only, we will study on this task: https://ledgerhq.atlassian.net/browse/LIVE-29648

OlivierFreyssinet (approved):

LucasWerey (approved):

jiyuzhuang (approved):

Justkant (approved):

ysitbon (approved):

PR #17770, opened 2026-05-25
[LWDM] feat(coin-bitcoin): [Zcash] Coin selection + prepareTransaction

cted-ledger:
### ✅ Checklist

  - Zcash shielded `prepareTransaction` / `getTransactionStatus` / `estimateMaxSpendable` are now functional for shielded transfer types
  - Shielded balance computation migrated from incremental-delta to direct sum of unspent notes (fixes self-send balance inflation)
  - Cross-scan spent detection via `knownNullifiers` / `spentKnownNullifiers` pipeline (fixes incremental sync balance bug)
  - No impact on transparent (Bitcoin legacy) flows -- adapter methods return `undefined` for transparent transfers

### 📝 Description

Implements the Zcash shielded transaction preparation pipeline in `coin-bitcoin`:

**Coin selection** (`coin-selection.ts` -- new module):
- Largest-first note selection with iterative ZIP-317 fee convergence
- `computeZip317Fee`, `computeLogicalActions`, `selectNotes`, `estimateMaxSpendableAmount`
- Exported `ZIP317_MINIMUM_FEE` constant for reuse

**Chain adapter methods** (`index.ts`):
- `prepareTransaction`: selects unspent Orchard notes, computes fee and change
- `getTransactionStatus`: validates balance, amount, selected notes
- `estimateMaxSpendable`: total unspent minus fee for spending all notes

**Balance migration** (`operations.ts`, `sync.ts`):
- `computeBalanceFromNotes`: direct sum of unspent notes (isSpent !== true && incoming/internal)
- Automatic fallback to incremental-delta for pre-upgrade data without `isSpent` fields
- `spentKnownNullifiers` from Rust marks previously-stored notes as spent during incremental sync

**Type extensions** (`types.ts`):
- `DecryptedOutput` / `DecryptedOutputRaw`: nullifier, rho, rseed, cmx, position (string), recipient, isSpent
- `SpendableNote`: selected note type for coin selection results (includes rho for `Note::from_parts`)
- `ZcashTransaction`: selectedNotes, zcashFee, changeAmount
- `SyncShieldedArgs` / `ShieldedSyncResult`: knownNullifiers / spentKnownNullifiers

**Full pipeline** (`engine.ts` -> `rehydrate.ts` -> `serialization.ts` -> `ZCash.ts` -> `ZCashIPC.ts` -> `contract.ts` -> `utility-entry.ts`):
- Enriched note fields (including `rho`) flow from Rust native engine through IPC to the JS balance computation
- `position` is a decimal string (not f64) to avoid precision loss on u64 round-trips
- `knownNullifiers` collected from existing unspent notes and passed to Rust for cross-scan spent detection
- `is_spent` (snake_case) -> `isSpent` (camelCase) conversion at the rehydration boundary

**Requires** `@ledgerhq/zcash-utils` >= 0.3.0 (enriched `ShieldedNote` with rho, spending fields, `knownNullifiers` / `spentKnownNullifiers` on `SyncParams` / `SyncStats`, position as decimal string).

### ❓ Context

- **JIRA link**: [LIVE-26922](https://ledgerhq.atlassian.net/browse/LIVE-26922)
- **Depends on**: [LIVE-26917](https://ledgerhq.atlassian.net/browse/LIVE-26917) (zcash-utils enriched sync)
- **Epic**: [LIVE-31212](https://ledgerhq.atlassian.net/browse/LIVE-31212) ([ZCASH][LWD] Shielded :: Private Send flows (COIN MODULE))

[LIVE-26922]: https://ledgerhq.atlassian.net/browse/LIVE-26922
Fixes LIVE-31251


libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/operations.ts
```diff
@@ -64,6 +64,72 @@ export function computeOutgoingFees(transactions: ShieldedTransaction[]): BigNum
...
+): { orchardBalance: BigNumber; saplingBalance: BigNumber } | undefined {
+  const allOrchardNotes = transactions.flatMap(tx => tx.decryptedData?.orchard_outputs ?? []);
+  const allSaplingNotes = transactions.flatMap(tx => tx.decryptedData?.sapling_outputs ?? []);
+
+  // Check if ANY note has the isSpent field — if none do, this is pre-upgrade data
+  const hasEnrichedNotes =
+    allOrchardNotes.some(n => n.isSpent !== undefined) ||
+    allSaplingNotes.some(n => n.isSpent !== undefined);
+
+  if (!hasEnrichedNotes) return undefined; // Caller falls back to delta approach
```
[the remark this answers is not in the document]
cted-ledger: Intentional — added a comment explaining the behavior. Legacy notes with isSpent === undefined pass the !== true filter and are counted as unspent. This is conservative: we'd rather overcount than undercount the balance.  
After a full re-sync with zcash-utils ≥ 0.3.0, all notes will have isSpent and the mixed case disappears. The delta fallback only fires when zero notes have isSpent (pure pre-upgrade state).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,93 @@ const zcashChainAdapter: ChainAdapter = {
...
+    if (tx.amount.lte(0)) {
+      errors.amount = new Error("Amount must be positive");
+    }
+    if (totalSpent.gt(orchardBalance)) {
+      errors.amount = new Error("Insufficient shielded balance");
+    }
+    if (!tx.selectedNotes) {
+      errors.amount = new Error("Transaction not prepared — call prepareTransaction first");
+    } else if (tx.selectedNotes.length === 0) {
+      errors.amount = new Error("No spendable notes available");
```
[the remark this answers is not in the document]
cted-ledger: Fixed — restructured as an if / else if / else if chain so only the first matching condition sets errors.amount. Priority: amount ≤ 0 → no selected notes → totalSpent > balance.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,93 @@ const zcashChainAdapter: ChainAdapter = {
...
-    // TODO: implement PCZT transaction preparation (ZIP-317 fee info)
-    return undefined;
+  prepareTransaction(account: Account, transaction: Transaction) {
+    const zcashAccount = account as ZcashAccount;
+    const tx = transaction as ZcashTransaction;
+    if (!isShieldedTransfer(tx)) return undefined; // Bitcoin legacy path
+
+    const notes = collectSpendableNotes(zcashAccount.privateInfo?.transactions ?? []);
+    const result = selectNotes(notes, tx.amount, tx.transferType);
+    if (!result) return Promise.resolve(tx); // Insufficient — status will report the error
```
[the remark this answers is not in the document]
cted-ledger: Fixed — prepareTransaction now returns selectedNotes: [] (empty array) when coin selection fails, instead of omitting the field. This makes the "prepared but insufficient" case distinguishable from "prepare never called (selectedNotes: undefined).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/sync.ts
```diff
@@ -63,7 +67,15 @@ export const zcashSyncShielded = (
...
     }
-    const { lastProcessedBlock, birthday } = acc.initialAccount?.privateInfo ?? {};
+    const { lastProcessedBlock, birthday, transactions } = acc.initialAccount?.privateInfo ?? {};
+
+    // Collect nullifiers of unspent notes from previous syncs so the engine
+    // can detect when previously-received notes are spent in this scan range.
+    const knownNullifiers: string[] = (transactions ?? []).flatMap(tx =>
+      (tx.decryptedData?.orchard_outputs ?? [])
+        .filter(n => n.isSpent !== true && n.nullifier != null)
+        .map(n => n.nullifier!),
```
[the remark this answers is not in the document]
cted-ledger: Fixed — wrapped in [...new Set(...)] to deduplicate before passing to the native engine.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (transferType === "shielded-to-transparent") {
+    // Transparent output is NOT an Orchard action.
+    // Orchard bundle: N spends + at most 1 change output (Orchard).
+    // nActions = max(spends, orchard_outputs) where orchard_outputs ∈ {0, 1}.
+    // Since spendCount ≥ 1, max(spendCount, 1) = spendCount. Equivalent but explicit.
+    return Math.max(spendCount, 1);
+  }
+  // transparent-to-shielded / transparent: no Orchard actions
+  return 0;
+}
```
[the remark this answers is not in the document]
cted-ledger: By design — prepareTransaction returns undefined for transparent source transfers, so computeLogicalActions is never called with transparent-to-shielded. The return 0 is defensive. Added a comment clarifying this.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
```diff
@@ -1,5 +1,25 @@
...
+    memo: raw.memo,
+    transfer_type: raw.transfer_type,
+    amount: new BigNumber(raw.amount),
+    ...(raw.nullifier != null && { nullifier: raw.nullifier }),
+    ...(raw.rseed != null && { rseed: raw.rseed }),
+    ...(raw.cmx != null && { cmx: raw.cmx }),
+    ...(raw.position != null && { position: raw.position }),
+    ...(raw.recipient != null && { recipient: raw.recipient }),
+    ...(raw.is_spent != null && { isSpent: raw.is_spent }),
+  };
```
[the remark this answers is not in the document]
cted-ledger: Fixed — removed the duplicate from serialization.ts and imported rehydrateOutput (now exported) from serialization/rehydrate.ts. Single source of truth.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,114 @@
+import { BigNumber } from "bignumber.js";
+import type { SpendableNote, ZcashTransferType } from "./types";
+
+// ZIP-317 constants
+const ZIP317_MARGINAL_FEE = 5_000; // per logical action
```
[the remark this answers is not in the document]
cted-ledger: Added — .changeset/sour-windows-shake.md included in the commit (@ledgerhq/coin-bitcoin, minor).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/native-engine/engine.ts
```diff
@@ -131,14 +134,15 @@ export async function startSyncJob(
...
-    const { blocksScanned, transactions } = await syncChunk(
+    const { blocksScanned, transactions, spentKnownNullifiers } = await syncChunk(
       native,
       {
         grpcUrl,
         network,
         viewingKey,
         chunkStart,
         chunkEnd,
+        ...(knownNullifiers && knownNullifiers.length > 0 && { knownNullifiers }),
```
semeano: `...(knownNullifiers?.length && { knownNullifiers }),`

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,94 @@ const zcashChainAdapter: ChainAdapter = {
...
+        warnings,
+        estimatedFees: new BigNumber(0),
+        amount: tx.amount,
+        totalSpent: tx.amount,
+        ...bitcoinExtras,
+      } satisfies TransactionStatus);
+    }
+
+    const orchardBalance = privateInfo.orchardBalance;
+    const fee = tx.zcashFee ?? new BigNumber(10_000); // default 2-action fee
```
semeano: `new BigNumber(10_000)` can you set this as a constant so it can be easily edited in the future?

cted-ledger: Done. Exported ZIP317_MINIMUM_FEE from coin-selection.ts (= ZIP317_GRACE_ACTIONS * ZIP317_MARGINAL_FEE = 10,000) and used it in getTransactionStatus.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/native-engine/engine.ts
cted-ledger: Done, simplified to ...(knownNullifiers?.length && { knownNullifiers })

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/operations.ts
```diff
@@ -64,6 +64,79 @@ export function computeOutgoingFees(transactions: ShieldedTransaction[]): BigNum
...
+/**
+ * Compute Orchard and Sapling balances by summing unspent notes.
+ * This is the authoritative balance method when enriched notes (with isSpent) are available.
+ * Falls back to undefined if no notes have the isSpent field (pre-upgrade data).
+ */
+export function computeBalanceFromNotes(
+  transactions: ShieldedTransaction[],
+): { orchardBalance: BigNumber; saplingBalance: BigNumber } | undefined {
+  const allOrchardNotes = transactions.flatMap(tx => tx.decryptedData?.orchard_outputs ?? []);
+  const allSaplingNotes = transactions.flatMap(tx => tx.decryptedData?.sapling_outputs ?? []);
```
may01: Handling sapling is misleading. Especially when we perform any kind of logic instead of stubs.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/operations.ts
```diff
@@ -64,6 +64,79 @@ export function computeOutgoingFees(transactions: ShieldedTransaction[]): BigNum
...
+
+  // Check if ANY note has the isSpent field — if none do, this is pre-upgrade data.
+  // For mixed datasets (some enriched, some legacy), legacy notes with
+  // isSpent === undefined pass the `!== true` filter and are included in the sum.
+  // This is intentionally conservative: we count notes whose spent status is
+  // unknown rather than excluding them (which would undercount the balance).
+  // After a full re-sync with zcash-utils >= 0.3.0, all notes will have isSpent.
```
may01: We handle legacy notes that are not available on production. I think it will be better to force resync in dev environment, and avoid logic for temporary decision.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/sync.ts
```diff
@@ -136,12 +153,37 @@ export function reduceShieldedSyncResult(
...
+
   const prevSapling = accumulated.accountUpdate.privateInfo?.saplingBalance ?? new BigNumber(0);
   const prevOrchard = accumulated.accountUpdate.privateInfo?.orchardBalance ?? new BigNumber(0);
 
-  const { deltaSapling, deltaOrchard } = computeProtocolDeltas(newTransactions);
-  const saplingBalance = prevSapling.plus(deltaSapling);
-  const orchardBalance = prevOrchard.plus(deltaOrchard);
+  // Prefer direct-sum from enriched notes with isSpent (authoritative).
+  // Falls back to incremental delta for pre-upgrade data.
+  const noteBalance = computeBalanceFromNotes(allShieldedTx);
```
may01: Should force resync. Without handling legacy approach

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    if (newFee.eq(fee)) {
+      // Fee converged
+      return { selectedNotes: selected, fee, changeAmount, totalInput };
+    }
+    fee = newFee; // Retry with updated fee
+  }
+
+  // Should not happen — fee is monotonically bounded, but defensive
+  return undefined;
```
may01, line 113: We should log unexpected behaviour if it suddenly will occure.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,118 @@
...
+      if (totalInput.gte(target)) break;
+      selected.push(note);
+      totalInput = totalInput.plus(note.amount);
+    }
+
+    if (totalInput.lt(target)) return undefined; // Insufficient balance
+
+    const changeAmount = totalInput.minus(amount).minus(fee);
+    const hasChange = changeAmount.gt(0);
+    const outputCount = hasChange ? 2 : 1; // recipient + optional change
```
may01, line 96: Claude:

> Producing sub-marginal_fee (5_000 zat) change is economically irrational (fee exceeds returned value) and creates near-unspendable dust notes for next time. Most Zcash wallets absorb dust into the fee.
> Action: introduce DUST_THRESHOLD = ZIP317_MARGINAL_FEE and if changeAmount < DUST_THRESHOLD, drop the change output and roll it into the fee.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,94 @@ const zcashChainAdapter: ChainAdapter = {
...
+    }
+
+    const orchardBalance = privateInfo.orchardBalance;
+    const fee = tx.zcashFee ?? new BigNumber(ZIP317_MINIMUM_FEE);
+    const totalSpent = tx.amount.plus(fee);
+
+    if (tx.amount.lte(0)) {
+      errors.amount = new Error("Amount must be positive");
+    } else if (!tx.selectedNotes || tx.selectedNotes.length === 0) {
+      // No notes selected — either insufficient balance or no spendable notes
```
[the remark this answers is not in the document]
cted-ledger: Fixed. Removed the misleading comment. prepareTransaction returns selectedNotes: [] when coin selection fails. Both cases (undefined and empty) now report the same error since this distinction added no value.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,94 @@ const zcashChainAdapter: ChainAdapter = {
...
     _parentAccount: Account | null | undefined,
-    _transaction: Transaction | null | undefined,
+    transaction: Transaction | null | undefined,
   ) {
-    // TODO: implement PCZT balance estimation
-    return undefined;
+    const zcashAccount = account as ZcashAccount;
+    const tx = transaction as ZcashTransaction | null | undefined;
+    const transferType = tx?.transferType ?? "shielded";
+    if (transferType === "transparent") return undefined;
```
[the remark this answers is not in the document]
cted-ledger: Fixed (same change as above). estimateMaxSpendable now returns undefined for both transparent and transparent-to-shielded.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,94 @@ const zcashChainAdapter: ChainAdapter = {
     return undefined;
   },
 
-  getTransactionStatus(_account: Account, _transaction: Transaction) {
-    // TODO: implement PCZT transaction validation + ZIP-317 fee estimation
-    return undefined;
+  getTransactionStatus(account: Account, transaction: Transaction) {
+    const zcashAccount = account as ZcashAccount;
+    const tx = transaction as ZcashTransaction;
+    if (!isShieldedTransfer(tx)) return undefined;
```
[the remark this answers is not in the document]
cted-ledger: Fixed. All three adapter methods now gate on transferType === "shielded" || "shielded-to-transparent" instead of isShieldedTransfer.
transparent-to-shielded and transparent both return undefined (Bitcoin legacy path), since they use transparent UTXOs, not Orchard notes.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (transferType === "shielded-to-transparent") {
+    // Transparent output is NOT an Orchard action.
+    // Orchard bundle: N spends + at most 1 change output (Orchard).
+    // nActions = max(spends, orchard_outputs) where orchard_outputs ∈ {0, 1}.
+    // Since spendCount ≥ 1, max(spendCount, 1) = spendCount. Equivalent but explicit.
+    return Math.max(spendCount, 1);
+  }
+  // transparent-to-shielded / transparent: fee is handled by the Bitcoin
+  // legacy path (prepareTransaction returns undefined for these types).
+  // Returning 0 is defensive — callers never reach here for transparent sources.
```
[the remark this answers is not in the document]
cted-ledger: The comment is now accurate: the adapter methods return undefined for transparent-to-shielded, so computeLogicalActions is indeed never called with that transfer type. The return 0 is purely defensive.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/__tests__/adapter.test.ts
```diff
@@ -55,38 +117,235 @@ describe("zcash chain adapter — transaction routing", () => {
...
+      // selectNotes returns undefined → prepareTransaction sets selectedNotes: []
+      expect(result.selectedNotes).toEqual([]);
+      expect(result.transferType).toBe("shielded");
+    });
+
+    it("returns a Promise for shielded-to-transparent (no spendable notes → original tx)", async () => {
+      // transparent-to-shielded: isShieldedTransfer = true but no orchard notes to spend
+      const account = makeZcashAccount({ transactions: [] });
+      const tx = makeTx("transparent-to-shielded", new BigNumber(1000));
+      const resultPromise = adapter.prepareTransaction!(account, tx);
```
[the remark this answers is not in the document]
cted-ledger: Fixed. Split into two tests: "returns undefined for transparent-to-shielded" (verifies legacy path) and "returns a Promise for shielded-to-transparent" (verifies coin selection with no notes).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/operations.ts
cted-ledger: Agreed. Removed all Sapling logic from computeBalanceFromNotes. The function now returns a single BigNumber (Orchard balance only). saplingBalance in privateInfo stays as-is from the accumulated state.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/operations.ts
cted-ledger: Removed the "resync-needed" path entirely.
Notes without isSpent are treated as unspent (conservative).
No legacy data exists in production (the feature hasn't shipped), so the complexity of a forced resync is not justified.
Dev and QA will need to reset their accounts...

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/sync.ts
cted-ledger: Done. The delta fallback and resync path are both removed. 

computeBalanceFromNotes returns BigNumber directly, no union type, no branching.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
cted-ledger: Added console.warn with amount, note count, and last fee value when selectNotes fails to converge after MAX_ITERATIONS.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
cted-ledger: Added DUST_THRESHOLD = ZIP317_MARGINAL_FEE (5,000 zats).
When changeAmount > 0 && changeAmount < DUST_THRESHOLD, the change is absorbed into the fee and changeAmount set to 0.
Avoids creating near-unspendable dust notes.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -1,14 +1,18 @@
...
 import type { ChainAdapter } from "../types";
 import type { BitcoinAddress, BitcoinXPub, SignerContext } from "../../signer";
-import type { Transaction } from "../../types";
+import type { Transaction, TransactionStatus } from "../../types";
 import { DmkSignerZcash } from "@ledgerhq/live-signer-zcash";
 import type { ZcashAddress, ZcashViewKey } from "@ledgerhq/live-signer-zcash";
 import { registerChainAdapter } from "../registry";
-import type { ZcashAccount, ZcashAccountRaw } from "./types";
+import type { ZcashAccount, ZcashAccountRaw, ZcashTransaction } from "./types";
+import { isShieldedTransfer } from "./types";
```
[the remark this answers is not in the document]
cted-ledger: fixed

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +78,95 @@ const zcashChainAdapter: ChainAdapter = {
...
+    const totalSpent = tx.amount.plus(fee);
+
+    if (tx.amount.lte(0)) {
+      errors.amount = new Error("Amount must be positive");
+    } else if (!tx.selectedNotes || tx.selectedNotes.length === 0) {
+      // No notes selected — either insufficient balance or no spendable notes
+      errors.amount = new Error("Insufficient shielded balance");
+    } else if (totalSpent.gt(orchardBalance)) {
+      errors.amount = new Error("Insufficient shielded balance");
+    }
```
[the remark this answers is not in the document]
cted-ledger: Fixed. prepareTransaction now checks tx.useAllAmount: when true, it computes the effective amount via estimateMaxSpendableAmount (total unspent - fee for spending all notes) and sets it on the returned transaction. getTransactionStatus skips the amount <= 0 check when useAllAmount is set. Added a test verifying the full flow: useAllAmount: true with amount: 0 produces a transaction with amount > 0, selected notes, and changeAmount = 0.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/sync.ts
```diff
@@ -63,7 +66,19 @@ export const zcashSyncShielded = (
...
-    const { lastProcessedBlock, birthday } = acc.initialAccount?.privateInfo ?? {};
+    const { lastProcessedBlock, birthday, transactions } = acc.initialAccount?.privateInfo ?? {};
+
+    // Collect nullifiers of unspent notes from previous syncs so the engine
+    // can detect when previously-received notes are spent in this scan range.
+    const knownNullifiers: string[] = [
+      ...new Set(
+        (transactions ?? []).flatMap(tx =>
+          (tx.decryptedData?.orchard_outputs ?? [])
+            .filter(n => n.isSpent !== true && n.nullifier !== undefined)
```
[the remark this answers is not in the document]
cted-ledger: Fixed. Added transfer_type === "incoming" || "internal" filter to knownNullifiers collection. Outgoing notes are not our spendable notes -- their nullifiers don't need cross-scan spent detection and would only bloat the payload.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +77,110 @@ const zcashChainAdapter: ChainAdapter = {
...
   },
 
-  prepareTransaction(_account: Account, _transaction: Transaction) {
-    // TODO: implement PCZT transaction preparation (ZIP-317 fee info)
-    return undefined;
+  prepareTransaction(account: Account, transaction: Transaction) {
+    const zcashAccount = account as ZcashAccount;
+    const tx = transaction as ZcashTransaction;
+    // Only handle flows with shielded inputs (Orchard note selection).
+    if (tx.transferType !== "shielded" && tx.transferType !== "shielded-to-transparent")
```
may01, line 170: How we will handle crafting of `transparent` and `transparent-> shielded `transaction? The chain adapter should substitute the current coin-bitcoin logic so the legacy implementation will be unreachable. 
As we discussed, it will be good to have transparent send logic to be handled in new approach along side with shielded. To be able not to support the legacy path.

cted-ledger: Agreed on the long-term direction -- the chain adapter should own all 4 flow types so the Bitcoin legacy path becomes unreachable for Zcash.                                                                                                                                                                                                                    
However, transparent send works in production today via the Bitcoin legacy path. If we handle transparent and transparent-to-shielded in the adapter now (before the V5 builder supports transparent inputs in CUSTOM-03b), we would break the existing transparent send.                                                                                                                                                         
                                                                                                                                                                                                                     
The return undefined is intentional to preserve backward compatibility: it lets the current Bitcoin path continue to handle transparent flows until the V5 builder is ready. The migration plan is:              
 1. CUSTOM-03b: V5 builder adds transparent input support in zcash-utils
 2. CUSTOM-06: signOperation orchestrates all 4 flows through the chain adapter                                                                                                                                   
 3. At that point the adapter handles everything and the legacy path is removed

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +77,110 @@ const zcashChainAdapter: ChainAdapter = {
...
+    const notes = collectSpendableNotes(zcashAccount.privateInfo?.transactions ?? []);
+
+    // When useAllAmount is set, compute the effective amount from max spendable
+    const effectiveAmount = tx.useAllAmount
+      ? estimateMaxSpendableAmount(notes, tx.transferType)
+      : tx.amount;
+
+    const result = selectNotes(notes, effectiveAmount, tx.transferType);
+    if (!result) {
+      return Promise.resolve({ ...tx, selectedNotes: [] } as ZcashTransaction);
```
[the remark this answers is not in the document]
cted-ledger: Fixed. The failure branch now explicitly clears zcashFee: undefined and changeAmount: undefined, and sets amount: effectiveAmount. No stale values leak from a prior successful prepare.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/sync.ts
```diff
@@ -136,12 +154,23 @@ export function reduceShieldedSyncResult(
...
+  // Mark previously-stored notes as spent using Rust's cross-scan detection.
+  // spentKnownNullifiers contains nullifiers of notes received in prior scans
+  // that were observed as spent inputs in the current scan range.
+  const spentNfs = result.spentKnownNullifiers ?? [];
+  if (spentNfs.length > 0) {
+    const spentSet = new Set(spentNfs);
+    for (const tx of allShieldedTx) {
+      for (const note of tx.decryptedData?.orchard_outputs ?? []) {
+        if (note.nullifier && spentSet.has(note.nullifier)) {
+          note.isSpent = true;
```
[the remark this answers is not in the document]
cted-ledger: Fixed. Instead of mutating note.isSpent in-place, we now clone only the transactions that need mutation (using spread + map). Objects from accumulated.accountUpdate.privateInfo.transactions are never modified.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +77,110 @@ const zcashChainAdapter: ChainAdapter = {
...
+    if (tx.transferType === "shielded-to-transparent" && !tx.recipient) {
+      errors.recipient = new Error("Recipient address is required for shielded-to-transparent");
+    }
+
+    if (amount.lte(0) && !tx.useAllAmount) {
+      errors.amount = new Error("Amount must be positive");
+    } else if (!tx.selectedNotes || tx.selectedNotes.length === 0) {
+      errors.amount = new Error("Insufficient shielded balance");
+    } else if (totalSpent.gt(orchardBalance)) {
+      errors.amount = new Error("Insufficient shielded balance");
```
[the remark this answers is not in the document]
cted-ledger: Added a consistency check: when selectedNotes is present and passes the length check, we verify sum(selectedNotes.amount) >= totalSpent. If not, errors.amount is set to "Selected notes do not cover amount + fee". This catches corrupted or stale selectedNotes without impacting the normal flow (prepareTransaction always produces valid selections).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/sync.ts
```diff
@@ -136,12 +154,31 @@ export function reduceShieldedSyncResult(
...
   ];
 
-  const prevSapling = accumulated.accountUpdate.privateInfo?.saplingBalance ?? new BigNumber(0);
-  const prevOrchard = accumulated.accountUpdate.privateInfo?.orchardBalance ?? new BigNumber(0);
+  // Mark previously-stored notes as spent using Rust's cross-scan detection.
+  // Clone transactions that need mutation to preserve immutability of the
+  // accumulated state (allShieldedTx contains refs to existing objects).
+  const spentNfs = result.spentKnownNullifiers ?? [];
+  if (spentNfs.length > 0) {
+    const spentSet = new Set(spentNfs);
```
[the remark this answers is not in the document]
cted-ledger: Fixed. The early-return path (newTransactions.length === 0) now processes spentKnownNullifiers before returning: it clones and marks affected notes as spent, and recomputes orchardBalance when the spent set is non-empty. This handles the case where a chunk has no new transactions but the stream stats report spent nullifiers from cross-scan detection.

pnpm-lock.yaml
```diff
@@ -61,23 +61,23 @@ catalogs:
...
+      specifier: 0.1.12
+      version: 0.1.12
     '@ledgerhq/speculos-device-controller':
       specifier: 0.2.3
       version: 0.2.3
     '@ledgerhq/zcash-utils':
-      specifier: 0.2.0
-      version: 0.2.0
+      specifier: 0.3.0
+      version: 0.3.0
```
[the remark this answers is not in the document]
cted-ledger: regenerated

pnpm-lock.yaml
```diff
@@ -508,9 +508,9 @@ patchedDependencies:
   '@bunli/core@0.9.1':
     hash: fbe5c853111cb92eab35c575f445da139fd7664c9be9bbca5ebbadfb0f292565
     path: patches/@bunli__core@0.9.1.patch
-  '@callstack/repack@5.2.3':
+  '@callstack/repack@5.2.5':
     hash: 8d7db4e65c8982b51b689b97af043f6e2c411605b19084c3781cb5ec5e635b60
-    path: patches/@callstack__repack@5.2.3.patch
+    path: patches/@callstack__repack@5.2.5.patch
   '@changesets/get-github-info@0.6.0':
```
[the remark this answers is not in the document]
cted-ledger: regenerated

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +77,124 @@ const zcashChainAdapter: ChainAdapter = {
...
+    const orchardBalance = privateInfo.orchardBalance;
+    const fee = tx.zcashFee ?? new BigNumber(ZIP317_MINIMUM_FEE);
+    const amount = tx.amount;
+    const totalSpent = amount.plus(fee);
+
+    // Recipient validation for shielded-to-transparent (transparent address required).
+    // For shielded-to-shielded, recipient validation is deferred to the PCZT builder.
+    if (tx.transferType === "shielded-to-transparent" && !tx.recipient) {
+      errors.recipient = new Error("Recipient address is required for shielded-to-transparent");
+    }
```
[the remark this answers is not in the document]
cted-ledger: Valid point. Full transparent address validation (t-addr format check) requires reusing isValidRecipient from src/logic.ts or validateRecipient from src/cache.ts, which pulls in Bitcoin-specific validation logic. For now the empty check catches the most common user error (missing recipient). Full format validation will be added as part of CUSTOM-06 (signOperation orchestration) when the transparent address codec is wired into the chain adapter

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/ipc/utility-entry.ts
```diff
@@ -85,6 +85,7 @@ async function handleStartSync(port: ParentPort, args: StartSyncArgs): Promise<v
         viewingKey: args.viewingKey,
         startBlockHeight: args.startBlockHeight,
         maxBatchSize: args.maxBatchSize,
+        ...(args.knownNullifiers && { knownNullifiers: args.knownNullifiers }),
```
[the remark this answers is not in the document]
cted-ledger: Good catch. Aligned with the other call sites by adding the .length > 0 check

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/native-engine/engine.ts
```diff
@@ -152,7 +160,15 @@ export async function startSyncJob(
     processedBlocks += blocksScanned;
     for (const tx of transactions) {
       allTransactions.push(mapNativeTx(tx));
+      // Collect nullifiers from incoming/internal Orchard notes discovered
+      // in this chunk so the next chunk can detect them as spent.
+      for (const note of tx.orchardNotes ?? []) {
+        if (note.nullifier && note.transferType !== "outgoing") {
+          accumulatedNullifiers.push(note.nullifier);
+        }
+      }
```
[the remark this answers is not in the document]
cted-ledger: Fixed. Changed accumulatedNullifiers from string[] to Set<string>. .add() deduplicates automatically. Spread to array ([...accumulatedNullifiers]) only when passing to syncChunk.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -74,23 +77,124 @@ const zcashChainAdapter: ChainAdapter = {
...
+      return undefined;
+
+    const notes = collectSpendableNotes(zcashAccount.privateInfo?.transactions ?? []);
+
+    // When useAllAmount is set, compute the effective amount from max spendable
+    const effectiveAmount = tx.useAllAmount
+      ? estimateMaxSpendableAmount(notes, tx.transferType)
+      : tx.amount;
+
+    const result = selectNotes(notes, effectiveAmount, tx.transferType);
```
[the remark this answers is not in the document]
cted-ledger: Fixed. Added early return when effectiveAmount <= 0 (e.g. useAllAmount with no spendable notes). Returns selectedNotes: [] immediately without calling selectNotes, avoids a pointless "send 0" path.

semeano (dismissed):

may01 (dismissed):

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,131 @@
+import { BigNumber } from "bignumber.js";
+import type { SpendableNote, ZcashTransferType } from "./types";
+
+// ZIP-317 constants
+export const ZIP317_MARGINAL_FEE = 5_000; // per logical action
+export const ZIP317_GRACE_ACTIONS = 2;
+/** Minimum fee (grace actions * marginal fee). */
+export const ZIP317_MINIMUM_FEE = ZIP317_GRACE_ACTIONS * ZIP317_MARGINAL_FEE; // 10_000
+/** Change below this threshold is absorbed into the fee (avoids near-unspendable dust notes). */
+export const DUST_THRESHOLD = ZIP317_MARGINAL_FEE; // 5_000
```
qperrot: why having DUST_THRESHOLD that is equal to another const ?
Also this is exported but I don't see it anywhere else

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,131 @@
...
+  spendableNotes: SpendableNote[],
+  amount: BigNumber,
+  transferType: ZcashTransferType,
+): CoinSelectionResult | undefined {
+  // Sort largest-first to minimize action count
+  const sorted = [...spendableNotes].sort((a, b) => b.amount.comparedTo(a.amount));
+
+  // Iterative fee resolution (converges in <= 3 rounds for practical sizes)
+  let fee = computeZip317Fee(ZIP317_GRACE_ACTIONS); // initial estimate: 2-action minimum
+  const MAX_ITERATIONS = 5;
```
qperrot: can't we put this MAX_ITERATIONS  on top ?

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -0,0 +1,131 @@
...
+      changeAmount = new BigNumber(0);
+    }
+
+    const hasChange = changeAmount.gt(0);
+    const outputCount = hasChange ? 2 : 1; // recipient + optional change
+    const spendCount = selected.length;
+    const logicalActions = computeLogicalActions(spendCount, outputCount, transferType);
+    const newFee = computeZip317Fee(logicalActions);
+
+    if (newFee.eq(fee) || newFee.lt(fee)) {
```
qperrot: can't we have isLessThanOrEqualTo  or lte ?

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
```diff
@@ -16,16 +17,31 @@ export function toZcashPrivateInfoRaw(info: ZcashPrivateInfo): ZcashPrivateInfoR
...
-          tx.decryptedData?.orchard_outputs.map(output => ({
-            ...output,
-            amount: output.amount.toString(),
-          })) || [],
-        sapling_outputs:
-          tx.decryptedData?.sapling_outputs.map(output => ({
-            ...output,
-            amount: output.amount.toString(),
-          })) || [],
+        orchard_outputs: (tx.decryptedData?.orchard_outputs ?? []).map(output => ({
```
qperrot: can't we have something like
`function mapDecryptedOutput(output: DecryptedOutput): DecryptedOutputRaw {
  return {
    memo: output.memo,
    transfer_type: output.transfer_type,
    amount: output.amount.toString(),
    ...(output.nullifier !== undefined && { nullifier: output.nullifier }),
    ...(output.rho       !== undefined && { rho: output.rho }),
    ...(output.rseed     !== undefined && { rseed: output.rseed }),
    ...(output.cmx       !== undefined && { cmx: output.cmx }),
    ...(output.position  !== undefined && { position: output.position }),
    ...(output.recipient !== undefined && { recipient: output.recipient }),
    ...(output.isSpent   !== undefined && { is_spent: output.isSpent }),
  };
}`

then
`orchard_outputs: (tx.decryptedData?.orchard_outputs ?? []).map(mapDecryptedOutput),
sapling_outputs: (tx.decryptedData?.sapling_outputs ?? []).map(mapDecryptedOutput),`

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/types.ts
```diff
@@ -41,12 +43,41 @@ export type DecryptedOutputRaw = {
...
   transfer_type: string;
   amount: BigNumber; // zatoshis
+  // Spending fields
+  nullifier?: string;
+  rho?: string;
+  rseed?: string;
+  cmx?: string;
+  position?: string;
+  recipient?: string;
+  isSpent?: boolean;
```
qperrot, line 63: can we avoid repetion and have type that we can reuse something like 
```type SpendingFields = {
  nullifier?: string; // 64-char hex
  rho?: string;       // 64-char hex
  rseed?: string;     // 64-char hex
  cmx?: string;       // 64-char hex
  position?: string;  // decimal string (avoids f64 precision loss)
  recipient?: string; // 86-char hex
};```

```export type DecryptedOutputRaw = SpendingFields & {
  memo: string;
  transfer_type: string;
  amount: string; // zatoshis
  is_spent?: boolean;
};```

```export type DecryptedOutput = SpendingFields & {
  memo: string;
  transfer_type: string;
  amount: BigNumber; // zatoshis
  isSpent?: boolean;
};```|


```export type SpendableNote = Required<SpendingFields> & {
  txid: string;
  outputIndex: number;
  amount: BigNumber;
};```

cted-ledger: Good idea. Will extract SpendingFields and use Required<SpendingFields> for SpendableNote. Avoids repeating the 6 fields across 3 types.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
cted-ledger: Agreed. Will extract mapDecryptedOutput(output: DecryptedOutput): DecryptedOutputRaw and reuse for both orchard_outputs and sapling_outputs.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
cted-ledger: Right, newFee.lte(fee) is cleaner. Will fix.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
cted-ledger: Agreed, will move MAX_ITERATIONS to the constants block at the top.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
cted-ledger: The value is the same today but the semantics differ: DUST_THRESHOLD is the minimum economically rational change, ZIP317_MARGINAL_FEE is the cost per action. They could diverge in theory. That said, the export is unnecessary since it's only used locally. Will remove the export and add a comment explaining the semantic link.

qperrot (approved):

PR #17785, opened 2026-05-25
[LWDM] fix(live-common): evict makeLoaderCache entries on rejection

gre-ledger:
### 📝 Description

A failing dynamic `import()` (HMR race, CDN blip on a chunk-served coin module) used to poison the loader cache for the whole session. Treat rejection as a not-cached outcome so the next call retries. Addresses Yoann's review comment on #17760.

### ❓ Context

- **JIRA or GitHub link**: LIVE-29187 — follow-up to #17760


gre-ledger:
=> some issue in wallet-cli which didn't move to Promise based loaders. i have to fix it first.

PR #17807, opened 2026-05-26
[LWM] fix(mobile): apply nav-passed transaction on (re)mount

gre-ledger:
### ✅ Checklist

- [x] **Covered by automatic tests.** — manual repro via Tezos (re-)delegation; no existing test covers this nav<->useBridgeTransaction edge.
  - Tezos (re-)delegation: changing the baker now actually applies.
  - All other `useTransactionChangeFromNavigation` consumers (Send / EVM EditTransaction / Bitcoin EditTransaction) unaffected — they already plant `route.params.transaction` in the `useBridgeTransaction` init, so the first dispatch becomes a no-op (`if (state.transaction === action.transaction) return state` in the reducer).
  - QA focus: Send/Edit-fees flows on Tezos, EVM, Bitcoin (re-edit fees several times in a row, on cold and warm screens).

### 📝 Description

Bug reproduced on the parent PR branch (LIVE-29187): on Tezos delegation, clicking the pen to change the baker and selecting a new one returned you to the summary but with the original baker still set, and the bridge rejected the transaction with ``delegate.unchanged`` (mapped to ``InvalidAddressBecauseAlreadyDelegated``).

Root cause is in ``useTransactionChangeFromNavigation``:

``` diff
- const navigationTxRef = useRef(navigationTransaction);
+ const navigationTxRef = useRef<typeof navigationTransaction>(undefined);
```

The previous form initialized the ref with the **current** ``route.params.transaction``. If the screen mounted (or remounted via Suspense / native-stack) while ``navigationTransaction`` was already set, the ``useEffect``'s ``ref.current !== navigationTransaction`` diff was false on the first run, so ``setTransaction`` was never called — silently dropping the value the previous screen tried to hand back. Starting the ref at ``undefined`` makes the first non-empty ``navigationTransaction`` always dispatch.

This is a latent bug pre-dating the parent PR; the parent's Suspense screenLayout wrappers (and async bridge timing) likely raised the odds of a remount on this navigation path on real devices, surfacing the issue. The Tezos summary is the only consumer whose ``useBridgeTransaction`` init doesn't seed ``transaction: route.params.transaction``, so it was the only one exposed.

### TLDR for callers

No call-site change required. After this fix, ``useTransactionChangeFromNavigation(setTransaction)`` is guaranteed to apply ``route.params.transaction`` on mount as well as on subsequent navigation updates. For new code, you can still pre-seed the transaction via the ``useBridgeTransaction`` init (as Send / Edit flows do) — that remains the recommended pattern, the hook is now just a safer backstop.

### ❓ Context

- **JIRA or GitHub link**: follow-up of [LIVE-29187](https://ledgerhq.atlassian.net/browse/LIVE-29187) (no dedicated ticket).
- **ADR link (if any)**: n/a.

[LIVE-29187]: https://ledgerhq.atlassian.net/browse/LIVE-29187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre-ledger:
cherry-picked.

PR #17901, opened 2026-05-28
[LWDM] feat(LIVE-30160): Celo pay with token - new send flow

lysyi3m:
### ✅ Checklist

- [x] **Covered by automatic tests.** Bridge helper, descriptor capability, plugin (render / selection / stale-id reset / hydration / unhydrated-label), and the fee-display branch on both LLD and LLM.
  - **Celo send flow (MVVM, LLD + LLM):** AMOUNT step gains a fee-currency picker; verify native CELO + allow-listed tokens (USDT, USDC, USDm, WETH, and regional mTokens — full list in `FEE_CURRENCY_OPTIONS`) and that estimated fees re-compute in the selected token's unit.
  - **Non-Celo coins on the AMOUNT step (LLD):** dialog height stays at `"fit"` unless an Amount plugin is registered for the current account. Sanity-check BTC / ETH / SOL / XRP for visual regressions.
  - **Classic Celo send flow:** shares the `coin-celo` bridge changes — verify the legacy token-fee selector still works.

### 📝 Description

Adapts the Celo "pay gas with tokens" feature — already shipped in the classic send flow — to the new MVVM send flow, while aligning with [ADR-034](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7043678209/Coin+modules+-+ADR-034+-+FlowWizard) (no `family === "..."` branching in shared UI).

**What changed**

- **Descriptor surface (live-common):** new `FeeDescriptor.getFeeCurrencyAccountId?(transaction)` capability + `SendFlowUiConfig.hasAmountPlugins` flag, so shared hooks and the layout consume coin-agnostic config.
- **Celo descriptor:** implements `getFeeCurrencyAccountId`; declares `amount.getPlugins: () => ["celoFeeCurrency"]`. Placeholder `customAssets` stub removed.
- **`coin-celo` bridge:** new `celoEstimateGas` helper that threads `feeCurrency` into raw `eth_estimateGas` (viem's wrapper does not expose it); used by `buildTransaction` and `getFeesForTransaction`.
- **Shared hooks:** desktop `useFeeInfo` + both `useNetworkFees` resolve the display unit from the selected fee sub-account via the new descriptor accessor — no family checks.
- **`SendFlowLayout` (LLD):** reads `uiConfig.hasAmountPlugins` to pin AMOUNT dialog height when a plugin is registered.
- **`CeloFeeCurrencyPlugin` (LLD + LLM):** new picker — Lumen `Menu` on desktop, `QueuedDrawer` on mobile. Filters sub-accounts by allow-list + positive balance; resets stale `feeCurrencyAccountId` (with hydration guard); resolves the trigger label from `feeCurrencyUnwrapped` so it stays correct during sub-account hydration.

### ❓ Context

- **JIRA:** [LIVE-30160](https://ledgerhq.atlassian.net/browse/LIVE-30160)
- **ADR:** [Coin modules — ADR-034 — FlowWizard](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7043678209/Coin+modules+-+ADR-034+-+FlowWizard)

[LIVE-30160]: https://ledgerhq.atlassian.net/browse/LIVE-30160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/components/SendFlowLayout.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/components/SendFlowLayout.tsx
@@ -9,11 +9,6 @@ import {
   type SendFlowStep,
   type SendFlowBusinessContext,
 } from "@ledgerhq/live-common/flows/send/types";
-import {
-  getAccountCurrency,
-  getMainAccount,
-} from "@ledgerhq/ledger-wallet-framework/account/helpers";
-import { sendFeatures } from "@ledgerhq/live-common/bridge/descriptor/send/features";
 import type { SendStepConfig } from "../types";
 import { SendHeader } from "./SendHeader";
 import { AnimatedHeight } from "./AnimatedHeight";
@@ -27,7 +22,7 @@ type SendFlowLayoutProps = Readonly<{
 
 export function SendFlowLayout({ isOpen, onClose }: SendFlowLayoutProps) {
   const wizard = useFlowWizard<SendFlowStep, SendFlowBusinessContext, SendStepConfig>();
-  const { state } = useSendFlowData();
+  const { state, uiConfig } = useSendFlowData();
 
   const currentStepConfig = wizard.currentStepConfig;
   const StepComponent = wizard.currentStepRenderer;
@@ -50,15 +45,8 @@ export function SendFlowLayout({ isOpen, onClose }: SendFlowLayoutProps) {
     [onClose, wizard.currentStep, sendFlowTrackingProperties],
   );
 
-  const accountCurrency = useMemo(() => {
-    if (!state.account.account) return undefined;
-    return getAccountCurrency(
-      getMainAccount(state.account.account, state.account.parentAccount ?? undefined),
-    );
-  }, [state.account.account, state.account.parentAccount]);
   const hasAmountPlugins =
-    wizard.currentStep === SEND_FLOW_STEP.AMOUNT &&
-    sendFeatures.getAmountPlugins(accountCurrency).length > 0;
+    wizard.currentStep === SEND_FLOW_STEP.AMOUNT && uiConfig.hasAmountPlugins;
 
   const dialogHeight = hasAmountPlugins ? "fixed" : (currentStepConfig?.height ?? "fixed");
 
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/components/SendFlowLayout.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/components/SendFlowLayout.tsx
@@ -4,7 +4,16 @@ import { cn } from "LLD/utils/cn";
 import { useFlowWizard } from "../../FlowWizard/FlowWizardContext";
 import { useSendFlowData } from "../context/SendFlowContext";
 import { FLOW_STATUS } from "@ledgerhq/live-common/flows/wizard/types";
-import type { SendFlowStep, SendFlowBusinessContext } from "@ledgerhq/live-common/flows/send/types";
+import {
+  SEND_FLOW_STEP,
+  type SendFlowStep,
+  type SendFlowBusinessContext,
+} from "@ledgerhq/live-common/flows/send/types";
+import {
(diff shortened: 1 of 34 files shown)


apps/ledger-live-desktop/src/mvvm/features/Send/hooks/useNetworkFees.ts
```diff
@@ -74,10 +74,14 @@ export function useNetworkFees({
     feePresetOptions,
   });
 
+  const feeCurrencyAccountId =
+    transaction.family === "celo" ? (transaction.feeCurrencyAccountId ?? null) : null;
```
dilaouid: This cannot happens in the new send flow architecture, we tend to go on a descriptor driven architecture
The descriptors will gives the differences and specificity for each different families (here; `libs/ledger-live-common/src/families/celo/descriptor`)

Then, the `uiConfig` exposed will have stuffs like `hasPayWithToken` (if I understand the PR goal correctly) which will allow your code to ignore the check for each family

tldr; the ultimate goal is to have a totally generic UI, and drop the old schemas of `families/celo`, `families/evm`etc which dealing each different family cases

apps/ledger-live-desktop/src/mvvm/features/Send/components/SendFlowLayout.tsx
```diff
@@ -41,7 +50,17 @@ export function SendFlowLayout({ isOpen, onClose }: SendFlowLayoutProps) {
     [onClose, wizard.currentStep, sendFlowTrackingProperties],
   );
 
-  const dialogHeight = currentStepConfig?.height ?? "fixed";
+  const accountCurrency = useMemo(() => {
```
dilaouid: We do have the account currency from `useSendFlowData()`, is that necessary to recall it here?

dilaouid (commented):
A few comment, basically due to the lack of explanation of how does the new send flow works and where it's leading
Here is an ADR of it: https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7043678209/Coin+modules+-+ADR-034+-+FlowWizard

(Nothing wrong with the plugins because EVM also uses plugins too, but they're also about to be moved to the descriptors in a way or another, so no family will be mentionned in the UI)

gre-ledger (dismissed):

apps/ledger-live-desktop/src/mvvm/features/Send/hooks/useNetworkFees.ts
lysyi3m: @dilaouid thanks for the ADR! that was a missing piece of context, indeed

three changes implemented:
- added a generic `FeeDescriptor.getFeeCurrencyAccountId?(transaction)` capability that exposes the fee sub-account ID without leaking family identity to the UI; Celo's descriptor implements it (the family-typed cast lives inside the descriptor where it belongs).
- added `SendFlowUiConfig.hasAmountPlugins` — precomputed in `getSendUiConfig` from `sendFeatures.getAmountPlugins(currency).length > 0`. The UI reads the boolean.
-  both `useNetworkFees`  hooks (LLD + LLM) and the desktop `SendFlowLayout` are now family-agnostic.

plugin files themselves still live under their named directory and keep an isCeloTransaction typeguard - kept as-is per your note

apps/ledger-live-desktop/src/mvvm/features/Send/components/SendFlowLayout.tsx
lysyi3m: you're right, addressed. 

followed the ADR pattern further: added `hasAmountPlugins: boolean` to `SendFlowUiConfig`, populated in `getSendUiConfig` from `sendFeatures.getAmountPlugins(currency).length > 0`. the layout now reads it straight from `useSendFlowData()`:

dilaouid (approved):

gre-ledger (approved):

PR #18048, opened 2026-06-01
[LWDM] feat(evm-staking): fetch pending rewards for cosmos-evm chains

YazhuEth:
## Summary

- Adds `RewardsStrategy` config + `fetchRewards()` populating `Stake.amountRewarded`. First impl: `cosmos-rest`, configured for Sei. Queries the Cosmos distribution REST endpoint after resolving the EVM address to its bech32 counterpart via the chain's address precompile.
- Fetch fired in parallel with the per-validator precompile loop so its latency overlaps the existing RPC reads; resolves to an empty map on failure to keep sync non-blocking.
- Wires the previously-empty per-row "Claim rewards" handler on desktop so it opens the claim modal preselected on the chosen validator.

## Context

Before this PR, the EVM staking fetcher only read delegated amounts from the staking precompile (`0x...1005`); rewards from the distribution module were never queried, so `Stake.amountRewarded` was implicitly always `0n`. The header "Claim rewards" button was therefore always disabled, and the per-row claim action did nothing (empty callback). This trio of issues was masked because the claim feature in the parent PR couldn't surface real reward values until they reached the UI.

- **JIRA or GitHub link**: [LIVE-31795](https://ledgerhq.atlassian.net/browse/LIVE-31795)

[LIVE-31795]: https://ledgerhq.atlassian.net/browse/LIVE-31795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- E2E
   - Desktop: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/e0991a62-61fe-44dc-8b15-8d1d4b29c244/
   - Mobile: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/2838c20c-2151-4c0e-86e7-36d372cecad1/

### DIFF
--- a/libs/coin-modules/coin-evm/src/staking/rewards.test.ts
+++ b/libs/coin-modules/coin-evm/src/staking/rewards.test.ts
@@ -1,0 +1,196 @@
+import { ethers } from "ethers";
+import { fetchRewards } from "./rewards";
+import { USEI_TO_EVM_SCALE } from "../utils";
+
+jest.mock("@ledgerhq/live-network", () => jest.fn());
+jest.mock("../config");
+jest.mock("../network/node/types");
+
+const mockNetwork = jest.mocked(require("@ledgerhq/live-network"));
+const mockGetCoinConfig = jest.mocked(require("../config").getCoinConfig);
+const mockIsExternalNodeConfig = jest.mocked(require("../network/node/types").isExternalNodeConfig);
+
+const EVM_ADDRESS = "0x1234567890abcdef1234567890abcdef12345678";
+const COSMOS_ADDRESS = "sei1defaultcosmosaddressfortest";
+
+type RewardCoin = { denom: string; amount: string };
+const restResponse = (rewards: Array<{ validator_address: string; reward: RewardCoin[] }>) => ({
+  data: { rewards },
+});
+
+describe("fetchRewards", () => {
+  let contractSpy: jest.SpyInstance;
+  let providerSpy: jest.SpyInstance;
+  let mockGetSeiAddr: jest.Mock;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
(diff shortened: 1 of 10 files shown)


libs/coin-modules/coin-evm/src/staking/rewards.ts
```diff
@@ -0,0 +1,90 @@
...
+type RewardCoin = { denom: string; amount: string };
+type ValidatorReward = {
+  validator_address: string;
+  reward: RewardCoin[];
+};
+type RewardsApiResponse = { rewards: ValidatorReward[] };
+
+// Truncate to the integer part to match what the chain actually pays out on
+// withdrawal (fractional sub-denom rewards are dropped on-chain).
+function parseDecimalIntegerPart(amount: string): bigint {
```
Moustafa-Koterba: you can put it into a utils file

libs/coin-modules/coin-evm/src/staking/rewards.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  data: { rewards },
+});
+
+describe("fetchRewards", () => {
+  let mockGetSeiAddr: jest.Mock;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockGetSeiAddr = jest.fn().mockResolvedValue(COSMOS_ADDRESS);
+    jest.spyOn(ethers, "JsonRpcProvider" as any).mockImplementation(() => ({}));
```
Moustafa-Koterba: Better use `mockReturnValue` if there is no logic
```suggestion
    jest.spyOn(ethers, "JsonRpcProvider" as any).mockReturnValue({});
```
same for others

Moustafa-Koterba (approved):

PR #18055, opened 2026-06-01
[LWDM] feat(cardano): add lastBlock CoinModule API method

lysyi3m:
### ✅ Checklist

- [x] **Covered by automatic tests.** Hermetic unit tests (mapping + error/boundary cases), an API delegation test, and a live-network integration test.
- [ ] **Impact of the changes:** Cardano CoinModule API only. Cardano is not yet wired into the generic coin-framework bridge, so there is no impact on the live sync/send flows. QA focus: none on current flows.
  - ...

### 📝 Description

Implements `lastBlock()` for the Cardano CoinModule ("Alpaca") API. It fetches the chain tip via the Ledger Cardano API `GET /v1/block/latest` (`{ blockHeight }`) and maps it to `BlockInfo`.

Backend limitation: `/v1/block/latest` exposes **only the tip height** — no hash, no timestamp, and there is no block-by-height endpoint. So:

- `height` ← `blockHeight` (guarded: a missing/`NaN`/non-positive value throws instead of leaking into `BlockInfo.height`).
- `hash` ← `""` (not exposed by the API; same approach as coin-canton, the other height-only backend).
- `time` ← `new Date()` — the tip is produced within one ~20s block interval, so it is approximated as "now". Valid only for the tip; `getBlock` / `getBlockInfo` remain `not supported` rather than fabricate historic block data.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18628](https://ledgerhq.atlassian.net/browse/LIVE-18628)
- **ADR link (if any)**:

[LIVE-18628]: https://ledgerhq.atlassian.net/browse/LIVE-18628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-modules/coin-cardano/src/api/getLatestBlock.ts
+++ b/libs/coin-modules/coin-cardano/src/api/getLatestBlock.ts
@@ -1,0 +1,15 @@
+import network from "@ledgerhq/live-network/network";
+import { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+import { CARDANO_API_ENDPOINT, CARDANO_TESTNET_API_ENDPOINT } from "../constants";
+import { isTestnet } from "../logic";
+import { APILatestBlock } from "./api-types";
+
+export async function fetchLatestBlock(currency: CryptoCurrency): Promise<APILatestBlock> {
+  const res = await network({
+    method: "GET",
+    url: isTestnet(currency)
+      ? `${CARDANO_TESTNET_API_ENDPOINT}/v1/block/latest`
+      : `${CARDANO_API_ENDPOINT}/v1/block/latest`,
+  });
+  return res && (res.data as APILatestBlock);
+}
--- a/.changeset/quick-lemons-rest.md
+++ b/.changeset/quick-lemons-rest.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-cardano": minor
+---
+
+Add lastBlock method to Coin Cardano CoinModule API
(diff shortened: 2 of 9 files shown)


libs/coin-modules/coin-cardano/src/api/getLatestBlock.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export async function fetchLatestBlock(currency: CryptoCurrency): Promise<APILatestBlock> {
+  const res = await network({
+    method: "GET",
+    url: isTestnet(currency)
+      ? `${CARDANO_TESTNET_API_ENDPOINT}/v1/block/latest`
+      : `${CARDANO_API_ENDPOINT}/v1/block/latest`,
+  });
+  return res && (res.data as APILatestBlock);
+}
```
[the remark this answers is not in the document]
lysyi3m: we're following existing pattern in coin-cardano API callers implementation. we're not going to change the approach mid-way and end up with two implementation of the network layer at the same time. this migration might happend at once later

henri-ly (approved):

libs/coin-modules/coin-cardano/src/api/getLatestBlock.ts
```diff
@@ -0,0 +1,15 @@
+import network from "@ledgerhq/live-network/network";
```
francois-guerin-ledger, line 1:
```suggestion
import network from "@ledgerhq/live-network";
```
Importing `/network` is deprecated. You should have access to a generic now, so you can drop casting to `APILatestBlock`.

lysyi3m: the ticket to track it is created https://ledgerhq.atlassian.net/browse/LIVE-31700

PR #18059, opened 2026-06-01
[LWDM] feat(cardano): add getBalance CoinModule API method

lysyi3m:
### ✅ Checklist

- [x] **Covered by automatic tests.** Hermetic unit tests (UTXO derivation, native/token/stake mapping, dRep reward rule, `computeAdaBalance` total/spendable incl. the floor-at-zero edge case, error cases), pagination tests, and a live-network API-level integration test through `createApi(...).getBalance` (native + token + options-rejection).
- [x] **Impact of the changes:** The CoinModule API method has no impact on live flows (Cardano isn't wired into the generic coin-framework). The shared balance-math helpers, however, are also called from the account bridge (`synchronisation.ts` `makeGetAccountShape`), which **is** used by LLD/LLM wallet sync. Behavior-preserving except one reconciled edge case: when held tokens require more backing ADA than the address holds **and** the account is dRep-delegated, spendable now follows the bridge definition (`spendable = max(0, utxosSum − minAda) + rewards`); `locked` stays within `[0, total]`. **QA focus:** Cardano account **balance** and **spendable balance**, especially accounts holding native tokens and/or staking rewards (delegated vs not delegated to a dRep).

### 📝 Description

Implements `getBalance(address)` for the Cardano CoinModule ("Alpaca") API. Returns `[native, ...stake, ...tokens]`:

- **Native + tokens** — derived from the UTXO set of the address's payment credential, fetched via the shared `getAllTransactionsByKeys` pagination primitive (the same one account sync uses): native ADA sum plus one balance per held token. `locked` = min-ADA backing the tokens + rewards not yet withdrawable (no dRep delegation).
- **Staking** — `/v1/delegation` (by stake credential) mapped to a separate native `Stake` entry: `delegate` = pool, `amountDeposited` = stake-key deposit, `amountRewarded` = rewards, `amount` = their sum. `claim_reward` is advertised only when delegated to a dRep (Conway rule). Omitted when there is no staking position.

The token min-ADA and the total/spendable split are owned by two shared helpers in `logic.ts` — `calculateMinAdaForTokens` and `computeAdaBalance` — called by both this method and the account bridge, so the math has a single source of truth. Structure and stake mapping follow the Tezos CoinModule precedent (liquid-delegation analog).

**Known limitation:** the framework passes a single `freshAddress` (no xpub), so native/token totals cover only that address's payment credential — a Cardano account spreads funds across many addresses. Account-level aggregation requires xpub support in the generic bridge (same constraint as `listOperations`). Staking is account-complete (the stake credential is shared across all the account's addresses).

### ❓ Context

- **JIRA or GitHub link**: [LIVE-18630](https://ledgerhq.atlassian.net/browse/LIVE-18630)
- **ADR link (if any)**:

[LIVE-18630]: https://ledgerhq.atlassian.net/browse/LIVE-18630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-modules/coin-cardano/src/logic/getBalance.integ.test.ts
```diff
@@ -0,0 +1,20 @@
+import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets";
+import { getBalance } from "./getBalance";
+
+describe("getBalance (integration)", () => {
```
hedi-edelbloute: Can we add a case with tokens ? also, I think we should have integ tests at the api level not logic level

libs/coin-modules/coin-cardano/src/logic/getBalance.ts
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Balances held by a Cardano address: native ADA + each token from its UTXOs, plus the
+ * account's staking position (when delegated) as a separate native balance entry.
+ *
+ * Caveat: native/token totals cover only the payment credential of the passed address.
+ * A Cardano account holds funds across many addresses (derived from its xpub), which the
+ * single-address CoinModule contract cannot enumerate. Staking is account-complete (the
+ * stake credential is shared across all the account's addresses).
+ */
+export async function getBalance(currency: CryptoCurrency, address: string): Promise<Balance[]> {
```
hedi-edelbloute, line 127: How difficult is it to rewire in current bridge implementation ?

lysyi3m: we've extracted what we could like computations and reused these helpers, see https://github.com/LedgerHQ/ledger-live/pull/18059/changes/fa5c5324e63e073d696bdf890534206483b45c99

for the full parity re-write I don't think there is a quick win, tbh

ishaba (dismissed):

ishaba (dismissed):

libs/coin-modules/coin-cardano/src/api/getTransactionsByPaymentKey.ts
hedi-edelbloute, line 1: In API folder we want only CoinModuleApi, move to logic or network folder if required

hedi-edelbloute (dismissed):

hedi-edelbloute (approved):

ishaba (approved):

may01 (approved):

PR #18083, opened 2026-06-02
[LWDM] fix(coin-monitoring): run CLI as ESM for ESM-only live-common

gre-ledger:
> [!NOTE]
> Stacked on top of #18028 (live-common ESM-only). Merge into that branch to turn its `coin-modules-monitoring` CI job green.

### ✅ Checklist

- [x] **Covered by automatic tests.** Existing `run.test.ts` passes; validated by running `monitor --currencies all --account-types pristine` end-to-end.
- [ ] **Impact of the changes:** `coin-modules-monitoring` tooling only. No app/runtime impact.

### 📝 Description

#18028 makes live-common **ESM-only**, which broke the `coin-modules-monitoring` CLI (`ERR_MODULE_NOT_FOUND: lodash/memoize`). The CLI ran as CJS and force-loaded every `@ledgerhq/*` as CJS; with live-common having no CJS build anymore, it loaded as native ESM and its bundler-targeted `lib-es` violated Node's strict-ESM rules.

Fix:
- Run the CLI as **native ESM** (`lib-es/cli.js`) and build ESM-only — so `@ledgerhq/*` resolve to a single ESM instance graph (fixes a `LiveConfig` singleton split that surfaced as "Config not set" on ~14 currencies).
- Rework `loader.mjs` to keep `@ledgerhq/*` as ESM and patch the bundler-targeted `lib-es`:
  - resolve extensionless / directory / bare-subpath imports (e.g. `lodash/memoize`)
  - inject the `type: "json"` import attribute Node requires
  - shim `require()` calls left in `lib-es` (e.g. `live-network`)
  - emulate `esModuleInterop` for third-party CJS deps (`lodash/fp`, `@apollo/client`, `bip32`, `casper-js-sdk`, `icon-sdk-js`)

Validated locally: `monitor --currencies all` runs every currency end-to-end (the lone skip was a transient `LedgerAPI4xx: Not Enough CU` API quota on mina, unrelated to module loading). `typecheck` + `jest` pass.

Screenshot 2026-06-02 at 12 08 29

### ❓ Context

- **JIRA or GitHub link**: LIVE-29648
- **ADR link (if any)**:

### DIFF
--- a/libs/coin-modules-monitoring/loader.mjs
+++ b/libs/coin-modules-monitoring/loader.mjs
@@ -1,62 +1,208 @@
-// Problems this loader solves when running CJS code that transitively pulls
-// in @mysten/ledgerjs-hw-app-sui (ESM-only):
-//
-// 1. @ledgerhq/* lib-es files use extensionless relative imports (e.g.
-//    `import './helpers'`) which Node's strict ESM resolver rejects.
-//
-// 2. @ledgerhq/* lib-es files are compiled for bundlers and contain `require()`
-//    calls that are invalid in native ESM context.
-//
-// 3. Some @ledgerhq/* packages (e.g. @ledgerhq/live-dmk-shared) only ship
-//    lib-es/, so the require→CJS swap doesn't apply — and their files contain
-//    directory imports like `export * from "./hooks"` which Node's ESM resolver
-//    rejects with ERR_UNSUPPORTED_DIR_IMPORT.
-//
-// Fix for (1) and (2): when @ledgerhq/* packages are resolved in ESM context,
-// force the `require` condition so Node picks lib/ (CJS) instead of lib-es/.
-//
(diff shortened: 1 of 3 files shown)


libs/coin-modules-monitoring/loader.mjs
```diff
@@ -1,62 +1,208 @@
...
+      return m ? `${m[1]}: ${m[2]}` : s;
+    })
+    .join(", ");
+}
+
+let tmpId = 0;
+// Rewrite a single-line `import` of a third-party CJS dep into require() + the
+// interop helpers tsc emits under esModuleInterop. Returns null to leave a line
+// untouched (ESM dep, our own code, relative import, etc.).
+function rewriteImportLine(line, parentUrl) {
```
gre-ledger, line 132: i don't like this. i am trying to find a better solution

gre-ledger:
https://github.com/LedgerHQ/ledger-live/actions/runs/26811141590

PR #18092, opened 2026-06-02
[LWM] fix: prevent duplicate transaction broadcast on device action re-render

qperrot:
### ✅ Checklist

  - ...

### 📝 Description
DeviceAction accepts a renderOnResult prop which is a render function. Because it was called inline during rendering, any re-render while the signed result was on screen would re-invoke the function — including the broadcast side-effect. On Hedera this surfaced visibly as DUPLICATE_TRANSACTION, but Tron would be equally vulnerable since its broadcast.ts just fires a raw HTTP call to the Tron network.

RenderOnResultWithJSX (lines 832–841) — a defensive wrapper at the DeviceAction level that uses a useState lazy initializer so renderOnResult is called exactly once on mount, not on every render. This is a belt-and-suspenders guard in case other callers of DeviceAction pass a side-effectful renderOnResult without their own ref guard.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-31607](https://ledgerhq.atlassian.net/browse/LIVE-31607)
- **ADR link (if any)**: 

---

[LIVE-31607]: https://ledgerhq.atlassian.net/browse/LIVE-31607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/yellow-frogs-talk.md
+++ b/.changeset/yellow-frogs-talk.md
@@ -1,0 +1,5 @@
+---
+"live-mobile": patch
+---
+
+Fix: prevent duplicate transaction broadcast on device action re-render, protecting all chains (including Tron) from double-submission via a mount-once guard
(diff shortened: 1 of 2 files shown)


OlivierFreyssinet (changes requested):
can you please list all of the impacted flows in the PR description ?

PR #18140, opened 2026-06-03
ci(sonar): faster checkout + skip Kotlin analyzer

dgreen-ledger:
## What

Two low-risk reductions to the **Sonar Cloud** job. Note up front: **Sonar is not a required check** (required = `OK`/`check-sync`, and `OK` doesn't depend on `scan-sonar`), so this does **not** speed up time-to-merge — it shortens the full-run "all-green" time and frees the shared `ledger-live-4xlarge` runner sooner.

The Sonar job's ~358s wall breaks down as ~85s git checkout + ~215s analyzer + ~30s setup/upload.

1. **Partial-clone checkout** (`filter: blob:none`). Keeps `fetch-depth: 0` (Sonar needs full history for SCM blame / new-code detection) but lazy-fetches blobs instead of pulling every file's content up front → much faster checkout on this large repo. Est. ~85s → ~15–25s.
2. **Disable the Kotlin analyzer** (`sonar.kotlin.file.suffixes=-`, mirroring the existing `c/cpp/objc` disables). This is a JS/TS repo; the Kotlin sensor only adds startup overhead. Est. ~few seconds.

## What this does NOT touch

The dominant cost is the **~120s sequential JS/TS analysis** of 12.5k files. I verified there is **no `sonar.javascript.maxThreads`** — SonarJS runs single-threaded through one Node bridge process, and SonarSource's guidance is to split into multiple projects. Reducing that ~120s requires either:
- ensuring the **SonarCloud PR analysis cache** is active (a SonarCloud project setting, not a repo config), or
- **splitting the monorepo into per-app Sonar projects** so each PR only analyzes affected sources (architectural).

## Estimated impact

~80–95s off the ~358s Sonar job (~22–26%), entirely in checkout + Kotlin. Non-merge-blocking, so the benefit is runner-capacity + perceived completion, not PR merge latency.

### DIFF
--- a/.github/workflows/scan-sonar-reusable.yml
+++ b/.github/workflows/scan-sonar-reusable.yml
@@ -1,6 +1,13 @@
 name: "[All Platforms] - Scan Sonar - Called"
 
 on:
+  # TEMPORARY (measurement only — remove before merge): allows dispatching the
+  # Sonar job standalone on a branch to benchmark checkout + analyzer timing.
+  workflow_dispatch:
+    inputs:
+      ref:
+        required: false
+        type: string
   workflow_call:
     inputs:
       should_download_desktop_coverage:
--- a/.github/workflows/scan-sonar-reusable.yml
+++ b/.github/workflows/scan-sonar-reusable.yml
@@ -49,6 +49,10 @@ jobs:
         with:
           ref: ${{ inputs.ref || github.sha }}
           fetch-depth: 0
+          # Partial clone: keep full history (Sonar needs it for SCM blame /
+          # new-code detection) but lazy-fetch blobs instead of downloading
+          # every file's content up front — much faster checkout on this repo.
+          filter: blob:none
           persist-credentials: false
 
       - name: fetch develop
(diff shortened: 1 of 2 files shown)


dgreen-ledger:
Closing — **measurement showed no safe quick win**, documented here for the record.

Benchmarked via a temporary `workflow_dispatch` on this branch (run 26874886462):

- **`filter: blob:none` (partial-clone checkout) breaks Sonar.** Checkout dropped 85s → 28s, but the scan **failed (exit 3)**: Sonar's SCM blame uses JGit, which doesn't lazy-fetch missing blobs — `MissingObjectException`. Shallow clone would fail the same way. The full-blob `fetch-depth: 0` checkout is effectively required for blame / new-code detection.
- **Analyzer can't be cut by config.** No `sonar.javascript.maxThreads` exists (SonarJS is single-threaded). The standalone dispatch ran as a *branch* scan (cold) and analyzed **10,385 files in ~11 min**, vs a real *PR* scan's **229 files in ~215s** — confirming the **SonarCloud PR analysis cache is already active and is the dominant optimization**.
- **Sensor disabling is negligible** — the Kotlin sensor was 38ms; IaC sub-analyzers (~7s) aren't cleanly disableable via properties.

**Conclusion:** the Sonar job is near its floor for a single-project monorepo. The only material lever left is **splitting into per-app SonarCloud projects** so each PR analyzes only affected sources — an architectural change, tracked separately if pursued. Also note Sonar is **not merge-blocking** (required checks are `OK`/`check-sync`), so this was full-run/runner-capacity only.

PR #18179, opened 2026-06-03
[LWDM] fix(aptos): fix prepare transaction [LIVE-31542]

semeano:
### ✅ Checklist

  - @ledgerhq/coin-aptos

### 📝 Description

Fix intermittent issues with fees calculations for Aptos transactions

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-31542

---

### DIFF
--- a/libs/coin-modules/coin-aptos/src/__tests__/bridge/prepareTransaction.test.ts
+++ b/libs/coin-modules/coin-aptos/src/__tests__/bridge/prepareTransaction.test.ts
@@ -2,6 +2,10 @@ import BigNumber from "bignumber.js";
 import { getEstimatedGas } from "../../bridge/getFeesForTransaction";
 import { getMaxSendBalance } from "../../bridge/logic";
 import prepareTransaction from "../../bridge/prepareTransaction";
+import {
+  APTOS_DELEGATION_RESERVE_IN_OCTAS,
+  MIN_COINS_ON_SHARES_POOL_IN_OCTAS,
+} from "../../constants";
 import { AptosAPI } from "../../network";
 import type { AptosAccount, Transaction } from "../../types";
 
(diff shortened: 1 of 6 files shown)


libs/coin-modules/coin-aptos/src/bridge/prepareTransaction.ts
```diff
@@ -74,8 +74,16 @@ const prepareTransaction = async (
   const aptosClient = new AptosAPI(account.currency.id);
   const tokenAccount = findSubAccountById(account, transaction?.subAccountId ?? "");
 
+  const estimationTransaction = transaction.useAllAmount
+    ? { ...transaction, amount: BigNumber(0) }
+    : transaction;
```
cted-ledger, line 79: Should the zeroing apply to staking modes too ? On-chain, withdraw(0) seems to abort (EWITHDRAW_ZERO_STAKE), and add_stake/unlock/reactivate_stake short-circuit on 0.
So I'm wondering whether the gas estimate stays meaningful there, or if this should be limited to send.
wdyt?

libs/coin-modules/coin-aptos/src/__tests__/bridge/prepareTransaction.test.ts
```diff
@@ -71,6 +71,24 @@ describe("Aptos prepareTransaction", () => {
...
+        fees: new BigNumber(2000),
+        estimate: { maxGasAmount: new BigNumber(200), gasUnitPrice: new BigNumber(10) },
+        errors: {},
+      });
+
+      await prepareTransaction(account, transaction);
+
+      const [, estimationTransaction] = (getEstimatedGas as jest.Mock).mock.calls[0];
+      expect(estimationTransaction.amount.isZero()).toBe(true);
+    });
```
cted-ledger, line 94: Should we cover a staking mode with useAllAmount (e.g. withdraw/unstake), to make the intended behaviour explicit there?

libs/coin-modules/coin-aptos/src/constants.ts
```diff
@@ -113,5 +113,5 @@ export enum TOKEN_TYPE {
   FUNGIBLE_ASSET = "fungible_asset",
 }
 
-export const DEFAULT_GAS = new BigNumber(200);
+export const DEFAULT_GAS = new BigNumber(400);
```
cted-ledger: Out of curiosity, what drove the 200 → 400 bump ?
Since it's now the simulation maxGasAmount ceiling, would it help to note where the value comes from?

cted-ledger:
for `useAllAmount`, this path simulates with the full balance while `prepareTransaction` now uses `0`...
Is that divergence intentional, or worth aligning ?

libs/coin-modules/coin-aptos/src/constants.ts
may01: Does this for tokens transfer or normal transfer. For normal the 200 GAS should be enough: https://aptos.dev/build/smart-contracts/error-codes#insufficient_balance_for_transaction_fee
Does it make sense to resolve the total gas amount with `gas_price` not `gas_unit`?

may01 (dismissed):

libs/coin-modules/coin-aptos/src/constants.ts
semeano: This was some left over from the test. I removed it

libs/coin-modules/coin-aptos/src/bridge/prepareTransaction.ts
semeano: Staking or any other staking related event are basically just a transaction with a function call. It's not that different from a regular send. In terms of gas estimation it doesn't change much.
This only affects the simulation input, keeps fees correct, doesn't change the final staked amount, and doesn't bypass staking validation.

may01 (approved):

PR #18200, opened 2026-06-04
feat(web-tools): set trustchain application path in playground

gre-ledger:
### ✅ Checklist

- [ ] npx changeset was attached. 
  - Wallet Sync Trustchain Playground only. No app/library runtime impact.

### 📝 Description

Screenshot 2026-06-04 at 12 50 05

Adds an **Application** picker in the playground Environment section to switch the trustchain applicationId, which drives the derivation path m/0'/<applicationId>'/…

- 16 — Ledger Sync (default)
- 17 — wallet-cli ring ( https://github.com/LedgerHQ/ledger-live/pull/17743 )

Cloud Sync SDK and Accounts Sync are Ledger Sync features, so they now only render when the application is 16. On other applications the playground exercises the Trustchain SDK only (e.g. to reproduce wallet-cli ring init flows), with a short note explaining why those sections are hidden.

### ❓ Context

- **ADR link (if any)**: N/A

### DIFF
--- a/apps/web-tools/src/trustchain/components/App.tsx
+++ b/apps/web-tools/src/trustchain/components/App.tsx
@@ -19,6 +19,7 @@ import { getSdk } from "@ledgerhq/ledger-key-ring-protocol";
 import { DisplayName, IdentityManager } from "./IdentityManager";
 import { AppSetDeviceId } from "./AppSetDeviceId";
 import { AppSetSupportedCurrencies } from "./AppSetSupportedCurrencies";
+import { AppSetApplicationId } from "./AppSetApplicationId";
 import { AppQRCodeCandidate } from "./AppQRCodeCandidate";
 import { TrustchainSDKContext, defaultContext } from "../context";
 import { DeviceInteractionVisibleContext } from "../deviceInteractionContext";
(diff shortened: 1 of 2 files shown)


apps/web-tools/src/trustchain/components/App.tsx
```diff
@@ -91,6 +97,24 @@ const App = () => {
...
 
+  // A trustchain is derived for a specific applicationId (its applicationPath is
+  // m/0'/<applicationId>'/…). Keep the selected applicationId and the loaded
+  // trustchain consistent so SDK calls never run against the wrong application:
+  // - when the user switches application, drop the now-mismatched trustchain so
+  //   they re-create/restore it under the new path;
+  // - when a trustchain for another application gets loaded (e.g. identity
+  //   switch), follow it by realigning the selected applicationId.
+  const selectedApplicationIdRef = useRef(applicationId);
+  useEffect(() => {
```
Sebastien-Dav1d: For a playground, it is not that important but
We could derive `applicationId` from the trustchain instead of the effect:

```tsx
const [selectedApplicationId, setSelectedApplicationId] = useState<number>(
  defaultContext.applicationId,
);

const applicationId = trustchain
  ? applicationIdOfPath(trustchain.applicationPath)
  : selectedApplicationId;

const handleSetApplicationId = useCallback(
  (id: number) => {
    setSelectedApplicationId(id);
    if (trustchain && applicationIdOfPath(trustchain.applicationPath) !== id) {
      setTrustchain(null);
    }
  },
  [trustchain, setTrustchain],
);
```

Then `<AppSetApplicationId setApplicationId={handleSetApplicationId} />`.

- user switches app → handler drops the mismatched trustchain
- trustchain loaded elsewhere → `applicationId` derives from it

No more `useEffect` / `selectedApplicationIdRef`, and the app can never drift from the loaded trustchain.

Sebastien-Dav1d (approved):

PR #18297, opened 2026-06-08
[LWM] feat: add post onboarding widget completed track event to lwm

mitchellv-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Add tracking event for post onboarding completion in lwm

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-31814

---

### DIFF
--- a/apps/ledger-live-mobile/src/logic/postOnboarding/__tests__/usePostOnboardingCompletionTracking.test.ts
+++ b/apps/ledger-live-mobile/src/logic/postOnboarding/__tests__/usePostOnboardingCompletionTracking.test.ts
@@ -1,0 +1,131 @@
+import { renderHook } from "@testing-library/react-native";
+import { DeviceModelId } from "@ledgerhq/types-devices";
+import { track } from "~/analytics";
+import { usePostOnboardingHubState } from "@ledgerhq/live-common/postOnboarding/hooks/index";
+import { usePostOnboardingHubStepperDisplay } from "../usePostOnboardingHubStepperDisplay";
+import { usePostOnboardingCompletionTracking } from "../usePostOnboardingCompletionTracking";
+
+jest.mock("@ledgerhq/live-common/postOnboarding/hooks/index", () => ({
+  usePostOnboardingHubState: jest.fn(),
+}));
+jest.mock("../usePostOnboardingHubStepperDisplay", () => ({
+  usePostOnboardingHubStepperDisplay: jest.fn(),
+}));
+
+const mockedTrack = track as jest.Mock;
+const mockedHubState = usePostOnboardingHubState as jest.Mock;
+const mockedStepper = usePostOnboardingHubStepperDisplay as jest.Mock;
+
+type Stepper = { areAllActionsCompleted: boolean; loading: boolean };
+
+const setStepper = ({ areAllActionsCompleted, loading }: Stepper) =>
+  mockedStepper.mockReturnValue({ areAllActionsCompleted, loading });
+
+// A single populated action is enough for the tracking hook's "ready" check.
+const ACTIONS_STATE = [{ id: "claimMock" }];
+
+describe("usePostOnboardingCompletionTracking", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockedHubState.mockReturnValue({
+      actionsState: ACTIONS_STATE,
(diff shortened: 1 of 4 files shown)


RobinVncnt (dismissed):

apps/ledger-live-mobile/src/logic/postOnboarding/usePostOnboardingCompletionTracking.ts
```diff
@@ -0,0 +1,35 @@
...
+import { usePostOnboardingHubStepperDisplay } from "./usePostOnboardingHubStepperDisplay";
+
+export function usePostOnboardingCompletionTracking(): void {
+  const { actionsState, deviceModelId } = usePostOnboardingHubState();
+  const { areAllActionsCompleted, loading } = usePostOnboardingHubStepperDisplay(actionsState);
+
+  const initializedRef = useRef(false);
+  const hasTrackedRef = useRef(false);
+
+  useEffect(() => {
```
LucasWerey, line 13: I'm not sure it's a good thing to do the tracking based on a useEffect, IMO it should be triggered after an event. It's a bit hacky to do it like this

mitchellv-ledger: I would normally agree but in this case the state changes can happen async and from various places and I think already has a state change which happens from a useEffect with repeated timer (a check if user has a positive balance).  Would be tricky to implement in a different way.

LucasWerey: But you are doing a useEffect on `areAllActionsCompleted` value that comes from a hook. You could pass a callback to this hook instead of doing a useEffect no? (I don't have the full picture regarding this so maybe not)

apps/ledger-live-mobile/src/logic/postOnboarding/__tests__/usePostOnboardingCompletionTracking.test.ts
```diff
@@ -0,0 +1,105 @@
+import { renderHook } from "@testing-library/react-native";
+import { DeviceModelId } from "@ledgerhq/types-devices";
+import { track } from "~/analytics";
+import { usePostOnboardingHubState } from "@ledgerhq/live-common/postOnboarding/hooks/index";
+import { usePostOnboardingHubStepperDisplay } from "../usePostOnboardingHubStepperDisplay";
+import { usePostOnboardingCompletionTracking } from "../usePostOnboardingCompletionTracking";
+
+jest.mock("~/analytics", () => ({ track: jest.fn() }));
```
LucasWerey: not needed it's already done in jestSetup

apps/ledger-live-mobile/src/logic/postOnboarding/usePostOnboardingCompletionTracking.ts
mitchellv-ledger: **Simple summary**: An event-driven callback won't work here because "all actions completed" is derived reactive state (some actions complete via a background balance poll or async checks with no user event), and the only hook that computes it is shared across 5 UI sites and would still need its own effect, so isolating the effect in usePostOnboardingCompletionTracking is the cleaner separation.

**Long answer**: You're right that an event-driven callback is usually cleaner than a useEffect. The problem here is that "all actions completed" isn't an event — it's derived reactive state. An action can be marked complete with no user interaction at all: assetsTransfer is completed by a 3-second debounced background balance poll (useAccountsWithFundsListener at the app root), and some actions resolve via an async subscription check (getIsAlreadyCompleted) or a reactive state check (getIsAlreadyCompletedByState). There's no single call site I can attach a callback to that covers all of these.

Passing a callback would mean putting it on usePostOnboardingHubStepperDisplay, since that's the hook that computes the aggregate. But that hook is shared by 5 UI call sites, and it would still need its own internal useEffect to detect the async transition — so it doesn't remove the effect, it just moves analytics-specific logic (transition detection + dedup + deviceModelId gating) into a display hook. Keeping it isolated in usePostOnboardingCompletionTracking felt like the better separation. 

Open to revisiting if you see a call site I'm missing!

RobinVncnt (approved):

LucasWerey (approved):

PR #18437, opened 2026-06-10
feat(wallet-cli): add optional BLE transport for Flex/Stax (Node, via DMK)

### LINKED ISSUE #18436: [Feature]: wallet-cli BLE transport for Flex/Stax (currently USB only)
### Library name

@ledgerhq/wallet-cli

### Describe your feature request

`@ledgerhq/wallet-cli` 1.0.2 registers the DMK transport over USB only (`apps/wallet-cli/src/device/node-webusb`, `register-dmk-transport.ts`). Flex and Stax are BLE first devices, and the main audience for a wallet CLI right now is agent and automation setups, where the host running the CLI usually does not have the device on a cable. The device stays next to the human, the agent runs on a server or a machine across the room.

I build an agent marketplace on Solana (AgenC) where agents settle real payments and the signer is a Flex over BLE on the DMK, so this exact flow runs in production for us, just not through wallet-cli yet. Things we hit in production that the BLE path needs to handle:

- one device session for the whole flow. Opening a second BLE session mid flow (one for app open, another for signing) leaves the channel busy and everything hangs after that
- auto open the target app from the dashboard through DMK, so the human only unlocks the device and never navigates menus
- machine readable error codes that distinguish device locked, app not open and BLE channel busy. An automation needs to decide between "ask the human to unlock" and "just retry", and today those cases look the same from the outside

Proposal: a `--transport usb|ble` flag, default usb so nothing changes for current users. The DMK already ships a BLE transport, so the bulk of the work is registering it plus the session handling above.

Happy to send the PR if there is interest. We already run this combination (DMK + BLE + Solana app auto open) in production and I would rather upstream it than keep it in our own stack.

signerless:
Closes #18436.

## What

Adds an optional Bluetooth Low Energy transport to `wallet-cli`, selectable with `WALLET_CLI_TRANSPORT=ble`. The default stays `usb`, so nothing changes for current users.

`@ledgerhq/device-management-kit` ships a Node transport for USB-HID/WebUSB only; its BLE transports target the browser (`device-transport-kit-web-ble`) and React Native (`device-transport-kit-react-native-ble`). There is no Node BLE transport, so the CLI cannot talk to a Flex/Stax over Bluetooth today. This adds one, backed by `@abandonware/noble`.

The motivation is the agent/automation use case from the issue: the host running the CLI usually does not have the device on a cable — the device stays with the human and pairs over BLE.

## How

- `src/device/node-ble/NodeBleTransport.ts` implements the DMK `Transport` contract and is registered in `createDeviceManagementKit()` only when the BLE transport is selected.
- Framing is delegated to DMK's injected `ApduSender`/`ApduReceiver` services with `channel = Nothing` and `padding = false`. With no channel, DMK emits exactly the `[0x05][seq][len][payload]` block layout Ledger BLE expects, so the transport writes raw frames to the GATT write characteristic and feeds notification buffers straight back to DMK — no hand-rolled framer.
- BLE service/characteristic UUIDs and device models come from `deviceModelDataSource.getBluetoothServicesInfos()`, so the transport stays in sync with the device models DMK knows about. The `0x08` GET-MTU handshake mirrors `@ledgerhq/hw-transport-node-ble`.
- `noble` is imported lazily, so USB-only runs never initialize the platform Bluetooth stack.
- On the BLE path the process exits explicitly after teardown: `noble` keeps a native libuv handle that never unrefs, so without this the process hangs after a command completes. USB unrefs its hotplug events and drains naturally, so that path is left untouched.
- `@abandonware/noble` + `@abandonware/bluetooth-hci-socket` added to the root `onlyBuiltDependencies` so the native binding builds on install.

The transport reuses the existing device-action layer, so commands work over BLE with no per-command changes: discovery, unlock/open-app prompts, rejection handling, and the multi-chain signers are all the same as USB.

## Tests

- Unit tests with a fake `noble` + stubbed DMK framing services: discovery filtering (ignores non-Ledger peripherals), `poweredOn` gating before scan, connect + MTU handshake + APDU round-trip, and idempotent disconnect.
- `typecheck`, `lint` (oxlint), and the full `bun test src/` suite are green.
- Verified live against a Ledger Flex over BLE: `WALLET_CLI_TRANSPORT=ble wallet-cli genuine-check` (device attestation) and `account discover solana` (account derivation) both complete end to end.

## Scope / follow-ups (separate)

Kept intentionally to "transport only" for an easy review. Two related UX gaps are tracked separately rather than bundled here:

- Device selection: like the USB path, this connects to the first discovered device. A `--device <name|id>` selector (and "don't auto-pick when more than one is present") would help when several devices are in range. The transport already surfaces each peripheral's `id` and `localName`, so the data is there.
- `genuine-check` returning `DeviceOnDashboardExpected` instead of navigating to the dashboard via DMK.

## Notes

`@abandonware/noble` has platform-specific native bindings, so a fully bundled standalone binary needs the same native-embed handling that `usb` gets in `embed-usb-native.ts`. This PR wires the transport for the from-source/`bun run` path; standalone packaging of the BLE native bindings can be a follow-up if there is appetite for shipping BLE in the published binaries.

### DIFF
--- a/.changeset/wallet-cli-ble-transport.md
+++ b/.changeset/wallet-cli-ble-transport.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/wallet-cli": minor
+---
+
+Add an optional Bluetooth Low Energy (BLE) transport to wallet-cli, selectable with `WALLET_CLI_TRANSPORT=ble` (default stays `usb`, so existing flows are unchanged). The BLE transport drives Flex/Stax/Nano X over the same DMK device-action layer as USB: framing is delegated to DMK's ApduSender/ApduReceiver services and BLE service/characteristic UUIDs come from `deviceModelDataSource`, so it stays in sync with the device models DMK knows about. This lets the CLI run on hosts where the device is paired over Bluetooth rather than cabled.
--- a/apps/wallet-cli/package.json
+++ b/apps/wallet-cli/package.json
@@ -35,6 +35,7 @@
     "format:check": "oxfmt -c ../../.oxfmtrc.json src bunli.config.ts --check"
   },
   "optionalDependencies": {
+    "@abandonware/noble": "^1.9.2-26",
     "@ledgerhq/wallet-cli-darwin-arm64": "workspace:*",
     "@ledgerhq/wallet-cli-linux-arm64": "workspace:*",
     "@ledgerhq/wallet-cli-linux-x64": "workspace:*",
--- a/apps/wallet-cli/src/cli.ts
+++ b/apps/wallet-cli/src/cli.ts
@@ -11,6 +11,7 @@ import "../.bunli/commands.gen";
 import bunliConfig from "../bunli.config";
 import { getCliProcessExitCode } from "./cli-process-exit-error";
 import { disposeWalletCliDmkTransportFully } from "./device/register-dmk-transport";
+import { resolveWalletCliTransportKind, type WalletCliTransportKind } from "./device/transport-kind";
 import AccountGroup from "./commands/account/index";
 import AssetsGroup from "./commands/assets/index";
 import SessionGroup from "./commands/session/index";
@@ -51,6 +52,17 @@ function normalizeNegatedFlags(argv: string[]): string[] {
 }
 
 if (import.meta.main) {
+  // Resolve (and validate) the transport once, up front: a bad
+  // WALLET_CLI_TRANSPORT fails fast here with a clear message, before any work,
+  // rather than throwing mid-shutdown after a command already ran.
+  let transportKind: WalletCliTransportKind;
+  try {
+    transportKind = resolveWalletCliTransportKind();
+  } catch (e) {
+    process.stderr.write(`${e instanceof Error ? e.message : String(e)}\n`);
+    process.exit(2);
+  }
+
   let exitCode = 0;
   try {
     exitCode = await runMain();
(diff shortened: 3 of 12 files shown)


signerless:
Thanks for the review — all three addressed in the latest push:

1. **noble in devDependencies** — moved to `optionalDependencies` rather than `dependencies`: BLE is opt-in and noble's native binding is platform-specific, so a USB-only install shouldn't fail when that binding can't build. `setupNodeBleTransport()` now catches a missing noble and surfaces a clear, actionable error instead of a raw module-resolution failure.
2. **resolveWalletCliTransportKind() throwing at shutdown** — the transport kind is now resolved and validated once at process start, before any command runs, so an invalid `WALLET_CLI_TRANSPORT` fails fast with a clear message. The shutdown path reuses that value and never throws.
3. **this.scanning set before startScanning() succeeds** — now set only after `startScanning()` returns, and reset in the catch path.

typecheck + oxlint (0 warnings) + the full `bun test src/` suite are green.

Justkant:
Hey thanks for the PR and sorry for not sending a message before this

I've been checking a bit the PR already and we think it could be integrated at some point but we're going to rework this a bit, especially for the options/flags and bundling part to make sure it works well on all platforms

We also want to move the transports back to the DMK repo at some point, so we'll see if we include a new transport in this repo or not

I'll let you know here when we include the changes in the CLI

signerless:
Thanks for the reply, always open to contributing

PR #18442, opened 2026-06-11
feat(wallet-cli): device discovery (devices) and selection (--device / WALLET_CLI_DEVICE)

### LINKED ISSUE #18438: [Feature]: wallet-cli device selection (--device) when multiple Ledgers are reachable
### Library name

@ledgerhq/wallet-cli

### Describe your feature request

wallet-cli connects to the **first discovered device** with no way to pick a specific one. In `connectFirstDevice` (`apps/wallet-cli/src/device/register-dmk-transport.ts`) the code is effectively:

```ts
const discovered = await firstValueFrom(dmk.listenToAvailableDevices({}).pipe(...));
const device = discovered[0]; // first one wins
```

With a single device this is fine, but when more than one Ledger is reachable on the active transport the choice is non-deterministic:

- Over BLE, several powered-on Flex/Stax in range — `discovered[0]` is whichever advertises first.
- Over USB, two cabled devices — the first enumerated wins.

There is no `--device` flag, and the `registerTransportModule` `open` handler already receives a `matchDeviceByName` argument that is currently ignored.

Proposal:

- A `--device <name|id>` option (and/or `WALLET_CLI_DEVICE` env) that filters discovered devices by `localName` or transport `id` before connecting.
- When more than one device is discovered and no selector is given, list the candidates (name + id) and ask the user to choose, instead of silently picking the first.

Both transports already surface each device's `id` and `name` via `TransportDiscoveredDevice`, so the selection data is available. This matters for automation/agent setups where a host may see several devices and must target a specific one deterministically.

Relates to the BLE transport work (#18436): the same first-device behavior applies to both USB and BLE, so a selector is a transport-agnostic improvement.

signerless:
Closes #18438.

> **Stacked on #18437 (BLE transport).** This branch builds on the Node BLE transport so the cross-transport `devices` listing and transport inference cover both USB and BLE. Until #18437 merges, this PR's diff includes the BLE commit — review that one first; once it lands, this diff reduces to just the device-discovery/selection commit.

## What

wallet-cli connected to whichever device was discovered first — non-deterministic when more than one Ledger is reachable (several over BLE, multiple cabled, or one of each), with no way to see what's connected and no way to target one. This adds discovery and selection, with the transport inferred from the chosen device.

## Pieces

- **`devices` command** (read-only) — scans every transport (USB and BLE) in parallel and lists each reachable Ledger with its name, transport, id, and model, without connecting or signing. Transports that can't initialize on the host are skipped; a hint reminds the user to unlock a missing device. Example:
  ```
  Ledger Nano S Plus  usb  7331…  nanoSP
  Solmaria            ble  17cb…  flex
  Missing a device? Unlock it (and open Bluetooth on Flex/Stax), then run again.
  ```
- **`--device <name|id>`** — a flag on every device command (account discover, genuine-check, receive, send, swap execute), plus a global `WALLET_CLI_DEVICE` env (the flag takes precedence). Matches by a **case-insensitive name substring** (so `--device nano` works without quoting a multi-word name) **or a transport-id prefix** (ids can be long; useful for stable BLE ids).
- **Transport inference** — when `WALLET_CLI_TRANSPORT` is not set, the connect path scans all transports, finds where the selector matches, and connects there, so `--device` works **without naming the transport**. With no selector the only reachable device is used; when several are reachable the CLI refuses to guess and lists the candidates (name, transport, id) so the user can disambiguate. `WALLET_CLI_TRANSPORT` still forces a transport when set.
- **Clean exit** — noble keeps a native libuv handle that never unrefs, so any run that initializes BLE (active transport or a cross-transport scan) restores the terminal cursor and exits explicitly after teardown; the pure-USB path keeps its graceful drain.

## Tests

- Unit tests for `resolveDeviceSelector`, `deviceMatchesSelector` (name substring + id prefix), `selectDiscoveredDevice` (single / ambiguous / no-match), and `--device` vs `WALLET_CLI_DEVICE` precedence.
- `typecheck`, `lint` (oxlint, 0 warnings), and the full `bun test src/` suite are green.
- Verified live on a Ledger Flex (BLE) and a Nano S Plus (USB): `devices` lists both with correct model/transport; `--device nano` / `--device sol` / `--device <id-prefix>` each connect to the right device with the transport inferred (no `WALLET_CLI_TRANSPORT`); `account discover` and `send` exercised end to end.

## Note on `--device` flag vs env

`WALLET_CLI_DEVICE` mirrors the existing `WALLET_CLI_TRANSPORT` convention (env, global default); `--device` is the per-command override. Both share one resolution path. Device ids are runtime-generated and not stable across sessions for USB (BLE ids are), so the name substring is the reliable cross-run selector for USB.

### DIFF
--- a/.changeset/wallet-cli-ble-transport.md
+++ b/.changeset/wallet-cli-ble-transport.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/wallet-cli": minor
+---
+
+Add an optional Bluetooth Low Energy (BLE) transport to wallet-cli, selectable with `WALLET_CLI_TRANSPORT=ble` (default stays `usb`, so existing flows are unchanged). The BLE transport drives Flex/Stax/Nano X over the same DMK device-action layer as USB: framing is delegated to DMK's ApduSender/ApduReceiver services and BLE service/characteristic UUIDs come from `deviceModelDataSource`, so it stays in sync with the device models DMK knows about. This lets the CLI run on hosts where the device is paired over Bluetooth rather than cabled.
--- a/.changeset/wallet-cli-device-selection.md
+++ b/.changeset/wallet-cli-device-selection.md
@@ -1,0 +1,9 @@
+---
+"@ledgerhq/wallet-cli": minor
+---
+
+Add device discovery and selection so the CLI no longer silently connects to whichever device was discovered first.
+
+- A read-only `devices` command lists every reachable Ledger across transports (USB and BLE) — name, transport, id, model — without connecting or signing, and hints to unlock a device that is missing.
+- `--device <name|id>` on every device command (account discover, genuine-check, receive, send, swap execute), or globally via `WALLET_CLI_DEVICE` (the flag takes precedence). It matches by a case-insensitive name substring (so `--device nano` works without quoting a multi-word name) or by a transport-id prefix.
+- The transport is inferred from the chosen device, so `--device` works without setting `WALLET_CLI_TRANSPORT`: the CLI scans all transports, finds where the selector matches, and connects there. With no selector it uses the only reachable device, or — when several are reachable — refuses to guess and lists the candidates.
--- a/apps/wallet-cli/package.json
+++ b/apps/wallet-cli/package.json
@@ -35,6 +35,7 @@
     "format:check": "oxfmt -c ../../.oxfmtrc.json src bunli.config.ts --check"
   },
   "optionalDependencies": {
+    "@abandonware/noble": "^1.9.2-26",
     "@ledgerhq/wallet-cli-darwin-arm64": "workspace:*",
     "@ledgerhq/wallet-cli-linux-arm64": "workspace:*",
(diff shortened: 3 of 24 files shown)


signerless:
Thanks for the review — addressed the latest round:

- **Error messages referenced `WALLET_CLI_DEVICE` even when the selector came from `--device`** — they're now selector-neutral (e.g. `No Ledger device matching "<x>". Run \`devices\` to list reachable Ledgers.`) and point at the `devices` command instead of a vague "unique device id".
- **`ListedDevice.name` vs optional `DiscoveredDevice.name`** — the mapping now defaults to `""` so the type stays honest.
- **Transport inference wasn't unit-tested** — extracted the pure decision into `chooseConnectTransport(found, selector)` and added tests for it (infers the match's transport, single-device, refuse-to-guess, no-match, and multi-transport-collision).
- **name matching** is a case-insensitive substring now (so `--device nano` works without quoting a multi-word name), plus transport-id prefix.

typecheck + oxlint (0 warnings) + the full `bun test src/` suite (355) are green.

PR #18485, opened 2026-06-11
fix(web-tools): rework UI with Lumen and fix account sync

gre-ledger:
### ✅ Checklist

- [ ] **Impact of the changes:** internal Web Tools app only (live.ledger.tools), no library/runtime impact.
  - QA focus: home page, /sync, /networkTroubleshoot, /crypto-icons, /trustchain.

### 📝 Description

The Web Tools UI had degraded after the CSS/Lumen migration (no padding, unstyled pages) and account flows on /sync were broken.

This reworks the UI on top of Lumen and fixes the sync support:

- redesign the home landing with Lumen tiles (light theme, centered title)
- drop unused tools: Domain TLV Parser, Derivation Paths, ETH Tx Tools
- restyle Network Troubleshooting, Synchronisation and Crypto Icons with Lumen
- load Tailwind, the Lumen config and the Inter font globally (was only pulled in as a side effect of the trustchain page)
- normalize the trustchain font sizes (base typography + Lumen max-width)
- register all coin modules so /sync has the same coin support as Ledger Live Desktop

Before this, /sync failed with "No coin module registered for family ..." and "No signer registered for network ..." because Web Tools never registered the coin modules. It now calls the same registration as Desktop.

home
tool page

> [!NOTE]
> Two trustchain unit tests fail on develop already (Lumen ESM not transformed by Jest); this PR does not change that and adds no new failures.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-32485
- **ADR link (if any)**: n/a


apps/web-tools/src/pages/sync.tsx
```diff
@@ -118,30 +120,39 @@ function App() {
...
+            <Spinner size={16} /> Synchronising…
+          </span>
+        ) : null}
+        {account ? (
+          <pre className="max-h-[60vh] overflow-auto rounded-lg border border-base bg-muted p-16 body-3 text-base">
+            <code>{JSON.stringify(account, null, 2)}</code>
+          </pre>
+        ) : !accountError && !isLoading ? (
+          <p className="body-2 text-muted">Enter an account id to synchronise.</p>
+        ) : null}
```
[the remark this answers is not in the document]
gre-ledger: it's not a problem becuase we do display the error and yes we should still be able to enter an account id 🤔

apps/web-tools/src/components/ToolPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+type ToolPageProps = {
+  title: string;
+  description?: string;
+  children: ReactNode;
+};
+
+export function ToolPage({ title, description, children }: ToolPageProps) {
+  return (
+    <ThemeProvider colorScheme="light">
```
Sebastien-Dav1d: Why not put the ThemeProvider at root level instead of putting it in every page ?

Sebastien-Dav1d: Why use light everywhere ? Why not system ?

Sebastien-Dav1d (dismissed):

apps/web-tools/src/components/ToolPage.tsx
gre-ledger: fixed

gre-ledger: not all pages are yet uniformized . eg. the repl have its own style etc.
it will take more time to uniformize but also i think a lot of the pages will join devtools effort at some point

apps/web-tools/src/live-common-setup.ts
```diff
@@ -12,6 +13,11 @@ LiveConfig.setAppinfo({
 
 setWalletAPIVersion(WALLET_API_VERSION);
 
+// Register every coin module (and its signer) so the tools have exactly the
+// same coin support as Ledger Live Desktop. Without this, account flows fail
+// with "No coin module registered for family ..." / "No signer registered ...".
+registerAllCoins();
```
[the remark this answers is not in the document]
gre-ledger: zcash is an exception.

Sebastien-Dav1d (approved):

PR #18512, opened 2026-06-12
[LWD] feat(LIVE-31786): support for Aleo token accounts in LWD

mdomanski-ext-ledger:
### ✅ Checklist

  - Aleo token accounts should be handled properly in LWD (tokens are not enabled yet)

### 📝 Description

This PR adds Aleo token account support to the LWD:
- AccountBalanceSummaryFooter now handles TokenAccount in addition to the main Aleo account, rendering the same three balance rows (available, transparent, private) by reading balances from the token account's fields. The repeated JSX for each row was extracted into a local BalanceRow component. Balance formatting was extracted into a shared formatAleoBalances helper in shared/utils.ts.
- BalanceSelector now accepts an optional subAccount prop and reads transparent/private balances from it when present, falling back to the main account's aleoResources. It also picks the correct currency unit from the sub-account when one is passed.
- Both AleoSendStepRecipient and SelfTransferStepRecipient gained token account awareness: when enableTokens is set in currency config they pass withSubAccounts, enforceHideEmptySubAccounts, and a filter to the account selector, and they pass the resolved sub-account down to BalanceSelector. The inline mode-switching logic in onChange was replaced with a call to the new applyAleoBalanceSourceChange helper, eliminating duplicated code between the two modals.
- useAccountChangeGuard replaced its hardcoded TRANSACTION_TYPE constant with a call to derivePublicTransactionMode, passing isTokenTx derived from subAccountId.
- StepSummaryRecipientValue now handles the case where the active account is a TokenAccount: it uses the token's name and icon as the display currency, and resolves
  the self-transfer recipient account correctly even when no address match is found.
- Send/Body.tsx was fixed to derive specific (coin-family overrides) from the main account's currency family rather than the active account's currency, so token
  accounts don't lose their family-specific UI steps. A small parenthesisation fix was also included there.
- Two new shared helpers were added to shared/utils.ts: applyAleoBalanceSourceChange (centralises public/private mode switching for both send and self-transfer) and formatAleoBalances (centralises formatting of the three balance values). Tests were expanded across all touched components.

### ❓ Context

- https://ledgerhq.atlassian.net/browse/LIVE-31786

---

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/modals/Send/Body.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/Send/Body.tsx
@@ -8,7 +8,6 @@ import { Trans, withTranslation } from "react-i18next";
 import { createStructuredSelector } from "reselect";
 import { UserRefusedOnDevice } from "@ledgerhq/errors";
 import { addPendingOperation, getMainAccount } from "@ledgerhq/live-common/account/index";
-import { isCryptoCurrency } from "@ledgerhq/live-common/currencies/helpers";
 import { getAccountCurrency } from "@ledgerhq/live-common/account/helpers";
 import useBridgeTransaction from "@ledgerhq/live-common/bridge/useBridgeTransaction";
 import { useAccountBridge } from "@ledgerhq/live-common/bridge/useAccountBridge";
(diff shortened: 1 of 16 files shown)


apps/ledger-live-desktop/src/renderer/modals/Send/Body.tsx
```diff
@@ -224,8 +224,11 @@ const Body = ({
   const [signed, setSigned] = useState(false);
   const currency = account ? getAccountCurrency(account) : undefined;
   const currencyName = currency ? currency.name : undefined;
+
+  const mainAccount = account ? getMainAccount(account, parentAccount) : null;
+  const mainCurrency = mainAccount ? getAccountCurrency(mainAccount) : undefined;
   const specific =
-    currency && isCryptoCurrency(currency) ? getLLDCoinFamily(currency.family) : null;
```
gre-ledger: please rebase with latest develop. this function no longer exists. `useLLDCoinFamily` will rely on suspense to lazy load the coin family.

gre-ledger (dismissed):

henri-ly (approved):

PR #18561, opened 2026-06-15
[LWDM] feat(market): use dedicated CVS tokenized-stock category for stocks

mcayuelas-ledger:
### ✅ Checklist

      - Market tab → **Stocks** category on Desktop and Mobile (asset discoverability): list now comes from the dedicated CVS `categories=tokenized-stock` param. Verify the right tokenized stocks show and unrelated assets do not.
      - Network: confirm only one `/v3/markets` request fires with the user's counter value (e.g. `to=eur`) — no spurious `to=usd` request while the supported counter-values list loads.
      - Other Market categories (All / Starred / trending) and search are unaffected.

### 📝 Description

Stocks filtering in the Market list was hardcoded: the client sent `filter=stock` to `/v3/markets` and then narrowed the result client-side with substring matching on currency ids/names (`xstock`, `tokenized-stock`, `prestocks`). This was brittle and not exhaustive.

The CVS backend now exposes a dedicated `tokenized-stock` category (via `/v3/categories`) that can be passed as the `categories` param to `/v3/markets`, exactly like trending categories. This PR maps the built-in **Stocks** tab to `categories=tokenized-stock` and removes the client-side id guessing entirely — the backend is now authoritative for the stock list.

While testing, a second issue surfaced: a spurious `to=usd` request was firing on every Market load. The counter-currency resolution fell back to `usd` while the supported counter-values list was still `undefined` (loading), then re-fired with the real currency once it resolved. The fallback now only applies once the list is loaded **and** confirms the currency is unsupported, so an EUR user fires only `to=eur`. The same guard was applied to Mobile (which previously had no supported-currency guard at all).

### ❓ Context

- **JIRA or GitHub link**: [LIVE-32194](https://ledgerhq.atlassian.net/browse/LIVE-32194), [LIVE-29907](https://ledgerhq.atlassian.net/browse/LIVE-29907)

---

[LIVE-32194]: https://ledgerhq.atlassian.net/browse/LIVE-32194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Market/hooks/useMarket.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/Market/hooks/useMarket.ts
@@ -24,9 +24,7 @@ import { addStarredMarketCoins, removeStarredMarketCoins } from "~/renderer/acti
 import { useMarketCategories } from "LLD/features/Market/hooks/useMarketCategories";
 import {
   getMarketCategoriesParam,
-  getMarketFilter,
   isBuiltInMarketListCategory,
-  isStockMarketCurrency,
 } from "@ledgerhq/live-common/market/utils/category";
 
 export function useMarket() {
@@ -68,10 +66,13 @@ export function useMarket() {
 
   const shouldDisplayLiveCompatible = filterBySupported || marketParams.liveCompatible;
 
+  // While the supported list is still loading we keep the user's counter value, and only fall
+  // back to usd once we know it is unsupported — otherwise a request fires with usd first.
+  const isCounterValueUnsupported =
+    !!supportedCounterCurrencies && !supportedCounterCurrencies.includes(settingsCounterValue);
+  const discoverabilityCounterCurrency = isCounterValueUnsupported ? "usd" : settingsCounterValue;
   const effectiveCounterCurrency = shouldDisplayAssetDiscoverability
-    ? supportedCounterCurrencies?.includes(settingsCounterValue)
-      ? settingsCounterValue
-      : "usd"
+    ? discoverabilityCounterCurrency
     : marketParams.counterCurrency;
 
   const resolvedMarketParams = { ...marketParams, counterCurrency: effectiveCounterCurrency };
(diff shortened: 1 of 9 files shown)


LucasWerey (approved):

apps/ledger-live-desktop/src/mvvm/features/Market/hooks/useMarket.ts
```diff
@@ -68,10 +66,13 @@ export function useMarket() {
 
   const shouldDisplayLiveCompatible = filterBySupported || marketParams.liveCompatible;
 
+  // While the supported list is still loading we keep the user's counter value, and only fall
+  // back to usd once we know it is unsupported — otherwise a request fires with usd first.
+  const isCounterValueUnsupported =
+    !!supportedCounterCurrencies && !supportedCounterCurrencies.includes(settingsCounterValue);
+  const discoverabilityCounterCurrency = isCounterValueUnsupported ? "usd" : settingsCounterValue;
```
gre-ledger, line 73: i'm a bit scared by this mecanism that exists in-place just for the Market feature. Shouldn't we fix it at the selector level that if a `settingsCounterValue` is yet not supported that we return a "safe" choice so Portfolio also benefit this?🤔  i'm a bit confused by the usecase / problem met before this was implemented.

gre-ledger (approved):
FYI **platform** team seems to only own this due to `apps/ledger-live-mobile/__tests__/handlers/market.ts` , => this test would deserve to be moved to the mvvm feature that it's testing, don't you think?

(on the other comment, I let you decide how you want to tackle this, because it's not really my code ownership)

apps/ledger-live-desktop/src/mvvm/features/Market/hooks/useMarket.ts
mcayuelas-ledger: Good question — the subtlety is that supportedCounterCurrencies here isn't a global "is this currency supported" signal. It comes from CoinGecko's
  ▎ /simple/supported_vs_currencies (via useMarketDataProvider), so it only describes what the market data backend accepts as vs_currency.

mcayuelas-ledger: also we will fix it this week regarding some data for the graph

PR #18569, opened 2026-06-15
[LWDM] chore: skip `SyncSuccessAllAccounts` tracking event when app was suspended

jnicoulaud-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Applications being suspended during account sync causes huge outliers in our reporting for `SyncSuccessAllAccounts`. 

This pull request implements suspension detection (but comparing expected timer execution time to actual time), and skips tracking when it occurs.

The implementation shares code with existing hang detection in place.

[Related analysis of MixPanel events](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7236911228/2026-06-15+-+Analysis+of+MixPanel+SyncSuccessAllAccounts+outliers)

### ❓ Context

- **JIRA or GitHub link**: 
- **ADR link (if any)**: 

---

### DIFF
--- a/libs/ledger-live-common/src/bridge/react/BridgeSync.tsx
+++ b/libs/ledger-live-common/src/bridge/react/BridgeSync.tsx
@@ -31,6 +31,8 @@ export type Props = {
   hydrateCurrency: (currency: CryptoCurrency) => Promise<any>;
   // an array of token ids to blacklist from the account sync
   blacklistedTokenIds?: string[];
+  // optional platform hook to mark suspension (e.g. React Native AppState background)
+  registerSuspensionListeners?: (markSuspended: () => void) => () => void;
 };
 
 type SyncJob = {
@@ -47,12 +49,14 @@ export const BridgeSync = ({
   prepareCurrency,
   hydrateCurrency,
   blacklistedTokenIds,
+  registerSuspensionListeners,
 }: Props): React.JSX.Element => {
   useHydrate({
     accounts,
     hydrateCurrency,
   });
   const sessionManager = useRef(createSyncSessionManager(trackAnalytics)).current;
+  useSuspensionPlatformListeners(sessionManager.markSuspended, registerSuspensionListeners);
   const [syncQueue, syncState] = useSyncQueue({
     accounts,
     prepareCurrency,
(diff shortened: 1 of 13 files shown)


libs/ledger-live-common/src/bridge/react/BridgeSync.tsx
```diff
@@ -114,6 +118,37 @@ export function resetStates() {
...
+
+  useEffect(() => {
+    const onVisibilityChange = () => {
+      if (typeof document !== "undefined" && document.hidden) {
+        markSuspendedRef.current();
+      }
+    };
+
+    if (typeof document !== "undefined") {
+      document.addEventListener("visibilitychange", onVisibilityChange);
```
gre-ledger, line 138: A few questions and a design concern on this approach:
  
Is this meant to cover Desktop as well? And if so, why is it needed there?

My understanding is that the suspension detection makes sense on React Native because the OS literally pauses the CPU when the app goes to background. and this is why on Mixpanel you observe crazy timing. But I think that's not true on Desktop: the app keeps running when it's unfocused. So I'd expect a lot of false positives there (we'd flag a "suspension" while the app is actually still running fine).

Also, I'd rather not have Desktop-specific logic living in live-common. More broadly, I would prefer not adding this much surface to the already-complex BridgeSync mechanism : the registerSuspensionListeners hook pushes a lot of handling down into live-common + live-promise.

I think there's a simpler interface to be found here: from the consumer side, we could just provide a `shouldSendTrackingEvent` predicate that BridgeSync can use to short-circuit, rather than registering listeners that live-common then has to orchestrate.

Concretely, I'd prefer we move syncSessionManager / driftWatchdog back into mobile (and Desktop too, if we genuinely want it covered there) and keep BridgeSync's interface minimal.

A secondary reason: the promise lib we'd be leaning on here is somewhat considered legacy/stable at this point : it's soon moving out of the repo : https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7019298849/Monorepo+Refactoring+Data+Architecture+steps

gre-ledger (changes requested):
some feedback on the arch design but happy to discuss irl

gre-ledger (changes requested):
After a quick analysis, this PR may fail the `oxfmt` format check, which we just made mandatory on `develop` (it was previously silent in CI).

Please run `pnpm format` on your branch before merging. Preferably rebase on latest `develop` first, as we also fixed some ongoing formatting drift in shared project files.

Also, please double-check that `mise` is set up on your side — the pre-commit hook should have caught this. If you're using an LLM to commit code, make sure it's not bypassing hooks with `--no-verify`.

I'm requesting changes to avoid an accidental merge — please re-request my review once fixed.

jnicoulaud-ledger:
Closing, implementation is not good

PR #18618, opened 2026-06-16
[LWD] fix(desktop): revert market banner favorites filter when last starred coin is removed

deepyjr:
### ✅ Checklist

      - Portfolio Market Banner: with the filter on "Favorites" and at least one starred coin, remove all favorites → the banner must revert to "Trending" instead of showing an empty favorites view (including after navigating away and back to the portfolio).
      - The ranking dropdown should reflect "Trending" after the reset.
      - Regular filter switching (Trending / Gainers / Losers / Favorites) still works and persists.
      - Behaviour with the asset discoverability flag both ON and OFF.

### 📝 Description

**Problem**: When the portfolio Market Banner filter is set to "Favorites" and the user removes their last starred coin, the banner stays stuck on an empty "Favorites" view. The ranking is persisted to disk, so the stale `"favorites"` value survives navigation (and app restarts) and the banner never recovers on its own.

**Solution**: `useMarketBannerViewModel` now resets the persisted ranking to `"trending"` when `"favorites"` is active but no coin is starred, and gates `isFavorites` on `hasStarred` so the trending performers render immediately (no blank banner while the reset propagates). The ranking dropdown reads the same persisted value, so it reflects "Trending" automatically. This mirrors the fix already shipped on mobile and follows the existing `useResetTimeRangeOnGraphRework` precedent on desktop.

### Before / After

https://github.com/user-attachments/assets/fd58656d-8d10-462b-95a0-31208364300c

### ❓ Context

- **JIRA or GitHub link**: [LIVE-32628](https://ledgerhq.atlassian.net/browse/LIVE-32628)

---

[LIVE-32628]: https://ledgerhq.atlassian.net/browse/LIVE-32628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/MarketBanner/hooks/useMarketBannerViewModel.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/MarketBanner/hooks/useMarketBannerViewModel.ts
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from "react";
+import { useCallback, useEffect, useMemo } from "react";
 import { useMarketPerformers } from "@ledgerhq/live-common/market/hooks/useMarketPerformers";
 import { useMarketData } from "@ledgerhq/live-common/market/hooks/useMarketDataProvider";
 import { filterMarketPerformersByAvailability } from "@ledgerhq/live-common/market/utils/index";
@@ -16,13 +16,14 @@ import {
   MARKET_PERFORMERS_SUPPORTED,
   MARKET_BANNER_REFRESH_RATE,
 } from "@ledgerhq/live-common/market/constants";
-import { useSelector } from "LLD/hooks/redux";
+import { useDispatch, useSelector } from "LLD/hooks/redux";
 import {
   counterValueCurrencySelector,
   starredMarketCoinsSelector,
 } from "~/renderer/reducers/settings";
 import {
   selectMarketBannerRanking,
+  setMarketBannerRanking,
   type MarketBannerRanking,
 } from "~/renderer/reducers/marketBanner";
 import { useWalletFeaturesConfig } from "@features/platform-feature-flags";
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/mvvm/features/MarketBanner/hooks/useMarketBannerViewModel.ts
```diff
@@ -139,12 +140,23 @@ export function useFavoritesBannerItems(options?: { skip?: boolean }): MarketBan
...
+
+  // Self-heal a stale persisted "favorites" ranking once the last starred coin is removed,
+  // otherwise the banner would stay stuck on an empty favorites view.
+  useEffect(() => {
+    if (ranking === "favorites" && !hasStarred) {
+      dispatch(setMarketBannerRanking("trending"));
+    }
+  }, [dispatch, ranking, hasStarred]);
+
+  const isFavorites = ranking === "favorites" && hasStarred;
```
mcayuelas-ledger: Any possibility to do it without effect?

apps/ledger-live-desktop/src/mvvm/features/MarketBanner/hooks/useMarketBannerViewModel.ts
```diff
@@ -139,12 +140,23 @@ export function useFavoritesBannerItems(options?: { skip?: boolean }): MarketBan
...
 
   const performers = usePerformersBannerItems(ranking, { skip: isFavorites });
   const favorites = useFavoritesBannerItems({ skip: !isFavorites });
```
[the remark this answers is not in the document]
claudiiafg: this seems like a good point, wdyt?

apps/ledger-live-desktop/src/mvvm/features/MarketBanner/hooks/useMarketBannerViewModel.ts
deepyjr: I've searched, and it seems that we don't have another "easy" option for that. Do you have an idea?

 What I like about the useEffect is that it resets the state from any path.

claudiiafg (approved):

PR #18628, opened 2026-06-16
feat(currencies): add @domain/api-currencies cvs + supported-fiats slice

gre-ledger:
### ✅ Checklist

- [ ] **Impact of the changes:** New isolated packages, not wired into LLD/LLM. Only the `web-tools` devtools host consumes them.
  - QA focus: the **Currencies** section of the `web-tools` dev-tools page (lists supported fiats from the Countervalues Service, with a Refetch button).

### 📝 Description

Start of the DDD evolution of [`libs/ledger-live-common/src/currencies/support.ts`](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/currencies/support.ts#L82) into an RTK Query model, following the new domain/feature package strategy.

**New packages**
- **`@domain/api-currencies`** (`domain/api/currencies/src/cvs/`) — the Countervalues Service sub-part:
  - `schema.ts` — Zod schema for the `/v3/supported/fiat` response (array of tickers)
  - `api.ts` — RTK Query `createApi` (`cvsApi`) with the `getSupportedFiats` query
  - `internals.ts` — OFAC currencies `Set` + filtering
  - `utils.ts` — `resolveSupportedFiats(tickers) → FiatCurrency[]` (resolve + OFAC filter + dedupe)
- **`@features/platform-currencies`** (`features/platform/currencies/`) — the `supportedFiats` redux slice, populated via `extraReducers` on `cvsApi.endpoints.getSupportedFiats.matchFulfilled`, storing the resolved, OFAC-filtered `FiatCurrency[]`; with selectors.
- **`@devtools/currencies`** — a devtools tool to inspect the supported fiats and trigger refetch.

**Changed**
- **`@domain/entity-currency-fiat`** — enriched with a by-ticker lookup (`FIAT_CURRENCIES_BY_TICKER` + `getFiatCurrencyByTicker`), since the CVS returns tickers while the registry is keyed by id. Uses the new Zod-first `FiatCurrency` entity (no `@ledgerhq/types-cryptoassets`).
- **`@devtools/registry`** — registered the `currencies` tool + `DevToolConfig` union member.
- **`apps/web-tools`** — store wires the `cvsApi` reducer + middleware and the slice; `useCurrenciesToolProps` hook; tool added to the dev-tools page.

**Showcase — devtools integration in `web-tools`:**

Image

#### Intentional, scoped divergences vs legacy `support.ts`
This is the *start* of the migration and is **not connected into the apps yet** — so the following are deliberately deferred (tracked under the epic / supported-set ADR, not regressions in this PR):
- **No local fallback list** when the CVS call fails or returns `[]` (legacy fell back to a 36-ticker hardcoded list). The slice stays empty on failure for now.
- **Registry-bounded resolution**: the new `@domain/entity-currency-fiat` registry currently has 20 fiats, so tickers it doesn't know yet are dropped. The registry is being onboarded incrementally.
- **OFAC list is duplicated** (not shared with `support.ts`) on purpose — the domain package must not depend on `@ledgerhq/live-common`. A single source of truth can be revisited once `support.ts` is retired.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-32269](https://ledgerhq.atlassian.net/browse/LIVE-32269) (epic [LIVE-32342](https://ledgerhq.atlassian.net/browse/LIVE-32342))
- **ADR link (if any)**: supported-set ADR — LIVE-31919 (pending)

[LIVE-32269]: https://ledgerhq.atlassian.net/browse/LIVE-32269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre-ledger:
## Claude analysis of next steps

Study of what it takes to migrate `apps/ledger-live-desktop` (LLD) and `apps/ledger-live-mobile` (LLM) onto this new currencies stack. **TL;DR: this PR ships the right foundation, but it is not enough on its own — app migration is _medium_ effort with real blockers, so it should not be wired into the apps yet.**

### How the apps consume supported fiats today
Both apps do essentially the same thing:
- **LLD** — `renderer/reducers/settings.ts → getsupportedCountervalues()` calls `listSupportedFiats()`, **appends BTC + ETH** as intermediaries, maps to `SupportedCountervaluesData = { value, label, currency: Currency }`, stores in `settings.supportedCounterValues`. Triggered fire-and-forget by `useInitSupportedCounterValues` (no await). Consumed by `CounterValueSelect` (Settings) + `CountervalueSelect` (Market). Not persisted. RTK Query reducers are spread via `lldRTKApiReducers`.
- **LLM** — `context/LedgerStore.tsx → updateSupportedCountervalues()` (with retry), same `{ value, ticker, label, currency }` shape + BTC/ETH, stored in `settings.supportedCounterValues`, persisted to MMKV via `DBSave`. RTK Query is already first-class (`llmRTKApi`, `applyLlmRTKApiMiddlewares`, `...llmRTKApiReducers`). Consumed by `CountervalueSettings` + Market.

### What is easy ✅
- **Store wiring** in both apps is an established pattern: register `cvsApi` reducer + middleware and the `supportedFiats` slice (LLM concatenates RTK Query middleware automatically).
- **`setCvsBaseUrl(getEnv("LEDGER_COUNTERVALUES_API"))`** at startup — trivial; both apps already depend on `live-env`.
- **Dispatching the fetch** at init (`getSupportedFiats.initiate()`); the "empty until loaded" behavior matches today's fire-and-forget model.
- **Countervalues API** only needs the `ticker`, which the new entity has.

### What is missing / blockers ❌
1. **Type is not a drop-in.** Apps store `currency: Currency` (legacy `FiatCurrency` from `@ledgerhq/types-cryptoassets`). Legacy→new is **not assignable** (legacy has no `id`, has `delisted`). New→legacy *is* mostly assignable, so a `new→legacy` adapter at the slice boundary is the realistic bridge — but that keeps legacy types in the apps (partial migration), or means migrating ~35 consumer files to the new entity.
2. **The list mixes crypto intermediaries (BTC/ETH).** `SupportedCountervaluesData` is not fiat-only. This stack is fiat-only; BTC/ETH still need to come from `@domain/entity-currency-crypto` (not wired here), so the app-facing list cannot be produced from this slice alone.
3. **Registry is incomplete → user-visible regression.** `@domain/entity-currency-fiat` has **20 fiats** vs legacy **36** (local fallback) / **~150** (cryptoassets). Switching now would drop selectable fiats (PLN, THB, TRY, AED…). Gated by the supported-set ADR (LIVE-31919) + registry onboarding (LIVE-31915 / LIVE-31916).
4. **No fallback on empty/failed CVS.** Legacy degrades to a 36-fiat local list; the new slice goes empty. Production parity requires a fallback (or equivalent) before any app switch.

### Suggested sequencing before app adoption
1. Complete the fiat registry to ≥ legacy parity + land the supported-set ADR (LIVE-31919).
2. Define the empty/failed-fetch fallback policy in `@features/platform-currencies`.
3. Provide a crypto-intermediaries source (BTC/ETH) for the countervalues picker.
4. Add a `new → legacy Currency` adapter (or migrate consumers) so `supportedCounterValues` keeps working.
5. Wire `cvsApi` + `supportedFiats` into LLD/LLM stores, replace `listSupportedFiats()` call sites, `setCvsBaseUrl` from env, dispatch the query at init.

### Verdict
This PR is the correct **foundation** (API + slice + entity lookup + a working devtools harness). It is intentionally **not** connected to the apps, and it should stay that way until items 1–4 above are addressed.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

devtools/currencies/src/types.ts
```diff
@@ -0,0 +1,13 @@
+import type { FiatCurrency } from "@domain/entity-currency-fiat";
+
+/** Props the host wires from the `supportedFiats` slice and the CVS query. */
+export interface CurrenciesToolProps {
+  /** Resolved, OFAC-filtered supported fiats. */
+  supportedFiats: FiatCurrency[];
+  /** Whether the supported-fiats query is currently in flight. */
+  isFetching: boolean;
+  /** Human-readable error message, when the query failed. */
+  error?: string;
```
ysitbon: Please use finite states instead of `isFetching`, `error`:

```ts
interface CurrenciesToolProps {
  ...
  status: RemoteFiatStatus
  ...
}

type RemoteFiatStatus = 
  | { type: "idle" }
  | { type: "fetching" } 
  | { type: "error", message: string }
```

features/platform/currencies/src/data/slice.ts
```diff
@@ -0,0 +1,17 @@
...
+const supportedFiatsSlice = createSlice({
+  name: "supportedFiats",
+  initialState: SUPPORTED_FIATS_INITIAL_STATE,
+  reducers: {},
+  extraReducers: builder => {
+    builder.addMatcher(cvsApi.endpoints.getSupportedFiats.matchFulfilled, (state, action) => {
+      state.supportedFiats = resolveSupportedFiats(action.payload);
+    });
+  },
+});
```
ysitbon: The entity should not know the api. The entity should stay pure:

```ts
const slice = createSlice({
  name: "supportedFiats",
  initialState: { list: [] as FiatCurrency[] },
  reducers: {
    setSupportedFiats(s, a: PayloadAction<FiatCurrency[]>) { s.list = a.payload; },
  },
});
```

The binding should be done on api side:

```ts
export const cvsApi = createApi({
  /* … */ 
  endpoints: build => ({
    getSupportedFiats: build.query<SupportedFiatTickers, void>({
      query: () => {{ url: "/v3/supported/fiat", /* … */ }),
      transformResponse: (raw: unknown) => SupportedFiatsResponseSchema.parse(raw),
      async onQueryStarted(_, { dispatch, queryFulfilled }) {      
        const { data } = await queryFulfilled;
        dispatch(setSupportedFiats(resolveSupportedFiats(data)));
      },
    }),
  }),
});
```

domain/api/currencies/src/cvs/config.ts
```diff
@@ -0,0 +1,17 @@
...
+ * stays free of any env/config dependency.
+ */
+export function setCvsBaseUrl(url: string): void {
+  cvsBaseUrl = url;
+}
+
+/** Current Countervalues Service base URL. */
+export function getCvsBaseUrl(): string {
+  return cvsBaseUrl;
+}
```
ysitbon: As we started with feature flags we should avoid imperative global config and we should provide a middleware-based solution:

```ts
function cvsBaseQuery(args, api, extraOptions) {
  return fetchBaseQuery({ baseUrl: (api.extra as CvsThunkExtra).cvsBaseUrl })(args, api, extraOptions);
} satisfies BaseQueryFn;

export const cvsApi = createApi({
  baseQuery: cvsBaseQuery,
  /* … */ 
})

export const cvsApiExtra = (baseUrl: string): CvsApiExtra => ({ cvsBaseUrl: baseUrl });
```
Using[custom default middleware config](https://redux-toolkit.js.org/api/getDefaultMiddleware#customizing-the-included-middleware), the app supplies the environment and the package owns the contract. 
```ts
// store
configureStore({
  reducer: { [cvsApi.reducerPath]: cvsApi.reducer, /* ... */ },
  middleware: gdm => gdm({ thunk: { extraArgument: cvsApiExtra(getEnv("LEDGER_COUNTERVALUES_API")) } })
    .concat(cvsApi.middleware),
});
```
Side note: we should found a mechanism for custom default middleware config to be able to chain multiple configs from different domain apis, probably an helper from shared packages.

ysitbon (commented):
`@features/platform-currencies` should not be created here. The actual content defined in this package should move to the existing `@domain/entity-currency-fiat` and indeed adjusted accordingly

features/platform/currencies/src/data/schema.ts
```diff
@@ -0,0 +1,15 @@
+import type { FiatCurrency } from "@domain/entity-currency-fiat";
+
+/** Redux state holding the supported fiats resolved from the Countervalues Service. */
+export interface SupportedFiatsState {
+  /** Resolved, OFAC-filtered list of supported fiat currencies. */
+  supportedFiats: FiatCurrency[];
+}
+
+/** Shape of any store state that includes the `supportedFiats` slice. */
+export type WithSupportedFiats = { supportedFiats: SupportedFiatsState };
```
ysitbon, line 10: This part makes structural access redundant

```ts
s.supportedFiats.supportedFiats
```

I'd prefer something more like this:

```ts
export interface SupportedFiatsState {
  currencies: FiatCurrency[];
}

export type WithSupportedFiats = { supportedFiats: SupportedFiatsState };
```

Which makes things clear enough:

```ts
s.supportedFiats.currencies
```

devtools/currencies/src/types.ts
gre-ledger: Done in 6b42e9bce2d — `CurrenciesToolProps` now exposes `status: RemoteFiatStatus = { type: "idle" } | { type: "fetching" } | { type: "error"; message: string }` instead of `isFetching`/`error`, and the component renders off the discriminant.

features/platform/currencies/src/data/slice.ts
gre-ledger: Done in 6b42e9bce2d. The slice is now pure (`setSupportedFiats(PayloadAction<FiatCurrency[]>)`, no `cvsApi` import). One nuance on "binding on the api side": the enforce-boundaries rule forbids `type:domain-api` from depending on a feature package, so `@domain/api-currencies` can't import the slice action — putting `onQueryStarted` literally inside `cvsApi` would be a circular/illegal dependency. So the binding lives in the **feature** layer (`features/platform/currencies/src/data/api.ts`) via `cvsApi.enhanceEndpoints({ endpoints: { getSupportedFiats: { onQueryStarted } } })`, which dispatches `setSupportedFiats(resolveSupportedFiats(data))`. Same effect, slice stays pure, deps stay features→domain. Consumers import `useGetSupportedFiatsQuery` from `@features/platform-currencies` so the binding runs.

domain/api/currencies/src/cvs/config.ts
gre-ledger: Done in 6b42e9bce2d — removed the global `setCvsBaseUrl`/`config.ts`. `cvsApi` now uses a custom `baseQuery` reading `(api.extra as CvsApiExtra).cvsBaseUrl`, with a `cvsApiExtra(baseUrl)` helper; web-tools supplies it via `getDefaultMiddleware({ thunk: { extraArgument: cvsApiExtra(getEnv("LEDGER_COUNTERVALUES_API")) } })`. Deferring the shared cross-domain-api extraArgument chaining helper to a follow-up as you suggested.

features/platform/currencies/src/data/schema.ts
gre-ledger: Done in 6b42e9bce2d — renamed to `SupportedFiatsState { currencies: FiatCurrency[] }`, so access is `state.supportedFiats.currencies`.

PR #18631, opened 2026-06-16
[LWDM] baby staking craft transaction

amaslakov:
### ✅ Checklist

  - Refactors the **shared** staking-message build path (`txToMessages`) used by **every** cosmos-family chain. QA should regression-check delegate / undelegate / redelegate / claim-reward / claim-compound on **Cosmos and Zenrock** — output is expected to be unchanged (bare for cosmos, `zrchain` relabel for zenrock).
  - **Babylon delegation stays disabled in config** (the enable is LIVE-31837, not in this PR), so there is **no user-facing Babylon change yet**; the wrapped-build path is exercised by unit tests only (incl. a byte-for-byte check against a real mainnet tx).
  - `claimRewardCompound` now **throws** on Babylon by design — its embedded delegate isn't epoching-wrapped yet (LIVE-31837).

### 📝 Description

Babylon's x/epoching module rejects bare cosmos-sdk staking messages: delegate/undelegate/redelegate must be wrapped in a `/babylon.epoching.v1.MsgWrapped*` envelope, with the inner staking msg nested under `msg` (amino + proto). This PR builds that envelope for Babylon — the send-side counterpart to the read-side positions work (LIVE-31279).

Chain-specific message types (cosmos-sdk vs `zrchain` vs epoching, and whether they're wrapped) now live as data on the chain classes (`COSMOS_/ZENROCK_/BABYLON_STAKING_MESSAGES`) instead of inline `currency.id` branches.

**Amino type:** the wrapped msgs use the proto type URL as the amino `type`, not a short `epoching/*` name — the protos set no `(amino.name)`, so amino-JSON falls back to the URL. Matches [Keplr](https://github.com/chainapsis/keplr-wallet/blob/master/packages/stores/src/account/babylon.ts) and Babylon's own SDK [`babylon-proto-ts`](https://github.com/babylonlabs-io/babylon-toolkit/blob/main/packages/babylon-proto-ts/src/lib/utils/amino.ts)

### ❓ Context

- **JIRA or GitHub link**: [LIVE-31280](https://ledgerhq.atlassian.net/browse/LIVE-31280?atlOrigin=eyJpIjoiNDUwNzBmZWE3MGMyNDUxNGFhMDg0Zjc4MDkyYWI1NWQiLCJwIjoiaiJ9)
- **ADR link (if any)**: 

---

[LIVE-31280]: https://ledgerhq.atlassian.net/browse/LIVE-31280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-modules/coin-cosmos/src/buildTransaction.unit.test.ts
+++ b/libs/coin-modules/coin-cosmos/src/buildTransaction.unit.test.ts
@@ -1,3 +1,11 @@
+import { AminoMsg, makeSignDoc, serializeSignDoc } from "@cosmjs/amino";
+import { ExtendedSecp256k1Signature, Secp256k1, sha256 } from "@cosmjs/crypto";
+import {
+  MsgWrappedBeginRedelegate,
+  MsgWrappedDelegate,
+  MsgWrappedUndelegate,
+  protobufPackage as epochingPackage,
+} from "@keplr-wallet/proto-types/babylon/epoching/v1/tx";
 import { Fee } from "@keplr-wallet/proto-types/cosmos/tx/v1beta1/tx";
 import BigNumber from "bignumber.js";
 import { MsgSend } from "cosmjs-types/cosmos/bank/v1beta1/tx";
@@ -10,9 +18,13 @@ import {
 import { TxBody, TxRaw } from "cosmjs-types/cosmos/tx/v1beta1/tx";
 
 import { buildTransaction, txToMessages } from "./buildTransaction";
+import Babylon, { BABYLON_STAKING_MESSAGES } from "./chain/Babylon";
+import Cosmos from "./chain/Cosmos";
+import Zenrock from "./chain/Zenrock";
 import { CosmosAccount, CosmosDelegationInfo, Transaction } from "./types";
 
 const veryBigNumber = new BigNumber(3333300000000000000000);
+const cosmos = new Cosmos();
 
 describe("txToMessages", () => {
   const transaction: Transaction = {} as Transaction;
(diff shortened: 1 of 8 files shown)


libs/coin-modules/coin-cosmos/src/buildTransaction.unit.test.ts
```diff
@@ -10,9 +18,13 @@ import {
...
 import { buildTransaction, txToMessages } from "./buildTransaction";
+import Babylon, { BABYLON_STAKING_MESSAGES } from "./chain/Babylon";
+import Cosmos from "./chain/Cosmos";
+import Zenrock from "./chain/Zenrock";
 import { CosmosAccount, CosmosDelegationInfo, Transaction } from "./types";
 
 const veryBigNumber = new BigNumber(3333300000000000000000);
+const cosmos = new Cosmos();
 
 describe("txToMessages", () => {
```
hedi-edelbloute, line 29: Should we check that zenrock, babylon, etc messages are right ones here ? and differ from cosmos base

hedi-edelbloute (dismissed):

libs/coin-modules/coin-cosmos/src/buildTransaction.unit.test.ts
amaslakov: Good point — that top block only drives the cosmos chain. The "correct types, and different from cosmos base" assertions live in dedicated per-chain blocks further down:
- babylon — txToMessages — babylon epoching wrapping: all three staking modes assert the /babylon.epoching.v1.MsgWrapped* amino type + proto typeUrl with the inner msg nested under msg; plus uses the canonical x/epoching proto type URLs… anchors them to the keplr package, and a real mainnet tx signature is recovered to prove the amino type is correct.
- zenrock — txToMessages — zenrock relabels staking msgs without wrapping: asserts the zrchain/* types, bare (non-wrapped).
- cosmos — txToMessages — the cosmos chain uses standard cosmos-sdk staking msgs (no wrapping), plus the whole main txToMessages block. zenrock was only covering delegate, so I've just added undelegate and redelegate cases to match. Thanks!

hedi-edelbloute (approved):

PR #18653, opened 2026-06-17
[LWM] feat(lwm): add confirmation step to new send flow

YazhuEth:
### ✅ Checklist

  - Mobile new send flow: success step shown after broadcast ("Transaction signed")
  - "View transaction" opens the operation details with the right account/operation
  - "Close" exits the send flow
  - Error step shown when the broadcast fails after signing (instead of a hardcoded success), with "Save logs" + "Try again"

### 📝 Description

After the signature step, the new send flow broadcasts the transaction and must show a confirmation. On mobile the confirmation step was a placeholder, so a broadcast failure after signing would still have shown a hardcoded success.

This PR implements the confirmation step on mobile. It reads the flow status directly from the send context (`state.flowStatus`, already set by the signature core) — no re-derivation — and renders accordingly, all on the same page:

- **SUCCESS** → full-screen "Transaction signed" with a green status glow, "View transaction" + "Close".
- **ERROR** (broadcast failure after signing) → full-screen error with a red status glow, the translated reason, "Save logs", "Try again" + "Close".

Cancellation / device errors keep being handled inside the signature (DIE) sheet and never reach this screen, so the confirmation only deals with the broadcast outcome (success or failure).

Implementation (mobile only — desktop and live-common are untouched):
- `ConfirmationStatusLayout` — shared dumb shell (top status glow + spot + title/description + bottom actions) used by both states.
- `ConfirmationScreenView` (success) and `ConfirmationErrorView` (error) — dumb views.
- `useConfirmationViewModel` — exposes the flow status, the concerned operation, and the view/retry/save-logs/close handlers.

### 📸 Screenshots

| Success | Error |
| ------ | ----- |
|image|image|

### ❓ Context

- **JIRA or GitHub link**: [LIVE-22993](https://ledgerhq.atlassian.net/browse/LIVE-22993)

---

[LIVE-22993]: https://ledgerhq.atlassian.net/browse/LIVE-22993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/features/Send/screens/Confirmation/hooks/useConfirmationViewModel.ts
+++ b/apps/ledger-live-mobile/src/mvvm/features/Send/screens/Confirmation/hooks/useConfirmationViewModel.ts
@@ -1,0 +1,46 @@
+import { useCallback } from "react";
+import { useNavigation } from "@react-navigation/native";
+import { ScreenName } from "~/const";
+import type { BaseNavigationComposite } from "~/components/RootNavigator/types/helpers";
+import useExportLogs from "~/components/useExportLogs";
+import { useSendFlowActions, useSendFlowData } from "../../../context/SendFlowContext";
+import type { SendFlowNavigationProp } from "../../../types";
+
+export function useConfirmationViewModel() {
+  const navigation = useNavigation<BaseNavigationComposite<SendFlowNavigationProp>>();
+  const { close, status: statusActions, operation } = useSendFlowActions();
+  const { state } = useSendFlowData();
+  const { account, parentAccount } = state.account;
+  const onSaveLogs = useExportLogs();
+
+  const optimisticOperation = state.operation.optimisticOperation;
+  const concernedOperation =
+    optimisticOperation?.subOperations?.find(op => op.accountId === account?.id) ??
+    optimisticOperation ??
+    null;
+
+  const onViewTransaction = useCallback(() => {
+    if (!account || !concernedOperation) return;
+    navigation.navigate(ScreenName.OperationDetails, {
+      accountId: account.id,
+      parentId: parentAccount?.id ?? undefined,
+      operation: concernedOperation,
+    });
+  }, [account, parentAccount, concernedOperation, navigation]);
+
+  const onRetry = useCallback(() => {
+    navigation.replace(ScreenName.SendFlowSignature);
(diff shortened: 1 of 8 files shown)


YazhuEth:
**E2E**
- Desktop : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/4b883ecd-639f-4252-91de-83108a40e881/
- Mobile: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/03f5550e-5ec4-4852-9686-fdf58e23c77a/

francois-guerin-ledger (dismissed):

libs/ledger-live-common/src/flows/send/confirmation/useSendFlowConfirmationCore.ts
dilaouid, line 1: Is it worth putting it in common ? I don't think it's used in LWM ?

apps/ledger-live-mobile/src/mvvm/features/Send/screens/Confirmation/hooks/useConfirmationViewModel.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!account || !concernedOperation) return;
+    navigation.navigate(ScreenName.OperationDetails, {
+      accountId: account.id,
+      parentId: parentAccount?.id ?? undefined,
+      operation: concernedOperation,
+    });
+  }, [account, parentAccount, concernedOperation, navigation]);
+
+  return {
+    canViewTransaction: Boolean(account && concernedOperation),
```
dilaouid, line 40: This is a bit dangerous !
Indeed we can trust the DIE for the failure / cancellation, but if an error happens after the signature or because of a broadcast failure and we pass the DIE, we might still want to see an error page and not a hardcoded success page

It's not likely to happens, but I know too less of the DIE to be confident at 100% on it 😔
(i know I said something else before, but better safe than sorry)

dilaouid:
I think we need more margins between the two buttons
| Figma | Your screen |
| ------ | ----- |
|image|image |

The spacing should be 16 and not 12

dilaouid (approved):

PR #18693, opened 2026-06-18
[LWDM] feat(LIVE-30909): aleo private tokens sync part 3

mateuszpalosz-ext:
### ✅ Checklist

  - Aleo token private sync

### 📝 Description

Base logic for Aleo tokens private sync. Part 3

### ❓ Context

- [https://ledgerhq.atlassian.net/browse/LIVE-30909](https://ledgerhq.atlassian.net/browse/LIVE-30909)

---

### DIFF
--- a/libs/coin-modules/coin-aleo/src/bridge/sync.ts
+++ b/libs/coin-modules/coin-aleo/src/bridge/sync.ts
@@ -9,6 +9,7 @@ import { encodeAccountId } from "@ledgerhq/ledger-wallet-framework/account/accou
 import { log } from "@ledgerhq/logs";
 import { concat, merge, Observable, of } from "rxjs";
 import { concatMap } from "rxjs/operators";
+import type { TokenCurrency } from "@ledgerhq/types-cryptoassets";
 import { SyncConfig, SYNC_TYPE_SHIELDED, SYNC_TYPE_TRANSPARENT } from "@ledgerhq/types-live";
 import type { TokenAccount } from "@ledgerhq/types-live";
 import invariant from "invariant";
@@ -20,6 +21,7 @@ import {
   isRecordScannerReady,
   splitPrivateAndPublicOperations,
   resolveConfig,
+  getCalTokens,
 } from "../logic/utils";
 import { aleoPrivateSyncProgress$ } from "./privateSyncProgress";
 import { accessProvableApi, fetchAllOwnedRecords, patchPublicOperations } from "../network/utils";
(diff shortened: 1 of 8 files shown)


libs/coin-modules/coin-aleo/src/bridge/tokens.ts
```diff
@@ -370,3 +394,465 @@ export async function resolveTokenSubAccounts({
...
+    if (isOutgoingRecord) {
+      // OUT: read the actual sent amount and recipient address from the transition inputs.
+      const outDetails = await getTokenOutDetails({ currency, record, viewKey });
+      if (outDetails.amount === null) {
+        log(
+          "aleo/buildSubAccountsFromPrivateRecords",
+          `Could not determine OUT amount for record ${record.commitment} (tx ${record.transaction_id}), falling back to 0`,
+        );
+      }
+      amount = outDetails.amount ?? new BigNumber(0);
```
henri-ly, line 786: I don't know if it's something that we should flag to the user that amount wasn't fetch properly.

libs/coin-modules/coin-aleo/src/bridge/sync.ts
```diff
@@ -106,9 +115,20 @@ export async function performPublicSync(
   latestAccountPublicOperations.operations.sort((a, b) => b.date.getTime() - a.date.getTime());
   latestAccountPublicOperations.tokenOperations.sort((a, b) => b.date.getTime() - a.date.getTime());
 
+  // Already-patched ops have modified senders/recipients that differ from raw API data.
+  // Filter them from the incoming ops — mergeOps then simply keeps the patched version
+  // from oldPublicOps untouched, and no patch-restoration pass is needed.
+  const patchedOpIds = new Set(
+    (oldPublicOps as AleoOperation[]).filter(op => op.extra?.patched).map(op => op.id),
```
henri-ly, line 122: Is there a scenario where patched op can't be refresh ? I mean if it's has been patched and somehow data need to be refresh a 2nd time, will we missed it ?

I guess it's a super edge and it can be ignore if you don't find my comment relevant.

libs/coin-modules/coin-aleo/src/bridge/sync.ts
```diff
@@ -276,20 +320,62 @@ export async function performPrivateSync(
       unspent: true,
       ...(signal && { signal }),
     }),
+    config.enableTokens
+      ? fetchAllOwnedRecords({
+          currency,
+          uuid: provableApi.uuid,
+          start: tokenSyncStartHeight,
+          // empty arrays opt out of the credits.aleo-only filter, returning records for all programs
+          programs: [],
```
henri-ly, line 329: So do we need to be more specific here ? When it will be a larger accounts we may have some tokens or spam tokens that we will not want ? and Avoid a 2nd filtering

henri-ly (approved):

libs/coin-modules/coin-aleo/src/bridge/tokens.ts
mdomanski-ext-ledger: We were thinking about breaking the sync as well, so user clearly sees that something went wrong, but in this case we are risking that accounts won't sync until the next LW release with a fix. 

With fallback to 0 it will look weird in the list of operations + we have a log that user could share, so it's less destructive 🙏

libs/coin-modules/coin-aleo/src/bridge/sync.ts
mdomanski-ext-ledger: It never happened before with native operations. The main reason why we set `patched: true` no matter what the results are is to avoid patching "broken" operations over and over with each sync. We just trust the data source (API and record scanner) and don't look back

libs/coin-modules/coin-aleo/src/bridge/sync.ts
mdomanski-ext-ledger: I'm afraid we would have to hardcode the tokens to achieve this 😞 right now we are fetching all records and then use CAL to filter out what shouldn't be there

mdomanski-ext-ledger (approved):

PR #18749, opened 2026-06-18
[LWDM] fix(LIVE-32725): show network currency for fees when sending from token accounts

vladyslavchupovskiy-ext-art:
### ✅ Checklist
  - **Desktop — Send modal**: network fee row now shows the native network currency icon and unit (e.g. XTZ) instead of the token currency (e.g. wUSDC) when sending from a token sub-account.
  - **Desktop — Sign Transaction modal**: same fix applied for consistency.
  - **Desktop — EVM family `AccountFooter`**: same fix for EVM token sends.
  - **Mobile — Tezos send summary**: `TezosFeeRow` now uses `getFeesCurrency` / `getFeesUnit` instead of the account currency for fee amount and countervalue.
  - **QA focus**: send flow from token sub-accounts (Tezos FA2 tokens, EVM ERC-20), verify fee icon, fee amount unit, and fiat countervalue all reflect the **network** currency, not the token.
### 📝 Description
When sending from a token sub-account, the network fee row incorrectly used the **account/token currency** for the currency icon and fee display. For example, sending wUSDC on Tezos showed the wUSDC icon and unit for fees, while fees are actually paid in XTZ.
This PR fixes the fee currency display by using `getFeesCurrency` / `getFeesUnit` (via the parent/main account) instead of `getAccountCurrency` for:
- `Send/AccountFooter` (desktop)
- `SignTransaction/AccountFooter` (desktop)
- `families/evm/AccountFooter` (desktop)
- `TezosFeeRow` (mobile)
A unit test was added for the Send modal `AccountFooter` to assert that a Tezos token sub-account displays **XTZ** as the fees currency.
### ❓ Context
- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-32725
- **ADR link (if any)**: N/A


apps/ledger-live-mobile/src/families/tezos/TezosFeeRow.tsx
```diff
@@ -3,33 +3,32 @@ import { View, StyleSheet, Linking } from "react-native";
...
 };
-export default function TezosFeeRow({ account, transaction }: Props) {
+export default function TezosFeeRow({ account, parentAccount, transaction }: Props) {
   const { colors } = useTheme();
   const extraInfoFees = useCallback(() => {
     Linking.openURL(urls.feesMoreInfo);
   }, []);
 
-  const unit = useAccountUnit(account);
+  const mainAccount = getMainAccount(account, parentAccount ?? undefined);
```
cted-ledger: getMainAccount runs an invariant that throws when account is a TokenAccount and parentAccount is missing (libs/ledger-wallet-framework/src/account/helpers.ts:12-18). The current render path always threads parentAccount (04-Summary → SendRowsFee wrapper → families/tezos/SendRowsFee → here), so there's no regression today. But the prop type is still parentAccount?: Account | null, so a direct usage (test, Storybook, future screen) with a token account and no parent would crash at render. Consider either moving this below the transaction.family early return, or tightening the contract.

apps/ledger-live-desktop/src/renderer/modals/Send/AccountFooter.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+describe("AccountFooter", () => {
+  const status = {
+    errors: {},
+    warnings: {},
+    estimatedFees: new BigNumber(603),
+    amount: new BigNumber(0),
+    totalSpent: new BigNumber(0),
+  } as TransactionStatus;
+
+  it("shows the native fees currency icon when sending from a token sub-account", () => {
```
cted-ledger, line 58: Nice coverage of the token sub-account case. Could you add a complementary case where account is a plain (non-token) Account to lock in that it still renders its own currency icon/unit? That guards against a future regression on the main-account path.

.changeset/friendly-bikes-beam.md
```diff
@@ -0,0 +1,6 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
```
cted-ledger: could you use `patch` instead of `minor` ?

apps/ledger-live-mobile/src/families/tezos/TezosFeeRow.tsx
vladyslavchupovskiy-ext-art: Fair point , thanks ! I would like to go with tightening the contract cause in that case we would be 100% sure that on the compile level no one will pass "TokenAccount" without parent account , sure from runtime perspective it won't happen ( I believe ) but still it's better to be safe than sorry :)

apps/ledger-live-desktop/src/renderer/modals/Send/AccountFooter.test.tsx
vladyslavchupovskiy-ext-art: Yeah makes sens , added in new commit

.changeset/friendly-bikes-beam.md
Justkant: we use minor everywhere in LW, not a big deal because we should have this job that takes care of updating the changeset files https://github.com/LedgerHQ/ledger-live/blob/923b5254aff495ec7145cd226a164a4917b4f839/tools/actions/composites/adjust-changeset-level/action.yml#L30

It's done like this to keep the patch available for hotfixes on libs for npm users IIRC

Justkant (approved):
LGTM but you might want to look at the comment from copilot on the changeset message

vladyslavchupovskiy-ext-art:
This issue gonna require global fix cause it's touching another coin families

PR #18764, opened 2026-06-19
[LWDM] refactor(swap): migrate fetchQuotes to RTK Query (LIVE-32805)

philipptpunkt:
## What

Migrates the swap `fetchQuotes` helper in live-common from a direct `axios` call to an **RTK Query** endpoint, per [LIVE-32805](https://ledgerhq.atlassian.net/browse/LIVE-32805) and the [Data Layer / RTK Query guideline](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6051954851).

## How

Split across the two layers the new architecture asks for, so the backend and the use case have separate owners:

- **Backend service** — `shared/api-services/src/services/swap/`. An endpoint-less `swapApi` (`reducerPath: "swapApi"`) owning the base URL, `X-Ledger-Client-Version`, auth, and the thunk `extraArgument` contract (`swapApiExtra` / `SwapApiExtraSchema`). This follows the pattern the package README prescribes — one store slice per backend, not per use case — so the other `SWAP_API_BASE` routes still in live-common (`/rate`, `/swap/status`, `/currencies/*`, the DEX routes) can later inject here instead of adding another `createApi`.
- **Use case** — `domain/api/swap-quotes` (`@domain/api-swap-quotes`). `swapApi.injectEndpoints({ fetchQuotes })` against `/quote`, plus the zod wire schemas (source of truth, with every type `z.infer`red from them), fixtures, and the store helpers. The package has no `libs/` dependency, so it needs no boundary exception.
- **live-common** — `wallet-api/Exchange/quotes/state-manager/*` and `service/types.ts` are now thin re-export shims; `service/fetchQuotes.ts` dispatches the endpoint imperatively. It runs server-side in the wallet-api `getQuotes` handler rather than in a React hook, so it obtains `dispatch` through a global setter (`setSwapQuotesStore` / `getSwapQuotesDispatch`). `forceRefetch: true` keeps quotes fresh; the subscription is held for the lifetime of the call and released in `finally`, which with `keepUnusedDataFor: 0` evicts the entry immediately.
- **App wiring** — both apps register `swapApi` from `@shared/api-services` in `rtkQueryApi.ts` and supply `swapApiExtra(…)` in `configureStore.ts`; `setSwapQuotesStore(store.dispatch)` is called from desktop `renderer/init.tsx:137` and mobile `state-manager/configureStore.ts:137`. wallet-cli has no app store, so `live-common-setup.ts` builds a standalone one.

## Behaviour changes

Three things genuinely change. All are in the changeset.

1. **`/quote` now goes through the authenticated base query** (`@shared/auth`), where the legacy axios call sent no credentials. Both apps already register an auth provider on their store's `extra`, so whether a request carries an `Authorization` header is controlled entirely by the **`lwdAuth`/`lwmAuth` feature flags**. They are off by default; enabling either makes `/quote` send the user's bearer token to the aggregator, and makes a **401** trigger the adapter's refresh-and-retry. (Until recently that set also included 403; develop narrowed it to 401 only, which removes the geo-block/duplicate-request concern raised in review.) Authenticating this request is in scope for the API-hardening work, and the flag provides the progressive rollout.

2. **Any aggregator HTTP error now resolves to an empty result**, so the caller surfaces the `noQuotes` global. Previously the shared `@ledgerhq/live-network` axios interceptor — installed on the *global* axios instance (`libs/live-network/src/network.ts:105-106`) — turned these into `LedgerAPI4xx`/`LedgerAPI5xx`, which reached the live app as an error. This is broader than a 5xx outage: a `400` from malformed params and a `401`/`403` that survives refresh-and-retry now also render as "no quotes for this pair", indistinguishable from a genuinely empty result. Only transport failures still reject, now with a named `SwapQuotesRequestFailed`. **This is user-visible and wants a PTX call.**

   HTTP status errors are deliberately left as RTK Query errors (rather than swallowed in the base query) so the 401 refresh-and-retry can fire; `fetchQuotes` maps them to the empty result afterwards.

3. **`SWAP_API_BASE` is now resolved once, at store construction**, instead of per request via `getSwapAPIBaseURL()`. Changing the env at runtime (developer settings, `experimental.tsx` localStorage keys) no longer affects in-flight quote requests until restart.

## Notes

- `X-Ledger-Client-Version` is set explicitly. The axios helper inherited it from the default instance, which live-network decorates globally; `fetch` inherits nothing. `User-Agent`, which live-network also sets on Node, is **not** restored — it is not settable from a browser context.
- `customHeaders` are kept out of the RTK Query cache key. Two consequences, both known: concurrent requests differing only by header share one in-flight call (RTK Query's `condition` returns early on `status === "pending"`, before `forceRefetch` is consulted), and the headers still travel in `originalArgs`, which desktop's redux logger records for `*/executeQuery/*` actions.
- The abort `signal` is not threaded through. RTK Query *can* cancel — `initiate()` returns a promise with `abort()` — but no caller supplies a signal, and live apps cannot: an `AbortSignal` does not survive the wallet-api boundary. `GetQuotesWireArgs` (already upstream in `libs/exchange-module`) omits it.
- Aggregator HTTP errors are no longer logged. The old interceptor emitted `log("network-error", …)`; combined with change 2 above, an outage is currently invisible in both the UI and support logs.

## Tests

- **`@domain/api-swap-quotes`** — 34, including msw integration tests that assert the outgoing query string and headers, the authenticated path (token sent when the provider yields one, no header while the flag is off), HTTP-error and transport-error mapping, and that `customHeaders` stay out of the cache key.
- **live-common quotes suite** — 223, covering the shims, the dispatch/store plumbing, error mapping, and regression coverage for repeated and concurrent identical requests (the subscription must be held, or `keepUnusedDataFor: 0` evicts the entry before the promise resolves and the result reads back empty).
- **Both apps** — the pinned `EXPECTED_REDUCER_PATHS` lists in `rtkQueryApi.test.ts` include `swapApi`, so a dropped or duplicated backend fails the build.

## Known follow-ups

Deliberately deferred rather than expanding this PR further. **These will land as a stacked PR on top of this one** rather than as further commits here, so the review already done is not invalidated:

- No `api.test.ts` for the swap service in `@shared/api-services`, which all four sibling services have — it matters more here, as the first consumer of `createAuthenticatedBaseQuery` outside `shared/auth`.
- The `globalThis` dispatch singleton. `GetQuotesContext` is an existing typed DI seam that every caller already builds; threading dispatch through it would remove the global and make a missing registration a compile error. At minimum, a per-app test asserting init calls `setSwapQuotesStore`.
- Concurrent-dedupe header crossing: fold a hash of `customHeaders` into the cache key, or make "headers must not affect the response body" an explicit contract.
- `customHeaders` and `sendAddress`/`receiveAddress` reaching desktop's exportable support log — redact before dispatch, or filter in the logger.
- No logging on aggregator HTTP errors (above).
- Fixtures are duplicated between `@domain/api-swap-quotes` and live-common because the package `exports` map has no `./fixtures` subpath.
- Workspace deps here use `workspace:^`; sibling domain packages use `workspace:*`.
- Until the auth flags are enabled, `shared/auth` warns per request when authentication cannot be obtained; the fix belongs in that adapter, not here.

[LIVE-32805]: https://ledgerhq.atlassian.net/browse/LIVE-32805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,128 @@
...
+export function buildQuotesParams(
+  providers: string[],
+  quotesInput: ResolvedQuotesInput,
+  counterValueCurrency: string,
+): Record<string, string> {
+  const params: Record<string, string> = {
+    amountFrom: quotesInput.amount,
+    displayLanguage: "en",
+    lang: "en",
+    theme: "dark",
```
Justkant: I see we kept those from the previous logic, should it really be kept ?
Does it impact the returned data ? Should we use the actual user lang ?
Or is it just leftovers from the live-app query params ?

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,128 @@
...
+  reducerPath: "swapQuotesApi",
+  baseQuery: fetchBaseQuery({
+    // The aggregator base URL is resolved per-request (from SWAP_API_BASE) in
+    // each endpoint's `query`, so the static base URL stays empty.
+    baseUrl: "",
+    // Treat every HTTP response as "successful" and let transformResponse map
+    // non-OK statuses to an empty result, matching the legacy helper which
+    // swallowed HTTP errors. Transport failures still surface as RTK errors.
+    validateStatus: () => true,
+  }),
```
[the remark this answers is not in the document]
Justkant: it's true that doing this seems a bit fragile or at least not really specific for the errors we handle

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/store.ts
```diff
@@ -0,0 +1,42 @@
...
+ * The reference is kept on `globalThis` to guarantee a single shared
+ * dispatch across all module instances, even when live-common is resolved
+ * to separate copies by lazy-loaded bundles.
+ */
+export type SwapQuotesDispatch = <T>(action: T) => Promise<unknown>;
+
+declare global {
+  // eslint-disable-next-line no-var
+  var __ledgerSwapQuotesDispatch: SwapQuotesDispatch | undefined;
+}
```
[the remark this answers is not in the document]
Justkant: you might want to copy the other places it's done like in the comment

Justkant (dismissed):
LGTM but I would look at the comment from copilot on the `validateStatus` and the one for the globalThis interface usage

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
philipptpunkt: I need to check with BE again, but I think they are required, however they're not used on BE side - that's why they're hardcoded right now...

philipptpunkt: Checked, this is not in use right now - left over from non-native providers. I'd say we can leave it like this and remove once BE made it optional.

apps/wallet-cli/src/live-common-setup.ts
```diff
@@ -4,6 +4,7 @@ import { registerCoinModules } from "@ledgerhq/live-common/coin-modules/registry
 import type { CoinModuleLoader } from "@ledgerhq/live-common/coin-modules/types";
 import { setWalletAPIVersion } from "@ledgerhq/live-common/wallet-api/version";
 import { WALLET_API_VERSION } from "@ledgerhq/live-common/wallet-api/constants";
+import { setupStandaloneSwapQuotesStore } from "@ledgerhq/live-common/wallet-api/Exchange/quotes/state-manager/standaloneStore";
```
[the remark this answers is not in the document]
philipptpunkt: ✅

libs/ledger-live-common/src/wallet-api/Exchange/quotes/service/fetchQuotes.ts
```diff
@@ -24,72 +28,33 @@ type FetchQuotesArgs = Omit<GetQuotesArgs, "data"> & {
...
-
-  if (quotesInput.networkFeesCurrencyId) {
-    searchParams.set("networkFeesCurrency", quotesInput.networkFeesCurrencyId);
-  }
-
-  if (quotesInput.slippage != null) {
-    searchParams.set("slippage", quotesInput.slippage.toString());
+  const { providers, data: quotesInput, headers: customHeaders } = args;
+  const dispatch = getSwapQuotesDispatch();
+
```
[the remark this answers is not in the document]
philipptpunkt: Fixed by wiring a store dispatch into wallet-cli's bootstrap. Added `setupStandaloneSwapQuotesStore()` in `live-common` (a minimal RTK store for `swapQuotesApi` that calls `setSwapQuotesStore`, mirroring cryptoassets' `setupCalClientStore`), and it's invoked at the top of live-common-setup.ts, which cli.ts imports for its side effect before any command runs. So `getQuotes` → `fetchQuotes` → `getSwapQuotesDispatch()` now has a registered dispatch and no longer throws.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/standaloneStore.test.ts
```diff
@@ -0,0 +1,61 @@
...
+describe("setupStandaloneSwapQuotesStore", () => {
+  const server = setupServer();
+
+  beforeAll(() => server.listen());
+  afterEach(() => server.resetHandlers());
+  afterAll(() => {
+    server.close();
+    // Clear the global dispatch registered during the tests.
+    globalThis.__ledgerSwapQuotesDispatch = undefined;
+  });
```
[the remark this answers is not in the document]
philipptpunkt: added a `beforeEach` that resets `globalThis.__ledgerSwapQuotesDispatch` (matching the existing currencies-store.test.ts pattern)

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,133 @@
...
+    // Treat every HTTP response as "successful" and let transformResponse map
+    // non-OK statuses to an empty result, matching the legacy helper which
+    // swallowed HTTP errors. Transport failures still surface as RTK errors.
+    validateStatus: () => true,
+    // Only parse the JSON body for 2xx responses. Aggregator error responses
+    // (4xx/5xx) frequently carry a non-JSON body; letting fetchBaseQuery parse
+    // it would surface as a PARSING_ERROR on `result.error` instead of letting
+    // transformResponse map the non-OK status to an empty result.
+    responseHandler: async (response: Response) => (response.ok ? response.json() : null),
+  }),
```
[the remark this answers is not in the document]
philipptpunkt: `responseHandler` now wraps `response.json()` in a `try/catch` and returns `null` on parse failure

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,142 @@
...
+        params: buildQuotesParams(providers, quotesInput, counterValueCurrency),
+        headers: {
+          Accept: "application/json",
+          ...(customHeaders ?? {}),
+        },
+      }),
+      transformResponse: transformFetchQuotesResponse,
+      // Quotes are time-sensitive: never retain them between requests.
+      keepUnusedDataFor: 0,
+    }),
```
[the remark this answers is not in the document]
philipptpunkt: not an issue for now and expected

deepyjr (dismissed):

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,142 @@
...
+ * RTK Query API for the swap quotes aggregator. Exposed as a single
+ * `fetchQuotes` query that replaces the legacy axios `fetchQuotes` helper.
+ *
+ * Consumed imperatively from the server-side `getQuotes` flow (not a React
+ * hook) via `dispatch(swapQuotesApi.endpoints.fetchQuotes.initiate(...))`,
+ * mirroring the `cryptoAssetsApi` (CAL client) pattern.
+ */
+export const swapQuotesApi = createApi({
+  reducerPath: "swapQuotesApi",
+  baseQuery: createAuthenticatedBaseQuery({
```
thesan: Thank you 🙏

henri-ly (dismissed):

philipptpunkt (commented):
> Claude went through this one, shape of the refactor is solid and the helper extraction is clean. One thing I'd call blocking, plus a couple of spots where the correctness claims aren't actually pinned by the tests.
> 
> :red_circle: Blocking — createAuthenticatedBaseQuery is the wrong base query here
> api.ts uses the authenticated base query, but the legacy axios call was unauthenticated. Checked against develop:
> • Nobody supplies extra.authSDK — not desktop configureStore.ts:42, not mobile, and setupStandaloneSwapQuotesStore passes no extra at all. So every /quote request hits the zod-parse failure → AuthenticatedBaseQueryMissingAuthSDKError → the catch at shared/auth/src/createAuthenticatedBaseQuery.ts:46, emitting a console.warn before falling back to plain fetchBaseQuery. Quotes refetch on every amount change, so that's warn-spam on a hot path.
> • Latent security change — once authSDK does land in app extra (libs/ledger-auth already exists), /quote silently starts carrying Authorization: <type> <accessToken> to a host resolved by getSwapAPIBaseURL(), which is env-overridable via SWAP_API_BASE. That contradicts the PR's own "wire request preserved byte-for-byte" goal.
> • Dead interaction with validateStatus — refreshAndRetryWhen keys off result.error, but validateStatus: () => true guarantees result.error is never set for 401/403. Refresh-and-retry can never fire, and a 401 would silently become "no quotes".
> • The wrapper is also untested here — the integration store in api.test.ts has no extra, so only the fallback path runs.
> 
> → Suggest plain fetchBaseQuery({ baseUrl: "", validateStatus, responseHandler }). Restores wire parity and drops the new @shared/auth dep from live-common + the lockfile churn.
> 
> :large_orange_circle: Worth fixing in the same pass
> • FetchQuotesArgs still accepts signal, now silently ignored. Omit<GetQuotesArgs, "data"> keeps it in the type while the body no longer destructures it. Your claim checks out — server.ts:782 calls getQuotes with GetQuotesWireArgs, which has no signal, so no live regression — but the type advertises cancellation it doesn't provide. Omit<GetQuotesArgs, "data" | "signal"> makes the compiler enforce the comment.
> • The "byte-for-byte" invariant isn't actually asserted. buildQuotesParams is checked with toMatchObject, which by construction can't catch a missing or extra param — and displayLanguage, lang, theme, networkFees are never asserted at all, exactly the constant params a refactor is most likely to drop. toEqual on the full object. Related: nothing checks the request that actually goes out (headers / query string). One msw handler reading request.url.searchParams + request.headers.get("accept") would pin both.
> 
> :white_circle: Nits
> • Unreachable branch: transformResponse only runs when a response arrived, so meta?.response?.status is always defined — the status === undefined branch and its test are dead. Reads as coverage it isn't.
> • Changeset omits @ledgerhq/wallet-cli, which is changeset-versioned (fixed group in .changeset/config.json) and is modified here.
> • Design note, take it or leave it: with forceRefetch: true + subscribe: false + keepUnusedDataFor: 0 there's no cache reuse, no hook consumers, no tag invalidation — so the store round-trip's entire benefit is guideline conformance, against a globalThis singleton, two wiring points, a standalone store, and every quote payload transiting the app middleware chain. I did verify it's not a leak: canTriggerUnsubscribe includes queryThunk.fulfilled, so the 0s removal timer fires and entries don't accumulate. Just overhead. Fine if the guideline mandates it — worth being explicit that that's the trade.
> 
> :white_check_mark: Things I liked
> • createApi from @reduxjs/toolkit/query rather than /query/react — right call, keeps react-redux out of wallet-cli (and the divergence from ofacGeoBlockApi is the right way round).
> • The responseHandler guard is a real improvement over axios, which would've thrown a TypeError from .filter on a 2xx non-array body. Good that the 502-HTML case has an integration test.
> • standaloneStore.ts correctly in entry, not ignoreUnimported.
> • standaloneStore.test.ts resetting the global in both beforeEach and afterAll is the right hygiene for a globalThis singleton, and the comment explaining why earns its keep.


### Status of each point:

**Already fixed**

| Point | Where |
|---|---|
| Dead `validateStatus` interaction | `validateStatus: () => true` and the custom `responseHandler` are **removed**. HTTP errors are now real RTK Query errors, so `refreshAndRetryWhen` can fire. `fetchQuotes` maps status-bearing errors to the legacy empty result. |
| `signal` still in the type | Narrowed to `Omit<GetQuotesArgs, "data" \| "signal">`. |
| Unreachable `status === undefined` branch | Gone — `transformFetchQuotesResponse(response)` no longer takes `meta`, and the dead test with it. |
| Changeset omits `@ledgerhq/wallet-cli` | Added. |
| Nothing asserts the outgoing request | Added an msw test reading `request.url.searchParams` + `headers.get("accept")`, plus one pinning caller-header override order. |
| `toMatchObject` can't catch a dropped param | Now `toEqual` on the full 14-param object, including `displayLanguage`/`lang`/`theme`/`networkFees`. |

**Also fixed, found separately** — a real bug your pass didn't reach: `initiate(..., { subscribe: false })` resolves to the *selected cache entry*, which `keepUnusedDataFor: 0` could evict before the promise settled. Repeated and concurrent identical quote requests returned **zero quotes** despite a successful 200 (reproduced: sequential ×3 → `{1, 0, 1}` with 3 HTTP hits; concurrent ×2 → `{0, 0}`). Fixed by holding the subscription and releasing it in `finally`; regression tests fail if reverted. Also re-added `X-Ledger-Client-Version` (lost with the axios default instance) and wrapped the thrown `FetchBaseQueryError` in a real `SwapQuotesRequestFailed`.

**Not changing** — see the inline replies on `api.ts` and `fetchQuotes.ts`.

One correction on the 👍 for `standaloneStore.test.ts`: that comment was factually wrong, so it's gone. See the inline note.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * Unlike the legacy axios call, `/quote` is authenticated: once `authSDK` is
+ * registered on the store's `extra` the request carries an `Authorization`
+ * header, and 401/403 trigger the adapter's refresh-and-retry. HTTP status
+ * errors are deliberately left as RTK Query errors so that retry can fire;
+ * `fetchQuotes` maps them to the legacy empty-result outcome.
+ */
+export const swapQuotesApi = createApi({
+  reducerPath: "swapQuotesApi",
+  baseQuery: createAuthenticatedBaseQuery({
```
philipptpunkt: > [!IMPORTANT]
> **Corrected 2026-07-31.** The original version of this comment claimed that no app registers an auth provider and that the adapter therefore falls back to an unauthenticated request. **That was wrong**, and the changeset repeated it. The wiring landed on develop while this branch was being rebased. Corrected text below; the changeset and the `swapQuotesApi` JSDoc are fixed in `94e9da4`.

**Keeping `createAuthenticatedBaseQuery` — this is a deliberate product decision, not an oversight.**

Raised with the team that owns the auth-hardening scope; authenticating `/quote` is the intent, and `createAuthenticatedBaseQuery` is the sanctioned adapter for it.

**How it actually behaves today** (verified against source at this head):

- Both apps **do** register an auth provider — `apps/ledger-live-desktop/src/state-manager/configureStore.ts:76` and `apps/ledger-live-mobile/src/state-manager/configureStore.ts:65` spread `authApiExtra({ authFeatureId: "lwdAuth" | "lwmAuth", … })` into the store's `thunk.extraArgument`. `getAuthProvider(api.extra)` resolves; there is no fallback.
- The gate is the **feature flag**: `shared/auth/src/api.ts:44` returns the untokenized `queryFn()` while `authEnabled` is false, and `lwdAuth`/`lwmAuth` default to `{ enabled: false }`. They are remote-config driven, so enabling either starts sending `Authorization: <type> <token>` to the aggregator with no code change. That is the intended progressive rollout.
- **`refreshAndRetryWhen` fires on 401 and 403** (`UNAUTHORIZED_STATUSES = new Set([401, 403])`). Worth flagging to whoever owns the flag: a swap-aggregator **403 is a plausible geo-block or entitlement response**, and it would trigger a token refresh plus a duplicate quote request on a hot path. That interaction deserves a look before the flag is enabled.

**Your `validateStatus` point was a real bug and is fixed.** `validateStatus: () => true` and the custom `responseHandler` are gone. HTTP status errors now surface as RTK Query errors so the retry can actually fire, and `fetchQuotes` maps them to the empty-result outcome via `getHttpStatus`. Your observation that a 401 would silently become "no quotes" was correct and no longer holds.

**The untested-wrapper point was also right.** Every test built its store without `extra`, so all requests took the missing-provider fallback — the opposite branch from production — and each logged a warning (36 warnings, ~1000 lines of stack traces per suite run). Tests now supply a stub provider matching the apps, plus two new cases: no `Authorization` while the flag is off, and the bearer token present once the provider yields one. Suite output is down to 23 lines. wallet-cli's standalone store also gets a no-op provider, so headless callers stop warning per request.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/service/fetchQuotes.ts
```diff
@@ -24,72 +59,52 @@ type FetchQuotesArgs = Omit<GetQuotesArgs, "data"> & {
...
+        counterValueCurrency,
+        customHeaders: customHeaders ? Object.fromEntries(customHeaders) : undefined,
+      },
+      // Bypass any cached entry: quotes are time-sensitive. The subscription is
+      // held only for the lifetime of this call and released in `finally`, so
+      // `keepUnusedDataFor: 0` still evicts the entry immediately. Without a
+      // subscriber the entry can be collected before this promise resolves,
+      // which reads back as an empty result. Concurrent identical requests are
+      // still de-duplicated by RTK Query.
+      { forceRefetch: true },
```
philipptpunkt, line 70: **Re the design note on overhead — agreed, and being explicit that it's the trade.**

You're right that there is no cache reuse, no hook consumers and no tag invalidation here, so the store round-trip buys guideline conformance rather than behaviour. That is the deliberate call: [LIVE-32805](https://ledgerhq.atlassian.net/browse/LIVE-32805) exists to move this onto the Data Layer / RTK Query guideline, and the `globalThis` singleton + two wiring points + standalone store are the cost of doing that for a call that runs server-side in the wallet-api handler rather than from a component.

Thanks for verifying the eviction independently — `canTriggerUnsubscribe` including `queryThunk.fulfilled` is exactly right, and there's now a test asserting `state.swapQuotesApi.queries` is empty after a request settles.

One correction to the options you were reading: `subscribe: false` is **gone**. It was not merely redundant, it was a bug — `initiate` resolves to the selected cache entry, and with `keepUnusedDataFor: 0` and no subscriber that entry could be evicted before the promise resolved. Repeated and concurrent identical requests returned zero quotes on a successful 200 (sequential ×3 → `{1, 0, 1}` with 3 real HTTP hits; concurrent ×2 → `{0, 0}`). The subscription is now held for the call and released in `finally`, which keeps the eviction you verified while fixing the read. Two regression tests cover it and fail if `subscribe: false` is restored.

domain/api/swap-quotes/src/store/store.standalone.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+jest.mock("../../../../exchange/swap", () => ({
+  getSwapAPIBaseURL: jest.fn(() => "https://swap.test"),
+}));
+
+describe("setupStandaloneSwapQuotesStore", () => {
+  const server = setupServer();
+
+  beforeAll(() => server.listen());
+  beforeEach(() => {
```
philipptpunkt, line 18: **Correcting one of the 👍 items:** the `afterAll` reset and its comment are gone, because the comment's justification was false.

It claimed the dispatch "persists across test files in a Jest worker" and that the reset stopped it leaking into other suites. Verified empirically with two probe files in this directory (one setting `globalThis.__probeLeak`, one reading it, `--runInBand`): the reader saw `undefined`. Jest gives every test **file** its own environment and its own `globalThis`, so there was no cross-file hazard and the `afterAll` was dead code justified by a comment that would mislead the next reader.

The `beforeEach` reset stays — it is genuinely needed, but for intra-file ordering only (so the "throws before setup" assertion doesn't depend on test order), and the comment now says that.

Same pass removed two other comments asserting precedents that don't exist: `standaloneStore.ts` cited `setupCalClientStore` from `@ledgerhq/cryptoassets/cal-client` (neither symbol nor package exists anywhere in the repo), and `store.ts` claimed it mirrors `setCryptoAssetsStore`, which uses a module-level `let` rather than `globalThis` — the opposite of what it does. It now cites `src/hw/index.ts`, which is the actual precedent.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/store.ts
```diff
@@ -0,0 +1,44 @@
...
+ * The reference is kept on `globalThis` to guarantee a single shared
+ * dispatch across all module instances, even when live-common is resolved
+ * to separate copies by lazy-loaded bundles. `registerTransportModule` in
+ * `src/hw/index.ts` uses the same approach for the same reason.
+ */
+import type { ThunkDispatch, UnknownAction } from "@reduxjs/toolkit";
+
+export type SwapQuotesDispatch = ThunkDispatch<unknown, unknown, UnknownAction>;
+
+declare global {
```
philipptpunkt: Related to the `globalThis` singleton discussion: the original `declare global { interface GlobalThis { … } }` augmented **nothing**. TypeScript derives `typeof globalThis` from global `var` declarations and never consults an interface by that name, so the `SwapQuotesDispatch` annotation was decorative and every access was implicitly `any` (it only compiled because live-common sets `noImplicitAny: false`).

Changed to `var __ledgerSwapQuotesDispatch: SwapQuotesDispatch | undefined`, which actually applies. Worth knowing that `src/hw/index.ts:51` and `src/manager/api.ts:45` have the same latent issue — out of scope here, but they're not the precedent to copy.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
philipptpunkt: Flagging explicitly since the earlier version of this thread may have been read: **I had this backwards.** I wrote that no app registers an auth provider so nothing is sent — both apps do register one (`authApiExtra` in each `configureStore.ts`), and the `lwdAuth`/`lwmAuth` remote flag is what decides whether the token goes out. The wiring landed on develop while this branch was being rebased, so a statement that was true when written stopped being true.

Corrected in `94e9da4`: changeset, `swapQuotesApi` JSDoc, and the comment above. The authenticated path now has test coverage (it previously had none — every test took the missing-provider fallback), which also removed ~1000 lines of warning noise per suite run.

One thing worth a second opinion before the flag is enabled: `refreshAndRetryWhen` retries on **403** as well as 401, and a swap-aggregator 403 is a plausible geo-block response — that would cause a spurious token refresh and a duplicate quote request.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,139 @@
...
+  quotesInput: ResolvedQuotesInput;
+  /** Fiat ticker (e.g. `"USD"`) the aggregator should use for quote countervalues. */
+  counterValueCurrency: string;
+  /**
+   * Optional caller-supplied headers, already flattened to a plain object.
+   * `serializeQueryArgs` keeps them out of the cache key so a live-app token
+   * cannot widen it. This is a cache-key guarantee only: RTK Query still
+   * records the unstripped args as `originalArgs` on the cache entry.
+   */
+  customHeaders?: Record<string, string>;
```
philipptpunkt: **Deferred: `customHeaders` still reach desktop's exportable support log.**

`serializeQueryArgs` keeps them out of the *cache key*, but RTK Query still records the unstripped args as `originalArgs` on the cache entry and on the action's `meta.arg`. Desktop's redux logger (`renderer/logger/index.ts:122-133`) copies `meta.arg` verbatim for any `*/executeQuery/*` action into the winston ring buffer that `saveLogs.ts` writes — the file users attach to support tickets.

So two things now land in an exportable artifact that never did under axios: the swap send/receive addresses and account ids, and whatever headers a live app supplied. `headers` on `GetQuotesArgs` is unvalidated pass-through from `custom.exchange.getQuotes`, so a partner token there would be written to that file.

Not fixed here because the containment options both reach outside this endpoint: either redact in desktop's shared logger (the more general fix, covering any future address-bearing endpoint), or stop putting `customHeaders` in the serialized arg at all. Worth its own ticket with the desktop owners rather than widening this PR again. Mobile is unaffected — no redux-action logging — and nothing here is persisted or sent to Datadog.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,139 @@
...
+ * carries an `Authorization` header; they are off by default. HTTP status
+ * errors are deliberately left as RTK Query errors so the adapter's 401/403
+ * refresh-and-retry can fire; `fetchQuotes` maps them to an empty result.
+ */
+export const swapQuotesApi = createApi({
+  reducerPath: "swapQuotesApi",
+  baseQuery: createAuthenticatedBaseQuery({
+    // The aggregator base URL is resolved per-request (from SWAP_API_BASE) in
+    // each endpoint's `query`, so the static base URL stays empty.
+    baseUrl: "",
```
philipptpunkt: **Deferred: no host allowlist on a request that can carry a token.**

`baseUrl` is empty and the absolute URL comes per-request from `getSwapAPIBaseURL()` → `getEnv("SWAP_API_BASE")`. Every other RTK Query API in the repo pins `baseUrl` once at `createApi` time. On desktop, `renderer/experimental.tsx` applies env keys from `localStorage` without filtering them against the `experimentalFeatures` allowlist, so a persisted entry can redirect `/quote`.

That is not remotely exploitable — it needs someone who can already write renderer `localStorage` or the process env — and it is unchanged from the axios implementation. What *is* new is that once `lwdAuth`/`lwmAuth` is enabled, the `Authorization` header follows wherever that value points, turning a local foothold into token exfiltration.

Leaving as-is because pinning the base URL changes how the env override works for the e2e harness, and because the mitigation belongs with the auth rollout rather than this migration. Flagging for whoever enables the flag: worth validating the resolved host before attaching credentials.

Related, same audience: `refreshAndRetryWhen` treats **403** as unauthorized alongside 401 (`shared/auth/src/createAuthenticatedBaseQuery.ts:21`). A swap-aggregator 403 is a plausible geo-block or entitlement response, so once the flag is on that would trigger a token refresh and a duplicate quote request on a hot path.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Reshape a successful `/quote` response body into a {@link FetchQuotesResult}.
+ * Only 2xx responses reach here; a body that is not an array yields an empty
+ * result rather than passing a malformed payload downstream. Non-OK statuses
+ * surface as RTK Query errors and are mapped by `fetchQuotes`.
+ */
+export function transformFetchQuotesResponse(response: unknown): FetchQuotesResult {
+  return splitQuotes(Array.isArray(response) ? (response as Array<RawQuote | RawQuoteError>) : []);
```
philipptpunkt: **Deferred: no runtime validation of the aggregator payload.**

The body is guarded only by `Array.isArray`, then cast, and rows are discriminated on `"code" in q` alone. Anything array-shaped flows into `normalizeQuote` and the fee/amount maths. The nearest precedent, `counterValues/state-manager/api.ts`, does this properly with a zod `responseSchema` plus `onSchemaFailure`.

Not fixed here because the permissiveness is inherited verbatim from the axios helper — this is not a regression — and adding schema validation changes what happens on a malformed payload (silently dropped rows today vs. a surfaced schema failure), which is a product decision about swap error surfaces rather than a transport concern. Reasonable follow-up while the code is warm.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
```diff
@@ -0,0 +1,139 @@
+import { createApi } from "@reduxjs/toolkit/query";
```
philipptpunkt: **Deliberate: `@reduxjs/toolkit/query`, not `/query/react`.**

Every other API registered in the apps' `rtkQueryApi.ts` imports from `/query/react`, so this is the odd one out. Keeping it: this endpoint is consumed imperatively from the wallet-api `getQuotes` handler and exports no hooks, and `/query/react` would pull react-redux into wallet-cli, which has no React.

The trade-off is that the app bundles a second copy of the RTKQ core. If someone would rather have the shared core than keep react-redux out of the CLI, switching the import and not exporting hooks is a one-line change — but it should be a conscious call, not drift.

philipptpunkt (commented):
Second pass over the review findings. Everything actionable inside this PR's scope is now fixed; the four items below are deliberate deferrals, recorded inline so they don't get re-raised.

**Fixed since the last round** (`94e9da4`, `8ddaa29`): corrected the auth claims in the changeset and JSDoc (both apps *do* register a provider — the gate is the `lwdAuth`/`lwmAuth` flag); added coverage for the authenticated path, which had none; removed the redundant `Accept` header and its tautological assertion; gave wallet-cli's standalone store a no-op provider; narrowed `getQuotes` to `GetQuotesWireArgs` so the public type stops advertising an abort `signal` it ignores; exported `resetSwapQuotesStore`; shared the `makeRawQuote`/`makeRawQuoteError` fixtures; moved the live-store tests beside `fetchQuotes`; covered the remaining error-shaping branches (`fetchQuotes.ts` 73.9% → 100% branch). Test-suite output dropped from ~1015 lines to 24, and auth warnings from 36 to 0.

Also corrected the PR description, which still described a `validateStatus` design that no longer exists and claimed byte-for-byte header parity that was never true.

deepyjr (dismissed):

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/standaloneStore.ts
ysitbon, line 1: I think you can move it to the domain too. Create a store dir and push it both `store` and `store.standalone`

domain/api/swap-quotes/src/types.ts
ysitbon, line 1: Here, most of the types should have their Schema first and then be inferred from the schemas. We want to prioritize the schema first. We gain typing, validation, and factories for free.

domain/api/swap-quotes/src/fixtures/quotesInput.ts
ysitbon, line 1: You can move the 2 *.fixtures files to a `fixtures` dir and use schema too

shared/api-services/src/services/swap/api.ts
```diff
@@ -0,0 +1,59 @@
+import {
+  createApi,
+  type BaseQueryFn,
+  type FetchArgs,
+  type FetchBaseQueryError,
+} from "@reduxjs/toolkit/query/react";
```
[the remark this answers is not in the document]
philipptpunkt: **Not changing this — though you're right that it reverses an earlier decision in this PR, so it's worth recording why rather than just resolving.**

This suggestion was taken on 20 Jul: the api imported `@reduxjs/toolkit/query` (core) specifically to keep `react-redux` out of wallet-cli. The later `@shared/api-services` refactor moved the swap `createApi` into that package, where all five services import `/query/react`, and re-introduced it.

Keeping it, for two reasons.

**1. The predicted failure does not occur.** The `/query/react` import lives in `@shared/api-services`, so `react-redux` is resolved relative to *that* package's RTK instance, not wallet-cli's. pnpm gives the two packages different instances:

| Package | RTK instance pnpm links | `require("@reduxjs/toolkit/query/react")` |
| --- | --- | --- |
| `shared/api-services` | `@reduxjs+toolkit@2.11.2_react-redux@9.2.0_react@19.1.4__react@19.1.4` (**peered**) | loads |
| `domain/api/swap-quotes` | `@reduxjs+toolkit@2.11.2` (unpeered) | would throw — but nothing there imports it; only `configureStore` and types from the core entry |

`pnpm-lock.yaml` records that peered version for `shared/api-services`, so CI resolves identically, and `.npmrc` sets `auto-install-peers=false` with `strict-peer-dependencies=false` — so nothing is quietly papering over a missing peer either.

So wallet-cli does not need to declare `react-redux`: the package that imports it declares it.

One limit on that claim: this was verified by module resolution from each package root, not by booting the CLI, so read "the CLI runs" as inferred rather than proven.

**2. Consistency in a package we don't own.** `@shared/api-services` is owned by @ledgerhq/platform, and its other four services (`cal`, `coinmarketcap`, `countervalues`, `push-devices`) all import `/query/react`. Making swap the single exception trades this question for "why is swap different", and the divergence would be invisible to whoever adds the next service.

**What we are accepting:** wallet-cli's dependency graph now carries `react-redux` and the React-coupled RTKQ entry it never uses. That is bundle weight in a dev CLI, not a break — `swapApi` exports no hooks, so nothing pulls the bindings into use.

If that weight does become a problem, the fix belongs in `@shared/api-services` as a core-only subpath export for headless consumers (your other comment's second option), not in `apps/wallet-cli/package.json`. That is a platform-owned change rather than something to bolt onto this PR.

domain/api/swap-quotes/src/store/store.standalone.ts
```diff
@@ -0,0 +1,38 @@
+import { configureStore } from "@reduxjs/toolkit";
+import { swapApi, swapApiExtra, type SwapApiExtra } from "@shared/api-services";
+import type { AuthProvider } from "@shared/auth";
```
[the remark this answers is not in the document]
philipptpunkt: Same root cause as [the thread on `shared/api-services/src/services/swap/api.ts:6`](https://github.com/LedgerHQ/ledger-live/pull/18764#discussion_r3712837549), answered in full there — not changing.

Short version: the `/query/react` import lives in `@shared/api-services`, which declares the `react-redux` peer and gets the peered RTK instance, so resolution happens there rather than in wallet-cli. This package imports only `configureStore` and types from the core `@reduxjs/toolkit` entry, so it never resolves `/query/react` itself.

apps/wallet-cli/src/live-common-setup.ts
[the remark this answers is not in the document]
philipptpunkt: Correcting the ✅ above, since the history is now misleading: this **was** fixed as suggested (core `@reduxjs/toolkit/query` entry, no hooks exported), but the later `@shared/api-services` refactor moved the swap `createApi` into that package and re-introduced `/query/react`.

It is now a deliberate keep rather than an oversight — rationale and the peer-resolution evidence are [here](https://github.com/LedgerHQ/ledger-live/pull/18764#discussion_r3712837549).

libs/ledger-live-common/src/wallet-api/Exchange/quotes/state-manager/api.ts
philipptpunkt: **Superseded — this comment no longer describes the code.**

The swap `createApi` has moved to `shared/api-services/src/services/swap/api.ts`, and this file is now a re-export shim. That service imports `/query/react` like its four sibling services, so the trade-off described above was decided the other way round: RTKQ core is shared, and wallet-cli's graph carries `react-redux` it does not use.

Current rationale, including why the missing-peer failure does not actually occur: [here](https://github.com/LedgerHQ/ledger-live/pull/18764#discussion_r3712837549).

ysitbon (approved):

deepyjr (approved):

apps/ledger-live-desktop/src/state-manager/configureStore.ts
```diff
@@ -69,6 +70,10 @@ const customCreateStore = ({
                 pushDevicesServiceUrl: getEnv("PUSH_DEVICES_SERVICE_URL"),
                 ledgerClientVersion: getEnv("LEDGER_CLIENT_VERSION"),
               }),
+              ...swapApiExtra({
+                swapApiBaseUrl: getEnv("SWAP_API_BASE"),
```
Justkant, line 74: nitpick: If we want to be able to change this at runtime in the settings, it would need to become a getter I think because right now this would be stale if changed after the init of the app, probably same for mobile

Not sure we need to change that now but if we want that to be dynamic and work as expected from the settings we might want to have a follow up task

libs/ledger-live-common/src/wallet-api/Exchange/quotes/service/fetchQuotes.ts
```diff
@@ -1,95 +1,91 @@
...
-    if (axios.isAxiosError(error) && error.response) {
-      return { rawQuotes: [], providerErrors: [] };
+    if (result.error) {
+      if (getHttpStatus(result.error) === undefined) {
+        throw new SwapQuotesRequestFailed(
+          `swap /quote request failed: ${describeError(result.error)}`,
+          result.error,
+        );
+      }
+      return emptyResult();
```
Justkant, line 84: **Behaviour change is real, and wider than the description suggests — wants PTX sign-off before merge.**

I checked the premise and it holds: `@ledgerhq/live-network` installs its interceptors on the *global* axios (`libs/live-network/src/network.ts:105-106`), so in the apps the old `axios.isAxiosError(error) && error.response` branch was effectively dead — 4xx/5xx arrived as `LedgerAPI4xx`/`LedgerAPI5xx` and were rethrown to the live app as an error.

So this now swallows more than a 502:
- a `400` from malformed params,
- a `401`/`403` that survives the auth adapter's refresh-and-retry,

both render as "no quotes for this pair", indistinguishable from a genuinely empty result.

Combined with the removed `log("network-error", …)` from the old interceptor, an aggregator outage becomes invisible in *both* the UI and support logs. I'd not defer the logging follow-up past this PR — it's the only remaining signal that this path failed.

domain/api/swap-quotes/src/api.ts
```diff
@@ -0,0 +1,111 @@
...
+  endpoints: build => ({
+    fetchQuotes: build.query<FetchQuotesResult, FetchQuotesQueryArgs>({
+      query: ({ providers, quotesInput, counterValueCurrency, customHeaders }) => ({
+        url: "/quote",
+        params: buildQuotesParams(providers, quotesInput, counterValueCurrency),
+        headers: { ...(customHeaders ?? {}) },
+      }),
+      // Keeps a live-app token out of the cache key. Consequence: concurrent
+      // requests differing only by `customHeaders` share one in-flight call.
+      serializeQueryArgs: ({ queryArgs }) => {
```
Justkant, line 100: **In-flight dedupe means one caller can receive a response fetched with another caller's headers.**

RTK Query's `condition` checks `status === 'pending'` *before* it honours `forceRefetch`, so two concurrent calls with identical params but different `customHeaders` share one request — and the second caller gets the body fetched with the first caller's headers. `headers?: Array<[string, string]>` is arbitrary live-app-supplied input (`libs/exchange-module/src/types.ts:139`), so this isn't hypothetical input we control.

Today that's probably benign, but it's only benign for as long as no header influences the response. The moment one does (partner key, affiliate id, experiment bucket) it's silently wrong data crossing between requests.

Two ways out: fold a hash of the headers into the cache key (keeps the token itself out of the key, which is the stated goal), or make "headers must not affect the response body" an explicit contract on `FetchQuotesQueryArgs` rather than a consequence noted in a comment.

shared/api-services/src/services/swap/api.ts
```diff
@@ -0,0 +1,59 @@
...
+      return headers;
+    },
+  })(args, api, extraOptions);
+};
+
+/**
+ * Endpoint-less swap aggregator api. Register it in the store; use cases add endpoints and tags to
+ * this same object — see {@link https://github.com/LedgerHQ/ledger-live/blob/develop/shared/api-services/README.md}.
+ */
+export const swapApi = createApi({
```
Justkant, line 52: **Missing `api.test.ts` for this service.**

All four sibling services have one (`cal`, `coinmarketcap`, `countervalues`, `push-devices`), and the package README's layout lists it explicitly: `api.test.ts # reducer path, zero endpoints, extraArgument validation`.

It matters more here than for the siblings, for two reasons:
- this is the **first** consumer of `createAuthenticatedBaseQuery` outside `shared/auth`, so the wiring of `prepareHeaders` *through* the auth wrapper (header survives both the token and no-token branches) is only covered indirectly, from another package;
- `swapApiExtra`'s fail-fast `parse` is what turns an empty env var into a throw at app init — that's untested at the layer that owns it.

The `@domain/api-swap-quotes` tests do exercise both paths, but the convention gap is in the package that owns the base query.

libs/ledger-live-common/src/wallet-api/Exchange/quotes/service/fetchQuotes.ts
```diff
@@ -1,95 +1,91 @@
...
-    Accept: "application/json",
-    ...(customHeaders ? Object.fromEntries(customHeaders) : {}),
-  };
+  const promise = dispatch(
+    swapQuotesApi.endpoints.fetchQuotes.initiate(
+      {
+        providers,
+        quotesInput,
+        counterValueCurrency,
+        customHeaders: customHeaders ? Object.fromEntries(customHeaders) : undefined,
```
Justkant, line 65: `customHeaders` are kept out of the cache key (good), but they still travel in `originalArgs`, and LLD's redux logger records the query arg for RTK Query actions:

```ts
// apps/ledger-live-desktop/src/renderer/logger/index.ts:122-129
if (action.type.includes("/executeQuery/") || …) {
  if (action.meta.arg) logData.queryArg = action.meta.arg;
}
```

So with redux logging enabled to gather a support log, live-app-supplied headers land in the file — along with `sendAddress`/`receiveAddress` from `quotesInput`. Redacting the headers before dispatch (or filtering them in the logger) would close the loop on the intent behind keeping them out of the key.

domain/api/swap-quotes/src/store/store.ts
```diff
@@ -0,0 +1,47 @@
...
+ * separate copies still share one dispatch.
+ */
+import type { ThunkDispatch, UnknownAction } from "@reduxjs/toolkit";
+
+export type SwapQuotesDispatch = ThunkDispatch<unknown, unknown, UnknownAction>;
+
+declare global {
+  // `var` is required: TypeScript builds `typeof globalThis` from global `var`
+  // declarations, so an `interface GlobalThis` would augment nothing.
+  var __ledgerSwapQuotesDispatch: SwapQuotesDispatch | undefined;
```
Justkant, line 20: On the `globalThis` singleton — I know it's on the follow-up list, but I'd push for doing it in this PR rather than after.

`GetQuotesContext` is already a typed DI seam that *every* caller of `getQuotes` builds (`accounts`, `spotPrices`, `locale`, `counterValueCurrency`, `deviceModelId`), so threading `dispatch` through it is a small change that removes the global entirely and makes a missing registration a compile error.

As it stands the failure mode is runtime-only: any host that registers the reducer but forgets `setSwapQuotesStore` throws on the first quote request. `rtkQueryApi.test.ts` pins the reducer paths in both apps, but nothing pins the startup call — so the half of the wiring that has no compile-time or test-time guard is exactly the half that's easy to forget. If the DI change is too much for this PR, a test per app asserting init calls `setSwapQuotesStore` is cheap insurance.

domain/api/swap-quotes/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "description": "Domain API client for the swap quotes aggregator: RTK Query endpoint for the /quote request",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "exports": {
+    ".": "./src/index.ts",
+    "./store": "./src/store/store.ts",
+    "./store.standalone": "./src/store/store.standalone.ts",
+    "./package.json": "./package.json"
```
Justkant, line 12: Nit: the fixtures are duplicated — `domain/api/swap-quotes/src/fixtures/*` and `libs/ledger-live-common/src/wallet-api/Exchange/quotes/fixtures/*` are near-identical (`makeQuotesInput` byte-for-byte; `makeRawQuote` differs only in parsing through the schema).

This `exports` map is what forces the copy — there's no `./fixtures` subpath, so live-common can't reach them. Adding one would let live-common import the schema-parsed versions and keep a single source of truth.

domain/api/swap-quotes/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "scripts": {
+    "typecheck": "tsc --noEmit",
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "coverage": "jest --coverage",
+    "unimported": "pnpm knip --directory ../../.. -W domain/api/swap-quotes"
+  },
+  "dependencies": {
+    "@reduxjs/toolkit": "catalog:",
+    "@shared/api-services": "workspace:^",
```
Justkant, line 24: Nit: every sibling domain package pins workspace deps as `workspace:*` (e.g. `domain/api/push-devices`, `currency-token`, `market-sentiment`) — `workspace:^` here is the odd one out. Applies to `@shared/auth` on the next line too.

Justkant (commented):
A few points beyond the base-URL one already raised. Architecture and test discipline look good — the `shared/api-services` service mirrors `cal/api.ts` closely, `SwapQuotesRequestFailed` matches the `errors` convention, and the schema-first types with fixtures parsed through the schema are a real drift guard. Inline comments below; two repo-level notes here:

**The PR description is out of date w.r.t. the diff.** The body places the endpoint in `libs/ledger-live-common/.../state-manager/api.ts` and lists "belongs in a `domain/api-swap-quotes` package" under *deferred* follow-ups — but the diff creates that package and live-common is now shims. The changeset gets it right; the body will mis-scope reviewers.

**No CODEOWNERS entry for the new paths.** `domain/api/swap-quotes/` and `shared/api-services/src/services/swap/` aren't covered, so PTX won't be auto-requested even though they own the other swap paths — and this PR explicitly wants a PTX call on the error-mapping change. (`libs/.../Exchange/quotes/` isn't owned today either, so this is pre-existing rather than introduced, but worth fixing while adding new directories.)

domain/api/swap-quotes/src/api.ts
```diff
@@ -0,0 +1,111 @@
...
+  // Validate each row against the one schema it should match, rather than the
+  // union: a union failure reports both branches and buries the real issue.
+  for (const [schema, group] of [
+    [RawQuoteSchema, result.rawQuotes],
+    [RawQuoteErrorSchema, result.providerErrors],
+  ] as const) {
+    for (const row of group) {
+      const parsed = schema.safeParse(row);
+      if (!parsed.success) {
+        console?.warn("swapQuotesApi: /quote row did not match the schema", parsed.error.issues);
```
Justkant, line 75: Nit, non-blocking: this re-parses every row of every response purely to emit a warning, on each quote poll across N providers. Almost certainly negligible next to the request itself, and the "never discard a row" rationale is sound — just noting it's per-poll work with no behavioural effect, so it's worth revisiting when the schema is tightened to drop-invalid.

(`console?.warn` matches the existing `shared/auth` idiom, so no objection there.)

Justkant (approved):

PR #18773, opened 2026-06-19
[LWDM] fix(market): support desktop countervalue fallbacks

deepyjr:
### ✅ Checklist

  - `@ledgerhq/live-common` market countervalue resolution and USD fallback helpers
  - Desktop Asset Detail price chart with unsupported fiat and crypto countervalues
  - Desktop Market list and Global Market Cap cards

### 📝 Description

This PR adds the shared market countervalue fallback used by desktop when the markets APIs cannot serve the selected countervalue directly. Unsupported fiats such as COP wait for the supported countervalue list before fetching, while crypto countervalues such as BTC/ETH fetch USD data and rescale it with the USD→countervalue spot rate.

On desktop, Asset Detail charts, the Market list, and the Global Market Cap card now use the shared fallback path and keep Ledger unit formatting.

image

image

image

image

### ❓ Context

- **JIRA or GitHub link**: [LIVE-32591](https://ledgerhq.atlassian.net/browse/LIVE-32591)

---

[LIVE-32591]: https://ledgerhq.atlassian.net/browse/LIVE-32591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/counterValues/hooks/useUsdToFiatRate.ts
```diff
@@ -8,6 +8,10 @@ export type UsdToFiatRate = {
   rate: number | null;
```
gre-ledger, line 8: i'm surprised to see a new `ledger-live-common/src/counterValues` and curious to understand why we didn't put the work in `live-countervalues-react` ? because that ledger-live-common/src/countervalues used to exist and eventually was externalized / decoupled into that dedicated library !

We will need to make sure to cover this when we work on a the next /domain/ entities and api for the countervalues rearchitecture.

https://ledgerhq.atlassian.net/browse/LIVE-32360

atm it seems platform is owning this folder , maybe shouldn't 🤔

gre-ledger: @mcayuelas-ledger do you think it is easy to make counterValues go back to the external lib, to facilitate the future migration to the DDD new arch, or would it create new dependencies to need to be added? maybe there were a reason we did it here 🤔 

To me live-common is legacy and it is always better to create new libs than adding more folders. we may need to write it more clearly into the project READMEs etc

mcayuelas-ledger: In itself, it is possible: we simply need to create a standalone folder. We only need a single dependency on RTK. The only problem is that `live-countervalues-react` uses the old version, whereas we want RTK; this is therefore the only obstacle to using it.

mcayuelas-ledger (approved):

gre-ledger (approved):

PR #18915, opened 2026-06-24
[LWDM] fix: tezos fees

estrauser-ledger:
### ✅ Checklist

  - ...

### 📝 Description

Accounts with smart contract interactions (e.g. sending XTZ to a contract that forwards it) had inflated native balances. Two issues in `listOperations`:

1. **Internal sub-transactions counted as account OUTs:** When a user sends XTZ to a contract, TzKT returns both the top-level tx (user → contract) and the internal txs (contract → third    party) under the user's operations. `listOperations` included all of them, double-counting the spent amount.

2. **Token ops duplicated parent fees:** Token transfer operations copied `tx.fees` from the parent native transaction. When fees were summed across sub-operations per tx hash, the same fee was counted twice (once on the native op, once on the token op).

#### Before

For an account sending 100,000 mutez to a contract that forwards 90,000 + 10,000:

| Operation | type | value | fees | source |
|-----------|------|-------|------|--------|
| user → contract | OUT | 100,000 | 17,873 | top-level |
| contract → addr1 | OUT | 90,000 | 0 | internal |
| contract → addr2 | OUT | 10,000 | 0 | internal |

Native impact: -(100,000 + 90,000 + 10,000) - 17,873 = **-217,873** (wrong — should be -117,873)
#### After

| Operation | type | value | fees | source |
|-----------|------|-------|------|--------|
| user → contract | OUT | 100,000 | 17,873 | top-level |
| ~~contract → addr1~~ | — | — | — | filtered |
| ~~contract → addr2~~ | — | — | — | filtered |

Native impact: -100,000 - 17,873 = **-117,873** (correct)

#### Changes

- **`keepNativeOp`**: Filter out transactions where the account is neither sender nor target (internal contract sub-operations)
- **`convertTokenOperation`**: Set `fees: 0n` — the fee is already on the parent native operation
- **Tests**: Use realistic addresses so transaction ops pass the new sender/target filter; add `status: "applied"` to the base `transfer` fixture

### ❓ Context

---

### DIFF
--- a/.changeset/ten-baboons-nail.md
+++ b/.changeset/ten-baboons-nail.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-tezos": minor
+---
+
+Fix native balance double-counting in listOperations by filtering out internal contract sub-transactions and zeroing token op fees
--- a/libs/coin-modules/coin-tezos/src/logic/listOperations.test.ts
+++ b/libs/coin-modules/coin-tezos/src/logic/listOperations.test.ts
@@ -388,6 +388,38 @@ describe("listOperations", () => {
     });
   });
 
+  it("filters out internal sub-transactions where account is neither sender nor target", async () => {
+    const contractAddress = "KT1WPEis2WhAc2FciM2tZVn8qe6pCBe9HkDp";
+    const thirdPartyAddress = "tz1Pci9FqMTqADRXqgD3ZDsdfEFcPqMTqA8D";
+
+    // Top-level: user sends to contract
+    const topLevelTx: APITransactionType = {
+      ...transfer,
+      id: 700,
+      amount: 100000,
+      target: { address: contractAddress },
+    };
+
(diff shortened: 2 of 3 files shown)


amaslakov (approved):
One thing: did you check that `listOperations.integ.test.ts` passes fine after this change? It asserts that `tx.fees => 0n`
https://github.com/LedgerHQ/ledger-live/blob/425667848de31fda5ef476c6398b38176f954534/libs/coin-modules/coin-tezos/src/logic/listOperations.integ.test.ts#L134

estrauser-ledger:
> One thing: did you check that `listOperations.integ.test.ts` passes fine after this change? It asserts that `tx.fees => 0n`

Yes :) 
Currently, there's a incoherent behavior where on token operations, fees are duplicated from the native op, and on other types, fees should be summed. so nnow it's impossible to compute balance from operations. It's what I'm trying to fix right now. So for transactions with tokens, I'll have one of them fees at 0 so they're not duplicated

PR #18944, opened 2026-06-24
[LWD] fix(lld): make full-screen Live App webview load resilient on cold start (LIVE-33113)

ysitbon:
### ✅ Checklist

  - **Shared `Web3AppWebview` code** → affects **every full-screen Live App**

### ❓ Context

- **JIRA or GitHub link**: [LIVE-33113](https://ledgerhq.atlassian.net/browse/LIVE-33113)
- **ADR link (if any)**: N/A

[LIVE-33113]: https://ledgerhq.atlassian.net/browse/LIVE-33113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/recover-webview-infinite-loader.md
+++ b/.changeset/recover-webview-infinite-loader.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": minor
+---
+
+Fix full-screen live apps (e.g. Recover) getting stuck on an infinite loader on cold start. The webview's `did-finish-load` listener was attached via an effect keyed on a stable ref, so on a cold start it could attach after the webview had already finished loading and miss the event, leaving `widgetLoaded` false forever. The listener now keys off the mounted webview node and recovers if the load already completed.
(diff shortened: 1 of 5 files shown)


ysitbon:
Addressed the review feedback in e1fac5920cb:

1. **Missed `dom-ready` on the recovery path** (WalletAPIWebview + PlatformAPIWebview) — good catch, this had a security implication: if the effect attached after load, `handleDomReady` never ran and the main-process manifest-domain `will-navigate` guard (wired via the `webview-dom-ready` IPC) was never installed. The recovery branch now replays `handleDomReady()` before `onLoad()`/`handleLoad()`. Safe because `dom-ready` always precedes `did-finish-load`, so a settled load (`!isLoading()`) implies both fired — and `getWebContentsId()`/`isLoading()` only succeed post-dom-ready (the `try/catch` covers the not-yet-ready case).

2. **Stale `oxlint-disable` on the PlatformAPIWebview effect** — removed. Confirmed `react-hooks/exhaustive-deps` stays clean with the explicit `[handleLoad, handleDomReady, webviewNode]` array.

PR #19007, opened 2026-06-25
[LWDM] feat(apps): inject domain crypto-currency registry at bootstrap (LIVE-32900)

ysitbon:
### ✅ Checklist

- [x] **Covered by automatic tests.** A guard test in `currencies.domain-parity.test.ts` injects the real domain array and asserts `getCryptoCurrencyById` resolves **every** id to the domain object; combined with #18933's ambiguous-ticker / scheme parity tests this covers the injected-store lookup path. The `"osmosis"` → `"osmo"` production fix (`coin-modules/registry.ts:21` over the cosmos loader `supportedCoins`) is covered transitively (`getCryptoCurrencyById("osmo")` resolves under injection). _Not_ in this PR: full functional e2e / app-boot — see QA focus below.
  - The runtime source of truth for crypto-currency **data** in all 5 apps switches from the bundled `@ledgerhq/cryptoassets` registry to `@domain/entity-currency-crypto` (values are parity-tested equal; bundled stays as fallback). **QA focus:** app bootstrap (no throw at `registerAllCoins()`), currency lists, add-account search, token accounts, send/receive (BTC + an EVM token), and countervalues for ambiguous tickers (ETH/BNB/DOT/XTZ/CRO).
  
### E2E run after merge
- https://github.com/LedgerHQ/ledger-live/actions/runs/28428767451
- https://github.com/LedgerHQ/ledger-live/actions/runs/28428777114

### 📝 Description

PR2 of the currency service-locator thread. [LIVE-32899](https://ledgerhq.atlassian.net/browse/LIVE-32899) made the `@ledgerhq/cryptoassets` currency registry injectable (accessors read `getInjectedCurrenciesStore() ?? bundledCurrenciesStore`). Until now nothing injected, so every app ran on the bundled data. This PR makes each app inject the **domain** registry at bootstrap, so the runtime source of truth for currency data becomes domain — the bundled data remains as the fallback and is kept iso by the LIVE-31915 parity test.

**What changed**

- **Injection (5 apps).** `setCryptoCurrenciesStore(Object.values(CRYPTO_CURRENCIES_REGISTRY) as unknown as CryptoCurrency[])` added at each app's `live-common-setup*.ts` bootstrap, before `registerAllCoins()` / `registerCoinModules()`:
  - desktop → `src/live-common-setup-base.ts` (shared by main + renderer)
  - mobile → `src/live-common-setup.ts`
  - cli → `src/live-common-setup-base.ts`
  - wallet-cli → `src/live-common-setup.ts`
  - web-tools → `src/live-common-setup.ts`
  - plus `@domain/entity-currency-crypto` added to each app's `package.json`.
  - _Note:_ not `config/bridge-setup.ts` (only desktop+mobile have it; it's the redux-coupled CAL **token** store, called late). `live-common-setup*.ts` exists for all 5 and runs early.
- **Type bridge.** The cast is required: legacy `coinType` is a real `enum CoinType` and `explorerId` a `LedgerExplorerId` union, so the domain type is structurally wider, not assignable. Values are parity-guaranteed, so the cast is sound.
- **Alias-key fix.** Once injected, `getCryptoCurrencyById` resolves only by canonical `.id` (no per-key fallback). The legacy alias `"osmosis"` (canonical id `"osmo"`) was migrated at 3 sites — most importantly the **production** path `libs/ledger-live-common/src/coin-modules/registry.ts:21`, which resolves every loader's `supportedCoins` through `getCryptoCurrencyById` at `registerAllCoins()` and would otherwise throw. `groestlcoin`/`lbry` aliases are never used as lookup keys, so they needed no change.

**Verification:** 5-app typecheck green; `pnpm lint:boundaries` green (apps may depend on `@domain/*`); cryptoassets parity/guard suite 423 pass.

> ⚠️ **Stacked on #18933 — do not merge yet.** This branch is based on [#18933](https://github.com/LedgerHQ/ledger-live/pull/18933) ([LIVE-33115](https://ledgerhq.atlassian.net/browse/LIVE-33115), order-independent `findCryptoCurrencyByTicker`). Without that fix, injecting the alphabetical domain array flips 5 ambiguous tickers. **Do not merge until #18933 lands in `develop`; then rebase onto `develop`** (the #18933 commits will drop out of this diff once it merges).

### ❓ Context

- **JIRA or GitHub link**: [LIVE-32900](https://ledgerhq.atlassian.net/browse/LIVE-32900) · depends on [LIVE-33115](https://ledgerhq.atlassian.net/browse/LIVE-33115) ([#18933](https://github.com/LedgerHQ/ledger-live/pull/18933)) · parent epic [LIVE-31901](https://ledgerhq.atlassian.net/browse/LIVE-31901)
- **ADR link (if any)**: n/a

[LIVE-32899]: https://ledgerhq.atlassian.net/browse/LIVE-32899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/cli/src/live-common-setup-base.ts
+++ b/apps/cli/src/live-common-setup-base.ts
@@ -6,6 +6,14 @@ import { registerAllCoins } from "@ledgerhq/live-common/coin-modules/load-all-co
 import { setWalletAPIVersion } from "@ledgerhq/live-common/wallet-api/version";
 import { WALLET_API_VERSION } from "@ledgerhq/live-common/wallet-api/constants";
 import BigNumber from "bignumber.js";
+import { setCryptoCurrenciesStore } from "@ledgerhq/cryptoassets";
+import {
+  CRYPTO_CURRENCIES_REGISTRY,
+  CRYPTO_CURRENCY_ALIASES,
+} from "@domain/entity-currency-crypto";
+
+// The domain registry is the runtime source of truth for currency data.
+setCryptoCurrenciesStore(Object.values(CRYPTO_CURRENCIES_REGISTRY), CRYPTO_CURRENCY_ALIASES);
 
 setWalletAPIVersion(WALLET_API_VERSION);
 
--- a/domain/entity/currency-crypto/src/schema.ts
+++ b/domain/entity/currency-crypto/src/schema.ts
@@ -98,7 +98,10 @@ export const CryptoCurrencySchema = z.object({
   explorerViews: z.array(ExplorerViewSchema),
   /** Ticker displayed on the device (when different from `ticker`). */
   deviceTicker: z.string().optional(),
-  /** Id used to connect to the Ledger explorer endpoint (when different from the currency id and ticker). */
+  /**
+   * Id used to connect to the Ledger explorer endpoint (when different from the currency id and ticker).
+   * @deprecated Kept only for backward compatibility; the explorer-id concept is being phased out.
+   */
   explorerId: z.string().optional(),
   /** Token standards supported by this chain (e.g. `["erc20"]`). */
   tokenTypes: z.array(z.string()).optional(),
--- a/libs/ledgerjs/packages/cryptoassets/src/currencies-store.ts
+++ b/libs/ledgerjs/packages/cryptoassets/src/currencies-store.ts
@@ -82,15 +82,40 @@ function buildCryptoCurrenciesStore(currencies: CryptoCurrency[]): CryptoCurrenc
  * Injects the crypto-currency registry from the canonical currency list.
  * This should be called once during application initialization.
  *
- * The caller only provides the currencies; the by-id/ticker/scheme indices and the
- * dev/prod arrays are derived here so they stay consistent by construction.
+ * The caller provides the currencies; the by-id/ticker/scheme indices and the dev/prod arrays are
+ * derived here so they stay consistent by construction. `aliases` maps a legacy alias key to a
+ * canonical id (e.g. `{ osmosis: "osmo" }`) so id lookups by that alias keep resolving, as they do
+ * against the bundled map which exposes the same legacy aliases.
  *
  * Uses globalThis to ensure a single shared reference across all module instances,
  * which is critical when coin-modules are lazy-loaded and may resolve to separate
  * module copies.
  */
-export function setCryptoCurrenciesStore(currencies: CryptoCurrency[]): void {
-  globalThis.__ledgerCryptoCurrenciesStore = buildCryptoCurrenciesStore(currencies);
+export function setCryptoCurrenciesStore(
+  currencies: CryptoCurrency[],
+  aliases: Record<string, string> = {},
+): void {
+  const store = buildCryptoCurrenciesStore(currencies);
+  for (const [alias, id] of Object.entries(aliases)) {
+    // Fail fast on a misconfigured alias map at injection time, rather than silently dropping it and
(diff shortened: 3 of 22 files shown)


apps/cli/src/live-common-setup-base.ts
```diff
@@ -6,6 +6,12 @@ import { registerAllCoins } from "@ledgerhq/live-common/coin-modules/load-all-co
 import { setWalletAPIVersion } from "@ledgerhq/live-common/wallet-api/version";
 import { WALLET_API_VERSION } from "@ledgerhq/live-common/wallet-api/constants";
 import BigNumber from "bignumber.js";
+import { setCryptoCurrenciesStore } from "@ledgerhq/cryptoassets";
+import { CRYPTO_CURRENCIES_REGISTRY } from "@domain/entity-currency-crypto";
+import type { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
+
+// Domain registry becomes the runtime source of currency data (cast bridges the parity-tested nominal type gap).
+setCryptoCurrenciesStore(Object.values(CRYPTO_CURRENCIES_REGISTRY) as unknown as CryptoCurrency[]);
```
gre-ledger: i'm a bit scared at accepting the cast here. why can't we make the typechecker happy here? it may hide real problem.

ysitbon: Dropped the cast entirely

domain/entity/currency-crypto/src/schema.ts
```diff
@@ -36,6 +36,39 @@ export const BitcoinLikeInfoSchema = z.object({
...
+  "zen",
+  "avax",
+  "eth",
+  "eth_sepolia",
+  "eth_hoodi",
+  "etc",
+  "matic",
+  "matic_amoy",
+  "bnb",
+]);
```
gre-ledger: let's see why we need this / confirm with coin team / because to me it's legacy to maintain this list. maybe we could afford a loosely typed string, but ok to stay compatible here.

gre-ledger (dismissed):
ok. one note on the type but not blocker

domain/entity/currency-crypto/src/schema.ts
ysitbon: Moving it as a string.

libs/ledgerjs/packages/cryptoassets/src/currencies-store.ts
```diff
@@ -82,15 +82,25 @@ function buildCryptoCurrenciesStore(currencies: CryptoCurrency[]): CryptoCurrenc
...
+  currencies: CryptoCurrency[],
+  aliases: Record<string, string> = {},
+): void {
+  const store = buildCryptoCurrenciesStore(currencies);
+  for (const [alias, id] of Object.entries(aliases)) {
+    const currency = store.cryptocurrenciesById[id];
+    if (currency) store.cryptocurrenciesById[alias] = currency;
+  }
+  globalThis.__ledgerCryptoCurrenciesStore = store;
 }
```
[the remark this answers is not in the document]
ysitbon: Addressed in f495213: setCryptoCurrenciesStore now fails fast at injection when an alias points to an unknown currency id or collides with an existing id, instead of silently dropping it. Added unit tests covering both cases.

gre-ledger (approved):

francois-guerin-ledger (approved):

mcayuelas-ledger (approved):

PR #19052, opened 2026-06-26
Feat/live 32903 send analytics

lpaquet-ledger:
### ✅ Checklist

  - ...

### 📝 Description

add send analytics to cli

### ❓ Context

https://ledgerhq.atlassian.net/browse/NTTVS-455

### DIFF
--- a/apps/wallet-cli/src/analytics/send-analytics.ts
+++ b/apps/wallet-cli/src/analytics/send-analytics.ts
@@ -19,11 +19,11 @@ export function trackSendStarted(p: {
   });
 }
 
-export function trackSendFailed(p: { errorCode: string; errorMessage: string }): void {
+export function trackSendFailed(p: { errorCode: string; errorName: string }): void {
   track("send_failed", {
     page: SEND,
     errorCode: p.errorCode,
-    errorMessage: p.errorMessage,
+    errorName: p.errorName,
   });
 }
 
--- a/apps/wallet-cli/src/analytics/send-analytics.ts
+++ b/apps/wallet-cli/src/analytics/send-analytics.ts
@@ -1,4 +1,4 @@
-import { track } from "../analytics/segment";
+import { track } from "./segment";
 
 const SEND = "Send";
 
@@ -19,10 +19,11 @@ export function trackSendStarted(p: {
   });
 }
 
-export function trackSendFailed(p: { errorCode: string }): void {
+export function trackSendFailed(p: { errorCode: string; errorMessage: string }): void {
   track("send_failed", {
     page: SEND,
     errorCode: p.errorCode,
+    errorMessage: p.errorMessage,
   });
 }
 
--- a/apps/wallet-cli/src/commands/send.ts
+++ b/apps/wallet-cli/src/commands/send.ts
@@ -155,25 +155,18 @@ async function runLiveSend({
         );
       }
 
-      try {
-        await runObservable({
-          source$: wallet.send(descriptor, intent, {
-            deviceId: WALLET_CLI_DMK_DEVICE_ID,
-            deviceModelId,
-          }),
-          onNext: event => out.sendEvent(event),
-          mapError: error =>
-            WalletCliDeviceError.fromKnownDeviceError(error, {
-              expectedApp: managerAppName,
-              rejectedContext: "sign",
-            }) ?? error,
-        });
-      } catch (error) {
-        if (error instanceof WalletCliDeviceError && error.state.code === "rejected") {
-          trackSendRejected({ network, device: deviceModelId });
-        }
-        throw error;
-      }
+      await runObservable({
+        source$: wallet.send(descriptor, intent, {
+          deviceId: WALLET_CLI_DMK_DEVICE_ID,
+          deviceModelId,
+        }),
+        onNext: event => out.sendEvent(event),
+        mapError: error =>
+          WalletCliDeviceError.fromKnownDeviceError(error, {
+            expectedApp: managerAppName,
+            rejectedContext: "sign",
+          }) ?? error,
+      });
 
       out.sendComplete();
       trackSendCompleted({ network, assetClass, amount, device: deviceModelId });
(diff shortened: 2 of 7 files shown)


apps/wallet-cli/src/analytics/send-analytics.ts
Justkant, line 1: I think we'll need to find a better place for those, I don't think they should live in the commands folder

We could create an analytics folder for this, can be done later

apps/wallet-cli/src/analytics/send-analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+    assetClass: p.assetClass,
+    dryRun: p.dryRun,
+    device: p.device,
+  });
+}
+
+export function trackSendFailed(p: { errorCode: string }): void {
+  track("send_failed", {
+    page: SEND,
+    errorCode: p.errorCode,
```
Justkant, line 25: And maybe the error message no ?

apps/wallet-cli/src/commands/send.ts
```diff
@@ -79,6 +89,15 @@ function buildIntentData(currencyId: string, flags: SendFlags) {
...
 }
 
+/**
+ * Best-effort native/token classification for analytics: a ticker matching the account's
+ * native currency is native, anything else (or a missing ticker) is treated as a token send.
+ */
+function classifySendAssetClass(currencyId: string): SendAssetClass {
+  const currency = findCryptoCurrencyById(currencyId);
+  return currency ? (currency.type === "CryptoCurrency" ? "native" : "token") : "token";
+}
```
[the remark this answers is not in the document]
Justkant: Looks like the copilot comment is correct, wdyt ?

apps/wallet-cli/src/commands/send.ts
```diff
@@ -211,25 +244,36 @@ export default defineCommand({
...
+          deviceTimeoutMs: flags["device-timeout"],
+          out,
+          network: ctx.network,
+          assetClass,
+          amount: flags.amount,
+        });
+      } catch (error) {
+        trackSendFailed({ errorCode: error instanceof Error ? error.message : String(error) });
+        throw error;
       }
```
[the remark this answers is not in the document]
Justkant: We need error name too

apps/wallet-cli/src/commands/send.ts
```diff
@@ -211,25 +244,36 @@ export default defineCommand({
...
+          managerAppName,
+          deviceTimeoutMs: flags["device-timeout"],
+          out,
+          network: ctx.network,
+          assetClass,
+          amount: flags.amount,
+        });
+      } catch (error) {
+        trackSendFailed({ errorCode: error instanceof Error ? error.message : String(error) });
+        throw error;
```
Justkant, line 285: I think we should consider some kind of util to wrap a run with start/completed/failed tracking

apps/wallet-cli/src/cli.ts
```diff
@@ -53,6 +53,7 @@ function normalizeNegatedFlags(argv: string[]): string[] {
 
 if (import.meta.main) {
   let exitCode = 0;
+  startAnalytics();
```
Justkant: You want to remove this it's duplicated below

apps/wallet-cli/src/commands/send.ts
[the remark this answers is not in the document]
lpaquet-ledger: Strange already did this before
Done

apps/wallet-cli/src/analytics/send-analytics.ts
lpaquet-ledger: Followed what was asked from tracking team but can add message

apps/wallet-cli/src/analytics/send-analytics.ts
lpaquet-ledger: done

apps/wallet-cli/src/commands/send.ts
Justkant: This has been done in the earn PR

apps/wallet-cli/src/commands/send.ts
```diff
@@ -121,20 +155,28 @@ async function runLiveSend({
...
+              expectedApp: managerAppName,
+              rejectedContext: "sign",
+            }) ?? error,
+        });
+      } catch (error) {
+        if (error instanceof WalletCliDeviceError && error.state.code === "rejected") {
+          trackSendRejected({ network, device: deviceModelId });
+        }
+        throw error;
+      }
```
Justkant: Do we really need that try/catch here instead of just using the try/catch below wrapping the runLiveSend ?

Putting it here means you will get the trackSendRejected and trackSendFailed for a single error

apps/wallet-cli/src/analytics/send-analytics.ts
Justkant: Sorry I said message but it could contain PII so better to only send the error name only with the code

apps/wallet-cli/src/commands/send.ts
```diff
@@ -211,25 +246,44 @@ export default defineCommand({
...
+          deviceTimeoutMs: flags["device-timeout"],
+          out,
+          network: ctx.network,
+          assetClass,
+          amount: flags.amount,
+        });
+      } catch (error) {
+        if (error instanceof WalletCliDeviceError && error.state.code === "rejected") {
+          const device = await getWalletCliDeviceModelId();
+          trackSendRejected({ network: ctx.network, device });
```
Justkant: Actually what I asked won't work I think, because the transport will be closed when we're done with the runLiveSend

We might be able to have this info by adding the deviceModelId in the WalletCliDeviceError, I'm checking a bit the code to see if it's a good solution or if we should just use the try/catch you did before

Justkant (dismissed):

Justkant (dismissed):

Justkant (approved):

PR #19053, opened 2026-06-26
feat(currencies): add @domain/api-currency-token CAL token client

ysitbon:
### ✅ Checklist

- [x] **Covered by automatic tests.** 82 unit tests: Zod schema parsing, converter parity against the crypto registry, persistence (extract/restore/hash-validation + Zod blob validation), and RTK request/header/staging-selection assertions via a `fetch` spy.
- [ ] **Impact of the changes:** New **private** package only — **not wired into any app**, so there is no runtime impact yet.
  - Adds `@domain/api-currency-token`; no existing code path changes. The legacy `cal-client` / `api-token-converter` remain untouched and in use. QA focus: none for this PR — it will be consumed later by [LIVE-31914](https://ledgerhq.atlassian.net/browse/LIVE-31914).

### 📝 Description

Relocates the **CAL (Crypto Asset List) token-data client** out of `libs/ledgerjs/packages/cryptoassets/src/cal-client/*` + `api-token-converter.ts` into a new **private** package **`@domain/api-currency-token`** (`domain/api/currency-token/`). Part of the per-entity split of the currencies API layer — token = CAL, crypto = DADA, fiat = CVS ([LIVE-32269](https://ledgerhq.atlassian.net/browse/LIVE-32269)).

**Ships**

- `cryptoAssetsApi` RTK Query API: `findTokenById`, `findTokenByAddressInCurrency`, `getTokensSyncHash`, `getTokensData`.
- Zod response schema (reusing `UnitSchema` from `@domain/entity-currency-unit`) and the `convertApiToken` converter — parent crypto currency resolved from `@domain/entity-currency-crypto`'s registry, output typed on `@domain/entity-currency-token`.
- Zod-validated RTK Query persistence helpers (`parsePersistedCAL`, cache extract/restore, CAL hash validation).

**Design**

- **Injected config** — service URLs, client version and an optional logger come from the store's thunk `extraArgument` via `calApiExtra(...)`; the package owns no `@ledgerhq/live-env` / `@ledgerhq/logs` / `@ledgerhq/types-cryptoassets` dependency.
- **Additive / dual-maintained** — legacy `cal-client` + `api-token-converter` stay in place (still imported by coin-modules + DADA); this is a parallel, retyped copy. Not yet wired into the apps (store binding is [LIVE-31914](https://ledgerhq.atlassian.net/browse/LIVE-31914); single-source gate [LIVE-32346](https://ledgerhq.atlassian.net/browse/LIVE-32346)).

### ❓ Context

- **JIRA or GitHub link**: [LIVE-31904](https://ledgerhq.atlassian.net/browse/LIVE-31904)
- **ADR link (if any)**: N/A

---

[LIVE-31914]: https://ledgerhq.atlassian.net/browse/LIVE-31914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-32269]: https://ledgerhq.atlassian.net/browse/LIVE-32269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/domain/api/currency-token/README.md
+++ b/domain/api/currency-token/README.md
@@ -1,0 +1,47 @@
+# @domain/api-currency-token
+
+Domain API client for **token currencies**, backed by the Crypto Asset List (CAL). RTK Query
+endpoints and helpers, typed on the Zod-first `@domain/entity-currency-token` entity. Owns no
+env/config/logging dependency.
+
+Relocation of `libs/ledgerjs/packages/cryptoassets/src/cal-client/*` and `api-token-converter.ts`.
+
+- `schema.ts` — Zod schema for the CAL `/v1/tokens` response (`ApiTokenResponseSchema`), reusing
+  `UnitSchema` from `@domain/entity-currency-unit`.
+- `types.ts` — query-arg and pagination types (`TokenByIdParams`, `TokenByAddressInCurrencyParams`,
+  `GetTokensDataParams`, `TokensDataWithPagination`).
+- `converter.ts` — `convertApiToken(apiToken)` → `TokenCurrency`; resolves the parent crypto
+  currency from `@domain/entity-currency-crypto`'s registry and drops tokens with an unknown parent.
+- `api.ts` — `cryptoAssetsApi` (`createApi`) with `findTokenById`, `findTokenByAddressInCurrency`,
+  `getTokensSyncHash`, `getTokensData`. Service URLs, client version and an optional logger are read
+  from the store's thunk `extraArgument`; the app supplies them via `calApiExtra(...)` so this
+  package owns no env/config dependency.
+- `persistence.ts` — Zod-validated RTK Query cache extraction/restoration and CAL hash validation.
+- `internals/` — **package-private** (not re-exported from `index.ts`): `constants.ts` (header
+  names, retry/page-size/persistence-version constants, the `TOKEN_TAGS` cache tags) and `utils.ts`
+  (the derived `output` projection + response transform helpers used by `api.ts`).
+
+Store wiring (app side):
+
+```ts
+configureStore({
+  reducer: { [cryptoAssetsApi.reducerPath]: cryptoAssetsApi.reducer },
+  middleware: gdm =>
+    gdm({
+      thunk: {
+        extraArgument: calApiExtra({
+          // Pass the staging URL here when running in staging mode — the package has no staging switch.
+          calServiceUrl: getEnv("CAL_SERVICE_URL"),
+          ledgerClientVersion: getEnv("LEDGER_CLIENT_VERSION"),
+        }),
+      },
+    }).concat(cryptoAssetsApi.middleware),
+});
+```
+
+When a store hosts several currency APIs (token, crypto, fiat), merge their extras with object
+spread — `{ ...calApiExtra({ … }), ...cryptoApiExtra({ … }) }` yields the combined `extraArgument` type.
+
+The imperative store builder and the slice binding (`onQueryStarted`) live in
+`@features/platform-currencies`, which keeps the domain api pure (the domain layer cannot depend on a
+feature package per the module boundaries).
--- a/domain/api/currency-token/src/api.ts
+++ b/domain/api/currency-token/src/api.ts
@@ -6,6 +6,7 @@ import {
   type FetchArgs,
   type FetchBaseQueryError,
 } from "@reduxjs/toolkit/query/react";
+import { z } from "zod";
 import type { TokenCurrency } from "@domain/entity-currency-token";
 import {
   type GetTokensDataParams,
@@ -22,6 +23,10 @@ import {
   MAX_RETRIES,
   TOKEN_OUTPUT_FIELDS,
   TOKEN_TAGS,
+  commitHeaderMissingError,
+  currencyNotFoundError,
+  fetchCurrencyError,
+  fetchError,
   transformTokensResponse,
   validateAndTransformSingleTokenResponse,
 } from "./internals";
@@ -32,24 +37,32 @@ import {
  * owns no env/config/logging dependency. The app picks the prod or staging URL — there is no
  * staging switch in here.
  */
-export interface CalApiExtra {
-  calServiceUrl: string;
-  ledgerClientVersion: string;
-  logger?: (...args: unknown[]) => void;
-}
+const CalApiExtraSchema = z.object({
+  calServiceUrl: z.string().min(1),
+  ledgerClientVersion: z.string().min(1),
+  logger: z.custom<(...args: unknown[]) => void>().optional(),
+});
+
+export type CalApiExtra = z.infer<typeof CalApiExtraSchema>;
 
 /**
  * Builds this package's slice of the thunk `extraArgument`. RTK leaves `extraArgument` untyped, so
- * this is the one compile-checked entry point ensuring the CAL config is complete and correctly named.
+ * this is the one compile- and runtime-checked entry point: `parse` fails fast at app init if the CAL
+ * config is incomplete (e.g. an env var resolved to an empty string).
  */
 export function calApiExtra(extra: CalApiExtra): CalApiExtra {
-  return extra;
+  return CalApiExtraSchema.parse(extra);
+}
+
+/** Extracts the {@link CalApiExtra} from the `extraArgument` of the API. */
+function getCalExtra(api: { extra: unknown }): CalApiExtra {
+  return api.extra as CalApiExtra;
 }
 
 /** Reads the injected {@link CalApiExtra} and delegates to {@link fetchBaseQuery}. Wrapped in `retry`. */
 const calBaseQuery: BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError> = retry(
   (args, api, extraOptions) => {
-    const extra = api.extra as CalApiExtra;
+    const extra = getCalExtra(api);
     return fetchBaseQuery({
       baseUrl: extra.calServiceUrl,
       prepareHeaders: headers => {
@@ -103,7 +116,7 @@ export const cryptoAssetsApi = createApi({
 
     getTokensSyncHash: build.query<string, string>({
       async queryFn(currencyId, queryApi) {
-        const extra = queryApi.extra as CalApiExtra;
+        const extra = getCalExtra(queryApi);
         try {
           const url = new URL("/v1/currencies", extra.calServiceUrl);
           url.searchParams.set("output", "id");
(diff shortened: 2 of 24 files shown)


domain/api/currency-token/src/types.ts
```diff
@@ -0,0 +1,50 @@
+import type { TokenCurrency } from "@domain/entity-currency-token";
+
+export interface TokenByIdParams {
+  id: string;
+}
+
+/** Parameters for the `getTokensData` query. */
+export interface TokenByAddressInCurrencyParams {
```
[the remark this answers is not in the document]
ysitbon: Good catch — fixed in 251265632f4: the JSDoc now reads `Parameters for the findTokenByAddressInCurrency query`.

domain/api/currency-token/src/types.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/** Response data for the `getTokensData` query, including pagination. */
+export interface TokensDataWithPagination {
+  /** The list of token currencies. */
+  tokens: TokenCurrency[];
+  pagination: TokenPagination;
+}
+
+/** Parameters for the `getTokensData` query with pagination. */
+export interface TokenPagination {
```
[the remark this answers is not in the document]
ysitbon: Fixed in 251265632f4 — relabeled as `Pagination metadata returned with getTokensData responses` (it is response metadata, not query params).

domain/api/currency-token/README.md
```diff
@@ -0,0 +1,47 @@
...
+Domain API client for **token currencies**, backed by the Crypto Asset List (CAL). RTK Query
+endpoints and helpers, typed on the Zod-first `@domain/entity-currency-token` entity. Owns no
+env/config/logging dependency.
+
+Relocation of `libs/ledgerjs/packages/cryptoassets/src/cal-client/*` and `api-token-converter.ts`.
+
+- `schema.ts` — Zod schema for the CAL `/v1/tokens` response (`ApiTokenResponseSchema`), reusing
+  `UnitSchema` from `@domain/entity-currency-unit`.
+- `types.ts` — query-arg and pagination types (`TokenByIdParams`, `GetTokensDataParams`,
+  `TokensDataWithPagination`, `TokensDataTags`).
```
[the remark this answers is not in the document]
ysitbon: Fixed in 251265632f4 — `TokensDataTags` was replaced by the package-private `TOKEN_TAGS`; removed it from the README and listed the actual public types instead.

domain/api/currency-token/README.md
```diff
@@ -0,0 +1,47 @@
...
+- `converter.ts` — `convertApiToken(apiToken)` → `TokenCurrency`; resolves the parent crypto
+  currency from `@domain/entity-currency-crypto`'s registry and drops tokens with an unknown parent.
+- `api.ts` — `cryptoAssetsApi` (`createApi`) with `findTokenById`, `findTokenByAddressInCurrency`,
+  `getTokensSyncHash`, `getTokensData`. Service URLs, client version and an optional logger are read
+  from the store's thunk `extraArgument`; the app supplies them via `calApiExtra(...)` so this
+  package owns no env/config dependency.
+- `persistence.ts` — Zod-validated RTK Query cache extraction/restoration and CAL hash validation.
+- `internals.ts` — **package-private** (not re-exported from `index.ts`): endpoint paths, header
+  names, retry/page-size/persistence-version constants, the derived `output` projection, and the
+  response transform helpers used by `api.ts` and its co-located test.
```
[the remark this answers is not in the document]
ysitbon: Fixed in 251265632f4 — the README now documents the `internals/` directory (`constants.ts` + `utils.ts`).

domain/api/currency-token/src/persistence.ts
```diff
@@ -0,0 +1,276 @@
...
+  ttl: number,
+): PersistedTokenEntry[] {
+  const now = Date.now();
+  return tokens.filter(token => now - token.timestamp < ttl);
+}
+
+/**
+ * Restores persisted tokens into the RTK Query cache. Validates persisted hashes against current
+ * hashes and skips restoring currencies whose hash changed (or could not be validated).
+ */
```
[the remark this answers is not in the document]
ysitbon: Fixed in 251265632f4 — clarified the doc: currencies with no stored hash are restored unconditionally (backward-compat); only a changed hash or a failed hash fetch skips restore.

domain/api/currency-token/src/internals/constants.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/** RTK Query cache tags exposed by the CAL token api. */
+export const TOKEN_TAGS = ["Tokens"] as const;
+
+/** RTK Query reducer path for the CAL token API — stable; it keys the persisted cache. */
+export const CAL_REDUCER_PATH = "cryptoAssetsApi";
+
+/** CAL endpoint paths, relative to the injected service base URL. */
+export const TOKENS_PATH = "/v1/tokens";
+export const CURRENCIES_PATH = "/v1/currencies";
```
gre-ledger: i understand the idea of constants and most of them make sense. 
but for the api paths, i would tend to prefer them inlined in the API side. wdyt?

domain/api/currency-token/src/internals/constants.ts
```diff
@@ -0,0 +1,29 @@
...
+export const DEFAULT_PAGE_SIZE = 1000;
+
+/** Request header carrying the Ledger client version on every CAL request. */
+export const CLIENT_VERSION_HEADER = "X-Ledger-Client-Version";
+
+/** Response header holding the CAL commit hash, used for token-cache invalidation. */
+export const COMMIT_HEADER = "X-Ledger-Commit";
+
+/** Response header carrying the pagination cursor for the next page. */
+export const NEXT_CURSOR_HEADER = "x-ledger-next";
```
gre-ledger: I would tend to rename the header with a strict name that is exactly the header name like this:

`HEADER_{uppercase}`

so

```
HEADER_X_LEDGER_CLIENT_VERSION
HEADER_X_LEDGER_COMMIT
HEADER_X_LEDGER_NEXT
```

(maybe even without the "HEADER_") , it's not very important feedback ^^

domain/api/currency-token/src/api.ts
```diff
@@ -0,0 +1,250 @@
...
+} from "./internals";
+
+/**
+ * Thunk `extraArgument` contract for {@link cryptoAssetsApi}. The app supplies the CAL
+ * service URLs, client version and an optional logger at store configuration time, so this
+ * package owns no env/config/logging dependency.
+ */
+export interface CalApiExtra {
+  calServiceUrl: string;
+  calServiceUrlStaging: string;
```
gre-ledger: i know it's like this in the historicaly version, but i'm challenging a bit the idea to have to set both the prod and staging urls. why can't we just accept a `calServiceUrl` as input and deal with the branching on the user land? (real question, maybe there are downsides, but i feel it's a bit an anti pattern i've seen lately, same on the fact that in env we define both prod and staging, it would be better to have the url actually set instead of a boolean switching two static env, no?)

domain/api/currency-token/src/api.ts
```diff
@@ -0,0 +1,250 @@
...
+  logger?: (...args: unknown[]) => void;
+}
+
+/** Builds the {@link CalApiExtra} object the store passes as thunk `extraArgument`. */
+export function calApiExtra(
+  calServiceUrl: string,
+  calServiceUrlStaging: string,
+  ledgerClientVersion: string,
+  logger?: CalApiExtra["logger"],
+): CalApiExtra {
```
gre-ledger: personally i don't see the benefit of this function. just inline the object in the userland side of that function. because with such number of args, it starts to be easy to mess up the order and without typecheck we can't know that you inverse url and version (for instance)

domain/api/currency-token/src/api.ts
```diff
@@ -0,0 +1,250 @@
...
+ * `extraArgument` at request time, then delegates to {@link fetchBaseQuery}. Wrapped in `retry`.
+ */
+const calBaseQuery: BaseQueryFn<string | StagingFetchArgs, unknown, FetchBaseQueryError> = retry(
+  (args, api, extraOptions) => {
+    const extra = api.extra as CalApiExtra;
+    let baseUrl = extra.calServiceUrl;
+    let fetchArgs: string | FetchArgs = args;
+    if (typeof args === "object") {
+      const { [STAGING]: staging, ...rest } = args;
+      if (staging) baseUrl = extra.calServiceUrlStaging;
```
gre-ledger: patterns feels overcomplex, i'm not sure why we need to branch the staging url at the request time too 🤔 is it only because it's the only way to grab the context? then maybe the actual url to be part of the args

gre-ledger (commented):
some of these feedback may be existing issue but i see an opportunity to do the fixes now rather than later.
I especially would like to see if we can move away from the pattern to have to send both prod and stg endpoint to the domain/api but instead we should only have one base url to set (if we think at the most canonical usage of such lib)
then it means we may have to figure out the staging branching somewhere else, here i'm not even sure where it's done.

domain/api/currency-token/src/internals/constants.ts
ysitbon: Done. Renamed to match the wire names: `HEADER_X_LEDGER_CLIENT_VERSION`, `HEADER_X_LEDGER_COMMIT`, `HEADER_X_LEDGER_NEXT`.

domain/api/currency-token/src/internals/constants.ts
ysitbon: Ok, I don't have a strong opinion on that. I inlined `/v1/tokens` and `/v1/currencies`

domain/api/currency-token/src/api.ts
ysitbon: Agreed, it's an anti-pattern. Switched to a single injected `calServiceUrl`, the app resolves prod vs staging and passes the URL.

domain/api/currency-token/src/api.ts
ysitbon: Agree and disagree 😄

**Agree** on the ergonomics. I'd forgotten to make it take a typed **object** instead of positional args, so url/version can't be transposed. Done: `calApiExtra({ calServiceUrl, ledgerClientVersion })`.

**Disagree** on dropping it. The store has a *single* `extraArgument` shared by several APIs (CAL, CVS, …), and RTK leaves `extraArgument` **untyped** (`unknown`) at the thunk boundary. Each base query just does `api.extra as XApiExtra`. So an inlined literal is checked against nothing: forget a field or mistype a key, and it still compiles, then shows up as `undefined` at request time. The per-package factory is the one place that **compile-checks** each API's config is complete and correctly named, instead of maintaining the merged bag blindly:

```ts
middleware: gdm =>
  gdm({
    thunk: {
      extraArgument: {
        ...calApiExtra({
          calServiceUrl: getEnv("CAL_SERVICE_URL"),
          ledgerClientVersion: getEnv("LEDGER_CLIENT_VERSION"),
        }),
        ...cvsApiExtra({ cvsBaseUrl: getEnv("LEDGER_COUNTERVALUES_API") }),
        // …other api extras
      },
    },
  }).concat(cryptoAssetsApi.middleware, cvsApi.middleware),
```

I also moved this rationale into the function's JSDoc so it doesn't only live in this thread.  And we could go one step further: instead of the identity function, back it with a Zod schema and `parse()` the input, that would fail fast with a clear error at app init if a field is missing/empty, rather than surfacing as `undefined` at request time.

domain/api/currency-token/src/api.ts
ysitbon: `calBaseQuery` no longer branches anything. It just reads `extra.calServiceUrl` for the base URL + sets the client-version header. The staging Symbol relay is gone.

domain/api/currency-token/src/api.ts
ysitbon: Did it 😄

gre-ledger (approved):

LucasWerey (approved):

PR #19087, opened 2026-06-29
[LWDM] feat(bitcoin): block manual fee below node min relay fee

YazhuEth:
### ✅ Checklist

      - Bitcoin-family send flow: BTC enforces a 1 sat/vB floor; other coins (LTC, DOGE, DASH, DGB, ZEC…) use their node's relay fee as the floor when the explorer exposes it (no floor otherwise).
      - Manual / custom fee entry in all four send flows (old + new, desktop + mobile).
      - Verify a manual fee below the node minimum is blocked in the form (not at broadcast), and a normal fee still goes through.

### 📝 Description

**Problem**: a user could enter a custom sat/vB below what the node will relay, which only failed later at broadcast with a "min relay fee not met" error.

**Solution**: fetch the node's min relay fee from Atlas (`/network` → `relay_fee`) and use it as the minimum for manual BTC fees in the send flow. When the node doesn't return it, we fall back to **1 sat/vB** (Bitcoin Core's default `minrelaytxfee`). The floor is stored on `networkInfo.relayFeePerByte` (BTC only) by `getAccountNetworkInfo`, then enforced in the form.

#### What happens per flow

| Flow | Input | Behaviour when fee < floor |
| --- | --- | --- |
| **Desktop – old** (`FeesField`) | integer-only | `FeeTooLow` shown directly under the fee input, Continue disabled |
| **Mobile – old** (`ScreenEditCustomFees`) | integer-only | inline `FeeTooLow` message added on the input + button disabled (also caught on the Summary) |
| **Desktop – new** (MVVM `CustomFees`) | decimals allowed | descriptor `minValue` → "Minimum is N" on the input + confirm disabled |
| **Mobile – new** (MVVM `CustomFees`) | decimals allowed | same shared descriptor → "Minimum is N" on the input + confirm disabled |

#### Why two guards

1. **`minValue` on the custom-fee descriptor** (`live-common`, used by both *new* flows): the front-line, input-level guard. It's the only layer that can show "Minimum is N" *where the user types*, and it's needed because the new flows accept decimals (e.g. `0.1`, which otherwise got floored to `0` and showed a confusing "fees required").

2. **`FeeTooLow` in `getTransactionStatus`** (`coin-bitcoin`): the backstop at transaction-validation level. `getTransactionStatus` runs for **every** flow and path (old flows, RBF / edit-transaction), so it invalidates a too-low fee even where no input-level guard exists. This matters as soon as the node returns a relay fee **> 1**: in the old flows the integer input happily accepts `1`, and only this backstop would reject it when the floor is e.g. `3`.

The two are complementary: `minValue` is UX (clear message at the input, new flows), `FeeTooLow` is correctness (uniform coverage across all flows/paths). Today the node returns `1 sat/vB`, so the practical effect is a robust 1 sat/vB floor + a fallback if Atlas is unavailable; the mechanism scales automatically if the node ever reports a higher relay fee.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-23486](https://ledgerhq.atlassian.net/browse/LIVE-23486)

---

[LIVE-23486]: https://ledgerhq.atlassian.net/browse/LIVE-23486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-modules/coin-bitcoin/src/getAccountNetworkInfo.ts
+++ b/libs/coin-modules/coin-bitcoin/src/getAccountNetworkInfo.ts
@@ -66,8 +66,13 @@ export async function getAccountNetworkInfo(account: Account): Promise<NetworkIn
     })),
     defaultFeePerByte: clamped[Math.floor(clamped.length / 2)] || new BigNumber(0),
   };
+  // BTC enforces a 1 sat/vB floor (Bitcoin Core's minrelaytxfee default); other coins use the node value as-is
+  const relayFeePerByte =
+    account.currency.id === "bitcoin" ? BigNumber.max(floorSatPerVB, 1) : floorSatPerVB;
+
   return {
     family: "bitcoin",
     feeItems,
+    relayFeePerByte,
   };
 }
--- a/.changeset/btc-min-relay-fee-floor.md
+++ b/.changeset/btc-min-relay-fee-floor.md
@@ -1,0 +1,8 @@
+---
+"@ledgerhq/coin-bitcoin": minor
+"@ledgerhq/live-common": minor
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
+Use the node's min relay fee as the minimum for manual Bitcoin-family fees in the send flow (BTC falls back to 1 sat/vB). A fee below it is now rejected in the form instead of at broadcast.
--- a/apps/ledger-live-desktop/static/i18n/en/app.json
+++ b/apps/ledger-live-desktop/static/i18n/en/app.json
@@ -7330,6 +7330,9 @@
     "FeeRequired": {
       "title": "Fees are required"
     },
+    "FeeTooLow": {
+      "title": "Fees are too low to be accepted by the network. Please increase them"
+    },
     "GasLessThanEstimate": {
       "title": "This may be too low. Please increase"
     },
--- a/apps/ledger-live-mobile/src/families/bitcoin/ScreenEditCustomFees.tsx
+++ b/apps/ledger-live-mobile/src/families/bitcoin/ScreenEditCustomFees.tsx
@@ -60,12 +60,16 @@ function BitcoinEditCustomFees({ navigation, route }: Props) {
   const bridge = useAccountBridge<BitcoinTransaction>(account, parentAccount);
   const [ownSatPerByte, setOwnSatPerByte] = useState(satPerByte ? satPerByte.toString() : "");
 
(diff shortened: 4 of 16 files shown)


YazhuEth:
**E2E**
- Desktop : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/b9bb86be-02fe-4bee-a4ca-6c738c8d1bfd/
- Mobile : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/ce977d05-ba0c-481d-8d63-e5812b2d01ff/

libs/coin-modules/coin-bitcoin/src/getAccountNetworkInfo.ts
```diff
@@ -66,8 +66,14 @@ export async function getAccountNetworkInfo(account: Account): Promise<NetworkIn
     })),
     defaultFeePerByte: clamped[Math.floor(clamped.length / 2)] || new BigNumber(0),
   };
-  return {
+  const networkInfo: NetworkInfo = {
     family: "bitcoin",
     feeItems,
   };
+
```
cted-ledger: This `if (currency.id === "bitcoin")` gate is duplicated in `getTransactionStatus`, and the optional `relayFeePerByte?` is what forces the fallback to be re-derived there. Could we set it for the whole family here (default sourced per-currency) and make it required on the live `NetworkInfo` type, so consumers just read it? (Optional on `NetworkInfoRaw` is fine for back-compat.)

YazhuEth: Done 👍 relayFeePerByte is now required on NetworkInfo (kept optional on Raw for back-compat) and set for the whole family. One caveat: the 1 sat/vB clamp is BTC-specific (Core's minrelaytxfee default), so other coins use the node value as-is (0 = no floor) instead of a bogus universal 1. The only remaining bitcoin check is in getTransactionStatus, for the networkInfo == null case.

YazhuEth:
**E2E**
- Desktop: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/31225619-e358-4eca-b751-7f152949c17b/
- Mobile: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/cb64b953-7ea5-44d2-91ef-f343a80c3cd1/

francois-guerin-ledger (approved):

PR #19179, opened 2026-07-01
test(B2CQA-5582): add e2e coverage for Swap Max/Balance/% buttons

kentoforik:
## Summary
- Adds Desktop e2e coverage for the Max/Balance/% quick-fill buttons on the Swap amount screen ([B2CQA-5582](https://ledgerhq.atlassian.net/browse/B2CQA-5582)), in its own `quickAmount.swap.spec.ts`.
- Covers: funded-account enabled state + Max tooltip, per-percentage tooltip copy (single asset), percentage fill math, native-coin insufficient-balance state, and zero-balance token disabled state (new `TokenAccount.ETH_USDC_2`).
- Extracts `expectAmountCloseTo`/`parseBalanceAmount` (relative-tolerance amount assertion + balance-text parsing) into a new `tests/utils/amountUtils.ts`, and dedupes the Max/percentage enabled-check, tooltip-read, and fill-amount logic in `SwapPage` behind shared helpers (`checkPercentageButtonsEnabled`, `checkPercentageFillsBalance`, `checkTooltipText`).

## Known issue (resolved)
- The "Percentage buttons fill the correct proportional amount of the balance" test previously showed flakiness when the percentage row's CSS `:hover` reveal failed to stick (seen with Electron's out-of-process webview, not specific to any one currency pair). Fixed by retrying the hover in `hoverAmountField()` until the row is confirmed visible, and by reading tooltips via a polling assertion scoped to the actual open-tooltip DOM node instead of a fixed wait.

## Test plan
- [x] `pnpm run typecheck` / `pnpm run lint` in `e2e/desktop` (clean locally)
- [x] Run `entrypoint.swap.spec.ts` / `quickAmount.swap.spec.ts` against Speculos and confirm all new/updated tests pass
- [x] Investigate and fix the percentage-row hover flakiness noted above

[B2CQA-5582]: https://ledgerhq.atlassian.net/browse/B2CQA-5582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/desktop/tests/specs/entrypoint.swap.spec.ts
cunhabruno, line 1: This spec is getting quite big, it could be better if you moved your changes to a new spec, maybe `amountButtons.spec` or something like that?

e2e/desktop/tests/specs/entrypoint.swap.spec.ts
```diff
@@ -311,6 +317,227 @@ for (const { fromAccount, toAccount, xrayTicket } of swapMax) {
...
+      await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+
+      await performSwapUntilQuoteSelectionStep(app, new Swap(fromAccount, toAccount, ""), "");
+
+      expect(await app.swap.isMaxToggleEnabled()).toBe(false);
+      expect(await app.swap.getMaxTooltipText()).toBe(
+        "You don't have enough balance including network fees",
+      );
+
+      expect(await app.swap.isPercentageEnabled("25%")).toBe(false);
```
cunhabruno: Those assertion are repeated in different places, I would create a function for that
Something like this:
```

@step("Check percentage buttons enabled = $0")
async checkPercentageButtonsEnabled(percentageKeys: PercentageKey[], expected: Boolean) {
  await this.hoverAmountField();               // hover once
  for (const key of this.percentageKeys) {
    expect(await this.isPercentageEnabled(this.percentageButtonTestId(key))).toBe(expected);
  }
}
```

e2e/desktop/tests/specs/entrypoint.swap.spec.ts
```diff
@@ -311,6 +317,227 @@ for (const { fromAccount, toAccount, xrayTicket } of swapMax) {
...
+      await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+
+      for (const { fromAccount, toAccount } of swapMaxBalancePairs) {
+        await performSwapUntilQuoteSelectionStep(app, new Swap(fromAccount, toAccount, ""), "");
+
+        const balanceText = await app.swap.getFromAccountBalanceText();
+        const balance = parseBalanceAmount(balanceText);
+
+        for (const percent of ["25%", "50%", "75%"] as const) {
+          await test.step(`${fromAccount.currency.name} - ${percent}`, async () => {
```
cunhabruno: test.step will double nest on allure report

e2e/desktop/tests/specs/entrypoint.swap.spec.ts
```diff
@@ -311,6 +317,227 @@ for (const { fromAccount, toAccount, xrayTicket } of swapMax) {
...
+            await app.swap.clickPercentage(percent);
+            const amountToSend = Number(await app.swap.getAmountToSend());
+            const expectedAmount = (balance * parseFloat(percent)) / 100;
```
cunhabruno: Prefer to not do maths at spec level maybe create a `checkPercentageFillsBalance` step?

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -574,6 +580,81 @@ export class SwapPage extends WebViewAppPage {
...
+    return !classAttr?.includes("cursor-not-allowed");
+  }
+
+  // Hovers the trigger and reads the currently open Radix tooltip. Waits for
+  // the tooltip to settle first: reading role="tooltip" immediately after
+  // hover() can still match a previous trigger's stale, not-yet-closed tooltip.
+  private async getVisibleTooltipText(hoverTrigger: () => Promise<void>) {
+    const webview = await this.getWebView();
+    await hoverTrigger();
+    await webview.waitForTimeout(500);
```
cunhabruno: I was able to get rid of this timeout and have a more reliable selector in this [commit](https://github.com/LedgerHQ/ledger-live/commit/4dc21f2745e42e2f4bfb5f9163ace09c739bb5c7), feel free to pick experiment

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -568,6 +576,104 @@ export class SwapPage extends WebViewAppPage {
...
+  }
+
+  // Max/% are Lumen `Tag` components rendered as plain <div>s: `disabled` only
+  // drives styling (never a real disabled/aria-disabled attribute), so
+  // enabled state must be read from the `cursor-not-allowed` class.
+  private async isEnabledByCursorClass(testId: string) {
+    const webview = await this.getWebView();
+    const classAttr = await webview.getByTestId(testId).getAttribute("class");
+    return classAttr !== null && !classAttr.includes("cursor-not-allowed");
+  }
```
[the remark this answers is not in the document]
kentoforik: Keeping the current behavior (missing `class` → not enabled). This branch is unreachable in practice — Lumen's `Tag`/`BaseTag` always sets either `cursor-pointer` or `cursor-not-allowed` via `cva()`, so `classAttr` is never actually null for these elements. Between the two options, failing the assertion on an unexpected DOM state (no class at all) is safer than silently treating it as enabled and letting a click proceed on a possibly-broken element.

cunhabruno (approved):

PR #19243, opened 2026-07-03
[LWDM] feat(back-11492): add configurable getBlock internal-tx source list

jprudent:
Implement [ADR-048](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7297957892) in `coin-evm`: make `getBlock` internal-tx fetching an explicit, per-network ordered source list instead of inferring strategy from explorer config type and hiding Erigon→geth fallback inside `NodeApi.traceBlock`.

### ✅ Checklist

  - `coin-evm` only — no consumer app or coin-modules per-network config in this PR.
  - **Default behaviour is preserved** (see semantics below): explorer-first, then node traces, then `empty` — but `empty` is **not** a blanket silencer for trace runtime errors.
  - Follow-up PR will pin Somnia (and others) to node-only in coin-service (coin-modules repo).
  - bump coin-module to es-2022

### 📝 Description

**Problem**

`getBlock` fetched internal transactions using a single hard-coded strategy: when an Etherscan-like explorer is configured, call the explorer first and fall back to node trace on error; otherwise use node trace directly. The node layer also hard-coded Erigon `trace_block` → geth `debug_traceBlockByNumber` inside one opaque `traceBlock` method.

This made the data path non-deterministic (latency/errors depend on which branch ran), hard to monitor, and impossible to skip a broken or rate-limited explorer without disabling it entirely. On Somnia, A4 pays the explorer timeout on every block even though node trace works — see [BACK-11492](https://ledgerhq.atlassian.net/browse/BACK-11492).

**Solution**

1. **`getBlockInternalTxsSources` on `EvmConfig`** — optional ordered list of sources: `trace_block`, `debug_traceBlockByNumber`, `explorer`, `empty`. `getBlock` tries each in order and stops at the first success.

2. **Compile-time-safe builder** — `internalTxSources()` enforces non-empty lists, no duplicates, and `empty` only as last element. Output is a branded `InternalTxSourceList` so only the builder can produce valid lists.

3. **Split `NodeApi.traceBlock`** into `traceBlockErigon` and `traceBlockGeth` — each maps to exactly one RPC. The Erigon→geth ordering now lives in the config list, not in the node layer.

4. **Testable pipeline** — `composeInternalTxsFetcher` (ordered loop) + `makeSourceFetchers` (dependency wiring) in `internalTransactionsFetcher.ts`.

**Default source list** (behaviour-preserving):

```ts
internalTxSources()
  .addSource("explorer")
  .addSource("trace_block")
  .addSource("debug_traceBlockByNumber")
  .addSource("empty")
  .build();
```

**Default empty / error semantics** (`runInternalTxSources`)

The loop distinguishes **skip** vs **remember-and-continue** vs **propagate**:

| Outcome | When | Next step |
|--------|------|-----------|
| **Success** | Any source resolves | Return map (short-circuit) |
| **Skip** (`SourceUnavailableError`) | Explorer not configured, explorer runtime failure (best-effort), trace method undefined, or all sources unavailable without `empty` | Try next source |
| **Remember + continue** | Trace RPC call fails at runtime (method was available) | Collect error, try next source (e.g. erigon error → geth) |
| **`empty` resolves** | Reached `empty` and **no** real trace error was remembered | Return empty map |
| **Propagate** | Reached `empty` with remembered trace errors, or list exhausted with remembered errors | Rethrow single error as-is, or `AggregateError` when multiple sources failed at runtime |

This preserves the old conditional the first implementation pass missed: structurally unsupported traces → empty (via skip + trailing `empty`); transient trace failures on an available method → propagate for retry — **not** silently swallowed by `empty`.

**TypeScript note:** `throwCollectedRealErrors` uses the built-in `AggregateError` (ES2021). `coin-evm/tsconfig.json` `lib` includes `es2021` so CI `build` / `typecheck` resolve the global (previously only `es2020`, which caused `TS2304: Cannot find name 'AggregateError'`).

**Example: node-only** (planned for Somnia in a follow-up):

```ts
getBlockInternalTxsSources: internalTxSources()
  .addSource("trace_block")
  .addSource("debug_traceBlockByNumber")
  .build(),
```

### ❓ Context

- **JIRA**: [BACK-11492](https://ledgerhq.atlassian.net/browse/BACK-11492)
- **ADR**: [ADR-048 — Configurable internal transactions fetch strategy for EVM getBlock](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7297957892)
- **Docs**: `libs/coin-modules/coin-evm/docs/getBlock-internal-txs-schema.md`

---

[BACK-11492]: https://ledgerhq.atlassian.net/browse/BACK-11492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-modules/coin-evm/src/logic/internalTransactionsFetcher.ts
```diff
@@ -0,0 +1,125 @@
...
+export type InternalTxsByHash = Map<string, BlockOperation[]>;
+
+export type SourceFetcher = (height: number) => Promise<InternalTxsByHash>;
+
+async function runInternalTxSources(
+  sources: readonly InternalTxSource[],
+  fetchers: Record<InternalTxSource, SourceFetcher>,
+  height: number,
+): Promise<InternalTxsByHash> {
+  let lastRealError: unknown;
```
jprudent: instead of a single error, compiles all errors collected in a array

libs/coin-modules/coin-evm/src/internalTxSources.test.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  DEFAULT_INTERNAL_TX_SOURCES,
+  internalTxSources,
+  type InternalTxSourceList,
+} from "./internalTxSources";
+
+describe("internalTxSources", () => {
```
jprudent, line 7: missing tests where inputs are invalid, empty list, empty not at end

jnicoulaud-ledger (dismissed):

libs/coin-modules/coin-evm/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "declarationMap": true,
     "lib": [
       "es2020",
```
gre-ledger: if need to introduce "es2021" you can therefore drop the former "es2020"

jprudent: ok thanks, 

by the way what are the guide lines regarding lang level ? is there any contraint in that matter ?

gre-ledger: @jprudent there is currently a grey area caused by fact we have a bunch of `public` libraries in our monorepo, we're progressively going to make them all private OR externalize them (into another repo)  https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6975684839/Ledger+Live+monorepo+package+visibility+inventory 

public lib have this grey area that we don't control where they are used and what could be the JS environment to target.

Once these are done, I think we can assume our main targets are the apps, and both desktop and mobile target ES2022 at the moment https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/tsconfig.json & https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-mobile/tsconfig.json

usually this choice inherits from underlying stack (Electron, React Native), **and at the moment it seems to be ES2022+ so we should be safe up to ES2022 🤞**

(no clear guideline exists in our doc today as i'm aware)

libs/coin-modules/coin-evm/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "declaration": true,
     "declarationMap": true,
     "lib": [
-      "es2020",
+      "es2022",
       "dom"
     ],
```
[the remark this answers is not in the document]
jprudent: acknowledged with Gre

Moustafa-Koterba (dismissed):

libs/coin-modules/coin-evm/src/network/node/rpc.common.ts
```diff
@@ -713,36 +713,31 @@ async function traceBlockErigon(
...
+    }
+    return trace;
+  });
 }
 
-async function callTraceBlock(
+async function traceBlockGethWithValidation(
   api: JsonRpcProvider,
-  blockHeight: number | "latest",
+  _currency: CryptoCurrency,
```
francois-guerin-ledger:
```suggestion
```
No need for that (and same for `traceBlockErigonWithValidation`). Can we take the opportunity to clean a bit ?

jprudent: I tried to apply your recommandation, but the result is ugly because it breaks the code symmetry:
- all NodeApi method interface starts with a currency parameter
- the code leverages that symmetry providing a generic builder (called `make`) to wrap the `withApi` around it 

=> So cleaning that unused parameter leads to a load more annoying boilerplate code.
I think it's a good compromise to keep it unused.

=> What I could do is to remove a level of indirection, see https://github.com/LedgerHQ/ledger-live/commit/421f885e05080e4faa33af389caf7736d062b49d This looks a good refactoring to me.

francois-guerin-ledger (approved):

Sebastien-Dav1d (approved):

PR #18594, opened 2026-06-16
[LWD] test(swap): adopt e2e swap history test for SwapTransactionStatusDialog modal

kentoforik:
## Summary

- Adds `data-testid` attributes to all relevant `SwapTransactionStatusDialog` components (dialog wrapper, title, date, status rows, amounts, provider, swap ID, network fees, receive account, explorer button with `data-href`)
- Introduces a new `SwapTransactionStatusDialog` page object (`e2e/desktop/tests/page/dialog/swap.transaction.status.dialog.ts`) replacing the old `OperationDrawer` swap assertions
- Updates the swap history e2e test to use the new page object and verify: title, date, sent/received amounts, network fees, receive account name, provider, swap ID, and "View in explorer" button URL

## Test plan

- [ ] Run `DISABLE_TRANSACTION_BROADCAST=1 pnpm e2e:desktop test:playwright entrypoint.swap.spec.ts` after rebuilding the desktop app
- [ ] Confirm `User should be able to see their swap history from the swap history page` passes and validates all new assertions
- [ ] Confirm no other tests in the file are skipped (`test.only` was removed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #19257, opened 2026-07-03
[LWM] test(swap): adopt e2e swap history test for SwapTransactionStatus drawer

kentoforik:
## Summary

- Adds `testID`s to all relevant `SwapTransactionStatus` drawer components (`TransactionHeader`, `StatusRow`/`StatusSection`, `DetailRow`/`DetailsSection`, `FooterSection`, `SwapTransactionStatusView`) — mirroring the desktop adoption in #18594, with `testID` placed directly on the leaf `Text` nodes (not wrapping `Box`es) so Detox can read their content on Android
- Introduces a new `SwapTransactionStatusDrawer` page object (`e2e/mobile/page/drawer/swapTransactionStatus.drawer.ts`), registered in the mobile `Application` page index
- Rewires `swapSeeHistoryOperations.spec.ts` / `swap.other.ts` to assert against the new drawer (title, date, sent/received status rows, network fees, receive account, provider, swap id, "view in explorer" button) instead of the old, now-unreachable `OperationDetails` screen
- Removes the dead `expectSwapDrawerInfos`/`swap-status`/`swap-operation-details-*` assertions from `swap.page.ts`

## Context

A new `SwapTransactionStatus` bottom-sheet drawer (LIVE-29443) replaced the old `OperationDetails` screen for swap-history row taps, but the e2e test was never updated — it kept waiting on the old screen's `swap-status` testID, which became unreachable dead code, causing a 60s timeout.

## Test plan

- [ ] Rebuild the Android e2e app (`pnpm build:android` from `e2e/mobile`) and run `pnpm test:android -- swapSeeHistoryOperations.spec.ts`
- [ ] Confirm `Swap history operations - Solana to Ethereum` passes end-to-end
- [ ] Confirm mobile app and e2e typechecks pass

e2e execution of swapSeeHistoryOperations.spec.ts: https://github.com/LedgerHQ/ledger-live/actions/runs/28664830400

e2e execution of the whole SWAP set on Mobile: https://github.com/LedgerHQ/ledger-live/actions/runs/28665394004

### DIFF
--- a/e2e/mobile/page/drawer/swapTransactionStatus.drawer.ts
+++ b/e2e/mobile/page/drawer/swapTransactionStatus.drawer.ts
@@ -1,0 +1,80 @@
+import { Step } from "jest-allure2-reporter/api";
+import { normalizeText } from "../../helpers/commonHelpers";
+import { SwapProvider } from "@ledgerhq/live-e2e-shared/enum/Provider";
+import { retryUntilTimeout } from "../../utils/retry";
+
+export type SwapTransactionStatusDetails = {
+  date: string;
+  sentAmount: string;
+  // Received amount depends on a live-computed final amount, not just the fixture's raw value: leave unset to only assert visibility.
+  receivedAmount?: string;
+  networkFees: string;
+  receiveAccount: string;
+};
+
+export default class SwapTransactionStatusDrawer {
+  scrollViewId = "swap-transaction-status-scroll-view";
+  titleId = "swap-transaction-title";
+  dateId = "swap-transaction-date";
+  sendRowId = "swap-transaction-status-send-row";
+  receiveRowId = "swap-transaction-status-receive-row";
+  sentAmountId = "swap-transaction-status-send-amount";
+  receivedAmountId = "swap-transaction-status-receive-amount";
+  networkFeesId = "swap-transaction-details-network-fees";
+  receiveAccountId = "swap-transaction-details-receive-account";
+  providerId = "swap-transaction-details-provider";
+  swapIdId = "swap-transaction-details-swap-id";
+  viewInExplorerButtonId = "swap-transaction-view-explorer-btn";
+
+  // Amounts/details render behind a loading Skeleton until the swap status view-model resolves,
+  // so the underlying text can briefly be empty right after the drawer opens: poll instead of reading once.
+  private async expectTextEventually(id: string, matcher: (text: string) => void) {
+    await retryUntilTimeout(async () => {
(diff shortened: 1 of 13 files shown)


CremaFR (dismissed):

philipptpunkt (dismissed):

e2e/mobile/page/drawer/swapTransactionStatus.drawer.ts
```diff
@@ -0,0 +1,80 @@
...
+  swapIdId = "swap-transaction-details-swap-id";
+  viewInExplorerButtonId = "swap-transaction-view-explorer-btn";
+
+  // Amounts/details render behind a loading Skeleton until the swap status view-model resolves,
+  // so the underlying text can briefly be empty right after the drawer opens: poll instead of reading once.
+  private async expectTextEventually(id: string, matcher: (text: string) => void) {
+    await retryUntilTimeout(async () => {
+      matcher(normalizeText(await getTextOfElement(id)));
+    });
+  }
```
abdurrahman-ledger: Can't we wait for some element visibility when text are loaded instead of adding  heavy retry logic ?

Also if one of the element appears (ex: first one dateId) doesn't it mean that this other are also loaded, so no need to retry for them ?

kentoforik: On the "heavy retry" part — you're right, it can be lighter. getTextOfElement already retries internally until the element exists (up to 60s), and every value here (date, amounts, fees, receive account, provider, swap id) only mounts its testID'd Text once the value is fully resolved — a Skeleton renders in its place otherwise. So the element literally doesn't exist until its content is final, which means the extra expectTextEventually retry-the-match wrapper was redundant on top of getTextOfElement's own wait. Simplified in last commit— same wait behavior, less code. Here is e2e run, verifying the fix: https://github.com/LedgerHQ/ledger-live/actions/runs/28797867205.

On "if one field is loaded, aren't the others too" — seems to be not. date and feesAmount come from the same underlying fetch (useSwapTransactionStatusController's details), but:
- the title/status rows require an additional accounts→currency lookup that date doesn't need,
receiveAccountName comes from a completely separate selector, and
providerData is a second async fetch that only kicks off once provider is already known (so it structurally resolves last).
So the fields can genuinely become ready at different times, and each still needs its own check. The good news is that check is now "free" — no custom retry code, just a direct getTextOfElement call per field.

abdurrahman-ledger (approved):

philipptpunkt (approved):

PR #19268, opened 2026-07-03
[LWDM] refactor(common): replace findCryptoCurrencyByTicker with id-anchored countervalue lookup

mcayuelas-ledger:
### ✅ Checklist

- [x] \`npx changeset\` was attached.
  - Counter-value currency resolution in settings reducers (desktop & mobile)
  - Market number formatting (countervalueFormatter, useMarket, useMarketAssets)
  - \`isCryptoCountervalue\` derivation in \`useResolveMarketCounterCurrency\`

### 📝 Description

The global \`findCryptoCurrencyByTicker\` function scans the entire crypto registry and is marked `@deprecated` because ticker strings are ambiguous — multiple currencies can share the same ticker. This caused non-deterministic resolution of crypto countervalues (BTC, ETH) in the settings reducer and in Market number formatting on both desktop and mobile.

A new closed set \`countervalueCryptoCurrencies\` (bitcoin + ethereum, resolved by stable id) and a matching \`findCountervalueCryptoCurrencyByTicker\` helper are introduced in \`currencies/support\`. All call-sites that previously used the deprecated function for countervalue resolution now use this scoped alternative. As a bonus, \`isCryptoCountervalue\` is now derived directly from the resolved currency's \`type\` field at the selector level and passed down as a prop, removing a redundant registry scan inside \`useResolveMarketCounterCurrency\`.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-33313](https://ledgerhq.atlassian.net/browse/LIVE-33313) · [LIVE-33314](https://ledgerhq.atlassian.net/browse/LIVE-33314)

---

[LIVE-33313]: https://ledgerhq.atlassian.net/browse/LIVE-33313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/currencies/support.ts
+++ b/libs/ledger-live-common/src/currencies/support.ts
@@ -1,5 +1,9 @@
-import { getFiatCurrencyByTicker, hasFiatCurrencyTicker } from "@ledgerhq/cryptoassets";
-import { FiatCurrency } from "@ledgerhq/types-cryptoassets";
+import {
+  getCryptoCurrencyById,
+  getFiatCurrencyByTicker,
+  hasFiatCurrencyTicker,
+} from "@ledgerhq/cryptoassets";
+import { CryptoCurrency, FiatCurrency } from "@ledgerhq/types-cryptoassets";
 import { getEnv } from "@ledgerhq/live-env";
 import { log } from "@ledgerhq/logs";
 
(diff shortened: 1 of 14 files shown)


LucasWerey (approved):

libs/ledger-live-common/src/currencies/support.ts
```diff
@@ -112,3 +116,25 @@ export async function listSupportedFiats(): Promise<FiatCurrency[]> {
...
+ * {@link countervalueCryptoCurrencies}. Unlike findCryptoCurrencyByTicker, this never scans
+ * the whole registry, so it can't return an arbitrary ambiguous-ticker winner.
+ */
+export function findCountervalueCryptoCurrencyByTicker(
+  ticker: string | undefined | null,
+): CryptoCurrency | undefined {
+  if (!ticker) return undefined;
+  const upperTicker = ticker.toUpperCase();
+  return countervalueCryptoCurrencies.find(c => c.ticker === upperTicker);
+}
```
gre-ledger, line 140: **our general goal is to drop `findCryptoCurrencyByTicker` and all similar pattern, not reintroducing it in live-common** because looking up a crypto by ticker isn't safe generally speaking (collisions exists / tickers also change over time, for that reason with countervalues api we no longer identify them by ticker).

The problem i see you're facing with **settings**, is the fact user's app.json have settings.counterValue being possibly "ETH" and "BTC".
what could be acceptable is to introduce (somewhere 🤔 ) a migration phase where when we load the app.json, we remap these to "ethereum" and "bitcoin", making the Settings counterValue model strictly having either a fiat's ticker OR a currency id. then you would be able to 

```diff
findFiatCurrencyByTicker(state.counterValue) ||
findCryptoCurrencyByTicker(state.counterValue) ||
- findCryptoCurrencyByTicker(state.counterValue) ||
+ findCryptoCurrencyById(state.counterValue) ||
getFiatCurrencyByTicker("USD")
```

For the market part, i'm not sure, but i think we need to solve one topic at a time if possible to reduce more and more the usages of `findCryptoCurrencyByTicker` so we can eventually sunset.

gre-ledger (changes requested):

PR #19301, opened 2026-07-06
ci(libs): make format:check mandatory on develop, skip on release branches

gre-ledger:
> [!NOTE]
> Completes #19210 which missed test-libs-reusable.yml.

### 📝 Description

#19210 replaced continue-on-error: true on format:check steps across 7 workflows but missed test-libs-reusable.yml — the one that covers coin-modules and ledger-live-common. This left a gap where PRs could merge into develop with oxfmt violations (e.g. #19225, #19256).

Applies the same pattern: skip the step on release branches (which don't have the format:check script) and let it fail-hard when targeting develop.

### 🔗 Context

- **JIRA / GitHub issue**: LIVE-31553
- **ADR** (if any): N/A

### DIFF
--- a/.github/workflows/test-libs-reusable.yml
+++ b/.github/workflows/test-libs-reusable.yml
@@ -165,10 +165,7 @@ jobs:
         run: pnpm i
       - name: Check formatting of affected libraries
         id: format-check-libs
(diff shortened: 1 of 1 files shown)


gre-ledger:
fixed by https://github.com/LedgerHQ/ledger-live/pull/19304

PR #19275, opened 2026-07-03
[LWDM] fix(asset-detail): support token deeplinks safely

sarneijim:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.**
- [x] **Impact of the changes:**
  - Desktop and Mobile `ledgerlive://market/...` / `ledgerlive://asset/...` deeplinks
  - Wallet 4.0 Asset Detail routing for Ledger token ids
  - Mobile Asset Detail and Analytics charts with non-finite market data

### 📝 Description

Token market deeplinks could only target coin ids reliably, so paths such as `ledgerlive://market/ethereum/erc20/usd_tether__erc20_` did not carry the Ledger token id and market state consistently into the Wallet 4.0 Asset Detail views. On Mobile, malformed or non-finite market chart/allocation data could also reach SVG path generation and trigger native RedBox errors such as `Invalid number formatting character 'N'`.

This PR adds a shared Ledger asset path parser in `@ledgerhq/asset-detail` and uses it from Desktop and Mobile deeplink handlers. Token deeplinks now pass the full Ledger token id and `ledgerIds` market state to Asset Detail, while legacy routes keep coin-only fallbacks. It also filters/sanitizes non-finite market series, line-chart values, balance formatting, Graph data, and Analytics ring/distribution values so chart renderers never receive `NaN`.

```ts
parseLedgerAssetPath("ethereum/erc20/usd_tether__erc20_");
// { currencyId: "ethereum", assetId: "ethereum/erc20/usd_tether__erc20_", ledgerIds: [...] }
```

Before: token deeplinks lost token context or fell back, and non-finite chart data could generate invalid SVG paths on Mobile.

After: token deeplinks open Asset Detail with token market state, and non-finite data is ignored or rendered with safe fallback values.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-27428](https://ledgerhq.atlassian.net/browse/LIVE-27428)

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

[...]

PR #19337, opened 2026-07-06
[LWDM] fix(charts): guard mobile charts against non-finite values

sarneijim:
## Summary

Split out from #19275 (token asset-detail deeplinks) at the reviewer's request — this PR contains **only** the chart-hardening changes, which are unrelated to deeplink URL sanitization.

Guards the mobile charts and the market chart-series builder against non-finite (`NaN` / `Infinity`) values so invalid data points no longer produce broken SVG paths:

- `components/Graph` — filter non-finite points, render an empty SVG for degenerate/empty input
- `LineChart/useLineChartViewModel` — sanitize series data (`NaN`/`Infinity` → `null`)
- `Analytics/RingChart/useRingChartViewModel` — skip non-finite distribution slices
- `AssetDetail/BalanceGraph/useBalanceGraphViewModel` — guard the value formatter
- `market/utils/buildMarketChartSeries` — drop non-finite `[timestamp, value]` points
- Distribution cards / analytics screens + tests

## Test plan

- [ ] `pnpm jest` for the touched mobile chart components and `buildMarketChartSeries`
- [ ] Verify charts render safely with empty / single-point / NaN-containing datasets

### DIFF
--- a/.changeset/mobile-charts-nonfinite-guard.md
+++ b/.changeset/mobile-charts-nonfinite-guard.md
@@ -1,0 +1,6 @@
+---
+"@ledgerhq/live-common": patch
+"live-mobile": patch
+---
+
+Guard mobile charts (line, ring, balance graph, distribution) and market chart series against non-finite values so invalid data points no longer produce broken SVG paths.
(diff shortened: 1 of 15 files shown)


sarneijim:
Closing: these mobile chart guards against non-finite (NaN/Infinity) values are defensive and not required by the token deeplinks work they were originally bundled with. The main distribution path is already guarded upstream (getAssetsDistribution's `sum !== 0`), and there's no reproduced crash justifying them right now. Dropping for now; can be reopened with a concrete repro if needed.

PR #19371, opened 2026-07-07
[LWDM] feat(zcash): signing orchestration [LIVE-26924]

semeano:
### 📝 Description

Implements signOperation for Zcash shielded transfers. Replaces the stub in the Zcash chain adapter with a full Observable flow: build PCZT (UtilityProcess) → sign on device via device-signer-kit-zcash → finalize → broadcast. Adds three IPC channels (buildTransaction, finalizeTransaction, broadcastTransaction) wired through main-host → utility-entry → engine, and adaptPcztForSigner() to normalise zcash-utils output (string→bigint value fields, undefined→null) for the device signer.

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/LIVE-26924


semeano:
E2E tests: https://github.com/LedgerHQ/ledger-live/actions/runs/29009946882

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -104,59 +370,191 @@ const zcashChainAdapter: ChainAdapter = {
...
+    account: Account,
+    deviceId: string,
+    transaction: Transaction,
+    signerContext: SignerContext,
+  ): Observable<SignOperationEvent> | undefined {
+    const zcashAccount = account as ZcashAccount;
+    const tx = transaction as ZcashTransaction;
+
+    // Transparent-only flows fall through to the Bitcoin legacy PSBT path.
+    if (!isShieldedTransfer(tx)) return undefined;
```
cted-ledger: Routing must be driven by the `zcashShielded` feature flag, not by `transferType` alone. Confirmed contract: flag ON ⇒ all flows (including Public→Public / transparent t→t) go through PCZT/V5; flag OFF ⇒ legacy Bitcoin for everything. Here `isShieldedTransfer` excludes `"transparent"`, so a t→t send always falls to the legacy path even when the flag is ON — the coin module never reads the flag (`useFeature`/`DefaultFeature` grep is empty in `libs/coin-modules`). The flag exists (`types-live/src/feature.ts:321`) and is already consulted in the UI (`ZcashTransferFromSelector.tsx`). Wire it into the routing decision here.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -104,59 +370,191 @@ const zcashChainAdapter: ChainAdapter = {
...
+
+    return new Observable<SignOperationEvent>(subscriber => {
+      let cancelled = false;
```
cted-ledger, line 435: Cancellation is cooperative between steps only, and on cancel the Observable calls `subscriber.complete()` (silent) rather than `subscriber.error(new CancelledError())`. Two consequences: (a) the long-running awaits (`buildTransaction` Halo2 proving ~15–30s, device sign, finalize) are not interrupted — the flag only discards the result once the in-flight promise resolves; (b) a cancelled flow is indistinguishable from "nothing to sign" for the consumer. Please confirm the `signAndBroadcast` consumer treats complete-without-`signed` as a cancel, and reconcile with the spec's "typed CancelledError / leaves no partial state" wording.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -104,59 +370,191 @@ const zcashChainAdapter: ChainAdapter = {
...
+        const transparentInputs = await mapTransparentInputs(zcashAccount, tx);
+        if (cancelled) return subscriber.complete();
+
+        // ── Step 4: build PCZT (Halo 2 proving + parsePczt in UtilityProcess) ─
+        const buildResult = await client.buildTransaction!({
+          grpcUrl,
+          ufvk,
+          network,
+          // Firmware logs but does not validate the seed fingerprint.
+          seedFingerprint: "00".repeat(32),
```
cted-ledger, line 502: Hardcoded placeholder seed fingerprint. The comment states the firmware logs but does not validate it — please confirm against the current `app-zcash` PCZT handler (`src/handlers/pczt.rs`). If the app team ever adds validation of this field, every PCZT built here silently becomes rejected on-device with no compile-time signal. Worth a follow-up to supply the real fingerprint rather than relying on the comment.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -1,23 +1,235 @@
...
+
+  return Promise.all(
+    utxos.map(async utxo => {
+      const address = utxo.address;
+      if (address === null || address === undefined || address === "") {
+        throw new Error(`UTXO is missing an address (txid: ${utxo.hash}, vout: ${utxo.outputIndex})`);
+      }
+      const derivInfo = addrMap.get(address);
+      if (!derivInfo) {
+        throw new Error(`No derivation found for UTXO address: ${address}`);
```
cted-ledger: This path is reachable in practice when a UTXO's address is outside the fetched receive/change set (`getAccountAddresses(0)`/`(1)` are gap-limit bounded) — e.g. an out-of-gap-limit or stale-sync UTXO. It's fail-closed (good), but the raw internal message leaks an address with no guidance, and there is no test covering it (nor the sibling "UTXO is missing an address" throw). Please add a test and consider a clearer user-facing error.

libs/coin-modules/coin-bitcoin/src/broadcast.ts
```diff
@@ -38,6 +46,13 @@ export const broadcast: AccountBridge<Transaction>["broadcast"] = async ({
     }
   }
 
+  // Chain adapters may own broadcasting for transactions the Bitcoin explorer
+  // cannot submit (e.g. Zcash shielded V5 txs, broadcast via gRPC). When the
+  // adapter takes over it returns the patched operation; otherwise we fall
+  // through to the standard explorer broadcast below.
+  const custom = getChainAdapter(account.currency.id).broadcast?.(account, signedOperation);
```
cted-ledger, line 51: The double-spend guard above is deliberately placed before this adapter delegation so Zcash Public→* flows that broadcast over gRPC stay protected — but that ordering is not covered by any test (the guard and the delegation are each tested only in isolation). Add a combined test: a signed op with non-empty `inputRefs` whose referenced UTXO is already spent, routed through an adapter, asserting the guard rejects before `broadcast` is called.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/__tests__/signOperation.test.ts
```diff
@@ -0,0 +1,943 @@
...
+    const signerContext = makeSignerContext();
+
+    const obs = callSignOperation(account, tx, signerContext) as Observable<SignOperationEvent>;
+    await expect(lastValueFrom(obs)).rejects.toThrow(/Missing zcashFee/);
+  });
+
+  it("surfaces device rejection as Observable error", async () => {
+    const account = makeAccount();
+    const tx = makeTx("shielded");
+    const rejectionError = new Error("UserRefusedOnDevice");
```
cted-ledger: Tautological test. It rejects with a fabricated generic `Error` whose message is the string "UserRefusedOnDevice" and asserts `.rejects.toThrow("UserRefusedOnDevice")` (substring match). Since `@ledgerhq/live-signer-zcash` is fully mocked here, the real `DmkSignerZcash.mapError` (errorCode `6985` → `new UserRefusedOnDevice()` from `@ledgerhq/errors`) is never exercised — this would still pass if the orchestration re-wrapped or stringified the typed error. Reject with a real `UserRefusedOnDevice` instance and assert `instanceof`, and add distinct tests for network-rejection and witness/proof-failure error types.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/__tests__/signOperation.test.ts
```diff
@@ -0,0 +1,943 @@
...
+
+    await expect(adapter.broadcast!(account, makeSignedOperation())).rejects.toThrow(
+      /not supported in this environment/,
+    );
+  });
+});
+
+// ── Suite 4: Cancellation ──────────────────────────────────────────────────
+
+describe("signOperation — cancellation", () => {
```
cted-ledger, line 899: Both cancellation tests are structurally incapable of failing. They populate `events` only through the `next` handler of a `Subscription` that has already been `unsubscribe()`d — RxJS guarantees `next` never fires again after unsubscribe, regardless of the production `cancelled` guards. These tests would pass identically if every `if (cancelled) return subscriber.complete();` in `index.ts` were deleted. Rewrite them to assert the downstream step's mock (`mockFinalizeTransaction`/`mockBroadcastTransaction`) is never invoked after cancellation, and add the missing 3rd checkpoint (cancel after `finalizeTransaction` resolves, before the `signed` event).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/__tests__/signOperation.test.ts
```diff
@@ -0,0 +1,943 @@
...
+    await new Promise(r => setTimeout(r, 20));
+
+    const signedEvents = events.filter(e => e.type === "signed");
+    expect(signedEvents).toHaveLength(0);
+  });
+});
+
+// ── Suite 5: IPC contract — new channels ──────────────────────────────────
+
+describe("ZCashIPC — new IPC channels", () => {
```
cted-ledger: This suite duplicates the new build/finalize/broadcast-channel assertions already covered in `ZCashIPC.test.ts` (re-imported here via `require`). It adds no incremental coverage and doubles the maintenance cost if the IPC payload shape changes. Remove it, or replace it with a scenario unique to the orchestration layer.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/__tests__/adapter.test.ts
```diff
@@ -99,24 +113,29 @@ function makeOrchardOutputNote(
...
   describe("signOperation", () => {
-    const allTransferTypes: ZcashTransferType[] = [
-      "transparent",
-      "transparent-to-shielded",
-      "shielded-to-transparent",
-      "shielded",
-    ];
-    it.each<ZcashTransferType>(allTransferTypes)(
-      "returns undefined for %s transactions (PCZT not yet implemented)",
+    it("returns undefined for transparent transfers (Bitcoin legacy path)", () => {
```
cted-ledger: This assertion locks in the current routing (transparent ⇒ legacy) as intentional, with no feature-flag dimension. Given the confirmed contract (flag ON ⇒ transparent goes through PCZT), this test will need to change — not just be supplemented — once toggle-driven routing lands. Parametrize it by flag state: ON ⇒ Observable (PCZT), OFF ⇒ undefined (legacy).

cted-ledger (commented):
Automated critical review (code + tests + specs). **No blockers.** The one structuring item: **signing routing must be driven by the `zcashShielded` feature flag** — flag ON ⇒ all flows (including Public→Public / transparent t→t) via PCZT/V5, flag OFF ⇒ legacy for everything. Today the coin module never reads the flag, so a transparent t→t always falls to legacy even with the flag ON (see inline on `index.ts`).

cted-ledger (changes requested):

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/coin-selection.ts
```diff
@@ -41,12 +41,144 @@ export function computeLogicalActions(
...
+ *   logical_actions = max(transparent_in, transparent_out)  // = transparent_in
+ *                   + max(orchard_in, orchard_out)          // = orchard_out
+ *
+ * `computeZip317Fee` floors the result at the 2-grace-action minimum.
+ */
+export function computeShieldingFee(
+  transparentInputCount: number,
+  orchardOutputCount: number,
+): BigNumber {
+  return computeZip317Fee(transparentInputCount + orchardOutputCount);
```
cted-ledger: **Fee model underprices the two mixed-pool flows — the native builder rejects them at runtime.**

For `transparent-to-shielded` and `shielded-to-transparent`, the ZIP-317 fee computed here omits the per-pool Orchard minimum-actions floor that the native builder enforces. In the builder (`ledger-zcash-utils`, `crates/zcash-crypto/src/craft.rs`), the Orchard bundle contributes its action count **independently** and is only then summed with the transparent leg:

```
orchard_actions     = (n_spends == 0 && n_orchard_outputs == 0)
                        ? 0
                        : max(ORCHARD_MIN_ACTIONS /*2*/, max(n_spends, n_orchard_outputs))
transparent_actions = max(t_in, t_out)
logical             = transparent_actions + orchard_actions
fee                 = MARGINAL_FEE * max(GRACE_ACTIONS, logical)
```

`computeShieldingFee(tIn, oOut) = computeZip317Fee(tIn + oOut)` collapses both pools into a single sum and applies only the outer grace floor, so the Orchard side is never floored to 2. Same effect for `computeLogicalActions(..., "shielded-to-transparent")` returning `max(spendCount, 1)`.

Worked example — 1 transparent input → 1 Orchard output, no shielded change (the default case):
- here: `computeZip317Fee(1 + 1)` → **10 000** zat
- builder: `transparent(1) + orchard(max(2, max(0,1))=2)` = 3 logical actions → **15 000** zat

The fee computed here is passed straight into the native `buildTransaction` (`index.ts`, `feeZat: tx.zcashFee!.toFixed(0)`), which strictly rejects any fee that is not exactly the required ZIP-317 fee (`fee != required_fee` and `builder_fee != fee` in `craft.rs`). So essentially every `transparent-to-shielded` / `shielded-to-transparent` send whose Orchard leg has fewer than 2 real actions — i.e. the default no-shielded-change case for both flows — will throw a `Craft` error out of the native module.

Pure `shielded` and pure `transparent` are unaffected: with the transparent leg (or the Orchard leg) absent, the outer grace floor of 2 coincides with the per-pool floor.

Note that `coin-selection.test.ts` currently asserts `expect(computeShieldingFee(1, 1).toNumber()).toBe(10_000)`, i.e. it locks in the incorrect value, so nothing in-repo catches this today.

**Suggested fix — make the builder's fee function the single source of truth rather than re-implementing it here.** The root issue is that two implementations of the same ZIP-317 formula must stay bit-identical, with a hard runtime failure on any drift. `zip317_fee` in `craft.rs` is already a standalone, pure function (no proving, no gRPC — plain arithmetic over per-pool action counts), and `build_transaction` already computes `required_fee` from it and returns the resulting fee. Rather than mirroring the formula in TS, expose `zip317_fee` as a cheap synchronous binding (e.g. `computeZip317Fee(counts)` via `zcash-ffi-node`) and call it wherever the wallet needs a fee (coin-selection, `getTransactionStatus`, `estimateMaxSpendable`). The wallet still selects the inputs/outputs (so the FR-4 "fee selected by ledger-live" intent is preserved), but the arithmetic lives in exactly one place and can never drift out of sync with what the builder will accept. Being pure, it can be a synchronous binding outside the UtilityProcess, so the iterative selection loop pays no IPC round-trip cost.

cted-ledger (approved):

claudiiafg (approved):

PR #19411, opened 2026-07-08
[LWDM] fix(algorand): not opt in asa error message

dilaouid:
### 📝 Description

Update LWDM to display the error message for a tx of USDT between two account, with the recipient which has not opt-in for the ASA in the recipient step instead of the amount step

image

### 🔗 Context

- **[JIRA / GitHub issue](https://ledgerhq.atlassian.net/browse/LIVE-34110)**  
- **ADR** (if any):

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/useAddressValidation.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/useAddressValidation.ts
@@ -14,6 +14,7 @@ import type {
   MatchedAccount,
   RecentAddress,
 } from "@ledgerhq/live-common/flows/send/recipient/types";
+import type { Transaction } from "@ledgerhq/live-common/generated/types";
 import type { CryptoCurrency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
 import type { Account, AccountLike, Operation } from "@ledgerhq/types-live";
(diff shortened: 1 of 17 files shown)


apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/useAddressValidation.ts
```diff
@@ -14,6 +14,7 @@ import type {
   MatchedAccount,
   RecentAddress,
 } from "@ledgerhq/live-common/flows/send/recipient/types";
+import type { Transaction } from "@ledgerhq/live-common/generated/types";
```
francois-guerin-ledger, line 17: Do we still have generated types ?

libs/ledger-live-common/src/flows/send/recipient/hooks/useRecipientSearchState.ts
```diff
@@ -28,15 +27,14 @@ export function useRecipientSearchState({
   const isBridgeInvalidAddress =
     bridgeRecipientError instanceof InvalidAddress && !isSelfTransferError;
 
+  const hasValidatedAddress =
+    result.status === "valid" || result.status === "ens_resolved" || result.status === "sanctioned";
```
francois-guerin-ledger, line 31:
```ts
[...].includes(...)
```
maybe ?

francois-guerin-ledger (approved):

PR #19431, opened 2026-07-09
fix(web-tools): reuse DMK session to avoid WebHID re-prompting on each device interaction

gre-ledger:
### 📝 Description

The DMK transport registered in web-tools/trustchain was calling startDiscovering on every
open() call, which triggers the WebHID/WebUSB browser permission dialog on every device
interaction. This made the /trustchain page nearly unusable — every button that talks to
the device popped a permission prompt.

Previous behaviour: each open() → startDiscovering → connect → new DmkCompatTransport.
No session was reused; the disconnect callback was always a no-op.

Fix: cache the DMK session across open() calls using the same pattern as
DeviceManagementKitTransport in live-dmk-desktop:
- On open(), check the cached session liveness via getDeviceSessionState + DeviceStatus.NOT_CONNECTED
- On a valid cached session, return it directly (no new discovery, no WebHID prompt)
- Subscribe to getDeviceSessionState to invalidate the cache on physical disconnect (next/error/complete)
- Store the subscription reference and unsubscribe it on cache clear (no leak)
- Guard handlers with sessionId check so stale subscriptions from a prior session cannot
  wipe a newer session cache

### 🔗 Context

- **JIRA / GitHub issue**: N/A (web-tools dev tooling fix)
- **ADR** (if any): N/A

### DIFF
--- a/apps/web-tools/src/trustchain/live-common-setup.ts
+++ b/apps/web-tools/src/trustchain/live-common-setup.ts
@@ -1,10 +1,10 @@
 import "../live-common-setup-network";
 import { registerTransportModule } from "@ledgerhq/live-common/hw/index";
-import { DeviceManagementKitBuilder } from "@ledgerhq/device-management-kit";
+import { DeviceManagementKitBuilder, DeviceStatus } from "@ledgerhq/device-management-kit";
 import { webHidTransportFactory } from "@ledgerhq/device-transport-kit-web-hid";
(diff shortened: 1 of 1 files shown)


gre-ledger:
I fixed all Copilot concerns. It entered a feedback loop that doesn't converge. i tested manually and it works.

thesan (approved):

PR #19534, opened 2026-07-13
ci(allure): send git branch with qaa-allure upload

cunhabruno:
### 📝 Description

The qaa-allure report server now accepts an optional `branch` form field on its
upload endpoint (`POST /api/v1/reports`, multipart/form-data). Sending it makes
the dashboard surface which git branch a run came from (Branch column, filter,
and a suffix in the report header). It's optional and backward-compatible.

This PR wires that field through the existing `upload-allure-report-qaa` composite:

- Added an optional `branch` input to the composite → `QAA_ALLURE_BRANCH` env →
  `--form-string "branch=…"` in `upload.sh`. It's only sent when non-empty, so
  the upload stays backward-compatible with the current server.
- Passed `inputs.ref || github.ref_name` — the ref actually checked out and
  tested — from the desktop + mobile E2E upload steps. (These workflows are
  `schedule`/`workflow_dispatch`/`workflow_call` only, never `pull_request`, so
  `github.head_ref` would always be empty; `inputs.ref || github.ref_name` is the
  same expression already used for the run-name and concurrency group.)
- Dropped the now-redundant branch entry from `tags`, since the dedicated
  `branch` field is the canonical place for it.

Kapture 2026-07-13 at 09 31 38

### ✅ CI runs (manual E2E validating the branch upload)

[- `@Mobile E2E • Manual • iOS & Android • support/qaa-allure-branch-field • cunhabruno`
](https://github.com/LedgerHQ/ledger-live/actions/runs/29234263610)
[@Desktop E2E • triggered on support/qaa-allure-branch-field by cunhabruno](https://github.com/LedgerHQ/ledger-live/actions/runs/29234242690)
### 🔗 Context

- **JIRA / GitHub issue**: N/A — experimental qaa-allure tooling
- **ADR** (if any): N/A

### DIFF
--- a/.github/workflows/test-mobile-e2e-reusable.yml
+++ b/.github/workflows/test-mobile-e2e-reusable.yml
@@ -1115,7 +1115,8 @@ jobs:
           api_key: ${{ secrets.QAA_ALLURE_API_KEY }}
           path: ios-test-artifacts
           name: "mobile-e2e-ios-${{ github.event_name == 'schedule' && 'nightly' || (inputs.smoke_tests && 'smoke' || 'manual') }}-#${{ github.run_id }}"
-          tags: ${{ format('["mobile","ios",{0},{1}]', toJSON(github.event_name == 'schedule' && 'nightly' || (inputs.smoke_tests && 'smoke' || 'manual')), toJSON(inputs.ref || github.ref_name)) }}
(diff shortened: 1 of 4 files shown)


dgreen-ledger (approved):
LGTM — clean, minimal, backward-compatible. The branch field pattern follows existing conventions in the workflow files.

VicAlbr (approved):

PR #19541, opened 2026-07-13
[LWD] refactor(desktop): reduce console.error in datadog + main process [platform]

gre-ledger:
### 📝 Description

Replace debug `console.error` with `console.warn` or drop entirely in `datadog/` and `main/` process files. Anonymization failure in `beforeSend` now returns `false` to avoid sending un-anonymized PII.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34349](https://ledgerhq.atlassian.net/browse/LIVE-34349)

[LIVE-34349]: https://ledgerhq.atlassian.net/browse/LIVE-34349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/datadog/config.ts
+++ b/apps/ledger-live-desktop/src/datadog/config.ts
@@ -41,6 +41,8 @@ function rewriteAsarUrlsRecursive(value: unknown, seen: Set<object>): void {
  * Builds the beforeSend callback for Datadog RUM / Log.
  * Drops events when opt-in is off or error message matches ignore list;
  * applies anonymization to the payload (parity with Sentry).
+ * If anonymization throws, the event is dropped (return false) to avoid
+ * sending potentially non-anonymized data.
  */
 export function buildBeforeSend(shouldSend: ShouldSendCallback) {
   return (event: unknown, _context?: unknown): boolean => {
@@ -55,14 +57,14 @@ export function buildBeforeSend(shouldSend: ShouldSendCallback) {
 
     try {
       anonymizer.filepathRecursiveReplacer(ev);
-    } catch (e) {
-      console.error("Datadog beforeSend: anonymization failed", e);
+    } catch {
+      return false;
     }
 
     try {
       rewriteAsarUrlsRecursive(ev, new Set());
     } catch (e) {
-      console.error("Datadog beforeSend: asar url rewrite failed", e);
+      console.warn("Datadog: asar URL rewrite failed (best-effort):", e);
     }
 
     return true;
--- a/apps/ledger-live-desktop/src/main/index.ts
+++ b/apps/ledger-live-desktop/src/main/index.ts
@@ -105,7 +105,7 @@ app.on("ready", async () => {
   // Defer extension installation to not block startup
   if (__DEV__) {
     setImmediate(() => {
-      installExtensions().catch(console.error);
+      installExtensions().catch(e => console.warn("Dev extensions install failed:", e));
     });
   }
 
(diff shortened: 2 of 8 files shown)


apps/ledger-live-desktop/src/datadog/config.ts
```diff
@@ -55,15 +55,13 @@ export function buildBeforeSend(shouldSend: ShouldSendCallback) {
 
     try {
       anonymizer.filepathRecursiveReplacer(ev);
-    } catch (e) {
-      console.error("Datadog beforeSend: anonymization failed", e);
+    } catch (_e) {
```
thesan: I think what sonarcloud means is that you can just right:
```suggestion
    } catch {
```

apps/ledger-live-desktop/src/datadog/config.ts
```diff
@@ -55,15 +55,13 @@ export function buildBeforeSend(shouldSend: ShouldSendCallback) {
...
+      return false;
```
thesan, line 61: Also here the behavior of the code was changed. It use to log and continue and now it exits the function. Is this intentional?

thesan (commented):
Since there are some minor behavior changes (even though it doesn't affect the UX) I think it could be nice to add a changeset for tracking sake.

apps/ledger-live-desktop/src/datadog/config.ts
gre-ledger: yes, if anonymization fails, i think we need to not report the error. actually i'm not sure we need to try catch at all. i never seen this code to fail, but it may protect from cycles issues (like non serializable objects / cycle between refs, ...) and in that case, the report in the first place would be not sendable 🤔

thesan (dismissed):

apps/ledger-live-desktop/src/datadog/config.ts
```diff
@@ -55,15 +55,13 @@ export function buildBeforeSend(shouldSend: ShouldSendCallback) {
 
     try {
       anonymizer.filepathRecursiveReplacer(ev);
-    } catch (e) {
-      console.error("Datadog beforeSend: anonymization failed", e);
+    } catch {
+      return false;
     }
```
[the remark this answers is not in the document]
gre-ledger: This is expected to drop event if we fail to anonymize it.

apps/ledger-live-desktop/src/main/index.ts
```diff
@@ -327,12 +327,11 @@ function setUserDataPath() {
...
 async function installExtensions() {
   // https://github.com/MarshallOfSound/electron-devtools-installer#usage
-  app.whenReady().then(() => {
-    installExtension([REDUX_DEVTOOLS, REACT_DEVELOPER_TOOLS], {
-      loadExtensionOptions: {
-        allowFileAccess: true,
-      },
-    }).catch(console.error);
+  await app.whenReady();
+  await installExtension([REDUX_DEVTOOLS, REACT_DEVELOPER_TOOLS], {
```
thesan, line 331: Isn't it risky to let this throw ?

gre-ledger: it's already getting caught on the usage side , which also is only in case of `__DEV__` and in in a setImmediate

```
      installExtensions().catch(e => console.warn("Dev extensions install failed:", e));
```

thesan (approved):

PR #19559, opened 2026-07-13
test(e2e-desktop): scope modal locators to the settled container

cunhabruno:
## Context

Desktop E2E `[Tezos] Receive` (`receive.address.spec.ts`) was intermittently failing on CI with a 120s timeout on `#address-field`, non-reproducible locally.

Root cause: the Receive modal opens with a 200ms entrance animation (`opacity 0→1`, `transform: scale(1.1)→scale(1)`). The `continue()` helper clicked the Continue button directly on the page, so the click could land during the animation before the step's transition was wired. Playwright reported the click as successful, but the app never advanced from the **Account** step — Speculos was never even engaged (device stayed on "Application is ready"), so `#address-field` never rendered. Coins whose spec asserts a warning/menu before `continue()` got a natural settle-wait, which is why only Tezos flaked.

## Change

Nest the modal sub-locators under the already-defined settled `container` locator (`[data-testid=modal-container][style*="opacity: 1"][style*="transform: scale(1)"]`), so every interaction waits for the entrance animation to finish first:

- `title`, `content`, `continueButton`, `closeButton`, `maxAmountCheckbox` → scoped to `container`
- `backdrop` stays page-scoped — it's the container's ancestor, not a descendant

## Validation

[@Desktop E2E • triggered on test/e2e-modal-settled-container-locators by cunhabruno](https://github.com/LedgerHQ/ledger-live/actions/runs/29253265176)

### DIFF
--- a/e2e/desktop/tests/component/modal.component.ts
+++ b/e2e/desktop/tests/component/modal.component.ts
@@ -6,13 +6,13 @@ export class Modal extends Component {
   readonly container = this.page.locator(
     '[data-testid=modal-container][style*="opacity: 1"][style*="transform: scale(1)"]',
   );
-  readonly title = this.page.getByTestId("modal-title");
-  readonly content = this.page.getByTestId("modal-content");
-  readonly continueButton = this.page.getByRole("button", { name: "continue" });
+  readonly title = this.container.getByTestId("modal-title");
+  readonly content = this.container.getByTestId("modal-content");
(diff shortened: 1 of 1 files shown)


e2e/desktop/tests/component/modal.component.ts
cunhabruno, line 1: Some properties are not being used in this component, but I will not spent much time on refactoring as this is the old modal and should be replaced soon

martijnhjk (approved):

PR #19561, opened 2026-07-13
[LWD] fix(LIVE-34367): re-enable db middleware after hard reset so identities are persisted

gre-ledger:
> [!NOTE]
> This fixes a bug found independently during work on LIVE-34028. The regression is pre-existing and unrelated to that migration.

### 📝 Description

After a **Reset App** (hard reset) on Ledger Live Desktop, `userId`, `datadogId` and `deviceIds` are never written back to `app.json` on the next launch — `jq .data.identities` returns `null`.

**Root cause:** `hardReset()` calls `disable()` on the DB middleware to prevent stale writes during `resetAll()`. The original implementation used a 1-second auto-re-enable timeout. On the post-reload init path, `initIdentities` runs and dispatches `initFromScratch` — but still within that 1-second window — so the middleware skips the write. No further state change on `identities` occurs, so the key is never written.

**Fix:** Replace the blind timeout with an explicit `enable()` call immediately after `await resetAll()` resolves. The middleware is re-enabled as soon as the file is cleared, before `initIdentities` runs.

```ts
// reset.ts
disableDBMiddleware();
await resetAll();       // IPC queue drains; file cleared
enableDBMiddleware();   // ← added: re-enable before init
resetStore();
```

No change to the persisted data shape or any other reset behaviour.

**Regression test:** build the app, trigger Reset App, relaunch, inspect `app.json`:
```sh
cat ~/Library/Application\ Support/Ledger\ Live\ Beta/app.json | jq .data.identities
# should return a valid object, not null
```

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34367](https://ledgerhq.atlassian.net/browse/LIVE-34367)
- **ADR** (if any): N/A

[LIVE-34367]: https://ledgerhq.atlassian.net/browse/LIVE-34367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/middlewares/__tests__/db.test.ts
+++ b/apps/ledger-live-desktop/src/renderer/middlewares/__tests__/db.test.ts
@@ -1,7 +1,7 @@
 import type { Dispatch, MiddlewareAPI, UnknownAction } from "@reduxjs/toolkit";
 import { setKey } from "~/renderer/storage";
 import type { State } from "../../reducers";
-import DBMiddleware from "../db";
+import DBMiddleware, { disable, enable } from "../db";
 
 jest.mock("~/renderer/storage", () => ({
   setKey: jest.fn(),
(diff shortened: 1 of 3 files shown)


gre-ledger:
## 📦 Triggered builds

Dispatched on `ledger-live-build` for `fix/LIVE-34367-reset-identities-not-persisted` (commit 8044ceb3058):

| Platform | Run |
|---|---|
| Desktop (macOS / Linux / Windows) | [run 29253589513](https://github.com/LedgerHQ/ledger-live-build/actions/runs/29253589513) |

Artifacts will be attached to the run on completion.

PR #19645, opened 2026-07-16
[LWDM] Implement readiness attribute in Accounts + link to Tezos revealed status

amaslakov:
### 📝 Description

Adds an optional generic `readiness` field to the base `Account` type — `readiness?: { ready: boolean; reason?: string }` — a chain-agnostic projection of whether an account is fully operational. It is persisted through account serialization and populated during sync via a new optional `BridgeApi.getAccountReadiness` hook. Tezos implements the hook: an account whose public key isn't revealed on-chain is reported as `{ ready: false, reason: "unrevealed" }`. Families without the hook leave `readiness` undefined.

Needed so the swap live app can read account readiness generically via wallet-api (LIVE-34259) without chain-specific logic — first consumer is Tezos CEX swap.

```ts
if (account.readiness?.ready === false) {
  // account.readiness.reason e.g. "unrevealed"
}
```

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34257](https://ledgerhq.atlassian.net/browse/LIVE-34257?atlOrigin=eyJpIjoiNWJlYTEzNmNiYjU0NGU2MDkwYjk1MmM0MTU1OTQ2ZTMiLCJwIjoiaiJ9)
- **ADR** (if any): 

[LIVE-34257]: https://ledgerhq.atlassian.net/browse/LIVE-34257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/tezos-account-readiness.md
+++ b/.changeset/tezos-account-readiness.md
@@ -1,0 +1,7 @@
+---
+"@ledgerhq/types-live": minor
+"@ledgerhq/ledger-wallet-framework": minor
+"@ledgerhq/live-common": minor
+---
+
+Add an optional `readiness` attribute to the base `Account` type (`{ ready: boolean; reason?: string }`), a generic cross-chain projection of whether an account is fully operational. It is persisted through account serialization and populated during sync via a new optional `BridgeApi.getAccountReadiness` hook. Tezos implements the hook: an account whose public key is not revealed on-chain is reported as `{ ready: false, reason: "unrevealed" }`. Families that do not provide the hook leave `readiness` undefined.
--- a/.changeset/tezos-account-readiness.md
+++ b/.changeset/tezos-account-readiness.md
@@ -2,6 +2,7 @@
 "@ledgerhq/types-live": minor
 "@ledgerhq/ledger-wallet-framework": minor
 "@ledgerhq/live-common": minor
+"@ledgerhq/coin-tezos": patch
 ---
 
-Add an optional `readiness` attribute to the base `Account` type (`{ ready: boolean; reason?: string }`), a generic cross-chain projection of whether an account is fully operational. It is persisted through account serialization and populated during sync via a new optional `BridgeApi.getAccountReadiness` hook. Tezos implements the hook: an account whose public key is not revealed on-chain is reported as `{ ready: false, reason: "unrevealed" }`. Families that do not provide the hook leave `readiness` undefined.
+Add an optional `readiness` attribute to the base `Account` type (`{ ready: boolean; reason?: string }`), a generic cross-chain projection of whether an account is fully operational. It is persisted through account serialization and populated during sync via a new optional `BridgeApi.getAccountReadiness` hook. Tezos implements the hook: an account whose public key is not revealed on-chain is reported as `{ ready: false, reason: "unrevealed" }`. Families that do not provide the hook leave `readiness` undefined. coin-tezos `getAccountByAddress` now coalesces concurrent same-address calls into a single request, so surfacing readiness during sync adds no redundant tzkt call.
--- a/libs/coin-modules/coin-tezos/src/network/tzkt.test.ts
+++ b/libs/coin-modules/coin-tezos/src/network/tzkt.test.ts
@@ -173,6 +173,20 @@ describe("tzkt network API", () => {
         }),
       );
     });
+
+    it("coalesces concurrent calls for the same address into a single request", async () => {
+      const account = { type: "user", address: "tz1dup", revealed: true } as APIAccount;
+      mockedNetwork.mockReturnValue(networkResponse(account) as ReturnType<typeof network>);
+
+      const [first, second] = await Promise.all([
+        api.getAccountByAddress("tz1dup"),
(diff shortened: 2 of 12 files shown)


Justkant:
Hey I don't think you need to add this info to the account just to have it for the wallet-api usage, you can check the plan in the description of this PR https://github.com/LedgerHQ/ledger-live/pull/19328

LMK if you prefer to keep the current implementation that touches the account instead of simply updating the wallet-api converter

amaslakov:
> Hey I don't think you need to add this info to the account just to have it for the wallet-api usage, you can check the plan in the description of this PR #19328
> 
> LMK if you prefer to keep the current implementation that touches the account instead of simply updating the wallet-api converter

@Justkant Digging in: `revealed` has to be produced at sync either way — the converter only sees what sync already wrote, and `tezosResources.revealed` isn't populated there — so a converter-only version would still need a sync-level write underneath it. We think `readiness` is meant to be generic (all accounts, default `true`), so we'd lean toward keeping it as a generic Account field set at sync rather than a passthrough on top of that same write — but happy to switch if you'd rather it live in the converter.

Justkant:
@amaslakov I don't mind keeping it in the Account it just seemed like more work for a similar result and it's only needed for the wallet-api at the moment but let's go with the Account no problem

~~Also could you make sure you target develop~~, I'm going to finish the PR to bump wallet-api so you'll be able to test

gre-ledger (dismissed):
(platform review: ok for adding concept in Account type)

gre-ledger:
⚠️ **Migration required — `@ledgerhq/cryptoassets` has been dropped** (see #20071).

Two things to fix:
- `import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets/index"` → `import { getCryptoCurrencyById } from "@domain/entity-currency-crypto"`
- `import { setCryptoAssetsStore } from "@ledgerhq/cryptoassets/state"` → **remove entirely** — store injection is gone, this is no longer needed.

See #20071 for reference.

Justkant:
A few follow-up notes from review, nothing blocking.

Checked the wallet-api boundary: `accountToWalletAPIAccount` drops `readiness` on `develop`, but #19919 stacks on this branch and adds the passthrough for both main accounts and token accounts (inheriting the parent's value), so nothing is needed here. Worth keeping the merge order in mind, since until that lands the attribute is persisted but not reachable by any live app, and the drop is silent (`readiness?` is optional on `WalletAPIAccount`, so nothing goes red).

### Retain-on-error in the shape merge

`...(readiness !== undefined ? { readiness } : {})` combined with `{ ...a, ...shape }` in `jsHelpers` means a failed readiness lookup keeps the previously persisted value instead of clearing it. `clearAccount` preserves `readiness` as well, so it also survives a syncHash-triggered clear.

That is safe for Tezos today, but only incidentally: `readinessPromise` is created before the `Promise.all`, so `getAccountInfo` and `getBalance` share the same coalesced `getAccountByAddress` promise, and `logic/getBalance` rethrows when that fetch fails, which fails the whole sync before any shape is applied. So "readiness lookup failed but sync succeeded" is currently unreachable.

It stops being unreachable for the first family whose readiness source is independent of `getBalance`. A short comment on that spread stating the retain-on-error intent would save the next implementer the trace.

### `tezosResources.revealed` is now vestigial

Only `families/tezos/types.ts` and `mock/account.ts` still reference `tezosResources`; nothing populates or reads it. This attribute supersedes it, so it could be dropped in a follow-up.

### Minor

`reason` being an open string matches wallet-api-core's own design, and we can always add `unrevealed` to its documented `AccountReadinessReason` vocabulary later.

Justkant (dismissed):

amaslakov:
## 🧪 Triggered test workflows

Manually dispatched on `LIVE-34257-types-live-live-common-implement-readiness-attribute-in-accounts-link-to-tezos-revealed-status` (commit 55b8c74), full suite on both surfaces:

| Workflow | Scope | Run |
|---|---|---|
| [Desktop] E2E Only | full suite, Speculos nanoSP | [run 30539850670](https://github.com/LedgerHQ/ledger-live/actions/runs/30539850670) |
| [Mobile] E2E Only | full suite, iOS & Android, nanoX | [run 30539991152](https://github.com/LedgerHQ/ledger-live/actions/runs/30539991152) |

Full suite rather than a tag-filtered subset: there is no `@tezos` tag to narrow to, mobile e2e has no tezos coverage, and `bridge/generic-coin-framework/getAccountShape.ts` + `ledger-wallet-framework/src/serialization/account.ts` sit on the sync/serialization path shared by every family — so the blast radius is wider than the `tezos` coin module the bot flagged.

`Coin Tester - tezos` is not re-dispatched: it passed on `2edb0f398f2`, and `55b8c74` differs only by a code comment.

Justkant (approved):

gre-ledger (approved):

ishaba (approved):

claudiiafg (approved):

PR #19662, opened 2026-07-16
[LWDM] feat(LIVE-30496): extract Aleo quick-amount tier logic into shared hook

vtaranushenko-ext-ledger:
### 📝 Description

Extract the Aleo private-send quick amount tier selection logic (Fast/Balanced/Full record boundaries) into `@ledgerhq/coin-aleo` and a shared `useAleoQuickAmountSelector` hook in `@ledgerhq/live-common`, and refactor the desktop QuickAmountSelector to consume it instead of duplicating the logic locally.

### 🔗 Context

https://ledgerhq.atlassian.net/browse/LIVE-30496

### DIFF
--- a/libs/coin-modules/coin-aleo/src/logic/utils.ts
+++ b/libs/coin-modules/coin-aleo/src/logic/utils.ts
@@ -1,8 +1,15 @@
 import BigNumber from "bignumber.js";
 import invariant from "invariant";
 import { log } from "@ledgerhq/logs";
+import { findCryptoCurrencyById } from "@ledgerhq/cryptoassets";
 import type { CryptoCurrency, TokenCurrency } from "@ledgerhq/types-cryptoassets";
-import type { Account, Operation, OperationType, TokenAccount } from "@ledgerhq/types-live";
+import type {
+  Account,
+  AccountLike,
+  Operation,
+  OperationType,
+  TokenAccount,
+} from "@ledgerhq/types-live";
 import type {
   Operation as CoinFrameworkOperation,
   MemoNotSupported,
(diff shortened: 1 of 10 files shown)


libs/ledger-live-common/src/families/aleo/react.ts
```diff
@@ -17,11 +20,161 @@ import {
...
+  patchAccountWithViewKey,
+  sumPrivateRecords,
+} from "./utils";
+import type { AleoAccount, AleoTokenAccount, AleoUnspentRecord, SigningStrategy } from "./types";
 import { aleoPrivateSyncProgress$ } from "./privateSyncProgress";
 import { MANDATORY_SYNC_POLLING_DELAY, PROGRESS_THROTTLE_INTERVAL_MS } from "./constants";
 
+const QUICK_AMOUNT_STRATEGIES: SigningStrategy[] = ["fast", "balanced", "full"];
+
+interface QuickAmountStrategyTile {
```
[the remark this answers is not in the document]
vtaranushenko-ext-ledger: QuickAmountStrategyTile is a same-file interface, so TS inlines it into the .d.ts without needing an export; confirmed via tsc --noEmit -p src/tsconfig.json with zero errors.

mateuszpalosz-ext (approved):

libs/coin-modules/coin-aleo/src/logic/utils.ts
```diff
@@ -980,6 +991,37 @@ export function sumPrivateRecords(records: AleoUnspentRecord[]): BigNumber {
...
+    },
+  };
+}
+
+export function isAleoAccount(acc: AccountLike): acc is AleoAccount | AleoTokenAccount {
+  if (acc.type === "Account") {
+    return acc.currency.family === "aleo";
+  }
+  return findCryptoCurrencyById(acc.token.parentCurrencyId)?.family === "aleo";
+}
```
henri-ly, line 1023: I find it weird isn't it the same thing ?

vtaranushenko-ext-ledger: acc.currency (base account) is a CryptoCurrency with family right on it, but acc.token (token account) is a TokenCurrency, which has no family field, only parentCurrencyId, so we have to resolve the parent via findCryptoCurrencyById to get its family.

henri-ly (approved):

PR #19688, opened 2026-07-16
docs(wallet-cli): focus ring key management skill on secure behavior

Justkant:
## Context

Addresses review feedback from @0pendev on the agent-skills sync PR [LedgerHQ/agent-skills#16](https://github.com/LedgerHQ/agent-skills/pull/16). The `ledger-wallet-cli` skill is the source of truth; the sync workflow will update the agent-skills PR once this merges to `develop`.

Review theme: *push agents to use the CLI with the best security practices we can provide.*

## Changes to `.agents/skills/ledger-wallet-cli/SKILL.md`

- **Drop all mention of `--unsecure-no-password`** — the skill now only describes password-protected ring provisioning, so the agent never surfaces or falls back to an unprotected ring.
- **`WALLET_PASS` is provisioned by the developer/user in the environment before the command runs** — the agent never sets, injects, chooses, or handles the value; it only references what the user has provisioned. Removed the "at the prompt" framing, which didn't fit the non-interactive agent flow.
- **`ring decrypt` output is sensitive** — added guidance to pipe output straight to its destination / never display the decrypted contents, since they'd land in the agent transcript, logs, and scrollback.
- **Remove the plaintext-backup suggestion** from the rotation limitation note — only re-encrypting under the new ring is advised.

Docs-only change to a skill file; no runtime code affected.

### DIFF
--- a/.agents/skills/ledger-wallet-cli/SKILL.md
+++ b/.agents/skills/ledger-wallet-cli/SKILL.md
@@ -98,9 +98,9 @@ All `--account` flags accept a session label (e.g. `ethereum-1`). Run `account d
 
 †TTY: whether the command requires an interactive terminal for user input.
 
-‡`ring init` prompts for a password unless `--unsecure-no-password` is passed. In non-interactive contexts inject `WALLET_PASS` from the OS keychain (never a literal — see [Non-TTY password injection](#ring--ledger-key-ring-lkrp)).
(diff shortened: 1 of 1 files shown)


0pendev (approved):

oscarchaix (approved):
Good changes thank you both @Justkant @0pendev

PR #19725, opened 2026-07-17
[LWDM] feat(live-env): decouple framework from definitions via injectDefinitions()

gre-ledger:
> [!IMPORTANT]
> **Breaking change** for `@ledgerhq/live-env` direct consumers (public libs, tests): `getEnv()` now throws until `injectDefinitions()` is called. Apps are unaffected (they go through `@shared/live-env` which calls it at module load). Public `libs/` and coin modules must use `getEnv<T>(name)` with an explicit type parameter.

### 📝 Description

Decouples `@ledgerhq/live-env` from its ~200 env var definitions, extracting them into a new workspace-private `@shared/live-env` DDD layer — the same pattern as `shared/feature-flags`.

The framework (`libs/env`) can now be extracted to `ts-libs` with no app-specific knowledge. Definitions stay in the monorepo under a clear team ownership structure (10 CODEOWNERS folders).

**Key design points:**

- `@ledgerhq/live-env` becomes a type-lossy framework: `getEnv(name: string): unknown`. It throws if any API is called before `injectDefinitions()`.
- `@shared/live-env` is the typed DDD layer: `getEnv<K extends EnvName>(name: K): InferDef<K>`. It calls `injectDefinitions(allDefinitions)` at module load (safe, guarded by `sideEffects: ["./src/index.ts"]` so bundlers don't strip it).
- Singleton state is stored on `globalThis.__ledgerLiveEnvState` (same pattern as `cryptoassets/src/state.ts`) so it survives module deduplication when coin modules are lazy-loaded.
- No rxjs dependency: the `changes` observable is a simple `Set<Listener>`-backed subscriber, keeping `libs/env` free of runtime dependencies.

**Consumer migration:**

```diff
- import { getEnv, EnvName } from "@ledgerhq/live-env";
+ import { getEnv, EnvName } from "@shared/live-env";
```

Public `libs/` that cannot import `@shared/*` keep importing from `@ledgerhq/live-env` but must add a type parameter:

```diff
- getEnv("LEDGER_CLIENT_VERSION")
+ getEnv<string>("LEDGER_CLIENT_VERSION")
```

React hook moved to its own package (no longer in `live-common`):

```diff
- import useEnv from "@ledgerhq/live-common/hooks/useEnv";
+ import { useEnv } from "@features/platform-env";
```

All app packages (`lld`, `llm`, `cli`, `wallet-cli`, `web-tools`) and all coin modules migrated in this PR. NX boundary enforcement updated with a temporary `BOUNDARY_EXCEPTIONS` entry (removed when `@ledgerhq/live-env` moves to `ts-libs`).

### 🔗 Context

- **JIRA / GitHub issue**: 
- **ADR**: n/a — follows existing pattern from `shared/feature-flags`


gre-ledger:
Superseded by stacked PRs #19747 (typed layer + migration) and #19748 (framework decoupling). Splitting for incremental review.

PR #19806, opened 2026-07-20
[LWDM] Live 34179 kaspa coin tester module

vladyslavchupovskiy-ext-art:
### 🔗 Context

- **JIRA / GitHub issue**: 
- **ADR** (if any):


gre-ledger:
⚠️ **Migration required — `@ledgerhq/cryptoassets` has been dropped** (see #20071).

This PR adds `"@ledgerhq/cryptoassets": "workspace:^"` in `package.json`, which will fail at install since the package no longer exists in the monorepo.

This is a coin-tester module (public lib) — don't import from `@domain/*`. Instead, access currencies through `@ledgerhq/coin-framework`: currency data is injected at bootstrap, coin modules should not resolve it directly.

See #20071 and #20100 for reference migrations.

gre-ledger:
⚠️ Please rebase on `develop` — #20076 just merged and we suspect it may break your `setEnv` calls (see the comment above for details).

PR #19814, opened 2026-07-21
fix(e2e-desktop): stabilize Entry Point - Asset Allocation swap test

kentoforik:
## Summary
- Fixes flaky failure of `Entry Point - Asset Allocation` in `entrypoint.swap.spec.ts` seen in [this CI run](https://github.com/LedgerHQ/ledger-live/actions/runs/29803004686), where the assertion expected `BTC` in the `to-account-coin-selector` but got the placeholder `"Choose asset"`.
- Root cause: the embedded swap rail resolves its default "to" currency asynchronously and can remount the widget after `checkAssetToContains` has already started asserting, so the check can catch the placeholder before the real value lands.
- Adds `waitForToAssetSelectorReady()` (mirrors the existing `checkIfToAssetIsAlreadySelected` pattern already in `swap.page.ts`) and calls it before the assertion in `checkAssetToContains`.

## Test plan
- [ ] `pnpm format` clean on the changed file
- [ ] `tsc --noEmit` shows no new errors introduced by this change
- [ ] Re-run the Desktop E2E workflow (or just `entrypoint.swap.spec.ts`) to confirm `Entry Point - Asset Allocation` passes reliably

### DIFF
--- a/e2e/desktop/tests/page/swap.page.ts
+++ b/e2e/desktop/tests/page/swap.page.ts
@@ -464,16 +464,30 @@ export class SwapPage extends WebViewAppPage {
   }
 
   @step("Wait for the to-account coin selector to be populated")
-  async waitForToAssetSelectorReady() {
-    const webview = await this.getWebView();
-    await webview.waitForFunction(
-      selectorTestId => {
-        const el = document.querySelector(`[data-testid='${selectorTestId}']`);
-        return Boolean(el?.textContent && el.textContent.trim() !== "Choose asset");
-      },
-      this.toAccountCoinSelector,
-      { timeout: 5_000 },
-    );
+  async waitForToAssetSelectorReady(timeout = 40_000) {
+    for (let attempt = 0; attempt < 2; attempt++) {
+      const webview = await this.getWebView();
+      try {
+        await webview.waitForFunction(
+          selectorTestId => {
+            const el = document.querySelector(`[data-testid='${selectorTestId}']`);
+            return Boolean(el?.textContent && el.textContent.trim() !== "Choose asset");
+          },
+          this.toAccountCoinSelector,
+          { timeout },
+        );
+        return;
+      } catch (error) {
+        // The embedded swap widget can remount (new webview instance) once its default
+        // currency resolves; if that happens mid-poll, retry once against the fresh webview
+        // instead of failing on the now-stale reference.
+        if (attempt === 0 && webview.isClosed()) {
+          this._webviewPage = undefined;
+          continue;
+        }
+        throw error;
+      }
+    }
   }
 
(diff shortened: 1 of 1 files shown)


e2e/desktop/tests/page/swap.page.ts
```diff
@@ -368,49 +368,44 @@ export class SwapPage extends WebViewAppPage {
     await expect(webview.getByTestId(this.fromAccountCoinSelector)).toContainText(asset);
   }
 
+  private async waitForSelectorPopulated(webview: Page, testId: string, timeout: number) {
```
martijnhjk: please consider declaring this private function at the top of the file below the private properties but before the step functions.

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -463,8 +458,32 @@ export class SwapPage extends WebViewAppPage {
...
+        await this.getWebView(),
+        this.toAccountCoinSelector,
+        timeout,
+      );
+    }
+  }
+
   @step("Check currency to swap to contains $0")
   async checkAssetToContains(expected: string) {
+    if (expected !== "Choose asset") {
```
martijnhjk: you could consider case insensitive:
```suggestion
    if (expected.toLowerCase() !== "choose asset") {
```

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -368,49 +368,44 @@ export class SwapPage extends WebViewAppPage {
     await expect(webview.getByTestId(this.fromAccountCoinSelector)).toContainText(asset);
   }
 
+  private async waitForSelectorPopulated(webview: Page, testId: string, timeout: number) {
+    await webview.waitForFunction(
+      selectorTestId => {
+        const el = document.querySelector(`[data-testid='${selectorTestId}']`);
+        return Boolean(el?.textContent && el.textContent.trim() !== "Choose asset");
```
martijnhjk: you could lowercase for case insensitive match here too

martijnhjk (approved):

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -89,6 +89,17 @@ export class SwapPage extends WebViewAppPage {
     this.page.getByTestId(`swap-history-to-amount-${swapId}`);
   private chooseAssetDrawer = new ChooseAssetDrawer(this.page);
 
+  private async waitForSelectorPopulated(webview: Page, testId: string, timeout: number) {
```
martijnhjk, line 92: non blocking comment: `webview: Page` is somewhat misleading as there is no guarantee the passed in page is a webview (could be any Page)

PR #19816, opened 2026-07-21
[LWM] fix: aleo testnet support in add account + token accounts in assets view

mdomanski-ext-ledger:
### 📝 Description

This PR fixes two independent Aleo bugs in mobile:
- quick actions (buy/sell/swap/stake) resolved the wrong family for token currencies because getFamilyByCurrencyId was called with the token's own id instead of its parent currency id 
- the Aleo add-account flow deduped new accounts against existing ones using freshAddress alone, which could incorrectly filter out accounts that share an address across different currencies - the key now combines currency.id:freshAddress.

Impact of changes

- Quick actions on token accounts — verify the correct actions show up and route to the correct family flow
- Aleo add-account flow (ViewKeyApproveScreen) — verify accounts aren't wrongly excluded as "existing" when they share a freshAddress with an account of another currency, and that true duplicates are still filtered out

### 🔗 Context

- https://ledgerhq.atlassian.net/browse/LIVE-34677
- https://ledgerhq.atlassian.net/browse/LIVE-34672

### DIFF
--- a/.changeset/pink-crabs-vanish.md
+++ b/.changeset/pink-crabs-vanish.md
@@ -1,0 +1,6 @@
+---
+"live-mobile": minor
+---
+
+fix: unique account key in aleo add account flow
+fix: missing getParentCurrencyId in quick actions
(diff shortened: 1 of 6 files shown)


LucasWerey (approved):
Just curious don't we need this also in the new AssetDetailView ?

henri-ly (approved):

PR #19842, opened 2026-07-21
[LWDM] fix: cannot read property length of undefined error when selling usdt…

qperrot:
… on tron

### 📝 Description

in prepareTransaction we have an if statement which is doing `if (transaction.votes.length)` but moonpay return votes as `undefined`.
The simple change is to change that if statement into `if(transaction.votes?.length)`

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-31312](https://ledgerhq.atlassian.net/browse/LIVE-31312)
- **ADR** (if any): 

[LIVE-31312]: https://ledgerhq.atlassian.net/browse/LIVE-31312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-modules/coin-tron/src/bridge/prepareTransaction.ts
+++ b/libs/coin-modules/coin-tron/src/bridge/prepareTransaction.ts
@@ -9,7 +9,7 @@ export const prepareTransaction: AccountBridge<
   const networkInfo =
     transaction.networkInfo || (await getTronAccountNetwork(account.freshAddress));
 
-  if (transaction.votes.length) {
+  if (transaction.votes?.length) {
     transaction.votes = await Promise.all(
       transaction.votes.map(async vote => ({
         ...vote,
--- a/.changeset/clever-coats-report.md
+++ b/.changeset/clever-coats-report.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-tron": patch
+---
+
+Fix: check if votes existe before checking the length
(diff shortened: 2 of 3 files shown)


libs/coin-modules/coin-tron/src/bridge/prepareTransaction.ts
```diff
@@ -9,7 +9,7 @@ export const prepareTransaction: AccountBridge<
   const networkInfo =
     transaction.networkInfo || (await getTronAccountNetwork(account.freshAddress));
 
-  if (transaction.votes.length) {
+  if (transaction.votes?.length) {
```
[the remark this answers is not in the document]
qperrot: yes but it's undefined on moonpay

hedi-edelbloute, line 12: Shouldn't we use !== undefined && > 0 or something like that ? to be more explicit

hedi-edelbloute (dismissed):

libs/coin-modules/coin-tron/src/bridge/prepareTransaction.ts
qperrot: as you wish but this actually should never happen 
Moonpay return undefined when it should be empty 
this fix it's so they can fix it on their side

henri-ly (approved):

PR #19843, opened 2026-07-21
[LWD] Redelegate / Undelegate modal — percentage buttons (25/50/75/100) apply amount only on second tap

amaslakov:
### 📝 Description

**Previous behaviour**: Amount presets (25/50/75/100%) and "send max" live in the input's renderRight; clicking one focuses the field. InputCurrency (the shared desktop amount input) syncs its display from the value prop in a useLayoutEffect that bailed out while the field was focused and held a typed value — so a preset was ignored once the user had edited the field. LIVE-34610 fixed the no-typing case; this closes the post-edit gap.

**Fix**: Track the last typed value (lastTypedRef) and keep the in-progress edit only when the incoming value is an echo of it; any external change now wins. Desktop only.

**Regression check**: InputCurrency.test.tsx: preset applies on first tap, preset overrides a typed value, and a typing echo is still preserved.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34511](https://ledgerhq.atlassian.net/browse/LIVE-34511?atlOrigin=eyJpIjoiMDcxMWU5ZGI4MGM2NDY5MWJkMWI5NzY0YjE2MDZiZmEiLCJwIjoiaiJ9)
- **ADR** (if any): 

[LIVE-34511]: https://ledgerhq.atlassian.net/browse/LIVE-34511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/InputCurrency.test.tsx
+++ b/apps/ledger-live-desktop/src/renderer/components/InputCurrency.test.tsx
@@ -1,10 +1,36 @@
-import React from "react";
+import React, { useState } from "react";
 import { BigNumber } from "bignumber.js";
 import { render, screen } from "tests/testSetup";
 import InputCurrency from "./InputCurrency";
 
 // Ethereum's "Gwei" fee unit — magnitude 9, i.e. at most 9 decimals.
 const gweiUnit = { name: "Gwei", code: "Gwei", magnitude: 9 };
+
+// Babylon-like unit (magnitude 6) for the percentage-preset scenarios.
+const babyUnit = { name: "Baby", code: "BABY", magnitude: 6 };
+
+// Mirrors the cosmos Amount field: presets live in renderRight and drive the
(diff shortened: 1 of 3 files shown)


apps/ledger-live-desktop/src/renderer/components/InputCurrency.test.tsx
LucasWerey, line 1: Hello, I made a fix for this issue yesterday. Is it still a thing?

amaslakov: @LucasWerey hey! Try to first click into the field and enter a value, then click a preset (e.g. 25%). For me this seems to still fail on develop

LucasWerey (dismissed):

LucasWerey (approved):

PR #19704, opened 2026-07-17
[LWD] test(e2e): fix entrypoint.swap flakiness with root-scoped swap POM

cunhabruno:
### 📝 Description

Fixes intermittent CI failures in `e2e/desktop/tests/specs/entrypoint.swap.spec.ts`.

**Problem:** on the Asset Detail page (aggregatedAssets ON) the embedded swap rail was located with a **page-wide** `swap-web-app-container-embedded` selector. When an embedded swap from the previously visited screen was still mounted, the selector could match more than one element — or resolve to the wrong screen's swap — causing flaky, timing-dependent failures.

**Fix:** scope the swap rail (and the Asset Detail page's own locators) to the current page root, adopting the component-based page-object pattern from #16706.

- **PageView**: add a `page-view-<path>` root testid that mirrors the route verbatim, giving each page an exact, unambiguous anchor.
- **`AppPage.pageView()`** helper + new **`SwapContainer`** component scoped to a parent locator (following the `AssetsTable` precedent).
- **`AssetDetailPage`**: takes the asset id, resolves its page root, and scopes its in-page locators to it. Portaled Lumen `Dialog`/`Menu` content (PnL dialog, favorite menu items) intentionally stay page-wide.
- **`index.ts`**: `assetDetail` is now an `assetId` factory; updated callers in `assets.aggregation` and `delegate` specs.

### ❓ Context

- **JIRA**: [QAA-1406](https://ledgerhq.atlassian.net/browse/QAA-1406)
- Failing job: https://github.com/LedgerHQ/ledger-live/actions/runs/29472883044/job/87539485233
- Pattern proposal / precedent: #16706

### ✅ Checklist

- [ ] `npx changeset` — n/a (test-only change, no package published)
- [x] Covered by automatic tests (this *is* an e2e test fix)
- [ ] **QA focus:** the Swap entry-point suite (Asset Allocation and Market → asset entry points) with `aggregatedAssets` enabled.

> Note: the base commit is currently unsigned (CI/local env couldn't reach the GPG agent); will re-sign before merge if required.

[QAA-1406]: https://ledgerhq.atlassian.net/browse/QAA-1406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #16706, opened 2026-04-23
[LWD] test(e2e): refactor wallet 4.0 assets page objects and update specs

cunhabruno:
### ✅ Checklist

- [ ] `npx changeset` was attached.
- [ ] **Covered by automatic tests.** 
- [ ] **Impact of the changes:** 
  - ...

### 📝 Description

- Refactor wallet 4.0 assets page objects: extract reusable `AssetSection` component and introduce `CryptoAssetsPage`, `MarketCoinPage`, and a shared `CommonScreen` base (with `page-header` + back button) to reduce duplication.
- Rework `assets.addresses.spec.ts` into three focused scenarios — empty state (placeholders + add-account flow), existing BTC/ETH accounts, and capped display on portfolios with >6 assets — and replace the single `LIVE-26275` TMS tag with actual xray ID

### ❓ Context

- **JIRA or GitHub link**: 
- **ADR link (if any)**: 
- https://ledgerhq.atlassian.net/wiki/spaces/QA/pages/7091945582/Playwright+Component-based+Page+Objects+-+pattern+proposal
- [CI](https://github.com/LedgerHQ/ledger-live/actions/runs/28008158002) with asset section enabled
- [CI](https://github.com/LedgerHQ/ledger-live/actions/runs/28008176098) with asset section disabled

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #19874, opened 2026-07-21
[LWD] test(e2e): fix entrypoint.swap flakiness with root-scoped swap POM

cunhabruno:
### 📝 Description

Re-lands the `entrypoint.swap` flakiness fix (previously #19704, reverted in #19873) with the semantic conflict resolved.

**Problem:** on the Asset Detail page (aggregatedAssets ON) the embedded swap rail was located with a **page-wide** `swap-web-app-container-embedded` selector. When an embedded swap from the previously visited screen was still mounted, the selector could match more than one element — or resolve to the wrong screen's swap — causing flaky, timing-dependent failures.

**Fix:** scope the swap rail (and the Asset Detail page's own locators) to the current page root, adopting the component-based page-object pattern from #16706.

- **PageView**: add a `page-view-<path>` root testid that mirrors the route verbatim, giving each page an exact, unambiguous anchor. The testid is derived in the view model (`getPageTestId` in `utils.ts` → `pageTestId`), keeping the View presentational.
- **`AppPage.pageView()`** helper + new **`SwapContainer`** component scoped to a parent locator (following the `AssetsTable` precedent).
- **`AssetDetailPage`**: takes the asset id, resolves its page root, and scopes its in-page locators to it. Portaled Lumen `Dialog`/`Menu` content (PnL dialog, favorite menu items) intentionally stay page-wide.
- **`index.ts`**: `assetDetail` is now an `assetId` factory; updated callers in `assets.aggregation` and `delegate` specs.

### 🔁 Why the previous PR was reverted

`develop` had trimmed the `PageView` view model's public surface so `pathname` was no longer exposed, but the reverted change still referenced it in `PageView` → `TS2304: Cannot find name 'pathname'`. This version threads a ready `pageTestId` through the view model instead, so the View no longer depends on a raw `pathname` prop.

### ❓ Context

- **JIRA**: [QAA-1406](https://ledgerhq.atlassian.net/browse/QAA-1406)
- Failing job (original): https://github.com/LedgerHQ/ledger-live/actions/runs/29472883044/job/87539485233
- Pattern precedent: #16706
- Previous PR / revert: #19704 / #19873

### ✅ Checklist

- [ ] `npx changeset` — n/a (test-only + internal testid, no package published)
- [x] Covered by automatic tests (this *is* an e2e test fix)
- [x] Ran locally against Speculos: `Entry Point - Asset Allocation` and `Market page - More than one account` both pass.
- [ ] **QA focus:** Swap entry-point suite (Asset Allocation and Market → asset entry points) with `aggregatedAssets` enabled.

[QAA-1406]: https://ledgerhq.atlassian.net/browse/QAA-1406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/desktop/tests/page/assetDetail.page.ts
+++ b/e2e/desktop/tests/page/assetDetail.page.ts
@@ -1,35 +1,59 @@
-import { expect } from "@playwright/test";
+import { expect, Locator, Page } from "@playwright/test";
 import { step } from "tests/misc/reporters/step";
 import { AppPage } from "./abstractClasses";
+import { SwapContainer } from "tests/component/swap/swap-container";
 
 export class AssetDetailPage extends AppPage {
-  private readonly header = this.page.getByTestId("asset-detail-header");
-  private readonly marketPrice = this.page.getByTestId("asset-detail-market-price");
-  private readonly marketPriceFiatVariation = this.page.getByTestId(
-    "asset-detail-market-price-fiat-variation",
-  );
-  private readonly totalBalance = this.page.getByTestId("asset-detail-total-balance");
-  private readonly addressList = this.page.getByTestId("asset-detail-address-list");
-  private readonly addressRows = this.page.locator(`[data-testid^="asset-detail-address-row-"]`);
-  private readonly addressRowBalances = this.page.locator(
-    `[data-testid^="asset-detail-address-balance-"]`,
-  );
-  private readonly addAddressAction = this.page.getByTestId("asset-detail-add-address");
-  private readonly transactionsSection = this.page.getByTestId("asset-detail-transactions-section");
-  private readonly transactionRows = this.page.locator(`[data-testid^="history-operation-row-"]`);
-  private readonly pnlCards = this.page.locator(`[data-testid^="asset-detail-pnl-card-"]`);
-  private readonly interactivePnlCard = this.page.getByTestId(
-    "asset-detail-pnl-card-unrealisedReturn",
-  );
-  private readonly pnlDetailDialog = this.page.getByTestId("pnl-detail-dialog");
-  private readonly optionsTrigger = this.page.getByTestId("asset-detail-header-options-trigger");
-  private readonly addFavoriteMenuItem = this.page.getByTestId("asset-detail-add-favorite");
-  private readonly removeFavoriteMenuItem = this.page.getByTestId("asset-detail-remove-favorite");
+  private readonly assetId: string;
+  private readonly assetDetailRoot: Locator;
+  readonly swapContainer: SwapContainer;
+
+  private readonly header: Locator;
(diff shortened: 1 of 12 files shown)


cunhabruno:
## 🧪 Triggered test workflow

Manually dispatched on `test/qaa-1406-fix-entrypoint-swap-flaky` (commit a76791432b1), full desktop suite:

| Workflow | Scope | Run |
|---|---|---|
| [Desktop] E2E Only | full suite, Speculos nanoSP | [run 29864939857](https://github.com/LedgerHQ/ledger-live/actions/runs/29864939857) |

e2e/desktop/tests/page/assetDetail.page.ts
```diff
@@ -1,35 +1,59 @@
...
-  // The staking section renders either the "earn banner" (not yet staked) or the "earn deposit"
-  // card (already staked); both open the same stake flow.
-  private readonly earnEntry = this.page
-    .getByTestId("asset-detail-earn-banner")
-    .or(this.page.getByTestId("asset-detail-earn-deposit"));
+
+  constructor(page: Page, assetId: string) {
+    super(page);
+    this.assetId = assetId;
+    this.assetDetailRoot = this.pageView(`asset/${this.assetId}`);
```
martijnhjk, line 36: Have you considered using getters for these instead? It will reduce the constructor and delay the execution until the property is called, eg:
```
get assetDetailRoot() {
  return this.pageView(`asset/${this.assetId}`);
}

get header(): Locator {
  return this.assetDetailRoot.getByTestId("asset-detail-header");
}
...
```

cunhabruno: Thanks for the suggestion! 🙏 I don't think getters buy us much here though. Playwright locators are already lazy: creating one just builds a selector string, and the DOM isn't queried until an action/assertion actually runs on it. So field/constructor assignment vs. a getter makes no difference to when the DOM is hit.

I'd rather keep constructor/field assignment here to stay consistent with the current POM pattern ([#16706](https://github.com/LedgerHQ/ledger-live/issues/16706)). If we do want to move to getters for some reason, let's do it as a separate, repo-wide change rather than diverging in this one file. 👍

martijnhjk: OK I was also suggesting in terms of readability. Smaller constructor and no need to declare the property explicitly and assign a value elsewhere. PRs would be less noisy too with getters because the property and the value are declared in one place so more contextual and less diff.

martijnhjk (approved):

LucasWerey (approved):

PR #19907, opened 2026-07-22
test(desktop-e2e): borrow loan driver + E2E setup hooks (QAA-1401)

VicAlbr:
### ✅ Checklist

- [ ] **Covered by automatic tests.** _This is QA E2E tooling — a headless driver that signs real transactions on Speculos against the staging Borrow API + Ethereum mainnet. It is validated by running the flows against staging and by static checks (typecheck / oxlint / oxfmt) + dispatcher smoke tests, not by unit tests._
  - Desktop Playwright E2E **tooling only** (`e2e/desktop/scripts/borrow/`, `scripts/e2e.ts`, `tests/utils/borrowSetup.ts`); no app/runtime code touched.
  - `borrow.spec.ts` open-loan describe gains an `afterAll(resetLoanState)` cleanup.
  - Real-mainnet driver: verify `pnpm e2e borrow open|close|repay|withdraw` against staging (SEED + COINAPPS + Docker), and the dispatcher (`pnpm e2e`, `--help`, unknown command, `pnpm e2e borrow`).

### 📝 Description

E2E coverage for the Borrow repay/withdraw flows (B2CQA-6073 / B2CQA-6080) needs a deterministic on-chain loan state to run against, and the open-loan spec needs a clean, constant environment across runs. On-chain loan state can't be mocked or snapshotted, so this adds a headless driver that creates and tears down real loan positions.

**Solution:**

- **Borrow driver** (`e2e/desktop/scripts/borrow/`) that drives the staging Borrow API and signs on **Speculos** (via `@ledgerhq/live-signer-evm` + the DMK Speculos transport — not the deprecated `apps/cli`). Flows: `open` / `close` / `repay` / `withdraw`, idempotent no-ops when there is nothing to do.
- A **`pnpm e2e` subcommand dispatcher** (`scripts/e2e.ts`); the driver runs as `pnpm e2e borrow <flow>`.
- Reusable **E2E hooks** (`tests/utils/borrowSetup.ts`): `ensureLoanOpen` (beforeAll precondition for future withdraw/close specs) and `resetLoanState` (before/after reset), both no-ops when broadcast is disabled. An `afterAll(resetLoanState)` is wired into the open-loan spec so `ETH_4` returns to zero each run.
- **Poll resilience**: transient `5xx`/network errors while polling action status are retried instead of aborting a run whose on-chain transactions already succeeded.

RPC defaults to `https://ethereum-rpc.publicnode.com` (override via `--rpc` / `EVM_RPC_URL`). Runs against **staging** (keyless) but builds **real Ethereum-mainnet** transactions — see `scripts/borrow/README.md` for funding, `SEED` / `COINAPPS` / Docker requirements, and the shared-`ETH_4` serialization / collision notes.

### ❓ Context

- **JIRA or GitHub link**: [QAA-1401](https://ledgerhq.atlassian.net/browse/QAA-1401)

---

[QAA-1401]: https://ledgerhq.atlassian.net/browse/QAA-1401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/desktop/tests/specs/borrow.spec.ts
```diff
@@ -77,6 +78,11 @@ test.describe("Borrow open loan", () => {
     },
   });
 
+  test.afterAll(async () => {
```
alexstapenka-ledger, line 90: README (scripts/borrow/README.md) recommends beforeAll(resetLoanState) for the open-loan spec so ETH_4 starts from zero, but borrow.spec.ts only wires afterAll. If a prior run leaves a partial/open position, the UI spec may hit hot-start instead of a clean open. Could we add beforeAll(resetLoanState) with the same 600s timeout, matching the README table?

e2e/desktop/tests/specs/borrow.spec.ts
alexstapenka-ledger, line 24: README documents that all borrow on-chain flows must not run in parallel on shared ETH_4 and suggests test.describe.configure({ mode: "serial" }) at the file level. borrow.spec.ts doesn't enforce this yet - fine with a single open-loan test today, but withdraw/close specs (6073/6080) will race on nonce/state if we merge without serial mode. Worth adding now before those specs land?

e2e/desktop/tests/specs/borrow.spec.ts
```diff
@@ -77,6 +78,11 @@ test.describe("Borrow open loan", () => {
...
+    test.setTimeout(600_000);
```
alexstapenka-ledger, line 91: resetLoanState can take ~6+ min (multiple mainnet txs). CI project timeout is 400s. test.setTimeout(600_000) inside afterAll should extend the hook, but we haven't validated this on the enable_broadcast lane. Could you confirm in a manual CI run that afterAll completes, or set the hook timeout at describe level?

libs/live-e2e-shared/src/borrow/borrowFlow.ts
```diff
@@ -0,0 +1,268 @@
...
+    }
+    if (!hash) throw new Error("Broadcast returned no hash");
+    console.log(`    broadcast: ${hash}`);
+    await submitTransaction(step.transactionId, hash);
+  }
+  if (!dryRun) await pollAction(action.actionId);
+}
+
+async function open(
+  executor: EvmSpeculosExecutor,
```
alexstapenka-ledger, line 87: UI open-loan (6065) may require an ERC20 approval step before deposit; the driver's open() only posts supply + borrow. If staging bundles approval into supply.steps[] we're fine - otherwise ensureLoanOpen / CLI open could fail on a fresh ETH_4 while the Playwright spec passes. Worth validating `pnpm e2e borrow open --dry-run` on a fully reset ETH_4?

VicAlbr: yep I test it several time should be ok

e2e/desktop/tests/specs/borrow.spec.ts
VicAlbr: Will run a CI when the PR is ready for review yep

libs/live-e2e-shared/src/borrow/types.ts
```diff
@@ -0,0 +1,33 @@
+export type BorrowAction = "supply" | "borrow" | "repay" | "withdraw";
```
martijnhjk, line 1: would it make sense to declare this closer to its use inside `e2e/desktop/scripts/borrow/borrowApi.ts`  for relevance?

libs/live-e2e-shared/src/borrow/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface PartnerActionStep {
+  transactionId: string;
+  signablePayload: string;
+  actionType: string;
+}
+
+export interface PartnerActionResponse {
```
martijnhjk, line 9: same for this, should it be declared where it is used and relevant?

martijnhjk:
Please consider renaming the file and command to something more identifiable and specific, eg `e2e-cli` or `cli-helper` (since it is already e2e/desktop scoped it might not need "e2e")? And is `scripts` the right directory or would it be clearer to have a separate `e2e/desktop/cli` directory? Finally, would it also be valuable for mobile? In which case a separate e2e cli package would be useful.

alexstapenka-ledger (dismissed):

libs/live-e2e-shared/src/borrow/types.ts
VicAlbr: Kept in `types.ts` on purpose — this file is the borrow module's shared type surface. Two of these (`OpenLoan`, `EvmSignablePayload`) are used across `borrowFlow.ts` / `evmSpeculos.ts`, so a shared types file is warranted, and I grouped the partner-API/wire DTOs together there so all the API shapes live in one discoverable place rather than some inline in `borrowApi.ts` and some in a file. Happy to inline the `borrowApi`-only ones (`BorrowAction`, `PartnerActionStep`, `PartnerActionResponse`) if you'd prefer — just leaned toward grouping for consistency.

libs/live-e2e-shared/src/borrow/types.ts
VicAlbr: Same rationale as above — grouped with the other partner-API DTOs in `types.ts` for a single discoverable type surface (some of them, like `OpenLoan` / `EvmSignablePayload`, are shared across the borrow files, so the file earns its place).

libs/live-e2e-shared/src/borrow/README.md
```diff
@@ -0,0 +1,150 @@
...
+  `--market-id` overrides only for a deliberate exception). Skips if a position already exists
+  (`--force` to override).
+- `close` — `repay` then `withdraw`. **No-op if no loan is open.** By default closes the
+  first open loan; pass `--all` to close every open loan, or `--market-id` to target one.
+- `repay` / `withdraw` — the individual halves of `close` (also honor `--all` / `--market-id`).
+
+## Usage
+
+```bash
+# from e2e/desktop
```
martijnhjk: would it be possible to encourage running from root instead to make this e2e mobile and desktop user friendly?

martijnhjk (approved):
🙌

PR #19785, opened 2026-07-20
test(e2e-mobile): verify swap form shown after closing swap success screen

kentoforik:
## Summary
- Adds `SwapPage.waitForSuccessAndClose()` (`e2e/mobile/page/trade/swap.page.ts`) which waits for the swap success screen and closes it via the header X (`NavigationHeaderCloseButton`, reusing `app.common.closePage()`), instead of tapping "proceed" (which navigates to Swap History).
- `runSwapTest` (`e2e/mobile/specs/swap/swap.ts`) now uses this close path and asserts `app.swap.expectSwapPage()` afterward, verifying the initial swap input form is shown again — matching real app behavior where closing (vs. continuing) returns to the swap form.

## Test plan
- [ ] Run a Detox swap spec that uses `runSwapTest` (e.g. `swapETH_SOL.spec.ts`) against Speculos and confirm the close tap + `expectSwapPage()` pass.
- [ ] Confirm `otherTestCases/swap.other.ts` (still using the old `waitForSuccessAndContinue()`) is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #19926, opened 2026-07-23
test(desktop): verify swap form shown after closing swap success drawer

kentoforik:
## Summary
- After `verifyExchangeCompletedTextContent` in `send.swap.spec.ts`, the test now closes the swap success drawer via its header X (`app.swapDrawer.closeDrawer()`, clicking `data-testid="drawer-close-button"` — the drawer is rendered by `LiveAppDrawer` → `SideDrawer`, titled "Confirm your Exchange") instead of leaving it open.
- Then verifies the initial swap input form is shown again via `app.swap.goAndWaitForSwapToBeReady(async () => {})`, reusing the existing helper as a pure "swap page is ready" check (no new page-object method needed).
- We assert the initial swap form rather than the Swap History tab: this suite forces `DISABLE_TRANSACTION_BROADCAST=1` (`setupEnv(true)`), which makes the swap live-app report the swap as cancelled internally and skip the History redirect it would otherwise trigger on a real broadcast — so History would never be reachable here regardless of timing.
- Mirrors the equivalent mobile change in #19785.

## Test plan
- [ ] Run a case from `send.swap.spec.ts` (e.g. `-g "Swap ETH to BTC"`) against Speculos/CI and confirm the drawer close + swap-page-ready check pass.

### DIFF
--- a/e2e/desktop/tests/specs/send.swap.spec.ts
+++ b/e2e/desktop/tests/specs/send.swap.spec.ts
@@ -245,6 +245,8 @@ for (const { fromAccount, toAccount, xrayTicket, tag, amount } of swaps) {
           await app.swapDrawer.verifyExchangeCompletedTextContent(
             swap.accountToCredit.currency.name,
           );
+          await app.swapDrawer.closeDrawer();
+          await app.swap.goAndWaitForSwapToBeReady(async () => {});
         }
       },
     );


kentoforik:
Consolidated into #19928 (single PR covering both desktop and mobile).

PR #19933, opened 2026-07-23
docs(git-conventions): update guidelines and commit lint (LIVE-27608)

LL782:
### 📝 Description

1.. Refine config for commitlint
2. Update Git Convention docs

The reference to `labeler.yml` has been removed from [docs/contributing/git-conventions.md](https://github.com/LedgerHQ/ledger-live/blob/develop/docs/contributing/git-conventions.md) 

### 🔗 Context

- JIRA: https://ledgerhq.atlassian.net/browse/LIVE-27608 
- ADR: https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7266992139

#### What is scope for?

Since the latest discussion with @thesan we agreed scope should not be tied to the paths of files changed and instead be chosen by a developer (or agent).

Consider the value of these two options:

`refactor(architecture): move counter-values into domains directory` (chosen by developer) vs
`refactor(desktop,mobile,shared): move counter-values into domains directory` (based on file changes)

The former makes the reason for the change clearer from the scope alone, and lets us filter PRs related to architecture.

The latter blurs architectural changes with work done directly on apps.

#### Scope required but open

- Failing to provide a scope will now `error` in commitlint
- We have chosen not to introduce an enum as the effort of maintaining it is not worth the benefits

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,10 +22,11 @@ The canonical rules live in [Git conventions](docs/contributing/git-conventions.
 ## Updating your branch with develop
 
 **Always prefer rebasing** unless your branch contains merge commits from
-sub-features:
+sub-features.
 
-- Small, self-contained branches -> rebase: `git rebase origin/develop`
-- Branches with cross-branch merges -> merge: `git merge origin/develop`
+- Small, self-contained branches -> rebase on `develop`.
+- Branches with cross-branch merges -> merge `develop` into them to stay up to
+  date.
 
 ## The PR Lifecycle
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,10 +19,10 @@ titles and merge commits.
 
 The canonical rules live in [Git conventions](docs/contributing/git-conventions.md).
 
-### Updating your branch with develop
+## Updating your branch with develop
 
 **Always prefer rebasing** unless your branch contains merge commits from
-sub-features
+sub-features:
 
 - Small, self-contained branches -> rebase: `git rebase origin/develop`
 - Branches with cross-branch merges -> merge: `git merge origin/develop`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,14 +19,13 @@ titles and merge commits.
 
 The canonical rules live in [Git conventions](docs/contributing/git-conventions.md).
 
-### Rebase & merge strategy
+### Updating your branch with develop
 
 **Always prefer rebasing** unless your branch contains merge commits from
-sub-features.
+sub-features
 
-- Small, self-contained branches -> rebase on `develop`.
-- Branches with cross-branch merges -> merge `develop` into them to stay up to
-  date.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -107,9 +83,11 @@ Tools and resources for building on Ledger are at the [Ledger Developer Portal](
 #### Request Copilot while still in Draft
 
 > [!TIP]
+>
```
LL782, line 87: **explanation: why I added an extra line**

My format on save was continually wrapping the image onto the same line as `[!TIP]` and breaking the layout.

docs/contributing/git-conventions.md
```diff
@@ -1,120 +1,109 @@
...
+branch: feat/ui-LIVE-1234-add-dark-mode-toggle
+commit: feat(ui): add dark mode toggle
+PR title: feat(ui): add dark mode toggle (LIVE-1234)
+```
 
-Use Conventional Commit types for commit messages, pull request titles and branch
-prefixes: `build`, `chore`, `ci`, `docs`, `feat`, `fix`, `perf`, `refactor`,
-`revert`, `style` and `test`.
+We will enforce this structure for commit messages and PR titles (which will be used for merge commits).
+This makes it easier to extract reliable data from Git history.
```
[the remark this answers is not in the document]
LL782: I'm happy to leave it as it is. We're in the process of introducing DangerJS to enforce it.

CONTRIBUTING.md
```diff
@@ -15,41 +15,17 @@ Thanks for contributing! These guidelines apply to internal and external contrib
...
 
 **Always prefer rebasing** unless your branch contains merge commits from
-sub-features.
+sub-features:
 
-- Small, self-contained branches -> rebase on `develop`.
-- Branches with cross-branch merges -> merge `develop` into them to stay up to
-  date.
+- Small, self-contained branches -> rebase: `git rebase origin/develop`
+- Branches with cross-branch merges -> merge: `git merge origin/develop`
```
gre-ledger: dangerous change without fetching first, it wouldn't "stay up to date" anymore.


`git pull --rebase origin develop` would be better, it's also fetching origin before rebasing.

alternatively the equivalent in individual commands is:

```
git fetch origin develop
git rebase origin develop
```

the merge strategy can be also --rebase or not --rebase.

generally speaking, i'm not sure why we need to update this doc in context of this PR 🤔

LL782: > generally speaking, i'm not sure why we need to update this doc in context of this PR 🤔

Good point @gre-ledger! I've reverted the changes.

The only remaining change is this:

```diff
- ### Rebase & merge strategy
+ ## Updating your branch with develop
```

My reasons for that change:

1. **`h3`->`h2`**: If it says an h3 under "## Branch, Commit & PR Conventions" it's hard to explain why those details don't move down into `git-conventions.md` with everything else. 
2. **Rewording the title:** I think this is clearer and less open to misunderstanding. When I first read "merge strategy" I imagined it was about merging PRs.

gre-ledger (approved):

PR #20014, opened 2026-07-24
ci(e2e): reduce nightly iOS simulators to 2

cunhabruno:
## Context

Nightly iOS E2E runs the full 12-shard suite with **3 simulators per shard**, which drove one runner from 8.7 GB → 2.5 GB free RAM and stalled the Detox WebSocket bridge (transient `beforeAll` failure in `walletAssets.spec.ts`, passed on retry). Reducing concurrent simulators relieves the memory pressure.

## Change

Cap nightly iOS to **2 simulators / 2 Detox workers**; everything else stays at 3.

- `test-mobile-e2e-reusable.yml`: job-level `E2E_WORKER_COUNT` on the iOS job = `2` for `schedule` (nightly), `3` otherwise. Feeds the simulator pre-boot step (`worker_count`) and is inherited by the Detox run step's process.
- `e2e/mobile/jest.config.js`: `maxWorkers` honors `E2E_WORKER_COUNT` when set, falling back to the previous `CI ? 3 : 1`.

Both the pre-boot count and the Detox worker count are reduced together — pre-booting 3 simulators would consume the memory even if only 2 were used.

## Scope

- **Nightly iOS** → 2 simulators + 2 workers
- **PR / manual iOS** → unchanged (3)
- **Android** (separate job, doesn't set the env) → unchanged (3)

## Not addressed

The same run surfaced a deterministic product bug (buy webview button stays `aria-disabled="true"`, modular drawer never opens in `navigateToBuyFromAssetPage_ETH.spec.ts`) — unrelated to memory, out of scope here.

### DIFF
--- a/.github/workflows/test-mobile-e2e-reusable.yml
+++ b/.github/workflows/test-mobile-e2e-reusable.yml
@@ -523,6 +523,7 @@ jobs:
       LANGUAGE: en_US.UTF-8
       LC_ALL: en_US.UTF-8
       SPECULOS_DEVICE: ${{ needs.determine-builds.outputs.ios_device }}
+      E2E_WORKER_COUNT: ${{ github.event_name == 'schedule' && '2' || '3' }}
     outputs:
       status: ${{ steps.run-ios.outcome }}
       artifact: ${{ steps.test-artifacts.outputs.artifact-id }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test-mobile-e2e-reusable.yml
```diff
@@ -523,6 +523,7 @@ jobs:
       LANGUAGE: en_US.UTF-8
       LC_ALL: en_US.UTF-8
       SPECULOS_DEVICE: ${{ needs.determine-builds.outputs.ios_device }}
+      E2E_WORKER_COUNT: ${{ github.event_name == 'schedule' && '2' || '3' }}
```
martijnhjk, line 526: so on demand triggered runs stay at 3?

martijnhjk (approved):

.github/workflows/test-mobile-e2e-reusable.yml
martijnhjk: sorry, you already explain this in the description 👍

Sebastien-Dav1d (approved):

PR #20018, opened 2026-07-26
[LWDM] feat(coin-vechain): add CoinModuleApi (Alpaca) for VET and VTHO (LIVE-34273)

pawell24:
### 📝 Description

Migrates the VeChain family to the **CoinModuleApi (Alpaca)** generic coin-module interface, adding it **alongside** the existing legacy account bridge. The new path is fully backward-compatible and **not enabled in the app** (no `genericCoinFrameworkFamilies` flag flip), so this PR is inert for end users. Modeled on the Kaspa migration (LIVE-33340), adapted for VeChain's account model and its two-token design: VET (native) + VTHO (VIP-180 token, which is also the gas currency).

`createApi` exposes the epic's method surface (LIVE-34264):

- Implemented: `getBalance` (VET + VTHO), `listOperations`, `lastBlock` / `getBlock` / `getBlockInfo`, `craftTransaction`, `estimateFees`, `combine`, `broadcast`, `validateIntent`, `validateAddress`.
- Not applicable on VeChain (throw): `getNextSequence` (nonce is random, not sequential), `craftRawTransaction`, `call`, and staking (`getStakes` / `getRewards` / `getValidators`).

VeChain-specific handling worth reviewer attention:

- **Gas is paid in VTHO, never in native VET.** A native (VET) `OUT` operation carries no fee, so the displayed VET debit equals the exact sent amount (the generic framework otherwise adds `fees` to `value` for native ops); the VTHO operation keeps the gas.
- **VTHO is both a sendable token and the gas currency**, so a max VTHO send reserves the gas fee out of the balance.
- **`estimateFees` tolerates the incomplete form states** the generic framework probes during `prepareTransaction` (missing/invalid recipient, zero amount) instead of throwing.
- **VIP-191 fee delegation**: `getBlock` reports the actual gas payer (`gasPayer`) rather than the origin.
- **Fee parameters are coerced to numbers** in `craftTransaction` before building the body — the generic coin-framework hands them back as `bigint`, which `JSON.stringify` cannot serialize. Surfaced by the coin-tester (LIVE-34282); covered by a unit test.

Type imports use `@ledgerhq/ledger-wallet-framework/types` (the renamed `coin-framework`), not the deprecated `@ledgerhq/types-cryptoassets`. Rebased on latest `develop`.

**Tests:** unit + msw (stub) + integration (real network). 429/429 unit/msw green, ~95% line coverage on the module. The coin-tester E2E harness (thor-solo, legacy + generic-adapter strategies) ships separately in LIVE-34282.

**Staking:** intentionally not implemented. It is not in the epic's method list (LIVE-34264), VeChain has never exposed staking in Ledger Live, and the reference Kaspa migration shipped without it. The task AC's mention of "staking" is boilerplate carried over from the cosmos template — to be confirmed with Product.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34273](https://ledgerhq.atlassian.net/browse/LIVE-34273)
- **ADR** (if any): —

[LIVE-34273]: https://ledgerhq.atlassian.net/browse/LIVE-34273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-modules/coin-vechain/src/logic/validateIntent.ts
```diff
@@ -0,0 +1,96 @@
...
+function isTokenAsset(intent: TransactionIntent): boolean {
+  return intent.asset.type !== "native";
+}
+
+function spendable(balances: Balance[], isToken: boolean): bigint {
+  const asset = balances.find(b =>
+    isToken ? b.asset.type !== "native" : b.asset.type === "native",
+  );
+  return (asset?.value ?? 0n) - (asset?.locked ?? 0n);
+}
```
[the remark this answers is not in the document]
pawell24: spendable can't go negative here: getBalance returns only { value, asset } and never sets locked (nothing in coin-vechain does, VeChain has no locking/staking), so asset.locked is always 0n and spendable = value (≥ 0). The negative-availableVtho path, and the spurious NotEnoughVTHO, is therefore unreachable. No clamp needed

libs/coin-modules/coin-vechain/src/logic/transaction/craftTransaction.ts
```diff
@@ -0,0 +1,121 @@
...
+async function resolveGasFees(
+  clauses: VechainSDKTransactionClause[],
+  sender: string,
+  customFees?: FeeEstimation,
+): Promise<GasFeeParams> {
+  const custom = customFees?.parameters as Partial<GasFeeParams> | undefined;
+  if (
+    custom?.gas !== undefined &&
+    custom.maxFeePerGas !== undefined &&
+    custom.maxPriorityFeePerGas !== undefined
```
cted-ledger, line 51: `resolveGasFees` only applies `customFees.parameters` when `gas`, `maxFeePerGas`, and `maxPriorityFeePerGas` are *all* defined — a partial override (e.g. just `gas`, which `estimateFees.ts` forwards on re-estimation) is silently ignored and falls back to a full re-estimate. This has been flagged by Copilot on the last few revisions.

`coin-multiversx`'s `craftTransaction.ts` (around line 135) applies a partial `gasLimit` override independently of the other fee fields — worth merging fields individually here instead of gating on all-or-nothing.

libs/coin-modules/coin-vechain/src/logic/transaction/craftTransaction.ts
```diff
@@ -0,0 +1,121 @@
...
+  };
+}
+
+// Craft an unsigned single-clause Thor tx (VET transfer, or VIP-180 transfer for VTHO). The returned
+// transaction is the JSON VechainSDKTransactionBody that combine() parses back — keep them in sync.
+export async function craftTransaction(
+  intent: TransactionIntent,
+  customFees?: FeeEstimation,
+): Promise<CraftedTransaction> {
+  if (!intent.recipient) {
```
cted-ledger, line 81: `intent.sender` is never validated before being used in I/O (`getAccount` at line 31, `estimateGas`/`buildTransactionBody` via `resolveGasFees` at line 85), unlike `recipient` here. A malformed sender wastes a network round trip and surfaces as an opaque error instead of a clear one.

`coin-multiversx`'s `craftTransaction.ts` validates `intent.sender` up front before any network call — worth mirroring that here.

libs/coin-modules/coin-vechain/src/logic/transaction/broadcast.ts
```diff
@@ -0,0 +1,10 @@
+import type { BroadcastConfig } from "@ledgerhq/coin-module-framework/api/types";
+import { submit } from "../../network";
+import type { VechainSDKTransaction } from "../../types";
+
+// Submit a signed tx (hex from combine) via Thor POST /transactions.
+export async function broadcast(tx: string, _broadcastConfig?: BroadcastConfig): Promise<string> {
```
cted-ledger, line 6: This function has no MSW test and no `.integ.test.ts` — only a fully-mocked unit test. `coin-kaspa` (this PR's stated reference model) has both `broadcast.integ.test.ts` and `broadcast.msw.test.ts`.

Since this is the function that actually submits a signed transaction, and broadcast is called out in the migration guidance as needing to never swallow an HTTP-200-with-error-body response, an MSW test covering that case seems worth adding.

libs/coin-modules/coin-vechain/src/api/index.integ.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      expect(Array.isArray(page.items)).toBe(true);
+      for (const op of page.items) {
+        expect(["IN", "OUT"]).toContain(op.type);
+        expect(typeof op.value).toBe("bigint");
+      }
+    });
+  });
+
+  it("craftTransaction fails clearly for an intent with no recipient", async () => {
```
cted-ledger, line 90: This is the only `craftTransaction` scenario in the whole integ suite, and it only covers the failure path (missing recipient). There's no successful craft → `combine` round trip anywhere in the PR (real network or MSW) exercising the full signed-transaction shape.

libs/coin-modules/coin-vechain/src/api/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  FeeEstimation,
+  Page,
+  Reward,
+  Stake,
+  TransactionIntent,
+  TransactionValidation,
+  Validator,
+} from "@ledgerhq/coin-module-framework/api/index";
+import { setCoinConfig, type VechainCoinConfig } from "../config";
+import { parseAddress } from "../common-logic";
```
cted-ledger: `validateAddress` is wired directly to `common-logic/parseAddress` from `api/`, bypassing `logic/` — there's no `logic/validateAddress.ts`, and `logic/index.ts` doesn't re-export one.

`coin-multiversx` wraps this in `logic/validateAddress.ts` and imports it into `api/index.ts` the same way as every other method — might be worth following that pattern for consistency with the api/ → logic/ dependency layering used elsewhere in this file.

cted-ledger (commented):
Reviewed the Alpaca migration for coin-vechain. Overall the structure follows the coin-kaspa reference well (logic/ subdirectories, jest unit/msw project split, coverage exclusions). Left inline comments on a few points: a fee-override bug already flagged by Copilot on prior revisions, some test-coverage gaps versus the coin-kaspa reference model, and two dependency/validation nits versus the coin-multiversx precedent. Nothing here looks blocking on its own, but the fee-override and broadcast/craft coverage points are worth addressing before this goes further (e.g. before the coin-tester PR builds on top of it).

gre-ledger:
ℹ️ `@ledgerhq/types-cryptoassets` is strongly deprecated and will be dropped soon (see #20100).

As a coin module (public lib), replace type imports with `@ledgerhq/coin-framework` equivalents — `CryptoCurrency`, `TokenCurrency`, `Unit`, `CryptoCurrencyId` are all re-exported from there.

No runtime behavior change needed, types only.

libs/coin-modules/coin-vechain/src/test/msw-setup.ts
```diff
@@ -0,0 +1,7 @@
+import { setEnv } from "@ledgerhq/live-env";
+import { TEST_VECHAIN_ENDPOINT } from "./msw.mock";
+
+// Runs as a jest `setupFiles` entry (before the test modules load), so `constants/env.ts` — which
+// reads `API_VECHAIN_THOREST` into a module-level const at import time — resolves to the test host
+// that the MSW handlers intercept.
+setEnv("API_VECHAIN_THOREST", TEST_VECHAIN_ENDPOINT);
```
gre-ledger, line 7: ⚠️  there are changes in @ledgerhq/live-env that i suspect may be breaking here. notably setEnv if still needed, must be set up with injectDefinitions (you can see examples where it's used in some other tests), because that lib is no longer self aware of its own definitions (list of envs).

also i notice you switched away from live-env in the rest of the lib, so is it still needed ?

Sorry mate but you may need to rebase on latest origin/develop to confirm the tests still works. I suspect they won't.

libs/coin-modules/coin-vechain/jest.config.js
```diff
@@ -20,11 +10,47 @@ module.exports = {
...
+  testTimeout: 30_000,
+  collectCoverageFrom: [
+    "src/**/*.ts",
+    "!src/**/*.test.ts",
+    "!src/**/*.spec.ts",
+    "!src/test/**/*.ts",
+    "!src/datasets/**/*.ts",
+    "!src/__snapshots__/**/*.ts",
+  ],
+  coverageReporters: ["json", ["lcov", { file: "lcov.info", projectRoot: "../../../" }], "text"],
```
cted-ledger, line 32: SonarCloud is currently failing this PR: "0.0% Coverage on New Code (required ≥ 80%)", despite ~95% local coverage per the description. Config here mirrors coin-kaspa's, so likely a report-ingestion issue for this new package (lcov path/nx aggregation/new-code baseline) rather than an actual test gap — worth checking the Sonar analysis logs before merge since it's a hard gate.

cted-ledger (commented):
Second pass — the 5 points from the previous review (partial fee override, sender validation, broadcast coverage, craft→combine round trip, validateAddress via logic/) are all fixed and verified against the current diff. All outstanding Copilot threads look resolved too. One new item below (Sonar), non-blocking otherwise.

cted-ledger (dismissed):

vladyslavchupovskiy-ext-art (dismissed):

libs/coin-modules/coin-vechain/src/logic/validateIntent.ts
```diff
@@ -0,0 +1,96 @@
...
+  TransactionIntent,
+  TransactionValidation,
+} from "@ledgerhq/coin-module-framework/api/index";
+import {
+  AmountRequired,
+  InvalidAddress,
+  InvalidAddressBecauseDestinationIsAlsoSource,
+  NotEnoughBalance,
+  RecipientRequired,
+} from "@ledgerhq/errors";
```
gre-ledger: could you import these from other locations? they should all exists in 

```suggestion
} from "@ledgerhq/ledger-wallet-framework/errors";
```

gre-ledger (changes requested):
in context of incoming sunset of @ledgerhq/errors, please import these error from their other moved location

semeano (dismissed):

gre-ledger (approved):

vladyslavchupovskiy-ext-art (approved):

cted-ledger (approved):

PR #20019, opened 2026-07-26
[LWDM] feat(coin-tester): add VeChain coin-tester with chainTag (LIVE-34282)

pawell24:
### 📝 Description

Adds `@ledgerhq/coin-tester-vechain`, a deterministic integration tester for the VeChain family, plus the small production change it requires. **Depends on the CoinModuleApi migration (#20018 / LIVE-34273)** and is based on that branch — GitHub will retarget this PR to `develop` once the migration merges.

The tester spins up a local `vechain/thor` solo node via Docker and replays a VET + VTHO scenario through **both** bridge strategies — the legacy account bridge and the generic coin-framework bridge (Alpaca / CoinModuleApi) — asserting equivalence. Scenario (4 transactions per strategy): Send VET, Send VTHO, Send max VET, Send max VTHO. The max cases assert the real max-send invariant (VET balance drains to zero; the max VTHO send moves >99% of the prior balance, holding back only the reserved gas), exercising the VeChain-specific gas reservation (VTHO is both the sent asset and the gas currency). Mirrors the Kaspa coin-tester (LIVE-34179).

**Supporting production change — configurable `chainTag`:** thor-solo's genesis chainTag is not the mainnet value (74) that `coin-vechain` hardcoded, so the tester needs it configurable. `coin-vechain` now resolves it through a single validated `getChainTag()` helper (`config.ts`): it accepts an integer single byte (0–255) and falls back to the mainnet tag (74) on a missing or invalid remote override, used by `craftTransaction.ts`, `createTransaction.ts` and family `setup.ts`. Production is unaffected: no `chainTag` in config falls back to 74 (VeChain Live is mainnet-only). live-common ships the mainnet tag as the production default.

**Run:** `pnpm coin:tester:vechain start` (requires a running Docker daemon).

**Review feedback addressed:** register `vechain` in the CI `Coin Tester` matrix; migrate the package to knip (minimal `exports`, drop `.unimportedrc.json`); add a `docker-compose` healthcheck and pin `vechain/thor:v2.4.4`; add the `Status: STABLE` README banner and root-script Run command; add a changeset for the `chainTag` change.

**Result:** both strategies green (legacy + generic-adapter), confirmed on CI. Out of scope for this scenario: multi-clause, VIP-191 fee delegation, staking (see the package README "Limitations").

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34282](https://ledgerhq.atlassian.net/browse/LIVE-34282)
- **Depends on**: #20018 ([LIVE-34273](https://ledgerhq.atlassian.net/browse/LIVE-34273) CoinModuleApi migration)
- **ADR** (if any): —

[LIVE-34282]: https://ledgerhq.atlassian.net/browse/LIVE-34282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre-ledger:
⚠️ Please rebase on `develop` — #20076 just merged and we suspect it may break your `setEnv` calls in `env.setup.ts` (you'll need `@shared/env` bootstrapped before calling `setEnv`).

gre-ledger (dismissed):

gre-ledger (dismissed):

semeano (dismissed):

libs/coin-tester-modules/coin-tester-vechain/src/thorNode.ts
```diff
@@ -0,0 +1,50 @@
...
+        return parseInt(genesis.id.slice(-2), 16);
+      }
+      lastError = `/blocks/best → ${res.status}`;
+    } catch (err) {
+      lastError = String(err);
+    }
+    await new Promise(r => setTimeout(r, 1000));
+  }
+  throw new Error(`waitForThorReady timed out: ${lastError}`);
+}
```
francois-guerin-ledger: We should not need this if the healthcheck passes.

libs/coin-tester-modules/coin-tester-vechain/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+services:
+  thor-solo:
+    image: vechain/thor:v2.4.4
```
francois-guerin-ledger: We must add the sha256 to pin the version. I am currently reworking coin testers to make them running offline. Pinning versions is mandatory so we can rely on caches to avoid re-downloading them.
See https://github.com/LedgerHQ/ledger-live/pull/20323

francois-guerin-ledger (changes requested):

deepyjr (dismissed):

knip.json
```diff
@@ -122,6 +122,9 @@
     },
     "./features/flow/contacts": {
       "entry": ["src/jest.native.ts"]
+    },
+    "./libs/coin-tester-modules/coin-tester-vechain": {
+      "entry": ["src/scenarii/vechain.ts"]
```
gre-ledger: normally it shouldn't be necessary to edit knip.json. can you check if it works without it ? can you also try to define the explicit entry point in package.json exports ? 

recently we managed to reduce this file exception list a lot and actually a skill was poisoning agents to update this file.

pawell24: Fixed, removed the knip.json entry, added an explicit ./scenarii/vechain export instead.

francois-guerin-ledger (dismissed):

gre-ledger (dismissed):

francois-guerin-ledger (approved):

cted-ledger (approved):

Sebastien-Dav1d (approved):

PR #20043, opened 2026-07-27
ci(harden): fix templating throughout composites

lewisd5:
### 📝 Description

Refactor composites to phase out direct use of templating inside run scripts

https://github.com/LedgerHQ/ledger-live/actions/runs/30261550988
https://github.com/LedgerHQ/ledger-live/actions/runs/30340459705
https://github.com/LedgerHQ/ledger-live/actions/runs/30340511850

### DIFF
--- a/tools/actions/composites/adjust-changeset-level/action.yml
+++ b/tools/actions/composites/adjust-changeset-level/action.yml
@@ -15,23 +15,29 @@ runs:
       shell: bash
       run: |
         valid_levels=("patch" "minor" "major")
-        if [[ ! " ${valid_levels[@]} " =~ " ${{ inputs.from_level }} " ]]; then
-          echo "Invalid from_level: ${inputs.from_level}. Must be one of ${valid_levels[@]}"
+        if [[ ! " ${valid_levels[@]} " =~ " ${INPUTS_FROM_LEVEL} " ]]; then
+          echo "Invalid from_level: ${INPUTS_FROM_LEVEL}. Must be one of ${valid_levels[@]}"
           exit 1
         fi
-        if [[ ! " ${valid_levels[@]} " =~ " ${{ inputs.to_level }} " ]]; then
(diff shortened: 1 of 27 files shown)


ypolishchuk-ledger (dismissed):

henri-ly (dismissed):

tools/actions/composites/adjust-changeset-level/action.yml
```diff
@@ -15,23 +15,29 @@ runs:
...
       run: |
-        echo "Adjusting changeset level from ${{ inputs.from_level }} to ${{ inputs.to_level }}"
+        echo "Adjusting changeset level from ${INPUTS_FROM_LEVEL} to ${INPUTS_TO_LEVEL}"
         for file in .changeset/*.md; do
-          if grep -q ": ${{ inputs.from_level }}" "$file"; then
+          if grep -q ": ${INPUTS_FROM_LEVEL}" "$file"; then
             # target the lines between the "---" yaml delimiters, replace e.g. ": patch" with ": minor"
-            sed -i '/---/,/---/s/: ${{ inputs.from_level }}/: ${{ inputs.to_level }}/' "$file"
-            echo "Updated $file to ${{ inputs.to_level }}"
+            sed -i '/---/,/---/s/: ${INPUTS_FROM_LEVEL}/: ${INPUTS_TO_LEVEL}/' "$file"
```
dgreen-ledger: Heads up, I think a sed expression is in single quotes won't expand. 
It wasn't an issue before because ${{ inputs.from_level }} gets interpolated by the runner before bash, so the value was already baked in as plain text inside the single quotes. So I think we needs double quotes around the sed expression now

dgreen-ledger (approved):

henri-ly (approved):

ypolishchuk-ledger (approved):

PR #20070, opened 2026-07-27
[LWDM] refactor(currencies): drop dead @ledgerhq/cryptoassets store injection (LIVE-35038)

ysitbon:
## What

Removes the now-dead `@ledgerhq/cryptoassets` currency/fiat store injection from the app bootstraps, drops the dependency from the apps, and cleans up the remaining stale references to the package in comments / JSDoc.

## Why

After the value migration, nothing in-app reads the legacy `@ledgerhq/cryptoassets` currency/fiat accessors (`getCryptoCurrencyById`, `getFiatCurrencyByTicker`, …) anymore — the runtime source of truth is the domain-backed wallet-framework currency resolver (`setCurrenciesResolver`). So `setCryptoCurrenciesStore` / `setFiatCurrenciesStore` injected the domain registry into a store **no consumer queried** — dead code.

## Changes

- Remove `setCryptoCurrenciesStore` / `setFiatCurrenciesStore` calls + imports from `apps/cli`, `apps/ledger-live-desktop`, `apps/ledger-live-mobile`, `apps/web-tools` (the `setCurrenciesResolver` framework injection — the live path — is untouched).
- Drop `@ledgerhq/cryptoassets` from those apps' `package.json`.
- Reword the remaining inline-comment / JSDoc mentions of `@ledgerhq/cryptoassets` (`domain/*`, `@features/platform-currencies`, `@ledgerhq/live-common`, `@ledgerhq/coin-bitcoin`, `@ledgerhq/live-countervalues`, `@ledgerhq/wallet-btc`).

## Out of scope

- The parity tests inside `@ledgerhq/cryptoassets` (dual-maintenance guard — retired with the package drop).
- `@ledgerhq/cryptoassets-evm-signatures` (a different package).

## Verification

- Repo-wide `pnpm nx run-many -t typecheck` green (104 projects).
- `@ledgerhq/cryptoassets` (value) now has **zero in-repo consumers** — 0 `package.json` deps, 0 source imports of any subpath (`/state`, `/abandonseed`, `cal-client`, …).
- `pnpm install --frozen-lockfile` clean.

PR #20071, opened 2026-07-27
[LWDM] refactor(currencies): drop the @ledgerhq/cryptoassets package (LIVE-31961)

ysitbon:
## What

Final step of the crypto-assets data-layer migration: **completely remove the legacy `@ledgerhq/cryptoassets` value package** and every remaining trace of it.

Stacked on #20070 (which removed the last runtime injection and dropped the dependency from the apps). Review/merge that one first — the base of this PR is its branch.

## Why

The registry, CAL client, converters and runtime store have all been migrated to the data layer (`@domain/entity-currency-*`, `@domain/api-currency-token`, `@features/platform-currencies`). After #20070 the value package has **zero in-repo consumers**: no `package.json` depends on it and no source imports any of its subpaths. There is nothing left to keep it alive for internally, so it is removed rather than kept as a frozen shell.

## Changes

- Delete `libs/ledgerjs/packages/cryptoassets` in full (static registry, CAL client, `api-token-converter`/`api-asset-converter`, runtime store, parity tests).
- Delete the two domain seed scripts (`domain/entity/currency-{crypto,fiat}/scripts/generate-currencies.mts`) — they only existed to seed the domain registry from the legacy one.
- Remove the package from `CODEOWNERS`, the Sonar CPD exclusion, and the root `package.json` `*:ljs` Nx globs.
- Drop the stale entries from six coin `.unimportedrc.json` files.
- Reword the cross-team docs, READMEs and inline comments that still referenced dual-maintenance / parity with the legacy package.

## Governance

This removes a **published** package. External consumers (vault / revault) pin their own published version and are unaffected by its removal from this repo. No changeset is added — the package is deleted, not versioned, and the remaining edits are docs/comments only.

## Verification

- Global typecheck green across all projects (`nx run-many -t typecheck`).
- Residual scan for `@ledgerhq/cryptoassets` / `ledgerjs/packages/cryptoassets` / the seed scripts / domain-parity tests returns zero (excluding unrelated `*-evm-signatures`, `types-cryptoassets`, and historical `CHANGELOG.md` / changeset prose).

No behaviour change: nothing consumes the package at runtime.


gre-ledger (commented):
:itshappening:

gre-ledger (approved):

henri-ly (approved):

LucasWerey (approved):

PR #20075, opened 2026-07-27
refactor(currencies): remove specialized re-exports from @domain/entity-currency (LIVE-34763)

ysitbon:
## What

`domain/entity/currency/src/index.ts` re-exported everything from the specialized entity packages (`unit` / `crypto` / `token` / `fiat`) via `export *`. This removes those re-exports so **`@domain/entity-currency` exposes only the cross-package unions**: `Currency` / `CurrencySchema` and `CryptoOrTokenCurrency` / `CryptoOrTokenCurrencySchema`.

## Why

The `export *` barrel gave every specialized type two import paths (`@domain/entity-currency` **and** its own package) — a barrel smell that knip flags and that blurs the dependency graph. The aggregate should exist **only** for the cross-package unions.

Import convention going forward:
- **Specialized** types ← their own package: `CryptoCurrency` ← `@domain/entity-currency-crypto`, `TokenCurrency` ← `@domain/entity-currency-token`, `FiatCurrency` ← `@domain/entity-currency-fiat`, `Unit` ← `@domain/entity-currency-unit`.
- **Unions** ← `@domain/entity-currency` only.

## Changes

- Remove the four `export *` re-exports from `src/index.ts` (kept the union imports + definitions).
- Drop the now-unused `@domain/entity-currency-unit` dependency (it was only referenced by the removed `export *`).
- Update the README and package description to the unions-only convention.

## No repointing needed

This ticket is described as the *final* step, after all consumers stop importing specialized symbols via the aggregate. On `develop` the aggregate already has **zero in-repo consumers**: no `package.json` depends on it and no file imports `@domain/entity-currency`. So the removal is fully self-contained — nothing to repoint. Once the re-exports are gone, importing a specialized symbol through the aggregate is a compile error, so the optional `no-restricted-imports` guard is redundant and left out.

## Verification

- `@domain/entity-currency` unit tests green (unions still parse crypto/token/fiat correctly).
- Global `nx run-many -t typecheck` green across all projects.
- `pnpm install --frozen-lockfile` consistent; lockfile diff is the single removed importer entry.

### DIFF
--- a/domain/entity/currency/src/index.ts
+++ b/domain/entity/currency/src/index.ts
@@ -3,12 +3,6 @@ import { CryptoCurrencySchema } from "@domain/entity-currency-crypto";
 import { TokenCurrencySchema } from "@domain/entity-currency-token";
 import { FiatCurrencySchema } from "@domain/entity-currency-fiat";
 
-// Re-export everything from the three currency entity packages and unit
-export * from "@domain/entity-currency-unit";
-export * from "@domain/entity-currency-crypto";
-export * from "@domain/entity-currency-token";
-export * from "@domain/entity-currency-fiat";
-
 /**
  * Discriminated union of {@link CryptoCurrency} and {@link TokenCurrency}.
  * Use when handling on-chain currencies (no fiat).
--- a/.changeset/drop-aggregate-currency-reexports.md
+++ b/.changeset/drop-aggregate-currency-reexports.md
@@ -1,0 +1,5 @@
+---
+"@domain/entity-currency": minor
+---
+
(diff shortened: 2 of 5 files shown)


domain/entity/currency/src/index.ts
```diff
@@ -3,11 +3,8 @@ import { CryptoCurrencySchema } from "@domain/entity-currency-crypto";
 import { TokenCurrencySchema } from "@domain/entity-currency-token";
 import { FiatCurrencySchema } from "@domain/entity-currency-fiat";
 
-// Re-export everything from the three currency entity packages and unit
-export * from "@domain/entity-currency-unit";
-export * from "@domain/entity-currency-crypto";
-export * from "@domain/entity-currency-token";
-export * from "@domain/entity-currency-fiat";
+// Cross-package currency unions only. Specialized types live in their own
+// @domain/entity-currency-{unit,crypto,token,fiat} packages — import those directly.
```
gre-ledger: i think we should just drop the code without leaving extra comment

gre-ledger (approved):

PR #16217, opened 2026-04-09
[LWDM] chore(coin-tron): remove preload and hydrate

francois-guerin-ledger:
### ✅ Checklist

- [x] `npx changeset` was attached.
- [x] **Covered by automatic tests.** 
- [ ] **Impact of the changes:** 
  - ...

### 📝 Description

Tron no longer loads the **full Super Representative list** during currency preload or when hydrating persisted preload data. That work was moved to on-demand calls, still backed by the existing LRU cache. Vote rows in the UI rely on names loaded during account sync instead of joining against the global SR list.

##### `@ledgerhq/coin-tron`

Deleted `libs/coin-modules/coin-tron/src/bridge/preload.ts`

##### `@ledgerhq/live-common`

- `useTronSuperRepresentatives` — removed (it fetched the entire SR list on mount).
- `useSortedSr` — removed (search/sort over the full SR list for voting UI).
- `formatVotes` — signature simplified: no second `superRepresentatives` argument; output no longer includes `validator` / `rank` and derives `isSR` from vote order; `name` comes from the vote payload (filled when building Tron resources during sync).

##### Apps (LLD / LLM)

Tron Votes and related screens updated to match the above (no global SR hook; display uses synced vote metadata).

#### Stacked change (already on this branch)

[#16228 — chore(coin-tron): do not fetch validators metadata](https://github.com/LedgerHQ/ledger-live/pull/16228) was merged into this branch. It changes `fetchSuperRepresentatives` to build the list from `/wallet/listwitnesses` only, removes per-witness enrichment (display names and brokerage), and drops the batched work that called into `accountNamesCache` for every witness.

Reviewer note
The vote flow may still pay a one-time cost to load the witness list when the user starts voting (on-demand, then cached). The earlier “temporary regression” framing referred to heavy per-witness metadata fetching; that part is addressed by the stacked PR above. If latency is still unacceptable in practice, capture timings or network traces so we can tune call sites or caching.

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-28683
- **ADR link (if any)**: 

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
[...]

PR #20111, opened 2026-07-28
[LWDM] refactor(solana): drop preload, lazily cache validators (LIVE-34639)

YazhuEth:
### 📝 Description

**The Solana validator list is now fetched on demand, instead of during the currency preload.**

Before, the whole Solana validator list was fetched during `CurrencyBridge.preload` and pushed into a global `BehaviorSubject`. That fetch is asynchronous and sat on the critical path of the scan account flow, so it slowed down adding an account. `preload` / `hydrate` are also deprecated.

Now the list is fetched lazily, the first time a screen needs it, behind a 15min LRU cache (the same freshness window the preload strategy used). Same approach as TRON in #16217.

**What changed**

- `coin-solana`: new `src/validators.ts` (`getSolanaValidators`) replaces `preload.ts` + `preload-data.ts`. Mainnet keeps its `profitableValidators` → `ledgerFirstValidators` ordering, testnet stays unfiltered, devnet still derives validators from `getVoteAccounts()`. `currencyBridge` is now just `{ scanAccounts, nftResolvers }`.
- `live-common`: `useValidators` and `useSolanaStakesWithMeta` keep the exact same signatures, so 7 of the 9 consuming UI files are untouched. Only `useSolanaPreloadData` is removed — use `useValidators` instead (both call sites migrated). `getAccountBannerState` now receives the validators as an argument instead of reading the global store.

**What users see**

On a cold start the list is empty for one render instead of until the preload completed. Every consumer already handled that: the delegation picker renders an empty list, operation details fall back to the raw vote address, and the stake banner stays hidden until the Ledger validator is known.

When a fetch fails, the last known list is kept rather than cleared — the preload store behaved the same way, since it only ever published on success, so a failed refresh left the previous value in place. The list is only empty when nothing has been loaded yet in the session. Staleness is bounded: rejections are not cached, so every screen mount retries, and a successful list is refreshed after 15min. A stale list cannot produce a bad transaction either, since `prepareTransaction` validates the vote account on-chain.

**Tests**

Nothing is broken — the suites were re-run after rebasing on `develop`:

- `coin-solana`: 412 unit tests, 39/39 suites. `validators.test.ts` replaces `preload.test.ts` and adds cache dedup / `clear()` / "failures are not cached".
- `live-common` solana: 37 tests, 6/6 suites, passing **with no network access at all**. `react.integration.test.ts` hits the real endpoint with no mocking; `react.test.ts` covers only what the live API cannot produce on demand (first-fetch failure, keeping the last known list on a later failure, retry on the next mount, and the stake metadata join); `banner.test.ts` passes validators as an argument instead of mocking the preload store.
- Typecheck clean on coin-solana, live-common, desktop and mobile.
- **Coin-tester**: passes for both the legacy and generic-adapter strategies, full staking flows included.

**E2E**: no impact, and nothing to update. No E2E spec exercises Solana delegation or the validator list — the only Solana mentions are the send flow and wallet-api, which never touch validators. Note the mocked bridge used under `PLAYWRIGHT_RUN` / `DETOX` was never the source of this list anyway: on mainnet the old `preloadWithAPI` also called the real `getValidators(cluster)` and only used the injected (mocked) `ChainAPI` for devnet. So E2E behaviour is unchanged.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34639](https://ledgerhq.atlassian.net/browse/LIVE-34639)
- Reference implementation for TRON: #16217
- The unrelated `getBalance.integ.test.ts` failure on this branch is fixed in #20135 (Agave 4.1 removed `warmupCooldownRate` from the RPC output); it is not caused by this PR.

[LIVE-34639]: https://ledgerhq.atlassian.net/browse/LIVE-34639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


YazhuEth:
**E2E**:
- Desktop : https://github.com/LedgerHQ/ledger-live/actions/runs/30349030584
- Mobile: https://github.com/LedgerHQ/ledger-live/actions/runs/30349055382

YazhuEth:
**E2E**:
- Desktop : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/90d72f5d-6171-4387-b8b1-99b45c85a564/
- Mobile: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/fb9ead7d-0e4a-456a-abfc-ada82c735119/

YazhuEth:
*E2E*
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/1219ec19-99fd-4f59-95ac-9363c8747741/
- http://ledger-live.allure.green.ledgerlabs.net/allure/reports/45a21ba1-899f-42bf-ba2d-093f5259334b/

YazhuEth:
**E2E**
D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/1c4453ec-6b26-4e16-9b96-3488692fa0ef/
M: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/38e6fd15-5801-4b7e-906a-d0b9c9af6d4c/

henri-ly (approved):

libs/coin-modules/coin-solana/src/validators.ts
```diff
@@ -0,0 +1,43 @@
...
+import type { ValidatorsAppValidator } from "./network/validator-app";
+import {
+  clusterByCurrencyId,
+  endpointByCurrencyId,
+  ledgerFirstValidators,
+  profitableValidators,
+} from "./utils";
+
+// one entry per solana currency id
+const VALIDATORS_CACHE = minutes(15, 3);
```
francois-guerin-ledger, line 16: What is the rational behind parameter `3` ?

francois-guerin-ledger (approved):

libs/coin-modules/coin-solana/src/validators.ts
YazhuEth: The second arg is the LRU entry count : `minutes(num, max)` --> `minutes(15, 3)` gives `{ max: 3, ttl: 900000 }`. The cache is keyed by currencyId and clusterByCurrencyId only accepts solana, solana_devnet and solana_testnet so there can never be more than 3 entries.

PR #20117, opened 2026-07-28
fix(celo): handle epoch-processing block during voting (LIVE-32861)

shazzzam:
## ✅ Checklist

- [ ] Tests
- [x] Bug fix
- [ ] Breaking change

## 📝 Description

Fixes **LIVE-32861** — Celo vote flow failed with an opaque **"RPC request Failed"**, blocking users from voting.

### Root cause
Since the Celo **L1→L2 migration**, epoch processing runs on-chain in a short recurring window (`startNextEpochProcess()` → `finishNextEpochProcess()`). During that window the `EpochManager` acts as a `Blocker`, and `Election.vote()` (via the `onlyWhenNotBlocked` modifier) reverts with `Contract is blocked from performing this action`. Our `eth_estimateGas` reverted and, because nothing classified it, surfaced as a generic RPC failure at the signing step.

This is a **transient on-chain state, not a defect** — the same vote succeeds once the window closes. The fix is a UX one: detect the block and show a specific, actionable message (wording approved by Product).

### Changes
- **coin-celo**
  - New `CeloEpochProcessingActive` error + `isEpochBlockRevert()` classifier for the block revert.
  - `getTransactionStatus` (primary): for `vote`, probe `Election.isBlocked()` — the same condition the on-chain modifier checks — and set an inline `amount` error, disabling **Continue** and showing the message *before* signing. A transient probe failure is swallowed so it never false-blocks.
  - `buildTransaction` (safety net, vote-only): map the block revert to `CeloEpochProcessingActive` instead of the raw RPC error, covering the race where the block starts after the status check.
- **i18n (en)**: added `errors.CeloEpochProcessingActive` (title + description) to LLD `app.json` and LLM `common.json`.
- **Tests**: `isEpochBlockRevert` cases; `getTransactionStatus` blocked / not-blocked / precedence-over-NotEnoughBalance / transient-probe; `buildTransaction` vote-block classification + non-vote passthrough.

### 🔗 Context

- **JIRA / GitHub issue**:  [LIVE-32861](https://ledgerhq.atlassian.net/browse/LIVE-32861)
- **ADR** (if any):

### DIFF
--- a/libs/coin-modules/coin-celo/src/bridge/getTransactionStatus.ts
+++ b/libs/coin-modules/coin-celo/src/bridge/getTransactionStatus.ts
@@ -1,4 +1,4 @@
-import { lockedGoldABI } from "@celo/abis";
+import { electionABI, lockedGoldABI } from "@celo/abis";
 import { isAddress } from "viem";
 import {
   AmountRequired,
@@ -12,7 +12,7 @@ import {
 import { findSubAccountById } from "@ledgerhq/ledger-wallet-framework/account/index";
 import { AccountBridge } from "@ledgerhq/types-live";
 import { BigNumber } from "bignumber.js";
-import { CeloAllFundsWarning, CeloGroupNotVotable } from "../errors";
+import { CeloAllFundsWarning, CeloEpochProcessingActive, CeloGroupNotVotable } from "../errors";
 import { getPendingStakingOperationAmounts, getVote } from "../logic";
 import { getCeloClient } from "../network/client";
 import { getRegistryAddressFor } from "../network/registry";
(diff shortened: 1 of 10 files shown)


libs/coin-modules/coin-celo/src/bridge/getTransactionStatus.ts
```diff
@@ -170,6 +170,29 @@ export const getTransactionStatus: AccountBridge<
...
+      const electionAddress = await getRegistryAddressFor("Election");
+      const blocked = await client.readContract({
+        address: electionAddress,
+        abi: electionABI,
+        functionName: "isBlocked",
+      });
+      if (blocked) {
+        errors.amount = new CeloEpochProcessingActive();
+      }
+    } catch {
```
amaslakov: Does it make sense to at least log here? So that some wrong registry address or a decoding failure don't get swallowed silently?
Something like:

```ts
} catch (error) {
  log("celo/getTransactionStatus", "isBlocked probe failed", { error });
}

amaslakov (dismissed):

amaslakov (approved):

PR #20121, opened 2026-07-28
[LWDM] getBlock support

amaslakov:
### 📝 Description

getBlock was stubbed as "not supported" (no per-block endpoint). The wallet-proxy now ships GET /v0/blockTransactionEvents/{blockHash} (proxy #180), so this implements getBlock(height): Promise<Block> per the coin-module-framework contract.

- info reuses getBlockInfo (height → hash → info).
- Each tx summary → BlockTransaction (hash, fees, feesPayer, failed).
- Tag-based events: Transferred → signed sender/recipient pair; TransferMemo → memo decoded into details; rejected/other → no ops. Empty block → [].
- Shared CBOR memo decoder extracted to memo.ts (also used by listOperations).

```ts
const api = createApi(config, "concordium");
const { info, transactions } = await api.getBlock(1_000_000);
```

Covered by unit + integ tests (integ hits live testnet; schema verified against a real transferWithMemo). Not wired into the apps — this is the CoinModuleApi surface, not the bridge.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-30911](https://ledgerhq.atlassian.net/browse/LIVE-30911?atlOrigin=eyJpIjoiYmNmNmZjODMxZDU3NDhjOWJiNTY1YWYzYzcxOGE5MGYiLCJwIjoiaiJ9)
- **ADR** (if any): 

[LIVE-30911]: https://ledgerhq.atlassian.net/browse/LIVE-30911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-modules/coin-concordium/src/logic/history/getBlock.ts
+++ b/libs/coin-modules/coin-concordium/src/logic/history/getBlock.ts
@@ -1,0 +1,104 @@
+import type {
+  Block,
+  BlockOperation,
+  BlockTransaction,
+} from "@ledgerhq/coin-module-framework/api/index";
+import { getBlockTransactionEvents } from "../../network/proxyClient";
+import type {
+  BlockTransactionEvent,
+  BlockTransactionSummary,
+  ConcordiumEventAddress,
+  TransferMemoEvent,
+  TransferredEvent,
+} from "../../types";
+import { getBlockInfo } from "./getBlockInfo";
+import { decodeMemo } from "./memo";
+
+const NATIVE_ASSET = { type: "native" } as const;
+
+/**
+ * Returns a full block (info + transactions) for a given height.
+ * `info` reuses the getBlockInfo logic (height → hash → info); transactions come
+ * from the per-block events endpoint and are mapped tag-by-tag.
+ */
+export async function getBlock(height: number, currencyId: string): Promise<Block> {
+  const info = await getBlockInfo(height, currencyId);
+  const summaries = await getBlockTransactionEvents(currencyId, info.hash);
+
+  return {
+    info,
+    transactions: summaries.map(toBlockTransaction),
+  };
+}
+
+function toBlockTransaction(summary: BlockTransactionSummary): BlockTransaction {
+  const { operations, memo } =
+    summary.result.outcome === "success"
+      ? mapEventsToOperations(summary.result.events, summary.hash)
+      : { operations: [] as BlockOperation[], memo: undefined };
+
+  return {
+    hash: summary.hash,
+    failed: summary.result.outcome === "reject",
+    fees: BigInt(summary.cost),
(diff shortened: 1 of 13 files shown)


libs/coin-modules/coin-concordium/src/types/network.ts
```diff
@@ -50,6 +50,72 @@ export interface BlockInfoResponse {
...
+export type BlockTransactionResult =
+  | { outcome: "success"; events: BlockTransactionEvent[] }
+  | { outcome: "reject"; rejectReason?: { tag: string; contents?: unknown } };
+
+/**
+ * A transaction summary as returned by /v0/blockTransactionEvents/{blockHash}.
+ * This is the raw node `TransactionSummary` shape (structured, tagged events),
+ * distinct from the flattened /v3/accTransactions shape used by listOperations.
+ */
+export interface BlockTransactionSummary {
```
lysyi3m, line 103: maybe `BlockTransactionEvent`? 

and then 

`export type BlockTransactionEventsResponse = BlockTransactionEvent[];` as `/v0/blockTransactionEvents/`'s response

libs/coin-modules/coin-concordium/src/logic/history/getBlock.ts
```diff
@@ -0,0 +1,104 @@
...
+const NATIVE_ASSET = { type: "native" } as const;
+
+/**
+ * Returns a full block (info + transactions) for a given height.
+ * `info` reuses the getBlockInfo logic (height → hash → info); transactions come
+ * from the per-block events endpoint and are mapped tag-by-tag.
+ */
+export async function getBlock(height: number, currencyId: string): Promise<Block> {
+  const info = await getBlockInfo(height, currencyId);
+  const summaries = await getBlockTransactionEvents(currencyId, info.hash);
```
lysyi3m, line 26: same as the type in types/network - we're introducing new concept of _summaries_, while fetch block transaction _events_. might be misleading in terms of maintainability

libs/coin-modules/coin-concordium/src/types/network.ts
amaslakov: `BlockTransactionSummary` isn't an event — it's a transaction record that contains events (result is `{ outcome: "success"; events: BlockTransactionEvent[] }` on success, `rejectReason` otherwise). It maps 1:1 to the node's `TransactionSummary`, which likewise contains events rather than being one. To make the `summary` -> `events` hierarchy clearer, I could rename the inner union `BlockTransactionEvent` -> `TransactionEvent`. Does that address it, or would you prefer a different split?

lysyi3m (dismissed):

lysyi3m (approved):

PR #20160, opened 2026-07-28
[LWM] fix(e2e): select mobile specs by tag not raw file content

ypolishchuk-ledger:
### ✅ Checklist

- [ ] `npx changeset` was attached. _Not required: this change is E2E/CI tooling only (no shipped app code) — consistent with prior fixes to this script (e.g. `9277cc2ad5e`) which merged without a changeset._
- [ ] **Covered by automatic tests.** _Validated with a local `node:test` suite (15 cases, kept out of this PR) plus real-CI validation (see Proof). Happy to wire the suite into CI if wanted._
  - Mobile E2E test selection when a `Filter pattern` is provided to the on-demand / scheduled mobile E2E workflows
  - A `swap` filter runs the swap suite (`e2e/mobile/specs/swap/**`) and no unrelated Wallet XP (`wallet40Q2`) specs
  - Tag filters (`@smoke`, `@family-evm`, device tags) still select the right specs
  - Single-file / folder targeting still works (e.g. `wallet40Q2/portfolio.spec.ts`)

### 📝 Description

**Problem**: The mobile E2E filter selects whole spec **files**, and a file was picked if the filter regex matched its **entire raw text** (`filterRegex.test(fs.readFileSync(file))`). So a filter like `swap` pulled in any spec that merely *contained* the substring "swap" anywhere — in a page-object method name (`pressQuickActionSwapButton`), an `it()` title, or even a **code comment**. In a recent run this added ~27 unrelated Wallet XP (`wallet40Q2`) tests to a `swap`-filtered run — including `assetDiscoverability.spec.ts`, which matched purely on a comment.

**Solution**: Match a spec only by its **file path** (preserves single-file / folder targeting) and its **declared `@`-tags** (`$Tag("@x")`, `tags: ["@x", ...]`, helper tag arrays). Comments, titles and method names can no longer accidentally select a spec. This mirrors the desktop Playwright `--grep` (title + tags) behaviour at the file-selection granularity Detox requires.

**Before / After** (real `e2e/mobile/specs`, filter = `swap`):

| Spec | Before | After |
| --- | --- | --- |
| `specs/swap/**` | ✅ 46 | ✅ 46 |
| `wallet40Q2/portfolio.spec.ts` (swap button / title) | ✅ selected | ❌ not selected |
| `wallet40Q2/assetDiscoverability.spec.ts` (swap in a **comment**) | ✅ selected | ❌ not selected |

**Behaviour change to note**: filtering by a TMS id (e.g. `B2CQA-2750`) no longer selects a spec, since it is neither a path nor a tag — this aligns mobile with desktop (which greps titles + tags, not TMS annotations).

### 🗂️ Relocation (also in this PR)

The selector previously lived, confusingly, in the app package. It now sits where it belongs:

- Selection logic → `e2e/tooling/filter/selectSpecs.mjs` (alongside `resolve.mjs` / `escaping.mjs`)
- Sharding orchestrator `shard-tests.mjs` → `e2e/mobile/scripts/` (next to its detox/jest config)
- CI call sites updated to the new path (`generate-shards-matrix` composite + `test-mobile-mock-reusable.yml`)

### 🔬 Proof (CI validation)

- [run 30437024839](https://github.com/LedgerHQ/ledger-live/actions/runs/30437024839) — *Determine Builds & Generate Shards* = **success**; job log confirms **46 swap specs selected, 0 Wallet XP (`wallet40Q2`)** using the relocated path.
- Earlier iOS & Android run: https://github.com/LedgerHQ/ledger-live/actions/runs/30375586240

### ⚠️ Before merge — required coordination

- ✅ **Temporary CI pin reverted** — `generate-shards-matrix` is back to `@develop` (the branch pin `9d43f66` was only used to validate the relocated path on-branch; restored in `ba23514`).
- **Rebase coordination:** because the composites resolve `@develop`, once this merges any in-flight branch that hasn't rebased will fail the shards step (develop's composite points at the new path while their old-path file still exists). Teams need to rebase past this merge.

### ❓ Context

- **JIRA or GitHub link**: N/A — internal E2E tooling fix (can create a ticket if required)

---


apps/ledger-live-mobile/scripts/shard-tests.mjs
```diff
@@ -45,15 +46,35 @@ function findTestFiles(dir) {
...
 }
 
-function filterTestFiles(files, testFilter) {
+// Extract the tags a spec *declares*, e.g. "@smoke", "@NanoSP", "@family-evm".
+// Detox/jest-allure2 tags are always authored as quoted string literals starting
+// with "@" — via `$Tag("@x")`, `tags: ["@x", ...]`, or arrays passed to helpers
+// like `runSwapTest(..., ["@ethereum", ...])`. Matching only these (instead of the
+// whole file text) prevents a spec from being selected just because the filter word
+// appears in a comment, a describe/it title, or a page-object method name
+// (e.g. `pressQuickActionSwapButton`).
```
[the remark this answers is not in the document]
ypolishchuk-ledger: Done

ysitbon (dismissed):

thesan (dismissed):

e2e/tooling/filter/selectSpecs.mjs
```diff
@@ -0,0 +1,75 @@
...
+        results.push(fullPath);
+      }
+    }
+  } catch (e) {
+    if (e.code === "ENOENT") return [];
+    console.error("[select-specs] Error reading directory:", dir, e);
+    throw new Error(`Failed to read directory ${dir}: ${e.message}`);
+  }
+  return results.sort(compareStrings);
+}
```
jeportie, line 47: I think there is some redundant sorting here :
I see 3 levels of sorting and for me the last one overrides all the others :

1. `entries.sort()`
2. `result.sort()` on every recursion
3. then the final `result.sort()`

I would suggest to collapse to a single sort.
Also i think the error path may cause some issues here. As the recursive call sits inside a try, failures gets re-caught and re-wrapped at every level on the way up. Collapsing to a single sort will also solve this as the recursion will sit outside the try.

I would suggest something like :

```suggestion
const isSpecFile = name =>
  name.endsWith(".spec.ts") && !name.endsWith(".skip.spec.ts");

function collectTestFiles(dir) {
  const out = [];
  let entries;
  try {
    entries = fs.readdirSync(dir, { withFileTypes: true });
  } catch (e) {
    if (e.code === "ENOENT") return [];
    throw new Error(`Failed to read directory ${dir}: ${e.message}`, { cause: e });
  }
  for (const entry of entries) {
    const fullPath = path.join(dir, entry.name);
    if (entry.isDirectory()) out.push(...collectTestFiles(fullPath));
    else if (isSpecFile(entry.name)) out.push(fullPath);
  }
  return out;
}

export function findTestFiles(dir) {
  return collectTestFiles(dir).sort(compareStrings);
}
```

ypolishchuk-ledger: Done

jeportie (dismissed):

ysitbon (dismissed):

thesan (dismissed):

jeportie (dismissed):

lewisd5 (dismissed):

lewisd5 (approved):

jeportie (approved):

ysitbon (approved):

PR #20161, opened 2026-07-28
docs(skills): add impacting-prs skill

gre-ledger:
`/impacting-prs inform ongoing PRs that they need to stop using @ledgerhq/errors lib`

Screenshot 2026-07-30 at 17 09 16

---

other example using Study mode:

```
/impacting-prs https://github.com/LedgerHQ/ledger-live/pull/20251
```

Screenshot 2026-07-30 at 17 21 30

Screenshot 2026-07-30 at 17 39 47

> [!NOTE]
> I'll personally give it a try on a real migration and will likely iterate on it a bit more. Anyone is free to try it locally too — feedback very welcome.

### 📝 Description

New `impacting-prs` skill: when a change lands in `develop` that every in-flight PR must follow (a package sunset, an API deprecation, a moved module, a new architectural rule), it studies which open PRs are impacted, explains the concrete impact per PR, and posts tailored comments — a blocking review when the breakage is already in `develop`, a heads-up when the artifact is only deprecated.

Born out of the `@ledgerhq/cryptoassets` sunset, where ~20 open PRs each had to be told what to migrate to, with the answer depending on the layer (a public lib cannot reach app-internal domain code, while `libs/coin-modules/*` go through the coin framework).

Workflow in 6 steps:

1. **Pin down the change** — ask for whose behalf we speak, the merged PRs implementing the migration, and dropped-vs-deprecated status; then read those diffs to derive a **per-layer migration matrix** (detection signal → replacement → severity), confirmed by the user before anything is scanned.
2. **Select PRs** — `scan-prs.sh` keeps only PRs targeting `develop` that are not already conflicting (those rebase anyway) and greps the detection pattern on **added lines only**, so a PR removing the old import is not flagged for doing the migration.
3. **Analyse** — drop false positives, map each file to a matrix row, and check the recommended API really exists on `develop`.
4. **Draft + approval** — hedged, on-behalf-of tone; nothing is posted before the user reads and approves every draft.
5. **Post** — one review per PR: short summary body + one comment anchored on the offending line, with a ` ```suggestion ` block when the fix is mechanical.
6. **Recap** — Slack-ready summary grouped by severity, linking each posted comment, plus what was skipped and why.

The scanner outputs the RIGHT-side line number for every hit, so comments can be anchored without recomputing positions:

```
pr	path	line	content
20034	libs/ledger-live-common/src/families/polkadot/react.ts	4	import type { CryptoCurrency } from "@ledgerhq/types-cryptoassets";
```

```bash
.agents/skills/impacting-prs/scan-prs.sh -p 'cryptoassets' -o "$TMPDIR/scan"
# → prs.json (eligible PRs) · diffs/<n>.diff · hits.tsv (pr ⇥ path ⇥ line ⇥ content)
```

Validated against the live repo: 56 eligible PRs scanned in ~13s, and the four anchors it reported were checked line-by-line against the PR head files (all exact). Two gotchas are baked in: GitHub computes mergeability lazily, so the first `gh pr list` returns `UNKNOWN` for most PRs and the script queries twice (that alone decides who gets skipped); and BSD `xargs -I` caps lines at 255 bytes, so diffs are fetched from inside the output dir with `GH_REPO` pinned.

Docs only — no product code, no changeset.

### 🔗 Context

- **JIRA / GitHub issue**: n/a — tooling
- **ADR** (if any): n/a

### DIFF
--- a/.agents/skills/impacting-prs/SKILL.md
+++ b/.agents/skills/impacting-prs/SKILL.md
@@ -1,0 +1,175 @@
+---
+name: impacting-prs
+description: Find which open PRs are impacted by a migration/sunset/refactor and notify their authors — blocking review when the old path is already gone from develop, heads-up comment when it is only deprecated or its removal is still in review, silent study report only when the replacement does not exist yet. Use when asked to check/flag/notify/block open PRs about a breaking change, deprecation or migration they need to follow.
+---
+
+# Notify open PRs impacted by an ongoing migration
+
+**When:** a change is landing (or has landed) that every in-flight PR must follow — a package sunset, an API deprecation, a moved module, a new architectural rule.
+
+**Three modes.** What gates the ping is **whether the replacement is already on `develop`**, not where the removal lives — an author can only act on advice they can commit today. Whether the old path is *already gone* only sets the severity:
+
+| Replacement on `develop`? | Old path on `develop`? | Mode | What we do |
+| --- | --- | --- | --- |
+| ✅ yes | gone | 🔴 **Blocking** | review requesting changes on each impacted PR |
+| ✅ yes | still there (deprecated, or removed only in a pending PR) | 🟠 **Heads-up** | comment on each impacted PR |
+| ❌ not yet | anything | 🔍 **Study** | **no ping.** Report the blast radius to the user, or as one comment on their own migration PR |
+
+The middle row is the common case and the one easy to get wrong: a sunset PR usually only deletes the *last legacy alias* of something whose successor landed weeks ago. The impacted authors can migrate right now, so ping them — the migration PR being unmerged is a reason to say "landing soon", not a reason to stay silent. Reserve 🔍 study for when the successor genuinely does not exist yet.
+
+**Hard rules:**
+
+- 🚫 Never post anything before the user has read and approved the drafts.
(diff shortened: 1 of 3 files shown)


gre-ledger:
- [x] it appears to me that the skill tend to want to comment again on a PR that was already commented on. we need to improve the skill so it skips the PRs that were already informed OR that already conflict with develop.

gre-ledger:
- [x] reinforce that the message need to be in English. when prompting in French it was about to send French comments XD

gre-ledger:
- [x] actually I would like to prefix a 🤖  icon before the message. so it would be like "🤖 On behalf of @gre-ledger, bla bla bla". ex https://github.com/LedgerHQ/ledger-live/pull/20181#pullrequestreview-4819075837 but also don't repeat on behalf of @gre-ledger  on the individual code comment.

gre-ledger:
- [x] today this skill focus on Heads Up OR Blocking and will directly ping on PRs. I think we also should allow a mode Study, that is typically in context of an ongoing draft with work that is not yet on develop and that other PRs can't actually migrate too. in such case, a comment report could be done directly on the ongoing PR of the user if any. otherwise just report to that prompt user.

ysitbon (approved):

PR #20164, opened 2026-07-28
[LWM] fix(counter-values): keep persisted countervalue on restart (LIVE-35110)

LucasWerey:
### ✅ Checklist

      - Countervalue persistence across app kill/restart, especially for **uncommon fiats not in the offline fallback list** (e.g. AMD, and other less-common ISO currencies).
      - Common fiats (USD, EUR, GBP…) still behave as before.
      - A genuinely unsupported/delisted countervalue still falls back to USD once the CVS supported-fiats query has settled.
      - App boot / store hydration path (`LedgerStore`) — no crash or regression in the countervalue init.

### 📝 Description

> ⚠️ **This is a suggested fix that I'll send to the platform team** (owners of the RTK-Query supported-fiats / countervalues migration) for review before merge — sharing the diagnosis and a proposed patch rather than merging directly.

**Problem** — After selecting a less common countervalue (e.g. Armenian Dram, AMD) in Settings, killing the app, and restarting it, the balance is displayed in USD again instead of the selected currency.

**Root cause** — The recent RTK-Query supported-fiats migration introduced a two-stage flow: the `supportedFiats` slice starts with a static **offline fallback list** (`FALLBACK_FIAT_TICKERS`, ~36 common fiats, `fiatsReady: false`) and is only replaced by the full CVS list once `useGetSupportedFiatsQuery()` resolves. At store bootstrap, `LedgerStore.updateSupportedCountervalues` ran the *"reset unsupported countervalue to USD"* guard **synchronously, before the CVS query had resolved** — i.e. against the fallback list only. Since AMD (and other uncommon fiats) are valid CVS fiats but absent from the fallback list, the guard reset the persisted countervalue to USD and re-persisted it on every launch.

**Solution** — Move the reset out of the boot path and into the reactive `InitialQueriesProvider` effect, which is already gated on `fiatsReady`, so the "unsupported → USD" decision is only made against the **authoritative CVS list**. Boot now only seeds `supportedCounterValues` from the fallback (harmless) and never mutates the user's countervalue.

- `LedgerStore.tsx` — `updateSupportedCountervalues` no longer resets the countervalue; it just seeds the supported list from the fallback fiats.
- `InitialQueriesContext.tsx` — the existing `fiatsReady`-gated effect now also resets the countervalue to USD only when the authoritative supported list does not contain it.
- `test-renderer.tsx` — inject `cvsApiExtra` (CVS base URL) into the test store's thunk `extraArgument`, mirroring production, so `useGetSupportedFiatsQuery` resolves against MSW in integration tests.

**Verification**
- 2 new integration tests (MSW-mocked `/v3/supported/fiat`, no live API):
  - persisted AMD is kept once CVS confirms it is supported (regression guard);
  - persisted fiat is reset to USD once CVS reports it unsupported (safety net preserved).
- Manually verified on simulator: with the fix, the balance renders in AMD (֏) after relaunch instead of resetting to USD.

Uploading Simulator Screen Recording - iPhone 17 Pro Max - 2026-07-28 at 23.05.18.mov…

### ❓ Context

- **JIRA or GitHub link**: [LIVE-35110](https://ledgerhq.atlassian.net/browse/LIVE-35110)

---

[LIVE-35110]: https://ledgerhq.atlassian.net/browse/LIVE-35110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/contexts/InitialQueriesContext.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/contexts/InitialQueriesContext.tsx
@@ -21,7 +21,7 @@ export function InitialQueriesProvider({ children }: React.PropsWithChildren) {
   // OFAC Geo Blocking
   const ofacQueryResult = ofacGeoBlockApi.useCheckQuery();
   // Boot-time fiat fetch: onQueryStarted → setFiats populates the supportedFiats slice
-  useGetSupportedFiatsQuery();
+  const { isSuccess: fiatsResolvedFromCvs } = useGetSupportedFiatsQuery();
 
   const fiatsReady = useSelector(selectSupportedFiatsReady);
 
@@ -29,13 +29,16 @@ export function InitialQueriesProvider({ children }: React.PropsWithChildren) {
   const fiats = useSelector(selectSupportedFiats);
   const counterValueCurrency = useSelector(counterValueCurrencySelector);
   useEffect(() => {
-    // Gate on fiatsReady so we act on the authoritative CVS list, not the boot-time
-    // fallback — resetting against the fallback would wipe valid but uncommon fiats.
     if (!fiatsReady) return;
     const supportedCounterValues = buildSupportedCounterValues(fiats);
     dispatch(setSupportedCounterValues(supportedCounterValues));
 
-    // Safety net: if the persisted counterValue is no longer supported by CVS, fall back to USD.
+    // Reset an unsupported persisted counterValue to USD, but only once CVS actually returned
+    // its list: fiatsReady also flips true on a CVS error/timeout (setFiatsReady runs in a
+    // finally), and then `fiats` is just the offline fallback — resetting there would wipe a
+    // valid uncommon fiat (e.g. AMD) and persist USD. A delisted fiat is corrected on the next
+    // successful load.
+    if (!fiatsResolvedFromCvs) return;
     const currentId = counterValueIdOf(counterValueCurrency);
     const isSupported = supportedCounterValues.some(
       ({ currency }) => counterValueIdOf(currency) === currentId,
@@ -43,7 +46,7 @@ export function InitialQueriesProvider({ children }: React.PropsWithChildren) {
     if (!isSupported) {
       dispatch(setCountervalue("USD"));
     }
-  }, [dispatch, fiats, fiatsReady, counterValueCurrency]);
+  }, [dispatch, fiats, fiatsReady, fiatsResolvedFromCvs, counterValueCurrency]);
   const ofacResult = useMemo(
     () => ({ blocked: ofacQueryResult.data ?? false, isLoading: ofacQueryResult.isLoading }),
     [ofacQueryResult.data, ofacQueryResult.isLoading],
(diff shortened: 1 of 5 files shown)


apps/ledger-live-mobile/src/mvvm/contexts/InitialQueriesContext.tsx
```diff
@@ -26,10 +27,23 @@ export function InitialQueriesProvider({ children }: React.PropsWithChildren) {
...
+
+    // Safety net: if the persisted counterValue is no longer supported by CVS, fall back to USD.
+    const currentId = counterValueIdOf(counterValueCurrency);
+    const isSupported = supportedCounterValues.some(
+      ({ currency }) => counterValueIdOf(currency) === currentId,
+    );
+    if (!isSupported) {
+      dispatch(setCountervalue("USD"));
+    }
+  }, [dispatch, fiats, fiatsReady, counterValueCurrency]);
```
ysitbon: We might be missing a case. Something like this may cover all:

```diff
-  useGetSupportedFiatsQuery();
+  const { isSuccess: fiatsResolvedFromCvs } = useGetSupportedFiatsQuery();
   ...
   useEffect(() => {
     if (!fiatsReady) return;
     const supportedCounterValues = buildSupportedCounterValues(fiats);
     dispatch(setSupportedCounterValues(supportedCounterValues));
-    // Safety net: if the persisted counterValue is no longer supported by CVS, fall back to USD.
+    // Reset only when CVS actually returned its list. fiatsReady also flips true on a CVS
+    // error/timeout (setFiatsReady runs in a finally) and then `fiats` is only the offline
+    // fallback — resetting there would wipe a valid uncommon fiat (e.g. AMD) and persist USD.
+    // A genuinely delisted fiat is corrected on the next successful load.
+    if (!fiatsResolvedFromCvs) return;
     const currentId = counterValueIdOf(counterValueCurrency);
     const isSupported = supportedCounterValues.some(
       ({ currency }) => counterValueIdOf(currency) === currentId,
     );
     if (!isSupported) dispatch(setCountervalue("USD"));
-  }, [dispatch, fiats, fiatsReady, counterValueCurrency]);
+  }, [dispatch, fiats, fiatsReady, fiatsResolvedFromCvs, counterValueCurrency]);
```

apps/ledger-live-mobile/src/mvvm/contexts/__tests__/InitialQueriesContext.test.tsx
```diff
@@ -46,6 +54,39 @@ describe("InitialQueriesContext", () => {
...
+    const { store } = render(
+      <InitialQueriesProvider>
+        <ContextSpy />
+      </InitialQueriesProvider>,
+      { overrideInitialState: withCounterValue("AMD") },
+    );
+
+    await waitFor(() => expect(store.getState().settings.counterValue).toBe("USD"));
+  });
+
```
ysitbon, line 88: And a pessimist test for the uncovered case
```ts
it("keeps the persisted counterValue when the CVS supported-fiats query fails", async () => {
  ofacResponse.mockResolvedValueOnce(HttpResponse.json({}));
  server.use(http.get(SUPPORTED_FIATS_URL, () => HttpResponse.error()));
  const { store } = render(
    <InitialQueriesProvider><ContextSpy /></InitialQueriesProvider>,
    { overrideInitialState: withCounterValue("AMD") },
  );
  await waitFor(() =>
    expect(contextSpy).toHaveBeenLastCalledWith(expect.objectContaining({ fiatsReady: true })));
  expect(store.getState().settings.counterValue).toBe("AMD");
});
```

LucasWerey: done

apps/ledger-live-mobile/src/mvvm/contexts/InitialQueriesContext.tsx
LucasWerey: done

ysitbon (approved):

deepyjr (approved):

PR #20177, opened 2026-07-29
[LWDM] refactor(coin-near): migrate NearBlocks indexer client to v3 (LIVE-34441)

pawell24:
### 📝 Description

NearBlocks published a new API generation and asked us to move off `/v1`. This migrates the four
indexer call sites in `coin-near` to `/v3`. Client-only change: the host (`API_NEARBLOCKS_INDEXER`)
and the bearer auth are unchanged, and the NEAR JSON-RPC node path is untouched.

| Call | Before | After |
|---|---|---|
| Transaction history | `/v1/account/{a}/txns-only` | `/v3/accounts/{a}/txns` |
| Gas price | `/v1/stats` | `/v3/stats` |
| Delegated validators | `/v1/kitwallet/staking-deposits/{a}` | `/v3/kitwallet/staking-deposits/{a}` |
| Validator list | `/v1/validators?per_page=&page=` | `/v3/validators?limit=&next=` |

Response handling:

- Every enveloped route is unwrapped from `{ data, meta }`, and `data: null` is handled (v3 types it
  as nullable). `/v3/kitwallet/*` is the exception and stays a bare array.
- `included_in_block_hash` → `block.block_hash`; `block.block_height` is now a string and is
  converted; the transfer amount moves from `actions[0].deposit` to `actions_agg.deposit`.
- Validators become a flat snake_case shape (`account_id`, `current_epoch_stake`, `fee_numerator`,
  `fee_denominator`). A single page is capped at 100, so reaching the same 200 validators needs a
  cursor walk. The walk is bounded by the requested total rather than by the cursor, so an indexer
  that keeps advertising a next page cannot loop the client indefinitely.
- `gas_price` is nullable in v3, so `getGasPrice` falls back to the node's JSON-RPC `gas_price`
  method rather than failing.
- Every nested object in a transaction is now typed optional and read defensively. `getOperations`
  maps a page through `Promise.all`, so one malformed transaction would otherwise discard the whole
  account history. Live data confirms `outcomes: {}` and `actions: []` do occur.

**Precondition (verified).** The origin behind `near-indexer.coin.ledger.com` serves `/v3`:
`GET /v3/stats` → `200`, `{"data":{"gas_price":"100000000","nodes_online":413, …}}` (2026-07-29).
No client-side API key is needed; the proxy injects it.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34441](https://ledgerhq.atlassian.net/browse/LIVE-34441)
- **ADR** (if any): n/a

[LIVE-34441]: https://ledgerhq.atlassian.net/browse/LIVE-34441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Allure reports

- Desktop (nanoSP): https://ledger-live.allure.green.ledgerlabs.net/allure/reports/db5491a7-3870-481f-83aa-1e68a4683069/
- iOS: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/492d6827-264c-46cd-a9ff-9e1326e28530/
- Android: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/112de07c-4f16-4611-b498-a77213c5ef05/

### DIFF
--- a/.changeset/near-nearblocks-indexer-v3.md
+++ b/.changeset/near-nearblocks-indexer-v3.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-near": minor
+---
+
+Migrate the NearBlocks indexer client from v1 to v3
--- a/libs/coin-modules/coin-near/src/api/indexer.integ.test.ts
+++ b/libs/coin-modules/coin-near/src/api/indexer.integ.test.ts
@@ -1,0 +1,85 @@
+import { setCoinConfig } from "../config";
+import { getOperations } from "./indexer";
+import { getGasPrice, getStakingPositions, getValidators } from "./node";
+
+const ACCOUNT_WITH_HISTORY = "nearkat.near";
+const DELEGATOR = "81afe80a9d91c82f66122c35ef400da709bde01eada5aae8d7a63bbf68f42040";
+
+describe("NearBlocks indexer (integration)", () => {
+  beforeAll(() => {
+    setCoinConfig(() => ({
+      status: { type: "active" },
+      infra: {
+        API_NEAR_PRIVATE_NODE: "https://near.coin.ledger.com/node",
+        API_NEAR_PUBLIC_NODE: "https://rpc.mainnet.near.org",
+        API_NEAR_INDEXER: "https://near.coin.ledger.com/indexer",
+        API_NEARBLOCKS_INDEXER: "https://near-indexer.coin.ledger.com",
+      },
+    }));
+  });
+
+  it("maps a real account history into operations", async () => {
+    const operations = await getOperations("accountId", ACCOUNT_WITH_HISTORY);
+
+    expect(operations.length).toBeGreaterThan(0);
+    expect(operations.length).toBeLessThanOrEqual(25);
+
+    operations.forEach(operation => {
+      expect(operation.hash).toMatch(/\S/);
+      expect(operation.value.isNaN()).toBe(false);
+      expect(operation.value.isNegative()).toBe(false);
+      expect(operation.fee.isNaN()).toBe(false);
+      expect(typeof operation.blockHeight).toBe("number");
+      expect(Number.isFinite(operation.blockHeight as number)).toBe(true);
+      expect(operation.blockHash).toMatch(/\S/);
+      expect(operation.date.getTime()).toBeGreaterThan(new Date("2020-01-01").getTime());
+      expect(operation.date.getTime()).toBeLessThan(Date.now() + 60_000);
+    });
+  });
+
+  it("reads a usable gas price", async () => {
+    const gasPrice = await getGasPrice();
+
+    expect(gasPrice).toMatch(/^\d+$/);
+    expect(BigInt(gasPrice) > 0n).toBe(true);
+  });
+
+  it("walks the cursor to collect the requested number of distinct validators", async () => {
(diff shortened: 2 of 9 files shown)


cted-ledger:
### Request: run the NEAR E2E suite before merge

The unit + integ coverage here is solid, and `test-coin-modules-integ` did run on this PR with `FILTER: near` (`indexer.integ.test.ts` and `preload.integ.test.ts` passed against the real indexer). But **no NEAR E2E test has run on this PR**, so the consumer side of the migration is currently unverified.

**Why nothing ran:** the NEAR specs aren't tagged `@smoke` (the only `@smoke` in `earn.v2.spec.ts` is on the Ethereum LiveApp test), so they're outside the smoke subsets that gate PRs. The `Ubuntu Mock` job sets `MOCK`, which bypasses `coin-near` entirely. The full suite only runs via `schedule` or `workflow_dispatch` on `[Desktop] - E2E Only - Scheduled/Manual`, i.e. after merge on the nightly.

**Why it's worth a run:** the existing NEAR specs exercise three of the four migrated call sites through the UI, including the one carrying the most new logic:

| Spec | Migrated call site exercised |
| --- | --- |
| `e2e/desktop/tests/specs/delegate.spec.ts` (NEAR_1, NEAR_2 → `ledgerbyfigment.poolv1.near`) | `/v3/validators` — the cursor walk must actually return the validator for the picker to find it by name |
| `e2e/mobile/specs/delegate/delegateNEAR.spec.ts` | same, mobile |
| `e2e/desktop/tests/specs/earn.v2.spec.ts` — "Hot start & Position - NEAR" | `/v3/kitwallet/staking-deposits/{a}` (requires a live stake position on NEAR_1) |
| `e2e/mobile/specs/earn/earnV2_hotStart_NEAR_1.spec.ts` | same, mobile |

`/v3/stats` is covered indirectly via `preload()`. For reference, `ledgerbyfigment.poolv1.near` currently sits at rank 7 in both the v1 top-200 and the new v3 cursor walk, so the validator these specs target is comfortably inside the 200 window and not exposed to boundary drift.

**One gotcha on the inputs:** with the default `build_type: testing`, the workflow sets `E2E_GENERATED_USERDATA_DIR` and restores `app.json` from the S3 cache keyed `develop/e2e-userdata-desktop-*`. `liveDataCommand` then short-circuits on `applyGeneratedUserdata` and never calls `runCliLiveData`, so the accounts are hydrated from a snapshot generated on `develop` (v1 code) and the transaction-history mapping — `/v3/accounts/{a}/txns`, i.e. the `actions_agg.deposit`, `block.block_hash` and string `block_height` remap — is **not** exercised. Passing `build_type: js` blanks that variable and forces a real CLI sync against this branch.

Suggested dispatch (`enable_broadcast` left at its default `false`, so nothing is broadcast):

```bash
gh workflow run test-ui-e2e-only-desktop.yml --repo LedgerHQ/ledger-live \
  -f ref=refactor/LIVE-34441-nearblocks-v3 \
  -f test_filter=@family-near \
  -f build_type=js
```

Happy to treat this as non-blocking if the team would rather let the nightly catch it post-merge — flagging it mainly because a data-layer swap like this has its whole blast radius in the NEAR accounts/staking UI, which is exactly what these specs cover.

cted-ledger (approved):

PR #20181, opened 2026-07-29
[LWD] feat(desktop): add qr code scanner to send recipient field

YazhuEth:
### 📝 Description

The new send flow on desktop had no way to scan a recipient QR code. The legacy flow had one, and the new flow on mobile already has one, so desktop was the gap.

This adds it, and with it the behaviour asked for in the ticket: the recipient field shows a QR code icon while it is empty, and swaps it for the clear cross as soon as the user types.

**Approach**

Lumen's `AddressInput` already handles the icon/cross swap on its own: it renders the built-in QR icon only when the input is empty, and `BaseInput` replaces it with the clear button once there is content. Desktop simply never passed `onQrCodeClick`. So the acceptance criterion needed no new UI logic, and the actual work was the camera surface behind the icon.

Clicking the icon renders the camera feed inline in the dialog body, right under the address input, replacing the recipient step body. The header (title, account summary, close) stays as is. The panel closes on a second click of the icon, on the first keystroke, and on a successful scan.

- `context/RecipientScannerContext.tsx` (new): desktop-local scanner state, mounted in `SendFlowLayout` so both the header (which owns the input) and the step (which owns the body) can read it.
- `screens/Recipient/components/RecipientQrScanner.tsx` (new): thin wrapper over the existing legacy `QRCodeCameraPickerCanvas` (jsQR + `getUserMedia`), sized for the 400px dialog. No new scanner code and no new dependency.
- `useSendHeaderModel.ts`: `handleQrCodeClick` (toggle + tracking), `handleRecipientInputChange` (closes the panel on typing), `handleScanPicked` (`decodeURIScheme` then `recipientSearch.setValue(address)`), and `isScannerOpen` forced false outside the recipient step.
- `SendHeader.tsx` / `RecipientScreen.tsx`: pass `onQrCodeClick`, render the panel, and hide the step body while scanning.

**Trade-offs worth a reviewer's attention**

- A scanned URI contributes only its address. Amount and currency are discarded, matching mobile's `onScanned` path, so the normal address validation, ENS resolution and sanction checks run untouched. The legacy desktop flow also prefilled the amount; that was dropped deliberately for parity.
- This is not a new wizard step. `SEND_FLOW_STEP` lives in live-common and both apps key exhaustive `Record<SendFlowStep, …>` configs off it, so adding `SCAN_RECIPIENT` would have forced dead entries into mobile, which keeps using its native scanner screen. The panel is desktop-local UI state instead.
- Analytics reuse the legacy event names `Send Flow QR Code Opened` / `Closed` so existing dashboards keep working.

**Tests**

Covered by automatic tests: 5 unit cases on the view model (toggle plus tracking, typing closes, scan sets the decoded address, no scanner outside the recipient step) and 3 integration cases (scan fills the input, icon/cross swap, typing closes the panel). The full Send suite passes, 242 tests across 26 suites. jsdom has no camera, so the test harness stubs the canvas picker module rather than faking `getUserMedia` and `jsqr`.

Not covered by automatic tests, and worth checking manually: the real camera path, the camera framing against the Figma spec, and the `NoAccessToCamera` state when permission is denied.

**QA focus**

- Recipient field of the new send flow: QR icon when empty, clear cross when filled, icon back after clearing.
- Camera panel: opens under the input, hides the recent recipients body, dialog height animates, closes on second click, on typing, and on a successful scan.
- Scanning a plain address QR and a `ethereum:0x…` / `bitcoin:…` URI QR: only the address lands in the field, then validation and ENS behave as usual.
- Camera released when the panel closes (no OS camera indicator left on) and when the dialog is closed mid scan.
- Camera permission denied: the error renders inside the panel.
- Amount step: the read-only recipient field must show no QR icon.

| Empty field: QR icon | Scanner open | Field filled: clear cross |
| -------------------- | ------------ | ------------------------- |
|  | Screenshot 2026-07-30 at 11 08 12||

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34431](https://ledgerhq.atlassian.net/browse/LIVE-34431)
- **ADR** (if any): n/a

[LIVE-34431]: https://ledgerhq.atlassian.net/browse/LIVE-34431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/context/RecipientScannerContext.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/context/RecipientScannerContext.tsx
@@ -1,0 +1,45 @@
+import React, { createContext, useCallback, useContext, useMemo, useState } from "react";
+import type { ReactNode } from "react";
+
+/**
+ * RecipientScannerContext
+ *
+ * Desktop-only UI state for the recipient QR code scanner. The QR icon lives in the flow header
+ * (SendHeader) while the panel replaces the recipient step body, so both need to read the same flag.
+ */
+
+type RecipientScannerContextValue = Readonly<{
+  isScannerOpen: boolean;
+  closeScanner: () => void;
+  toggleScanner: () => void;
+}>;
+
+const RecipientScannerContext = createContext<RecipientScannerContextValue | null>(null);
+
+type RecipientScannerProviderProps = Readonly<{
+  children: ReactNode;
+}>;
+
+export function RecipientScannerProvider({ children }: RecipientScannerProviderProps) {
+  const [isScannerOpen, setIsScannerOpen] = useState(false);
+
+  const closeScanner = useCallback(() => setIsScannerOpen(false), []);
+  const toggleScanner = useCallback(() => setIsScannerOpen(previous => !previous), []);
+
+  const value = useMemo(
+    () => ({ isScannerOpen, closeScanner, toggleScanner }),
+    [isScannerOpen, closeScanner, toggleScanner],
+  );
+
+  return (
+    <RecipientScannerContext.Provider value={value}>{children}</RecipientScannerContext.Provider>
+  );
+}
+
+export function useRecipientScanner(): RecipientScannerContextValue {
+  const context = useContext(RecipientScannerContext);
+  if (!context) {
+    throw new Error("useRecipientScanner must be used within a RecipientScannerProvider");
+  }
+  return context;
+}
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/hooks/useSendHeaderModel.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/hooks/useSendHeaderModel.ts
@@ -1,4 +1,5 @@
 import { SendFlowStep, SEND_FLOW_STEP } from "@ledgerhq/live-common/flows/send/types";
+import { decodeURIScheme } from "@ledgerhq/live-common/currencies/index";
 import { t } from "i18next";
 import { useMemo, useCallback, useRef } from "react";
 import { useFlowWizard } from "../../FlowWizard/FlowWizardContext";
@@ -14,8 +15,9 @@ import {
 import { SendStepConfig } from "../types";
 import BigNumber from "bignumber.js";
 import { useMaybeAccountName } from "~/renderer/reducers/wallet";
-import { trackPage } from "~/renderer/analytics/segment";
+import { track, trackPage } from "~/renderer/analytics/segment";
 import { getSendFlowTrackingProperties } from "../utils/tracking";
+import { useRecipientScanner } from "../context/RecipientScannerContext";
 
 type UseSendHeaderModelParams = Readonly<{
   availableText: string;
@@ -26,6 +28,10 @@ type UseSendHeaderModelResult = Readonly<{
   descriptionText: string | undefined;
   handleBack: () => void;
   handleRecipientInputClick: () => void;
+  handleRecipientInputChange: (value: string) => void;
+  handleQrCodeClick: () => void;
+  handleScanPicked: (code: string) => void;
+  isScannerOpen: boolean;
   showBackButton: boolean;
   showRecipientInput: boolean;
   showMemoControls: boolean;
@@ -41,6 +47,7 @@ export function useSendHeaderModel({
   const wizard = useFlowWizard<SendFlowStep, SendFlowBusinessContext, SendStepConfig>();
   const { state, uiConfig, recipientSearch, isRecipientAddressComplete } = useSendFlowData();
   const { close, transaction } = useSendFlowActions();
+  const { isScannerOpen, closeScanner, toggleScanner } = useRecipientScanner();
 
   const currencyName = state.account.currency?.ticker ?? "";
   const accountName = useMaybeAccountName(state.account.account ?? undefined);
@@ -89,6 +96,8 @@ export function useSendHeaderModel({
   const descriptionText = showTitle && showAvailable && accountSummary ? accountSummary : "";
 
   const handleBack = useCallback(() => {
+    closeScanner();
+
     // Per-step state cleanup that runs regardless of whether navigation uses backTarget
     // or goToPreviousStep, so floating steps and regular steps are treated uniformly
     if (currentStep === SEND_FLOW_STEP.AMOUNT) {
@@ -117,7 +126,7 @@ export function useSendHeaderModel({
     } else {
       close();
     }
-  }, [backTarget, close, currentStep, navigation, resetViewState, transaction]);
+  }, [backTarget, close, closeScanner, currentStep, navigation, resetViewState, transaction]);
 
   const addressInputValue = useMemo(() => {
     if (isRecipientStep) return recipientSearch.value;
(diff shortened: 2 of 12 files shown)


apps/ledger-live-desktop/src/mvvm/features/Send/hooks/useSendHeaderModel.ts
```diff
@@ -136,6 +143,37 @@ export function useSendHeaderModel({
     handleBack();
   }, [handleBack, isAmountStep, recipientSearch, state.recipient]);
 
+  const showScanner = isScannerOpen && isRecipientStep;
+
+  useEffect(() => {
```
dilaouid: I think we can avoid an useEffect here

apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/components/RecipientQrScanner.tsx
```diff
@@ -0,0 +1,23 @@
...
+const CAMERA_HEIGHT = 260;
+const CAMERA_CENTER_SIZE = 180;
+
+type RecipientQrScannerProps = Readonly<{
+  onPick: (code: string) => void;
+}>;
+
+export function RecipientQrScanner({ onPick }: RecipientQrScannerProps) {
+  return (
+    <div className="flex justify-center px-16 pb-16" data-testid="send-recipient-qr-scanner">
```
dilaouid: The close button does nothing here and is not linked to any `onClick` element

apps/ledger-live-desktop/src/mvvm/features/Send/context/RecipientScannerContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+const RecipientScannerContext = createContext<RecipientScannerContextValue | null>(null);
+
+type RecipientScannerProviderProps = Readonly<{
+  children: ReactNode;
+}>;
+
+export function RecipientScannerProvider({ children }: RecipientScannerProviderProps) {
+  const [isScannerOpen, setIsScannerOpen] = useState(false);
+
+  const openScanner = useCallback(() => setIsScannerOpen(true), []);
```
dilaouid: I don't think this is useful ?

apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/components/RecipientQrScanner.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import QRCodeCameraPickerCanvas from "~/renderer/components/QRCodeCameraPickerCanvas";
```
dilaouid: I think this is mixing styled-component with Lumen, can we create a component for this instead of importing a legacy one?

apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/useQrCodeScanner.ts
```diff
@@ -0,0 +1,105 @@
...
+      if (cancelled) {
+        stream.getTracks().forEach(track => track.stop());
+        return;
+      }
+
+      video.srcObject = stream;
+      try {
+        await video.play();
+      } catch (e) {
+        logger.error(e);
```
dilaouid, line 102: Can we not throw a `NoAccessToCamera` here ? maybe so fail won't be hiden

apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/useQrCodeScanner.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    const canvas = document.createElement("canvas");
+    const context = canvas.getContext("2d");
+
+    const decodeFrame = () => {
+      if (!context || !video.videoWidth || !video.videoHeight) return;
+
+      canvas.width = video.videoWidth;
+      canvas.height = video.videoHeight;
+      context.drawImage(video, 0, 0, canvas.width, canvas.height);
```
dilaouid: Here we are using a lot of performance just to scan a qr code 😿
each `DECODE_INTERVAL_MS` (250ms) we scan one picture, so it means:, 4 times per second we:
- take the whole resolution of the cam (let's say it's a 4K camera the user is very rich)
- reassign `canvas.width/height`
- fill `getImageData` RGB**_A_** array
- `jsQR` will scan every pixel of this picture

In a 4K camera, one image is almost 30mo
With 4 scan/second, we go through 120mo/s of data to scan, without compting on the `jsQR` work

My suggestion will be to increase `DECODE INTERVAL MS` , always keeping the same ratio and resize the canva only on changes:

```ts
const MAX_DECODE_WIDTH = 640;

const decodeFrame = () => {
  if (
    !context ||
    video.readyState < HTMLMediaElement.HAVE_CURRENT_DATA ||
    !video.videoWidth ||
    !video.videoHeight
  ) {
    return;
  }

  const scale = Math.min(1, MAX_DECODE_WIDTH / video.videoWidth);
  const width = Math.round(video.videoWidth * scale);
  const height = Math.round(video.videoHeight * scale);

  if (canvas.width !== width || canvas.height !== height) {
    canvas.width = width;
    canvas.height = height;
  }

  context.drawImage(video, 0, 0, width, height);

  const image = context.getImageData(0, 0, width, height);
  const code = jsQR(image.data, width, height);

  if (code?.data) {
    onPickRef.current(code.data);
  }
};
```


NIT SUGGESTION; And above we can also expect a smaller resolution of the camera, we don't need to take the whole resolution of the native camera it's only for a qr code scanning, like for example:

```ts
getUserMedia({
  video: {
    facingMode: { ideal: "environment" },
    width: { ideal: 640 },
    height: { ideal: 480 },
  },
});
```

dilaouid (dismissed):

YazhuEth:
*E2E*
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/78a25772-af2d-4503-a47c-334a901c542a/
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/4fb51ef7-2c74-4069-aded-5bd98aa53741/

apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/useQrCodeScanner.ts
```diff
@@ -0,0 +1,126 @@
+import { useCallback, useEffect, useRef, useState } from "react";
+import jsQR from "jsqr";
+import { NoAccessToCamera } from "@ledgerhq/errors";
```
gre-ledger: `@ledgerhq/errors` is being sunset — `NoAccessToCamera` now lives in `@ledgerhq/live-common/errors`:

```suggestion
import { NoAccessToCamera } from "@ledgerhq/live-common/errors";
```

(on behalf of @gre-ledger)

apps/ledger-live-desktop/src/mvvm/features/Send/screens/Recipient/hooks/__tests__/useQrCodeScanner.test.tsx
```diff
@@ -0,0 +1,223 @@
+import React from "react";
+import { createRoot } from "react-dom/client";
+import { act } from "tests/testSetup";
+import { NoAccessToCamera } from "@ledgerhq/errors";
```
gre-ledger: `@ledgerhq/errors` is being sunset — `NoAccessToCamera` now lives in `@ledgerhq/live-common/errors`:

```suggestion
import { NoAccessToCamera } from "@ledgerhq/live-common/errors";
```

(on behalf of @gre-ledger)

gre-ledger (commented):
On behalf of @gre-ledger: `@ledgerhq/errors` is being sunset (LIVE-35072, see #20040) and should disappear soon — see the 2 inline notes.

Nothing blocking today. The desktop replacement for `NoAccessToCamera` is `@ledgerhq/live-common/errors`. Let me know if this does not apply.

gre-ledger:
(i'm still WIP on this skill 😆  https://github.com/LedgerHQ/ledger-live/pull/20161#issuecomment-5131255221 )

gre-ledger (approved):

dilaouid (approved):

PR #20189, opened 2026-07-29
[LWDM] feat(send): rework network fee display to include crypto amounts (LIVE-34432)

YazhuEth:
### ✅ Checklist

- [x] **Covered by automatic tests.** _Unit tests cover the shared formatter, the core hook (all three display modes plus the zero-fee and error guards), the per-preset values hook, and both apps' view models. The mobile row component is tested directly; the desktop row markup is covered by `SendFlow.integration.test.tsx`. The two-tone styling itself needs a visual check._
  - **ETH / EVM**: the fee row must follow the amount input's fiat ⇄ crypto toggle, and the fee selector presets must show both amounts.
  - **SOL, XRP**: fees are not editable, so the row must show the fiat value followed by the native amount in the muted colour, with no `•` separator and no strategy label. The row must not be clickable.
  - **TRX**: this is a behaviour change, not a regression. Tron previously showed `$0.10 • 0.00056 TRX` via a coin specific flag. It now goes through the generic non editable rule, so the `•` disappears and the TRX amount becomes muted. Please confirm the Tron fee explanation drawer still works.
  - **BTC / Kaspa**: presets must still show the fee rate legend (`12 sat/vB`), not the amounts. Coin control must stay reachable, and the coin control footer fee row must still render.
  - Zero and unknown fees: a covered fee should read `0 USD 0 TRX`, while an unknown fee (estimation skipped because the transaction has errors) must still read `--` rather than a confirmed zero.

### 📝 Description

**Problem.** In the new send flow, the network fee row showed a fiat only value, falling back to the native amount only when no countervalue rate existed. Fee preset sublabels were fiat only too. Users signing a transaction could not see what they would actually pay in the network's own currency, which matters most for the coins whose fee they cannot adjust at all.

**Solution.** The fee display now depends on whether the user can change the fee:

- **Fee editable** (presets, custom fees, or coin control available): the row shows a single value that follows the amount input's fiat ⇄ crypto toggle. `$0.10 • Fast` in fiat mode, `0.032938 ETH • Fast` in crypto mode.
- **Fee not editable**: the row always shows both values, fiat in the base colour followed by the native amount in the muted colour, and no strategy label. The row is the only place that fee is visible, so it should not hide either half.
- **Fee selector**: each preset sublabel shows both amounts, `$0.03 · 0.000329 ETH`. Coins that price fees by rate (Bitcoin, Kaspa) keep their `12 sat/vB` legend instead, since the rate is what users compare presets by there.

Almost all of the work lives in the shared core so both apps change once:

- `flows/send/utils/networkFeesDisplay.ts`: the two formatters collapse into one `formatFeesValue({ mode })` with modes `"fiat" | "crypto" | "both"`, returning a main value plus an optional secondary native amount (split rather than pre joined, so the two halves can be styled differently).
- `flows/send/hooks/useFeePresetValuesCore.ts` (renamed from `useFeePresetFiatValuesCore`): each preset now carries `{ fiat, crypto }`. The bridge estimation path already computed the native amount per preset and discarded it.
- `flows/send/utils/feeSelectorOptions.ts`: `FeeStrategyOption` gains `sublabelCrypto`, and a new `feeStrategySublabel()` resolves legend versus amounts in one place shared by both apps.
- `flows/send/hooks/useNetworkFeesCore.ts`: takes the display mode as a parameter (injected like `fiatUnit` and `locale` already are, rather than reading context, so the hook stays testable without mounting a provider) and derives editability from the descriptor driven UI config.

The per coin `FeeDescriptor.showFeeCurrencyAmount` flag, which only Tron used, is **deleted**. Tron's fees are not editable, so it now gets the two value display from the generic rule. This removes a coin specific flag in favour of the underlying mechanism, per the coin families contract.

The desktop change threads one extra prop (`feesRowSecondaryValue`) through the existing flat prop chain (`AmountFooter` → `AmountScreenView` → `CoinControlFooter`), which accounts for most of the remaining touched files.

### 📸 Screenshots

**Fee editable (ETH): the row follows the amount input's fiat ⇄ crypto toggle**

| Fiat mode | Crypto mode |
| --------- | ----------- |
|Screenshot 2026-07-29 at 16 55 35| Screenshot 2026-07-29 at 16 55 14|

**Fee selector, and a coin whose fee is not editable**

| Fee selector: both amounts per preset (ETH) | Fee not editable: both values (SOL) |
| ------------------------------------------- | ----------------------------------- |
| Screenshot 2026-07-29 at 16 55 24|Screenshot 2026-07-29 at 16 56 46|

### ❓ Context

- **JIRA or GitHub link**: [LIVE-34432](https://ledgerhq.atlassian.net/browse/LIVE-34432)

---

[LIVE-34432]: https://ledgerhq.atlassian.net/browse/LIVE-34432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/network-fee-display-crypto-and-fiat.md
+++ b/.changeset/network-fee-display-crypto-and-fiat.md
@@ -1,0 +1,7 @@
+---
+"@ledgerhq/live-common": patch
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
+Show the network fee in the currency users care about in the new send flow. When the fee is editable the row follows the amount input's fiat/crypto toggle; when it is not, the row shows the fiat value alongside the native amount, since it is the only place that fee is visible. Fee presets now sub-label both amounts, except coins priced by fee rate (Bitcoin, Kaspa) which keep their sat/vB legend.
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/hooks/__tests__/useNetworkFees.test.ts
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/hooks/__tests__/useNetworkFees.test.ts
@@ -26,10 +26,19 @@ jest.mock("@ledgerhq/live-common/bridge/descriptor/send/features", () => ({
     canEstimateFeePresetsWithZeroAmount: jest.fn(() => false),
     getFeePresetOptions: jest.fn(() => []),
     getFeeCurrencyAccountId: jest.fn(() => null),
-    showFeeCurrencyAmount: jest.fn(() => false),
-    getDefaultStrategyPatch: jest.fn(() => ({ feesStrategy: undefined, fees: undefined })),
+    getDefaultStrategyPatch: jest.fn(() => ({
+      feesStrategy: undefined,
+      fees: undefined,
+    })),
     hasFeeRateLegend: jest.fn(() => false),
   },
+}));
+let mockDisplayMode: "fiat" | "crypto" = "fiat";
+jest.mock("@ledgerhq/live-common/flows/send/amount/SendAmountDisplayModeContext", () => ({
+  useSendAmountDisplayMode: () => ({
+    displayMode: mockDisplayMode,
+    setDisplayMode: jest.fn(),
+  }),
 }));
 jest.mock("@ledgerhq/live-countervalues-react", () => ({
   ...jest.requireActual("@ledgerhq/live-countervalues-react"),
@@ -51,6 +60,7 @@ function isAccount(account: unknown): account is Account {
 
 function buildBaseParams(overrides?: {
   transaction?: Partial<Transaction>;
+  status?: Partial<TransactionStatus>;
   uiConfig?: {
     hasFeePresets?: boolean;
     hasCustomFees?: boolean;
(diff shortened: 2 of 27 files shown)


YazhuEth:
**E2E**
- https://github.com/LedgerHQ/ledger-live/actions/runs/30463990309
- https://github.com/LedgerHQ/ledger-live/actions/runs/30463999306

YazhuEth:
E2E
- https://github.com/LedgerHQ/ledger-live/actions/runs/30541129347
- https://github.com/LedgerHQ/ledger-live/actions/runs/30541143120

YazhuEth:
*E2E*
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/978ab7d1-472e-4661-8db4-20edb08f11b9/
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/1e782988-3aa5-4019-b6c4-7c3bcf51b4b4/

dilaouid:
Should we keep the Network fee in [FIAT] if we switched the mode to crypto? 🤔

image

dilaouid (approved):

PR #20190, opened 2026-07-29
[LWDM] fix(lwdm): new send flow wait for valid address to display memo

dilaouid:
### 📝 Description

Previously, the new Send flow showed the memo field as soon as the recipient address looked complete, even before bridge validation finished or confirmed the address was actually valid. This code introduces `isAddressValid`, which requires a successful bridge validation result with no recipient error, and uses it (instead of `isAddressComplete`) to control when the memo UI appears on lwdm

https://github.com/user-attachments/assets/f2b02506-fcff-40d5-93d2-5782e3415e11

### 🔗 Context

- **[JIRA / GitHub issue](https://ledgerhq.atlassian.net/browse/LIVE-34975)**: 
- **ADR** (if any):

### DIFF
--- a/.changeset/wise-apricots-compete.md
+++ b/.changeset/wise-apricots-compete.md
@@ -1,0 +1,7 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
+"@ledgerhq/live-common": minor
+---
+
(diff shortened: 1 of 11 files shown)


qperrot (approved):
LGTM

PR #20192, opened 2026-07-29
test(e2e-desktop): soft assert with screenshots

martijnhjk:
### 📝 Description

Introduces soft assert support with screenshot on each soft assert failure.

[Regression](https://github.com/LedgerHQ/ledger-live/actions/runs/30540004490) - language change with soft asserts and add accounts as benchmark.

[Failing Regression](https://github.com/LedgerHQ/ledger-live/actions/runs/30540384304) - intentional failures to show reporting example.

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/QAA-1061

### DIFF
--- a/e2e/desktop/tests/utils/softExpect.ts
+++ b/e2e/desktop/tests/utils/softExpect.ts
@@ -1,0 +1,77 @@
+import { expect, test } from "@playwright/test";
+
+import type { Expect, Page } from "@playwright/test";
+
+export type SoftExpectMatcher = Expect["soft"];
+
+export type SoftExpectOptions = {
+  timeout?: number;
+};
+
+export type SoftExpectParams = {
+  page: Page;
+  assertion: (soft: SoftExpectMatcher) => Promise<void>;
+};
+
+/**
+ * Runs a soft assertion and attaches a screenshot when it fails.
+ *
+ * Playwright soft failures are recorded without stopping the test and reported at the end.
+ * This helper attaches a screenshot as soon as the soft assertion fails.
+ *
+ * The callback receives a configured soft expect instance with a dedicated timeout to avoid bloating overall test runtime.
+ * The function uses the injected matcher — not global `expect.soft` — so that timeout applies.
+ *
+ * @param params.page - Main window to screenshot on failure.
+ * @param params.assertion - Callback using the injected soft expect matcher (`soft`).
+ * @param options.timeout - Optional soft assertion timeout in ms.
+ *
+ * @example
+ * // In page objects extending `PageHolder`:
+ * await this.softExpect(async soft => {
+ *   await soft(this.header).toContainText("My Header");
+ * });
+ *
+ * @example
+ * // In page objects extending `PageHolder` with a custom timeout:
+ * await this.softExpect(async soft => {
+ *   await soft(this.header).toContainText("My Header");
+ * }, { timeout: 10_000 });
+ *
+ * @example
+ * // Direct import (two-arg: params, options):
+ * await softExpect(
+ *   {
+ *     page,
+ *     assertion: async soft => {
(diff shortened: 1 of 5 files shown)


e2e/desktop/tests/specs/benchmark.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  @step("Soft fail")
+  async softExpectFail({ page }: { page: Page }) {
+    await this.softExpect(page).toHaveTitle("failmesoft");
+  }
+}
+
+const myPage = new MyPage();
+
+test.describe("Hard assert", () => {
```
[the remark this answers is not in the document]
martijnhjk: outdated, not relevant

e2e/desktop/tests/utils/softExpect.ts
```diff
@@ -0,0 +1,70 @@
...
+  const softExpectInstance = expect.configure({ soft: true, timeout });
+  await assertion(softExpectInstance);
+
+  if (testInfo.errors.length > errorCountBefore) {
+    const screenshot = await page.screenshot();
+    await testInfo.attach("Soft failure screenshot", {
+      body: screenshot,
+      contentType: "image/png",
+    });
+  }
```
cunhabruno: Two things in this block, both fixable in one edit.

**1. `page.screenshot()` is unguarded.** If it throws (window closed, renderer hung, screenshot timeout), `softExpect` throws and the test aborts at the assertion specifically designed *not* to abort it — the remaining soft checks never run. Not far-fetched, since we only reach this line after a 5s matcher timeout, which correlates with a stuck renderer.

**2. Every attachment shares one name.** The language-change spec runs 5 soft checks per test (`settings.spec.ts:263-267`), so a translation regression produces 5 attachments all called `Soft failure screenshot`, with nothing tying each one to the assertion that failed.

To be fair, the context isn't *entirely* lost: `testInfo.attach()` associates the attachment with the enclosing step (`this._attach(a, this._parentStep()?.stepId)` in `worker/testInfo`), and `allure-playwright` preserves that mapping via `attachmentTargetByStep`. `detail: false` doesn't break it either — that filter only drops step categories outside `["test.step", "attach", "test.attach"]`, and the `@step` decorator produces `test.step`. So in the Allure tree each PNG already sits under a uniquely named node like *"Expect theme row title to be Mode"*. The ambiguity is confined to the flat test-level attachment list, downloaded filenames, and CI's `github`/`list` reporters.

Indexing by position in `testInfo.errors` closes that gap deterministically — `errorCountBefore` already *is* that index, so `#3` is the 3rd entry in the error list in every reporter:

```suggestion
  if (testInfo.errors.length === errorCountBefore) return;

  try {
    const screenshot = await page.screenshot();
    await testInfo.attach(`Soft failure #${errorCountBefore + 1} screenshot`, {
      body: screenshot,
      contentType: "image/png",
    });
  } catch (error) {
    console.warn(`[softExpect] could not attach screenshot: ${error}`);
  }
```

I'd avoid deriving the name from the error message: in Playwright 1.60 `expectation` is the literal placeholder `"expected"`, so line 1 is byte-identical across all five failures (`expect(locator).toContainText(expected) failed`); the `Locator:` line also collides between `expectCounterValueRowCharacterSet` and `expectCounterValueRowTranslation`, which both target `setting-countervalue-dropDown`; and the message carries ANSI escapes.

One edge case: if a single callback contains several soft assertions and two fail, `errors` grows by 2 but you get one screenshot labelled `#N`. Probably fine given the docs steer toward one assertion per callback — otherwise `` `#${errorCountBefore + 1}–${testInfo.errors.length}` ``.

martijnhjk: pushed both suggestions thank you 🌮 !

cunhabruno (approved):

PR #20198, opened 2026-07-29
[LWDM] fix(errors): resolve esm import crash from extensionless specifiers

### LINKED ISSUE #15967: [Bug]: `@ledgerhq/errors` crashes when imported from an ESM module
### Impacted Library name

@ledgerhq/errors

### Impacted Library version

11.11.0

### Describe the bug

Importing this library from an ESM module leads to import errors that crash your application. I believe it's due to a misconfiguration of typescript or your build system.

### Expected behavior

It shouldn't throw on imports.

### Additional context

Here's a minimal reproduction: https://github.com/alcuadrado/ledgerhq-errors-repro

reservedbytes:
`@ledgerhq/errors` cannot be imported from an esm module, it crashes with `ERR_MODULE_NOT_FOUND`

the esm build uses `--moduleResolution bundler`, which emits relative specifiers exactly as written,
so `lib-es/index.js` ships `from "./helpers"` and node esm cannot resolve it

this adds the `.js` extension to the two specifiers in this package,
cjs is unaffected since `require("./helpers.js")` resolves the same either way,
and the jest configs get the usual `moduleNameMapper` so they still land on `helpers.ts`

checked `@ledgerhq/errors@6.37.0` from npm, not just the 6.32.0 the reporter's lockfile pins,
and after the change an esm consumer imports it while a `require()` consumer still works

Closes #15967

### DIFF
--- a/.changeset/errors-esm-import-extensions.md
+++ b/.changeset/errors-esm-import-extensions.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/errors": patch
+---
+
+fix esm import crash caused by relative specifiers missing their file extension
--- a/libs/ledgerjs/packages/errors/jest.config.ts
+++ b/libs/ledgerjs/packages/errors/jest.config.ts
@@ -7,4 +7,7 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 export default {
   ...baseConfig,
   rootDir: __dirname,
+  moduleNameMapper: {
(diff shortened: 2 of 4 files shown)


gre-ledger (changes requested):
@ledgerhq/errors is deprecated and no longer maintained, in fact we will be sunset the library very soon (drop it from the project). so i'm sorry but I don't think we will take this fix, you can instead stop depending on it (see README on the library on how to not use the library)

PR #20197, opened 2026-07-29
[LWDM] refactor(cvs): drop boot fiat fetch and InitialQueriesContext

ysitbon:
## Summary

- **Root-cause fix for [LIVE-35110](https://ledgerhq.atlassian.net/browse/LIVE-35110)** (CVS counter-value reset bug). Supersedes hotfix PR #20164.
- Deletes `InitialQueriesContext`/`InitialQueriesProvider` from mobile entirely. Rewires consumers to direct selectors and RTK Query hooks.
- `supportedCounterValues` is no longer a persisted Redux field — it's a derived selector `createSelector(selectSupportedFiats, buildSupportedCounterValues)` backed by the `supportedFiats` slice, so boot-time CVS resets are impossible.
- Adds a read-time OFAC guard in `counterValueCurrencyLocalSelector` so compliance is preserved without the deleted boot reset.
- Both apps now lazy-load `useGetSupportedFiatsQuery()` from the two picker screens that need it.

## Changes

**Mobile (source)**
- Delete `InitialQueriesContext.tsx` / `InitialQueriesProvider`
- `WaitForAppReady` → `selectRemoteFlagsReady` + `ofacGeoBlockApi.useCheckQuery()`
- `AppGeoBlocker` → direct `ofacGeoBlockApi.useCheckQuery()`
- `LedgerStore.tsx` → remove `updateSupportedCountervalues` call and `InitialQueriesProvider` wrapper
- `reducers/settings.ts` → derived `supportedCounterValuesSelector`, OFAC guard
- `reducers/types.ts` → remove `supportedCounterValues` from `SettingsState`
- `actions/settings.ts` + `actions/types.ts` → remove `setSupportedCounterValues` / `SET_SUPPORTED_COUNTER_VALUES`
- `CountervalueSettings.tsx` + `useMarketCurrencySelectViewModel.ts` → lazy `useGetSupportedFiatsQuery()`

**Desktop (source)**
- `renderer/init.tsx` → remove boot-time `getSupportedFiats.initiate` dispatch
- `CounterValueSelect.tsx` + `CountervalueSelect.tsx` → lazy `useGetSupportedFiatsQuery()`

**Tests**
- `WaitForAppReady.test.tsx` — full rewrite off the deleted context (4 cases)
- `AppGeoBlocker.test.tsx` — rewrite using `ofacGeoBlockApi` mock (3 cases)
- `changeCurrency.integration.test.tsx` — seeds `supportedFiats` slice directly with real `FiatCurrency` objects
[...]

PR #20216, opened 2026-07-30
[LWDM] feat(zcash): ironwood support on lw [LIVE-32890]

semeano:
### 📝 Description

Ironwood pool in the Ledger Live Desktop shielded UI: surface the Ironwood shielded balance and target the Ironwood pool in the send modal.

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/LIVE-32890

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountBalanceSummaryFooter.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountBalanceSummaryFooter.tsx
@@ -19,11 +19,15 @@ import { openModal } from "~/renderer/actions/modals";
 import type { Currency } from "@ledgerhq/wallet-btc/index";
 import type { ZcashAccount } from "@ledgerhq/live-common/families/bitcoin/types";
 import type { TokenAccount } from "@ledgerhq/types-live";
-import type { ZcashSyncState } from "@ledgerhq/coin-bitcoin/chain-adapters/zcash/types";
+import type { ZcashSyncState } from "@ledgerhq/coin-zcash/network/types";
 import {
   ZCASH_CHECK_OUTDATED_SYNC_INTERVAL,
   ZCASH_OUTDATED_SYNC_INTERVAL_MINUTES,
-} from "@ledgerhq/coin-bitcoin/chain-adapters/zcash/constants";
+} from "@ledgerhq/coin-zcash/constants";
+import {
+  getPrivateBalance,
+  getTransparentBalance,
+} from "@ledgerhq/coin-zcash/logic/account/balance";
 import { selectShieldedSubscriptions } from "~/renderer/reducers/shieldedSyncSubscriptions";
 import { useZcashShieldedSync } from "./useZcashShieldedSync";
 
@@ -201,10 +205,6 @@ const AccountBalanceSummaryFooter = ({ account }: Props) => {
   const showPrivateBalanceComponent = useFeature("zcashShielded")?.enabled;
 
   const privateInfo = "privateInfo" in account ? account.privateInfo : null;
-  const { orchardBalance, saplingBalance } = privateInfo ?? {
-    orchardBalance: BigNumber(0),
-    saplingBalance: BigNumber(0),
-  };
   const syncState = privateInfo?.syncState ?? "disabled";
   const previousSyncState = usePrevious(syncState);
   const lastSync = privateInfo?.lastSyncTimestamp ? new Date(privateInfo.lastSyncTimestamp) : null;
@@ -308,9 +308,14 @@ const AccountBalanceSummaryFooter = ({ account }: Props) => {
     locale,
   };
 
+  // Each label is derived from its own source rather than one from another, so
+  // they stay correct regardless of which module (coin-bitcoin flag-off adapter
+  // vs coin-zcash) last wrote `account.balance` — a toggle or a pre-first-sync
+  // flag-ON no longer yields an under-reported or negative transparent balance.
+  const bitcoinResources = "bitcoinResources" in account ? account.bitcoinResources : undefined;
+  const _transparentBalance = getTransparentBalance(bitcoinResources?.utxos);
+  const _privateBalance = getPrivateBalance(privateInfo);
   const _availableBalance = balance ?? BigNumber(0);
-  const _privateBalance = orchardBalance.plus(saplingBalance);
-  const _transparentBalance = _availableBalance.minus(_privateBalance);
 
   const transparentBalanceLabel = formatCurrencyUnit(unit, _transparentBalance, formatConfig);
   const privateBalanceLabel = formatCurrencyUnit(unit, _privateBalance, formatConfig);
--- a/apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashTransferFromSelector.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashTransferFromSelector.tsx
@@ -1,13 +1,16 @@
 import React, { useEffect } from "react";
 import styled from "styled-components";
 import { Trans } from "react-i18next";
-import BigNumber from "bignumber.js";
 import { useFeature } from "@features/platform-feature-flags";
 import { useAccountBridge } from "@ledgerhq/live-common/bridge/useAccountBridge";
 import { formatCurrencyUnit } from "@ledgerhq/live-common/currencies/index";
 import type { Account } from "@ledgerhq/types-live";
 import type { Transaction, ZcashAccount } from "@ledgerhq/live-common/families/bitcoin/types";
-import type { ZcashTransaction } from "@ledgerhq/coin-bitcoin/chain-adapters/zcash/types";
+import type { Transaction as ZcashTransaction } from "@ledgerhq/coin-zcash/types";
+import {
+  getPrivateBalance,
+  getTransparentBalance,
+} from "@ledgerhq/coin-zcash/logic/account/balance";
 import { useSelector } from "LLD/hooks/redux";
 import { localeSelector } from "~/renderer/reducers/settings";
 import Box from "~/renderer/components/Box/Box";
@@ -82,26 +85,26 @@ const ZcashTransferFromSelector = ({ account, transaction, onChange }: Props) =>
   const locale = useSelector(localeSelector);
   const unit = useAccountUnit(account);
 
-  const tx = transaction as ZcashTransaction;
+  const tx = transaction as unknown as ZcashTransaction;
 
   useEffect(() => {
     if (!active) return;
     if (tx.sender !== undefined) return;
-    onChange(bridge.updateTransaction(tx, { sender: "public" }) as Transaction);
+    onChange(bridge.updateTransaction(tx, { sender: "public" }) as unknown as Transaction);
   }, [active, tx.sender, bridge, onChange, tx]);
 
   if (!active) return null;
 
-  const sender: Sender = tx.sender ?? "public";
+  const sender: Sender = tx.sender === "private" ? "private" : "public";
 
-  const privateInfo = (account as ZcashAccount).privateInfo;
+  const zcashAccount = account as ZcashAccount;
+  const privateInfo = zcashAccount.privateInfo;
   const fvkAvailable = Boolean(privateInfo?.ufvk);
-  const orchardBalance = privateInfo?.orchardBalance ?? BigNumber(0);
-  const saplingBalance = privateInfo?.saplingBalance ?? BigNumber(0);
-
-  const totalBalance = account.balance ?? BigNumber(0);
-  const privateBalance = orchardBalance.plus(saplingBalance);
-  const transparentBalance = totalBalance.minus(privateBalance);
+  // Same helpers as AccountBalanceSummaryFooter: private = Ironwood only
+  // (deprecated Orchard/Sapling send flows are gone), transparent = own UTXOs
+  // (not balance − private, which diverges across flag/module provenance).
+  const privateBalance = getPrivateBalance(privateInfo);
+  const transparentBalance = getTransparentBalance(zcashAccount.bitcoinResources?.utxos);
 
   const formatConfig = { showCode: true, discreet, locale };
   const transparentLabel = formatCurrencyUnit(unit, transparentBalance, formatConfig);
@@ -109,7 +112,7 @@ const ZcashTransferFromSelector = ({ account, transaction, onChange }: Props) =>
 
   const select = (next: Sender) => {
     if (next === sender) return;
-    onChange(bridge.updateTransaction(tx, { sender: next }) as Transaction);
+    onChange(bridge.updateTransaction(tx, { sender: next }) as unknown as Transaction);
   };
 
   return (
--- a/apps/ledger-live-desktop/src/renderer/families/bitcoin/__tests__/ZcashTransferFromSelector.test.tsx
+++ b/apps/ledger-live-desktop/src/renderer/families/bitcoin/__tests__/ZcashTransferFromSelector.test.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import BigNumber from "bignumber.js";
-import { DEFAULT_ZCASH_PRIVATE_INFO } from "@ledgerhq/coin-bitcoin/chain-adapters/zcash/constants";
+import { DEFAULT_ZCASH_PRIVATE_INFO } from "@ledgerhq/coin-zcash/constants";
 import { render, screen, fireEvent, withFlagOverrides } from "tests/testSetup";
 import { createFixtureAccount } from "@ledgerhq/coin-bitcoin/fixtures/common.fixtures";
 import { CryptoCurrency } from "@domain/entity-currency-crypto";
@@ -21,6 +21,16 @@ jest.mock("@ledgerhq/live-common/bridge/useAccountBridge", () => ({
 }));
 
 const baseAccount = createFixtureAccount();
+
+const makeUtxo = (value: number) => ({
+  hash: "",
+  outputIndex: 0,
+  blockHeight: 1,
+  address: "",
+  value: new BigNumber(value),
+  rbf: false,
+  isChange: false,
+});
 
 const buildAccount = (overrides: Partial<typeof DEFAULT_ZCASH_PRIVATE_INFO> = {}, isZcash = true) =>
   ({
@@ -58,14 +68,14 @@ describe("ZcashTransferFromSelector", () => {
     mockedUseAccountUnit.mockReturnValue({ code: "ZEC", name: "Zcash", magnitude: 8 });
   });
 
-  it("renders both cards with Public active by default and persists the default to the transaction", () => {
+  it("renders Public and Private cards with Public active by default and persists the default to the transaction", () => {
     const onChange = renderSelector(buildAccount(), {});
 
     expect(screen.getByTestId("zcash-transfer-from-selector")).toBeVisible();
     expect(screen.getByTestId("transfer-from-public")).toBeVisible();
     expect(screen.getByTestId("transfer-from-private")).toBeVisible();
+    expect(screen.queryByTestId("transfer-from-ironwood")).not.toBeInTheDocument();
 
-    // Public default is written to the transaction state once on mount.
     expect(onChange).toHaveBeenCalledTimes(1);
     expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ sender: "public" }));
   });
@@ -77,7 +87,7 @@ describe("ZcashTransferFromSelector", () => {
 
   it("selects Private when the Private card is clicked", () => {
     const onChange = renderSelector(buildAccount({ ufvk: "uview-test" }), {});
-    onChange.mockClear(); // ignore the on-mount default write
+    onChange.mockClear();
 
     fireEvent.click(screen.getByTestId("transfer-from-private"));
 
@@ -94,27 +104,86 @@ describe("ZcashTransferFromSelector", () => {
     expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ sender: "public" }));
   });
 
-  it("keeps the Private card clickable in every sync state (never disabled)", () => {
-    const onChange = renderSelector(buildAccount({ syncState: "disabled", ufvk: "uview-test" }), {
-      sender: "public",
-    } as Partial<Transaction>);
-    onChange.mockClear();
-
-    fireEvent.click(screen.getByTestId("transfer-from-private"));
-
-    expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ sender: "private" }));
-  });
-
-  it("shows the unavailable hint on the Private card when no FVK is available", () => {
+  it("disables the Private card and shows why when no FVK is available", () => {
     renderSelector(buildAccount({ ufvk: null }), { sender: "public" } as Partial<Transaction>);
+    expect(screen.getByTestId("transfer-from-private")).toBeDisabled();
     expect(screen.getByTestId("transfer-from-private-unavailable")).toBeInTheDocument();
   });
 
-  it("hides the unavailable hint when an FVK is available", () => {
+  it("enables the Private card and hides the unavailable hint when an FVK is available", () => {
     renderSelector(buildAccount({ ufvk: "uview-test" }), {
       sender: "public",
     } as Partial<Transaction>);
+    expect(screen.getByTestId("transfer-from-private")).not.toBeDisabled();
     expect(screen.queryByTestId("transfer-from-private-unavailable")).not.toBeInTheDocument();
+  });
+
+  it("shows the spendable Ironwood balance on the Private card", () => {
+    renderSelector(
+      buildAccount({
+        ironwoodBalance: new BigNumber(50_000_000),
+        orchardBalance: new BigNumber(0),
+        saplingBalance: new BigNumber(0),
+        ufvk: "uview-test",
+      }),
+      {},
+    );
+
+    expect(screen.getByTestId("transfer-from-private")).toHaveTextContent(/0\.5 ZEC/);
+  });
+
+  it("shows 0 ZEC on the Private card when the Ironwood balance is 0", () => {
+    renderSelector(
+      buildAccount({
+        ironwoodBalance: new BigNumber(0),
+        orchardBalance: new BigNumber(50_000_000),
+        saplingBalance: new BigNumber(0),
+        ufvk: "uview-test",
+      }),
+      {},
+    );
+
+    expect(screen.getByTestId("transfer-from-private")).toHaveTextContent(/0 ZEC/);
+  });
+
+  it("sources the Public balance from UTXOs via getTransparentBalance, ignoring deprecated pools", () => {
+    renderSelector(
+      {
+        ...buildAccount({
+          orchardBalance: new BigNumber(30_000_000),
+          saplingBalance: new BigNumber(20_000_000),
+          ironwoodBalance: new BigNumber(10_000_000),
+          ufvk: "uview-test",
+        }),
+        // Transparent comes from own UTXOs (0.4 ZEC), not balance − ironwood —
+        // same helpers as AccountBalanceSummaryFooter.
+        bitcoinResources: { utxos: [makeUtxo(40_000_000)] },
+        balance: new BigNumber(50_000_000),
+      } as Account,
+      {},
+    );
+
+    expect(screen.getByTestId("transfer-from-public")).toHaveTextContent(/0\.4 ZEC/);
+  });
+
+  it("does not derive the Public balance by subtracting private from account.balance", () => {
+    renderSelector(
+      {
+        ...buildAccount({
+          ironwoodBalance: new BigNumber(50_000_000),
+          ufvk: "uview-test",
+        }),
+        // Stale provenance: balance − ironwood would wrongly yield 0.4 ZEC;
+        // UTXOs keep Public correct at 0.1 ZEC.
+        balance: new BigNumber(90_000_000),
+        bitcoinResources: { utxos: [makeUtxo(10_000_000)] },
+      } as Account,
+      {},
+    );
+
+    expect(screen.getByTestId("transfer-from-public")).toHaveTextContent(/0\.1 ZEC/);
+    expect(screen.getByTestId("transfer-from-public")).not.toHaveTextContent(/0\.4 ZEC/);
+    expect(screen.getByTestId("transfer-from-private")).toHaveTextContent(/0\.5 ZEC/);
   });
 
   it("renders nothing when the zcashShielded feature flag is off", () => {
--- a/libs/coin-modules/coin-zcash/src/bridge/operations.ts
+++ b/libs/coin-modules/coin-zcash/src/bridge/operations.ts
@@ -201,6 +201,11 @@ export function collectIronwoodSpendableNotes(
 
 /**
  * Converts raw shielded transactions to BtcOperation format.
+ *
+ * Operation value uses the Ironwood pool only, matching `getPrivateBalance` /
+ * `account.balance`. Orchard and Sapling notes still drive `type` for history
+ * labelling but are excluded from value so ops reconcile with the spendable
+ * private balance.
  */
 export function convertShieldedTransactionsToOperations(
   shieldedTransactions: ShieldedTransaction[],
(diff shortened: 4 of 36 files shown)


apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashTransferFromSelector.tsx
```diff
@@ -16,7 +16,7 @@ import Label from "~/renderer/components/Label";
 import Discreet, { useDiscreetMode } from "~/renderer/components/Discreet";
 import { useAccountUnit } from "~/renderer/hooks/useAccountUnit";
 
-type Sender = "public" | "private";
+type Sender = "public" | "ironwood";
```
cted-ledger: UI-01 FR5 defines this as `transaction.sender = "public" | "private"`, and that value is the contract between the send modal and both Zcash coin-modules. `@ledgerhq/coin-zcash` (#20197) implements it literally: `deriveZcashTransferType` there takes `"public" | "private"`, so an `"ironwood"` sender falls through to the default branch and yields `transparent` — the modal would show the Ironwood card while the transaction spends transparent UTXOs, with no error anywhere.

Suggestion: keep `"private"` as the value and resolve the pool in the coin-module, where the account's notes are known.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashTransferFromSelector.tsx
```diff
@@ -92,20 +86,23 @@ const ZcashTransferFromSelector = ({ account, transaction, onChange }: Props) =>
 
   if (!active) return null;
 
-  const sender: Sender = tx.sender ?? "public";
+  const sender: Sender = tx.sender === "ironwood" ? "ironwood" : "public";
```
cted-ledger: Any transaction that already carries `sender: "private"` (a persisted draft, or a flow that set it before this change) renders as Public here while the transaction still says private — the card shown and the pool signed disagree. If the `"ironwood"` value is kept, this needs a normalisation on mount rather than a read-time coercion.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashTransferFromSelector.tsx
```diff
@@ -140,26 +137,21 @@ const ZcashTransferFromSelector = ({ account, transaction, onChange }: Props) =>
...
+          aria-pressed={sender === "ironwood"}
           disabled={!fvkAvailable}
-          onClick={() => select("private")}
-          data-testid="transfer-from-private"
+          onClick={() => select("ironwood")}
+          data-testid="transfer-from-ironwood"
         >
           <CardTitle>
-            <Trans i18nKey="zcash.shielded.send.transferFrom.private.title" />
+            <Trans i18nKey="zcash.shielded.send.transferFrom.ironwood.title" />
```
cted-ledger: This is where the pool gets named in the UI, which IW-UI-01 FR2 rules out: *"the shielded send targets the Ironwood pool, reusing the existing send-modal patterns (UI-01..05). No pool is named in the signing/transfer view"*. "Ironwood" is a consensus-level detail the user cannot act on — post-NU6.3 it is simply what "Private" means.

Keeping `transferFrom.private.title` / `.subtitle` also avoids the i18n churn: the nine other locales still ship `transferFrom.private.*` and none defines `transferFrom.ironwood.*`.

cted-ledger (commented):
Thanks for wiring the Ironwood send path — the adapter side is solid, and `getPrivateBalance` in the summary footer is exactly what IW-UI-01 asks for (FR1: surface the Ironwood balance in the existing shielded display).

My main concern is the transfer-from selector, which goes against two validated specs.

**IW-UI-01** (Ironwood pool UI) is explicit in its scope: *"No dual-pool UI. […] there is no dual-pool display, no Orchard-closure messaging, no send guard, and no 'Move to Ironwood' action — the shielded UI simply reflects the Ironwood pool"*, and FR2: *"the shielded send targets the Ironwood pool, reusing the existing send-modal patterns (UI-01..05). **No pool is named in the signing/transfer view**"*. Choosing Ironwood is a routing decision that belongs to the coin-module (IW-COIN-01 FR2: *"route shielded sends to the Ironwood pool"*), not a choice we surface to the user — who has no Orchard funds to choose from in the first place, which is the very premise of both specs.

**UI-01** (transfer-from selectors) locks the contract this component implements: two cards, Public balance and Private balance (FR2), the source stored as `transaction.sender = "public" | "private"` (FR5), the Private card never disabled whatever the sync state (FR4 — banners from UI-03 do the gating), and an explicit unavailable state when no FVK is present (FR6, also an acceptance criterion).

Concretely, I'd keep the card labelled Private/Shielded and the `sender` value `"private"`, and let `deriveZcashTransferType` map a private sender to the Ironwood flows. That keeps the protocol detail where it belongs and leaves the UI contract untouched.

One cross-module consequence worth flagging: `@ledgerhq/coin-zcash` (#20197, same `zcashShielded` routing) implements the UI-01 contract and only knows `"public" | "private"`. A transaction carrying `sender: "ironwood"` falls through its default branch and derives `transparent` — the send would spend transparent UTXOs instead of shielded notes, silently. Whatever we settle on has to be identical in both modules, since the flag can route either way.

Two more points outside the UI, in the line comments: the `transparent-to-shielded` routing, and the zcash-utils version this branch builds against.

Nothing here blocks the adapter work, which I'd happily see land as-is.

apps/ledger-live-desktop/src/renderer/families/bitcoin/__tests__/ZcashTransferFromSelector.test.tsx
```diff
@@ -58,63 +58,95 @@ describe("ZcashTransferFromSelector", () => {
...
     fireEvent.click(screen.getByTestId("transfer-from-public"));
 
     expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ sender: "public" }));
   });
 
-  it("keeps the Private card clickable in every sync state (never disabled)", () => {
-    const onChange = renderSelector(buildAccount({ syncState: "disabled", ufvk: "uview-test" }), {
+  it("disables the Ironwood card when no FVK is available", () => {
+    renderSelector(buildAccount({ ufvk: null }), { sender: "public" } as Partial<Transaction>);
+    expect(screen.getByTestId("transfer-from-ironwood")).toBeDisabled();
```
cted-ledger: UI-01 is explicit on this state: FR4 says the Private card is *"selectable in all sync states — it is never disabled"* (the UI-03 banners do the gating), and FR6 requires the card to *indicate* that the private balance is unavailable when no FVK is present — both are acceptance criteria of that spec.

This revision removes the `Unavailable` hint, its i18n key and the two tests that covered it, while keeping `disabled={!fvkAvailable}`: the card ends up greyed out with nothing telling the user why. Worth restoring the hint (and ideally dropping the `disabled`, per FR4).

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/types.ts
```diff
@@ -234,7 +234,7 @@ export type ZcashTransferType =
 export type ZcashTransaction = Transaction & {
   transferType: ZcashTransferType;
   /** Source balance pool selected on the Recipient step. */
-  sender?: "public" | "private";
+  sender?: "public" | "private" | "ironwood";
```
cted-ledger: This widening is public API of `@ledgerhq/coin-bitcoin`, and it has to stay in step with `@ledgerhq/coin-zcash`, which the same `zcashShielded` flag can route to. If we keep the UI-01 contract (`"public" | "private"`), nothing changes here and both modules stay compatible.

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/index.ts
```diff
@@ -674,7 +662,10 @@ const zcashChainAdapter: ChainAdapter = {
           spends: mapSpends(tx.selectedNotes!),
           transparentInputs,
           outputs: mapOutputs(tx),
-        });
+        };
+        const buildResult = isIronwoodFlow
```
cted-ledger: `transparent-to-shielded` stays on the V5 builder here, and I think it needs to go through `buildIronwoodTransaction` too: post-NU6.3 a unified recipient is credited in the Ironwood pool, so a shielding transaction carries an Ironwood bundle — which is what IW-COIN-01 FR2 and IW-UTILS-02 describe (shielded sends built by the V6 Ironwood builder).

This is what we hit on device on the coin-zcash side: t→z through the V5 builder fails, and routing it to the V6 builder then needed the anchor to come from the commitment-tree frontier because Zaino does not serve `GetSubtreeRoots` for Ironwood (fixed in ledger-zcash-utils#36). Happy to walk through the traces if useful.

.changeset/warm-cycles-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"ledger-live-desktop": minor
```
cted-ledger: Two release-side points.

This PR changes the public surface of `@ledgerhq/coin-bitcoin` (new IPC client method, widened `sender`, Ironwood build routing) but the changeset only bumps `ledger-live-desktop` — Copilot raised this twice and it still stands.

Also, the catalog in `pnpm-workspace.yaml` still pins `@ledgerhq/zcash-utils@1.1.0`, which predates the NU6.3 fixes (ledger-zcash-utils#36, published as 1.1.1): V5 format pinning, Orchard proving/verifying keys selected from the consensus branch, change added through the builder's change API, and the frontier-derived shielding anchor. Against 1.1.0 the Ironwood path cannot complete on NU6.3 whichever builder it reaches. The bump is on the coin-zcash branch; rebasing on develop once it lands is enough.

libs/coin-modules/coin-zcash/src/constants.ts
```diff
@@ -3,7 +3,7 @@ import type { ZcashPrivateInfo } from "./network/types";
 
 export const ZCASH_LOG_TYPE = "zcash";
 export const ZCASH_GRPC_URL_TESTNET = "https://testnet.zec.rocks";
-export const ZCASH_GRPC_URL_MAINNET = "https://zec-indexer.coin.ledger-test.com";
+export const ZCASH_GRPC_URL_MAINNET = "https://zec.rocks:443";
```
may01: Please update `docs/services.md` with relevant link

may01 (commented):
Please check this comments on tests:

> [BLOCKER] signOperation.test.ts will fail (TypeScript + logic)
> 
> Lines 185–236: the existing test matrix includes ["shielded", "z→z", "v1"] and ["shielded-to-transparent", "z→t", "v1"]. The PR adds both to IRONWOOD_BUNDLE_TRANSFER_TYPES, so they'll call the V6 builder. The test expects the V1 builder (mockCraftTransaction) — both assertions will fail.
> Lines 190–191, 302: literal values "ironwood" / "ironwood-to-transparent" are used as Transaction["transferType"] after those members are removed from the type — TypeScript compile error.
> [BLOCKER] getTransactionStatus.test.ts will fail (TypeScript + logic)
> 
> Line 88: "ironwood" / "ironwood-to-transparent" cast as ZcashTransferType — compile error after type removal.
> Lines 356–374: fixture orchardBalance: 20_000, ironwoodBalance: 10_000_000, sends amount: 30_000 with transferType: "shielded". Old code routed to orchardBalance (insufficient → error). New code always uses ironwoodBalance (covers 30k → no error). Test assertion fails.
> [BLOCKER] ZcashTransferFromSelector.test.tsx will fail
> 
> Lines 108–111: screen.getByTestId("transfer-from-private-unavailable") throws when the element is absent. The PR removes the JSX carrying that data-testid. If the new UX relies on the pre-existing disabled={!fvkAvailable} attribute instead, the test must be updated.

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountBalanceSummaryFooter.tsx
```diff
@@ -24,6 +24,7 @@ import {
   ZCASH_CHECK_OUTDATED_SYNC_INTERVAL,
   ZCASH_OUTDATED_SYNC_INTERVAL_MINUTES,
 } from "@ledgerhq/coin-bitcoin/chain-adapters/zcash/constants";
+import { getPrivateBalance } from "@ledgerhq/coin-bitcoin/chain-adapters/zcash/balance";
```
cted-ledger: **issue (blocker):** this imports `getPrivateBalance` from `@ledgerhq/coin-bitcoin`, but the `coin-bitcoin` copy is no longer part of this PR — so it still returns `orchard + sapling + ironwood`, while the balance it is subtracted from comes from `coin-zcash` and now excludes Orchard/Sapling.

The two are always paired at runtime: this component early-returns unless `zcashShielded` is enabled (line 293-299), and that is precisely the flag under which `resolveFamily()` routes the account to `coin-zcash` (`libs/ledger-live-common/src/bridge/impl.ts:37`, `bridge/zcashRouting.ts:14`). `coin-zcash`'s sync sets `account.balance` via its own `computeZcashBalance` (`coin-zcash/src/bridge/sync.ts:412,657,725`), which after this PR is `transparent + ironwood`.

So for an account with Orchard or Sapling notes, at line 310:

```
_transparentBalance = (transparent + ironwood) − (orchard + sapling + ironwood)
                    = transparent − orchard − sapling
```

→ the "Transparent balance" label is under-reported and goes **negative** whenever `orchard + sapling > transparent`. And the "Private balance" label here counts all three pools, while `ZcashTransferFromSelector` shows Ironwood only — the same account shows two different private balances depending on the screen.

This looks like a leftover from the earlier revision that also changed `coin-bitcoin/src/chain-adapters/zcash/balance.ts`: dropping that half made the import point at a now-divergent definition. Sourcing the helper from `@ledgerhq/coin-zcash` (consistent with `SendModalTitle.tsx`, which this PR switches to `@ledgerhq/coin-zcash/types`) would restore the pairing.

An assertion that the three labels sum consistently for an account holding Orchard/Sapling notes would have caught this — the new test in `__tests__/AccountBalanceSummaryFooter.test.tsx` sets `balance` by hand to `transparent + ironwood`, so it encodes the coin-zcash convention while the component reads the coin-bitcoin one, and passes either way.

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountBalanceSummaryFooter.tsx
```diff
@@ -309,7 +306,7 @@ const AccountBalanceSummaryFooter = ({ account }: Props) => {
   };
 
   const _availableBalance = balance ?? BigNumber(0);
-  const _privateBalance = orchardBalance.plus(saplingBalance);
+  const _privateBalance = getPrivateBalance(privateInfo);
```
cted-ledger, line 317: **suggestion:** beyond the module-mismatch on the import above, deriving the transparent figure as `balance − private` (line 310) couples this label to whichever pool set `getPrivateBalance` happens to include. That is what turns a definition change in another package into a wrong — possibly negative — number on screen, with no failing test here.

`getTransparentBalance(utxos)` already exists (`coin-zcash/src/logic/account/balance.ts:17-20`, and its `coin-bitcoin` twin) and is currently only called inside the syncs. Computing the transparent balance from its own source rather than by subtraction would make both labels independent of the aggregation choice, and would keep working through the next pool migration.

libs/coin-modules/coin-zcash/src/logic/account/balance.ts
```diff
@@ -20,12 +20,18 @@ export function getTransparentBalance(
...
+ * are deliberately excluded so this value matches the spendable "Private
+ * balance" shown in the send modal (see ZcashTransferFromSelector) instead of
+ * exceeding it by any residual Orchard/Sapling notes.
+ */
 export function getPrivateBalance(privateInfo: PrivateBalances | undefined | null): BigNumber {
   if (!privateInfo) return new BigNumber(0);
-  return (privateInfo.orchardBalance ?? new BigNumber(0))
-    .plus(privateInfo.saplingBalance ?? new BigNumber(0))
-    .plus(privateInfo.ironwoodBalance ?? new BigNumber(0));
+  return privateInfo.ironwoodBalance ?? new BigNumber(0);
```
cted-ledger, line 38: **issue:** two consequences of this change that the diff doesn't follow through on.

**1. `account.balance` stops reconciling with the sum of the account's operations.** `coin-zcash/src/bridge/operations.ts` is untouched: the operation `value` is still computed from `allNotes`, which spans `orchard_outputs` + `sapling_outputs` + `ironwood_outputs` (`operations.ts:211-217`). So an account holding Orchard notes gets history rows crediting value that `balance` — and `spendableBalance`, same value — never reflects. That's independent of the product decision: the decision says don't count those pools in the balance, it doesn't say leave them counted in the operations. Either exclude the deprecated pools from the operation value too, or keep them in `balance` and restrict only `spendableBalance`.

**2. The two packages now disagree on an account's balance.** `coin-bitcoin/src/chain-adapters/zcash/balance.ts` keeps summing all three pools, so the same account is worth different amounts depending on which module serves it. The differential test in `libs/ledger-live-common/src/families/zcash/accounting.differential.test.ts` exists precisely to catch adapter-vs-`coin-zcash` divergence, but it only covers `prepareTransaction` / `getTransactionStatus` / `estimateMaxSpendable` — not `getPrivateBalance` / `computeZcashBalance`. Extending it to the balance helpers would document the divergence as intentional (or catch it if it isn't); as it stands the disagreement is silent, and it is what the footer import above trips over.

**nitpick:** the comment says the deprecated pools are excluded "so this value matches the spendable Private balance shown in the send modal". That is an argument about spendability; `balance` and `spendableBalance` are separate fields for that reason, and `coin-zcash/src/bridge/sync.ts` sets both to the same value. Worth stating explicitly in the comment that hiding the funds from the total is the intended product behaviour, not just a by-product of matching the modal — otherwise the next reader will try to "fix" it.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashTransferFromSelector.tsx
```diff
@@ -92,20 +86,21 @@ const ZcashTransferFromSelector = ({ account, transaction, onChange }: Props) =>
...
-  const saplingBalance = privateInfo?.saplingBalance ?? BigNumber(0);
+  // Only the Ironwood pool is spendable from the send flow; Sapling/Orchard send
+  // flows are deprecated, so the private card shows the Ironwood balance and the
+  // total balance excludes the deprecated pools (see getPrivateBalance).
+  const ironwoodBalance = privateInfo?.ironwoodBalance ?? BigNumber(0);
 
   const totalBalance = account.balance ?? BigNumber(0);
-  const privateBalance = orchardBalance.plus(saplingBalance);
-  const transparentBalance = totalBalance.minus(privateBalance);
+  const transparentBalance = totalBalance.minus(ironwoodBalance);
```
cted-ledger: **nitpick:** this is the third spelling of "private balance" in the diff — here it's a direct `privateInfo?.ironwoodBalance` read (line 96), where the footer goes through `getPrivateBalance` and both then derive transparent by subtraction.

This one is *correct* under the flag (the component only renders when `shieldedEnabled`, so `account.balance` comes from `coin-zcash` and equals `transparent + ironwood`) — but it is correct by coincidence of two independent definitions agreeing, which is what breaks in the footer. Routing both components through the same helper, from the same package, would collapse the three back into one.

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -8917,8 +8917,7 @@
           },
           "private": {
             "title": "Private balance",
-            "subtitle": "Shielded",
-            "unavailable": "Private balance unavailable"
+            "subtitle": "Shielded"
```
cted-ledger: **issue (minor, UX):** dropping the `unavailable` key removes the only explanation the user got for a non-actionable Private card.

Per the new test (`ZcashTransferFromSelector.test.tsx`, "disables the Private card when no FVK is available"), the card is still disabled when `fvkAvailable` is false — it now just shows a balance and silently refuses to be selected, with no reason given. This reads like collateral from the Ironwood rework rather than an intentional UX change; if the wording was the problem, replacing the string seems better than removing it.

libs/coin-modules/coin-zcash/src/bridge/serialization.ts
```diff
@@ -61,19 +61,21 @@ export function fromBitcoinOutputRaw([
 export function toBitcoinResourcesRaw(r: BitcoinResources): BitcoinResourcesRaw {
   return {
     utxos: r.utxos.map(toBitcoinOutputRaw),
-    walletAccount: r.walletAccount && wallet.exportToSerializedAccountSync(r.walletAccount),
+    ...(r.walletAccount && {
```
cted-ledger, line 64: **nitpick (out of scope):** this `walletAccount` conditional-spread rework, the paired `walletAccount?: WalletAccount | undefined` → `walletAccount?: WalletAccount` change in `types/bridge.ts`, and the new `coin:zcash` script in `package.json` have nothing to do with restricting the shielded send flow to Ironwood.

They look harmless (the `exactOptionalPropertyTypes`-style motivation is understandable), but they widen the review surface of a PR that also carries a user-visible balance-semantics change. Splitting them out would make both halves easier to review and to revert independently.

.changeset/zcash-ironwood-only-shielded-send.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ledgerhq/coin-zcash": minor
+"@ledgerhq/coin-bitcoin": minor
+"@ledgerhq/live-common": patch
+"ledger-live-desktop": minor
+---
+
+Restrict the Zcash shielded send flow to the Ironwood pool only.
```
cted-ledger: **issue:** two things to fix here.

1. `"@ledgerhq/coin-bitcoin": minor` is now stale — no `coin-bitcoin` file is in the diff anymore, so this publishes a needless minor bump of the shared UTXO module (which every other UTXO chain depends on).
2. The description covers only the send-flow restriction, but the PR also changes what the account **balance** means: `getPrivateBalance` / `computeZcashBalance` now exclude the Orchard and Sapling pools, so `account.balance` and `spendableBalance` drop for any account holding notes there. That's the more user-visible half and the one a release reader needs to see.

**nitpick:** no trailing newline at end of file.

cted-ledger (commented):
Reviewed at `2bc86eb`. The product decision is taken as given: restricting the spendable shielded pool to Ironwood, and excluding the deprecated Orchard/Sapling pools from the displayed balance, is accepted and not re-litigated here. The comments below are the technical consequences that remain **under that decision**.

**Blocker** — now that the `coin-bitcoin` changes are out of the diff, `AccountBalanceSummaryFooter.tsx` still imports `getPrivateBalance` from `@ledgerhq/coin-bitcoin`, where it is unchanged and still sums all three pools. The footer only renders behind `zcashShielded`, which is exactly the flag that routes the account to `coin-zcash` (`bridge/impl.ts:37`), whose `account.balance` now excludes Orchard/Sapling. So the two sides no longer agree:

- "Transparent balance" = `(transparent + ironwood) − (orchard + sapling + ironwood)` = `transparent − orchard − sapling` → under-reported, and **negative** as soon as the account holds Orchard or Sapling notes.
- "Private balance" in the footer = all three pools, while the send modal's Private card = Ironwood only. Two different figures for the same thing, on two screens.

Details inline on the import.

**Other findings**
- `balance` no longer reconciles with the sum of the account's operations: `coin-zcash/src/bridge/operations.ts:211-217` still values shielded operations over all three pools.
- The two packages now disagree on what an account's balance is, and the differential test that exists to catch exactly that doesn't cover the balance helpers.
- The changeset declares a `coin-bitcoin` minor bump although no `coin-bitcoin` file is touched anymore, and doesn't mention the balance-semantics change.
- A few drive-by changes unrelated to Ironwood, and one small UX regression (the removed "unavailable" hint).

Scope note: all of this is bounded by `zcashShielded` (default `false`, `bridge/zcashRouting.ts:14`), so no GA regression — but it lands squarely on the flagged cohort, which is where Orchard notes actually sit, since pre-NU6.3 shielding landed there.

Refs read: `origin/develop` and this PR head `2bc86eb`.

apps/ledger-live-desktop/src/renderer/families/bitcoin/AccountBalanceSummaryFooter.tsx
cted-ledger: **suggestion:** following up on the above, because the flag deserves an explicit note — switching the import alone would not be enough.

First, to be precise about the flag-off path: the single call site is line 309, which sits **after** the early return at 293-299, so with `zcashShielded` off the component returns `null` and `getPrivateBalance` is never evaluated. The `coin-bitcoin` import serves no flag-off path here. `coin-bitcoin` already handles flag-off on its own side, in the adapter rather than in the UI — `chain-adapters/zcash/index.ts:472-481` returns the transparent balance only when `isZcashShieldedEnabled()` is false, precisely so the displayed balance can't exceed the transparent max spendable.

The part that does need handling in the UI: **`account.balance` is persisted, and its provenance depends on the flag state at the last sync, not at render time.** With the toggle flipped ON → OFF → ON without a re-sync in between, the stored balance was written by the `coin-bitcoin` adapter (transparent-only, per the lines above), while this component renders and subtracts a private balance from it — giving an under-reported or negative "Transparent balance" whichever package the helper is imported from. Same class of mismatch on a fresh flag-ON before the first `coin-zcash` sync completes.

So the robust shape is to stop deriving one label from another:

- transparent ← `getTransparentBalance(bitcoinResources.utxos)`, from its own source;
- private ← the pool the send flow actually spends (Ironwood), from `privateInfo`;
- available ← `account.balance`, displayed as-is.

Each label then holds regardless of which module last wrote the balance, and the three stay consistent across a toggle without waiting for a re-sync. If the labels must keep being derived from `account.balance` instead, then the component needs to read the flag the same way the routing does (`bridge/zcashRouting.ts` / `bridge/impl.ts:37`) **and** handle the stale-provenance case explicitly — clamping at zero is not enough, since a wrong-but-positive figure is worse than a visibly missing one.

cted-ledger (commented):
Re-reviewed at `593231c`. **praise:** every point from the previous round is addressed, and two of them better than suggested — the footer and the selector now both source `getPrivateBalance` / `getTransparentBalance` from `coin-zcash` and derive the transparent figure from the account's own UTXOs, and the differential suite now *asserts* the balance divergence between the two packages (the 50,000 gap) instead of leaving it silent. The changeset's stale `coin-bitcoin` bump is gone and it mentions the balance/operations change; the Private-card hint is back.

I also checked the 8 files migrated from `coin-bitcoin` to `coin-zcash` imports: three of them pull **runtime** constants (`ZCASH_SHIELDED_TX_TYPES`, `DEFAULT_ZCASH_PRIVATE_INFO`, `ZCASH_ACTIVATION_DATE*` + the two sync intervals). I diffed each against its `coin-bitcoin` twin at this head — byte-identical, so the migration is neutral on the flag-off path where the adapter still produces the data. `SendModalTitle` also reads `transferType` only when `shieldedEnabled`, so the narrowed 4-key union can't be indexed with a legacy `ironwood*` value.

The new `operations.ts` value logic is where I have findings — three defects in the branch selection, inline. Plus one product-level consequence of "Ironwood-only value" that I don't think the PR should absorb silently; it's on the doc comment and it's the only one I'd call a decision rather than a bug.

Nothing else outstanding from my side except the out-of-scope drive-bys (`serialization.ts`, `types/bridge.ts`, `package.json`), which still stand as a suggestion, not a blocker.

libs/coin-modules/coin-zcash/src/bridge/operations.ts
```diff
@@ -201,32 +201,44 @@ export function collectIronwoodSpendableNotes(
...
     } else if (txType.endsWith("_OUT")) {
-      // Outgoing value spans both destinations the funds can reach — other shielded
-      // addresses and transparent ones — and includes the fee, the convention the
-      // transparent operations and the optimistic operation both follow.
-      value = sumNotes("outgoing").plus(deshieldedValue(tx)).plus(fee);
+      const hasIronwoodNotes = ironwoodNotes.length > 0;
+      const hasDeprecatedPoolNotes =
+        (tx.decryptedData?.orchard_outputs?.length ?? 0) > 0 ||
+        (tx.decryptedData?.sapling_outputs?.length ?? 0) > 0;
+      if (hasIronwoodNotes || txType === "SHIELDED_TX_IRONWOOD_OUT") {
```
cted-ledger: **issue:** the `|| txType === "SHIELDED_TX_IRONWOOD_OUT"` disjunct is dead code.

`getTxType` (`:50-76`) can only return an Ironwood type through `shieldedTypeByProtocol` (`:22-31`) or `deshieldedType` (`:33-44`), and **both require a non-empty `ironwoodOutputs`** — the former via `hasDirection(ironwoodOutputs, direction)`, the latter via an explicit `ironwoodOutputs.length > 0`. Since `hasIronwoodNotes` on line 227 reads the same `tx.decryptedData.ironwood_outputs` array, `txType === "SHIELDED_TX_IRONWOOD_OUT"` implies `hasIronwoodNotes === true`, so the second operand can never be the one that makes this condition true.

Harmless at runtime, but it reads as a guard against a case that exists, which will mislead the next reader of this branch. Worth dropping.

libs/coin-modules/coin-zcash/src/bridge/operations.ts
```diff
@@ -201,32 +201,44 @@ export function collectIronwoodSpendableNotes(
...
     let value = new BigNumber(0);
     if (txType.endsWith("_IN")) {
-      value = sumNotes("incoming");
+      value = sumIronwoodNotes("incoming");
     } else if (txType.endsWith("_OUT")) {
-      // Outgoing value spans both destinations the funds can reach — other shielded
-      // addresses and transparent ones — and includes the fee, the convention the
-      // transparent operations and the optimistic operation both follow.
-      value = sumNotes("outgoing").plus(deshieldedValue(tx)).plus(fee);
+      const hasIronwoodNotes = ironwoodNotes.length > 0;
```
cted-ledger: **issue:** `hasIronwoodNotes` is a presence check with no direction filter, while the `txType` it is paired with is chosen *per direction*.

`shieldedTypeByProtocol` (`:22-31`) returns `SHIELDED_TX_ORCHARD_OUT` whenever `hasDirection(ironwoodOutputs, "outgoing")` is false and Orchard has an outgoing note — **even if `ironwood_outputs` is non-empty**, e.g. it holds only an `incoming` or `internal` entry. In that state, line 231 still takes the Ironwood branch and computes `value = sumIronwoodNotes("outgoing") + deshieldedValue(tx) + fee`: `sumIronwoodNotes("outgoing")` is 0, but `deshieldedValue + fee` lands on an **Orchard-labelled** operation, driven purely by the presence of an unrelated Ironwood note.

So the pool that decides the *label* and the pool that decides the *value* can disagree. Aligning the two would mean filtering by direction here (`hasDirection(ironwoodNotes, "outgoing")`) rather than by length, or keying the whole branch off `txType` instead of off note presence.

No test covers this: `"op.value excludes Orchard and Sapling notes so it matches Ironwood-only balance"` is the only mixed-pool case and it's an `_IN`, where Ironwood's own directional match wins the label. Whether `ledger-zcash-utils` can actually emit that combination for one tx I can't verify from this repo — but the TS logic admits it, and the cheap fix doesn't depend on the answer.

libs/coin-modules/coin-zcash/src/bridge/operations.ts
```diff
@@ -201,32 +201,44 @@ export function collectIronwoodSpendableNotes(
...
+      const hasDeprecatedPoolNotes =
+        (tx.decryptedData?.orchard_outputs?.length ?? 0) > 0 ||
+        (tx.decryptedData?.sapling_outputs?.length ?? 0) > 0;
+      if (hasIronwoodNotes || txType === "SHIELDED_TX_IRONWOOD_OUT") {
+        // Outgoing value spans both destinations the funds can reach — other shielded
+        // addresses and transparent ones — and includes the fee, the convention the
+        // transparent operations and the optimistic operation both follow.
+        value = sumIronwoodNotes("outgoing").plus(deshieldedValue(tx)).plus(fee);
+      } else if (!hasDeprecatedPoolNotes) {
```
cted-ledger: **issue:** this branch assigns a non-zero value to a spend whose pool is, by its own admission, unknown — and the comment two lines below ("Orchard/Sapling outs: value stays 0") is therefore only true when *some* deprecated note survives.

When a whole-note z→t leaves no change at all, all three note arrays are empty. Nothing else on `ShieldedTransaction` / `DecryptedTransaction` (`network/types.ts:88-113`) carries pool provenance — `transparentOut` is a bare aggregate, and the per-note nullifier/position data in `SpendingFields` / `DecryptedOutput` doesn't exist for a spend that produced no note. So a **historical Orchard** whole-note z→t is indistinguishable from an Ironwood one here, `deshieldedType` labels it `SHIELDED_TX_ORCHARD_OUT` (`:33-44`, Orchard is the default fallback), and this branch then gives it `deshieldedValue + fee`.

That is an Orchard-pool spend contributing value to an Ironwood-only total — exactly what the doc comment on line 204-208 says is excluded. And the test `"op.value for an Ironwood z→t send that left no change is still the transparent amount plus the fee"` locks it in as expected behaviour (`type = SHIELDED_TX_ORCHARD_OUT`, `value = 1_000_000`), so the ambiguity is asserted rather than flagged.

Either the label should follow the value (treat the no-notes case as Ironwood and type it `IRONWOOD_OUT`), or the value should follow the label (0, like every other Orchard out) — but the current pairing of an Orchard label with an Ironwood-counted value is the one combination that can't be right either way. If it has to stay ambiguous, the test name and the comment should say so explicitly instead of calling the fixture "an Ironwood z→t send" when nothing in it is Ironwood.

libs/coin-modules/coin-zcash/src/bridge/operations.ts
```diff
@@ -201,32 +201,44 @@ export function collectIronwoodSpendableNotes(
 
 /**
  * Converts raw shielded transactions to BtcOperation format.
+ *
+ * Operation value uses the Ironwood pool only, matching `getPrivateBalance` /
```
cted-ledger: **question (product-level, worth an explicit decision):** making the operation value Ironwood-only does reconcile ops with `account.balance` — which is what I asked for — but it buys that consistency by having the history under-report real movements, and I'd rather that trade-off were chosen than inherited.

Two cases:

- **Orchard z→t to an external address.** `value = 0` here, and no transparent operation compensates: `mapTxToOperations` (`bridge/sync.ts:142-269`) only emits ops for account-owned addresses, so nothing in the account's history records that the funds left. The two pipelines never merge either — shielded op ids are `encodeOperationId(accountId, tx.id, "SHIELDED_TX_*")` (`:234`) vs `"IN"|"OUT"` on the transparent side (`sync.ts:220,253`), and `mergeOps` dedups by `op.id`. A real payment shows as a 0-amount row.
- **Incoming Orchard/Sapling.** Same, `value = 0`; the row still renders (the type stays `SHIELDED_TX_ORCHARD_IN` and `ConfirmationCheck.tsx:136` maps it to an icon), showing 0 for value actually received.

So the previous inconsistency (ops crediting value the balance ignored) is replaced by a history that displays 0 for genuine inflows and outflows. Both are wrong in the reader's eyes; this one is arguably worse, because a 0-amount row looks like a fact rather than a discrepancy.

The alternative that keeps both honest is to leave the operation value as the real amount and restrict only the **balance** (which is the product decision as I understood it: hide the deprecated pools from what is *spendable*, not deny that the funds moved). If the current shape is deliberate, it's worth stating in this comment why the history is expected to show zeros, so nobody "fixes" it later — and worth checking with whoever owns the spec whether a 0-value row is acceptable for a real payment, or whether those operations should be filtered out of the list entirely rather than shown at 0.

cted-ledger (approved):

may01 (approved):

PR #20204, opened 2026-07-29
[LWDM] chore(LIVE-32915): migrate @ledgerhq/errors in coin-modules

gre-ledger:
### 📝 Description

Removes direct imports of @ledgerhq/errors in libs/coin-modules and libs/coin-tester-modules, replacing them with native error classes from @ledgerhq/ledger-wallet-framework/errors, @ledgerhq/hw-transport/errors, or local modules.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35183](https://ledgerhq.atlassian.net/browse/LIVE-35183)
- **ADR** (if any): N/A

[LIVE-35183]: https://ledgerhq.atlassian.net/browse/LIVE-35183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #20221, opened 2026-07-30
[LWDM] feat(coin-zcash): serve Zcash accounts with a standalone coin-module (LIVE-34556)

cted-ledger:
### 📝 Description

Zcash accounts are served today by a chain-adapter inside `@ledgerhq/coin-bitcoin`,
which reuses the Bitcoin account/transaction model and its legacy transparent
signing path. This branch introduces a standalone `@ledgerhq/coin-zcash`
coin-module and routes Zcash to it, so that every flow — transparent `t→t`
included — is built and signed as a PCZT, with no PSBT fallback.

**Routing is behind the existing `zcashShielded` feature flag, OFF by default**:
with the flag off, `bridge/impl.ts` keeps resolving Zcash through the
`coin-bitcoin` adapter exactly as before, so this changes nothing for users until
the flag is turned on. No new flag was introduced.

The module was ported to mirror the adapter rather than to redesign it, which is
what keeps the migration reviewable. Two differential test suites pin that
equivalence and are the automated check against regression: they run the same
inputs through `coin-zcash` and through the `coin-bitcoin` adapter it came from,
and assert both agree on accounting (balances, operation types, spendable
amounts) and on account serialization. Where the two legitimately differ — the
ZIP-317 fee model, and the absence of a UTXO-picking strategy — the difference is
asserted rather than smoothed over.

Also in this branch:

- **Ironwood / V6 crafting.** The flows that carry an Ironwood bundle (an Ironwood
  spend, or an Ironwood output — which includes `t→z`, since a unified recipient
  resolves to the Ironwood pool from NU6.3) are dispatched to the V6 builder;
  Orchard flows and `t→t` stay on the V5 builder.
- **ZIP-317 transparent fee fix (LIVE-35152)**, in the `coin-bitcoin` Zcash
  adapter. Transparent sends were rejected from the mempool for "unpaid actions
  is higher than the limit": a single account-wide sat/vByte rate cannot express a
  per-action fee with a flat floor. The rate is now tightened to the transaction's
  actual layout through a new `ChainAdapter.resolveFeePerByte` hook. Carries its
  own changeset.
- **Desktop wiring.** The Zcash native engine runs in its own utility process; the
  `zcash:` IPC contract and its rspack bundle now belong to `coin-zcash`. The
  Bitcoin send UI no longer reads a UTXO strategy from a transaction that has
  none.

### 🔗 Context

- **JIRA / GitHub issue**: LIVE-34556 (standalone module), LIVE-35152 (ZIP-317
  transparent fee)
- **Depends on**: `LedgerHQ/ledger-zcash-utils#36` — NU6.3 activated on Mainnet on
  2026-07-28, and three behaviours of the native builder changed at that height,
  so crafting fails without those fixes. `@ledgerhq/zcash-utils` must be released
  and bumped here before the flag is enabled on Mainnet.

### Known limits

- A V6 PCZT reaches the device and stops there: the DMK Zcash signer kit announces
  a v1 container and carries no Ironwood serialization (tracked separately as the
  PCZT v2 Ironwood signing work).
- A shielded send to a third party (`z→z`) has no builder: it would need Orchard
  spends alongside an Ironwood output, and the V6 builder takes Ironwood spends
  only. From NU6.3 the Orchard pool is spendable but no longer creditable.

### Test plan

- [x] `pnpm test` for `@ledgerhq/coin-zcash` — 401 pass, `typecheck` and `lint`
      clean
- [x] Differential suites green against the `coin-bitcoin` adapter
- [x] Mainnet, flag ON: sync, `t→t` signed and broadcast, `t→z` crafts a V6 PCZT
      and reaches the device
- [ ] Regression pass with the flag OFF (Zcash unchanged through `coin-bitcoin`)
- [ ] Mobile is untouched and still routes through `coin-bitcoin`


libs/coin-modules/coin-zcash/src/bridge/getTransactionStatus.ts
```diff
@@ -0,0 +1,86 @@
+import { BigNumber } from "bignumber.js";
+import { NotEnoughBalance } from "@ledgerhq/errors";
```
gre-ledger: `@ledgerhq/errors` is being sunset — for coin-modules the replacement is `@ledgerhq/ledger-wallet-framework/errors`:

```suggestion
import { NotEnoughBalance } from "@ledgerhq/ledger-wallet-framework/errors";
```

(on behalf of @gre-ledger — feel free to ignore if this file is handled elsewhere)

gre-ledger (commented):
On behalf of @gre-ledger: `@ledgerhq/errors` is being sunset (LIVE-35072, see #20040) and should disappear soon, so this PR may need a small follow-up — see the 4 inline notes below.

Nothing blocking today, mostly to avoid adding new usages that will need migrating again. The pattern for `coin-modules` is `@ledgerhq/ledger-wallet-framework/errors`. Let me know if any of these don't apply.

libs/coin-modules/coin-zcash/src/bridge/statusHelpers.ts
```diff
@@ -0,0 +1,56 @@
+import { BigNumber } from "bignumber.js";
+import { InvalidAddress, RecipientRequired } from "@ledgerhq/errors";
```
gre-ledger: `@ledgerhq/errors` is being sunset — for coin-modules the replacement is `@ledgerhq/ledger-wallet-framework/errors`:

```suggestion
import { InvalidAddress, RecipientRequired } from "@ledgerhq/ledger-wallet-framework/errors";
```

(on behalf of @gre-ledger)

libs/coin-modules/coin-zcash/src/logic/transaction/broadcast.ts
```diff
@@ -0,0 +1,79 @@
+import { InvalidTransactionError } from "@ledgerhq/errors";
```
gre-ledger: `@ledgerhq/errors` is being sunset — for coin-modules the replacement is `@ledgerhq/ledger-wallet-framework/errors`:

```suggestion
import { InvalidTransactionError } from "@ledgerhq/ledger-wallet-framework/errors";
```

(on behalf of @gre-ledger)

libs/coin-modules/coin-zcash/src/logic/transaction/broadcast.unit.test.ts
```diff
@@ -0,0 +1,137 @@
+/**
+ * Tests for the double-spend guard and the gRPC submission it protects. The
+ * engine client is mocked: what matters is which outpoints are checked against
+ * the explorer, and when the submission is refused.
+ */
+import { InvalidTransactionError } from "@ledgerhq/errors";
```
gre-ledger: `@ledgerhq/errors` is being sunset — for coin-modules the replacement is `@ledgerhq/ledger-wallet-framework/errors`:

```suggestion
import { InvalidTransactionError } from "@ledgerhq/ledger-wallet-framework/errors";
```

(on behalf of @gre-ledger)

libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -25,6 +25,27 @@ import {
   loadBridgeExtensionsForFamily,
 } from "../coin-modules/registry";
 import { defaultBridgeExtensions } from "./defaultBridgeExtensions";
+import { isZcashShieldedEnabled } from "@ledgerhq/coin-zcash/constants";
```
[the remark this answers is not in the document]
cted-ledger: Fixed in a87fe8b. The flag mirror moved out of the coin module: `libs/ledger-live-common/src/bridge/zcashRouting.ts` now owns `setZcashShieldedEnabled` / `isZcashShieldedEnabled`, and `bridge/impl.ts` imports it from there — so shared bridge code no longer references `@ledgerhq/coin-zcash` at all, at type level or at runtime. `libs/ledger-live-common/src/coin-modules/no-coin-eager-imports.test.ts` passes.

The mirror has to live in live-common rather than in the coin module precisely because `bridge/impl.ts` is the reader, and the host app is the only layer that can resolve a React feature flag — same wiring as `setSuiGraphqlEnabled` / `setCosmosLdmkEnabled`.

libs/coin-modules/coin-zcash/src/walletBtcCurrency.ts
```diff
@@ -0,0 +1,21 @@
+import type { CryptoCurrency } from "@ledgerhq/ledger-wallet-framework/types";
+import { getCryptoCurrencyById } from "@ledgerhq/ledger-wallet-framework/currencies";
+import { getEnv } from "@ledgerhq/live-env";
+import type { WalletBtcCurrency } from "@ledgerhq/wallet-btc/crypto/types";
```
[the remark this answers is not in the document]
cted-ledger: Fixed in a87fe8b: the file moved to `libs/coin-modules/coin-zcash/src/walletBtcCurrency.ts`, out of the `**/src/network/**` scope the rule covers — which is also where `coin-bitcoin` keeps its equivalent, so the layout now matches the module this code was ported from. `oxlint src` on the module reports no `no-restricted-imports` warning for that file any more.

libs/coin-modules/coin-zcash/src/network/ZCash.ts
```diff
@@ -0,0 +1,240 @@
...
+        transactionDetailsJob(grpcUrl, requests, network, ufvk),
+    }),
+
+    syncShielded(syncArgs: SyncShieldedArgs): Observable<ShieldedSyncResult> {
+      return new Observable<ShieldedSyncResult>(subscriber => {
+        const validationError = deps.validateStartSyncArgs(syncArgs);
+        if (validationError) {
+          subscriber.error(validationError);
+          return;
+        }
```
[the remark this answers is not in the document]
cted-ledger: Fixed in a87fe8b — the message is wrapped now: `subscriber.error(new Error(validationError))`, since the validator reports a plain string while every consumer of the stream reads `name` and stack off an `Error`. The corresponding case in `ZCash.test.ts` asserts an `Error` instance rather than the raw string.

libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -25,6 +25,27 @@ import {
...
+  }
+  return currency.family;
+}
+
+// Cache key for the bridge caches below: the resolved family, plus the
+// currencyId for zcash, the one family whose resolution can flip mid-session.
+function bridgeCacheKey(currency: CryptoCurrency): string {
+  const family = resolveFamily(currency);
+  return currency.id === "zcash" ? `${family}:${currency.id}` : family;
+}
```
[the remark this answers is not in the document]
cted-ledger: Good catch — fixed in a87fe8b. `clearBridgeCache(family)` now deletes both the plain `family` key and every `` `${family}:` ``-prefixed entry, across the three caches (currency, account, mock) plus the `family|` bridge-cache keys. Without it a Zcash entry stored as `zcash:zcash` survived the eviction, and since rejections stay cached, a transient load failure could not be retried in-session.

Regression test in `libs/ledger-live-common/src/bridge/__tests__/impl.zcash-routing.test.ts`: after `clearBridgeCache("zcash")`, `getAccountBridge` hands back a fresh promise instead of the memoized one.

apps/ledger-live-desktop/src/renderer/families/bitcoin/useBitcoinPickingStrategy.tsx
```diff
@@ -19,7 +19,7 @@ type BitcoinStrategyResult = {
   item: Option | undefined | null;
   options: Array<Option>;
 };
-const useBitcoinPickingStrategy = (strategy: number): BitcoinStrategyResult => {
+const useBitcoinPickingStrategy = (strategy: number | undefined): BitcoinStrategyResult => {
   const [item, setItem] = useState(() => {
     return options.find(o => bitcoinPickingStrategy[o.value] === strategy);
   });
```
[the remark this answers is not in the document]
cted-ledger: Leaving this as is, deliberately.

The `if (newItem)` guard is `develop`'s behaviour, not something this PR introduces — the only change here is widening the parameter to `number | undefined`. The guard already applies on `develop` to `CUSTOM`, which is filtered out of `options`, so clearing the state on a no-match would change what a Bitcoin account with a custom strategy displays. That is a behaviour change to the Bitcoin fee selector, out of scope for a Zcash PR.

For the `undefined` case specifically, nothing can go stale on screen: `SendAmountFields` computes `hideFeeSelection` for a Zcash account and returns `null` before the picker is ever rendered, so the hook's `item` is not read in that state. The widening exists only because the hook is called above that early return, which is what made a `coin-zcash` transaction (no `utxoStrategy`) crash on `strategy` — see the regression test added in `__tests__/SendAmountFields.test.tsx`.

libs/coin-modules/coin-zcash/src/network/ipc/utility-entry.ts
```diff
@@ -0,0 +1,284 @@
...
+  try {
+    await startSyncJob(
+      {
+        grpcUrl: args.grpcUrl,
+        network: args.network,
+        viewingKey: args.viewingKey,
+        startBlockHeight: args.startBlockHeight,
+        maxBatchSize: args.maxBatchSize,
+        ...(args.knownNullifiers?.length && { knownNullifiers: args.knownNullifiers }),
+      },
```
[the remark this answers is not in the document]
cted-ledger: Changed to an explicit ternary, so the spread operand is always an object. It also matches the form already used in `src/network/engine.ts`.

The TypeScript part of the report does not hold though: spreading a `number` is legal (a primitive contributes no own enumerable properties), and `pnpm --filter @ledgerhq/coin-zcash typecheck` is green on the original line — which is byte-identical to `coin-bitcoin`'s `src/chain-adapters/zcash/ipc/utility-entry.ts:104`, the code this file was ported from. Runtime behaviour was already correct too, since `0` and `undefined` both spread to nothing. So this is a readability change, not a bug fix.

libs/coin-modules/coin-zcash/src/bridge/getTransactionStatus.ts
cted-ledger: Thanks — already done, in 6a1552b: `NotEnoughBalance` here, `InvalidAddress` / `RecipientRequired` in `bridge/statusHelpers.ts` and `InvalidTransactionError` in `logic/transaction/broadcast.ts` (+ its test) all come from `@ledgerhq/ledger-wallet-framework/errors` now, and `@ledgerhq/errors` is out of the module's `package.json`. The four Zcash-specific classes stay local in `src/types/errors.ts`.

One note for the `logic/**` ones: those imports warned under `libs/coin-modules/.oxlintrc.json`, whose allowlist for `api|logic|network` did not include the framework. #20204 added it, merged earlier today, so the warnings clear on the next rebase — no carve-out needed on this side.

libs/coin-modules/coin-zcash/src/bridge/statusHelpers.ts
cted-ledger: Done in 6a1552b — this file takes the error from `@ledgerhq/ledger-wallet-framework/errors`, and `@ledgerhq/errors` is no longer a dependency of the module.

libs/coin-modules/coin-zcash/src/logic/transaction/broadcast.ts
cted-ledger: Done in 6a1552b — this file takes the error from `@ledgerhq/ledger-wallet-framework/errors`, and `@ledgerhq/errors` is no longer a dependency of the module.

libs/coin-modules/coin-zcash/src/logic/transaction/broadcast.unit.test.ts
cted-ledger: Done in 6a1552b — this file takes the error from `@ledgerhq/ledger-wallet-framework/errors`, and `@ledgerhq/errors` is no longer a dependency of the module.

deepyjr (approved):

Sebastien-Dav1d (approved):

ysitbon (approved):

henri-ly (approved):

PR #20235, opened 2026-07-30
[LWDM] test(cosmos): run zenrock integration test weekly (flaky node)

ishaba:
### 📝 Description

`zenrock.integration.test.ts` fails every per-PR/daily Integration Tests run — 22× `API HTTP 502` from its public LCD `zenrock.api.m.stavr.tech`. zenrock has no paid node provider and `zenrock.coin.ledger.com` is being removed, so there is no stable endpoint to point at.

Safe quick fix to unblock affected PR's.

### 🔗 Context

- **JIRA / GitHub issue**: _n/a_ 
- **ADR** (if any): _n/a_

### DIFF
--- a/libs/ledger-live-common/jest.config.ts
+++ b/libs/ledger-live-common/jest.config.ts
@@ -24,6 +24,7 @@ const weeklyIntegrationTests = [
   "src/families/cosmos/datasets/stargaze.integration.test.ts",
   "src/families/cosmos/datasets/quicksilver.integration.test.ts",
   "src/families/cosmos/datasets/xion.integration.test.ts",
+  "src/families/cosmos/datasets/zenrock.integration.test.ts",
   "src/families/mina/bridge.integration.test.ts",
 ];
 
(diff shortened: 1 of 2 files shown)


libs/ledger-live-common/jest.config.ts
```diff
@@ -24,6 +24,7 @@ const weeklyIntegrationTests = [
   "src/families/cosmos/datasets/stargaze.integration.test.ts",
   "src/families/cosmos/datasets/quicksilver.integration.test.ts",
   "src/families/cosmos/datasets/xion.integration.test.ts",
+  "src/families/cosmos/datasets/zenrock.integration.test.ts",
   "src/families/mina/bridge.integration.test.ts",
```
gre-ledger, line 28: the fact we need to maintain this list is problematic (scalability, #team-platform ping,..)
let's see if coin-integration team eventually have a solution to reduce the need for these integration tests OR fine a better way to not have to do custom config 🙏

gre-ledger (approved):

PR #20241, opened 2026-07-30
fix(wallet-cli): print amounts with an ascii space

Justkant:
## Summary

wallet-cli printed every amount with a non-breaking space (U+00A0) between the value and the ticker, inherited from live-common's `formatCurrencyUnit` (`libs/live-currency-format/src/formatCurrencyUnit.ts:7`, joined into the suffix at line 211).

NBSP is the right choice for the desktop and mobile GUIs, so the shared helper is deliberately left untouched. Instead, a new CLI boundary `formatCliCurrencyUnit` in `src/wallet/formatter/currency.ts` normalizes U+00A0 and U+202F to a plain space, and all five of wallet-cli's direct `formatCurrencyUnit` call sites now go through it.

The character is invisible in a terminal but breaks `grep`, `awk`, `cut` and anything consuming `--output json`.

Affected surfaces:

- `balances` (human and JSON)
- `operations` (human and JSON)
- `earn positions`
- `send --dry-run` and the `send` "prepared" event
- the earn deposit/withdraw dry runs

`swap quote` was already clean.

## Test plan

- [x] `pnpm test` in `apps/wallet-cli`: 827 pass, 0 fail
- [x] Typecheck, lint and format all clean
- [x] Device-free regression suite D went from `passed 36/41 failed 5 skipped 1` (failing D4, D5, D6, D23, D24) to `passed 41/41 failed 0 skipped 1`, verified by stashing the fix and re-running against the same session and network

## Notes

No changeset: `@ledgerhq/wallet-cli` is `private: true` and nothing under `libs/` changed.

### DIFF
--- a/apps/wallet-cli/src/wallet/README.md
+++ b/apps/wallet-cli/src/wallet/README.md
@@ -17,13 +17,18 @@ wallet/
   models.ts          ← serializable types + Zod schemas
   index.ts           ← WalletAdapter (routing)
   intents/           ← TransactionIntent schemas (bitcoin, evm, solana)
-  formatter/         ← HumanFormatter and JsonFormatter
+  formatter/         ← HumanFormatter, JsonFormatter, and currency.ts
   compatibility/
     bridge.ts        ← BridgeAdapter (full sync via live-common bridge)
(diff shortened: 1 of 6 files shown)


Justkant:
Reopening for now, we'll handle that soon

PR #20242, opened 2026-07-30
test(wallet-cli): add the release regression harness and plan

Justkant:
> **Stacked on #20241** — this PR targets `fix/wallet-cli-ascii-space-amounts`, not `develop`. Review and merge #20241 first.

## Summary

Adds the wallet-cli release regression plan, and the harness that automates the part of it needing no device:

- `docs/regression/README.md` — entry point: what CI already covers, prerequisites, how to run a campaign.
- `docs/regression/no-device.md` — the case tables the harness automates.
- `docs/regression/with-device.md` — the cases a person has to run with a Ledger on USB (wallet-cli is USB/DMK only, no Speculos support).
- `docs/regression/releases/2.1.0.md` — per-release focus: what changed and which cases it puts at risk.
- `scripts/regression/*.ts` — the harness itself, suites A–D, in bun/TypeScript.

The harness needs only bun — no `jq`, no `bash`, no GNU `timeout`, the last of which stock macOS does not ship — so it runs on any contributor's machine, on macOS, Linux and Windows alike. Case ids it prints match the tables in `docs/regression/no-device.md`.

Windows specifics handled:

- Binary resolution covers `dist/windows-x64/cli.exe`.
- Env isolation also overrides `USERPROFILE`/`LOCALAPPDATA`/`APPDATA`, since `stateDir` from `@bunli/utils` reads `%LOCALAPPDATA%` on win32 and ignores `XDG_STATE_HOME`.
- The POSIX file-mode cases skip with a reason on Windows and as root.

Also:

- A `--session <dir>` flag to pin suite D at a specific session directory instead of the developer's real one.
- A missing session degrades to per-case skips instead of aborting the suite.
- `tsconfig.json` adds `scripts/regression/*.ts` to `include` so the harness is typechecked.

The cases were first prototyped as bash scripts that were never committed — which is why the case ids look pre-established, and why the parity numbers below are evidence rather than a self-comparison.

## Test plan

- [x] Hermetic suites B and C: `passed 44/44 failed 0 skipped 0`, identical to the bash prototype on the same machine
- [x] Suite D: `passed 41/41 failed 0 skipped 1` once #20241 is applied. Without it, suite D shows five NBSP-related failures (D4, D5, D6, D23, D24) — hence the stacking.

## Notes

No changeset: `@ledgerhq/wallet-cli` is `private: true` and nothing under `libs/` changed.

### DIFF
--- a/apps/wallet-cli/README.md
+++ b/apps/wallet-cli/README.md
@@ -105,6 +105,15 @@ pnpm start -- <command> [args]
 
 If `USER_ID` is unset, it defaults to `wallet-cli` so DMK firmware distribution salt stays stable for this CLI (`env-setup.ts`).
 
+## Regression testing
+
+Before a release, work through the [regression plan](docs/regression/README.md). Cases that
+need no device are automated (`bun scripts/regression/run.ts`); the rest need a Ledger on USB
+and a person to confirm on-screen. Per-release focus lives in
+[`docs/regression/releases/`](docs/regression/releases/).
+
+Adding or changing a command? Update the matching case table in the same PR.
+
 ## Relation to `ledger-live` CLI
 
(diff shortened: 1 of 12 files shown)


Justkant:
Reopening for now, we'll handle that soon

PR #20257, opened 2026-07-30
[LWDM] fix(coin-modules): add coverageProvider babel to coin-canton and coin-concordium

gre-ledger:
`workerThreads: true` causes `generateEmptyCoverage` in jest 30 to produce V8-format coverage output instead of Istanbul format, crashing `addFileCoverage` with "Invalid file coverage object, missing keys, found:data". Adding explicit `coverageProvider: "babel"` fixes this.


gre-ledger:
Closing — the fix (coverageProvider: babel) doesn't actually change the NX hash so coin-canton:coverage was hitting remote cache, not running fresh. The coverage error is pre-existing and not caused by this migration. Will track separately if needed.

PR #20289, opened 2026-07-31
[LWDM] fix(coin-framework): improve nonce check to avoid error at broadcast

qperrot:
### 📝 Description

When two transactions are sent in quick succession, the second one can be
rejected by the node with **"nonce too low"** (or trigger an unintended
replacement). This happens because the sequence/nonce source (indexer or a
load-balanced RPC) lags behind a just-broadcast transaction and briefly returns
a nonce that was already consumed by the first send.

Separately, when an intent carries an **explicit** nonce (dapp / WalletConnect /
speed-up flows), a stale value was only rejected *after* signing — the user went
through the whole device flow before hitting the error.

## What this PR does

1. **`nextSequenceWithPending` (generic-coin-framework)** — at sign time, derive
   the next sequence from **both** the network source and the locally-tracked
   pending operations: `max(networkSequence, highestPendingSequence + 1)`.
   Pending ops are tracked optimistically right after broadcast and carry their
   `transactionSequenceNumber`, so the next send no longer reuses a nonce. It is
   deterministic and self-corrects: once the network source catches up,
   `networkSequence` wins again.

2. **`validateNonce` (coin-evm `validateIntent`)** — when an intent carries an
   explicit sequence, validate it against the sender's on-chain nonce **before**
   signing and surface `"nonce is too low"` up front. Best-effort: a failed nonce
   fetch never blocks the user. When no explicit sequence is set, nothing is
   checked (crafting fetches a fresh nonce).

## Changes

- `generic-coin-framework/utils.ts` — add `nextSequenceWithPending()`
- `generic-coin-framework/signOperation.ts` — use it when the intent has no explicit sequence
- `coin-evm/logic/validateIntent.ts` — add the pre-sign `validateNonce()` guard

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35248](https://ledgerhq.atlassian.net/browse/LIVE-35248)
- **ADR** (if any): 

[LIVE-35248]: https://ledgerhq.atlassian.net/browse/LIVE-35248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/quick-nonces-heal.md
+++ b/.changeset/quick-nonces-heal.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": minor
+---
+
+Prevent "nonce too low" errors on rapid consecutive sends by deriving the next sequence from both the network source and locally-tracked pending operations.
--- a/libs/ledger-live-common/src/bridge/generic-coin-framework/signOperation.ts
+++ b/libs/ledger-live-common/src/bridge/generic-coin-framework/signOperation.ts
@@ -3,7 +3,12 @@ import { SignerContext } from "@ledgerhq/ledger-wallet-framework/signer";
 import type { Account, DeviceId, SignOperationEvent, AccountBridge } from "@ledgerhq/types-live";
 import { getCoinModuleApi } from "./api";
 import { getBridgeApi } from "./bridge";
-import { bigNumberToBigIntDeep, buildOptimisticOperation, transactionToIntent } from "./utils";
+import {
+  bigNumberToBigIntDeep,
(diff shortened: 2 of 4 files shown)


dilaouid (dismissed):

francois-guerin-ledger (approved):

qperrot:
https://ledger-live.allure.green.ledgerlabs.net/allure/reports/91154840-4476-4f36-8a0e-19529647c589/

PR #20296, opened 2026-08-01
[LWDM] feat(coin-near): add CoinModuleApi for native NEAR and staking

pawell24:
### 📝 Description

Exposes the CoinModuleApi surface for NEAR alongside the existing account bridge, covering the
native asset and delegated staking (stake / unstake / withdraw).

The generic-framework flag stays off, so Ledger Live keeps routing NEAR through the account bridge
and user-facing behaviour is unchanged. Enabling it needs UI work on the NEAR screens and lands
separately.

The account bridge and the new surface now share their logic rather than duplicating it:

- `logic/fees.ts` holds the fee formula, used by both `getFeesForTransaction` and `estimateFees`.
- `logic/actions.ts` holds the action builder, used by both `buildTransaction` and
  `craftTransaction`.
- `network/protocolConfig.ts` derives gas and storage costs from the protocol config behind a
  30-minute cache. The new path has no preload step, and the preload defaults are zeros, so
  sourcing costs from them would have produced a zero fee.

`src/api/` moved to `src/network/`, since `src/api/index.ts` is now the module factory.

**Scope**: the ticket AC also lists tokens. NEAR has no NEP-141 support anywhere in Ledger Live,
including the account bridge, so there is nothing to migrate. Adding it is separate work.

**Testing**: 213 unit tests, including 7 that assert the crafted transaction is byte-identical to
the account bridge's, plus 40 integration tests against mainnet. Both paths were also exercised
manually in Ledger Live Desktop (balances, history, fees, validation, send-max, device signing and
a staking broadcast on mainnet); the values matched the account bridge to 8 decimals.

Enabling the framework for NEAR also needs a one-line fix in
`generic-coin-framework/utils.ts`, where an optimistic operation's fee is converted with
`toString()` instead of `toFixed()`. BigNumber switches to exponential notation from 1e21 up and
`BigInt` cannot parse it, which a NEAR staking fee (24 decimals) reaches at 0.0015. It is not
specific to NEAR, so it ships on its own.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34284](https://ledgerhq.atlassian.net/browse/LIVE-34284)
- **ADR** (if any): n/a

[LIVE-34284]: https://ledgerhq.atlassian.net/browse/LIVE-34284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-modules/coin-near/src/api/parity.integ.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    const liquid = balance
+      .minus(nearResources.stakedBalance)
+      .minus(nearResources.availableBalance)
+      .minus(nearResources.pendingBalance);
+
+    expect(native.value).toBe(BigInt(liquid.toFixed(0)));
+    expect(native.value).toBeLessThanOrEqual(BigInt(balance.toFixed(0)));
+  }, 120_000);
+
```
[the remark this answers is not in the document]
pawell24: Correct, fixed in cc9d58f. It passed only because `nearkat.near` holds no stake, so `liquid == balance` and both semantics matched. Also added an assertion on `locked`, since `value` alone cannot tell them apart on a stake-free account.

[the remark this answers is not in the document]
pawell24: Correct, fixed in cc9d58f. It passed only because nearkat.near holds no stake, so liquid == balance and both semantics matched. Also added an assertion on locked, since value alone cannot tell them apart on a stake-free account.

libs/ledger-live-common/src/coin-modules/loaders.ts
```diff
@@ -301,6 +301,7 @@ export const coinModuleLoaders: CoinModuleLoader[] = [
     family: "near",
     supportedCoins: ["near"],
     loadSetup: () => import("../families/near/setup"),
+    loadLocalApi: () => import("../families/near/coinModuleApi").then(m => m.createLocalNearApi),
```
cted-ledger, line 306: **issue (blocks the flag-flip, not this PR):** no `loadBridgeApi`, so `getBridgeApi()` returns `{}` (`bridge.ts:7`) and staking won't work:

- `stakingSupported` is read off the **BridgeApi** (`getAccountShape.ts:346`), not the CoinModuleApi → `getValidators()` never runs. `api/index.ts:38` claims otherwise; the flag has to live wallet-side (cf. `evm/bridge/api.ts:209`).
- No `computeIntentType` → `defaultComputeIntentType` rejects `withdraw` (`utils.ts:426-436`), which the desktop flow sets (`WithdrawingFlowModal/Body.tsx:81`) → throws.
- No `usesStakingPositions` → `nearResources` never populated.

**Fix:** add `families/near/bridge/api.ts` + wire `loadBridgeApi`. If deferred to the flag-flip PR, drop the `stakingSupported` field and its comment.

libs/coin-modules/coin-near/src/network/protocolConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+const fetchActionCosts = async (): Promise<NearActionCosts> => {
+  const protocolConfig = await getProtocolConfig();
+
+  if (!protocolConfig) {
+    throw new NearProtocolConfigNotLoaded();
+  }
+
+  const { storage_amount_per_byte, transaction_costs } = protocolConfig.runtime_config;
```
cted-ledger: **issue:** two fields of the object you destructure aren't read, and mainnet is on protocol 86 where both apply (`rpc.mainnet.near.org/status`):

- **`min_gas_purchase_price`** — nearcore floors the gas price on the *execution* half only (`runtime/runtime/src/config.rs:459-468`), while `logic/fees.ts:42` applies one `gas_price` to both. It goes `0 → 1e9` at protocol 85, ~10× the current price: ≈ **0.007 NEAR** short on a send to an implicit account, and a send-max the runtime rejects.
- **`account_creation_charge`** (0.007 N) — the activation minimum at `validateIntent.ts:68` is 0.00182 N, ~4× low.

Both are `#[serde(default)]` and `0` before protocol 85, so reading them is version-safe.

**Fix:** add both to `NearActionCosts`; `max(gasPrice, minGasPurchasePrice)` on the `executionFee` term only; minimum becomes `max(storageCost × NEW_ACCOUNT_SIZE, accountCreationCharge)`. Inherited from the bridge — a separate ticket is fine.

libs/coin-modules/coin-near/src/logic/fees.ts
```diff
@@ -0,0 +1,43 @@
...
+// account also pays for creating it and adding its access key.
+export const computeFees = ({
+  mode,
+  recipient,
+  useAllAmount,
+  gasPrice,
+  costs,
+}: FeeInput): BigNumber => {
+  if (STAKING_MODES.has(mode)) {
+    return getStakingFees({ mode, useAllAmount: useAllAmount ?? false }, gasPrice);
```
cted-ledger, line 29: **issue:** `getStakingFees` (`src/logic.ts:175-182`) divides by 10, but the runtime locks the **whole** prepaid gas at conversion time (`runtime/runtime/src/config.rs:447-467`) and refunds the unburnt part afterwards. For `deposit_and_stake` (125 TGas) the real hold is ~10× the fee shown.

Unchanged from `develop`, so not a regression — what changes is the blast radius: `computeFees` now feeds `estimateFees` and `validateIntent`, where this decides `NotEnoughBalance` and the send-max ceiling. A staking send-max computed from a 10×-low fee fails at broadcast.

**Fix:** charge the full prepaid gas and let the refund show up as it does on-chain. Worth its own ticket — but please confirm it's intentional, since `validateIntent` now inherits it silently.

libs/coin-modules/coin-near/src/logic/getBalance.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Its value is the account total, delegated buckets included, which is what the account bridge
+ * reports as the account balance. `locked` is the non-spendable part of it: the staked, pending
+ * and withdrawable buckets plus the storage deposit and minimum-balance buffer, capped at the
+ * value so it can never claim more is locked than the account holds. `value - locked` then
+ * reproduces the account bridge's spendable balance, which floors at zero. This mirrors
+ * coin-tezos, which likewise reports the total as `value` and the frozen part as `locked`.
+ */
+export async function getBalance(address: string): Promise<Balance[]> {
+  const { balance, nearResources } = await getAccount(address);
```
cted-ledger, line 21: **issue:** `getAccount` derives `storageUsageBalance` from `getCurrentNearPreloadData().storageCost` (`src/network/node.ts:60`) — preload state this path never populates. It's the one place the PR's own rule ("the new path has no preload step") isn't applied.

It gives the right number only because the fallback in `src/constants.ts:1` currently equals the live protocol value. If that moves, the api path's `locked` — and the spendable balance the send flow enforces — silently diverges from the bridge's while both run.

`api/parity.integ.test.ts` can't catch it: `beforeAll` calls `preload()` (l.61), and the zeroed-costs case (l.142) asserts on `estimateFees` only.

**Fix:** read it from `getActionCosts()`, as `logic/transaction/validateIntent.ts:68` already does, and assert `locked` in the zeroed-preload parity case.

libs/coin-modules/coin-near/src/logic/staking/getStakes.ts
```diff
@@ -0,0 +1,64 @@
...
+      const actions: StakeAction[] = ["undelegate"];
+      items.push({
+        uid: `${address}:${validatorId}:staked`,
+        address,
+        delegate: validatorId,
+        state: "active",
+        actions,
+        asset: { type: "native" },
+        amount: BigInt(staked.toFixed(0)),
+        amountDeposited: BigInt(staked.toFixed(0)),
```
cted-ledger: **issue:** `amountDeposited = amount` asserts `amountRewarded = 0` (the framework contract is `amount = amountDeposited + amountRewarded`). But `get_account_staked_balance` derives from `stake_shares` (`near/core-contracts` `staking-pool/src/lib.rs:328,374`): rewards compound into the shares and the principal isn't recoverable — so any delegation with accrued rewards reports them as principal.

**Fix:** omit `amountDeposited` (it's optional) rather than asserting a split NEAR can't express. Same at l.37-38 and l.50-51 — your comment at l.6-8 already makes the point.

libs/coin-modules/coin-near/src/logic/transaction/validateIntent.ts
```diff
@@ -0,0 +1,211 @@
...
+    errors.amount = new NotEnoughBalance();
+  } else if (amount <= 0n && !intent.useAllAmount) {
+    errors.amount = new AmountRequired();
+  } else if (recipient.isNewAccount && amount < BigInt(recipient.storageCost.toFixed(0))) {
+    errors.amount = new NearActivationFeeNotCovered(undefined, {
+      formattedNewAccountStorageCost: recipient.formattedStorageCost,
+    });
+  }
+
+  if (intent.useAllAmount && balances.some(b => b.stake !== undefined)) {
```
cted-ledger: **issue:** `extractBalances` never carries `stake` (`generic-coin-framework/utils.ts:167-200`), so this condition is always false and the warning can't fire. The bridge raised it on `nearResources.stakingPositions.length > 0 && useAllAmount`, so send-max with an active delegation used to warn and now doesn't.

`logic/staking/pooledAmount.ts:16-19` documents exactly this — it just hasn't reached `validateSend`.

**Fix:** reuse the `pooledAmount` network fallback, or derive from `locked > reserve`.

cted-ledger (commented):
## Review: `coin-near` CoinModuleApi migration

**Nothing blocks the merge as scoped.** The flag is off, and the `logic/fees.ts` + `logic/actions.ts` extraction and the `preload.ts` rework are behaviour-preserving for the live bridge path — checked line by line, including the asymmetric `withdraw`+`useAllAmount` gas multiplier and the `.force()` that keeps refresh-on-every-preload. `spendable()` reproduces the bridge's reserve exactly. The byte-parity suite and `validateIntent`'s staking edge cases are both above the bar for this migration class.

8 inline comments, in three groups:

- **Staking won't work at flag-flip** — no `BridgeApi` registered (`loaders.ts`), `getStakes` asserts a split NEAR can't express (`getStakes.ts`), `NearRecommendUnstake` can't fire (`validateIntent.ts`). Plus the inherited `÷ 10` on staking fees (`fees.ts`), which this PR promotes from display to decision.
- **Two protocol params missing, mainnet already past them** — `min_gas_purchase_price` + `account_creation_charge` (`protocolConfig.ts`). Inherited, but this is where the derivation now lives.
- **Structure & tests** — ADR-049 chose a different `logic/` shape (`actions.ts`); `logic/` has no MSW and no integ coverage (`broadcast.test.ts`).

**Coverage: UNVERIFIED** — I didn't execute the suites, so the 213/40 figures and the 80% threshold are neither confirmed nor contradicted. **Size** is fine as one unit (40 non-test files, 6 pure renames). **Tokens descope** holds — there's no NEP-141 path anywhere in the module; worth a line on LIVE-34284 so the AC matches the delivery.

libs/coin-modules/coin-near/src/logic/actions.ts
```diff
@@ -0,0 +1,58 @@
+import * as nearAPI from "near-api-js";
```
cted-ledger, line 1: **issue:** [ADR-049](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7324958730) (CF, **ACCEPTED**) chose **Option B** for `logic/`: flat, all tests in `logic/tests/`. Reference is `coin-solana` (no `logic/index.ts` either — `api/` imports each function directly). This PR builds the nested shape, i.e. the option not chosen, plus a 4th `staking/` folder.

Bad luck more than bad work — the ADR landed days ago and the boilerplate plus the last three merged migrations are all still nested. Two heads-ups before moving anything:

- The ADR contradicts itself (§5 says co-locate, Decision Outcome says `logic/tests/`). Confirm with its author so 20 files don't move twice.
- Keep `.integ.test.ts`, **not** the ADR examples' `.integration.test.ts` — `jest.integ.config.js` matches `.integ.test.ts$` and `jest.config.js` ignores `.integration.test.ts`, so such files run under **neither** config.

libs/coin-modules/coin-near/src/logic/tests/broadcast.unit.test.ts
```diff
@@ -0,0 +1,26 @@
+import { broadcastTransaction } from "../../network";
```
cted-ledger, line 1: **issue:** `logic/` has only one of the three test levels — **0 MSW, 0 integ** under `logic/`, vs solana 7/10, kaspa 8/11, multiversx 0/9. (This file is named `*.test.ts` but is a pure unit test; the convention reserves that suffix for network-free integration i.e. MSW — `coin-modules/SKILL.md:25`.)

`network/` **is** MSW-backed and integ exists at api level, so it's the middle layer that's uncovered — and that's where this migration class breaks: cursor propagation, `undefined`-instead-of-`[]`, HTTP-200-with-an-error-body (this very function never checks the hash is non-empty). Today those paths only run against mainnet.

Also `jest.config.js` loads no `@ledgerhq/disable-network-setup` (already a devDep, unused) — `coin-solana` does it in a single project, so it's a one-line add, no two-project split needed. Without it, unmatched requests in the MSW suites hit the real endpoint instead of failing.

Minor: `lastBlock` has no suite of its own (it's in `getBlockInfo.test.ts:33-39`); `pooledAmount` has none.

libs/coin-modules/coin-near/src/logic/transaction/validateIntent.ts
```diff
@@ -0,0 +1,211 @@
+import type {
+  Balance,
+  FeeEstimation,
+  TransactionValidation,
+} from "@ledgerhq/coin-module-framework/api/index";
+import { formatCurrencyUnit } from "@ledgerhq/coin-module-framework/currencies/index";
+import { getCryptoCurrencyById } from "@ledgerhq/ledger-wallet-framework/currencies";
```
francois-guerin-ledger: We should avoid importing `getCryptoCurrencyById` here. Prefer hardcoding `.units[0]` and `.name`.

libs/coin-modules/coin-near/src/logic/transaction/validateIntent.ts
```diff
@@ -0,0 +1,211 @@
...
+import {
+  AmountRequired,
+  InvalidAddress,
+  InvalidAddressBecauseDestinationIsAlsoSource,
+  NotEnoughBalance,
+  RecipientRequired,
+} from "@ledgerhq/ledger-wallet-framework/errors";
```
francois-guerin-ledger:
```suggestion
} from "@ledgerhq/coin-module-framework/errors";
```
Import those errors from `coin-module-framework`.

libs/ledger-live-common/src/coin-modules/loaders.ts
pawell24: Agreed, deferred to the flag-flip/enablement PR, as you noted. Tracked as follow-up, not forgotten.

francois-guerin-ledger (approved):

Sebastien-Dav1d (approved):

henri-ly (approved):

PR #20297, opened 2026-08-02
[LWDM] feat(coin-tester): add the NEAR coin-tester

pawell24:
### 📝 Description

Adds `@ledgerhq/coin-tester-near`, a deterministic integration tester for the NEAR family. It boots a local `near-sandbox` node and runs one scenario through both the legacy account bridge and the generic coin-framework bridge, so the two surfaces are compared against each other on every run.

The scenario covers a transfer to an existing named account, a transfer to a fresh implicit account, and the full staking cycle: delegate, undelegate, and withdraw after the unlock period.

No Docker. `near-sandbox` downloads a native `neard` build for the host platform, so it runs on the standard CI runner and on a laptop with no extra setup. `staking_pool.wasm` is pinned to a commit and verified against its sha256, so neither upstream nor a stale cache can change the contract the scenario delegates to.

**Three changes outside the tester package** are needed for the generic-adapter half of the run:

- `families/near/bridge/api.ts` (new): NEAR names its staking modes `stake` / `unstake` / `withdraw`, while the framework's default whitelist knows `stake` / `unstake` / `finalize_unstake` and throws on anything else. Without the mapping, `withdraw` cannot be built through the framework at all.
- `generic-coin-framework/createTransaction.ts`: NEAR joins the Cardano branch, since a NEAR nonce belongs to an access key rather than the account, so the default transaction must be signable without callers setting a sequence.
- `generic-coin-framework/utils.ts`: `buildOptimisticOperation` converted fees with `BigNumber.toString()`, which switches to exponential notation from 1e21 up and cannot be parsed by `BigInt`. At NEAR's 24 decimals a staking fee crosses that threshold at 0.0015, so every staked operation threw. Now uses `toFixed()`, with a regression test asserting the boundary directly.

The last one is a shared-framework fix affecting any family with a large-magnitude fee. It is bundled here rather than split out because the tester is red without it.

### What the first end-to-end run caught

Four contract mismatches that unit tests had missed, all wrong assumptions about the indexer contract rather than bugs in the coin module:

| Finding | Why unit tests missed it |
| --- | --- |
| `/v3/stats` must be stubbed | `getGasPrice`'s node-RPC fallback only covers a response without a `gas_price`, not a request that throws |
| `staking-deposits` returns a bare array | the module maps over the body itself, with no `{ data }` envelope |
| `loadBridgeApi` is required | the framework reads `computeIntentType` from the loader registry, not from the caller |
| a fresh account is not immediately final | the module queries with `finality: "final"` while a new account is first visible only optimistically |

### Notes for reviewers

- Most of the diff is test code and the generated lockfile entry. The shipped logic is `bridge/api.ts` (30 lines) plus two small framework edits.
- `families/near/bridge/api.ts` gets the `bridge/api.test.ts` that the other nine families already have.
- Runtime is about 52 s per strategy. The coin-tester job is `continue-on-error: true`, so it reports without gating the merge.
- **Based on #20296.** The base branch will retarget to `develop` automatically once that merges.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34293](https://ledgerhq.atlassian.net/browse/LIVE-34293)
- **ADR** (if any): n/a

[LIVE-34293]: https://ledgerhq.atlassian.net/browse/LIVE-34293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-tester-modules/coin-tester-near/src/scenarii.test.ts
+++ b/libs/coin-tester-modules/coin-tester-near/src/scenarii.test.ts
@@ -1,0 +1,21 @@
+import { executeScenario } from "@ledgerhq/coin-tester/main";
+import { killSandbox, scenarioNear } from "./scenarii/near";
+
+// Both strategies run the same 6-transaction scenario end to end (sandbox startup, staking pool
+// deploy, epoch fast-forward); the default 5s is nowhere near enough for either.
+jest.setTimeout(20 * 60 * 1000);
+
+["exit", "SIGINT", "SIGQUIT", "SIGTERM", "SIGUSR1", "SIGUSR2", "uncaughtException"].forEach(e =>
+  process.on(e, () => {
+    void killSandbox().catch(() => {});
+  }),
+);
+
+describe.each([["legacy"], ["generic-adapter"]] as const)(
+  "NEAR Deterministic Tester (%s strategy)",
+  strategy => {
+    it("scenario NEAR", async () => {
+      await executeScenario(scenarioNear, strategy);
+    });
+  },
+);
--- a/libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
+++ b/libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
@@ -1,0 +1,204 @@
+import BigNumber from "bignumber.js";
+import type { Account } from "@ledgerhq/types-live";
+import { LiveConfig } from "@ledgerhq/live-config/LiveConfig";
+import type { Scenario, ScenarioTransaction } from "@ledgerhq/coin-tester/main";
+import type { GenericTransaction } from "@ledgerhq/live-common/bridge/generic-coin-framework/types";
+import type { SetupServerApi } from "msw/node";
+import {
+  EPOCHS_TO_UNLOCK_BLOCKS,
+  IMPLICIT_RECIPIENT_ID,
+  NAMED_RECIPIENT_ID,
+  NEAR,
+  POOL_ID,
+  RECIPIENT_BALANCE,
+  SENDER_BALANCE,
+  SENDER_ID,
+  coinConfig,
+  makeAccount,
+} from "../fixtures";
+import { getBridges } from "../helpers";
+import { recordTransaction, resetIndexer, startIndexer } from "../indexer";
+import { startSandbox, type SandboxHandle } from "../sandbox";
+import { buildSigners, randomKeyPair } from "../signer";
+import { deployStakingPool, pingPool } from "../stakingPool";
+
+let sandbox: SandboxHandle | undefined;
+let indexer: SetupServerApi | undefined;
+
+/** Index of the next transaction, so the unstake lock can be skipped before the withdrawal. */
+let step = 0;
+const WITHDRAW_STEP = 4;
+
+const has = (account: Account, type: string): boolean =>
+  account.operations.some(operation => operation.type === type);
+
+const countOf = (account: Account, type: string): number =>
+  account.operations.filter(operation => operation.type === type).length;
+
+const ONE_NEAR = new BigNumber((1n * NEAR).toString());
+const FIVE_NEAR = new BigNumber((5n * NEAR).toString());
+const TEN_NEAR = new BigNumber((10n * NEAR).toString());
+
+type NearScenarioTransaction = ScenarioTransaction<GenericTransaction, Account>;
+
+// Asserted against the balance, not the operation's `value`: the two bridges disagree on whether
+// the fee is inside `value`, but `balance = previous - amount - fee` holds for both.
+function makeTransactions(): NearScenarioTransaction[] {
+  return [
+    {
+      name: "transfer to a named account",
+      mode: "send",
+      recipient: NAMED_RECIPIENT_ID,
+      amount: ONE_NEAR,
+      expect: (previous, current) => {
+        const [latestOp] = current.operations;
+        expect(latestOp.type).toBe("OUT");
+        expect(latestOp.recipients).toContain(NAMED_RECIPIENT_ID);
+        expect(latestOp.fee.gt(0)).toBe(true);
+        expect(current.balance).toEqual(previous.balance.minus(ONE_NEAR).minus(latestOp.fee));
+        expect(countOf(current, "OUT")).toBe(1);
+      },
+    },
+    {
+      name: "transfer to an implicit account that does not exist yet",
+      mode: "send",
+      recipient: IMPLICIT_RECIPIENT_ID,
+      amount: ONE_NEAR,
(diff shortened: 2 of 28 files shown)


gre-ledger (dismissed):

libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
```diff
@@ -0,0 +1,181 @@
...
+const countOf = (account: Account, type: string): number =>
+  account.operations.filter(operation => operation.type === type).length;
+
+const ONE_NEAR = new BigNumber((1n * NEAR).toString());
+const FIVE_NEAR = new BigNumber((5n * NEAR).toString());
+const TEN_NEAR = new BigNumber((10n * NEAR).toString());
+
+// Asserted against the balance, not the operation's `value`: the two bridges disagree on whether
+// the fee is inside `value`, but `balance = previous - amount - fee` holds for both.
+function scenarioTransactions(): ScenarioTransaction<GenericTransaction, Account>[] {
```
qperrot: check on other coin testers for consistency
`function makeTransactions(): XrpScenarioTransaction[]`

libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
```diff
@@ -0,0 +1,181 @@
...
+      recipient: POOL_ID,
+      amount: TEN_NEAR,
+      expect: (previous, current) => {
+        const [latestOp] = current.operations;
+        expect(latestOp.type).toBe("STAKE");
+        expect(latestOp.fee.gt(0)).toBe(true);
+        expect(previous.spendableBalance.minus(current.spendableBalance).gte(TEN_NEAR)).toBe(true);
+        // The stake leaves the spendable balance but stays in the total, which only pays the fee.
+        expect(previous.balance.minus(current.balance).lt(TEN_NEAR)).toBe(true);
+        expect(current.spendableBalance.lt(current.balance)).toBe(true);
```
qperrot, line 91: can we have better check like 
`xpect(current.balance).toStrictEqual(previous.balance.minus(latest.value));`

libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
```diff
@@ -0,0 +1,181 @@
...
+      amount: FIVE_NEAR,
+      expect: (previous, current) => {
+        const [latestOp] = current.operations;
+        expect(latestOp.type).toBe("WITHDRAW_UNSTAKED");
+        expect(latestOp.fee.gt(0)).toBe(true);
+        expect(current.spendableBalance.gt(previous.spendableBalance)).toBe(true);
+      },
+    },
+  ];
+}
```
qperrot, line 135: Would be nice to have a send max as well

libs/coin-tester-modules/coin-tester-near/src/scenarii.test.ts
```diff
@@ -0,0 +1,13 @@
...
+import { scenarioNear } from "./scenarii/near";
+
+jest.setTimeout(20 * 60 * 1000);
+
+describe.each([["legacy"], ["generic-adapter"]] as const)(
+  "NEAR Deterministic Tester (%s strategy)",
+  strategy => {
+    it("scenario NEAR", async () => {
+      await executeScenario(scenarioNear, strategy);
+    });
```
qperrot, line 19: why jest.setTimeout ?

other coin tester are doing
```
["exit", "SIGINT", "SIGQUIT", "SIGTERM", "SIGUSR1", "SIGUSR2", "uncaughtException"].forEach(e =>
  process.on(e, () => {
    killChainSimulator().catch(() => {});
  }),
);
```

pnpm-lock.yaml
```diff
@@ -22812,7 +22876,7 @@ packages:
       metro-react-native-babel-preset: '*'
       react: 19.1.4
       react-dom: 19.1.4
-      webpack: '*'
+      webpack: ^5.89.0
```
qperrot: is this change related to your PR ?
double check if it hasn't been changed by mistake

pnpm-lock.yaml
```diff
@@ -25156,7 +25220,7 @@ packages:
     resolution: {integrity: sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==}
     engines: {node: '>=6'}
     peerDependencies:
-      rxjs: '*'
+      rxjs: ^5.5.10
```
qperrot: is this change related to your PR ?
double check if it hasn't been changed by mistake

libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
pawell24: Renamed to makeTransactions with a NearScenarioTransaction type alias, matching xrp

libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
pawell24: Tightened to expect(current.balance).toStrictEqual(previous.balance.minus(latestOp.fee)) plus the same for spendableBalance

libs/coin-tester-modules/coin-tester-near/src/scenarii/near.ts
pawell24: Added a "send max to a named account" transaction at the end, drains spendableBalance to zero

libs/coin-tester-modules/coin-tester-near/src/scenarii.test.ts
pawell24: Kept the timeout (sandbox startup + staking pool deploy + 2 strategies need more than the default 5s). Added the missing process.on(...) cleanup via a new exported killSandbox(), same pattern as xrp/cosmos

knip.json
```diff
@@ -249,6 +249,9 @@
     },
     "./libs/coin-tester-modules/coin-tester-multiversx": {
       "entry": ["src/scenarii/multiversx.ts"]
+    },
+    "./libs/coin-tester-modules/coin-tester-near": {
+      "entry": ["src/scenarii/near.ts"]
```
gre-ledger: please remove this addition. it is not necessary. if it breaks again, you would need to actually fix the packahe.json entry points (exports)

francois-guerin-ledger (approved):

henri-ly (approved):

gre-ledger (approved):

PR #20281, opened 2026-07-31
[LWD] feat(contacts): render desktop contact edit and delete actions

claudiiafg:
### 📝 Description

Renders edit and delete contact actions on Desktop contact detail. Adds the shared `@features/flow-contacts` edit/delete UI and wires it in the Desktop MVVM layer. Consumes existing shared state (`useContactDetailActionsViewModel`, delete lifecycle, signer-required flag) without reimplementing Contacts rules.

### Changes

**Shared (`@features/flow-contacts`)**
- `EditContact` step — rename dialog, validation, apply-changes flow
- `useContactDetailEditDeleteFlowViewModel` — orchestrates edit, delete, and signer states
- `ContactsDeleteContactDialog` / `ContactsEditSignerDialog` (web + native exports)
- `ContactDetailActions` — pencil + trash icon buttons on detail pane (delete hidden for Me)
- Unit tests for rename view model

**Desktop (`ledger-live-desktop`)**
- Redux ports for `renameContact` / `deleteContact`
- `useContactDetailEditDeleteAdapter` + pane adapter wiring
- Mount rename, delete, and signer dialogs in `ContactsView`
- English i18n for edit/delete/signer copy
- On delete success → returns to Me contact detail

### Behaviour

| Contact | Edit | Delete |
|---------|------|--------|
| Me | ✅ Rename name | ❌ Hidden |
| Saved contact | ✅ | ✅ + confirmation dialog |

- Contact **with addresses** → mocked signer dialog before rename
- Contact **without addresses** → rename dialog opens directly

https://github.com/user-attachments/assets/ece5302d-c8f8-4c69-9511-9d6b8702c8d5

### 🔗 Context

[LIVE-33943](https://ledgerhq.atlassian.net/browse/LIVE-33943)

[LIVE-33943]: https://ledgerhq.atlassian.net/browse/LIVE-33943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #20321, opened 2026-08-03
[LWDM] feat(contacts): block duplicate contact names (LIVE-35066)

deepyjr:
### ✅ Checklist

- [x] **Covered by automatic tests.** _Desktop and Mobile integration commands are blocked in this workspace before execution by generated dependencies unavailable locally._
  - Contacts cannot be created with an existing normalized name.
  - Desktop and Mobile display the same validation message before any creation action.

### 📝 Description

The shared Contacts domain now recognizes duplicate names after trimming, NFC normalization, and case folding.

The Add Contact flow reads existing Contacts, disables confirmation for duplicates, and validates again before calling the creation port. Both platforms provide the English duplicate-name message.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-35066](https://ledgerhq.atlassian.net/browse/LIVE-35066)

---

[LIVE-35066]: https://ledgerhq.atlassian.net/browse/LIVE-35066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Contacts/__integrations__/Contacts.integration.test.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Contacts/__integrations__/Contacts.integration.test.tsx
@@ -281,6 +281,25 @@ describe("Contacts integration", () => {
     });
   });
 
+  it("should block a duplicate contact name and allow a unique replacement", async () => {
+    const me = mockMeContact();
+    const ada = mockContact({ id: "contact-ada", name: "Ada" });
+    const { user } = renderContactsScreen({ contacts: { contacts: [me, ada] } });
+
+    await user.click(screen.getByTestId("contacts-add-contact"));
+    const input = screen.getByTestId("contacts-add-contact-name-input");
+
+    fireEvent.change(input, { target: { value: " ada " } });
+
+    expect(screen.getByText("This contact name is already in use.")).toBeVisible();
+    expect(screen.getByTestId("contacts-add-contact-save")).toBeDisabled();
+
+    fireEvent.change(input, { target: { value: "Ben" } });
+
+    expect(screen.queryByText("This contact name is already in use.")).not.toBeInTheDocument();
+    expect(screen.getByTestId("contacts-add-contact-save")).toBeEnabled();
+  });
+
   it("should filter saved contacts when searching", async () => {
     const { user } = renderContactsScreen(populatedContactsPageState);
 
--- a/domain/entity/contact/src/validation.ts
+++ b/domain/entity/contact/src/validation.ts
@@ -1,6 +1,7 @@
 import {
   ContactAddressLabelTooLongError,
   DuplicateContactAddressLabelError,
+  DuplicateContactNameError,
   InvalidContactAddressLabelError,
   InvalidContactNameError,
 } from "./errors";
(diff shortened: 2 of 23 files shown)


domain/entity/contact/src/validation.ts
```diff
@@ -13,29 +15,55 @@ import type { ContactAddressLabel, ContactName } from "./types";
...
+
+  if (!parsed.success) {
+    return INVALID_CONTACT_NAME_ERROR_NAME;
+  }
+
+  return existingNames.some(
+    name => normalizeContactNameForComparison(name) === normalizeContactNameForComparison(parsed.data),
+  )
+    ? DUPLICATE_CONTACT_NAME_ERROR_NAME
+    : null;
```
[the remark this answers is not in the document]
deepyjr: Fixed in d11a1993b19. The normalized draft name is now computed once before checking the existing names, matching the address-label validation pattern.

apps/ledger-live-desktop/src/mvvm/features/Contacts/__integrations__/Contacts.integration.test.tsx
```diff
@@ -298,6 +298,35 @@ describe("Contacts integration", () => {
...
+      <MemoryRouter initialEntries={["/contacts"]}>
+        <Routes>
+          <Route path="/contacts" element={<ContactsScreen />} />
+        </Routes>
+      </MemoryRouter>,
+      {
+        skipRouter: true,
+        initialState: contactsPageInitialState({ contacts: { contacts: [me, ada] } }),
+      },
+    );
```
mcayuelas-ledger: on @claudiiafg 's PR she will add an helper for that to not duplicate it

deepyjr: Rebased onto develop after #20281 merged. This test now uses the shared `renderContactsScreen` helper instead of duplicating the render setup.

domain/entity/contact/src/validation.ts
```diff
@@ -13,29 +15,56 @@ import type { ContactAddressLabel, ContactName } from "./types";
...
+  existingNames: readonly ContactName[] = [],
+): ContactName {
   const parsed = ContactNameSchema.safeParse(draftName.trim());
 
   if (!parsed.success) {
     throw new InvalidContactNameError();
   }
 
+  if (getContactNameValidationError(parsed.data, existingNames) !== null) {
+    throw new DuplicateContactNameError();
```
gre-ledger, line 110: I find this code conceptually weird or maybe getContactNameValidationError is not a good name, but i'm not sure why we assume that getContactNameValidationError returning non null is exactly a `DuplicateContactNameError`. Shouldn't you let a validation function throwing an error OR returning a `{ success:bool, error: Error | null }` kind of shape? 

why bother returning the error name if at the end all falls into DuplicateContactNameError.

Generally speaking, I also find it a bit weird in this package that we have errors.ts that not only return classes but also constants. While I understand it can be used to code share the label for `error.name === NAME` matching it's probably the only case i would see such label used. otherwise why bother with the class if we extrapolate 🤔 

anyway, i reco to look at other errors.ts in the project, because the current pattern of this lib is a bit diverging (not saying it's not better, but i'm more thinking general consistency)

deepyjr: @gre-ledger Thanks for the feedback. I agree the current error handling may not match the patterns used elsewhere in the codebase. I think I may have misunderstood the intended approach, so I’ll look at similar errors.ts implementations and validation flows in other folders.

mcayuelas-ledger (dismissed):

mcayuelas-ledger (approved):

gre-ledger (approved):

PR #20114, opened 2026-07-28
[LWDM] feat(coin-bitcoin): add Ironwood pool support [LIVE-32889]

semeano:
### 📝 Description

Extends the Zcash adapter in @ledgerhq/coin-bitcoin to track and spend
Ironwood pool notes alongside Sapling and Orchard. Includes balance
aggregation, note collection, ZIP-317 fee modelling, IPC plumbing for
buildIronwoodTransaction (PCZT flow), nullifier tracking, serialization
round-trips, and a guarded signOperation stub that errors until
finalizeIronwoodTransaction lands in @ledgerhq/zcash-utils. Flips
orchardOnly to false now that the NU6.3 NAPI is merged.

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/LIVE-32889

PR #20328, opened 2026-08-03
[LWDM] feat(zcash): add optimistic note reservation [LIVE-34949]

semeano:
### 📝 Description

Selected Ironwood notes are now reserved in a session-scoped in-memory store immediately after buildSignedOperation, so back-to-back shielded sends before the first confirms will select disjoint note sets. The store is reconciled on every shielded sync pass via reduceShieldedSyncResult.

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/LIVE-34949

### DIFF
--- a/libs/coin-modules/coin-zcash/src/bridge/note-reservation.ts
+++ b/libs/coin-modules/coin-zcash/src/bridge/note-reservation.ts
@@ -1,10 +1,32 @@
+import { getEnv } from "@ledgerhq/live-env";
+
 /**
  * Session-scoped, in-memory reservation store for shielded note nullifiers.
  *
- * After a shielded send is built and signed, the selected notes are reserved
- * here so that a concurrent or immediately-following send cannot select the
- * same notes. The store is reconciled against on-chain reality on every
- * shielded sync pass via reconcileReservations.
+ * After a shielded send is built and signed, the notes it spends are reserved
+ * under the hash of the operation spending them, so that a concurrent or
+ * immediately-following send cannot select the same notes. Selection filters the
+ * account's reserved nullifiers out of the spendable pool (prepareTransaction,
+ * estimateMaxSpendable).
+ *
+ * A reservation is released only on evidence about its own operation:
+ *
+ * - the scan reports its notes spent on-chain — `releaseConfirmedNullifiers`,
+ *   from the shielded sync reducer;
+ * - its operation has a confirmed counterpart in `account.operations`, the same
+ *   signal `reconcileConfirmedPendingOperations` retires the optimistic
+ *   operation on — `releaseRetiredReservations`, from `postSync`;
+ * - it outlived OPERATION_OPTIMISTIC_RETENTION, measured from signing time just
+ *   as `shouldRetainPendingOperation` measures the optimistic operation's own
+ *   date: a broadcast that never confirmed stops holding notes at the moment
+ *   Ledger Wallet itself stops believing in the operation — same function;
+ * - broadcast failed, so nothing can confirm — `releaseReservation`, from the
+ *   bridge's `broadcast`.
+ *
+ * A sync reaching the chain tip is deliberately not one of them: catching up to
+ * the tip a run started from is the outcome of nearly every poll of an account
+ * that is not backlogged, and says nothing about a spend in flight, which needs
+ * a block (~75s) at the very least to confirm.
  *
  * Scope: single-user desktop/mobile session. The store is module-level and
  * therefore shared for the lifetime of the process. It is not persisted across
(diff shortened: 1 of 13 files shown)


libs/coin-modules/coin-zcash/src/bridge/note-reservation.ts
```diff
@@ -0,0 +1,88 @@
...
+  syncComplete: boolean,
+): void {
+  const set = _reserved.get(accountId);
+  if (!set) return;
+
+  for (const nf of confirmedSpentNullifiers) {
+    set.delete(nf);
+  }
+
+  if (syncComplete) {
```
cted-ledger: **🔴 Critical: reservation is released on ordinary sync completion, not on spend confirmation (confidence: 92)**

`remainingBlocks` is computed in `network/engine.ts` (`syncShielded`) as `endHeight - chunkEnd`, where `endHeight` is fetched once via `getChainTip()` at the *start* of that sync run. It hits 0 as soon as the scan catches up to the tip it started at — the normal outcome of essentially every periodic sync poll once an account isn't backlogged, not a rare "fully synced" event. `syncComplete` here therefore fires far more often, and far sooner, than "this spend has confirmed on-chain" (≥ 1 Zcash block, ~75s, often more).

The header comment above (`note-reservation.ts:17-20`) concedes exactly this ("clears the reservation prematurely") but frames it as an edge case — given the `remainingBlocks` semantics, it looks like the common case for any account that syncs periodically while a send is in flight.

The validated spec (`IW-COIN-03`) explicitly required reconciliation to "hang off the existing optimistic-operation lifecycle rather than define its own," naming `reconcileConfirmedPendingOperations`/`postSync` as the mechanism to reuse so a reservation is released only when its specific operation is confirmed retired. This `set.clear()` on `syncComplete` is a different, self-defined rule ("this sync run reached its own start-of-run tip") that doesn't track per-operation confirmation — which is the root cause of the premature-release risk above. The acceptance-criteria test ("two back-to-back sends... without an intervening sync") doesn't exercise this path, so the green suite here isn't evidence the double-spend race is actually closed.

**Suggestion**: tie release to the confirmed-operation lifecycle the spec points at (or equivalent), not to "sync reached tip," before merging — this is shielded-fund signing logic.

libs/coin-modules/coin-zcash/src/bridge/signOperation.ts
```diff
@@ -123,6 +124,12 @@ export const buildSignOperation =
         );
 
         const fee = new BigNumber(buildResult.feeZat);
+
+        const ironwoodNullifiers = transaction.selectedNotes.map(n => n.nullifier);
```
cted-ledger, line 128: **🟡 Warning: reservation (and the field below) also captures Orchard dev/test notes, against an explicit spec constraint (confidence: 85)**

`IW-COIN-03`'s Constraints: "Ironwood pool only. Do NOT retrofit the Orchard dev/test path; do NOT introduce dual-pool logic." This reserves `transaction.selectedNotes` under the name `ironwoodNullifiers` regardless of `transferType`. Per `prepareTransaction.ts`, `"shielded"`/`"shielded-to-transparent"` populate `selectedNotes` from **Orchard** notes (`collectSpendableNotes`), not Ironwood ones — the new `signOperation.test.ts` case even exercises this with `transferType: "shielded"` and confirms Orchard nullifiers land in the same store.

Currently inert (the only two readers of `getReservedNullifiers`, in `estimateMaxSpendable.ts`/`prepareTransaction.ts`, filter exclusively against Ironwood notes), but it's a real constraint violation and a misleading variable name — any future Ironwood-agnostic consumer of this store would silently inherit Orchard dedup nobody asked for.

**Suggestion**: gate this on the Ironwood transfer types specifically, or rename to reflect that it's intentionally pool-agnostic if that's the actual intent.

libs/coin-modules/coin-zcash/src/bridge/signOperation.ts
```diff
@@ -137,6 +144,7 @@ export const buildSignOperation =
           date: new Date(),
           extra: {
             zcashShielded: true,
+            ...(ironwoodNullifiers.length > 0 && { shieldedNullifiers: ironwoodNullifiers }),
```
cted-ledger, line 147: **🟡 Warning: `shieldedNullifiers` has no reader anywhere (confidence: 84)**

The spec asks this to "mirror the existing transparent `inputRefs` / pending-spent model so the `broadcast()` double-spend guard and conflict-dedup consumers work for Ironwood notes." `inputRefs`'s transparent analog is read by `coin-bitcoin`'s `broadcast.ts` for exactly that guard. A repo-wide search for `shieldedNullifiers` outside this PR's own tests finds no reader — it's write-only. The actual filtering happens entirely through the in-memory `note-reservation.ts` store, disconnected from this persisted field.

Might be an intentional placeholder for a follow-up (e.g. surviving a restart), but as shipped it doesn't yet deliver the "mirrors `inputRefs`" requirement's actual payoff. Worth a one-line note in the PR description either way.

libs/coin-modules/coin-zcash/src/bridge/note-reservation.ts
```diff
@@ -0,0 +1,88 @@
...
+ * The transparent analog is inputRefs written into operation.extra; the
+ * reservation store provides the same intra-session dedup for shielded spends.
+ *
+ * Risk note: a complete sync (remainingBlocks === 0) triggered immediately after
+ * signOperation but before broadcast confirmation clears the reservation
+ * prematurely. The mempool TTL and node-side nullifier rejection provide a safety
+ * net; this store is an intra-session dedup helper, not the sole double-spend guard.
+ */
+
+const _reserved = new Map<string, Set<string>>();
```
cted-ledger: **🟢 Suggestion: unbounded growth risk in the reservation `Map`, low severity (confidence: 80)**

`reconcileReservations` only drops an account's entry on `syncComplete` or once its set is already empty. An account that syncs only while permanently backlogged (`remainingBlocks > 0` throughout) never gets its entry released, and the outer `Map` accumulates one entry per account touched over the process lifetime. Minor given the typical number of Zcash accounts per user — flagging for awareness, not a blocker.

cted-ledger (commented):
## Code Review: `feat(zcash): add optimistic note reservation [LIVE-34949]`

Reviewed against the validated spec (`IW-COIN-03-ironwood-note-reservation.md`), its blocking dependency spec (`IW-COIN-01`), and JIRA LIVE-34949. Inline comments below cover the line-level findings; one finding is scope-level and doesn't anchor to a line, so it's here:

### 🔴 Critical — fix lands outside the ticket's and spec's stated repo path (confidence: 88)

Both JIRA LIVE-34949 ("Target repo: ledger-live, coin-bitcoin chain-adapters/zcash") and the validated spec `IW-COIN-03` (Data sources: `chain-adapters/zcash/index.ts`, `chain-adapters/zcash/coin-selection.ts`, `chain-adapters/zcash/operations.ts`) target `libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/`. This PR touches only `libs/coin-modules/coin-zcash/src/bridge/` — the newer standalone module, which coexists with `coin-bitcoin`'s chain-adapter behind the `zcashShielded` flag (`libs/ledger-live-common/src/bridge/zcashRouting.ts`, off by default). `coin-bitcoin` has its own, separately-added Ironwood send path (#20114) that would carry the identical double-spend race described in this ticket, and nothing here touches it.

This silently narrows LIVE-34949's own stated scope without updating the ticket or the spec. Depending on which package actually serves production Zcash accounts (unresolved from this PR/spec alone), there's a real risk the production-serving path ships without this fix. Please confirm which path is production-bound and either link a follow-up ticket for `coin-bitcoin` or correct LIVE-34949's scope description.

---
*2 critical, 2 warning, 1 suggestion findings (repo review convention per `.agents/agents/code-reviewer.md`). See inline comments for the rest.*

libs/coin-modules/coin-zcash/src/bridge/note-reservation.ts
semeano: Fixed.

libs/coin-modules/coin-zcash/src/bridge/signOperation.ts
semeano: I don't think this one holds — `selectedNotes` is never populated from Orchard notes. Maybe you were looking into old coin-bitcoin chain adapter?

`prepareTransaction` sources the pool via `collectIronwoodSpendableNotes`, not
`collectSpendableNotes`, and that was already the case before this PR (the only change
here is the `reserved` filter).

libs/coin-modules/coin-zcash/src/bridge/signOperation.ts
semeano: Fixed

libs/coin-modules/coin-zcash/src/bridge/note-reservation.ts
semeano: This comment is against reconcileReservations from the first commit of the branch; it was removed in 733bf4df849 ("tie note reservation release to operation lifecycle"). There's no sync-completion-gated release any more — reaching the tip is deliberately not treated as evidence about a spend in flight.

semeano:
Everything was addressed except the following comments:

___

> ### 🔴 Critical — fix lands outside the ticket's and spec's stated repo path (confidence: 88)

The conclusion doesn't hold anymore though, for a concrete reason: the coin-bitcoin Ironwood send path no longer exists.

___

> **🟡 Warning: reservation (and the field below) also captures Orchard dev/test notes, against an explicit spec constraint (confidence: 85)**

I don't think this one holds — `selectedNotes` is never populated from Orchard notes. Maybe you were looking into old coin-bitcoin chain adapter?

`prepareTransaction` sources the pool via `collectIronwoodSpendableNotes`, not
`collectSpendableNotes`, and that was already the case before this PR (the only change
here is the `reserved` filter).

___

> **🟢 Suggestion: unbounded growth risk in the reservation `Map`, low severity (confidence: 80)**

This comment is against reconcileReservations from the first commit of the branch; it was removed in 733bf4df849 ("tie note reservation release to operation lifecycle"). There's no sync-completion-gated release any more — reaching the tip is deliberately not treated as evidence about a spend in flight.

may01 (approved):

PR #20332, opened 2026-08-03
[LWDM] fix(coin-hedera): support multi-transfers from single tx (LIVE-24949)

vtaranushenko-ext-ledger:
### 📝 Description

This PR fixes Hedera transactions parsing to cover multi-transfers that can be done with SDK. It also adjusts the sync cursor to skip fetching data that we already have.

https://hashscan.io/mainnet/transaction/1760510879.300860000 was incomplete before:

| hedera-before-1 | hedera-before-2 |
|---|--|

After:

| hedera-after-1 | hedera-after-2 | hedera-after-3 | hedera-after-4 |
|---|--|--|--|

### 🔗 Context

https://ledgerhq.atlassian.net/browse/LIVE-24949

### DIFF
--- a/.changeset/shy-lobsters-crash.md
+++ b/.changeset/shy-lobsters-crash.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-hedera": minor
+---
+
+Fix Hedera transactions that pack several fund movements being reported as a single operation, and fix the sync cursor skipping or duplicating operations.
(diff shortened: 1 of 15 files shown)


mdomanski-ext-ledger:
Closing it for now until we are sure we won't break anything on Framework level with duplicated fees (to be confirmed with Julien, maybe it's wise to wait for `getTransactions` refactor)

PR #20341, opened 2026-08-03
feat(ddd): split api backend access from use case (LIVE-35298)

ysitbon:
## Context

[LIVE-35298](https://ledgerhq.atlassian.net/browse/LIVE-35298) — epic [LIVE-35297](https://ledgerhq.atlassian.net/browse/LIVE-35297)

`domain/api/*` packages are organised **per use case**, and each one called `createApi(...)` with its endpoints inline. `createApi` bundles *reaching a backend* (base URL, headers, retry, auth) together with *what you ask it for*, so organising by use case forces the reaching half to be re-declared per use case — and two packages hitting the same backend end up with two store slices, two caches and two middlewares for one service.

That was already the case: `@domain/api-altcoins-sentiment` and `@domain/api-market-sentiment` both call CoinMarketCap on the same `CMC_API_URL`, duplicating a near-identical base query, an identical `extraArgument` schema under two names, and identical cache-timing constants.

## What changed

Per the [RTK Query code-splitting guide](https://redux-toolkit.js.org/rtk-query/usage/code-splitting#injecting-endpoints), the two halves are split. **`@shared/api-services`** holds one endpoint-less `createApi({ endpoints: () => ({}) })` per backend, one directory each:

```
shared/api-services/src/services/<service>/
  api.ts  api.test.ts  constants.ts  schema.ts  types.ts  index.ts
```

| Service | Reducer path | Injectors |
|---|---|---|
| `cal` | `calApi` | `currency-token` |
| `coinmarketcap` | `coinMarketCapApi` | `altcoins-sentiment`, `market-sentiment` |
| `countervalues` | `countervaluesApi` | `currency-fiat` |
| `push-devices` | `pushDevicesApi` | `push-devices` |

Use-case packages add to that same api — `injectEndpoints` for endpoints, `enhanceEndpoints({ addTagTypes })` for their own cache tags. Both mutate and return the same object, so one reducer, one middleware and one cache serve every use case:

```ts
export const FEAR_AND_GREED_TAGS = ["FearAndGreedLatest"] as const;

export const marketSentimentApi = coinMarketCapApi
  .enhanceEndpoints({ addTagTypes: FEAR_AND_GREED_TAGS })
  .injectEndpoints({ endpoints: build => ({ /* … */ }) });
```

Both app registries now register the service apis, so `rtkQueryApi.ts` reads as a list of the backends the app talks to rather than of its use cases (11 entries → 10, since the CoinMarketCap pair collapsed).

## Where the seam falls

| Concern | Owner |
|---|---|
| Base URL, headers, retry, auth, `extraArgument`, `reducerPath` | `@shared/api-services` |
| Endpoints, wire schemas, transforms, **cache tags**, hooks, persistence | `@domain/api-*` |
| Polling intervals | the app feature's `constants.ts` |

Two placements worth calling out, both corrections made during review:

- **Cache tags belong to the use case.** `injectEndpoints` doesn't accept `tagTypes`, which suggests the shared api must declare every tag upfront. It doesn't — `enhanceEndpoints({ addTagTypes })` widens the tag union in place, so a tag stays next to the endpoints that provide it and adding a use case never means editing a shared file. Verified against RTK 2.11.2 with a throwaway probe: same-object return, two callers can each add tags, and invalidation fires end-to-end.
- **Polling intervals belong to the app.** `pollingInterval` is a view refresh decision, so it now lives in each feature's `constants.ts`, following the existing `ASSETS_PRICE_REFRESH_INTERVAL_MS` convention. Only `keepUnusedDataFor` stays domain-side, as a private const.

Because the shared package ended up holding no domain vocabulary at all — no entity schemas, no tags, no business terms, and zero workspace dependencies — it sits in `shared/` rather than `domain/`. Had the tags stayed it couldn't, since `shared/*` must not import `domain/*`.

## Notes for review

- **No endpoint behaviour changes.** Bodies, names, tags, transforms and cache settings were relocated verbatim.
- **The split is unconditional**, not a thing to reach for once a second use case appears — CAL, Countervalues and Push Devices each have one injector today and still go through it, so the second is a one-line addition rather than a migration.
- **`extraArgument` builders kept their names** (`calApiExtra`, `cvsApiExtra`, `pushDevicesApiExtra`), so call sites are unchanged and only import paths moved. `pushDevicesApiExtra` is re-exported from its use-case package, so neither `configureStore.ts` changed for it.
- **Reducer paths renamed** after their backend. Safe: the CAL persisted blob is a version-pinned `{ version, tokens, hashes }` format containing no reducer path, stored under the storage key `cryptoAssets`; slice access is a dynamic `state[api.reducerPath]` read and restore goes through `util.upsertQueryEntries`. Nothing is orphaned, no cache goes cold.
- **A tag-less api has a narrower reducer state type.** `WithCryptoAssetsApi` was typed on the injected `cryptoAssetsApi`, whose state gained `provided.tags.Tokens`, and stopped accepting the app's `State` (which comes from the tag-less `calApi`). Typecheck caught it at `db.ts:61`; it's now typed on `calApi`. Any helper typed on an injected reference has this problem — documented in both READMEs.
- **`DBSave.ts` now uses the dynamic reducer path.** It read `state.cryptoAssetsApi` as a property access, which a string-literal grep missed and typecheck caught — a literal-only search would have shipped a silent break in CAL cache persistence.
- **Injection is a module-level side effect.** An endpoint exists only once its use-case module is evaluated as a *value* import. Apps register the service api; features import hooks from the use-case package.
- **Naming:** an earlier revision called these `@domain/api-transport-*`. Dropped — "transport" is overwhelmingly a device term here (449 files touch `@ledgerhq/hw-transport*`, and `hw-transport-http` already means "device over HTTP"). One `api-services` package removes the need for any qualifier.

## `@domain/api-pay-card` is a holdout, not an exception

It's the only backend still declaring its own `createApi`, because its base query resolves mock responses keyed by endpoint URL and typed from its own wire schemas — it has to be made transport-only first. **Do not copy it.** The docs say so explicitly; the Pay Card team will migrate it.

## Verification

- `pnpm nx run-many -t typecheck` — 118 projects, clean
- `pnpm nx run-many -t lint` — 146 projects, 0 errors
- Tests across all affected packages, including `@domain/api-currency-token` (89, full persistence suite) and `@features/platform-currencies` (47, covering `buildCryptoAssetsStore`'s direct `.endpoints.*.initiate` dispatch and `buildStandaloneCryptoAssetsStore`'s CAL-only partial config)
- Both apps typecheck; 228 desktop reducer/store/middleware suites, 8 mobile store/registry/`DBSave` suites (the two persistence writers), and the 14 mobile FearAndGreed/AltcoinSeason integration tests
- New `rtkQueryApi.test.ts` in each app pins the exact expected reducer-path set, since a wrong import or dropped api changes the store shape without failing typecheck

**Still needs a runtime smoke test** — injection being a module-level side effect means a missing endpoint is invisible to typecheck. Worth checking the altcoin season index, fear & greed, CAL token lookups and fiat loading on desktop.

## Follow-ups

- `@domain/api-aggregated-assets` (on develop) is a placeholder for **DADA**, currently `assetsDataApi` in live-common. When it migrates, its base belongs in `@shared/api-services` as `src/services/dada/`.
- `marketApi`, `counterValuesApi` and `ofacGeoBlockApi` in live-common all share `LEDGER_COUNTERVALUES_API` with `countervaluesApi` — they should inject rather than add a fourth `createApi` against the same backend.
- `ledgerClientVersion` is declared by both CAL and Push Devices with different validators (`.min(1)` vs `.trim().min(1)`). Both moved verbatim; unifying is a behaviour change for its own ticket.

[LIVE-35298]: https://ledgerhq.atlassian.net/browse/LIVE-35298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-35297]: https://ledgerhq.atlassian.net/browse/LIVE-35297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/state-manager/configureStore.ts
```diff
@@ -59,10 +58,7 @@ const customCreateStore = ({
               ...cvsApiExtra({
                 countervaluesServiceUrl: getEnv("LEDGER_COUNTERVALUES_API"),
               }),
-              ...marketSentimentApiExtra({
-                coinMarketCapApiUrl: getEnv("CMC_API_URL"),
-              }),
-              ...altcoinsSentimentApiExtra({
+              ...coinMarketCapApiExtra({
                 coinMarketCapApiUrl: getEnv("CMC_API_URL"),
```
gre-ledger, line 66: nice gain here, so we don't have to repeat injecting the url conf

gre-ledger (dismissed):

ysitbon:
## 🧪 Triggered test workflows

Manually dispatched on `refactor/injectable-endpoints` (commit 2188c4d5e4), scoped to the diff:

| Workflow | Scope | Run |
|---|---|---|
| [Desktop] E2E Only | full suite, Speculos nanoSP | [run 30856085543](https://github.com/LedgerHQ/ledger-live/actions/runs/30856085543) |
| [Mobile] E2E Only | full suite, Android only, nanoX | [run 30856087715](https://github.com/LedgerHQ/ledger-live/actions/runs/30856087715) |

**Coin Tester deliberately not dispatched** — the diff touches no `libs/` file, so no coin module, bridge or family logic is affected and its auto-detection would select nothing.

**Why full suite rather than a tag filter:** this rewires both apps' RTK Query store root, CAL cache persistence and the CoinMarketCap-backed widgets. A store misconfiguration or an unresolved `injectEndpoints` is invisible to typecheck and fails broadly rather than per-coin, so `@bitcoin`/`@family-evm`-style filters would target the wrong surface.

**Why Android only on mobile:** every mobile change is JS-level (store registry, `DBSave.ts`, feature constants). Nothing is platform-specific, so an iOS leg would duplicate the signal at double the cost.

mcayuelas-ledger (approved):

Sebastien-Dav1d (approved):

PR #20345, opened 2026-08-03
[LWDM] chore(aggregated-assets): move domain layer behind shims (LIVE-35226)

LucasWerey:
### ✅ Checklist

- [x] **Covered by automatic tests.** _The characterization tests from LIVE-35224 are the gate: 18 suites, **216 tests, passing unmodified**. The moved packages add **135** of their own (84 api, 33 api-services, 13 interest-rate, 5 aggregated-asset)._
      - **Behaviour-neutral by design.** Zero consumer files change; all 82 consumers keep working through one-line re-export shims at the old paths.
      - One deliberate exception, added during review: the synthesised currency in `convertApiAssets` is now validated by `CryptoCurrencySchema` — see *Review round* below.
      - QA focus if anything is spot-checked: **Market** (price / 24h change badges), **Portfolio** (distribution), and the **asset/network selectors** — the surfaces fed by the relocated cache selectors. Plus the **Stocks** and **Stablecoins** sections, whose MSW fixtures were refactored (proven byte-identical).

### 📝 Description

**Problem.** `libs/ledger-live-common/src/dada-client` is being migrated into the DDD layers under [LIVE-35223](https://ledgerhq.atlassian.net/browse/LIVE-35223). The packages were scaffolded empty in [#20285](https://github.com/LedgerHQ/ledger-live/pull/20285); this PR moves the domain layer into them.

**Solution.** Relocate the entity types and the API layer, leaving shims behind so consumer retargeting is a separate, independently revertable step (LIVE-35228 / 35229 / 35230).

| Package | Now contains |
| --- | --- |
| `@domain/entity-aggregated-asset` | `CryptoAssetMetaSchema` + inferred `CryptoAssetMeta` |
| `@domain/entity-interest-rate` | `InterestRateSchema`, `ApyTypeSchema` + inferred types |
| `@domain/api-aggregated-assets` | `schema.ts` (wire), `types.ts`, `transforms.ts`, `requests.ts`, `accessors.ts`, `api.ts`, `errors.ts`, `pagination.ts`, `constants.ts`, `internals/`, `fixtures/` |
| `@shared/api-services` | `services/dada/` — the endpoint-less `createApi` base |

14 files under `dada-client` became one-line re-exports. `libs/ledger-live-common` gains the packages as workspace deps — it already declared 8 `@domain/*` deps, so this is the established pattern.

#### Decisions worth reviewing

- **The RTK base lives in `@shared/api-services`, and this package injects into it.** `services/dada/` owns `createApi({ reducerPath: DADA_REDUCER_PATH, tagTypes: [], endpoints: () => ({}) })`; the domain package does `dadaApi.enhanceEndpoints({ addTagTypes }).injectEndpoints(...)`. One reducer, one middleware, one cache slice per backend. Splitting the single `injectEndpoints` call into three per-use-case modules is [LIVE-35301](https://ledgerhq.atlassian.net/browse/LIVE-35301).
- **The entities are zod-first.** `domain/entity/README.md` mandates it: `schema.ts` holds the schema and the TS types are `z.infer<>`. Both entity packages follow the `entity/market-sentiment` layout.
- **`reducerPath` stays the literal `"assetsDataApi"`, pinned by a test.** `createCurrencyDataSelector` hand-scans that key as a string and three Storybook files preload it, so a rename would silently return `undefined` for every market and interest-rate lookup **with no type error**. Any rename belongs to LIVE-35301.
- **`NetworkInfo` and `CurrenciesOrder` land in the api package, not an entity.** A network *is* a chain, already modelled by `@domain/entity-currency-crypto`, so `NetworkInfo` stays a wire type resolving to that rather than duplicating the concept. `CurrenciesOrder` is `{ key, order, metaCurrencyIds }` — server sort metadata, not a business object. Agreed with @gre and Yoann; see [DADA DDD compliant](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7382761502/DADA+DDD+compliant).
- **`internals/` really is internal.** Nothing in it is exported from the barrel. `market.ts` lives there too: `dadaIdToMarketId()` and the market item type are **copied** from live-common rather than imported, because a `domain/*` package must not import legacy `libs/*`. The drift risk is documented in the file — `PartialMarketItemResponse` is `Partial<MarketItemResponse>`, so every field is optional and future divergence will **never** produce a type error.
- **`pagination.ts` is public and lives here, not in the platform layer.** It was moved out and back during review. `mergeAssetsDataPages` is consumed by the platform hooks, and it is structurally the same kind of helper as `errors.ts`/`parseError` — which already lives in `domain/api` — so consistency won.

#### Review round (@ysitbon)

| Feedback | Outcome |
| --- | --- |
| Possible loop in `collectAllByCategory` | **Confirmed, pre-existing** → [LIVE-35503](https://ledgerhq.atlassian.net/browse/LIVE-35503) |
| Use `CryptoCurrencySchema` instead of parsing the id alone | Done |
| Keep the market type internal to the package | Done — zero external consumers, see below |
| Use `getCryptoCurrencyById` for the mocks | Done |
| Mocks under a `fixtures/` folder | Done — export names unchanged |
| One utils file | Done for the five pure helpers |
| Every function tested | Done — five had **no** test anywhere |
| Duration const → Temporal, and out of `types.ts` | Moved to `constants.ts`; **Temporal declined**, see below |
| Brand `fetchAt` / `currencyId` with `@shared/schema-primitives` | Implemented, then **reverted** → LIVE-35232 |
| `fetchAssetsPage` internal, *"à voir la pertinence de l'implém"* | Move done; the design half → LIVE-35301 |

**The market type was fully internalisable.** I had argued it wasn't, citing 7 consumers of `dadaIdToMarketId` and 5 of `PartialMarketItemResponse` — those counts were for the *legacy live-common* declarations. **Zero** files import either from `@domain/api-aggregated-assets`, so dropping the barrel export cost nothing. Worth flagging the real finding: `dadaIdToMarketId` was **duplicated, not moved** — an identical copy remains at `market/utils/index.ts:129`, and only that copy had a test. De-duplicating is retarget work.

**Branding was reverted deliberately.** `DateTimeIsoSchema` and the crypto-or-token id union are correct, but branding them broke `tsc` in three characterization fixtures, which had to be rebuilt through `InterestRateSchema.parse(...)` to compile. Those fixtures are the epic's safety net and must pass unmodified, so the branding moved to LIVE-35232 where consumer churn is already expected. Note for whoever picks it up: verify with CI's command (`tsc --noEmit -p src/tsconfig.json --customConditions node`), because the root `tsconfig.json` reports zero errors on code that fails CI.

**Temporal was declined, with evidence.** V8 in node 24.14 has it behind `--harmony-temporal`, default off and marked *in progress / experimental*; `typeof Temporal` is `undefined` without the flag, and there is no `temporal-polyfill` anywhere in the monorepo. Mobile runs Hermes and desktop runs Electron — neither exposes it either. It would mean a polyfill shipped into the mobile bundle, and since `keepUnusedDataFor?: number` takes plain seconds, `Temporal.Duration.from({ days: 1 }).total({ unit: "seconds" })` just returns `86400`. Written as `24 * 60 * 60` in `constants.ts` instead. Temporal's real home here is parsing `fetchAt` — `DateTimeIsoSchema`'s own doc comment already anticipates `Temporal.Instant.from()`.

**The fetch helpers were the cosmetic half.** `resolveBaseUrl` and `fetchAssetsPage` moved to
`internals/` — nothing outside the package used them. But both exist only because the shared base
is `fetchBaseQuery({ baseUrl: "" })`, so every endpoint resolves its own absolute url and the
fan-out endpoints hand-roll `fetch`. RTK passes `baseQuery` as `queryFn`'s fourth argument, so
that is fixable, and the bypass is what forces `assertDadaApiUrl` to exist, drops RTK's
`AbortSignal`, and makes a DADA 5xx surface as `FETCH_ERROR` instead of a numeric status — so
`isNetworkError()` reports true for a server error while the paginated endpoints classify the same
failure correctly. Latent today: the only consumer chain is `useChunkedAssetsData` →
`useAssetDistribution`, which never reads `errorInfo`. Now acceptance criteria on LIVE-35301.

#### The one behaviour change, stated plainly

`convertApiAssets` now builds the synthesised currency through `CryptoCurrencySchema.parse(...)` rather than validating only its id. `parse` and not `safeParse` is deliberate — it preserves the contract the characterization tests pin, where an unusable currency surfaces as a query error rather than vanishing.

It does widen what can fail a response: `CryptoCurrencySchema` requires `units.min(1)` while the wire type does not, so an asset arriving with `units: []` would now fail the whole response. That is recorded as an explicit acceptance criterion on [LIVE-35233](https://ledgerhq.atlassian.net/browse/LIVE-35233), which converts this path to per-item drop-and-count.

#### What was deliberately *not* touched

- The lenient conversion paths — unconvertible tokens are still silently dropped, and cryptos missing from the local CAL are still *synthesised* rather than dropped. That leniency is load-bearing and carries a comment saying so.
- `getChunkedAssetsData` still succeeds if **any** chunk resolves. Portfolio distribution depends on it.
- An asset with `id: ""` still fails the *entire* query. Pinned as current behaviour; fixed in LIVE-35233.
- Base URL resolution still calls `getEnv("DADA_API_*")` from `domain/api`, where `domain/api/README.md` assigns it to `@shared/api-services`. Not behaviour-neutral, so it is scoped to LIVE-35301.
- The apps still register the use-case api rather than `dadaApi`. Scoped to LIVE-35230.

#### Fixtures

The three mocks moved to `src/fixtures/`, keeping the `.mock.ts` suffix so knip's `**/*.mock.*` ignore still applies and the `./mock*` export names stay unchanged — no consumer moved.

- **Currencies come from the CAL registry.** These are transformed-shape `AssetsData`, so `cryptoOrTokenCurrencies` holds real entities — what `getCryptoCurrencyById` returns. The three hand-built ones had drifted: `injective` claimed `family: "injective"` with one unit where the registry says `"cosmos"` with two. `getCryptoCurrencyById` over `findCryptoCurrencyById` on purpose: it throws, which is what a fixture should do if the CAL drops an id.
- **`buildCategoryResponse` replaces the hand-written category fixtures**, so a new category is a list of assets rather than another file. Stablecoins collapses to a ticker list. Stocks keeps explicit slugs because they are **not** derivable from the tickers — `AAPLX` is `applex`, not `aaplx`. I caught that by serialising both fixtures before the refactor and diffing after; with explicit slugs the output is **byte-identical**, so the MSW handlers and the desktop Portfolio integration test see no change.

#### Verification

- **Characterization tests: 18 suites, 216 tests, unmodified.** Their stack traces resolve into `domain/api/aggregated-assets/src/transforms.ts`, which confirms they exercise the moved code through the shims rather than passing vacuously.
- 135 in-package tests. Five functions previously had **no** test anywhere in the repo: `resolveBaseUrl`, `fetchAssetsPage`, `allSettled`, `assertDadaApiUrl`, and this package's `dadaIdToMarketId`. `collectAllByCategory` was exercised only indirectly and now has direct tests. `assertDadaApiUrl` mattered most — it is the host guard for the endpoints that build their own `fetch` instead of going through `baseQuery`.
- `libs/ledger-live-common` typechecks with **0 errors** under CI's exact command; all five packages typecheck.
- knip clean in the package and in live-common. The lockfile diff is **purely additive** (51 insertions, 0 deletions) and `pnpm install --frozen-lockfile` succeeds.

#### Pre-existing issues encountered, not caused by this PR

- **Desktop jest cannot run locally** — `libs/ui/packages/react/lib` is unbuilt, so all 62 suites fail in `tests/jestSetup.js` on `@ledgerhq/react-ui/assets/fonts`. Verified consumers via desktop **typecheck** instead.
- **One desktop typecheck error at `dada-client/entities/selectorUtils.ts:34`** (untyped cache scan). That file is byte-identical to develop and imports only `@reduxjs/toolkit`, so it is pre-existing — and a fitting demonstration of why this code is guarded by tests rather than types.
- `libs/asset-detail` reports 118 implicit-any errors, all in unrelated live-common files. None mention this code.
- SonarCloud's quality gate fails on relocated code. The in-workflow `Sonar Cloud` job passes; the findings predate this PR.

### ❓ Context

- **JIRA**: [LIVE-35226](https://ledgerhq.atlassian.net/browse/LIVE-35226) (epic: [LIVE-35223](https://ledgerhq.atlassian.net/browse/LIVE-35223))
- **Architecture**: [DADA DDD compliant](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7382761502/DADA+DDD+compliant) (WXP)
- **Predecessors**: [#20276](https://github.com/LedgerHQ/ledger-live/pull/20276) (characterization tests) · [#20285](https://github.com/LedgerHQ/ledger-live/pull/20285) (scaffolding)
- **Filed from this review**: [LIVE-35503](https://ledgerhq.atlassian.net/browse/LIVE-35503) (unbounded category walk)
- **Next**: LIVE-35227 moves the platform layer (hooks, cache selectors, discovery)

---

[LIVE-35223]: https://ledgerhq.atlassian.net/browse/LIVE-35223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


domain/entity/interest-rate/src/schema.ts
```diff
@@ -0,0 +1,11 @@
+/** An interest rate attached to one currency. */
+export interface InterestRate {
+  /** Currency identifier */
+  currencyId: string;
+  /** Interest rate value */
```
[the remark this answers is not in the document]
LucasWerey: You're right — fixed in 22b1c04.

`domain/entity/README.md` is explicit: *"Use `zod` for schema definition"*, with `schema.ts` and `schema.test.ts` listed as required, and *"The schema is always required — it is the canonical data model and the primary reason the package exists."* Shipping plain interfaces diverged from that.

Now follows the `domain/entity/market-sentiment` layout: `schema.ts` holds the Zod schema, `types.ts` infers the type via `z.infer`, `schema.test.ts` covers validation, and the barrel re-exports both.

Worth noting **why this was safe to do inside a behaviour-neutral relocation**, since my initial reasoning for deferring it was wrong. I had assumed adopting Zod meant adopting branded types, which would force a `.parse()` at the boundary and break consumers building these shapes from raw JSON. But a plain `z.object()` without `.brand()` infers a structurally identical type — so the schema can exist and be exported without anything being obliged to parse yet. Confirmed: `libs/ledger-live-common` still typechecks with **0 errors** and the characterization tests pass **unmodified** (21 suites, 245 tests).

Applying `.parse()` at the api boundary remains LIVE-35232, which is where the drop-invalid-item semantics get decided.

Two loosenesses are deliberate and now documented *and tested* rather than tightened:

- `InterestRate.type` stays `z.string()` rather than the `ApyType` enum. DADA genuinely sends kinds outside `"NRR" | "APY" | "APR"` and `useInterestRatesByCurrencies` drops them — behaviour pinned in #20276. Narrowing here would claim a guarantee the wire doesn't give.
- `fetchAt` stays `z.string()` rather than `DateTimeIsoSchema`. I checked: it has **zero** production reads — every occurrence is a mock, fixture or Storybook story. Validating a field nobody consumes could only discard otherwise-good rates.

domain/entity/aggregated-asset/src/schema.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * An aggregated asset: one logical asset grouping several per-network currencies.
+ *
+ * `assetsIds` maps a network id to the currency id that represents this asset on that network,
+ * which is what makes it "aggregated" rather than a single currency.
```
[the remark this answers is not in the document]
LucasWerey: You're right — fixed in 22b1c04.

`domain/entity/README.md` is explicit: *"Use `zod` for schema definition"*, with `schema.ts` and `schema.test.ts` listed as required, and *"The schema is always required — it is the canonical data model and the primary reason the package exists."* Shipping plain interfaces diverged from that.

Now follows the `domain/entity/market-sentiment` layout: `schema.ts` holds the Zod schema, `types.ts` infers the type via `z.infer`, `schema.test.ts` covers validation, and the barrel re-exports both.

Worth noting **why this was safe to do inside a behaviour-neutral relocation**, since my initial reasoning for deferring it was wrong. I had assumed adopting Zod meant adopting branded types, which would force a `.parse()` at the boundary and break consumers building these shapes from raw JSON. But a plain `z.object()` without `.brand()` infers a structurally identical type — so the schema can exist and be exported without anything being obliged to parse yet. Confirmed: `libs/ledger-live-common` still typechecks with **0 errors** and the characterization tests pass **unmodified** (21 suites, 245 tests).

Applying `.parse()` at the api boundary remains LIVE-35232, which is where the drop-invalid-item semantics get decided.

Two loosenesses are deliberate and now documented *and tested* rather than tightened:

- `InterestRate.type` stays `z.string()` rather than the `ApyType` enum. DADA genuinely sends kinds outside `"NRR" | "APY" | "APR"` and `useInterestRatesByCurrencies` drops them — behaviour pinned in #20276. Narrowing here would claim a guarantee the wire doesn't give.
- `fetchAt` stays `z.string()` rather than `DateTimeIsoSchema`. I checked: it has **zero** production reads — every occurrence is a mock, fixture or Storybook story. Validating a field nobody consumes could only discard otherwise-good rates.

pnpm-lock.yaml
```diff
@@ -23081,7 +23121,7 @@ packages:
       metro-react-native-babel-preset: '*'
       react: 19.1.4
       react-dom: 19.1.4
-      webpack: ^5.89.0
+      webpack: '*'
     peerDependenciesMeta:
```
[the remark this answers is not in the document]
LucasWerey: Fixed in 22b1c04 — the lockfile diff is now **48 insertions, 0 deletions**, containing only the three new workspace importers.

You were right to keep pushing on this. I'd previously concluded the churn was unavoidable pnpm re-resolution triggered by the workspace set changing, and said so on an earlier PR. That was wrong. I tested it properly this time: reverting the two peer-range rewrites by hand and then running `pnpm install --frozen-lockfile` **passes**, which proves they were cosmetic rather than required.

The two reverted:

- `any-observable@0.3.0`: `rxjs: ^5.5.10` (pnpm had rewritten it to `'*'`)
- `@storybook/addon-react-native-web@0.0.29`: `webpack: ^5.89.0` (rewritten to `'*'`)

Neither package is touched by this PR, and `--frozen-lockfile` is what CI enforces, so keeping develop's values is both correct and reviewable.

For the record on the other half of your note: `packageManager` pins `pnpm@10.24.0` and that is the version used, so the rewrites weren't a version mismatch — pnpm just normalises those ranges opportunistically whenever it rewrites the file.

pnpm-lock.yaml
```diff
@@ -25425,7 +25465,7 @@ packages:
     resolution: {integrity: sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==}
     engines: {node: '>=6'}
     peerDependencies:
-      rxjs: ^5.5.10
+      rxjs: '*'
```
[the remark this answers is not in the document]
LucasWerey: Fixed in 22b1c04 — the lockfile diff is now **48 insertions, 0 deletions**, containing only the three new workspace importers.

You were right to keep pushing on this. I'd previously concluded the churn was unavoidable pnpm re-resolution triggered by the workspace set changing, and said so on an earlier PR. That was wrong. I tested it properly this time: reverting the two peer-range rewrites by hand and then running `pnpm install --frozen-lockfile` **passes**, which proves they were cosmetic rather than required.

The two reverted:

- `any-observable@0.3.0`: `rxjs: ^5.5.10` (pnpm had rewritten it to `'*'`)
- `@storybook/addon-react-native-web@0.0.29`: `webpack: ^5.89.0` (rewritten to `'*'`)

Neither package is touched by this PR, and `--frozen-lockfile` is what CI enforces, so keeping develop's values is both correct and reviewable.

For the record on the other half of your note: `packageManager` pins `pnpm@10.24.0` and that is the version used, so the rewrites weren't a version mismatch — pnpm just normalises those ranges opportunistically whenever it rewrites the file.

LucasWerey:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345) **Quality Gate failed**
> Failed conditions  [11.3% Coverage on New Code](https://sonarcloud.io/component_measures?id=LedgerHQ_ledger-live&pullRequest=20345&metric=new_coverage&view=list) (required ≥ 80%)  [3 New Code Smells](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345) (required ≤ 1)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Won't be fixed here as we are only moving things

domain/entity/interest-rate/src/schema.ts
```diff
@@ -0,0 +1,21 @@
...
+/** An interest rate attached to one currency. */
+export const InterestRateSchema = z.object({
+  /** Currency identifier */
+  currencyId: z.string(),
+  /** Interest rate value */
+  rate: z.number(),
+  /** Type of rate (NRR, APR, APY, etc.) — intentionally wider than ApyType, see above */
+  type: z.string(),
+  /** Timestamp when the rate was fetched */
+  fetchAt: z.string(),
```
ysitbon, line 20: I don't know the type but is it compatible with DatetimeIso from `@shared/schema-primitive` package?

domain/entity/interest-rate/src/schema.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * DADA sends values outside this set, so {@link InterestRateSchema} keeps `type` as a plain string
+ * and consumers narrow to this union, dropping anything unrecognised.
+ */
+export const ApyTypeSchema = z.enum(["NRR", "APY", "APR"]);
+
+/** An interest rate attached to one currency. */
+export const InterestRateSchema = z.object({
+  /** Currency identifier */
+  currencyId: z.string(),
```
ysitbon, line 14: I don't know the constraint but can't it be constrained to currency ids branded types or is it completly different types?

domain/entity/aggregated-asset/src/schema.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export const CryptoAssetMetaSchema = z.object({
+  /** Asset identifier */
+  id: z.string(),
+  /** Asset ticker symbol */
+  ticker: z.string(),
+  /** Asset display name */
+  name: z.string(),
+  /** Map of network IDs to their corresponding asset IDs */
+  assetsIds: z.record(z.string(), z.string()),
```
ysitbon, line 17: Keys can't be related to currency ids branded types too or different ids?

domain/api/aggregated-assets/src/api.ts
```diff
@@ -0,0 +1,292 @@
...
+  );
+}
+
+export function fetchAllAssetCurrencyIdsByCategory(queryArg: GetAssetsByCategoryParams) {
+  return collectAllByCategory(queryArg, data =>
+    Object.values(data.cryptoAssets).flatMap(meta => Object.values(meta.assetsIds)),
+  );
+}
+
+export const assetsDataApi = createApi({
```
ysitbon: use injectEndpoints. Empty api should be registered on @shared/api-services

domain/api/aggregated-assets/src/index.ts
ysitbon, line 1: This is a bit messy. It should be more something like export * from "file" all the rest you don't want to expose should go on internals files or dir.

domain/api/aggregated-assets/src/index.ts
```diff
@@ -1 +1,51 @@
...
+/*
+ * Consumed by the feature layer's page-merging and discovery helpers, so exported rather than
+ * kept internal despite living under internals/.
+ */
+export { chunkCurrencyIds } from "./internals/chunkCurrencyIds";
+export type { CurrencyIdChunks } from "./internals/chunkCurrencyIds";
+export { deepMergeCryptoAssets } from "./internals/deepMergeCryptoAssets";
+export { mergeAssetsDataPages } from "./internals/mergeAssetsDataPages";
+export { dadaIdToMarketId } from "./internals/market";
+export type { MarketItemResponse, PartialMarketItemResponse } from "./internals/market";
```
ysitbon: Files under internals should not be exposed

domain/api/aggregated-assets/src/api.ts
```diff
@@ -0,0 +1,292 @@
...
+function emptyAssetsData(): AssetsData {
+  return {
+    cryptoAssets: {},
+    networks: {},
+    cryptoOrTokenCurrencies: {},
+    interestRates: {},
+    markets: {},
+    currenciesOrder: { metaCurrencyIds: [], key: "", order: "" },
+  };
+}
```
ysitbon: Could go on internals and could be tested there

domain/api/aggregated-assets/src/api.ts
```diff
@@ -0,0 +1,292 @@
...
+    return { data: collected };
+  } catch (error) {
+    return {
+      error: {
+        status: "FETCH_ERROR",
+        error: error instanceof Error ? error.message : "Unknown error",
+      },
+    };
+  }
+}
```
ysitbon: Could go to internals too

domain/api/aggregated-assets/src/api.ts
```diff
@@ -0,0 +1,292 @@
...
+  return collectAllByCategory(queryArg, data =>
+    Object.values(data.cryptoAssets).map(a => a.ticker),
+  );
+}
+
+export function fetchAllAssetCurrencyIdsByCategory(queryArg: GetAssetsByCategoryParams) {
+  return collectAllByCategory(queryArg, data =>
+    Object.values(data.cryptoAssets).flatMap(meta => Object.values(meta.assetsIds)),
+  );
+}
```
ysitbon: Maybe better to move them into an `accessors.ts` file

domain/api/aggregated-assets/src/api.ts
LucasWerey: I've a dedicated Jira task for it: [LIVE-35301](https://ledgerhq.atlassian.net/browse/LIVE-35301) 
As the task here is just to move things

domain/entity/aggregated-asset/src/schema.ts
LucasWerey: Will do it in this task [LIVE-35232](https://ledgerhq.atlassian.net/browse/LIVE-35232)

domain/entity/interest-rate/src/schema.ts
LucasWerey: Done

domain/api/aggregated-assets/src/index.ts
LucasWerey: Fixed

domain/api/aggregated-assets/src/api.ts
LucasWerey: Done

domain/api/aggregated-assets/src/index.ts
LucasWerey: Done

domain/api/aggregated-assets/src/api.ts
LucasWerey: Done

domain/api/aggregated-assets/src/api.ts
LucasWerey: Done

domain/entity/interest-rate/src/schema.ts
LucasWerey: Done; I used an union of both brands rather than CryptoCurrencyIdSchema

domain/entity/interest-rate/src/schema.ts
LucasWerey: Correction on my "Done" above — I reverted it in 878ad8ff.

`DateTimeIsoSchema` *is* the right type (DADA sends RFC 3339, and all four timestamp shapes it produces pass the schema, 6-digit fractional seconds included). The problem was downstream: branding `fetchAt` broke `tsc` in three characterization fixtures, which had to be rebuilt through `InterestRateSchema.parse(...)` to compile.

Those fixtures are the safety net for the whole epic and the deal is that they pass **unmodified** — spending that on a nice-to-have inside a move-only PR is a bad trade, so it goes to [LIVE-35232](https://ledgerhq.atlassian.net/browse/LIVE-35232), which brands `assetsIds` and already expects to touch consumers.

Worth recording that the reason it looked cheap at first is that CI typechecks live-common with `-p src/tsconfig.json --customConditions node`, and the root config I checked first reported zero errors.

domain/entity/interest-rate/src/schema.ts
LucasWerey: Same story as the `fetchAt` thread — reverted in 878ad8ff, moved to [LIVE-35232](https://ledgerhq.atlassian.net/browse/LIVE-35232).

To actually answer the question though, since it changes what we implement there: it is **not** `CryptoCurrencyIdSchema`. DADA keys `interestRates` by both crypto ids (`ethereum`) and token ids (`ethereum/erc20/usd_tether`), so it needs `z.union([CryptoCurrencyIdSchema, TokenCurrencyIdSchema])`. That is the same union `assetsIds` needs on its *value* side, so the two are one piece of work — another reason to do them together rather than half here.

domain/api/aggregated-assets/src/api.ts
LucasWerey: Correcting myself here — this is **done**, not deferred. I pointed at LIVE-35301 before checking, and `@shared/api-services` already existed (its README even names this migration). Shipped in d4f207e1.

`shared/api-services/src/services/dada/` now holds the endpoint-less base with the frozen `reducerPath`, and `domain/api/aggregated-assets` does `dadaApi.enhanceEndpoints({ addTagTypes }).injectEndpoints(...)`. One reducer, one middleware, one cache slice — which `createCurrencyDataSelector` depends on, since it hand-scans `state.assetsDataApi.queries` by string.

Two things genuinely left, both needing consumer changes so out of a move-only PR:
- the base URL is still resolved in `domain/api` via `getEnv("DADA_API_*")`, where it should reach the service as `extraArgument`
- the apps still register the use-case api rather than `dadaApi`

LIVE-35301 is now scoped to splitting the single `injectEndpoints` call into three per-use-case modules, which is the part that actually needed its own revert.

LucasWerey:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345) **Quality Gate failed**
> Failed conditions  [27.5% Coverage on New Code](https://sonarcloud.io/component_measures?id=LedgerHQ_ledger-live&pullRequest=20345&metric=new_coverage&view=list) (required ≥ 80%)  [4 New Code Smells](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345) (required ≤ 1)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Won't be addressed in this PR

domain/api/aggregated-assets/src/fixtures/assetsData.mock.ts
ysitbon, line 1: If just for package test move it to a fixtures dir

ysitbon (dismissed):

ysitbon (approved):

LucasWerey:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345) **Quality Gate failed**
> Failed conditions  [66.3% Coverage on New Code](https://sonarcloud.io/component_measures?id=LedgerHQ_ledger-live&pullRequest=20345&metric=new_coverage&view=list) (required ≥ 80%)  [4 New Code Smells](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345) (required ≤ 1)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20345)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Won't fix now

mcayuelas-ledger (approved):

PR #20349, opened 2026-08-03
[LWDM] fix(zcash): restrict raw source tx override to V5

cted-ledger:
### 📝 Description

Signing a Zcash transparent transaction failed with "Something went wrong", no prompt ever reaching the device, whenever one of the selected UTXOs came from a V4 transaction. Reported in production as PROD-12599 and reproduced on `develop`.

**Previous behaviour:** `DmkSignerZcash` hands the raw source-transaction bytes to the signer kit through `serializedPreviousTransactionOverride`. That override exists for V5 sources, whose Orchard bundle the kit's own serialization strips — which would make the device commit to a wrong ZIP-244 txid. It was applied to every version, but the kit chunks those bytes expecting Ledger's internal serialization, whose V4 header carries a consensus branch id that the on-chain bytes do not have. It therefore read the input count four bytes late and threw while chunking that input, so the transaction died mid-flight, after any earlier V5 inputs had already been streamed. `mapError` then flattened the untagged error into a message-less `Error`, which is why the UI could only report "Something went wrong".

**Fix:** the override is now restricted to the versions that can carry a shielded bundle, V5 and above. A V4 source transaction goes back through the serialization path that has always handled it, its Sapling fields travelling in `extraData`. Untagged device-action errors keep their context instead of being flattened, so a failure inside a task names itself in the logs.

**Regression cover:** unit tests assert that the override is applied for V5, omitted for V4, and omitted when the version field is too short to read; a multi-input case mirrors the reported logs, with two V5 sources and one V4 among them, the V4 being the only one skipping the override. The error-mapping fallbacks are covered as well.

Verified on device: the account that reproduced the failure now signs.

This is one half of the Zcash signing problem. The other half — the device computing a wrong txid for a V5 source carrying an Orchard bundle, which surfaced as a 504 on broadcast — is fixed device-side in LedgerHQ/device-sdk-ts#1702 and will reach this repo through a `@ledgerhq/device-signer-kit-zcash` bump.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35215](https://ledgerhq.atlassian.net/browse/LIVE-35215), reported in production as [PROD-12599](https://ledgerhq.atlassian.net/browse/PROD-12599)
- **Related**: LedgerHQ/device-sdk-ts#1702 — device-side txid fix ([LIVE-35258](https://ledgerhq.atlassian.net/browse/LIVE-35258))

[PROD-12599]: https://ledgerhq.atlassian.net/browse/PROD-12599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-35215]: https://ledgerhq.atlassian.net/browse/LIVE-35215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/PROD-12599-zcash-v4-source-tx-override.md
+++ b/.changeset/PROD-12599-zcash-v4-source-tx-override.md
@@ -1,0 +1,14 @@
+---
+"@ledgerhq/live-signer-zcash": patch
+"live-mobile": patch 
+---
+
+Fix Zcash sends failing before the device prompt when the coin being spent came from a V4-format transaction (PROD-12599).
+
+Signing a transparent Zcash transaction whose input came from a V4 (Sapling-format) transaction failed immediately: no review screen appeared on the device, and Ledger Live showed "Something went wrong" with no detail. V4 is still valid on mainnet and still emitted by exchanges and older wallets, while Ledger Live itself emits V5 — so an account funded from within Ledger Live never hit this, which is why the failure looked intermittent. It is in fact deterministic, decided by which software created the funding transaction.
+
+`serializedPreviousTransactionOverride` carries the source transaction's raw on-chain bytes so the device can compute the correct ZIP-244 txid for a V5 transaction, whose Orchard bundle the signer kit's serialization would otherwise strip. It was being set for every version. The kit chunks a V4 transaction expecting Ledger's internal serialization, whose header carries a consensus branch id absent from the on-chain bytes; given those bytes it read the input count four bytes late and threw while chunking that input. The override is now restricted to the versions that need it, and a V4 source transaction goes back through the serialization path that has always handled it, its Sapling fields travelling in `extraData` as before.
+
+The decision is made per input, which is what a send spending several coins looks like from the outside: the V5 inputs are chunked and their trusted inputs obtained from the device first, then the transaction dies when the V4 input's turn comes. The device has already answered several times by then, yet no review screen is ever reached — so the failure looks like a device problem rather than a serialization one.
+
+Untagged device action errors are also no longer flattened into a message-less `Error`, so a failure inside a device action task names itself in the logs instead of surfacing only as "Something went wrong".
--- a/libs/coin-modules/coin-bitcoin/src/prepareTransaction.ts
+++ b/libs/coin-modules/coin-bitcoin/src/prepareTransaction.ts
@@ -83,8 +83,6 @@ export const prepareTransaction: AccountBridge<Transaction>["prepareTransaction"
     invariant(networkInfo.family === "bitcoin", "bitcoin networkInfo expected");
   }
 
-  const feePerByte = inferFeePerByte(transaction, networkInfo);
-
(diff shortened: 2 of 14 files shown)


libs/coin-modules/coin-bitcoin/src/prepareTransaction.ts
```diff
@@ -100,6 +98,24 @@ export const prepareTransaction: AccountBridge<Transaction>["prepareTransaction"
...
+
+  const inferredFeePerByte = inferFeePerByte(transaction, networkInfo);
+
+  // A chain whose fee is not a rate per vByte resolves the rate from the layout
+  // this transaction resolves to (see ChainAdapter.resolveFeePerByte). It runs on
+  // the merged exclusions, so the rate is resolved against the very UTXO set the
+  // transaction will be built from, and before the change detection below, so
+  // repeated preparations settle at a fixed point instead of re-deriving a rate
+  // that inferFeePerByte would discard.
+  const feePerByte =
```
qperrot, line 111: Are we sure this won't break bitcoin ?

pnpm-lock.yaml
```diff
@@ -25063,7 +25063,7 @@ packages:
     resolution: {integrity: sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==}
     engines: {node: '>=6'}
     peerDependencies:
-      rxjs: ^5.5.10
+      rxjs: '*'
```
qperrot, line 25066: is this normal change ?

.changeset/PROD-12599-zcash-v4-source-tx-override.md
```diff
@@ -0,0 +1,14 @@
+---
+"@ledgerhq/live-signer-zcash": patch
+"live-mobile": patch
```
qperrot, line 3: no mobile change it can be removed

cted-ledger: Mandatory to trigger mobile build
It is a pre-release PR

pnpm-lock.yaml
cted-ledger: result after fresh "pnpm install"

fAnselmi-Ledger (approved):

libs/coin-modules/coin-bitcoin/src/prepareTransaction.ts
cted-ledger: yes, thanks to the ternary below. 
bitcoin still get the same behaviour with:
```ts
inferFeePerByte(transaction, networkInfo);
```

francois-guerin-ledger (approved):

PR #20367, opened 2026-08-03
test(e2e-mobile): add Borrow Detox E2E coverage (LIVE-35015)

alexstapenka-ledger:
### ✅ Checklist

  - Mobile Borrow Detox E2E only (e2e/mobile/specs/borrow/) — no production app UI changes
  - Speculos/DMK signing infra used by on-chain mobile E2E (initUtil, speculosUtils) — borrow opts into Speculos recycle via recycleSpeculosAfterCliOnApp
  - On-chain scenarios require DISABLE_TRANSACTION_BROADCAST=0, mainnet ETH_4, and Borrow staging API availability

### 📝 Description

**Problem:** Desktop Borrow E2E (cold start, open loan, repay, withdraw) was merged, but mobile had no equivalent Detox coverage for LIVE-35015 / TMS cases B2CQA-6062, B2CQA-6065, B2CQA-6073, B2CQA-6080.

**Solution:** Add a single mobile spec file (desktop parity) with shared runners, page object, and portfolio entry helpers. On-chain flows reuse @ledgerhq/live-e2e-shared/borrow/borrowSetup for chain state setup/teardown.

Key mobile-specific work:
- Borrow webview stays mounted when navigating away — recovery via portfolio deeplink, hot-start intro dismiss, and stale-state loops in borrow.page.ts
- Speculos signing parity with swap/borrow desktop (refreshSpeculosForSigning, DMK session release, opt-in CLI Speculos recycle)
- Best-effort afterAll loan cleanup so staging Borrow API flakes (403/timeout) do not fail a green suite
- Borrow-only feature flags (pnl, analytics consent off) scoped to BORROW_FEATURE_FLAGS

**Run locally:**

    cd e2e/mobile
    export MOCK=0 DISABLE_TRANSACTION_BROADCAST=0 SPECULOS_DEVICE=nanoX
    pnpm test:android specs/borrow/

### ❓ Context

- **JIRA or GitHub link**: [LIVE-35015](https://ledgerhq.atlassian.net/browse/LIVE-35015)

---

[LIVE-35015]: https://ledgerhq.atlassian.net/browse/LIVE-35015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


VicAlbr:
Needs https://github.com/LedgerHQ/ledger-live/pull/20382 to be merged first

e2e/mobile/utils/speculosUtils.ts
```diff
@@ -346,10 +357,83 @@ async function waitForBridgeEnv(
...
+    } catch {
+      await new Promise(resolve => setTimeout(resolve, 500));
+    }
+  }
+
+  throw new Error(`Speculos HTTP not ready on port ${port} after ${timeoutMs}ms`);
+}
+
+/** Re-register Speculos in the app and verify HTTP reachability (no Speculos relaunch). */
+export async function prepareSpeculosForSigning(): Promise<void> {
```
ypolishchuk-ledger: this is not un use anywhere

e2e/mobile/utils/speculosUtils.ts
```diff
@@ -346,10 +357,83 @@ async function waitForBridgeEnv(
...
+  const entries = Array.from(speculosDevices.entries());
+  return entries.at(-1)?.[0];
+}
+
+export function getActiveSpeculosAddress(): string | undefined {
+  const port = getActiveSpeculosPort();
+  return port === undefined ? undefined : getKnownSpeculosAddress(port);
+}
+
+export async function ensureSpeculosRegisteredInApp(): Promise<void> {
```
ypolishchuk-ledger: this is not in use anywhere

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,693 @@
...
+    }
+  }
+
+  private async waitForLoanOverviewReady(): Promise<void> {
+    await waitForCurrentWebviewUrlToContain("/loanoverview/", 20_000);
+    // Overview wrapper test id is not always mounted; Repay CTA is the stable marker (desktop parity).
+    await waitForWebElementToBeEnabled(this.repayButtonId, 60_000);
+  }
+
+  private async waitForWithdrawOverviewReady(): Promise<void> {
```
ypolishchuk-ledger: this is not in use anywhere

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Borrow mobile E2E — single spec file (desktop parity: e2e/desktop/tests/specs/borrow.spec.ts).
+ *
+ * On-chain scenarios share ETH_4 mainnet state; keep describes in this order so hooks do not race.
```
ypolishchuk-ledger: Please make the ordering dependency explicit (e.g. a single `it` with sub-steps, or ensureRepayTestPrecondition-style guards at the start of the withdraw test) rather than relying on file position. Because a failure/retry of the first test leaves the second in an undefined state, and they can't be run in isolation.

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  runBorrowColdStartTest,
+  runBorrowOpenLoanTest,
+  runBorrowRepayWithdrawTests,
+} from "./borrow";
+
+const borrowTags = ["@NanoSP", "@NanoX", "@Stax", "@Flex", "@NanoGen5", "@ethereum", "@family-evm"];
+
+// Sonar S2187: Jest entrypoint — scenarios register describe/it via runBorrow* in ./borrow.ts.
+describe("Borrow mobile E2E", () => {
+  it("registers borrow scenario runners", () => {
```
ypolishchuk-ledger: Understood re: Sonar S2187, but this asserts nothing meaningful and will show as a passing test in reports, muddying signal. Prefer the documented Sonar suppression (// NOSONAR / ignore rule) over a placeholder assertion, or move the runners so the file legitimately owns real its.

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,693 @@
...
+  }
+
+  /** Close intro on hot start (existing loan). Do not tap "Simulate my loan". */
+  private async dismissIntroModalForHotStart(): Promise<boolean> {
+    if (!(await this.isWebTestIdVisible(this.introModalId))) {
+      return false;
+    }
+
+    const closeXpaths = [
+      `//*[@data-testid='${this.introModalId}']//button[contains(@aria-label, 'Close') or contains(@aria-label, 'close')]`,
```
ypolishchuk-ledger: These locators seem very fragile, is there a way to use more reliable locators with IDs?

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,693 @@
...
+      if (await this.isLoansDashboardReady()) {
+        return;
+      }
+
+      if (await this.dismissIntroModalForHotStart()) {
+        continue;
+      }
+
+      await this.recoverFromStaleBorrowWebview();
+      await new Promise(resolve => setTimeout(resolve, 1_000));
```
ypolishchuk-ledger: In several places in this class you use fixed setTimeout sleeps, please prefer polling on a condition (you already have isWebTestIdVisible/isWebElementEnabledByTestId) instead of blind waits.

e2e/mobile/utils/speculosUtils.ts
alexstapenka-ledger: fixed

e2e/mobile/utils/speculosUtils.ts
alexstapenka-ledger: fixed

e2e/mobile/page/trade/borrow.page.ts
alexstapenka-ledger: fixed

e2e/mobile/specs/borrow/borrow.spec.ts
alexstapenka-ledger: fixed

e2e/mobile/specs/borrow/borrow.spec.ts
alexstapenka-ledger: fixed

e2e/mobile/page/trade/borrow.page.ts
alexstapenka-ledger: fixed

e2e/mobile/page/trade/borrow.page.ts
alexstapenka-ledger: fixed

alexstapenka-ledger:
https://github.com/LedgerHQ/ledger-live/actions/runs/30988898488
image
image

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,836 @@
...
+      await waitForElementById(app.send.summaryContinueEnabledButtonId, 5_000);
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
+  private async isWebTestIdVisible(testId: string): Promise<boolean> {
+    try {
+      await waitWebElementByTestId(testId, { timeout: 2_000, throwOnTimeout: true });
```
cunhabruno: Why not just using `waitWebElementByTestId` with `throwOnTimeout` false, so you dont this this new function?

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,836 @@
+import { Step } from "jest-allure2-reporter/api";
+import { Account } from "@ledgerhq/live-e2e-shared/enum/Account";
+import { isAndroid } from "../../helpers/commonHelpers";
+import { WebElementHelpers } from "../../helpers/elementHelpers";
+import { refreshSpeculosForSigning } from "../../utils/speculosUtils";
+
+const BORROW_SPECULOS_APP = Account.ETH_4.currency.speculosApp.name;
```
cunhabruno: The spec already owns `openLoanAccount`  the page object shouldn't hardcode the currency.

e2e/mobile/utils/speculosUtils.ts
```diff
@@ -346,10 +357,68 @@ async function waitForBridgeEnv(
...
+    throw new Error("No active Speculos port");
+  }
+
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    try {
+      await fetchCurrentScreenTexts(port);
+      return;
+    } catch {
+      await new Promise(resolve => setTimeout(resolve, 500));
```
cunhabruno: `delay(ms)` already exists in `helpers/commonHelpers.ts` and is used by other page objects.

e2e/mobile/helpers/commonHelpers.ts
```diff
@@ -95,6 +95,26 @@ export async function launchApp(customConfig: Detox.DeviceLaunchAppConfig = {})
...
+
+  log.warn("E2E Bridge not responding — relaunching app to restore WebSocket");
+  const port = await launchApp({ newInstance: true });
+  await device.reverseTcpPort(port);
+
+  const deadline = Date.now() + 60_000;
+  while (Date.now() < deadline) {
+    const envsAfterRelaunch = await getEnvs();
+    if (envsAfterRelaunch) return;
+    await new Promise(resolve => setTimeout(resolve, 500));
```
cunhabruno: same here, use Delay()

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,836 @@
...
+    try {
+      await waitWebElementByTestId(testId, { timeout: 2_000, throwOnTimeout: true });
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
+  private async pollInterval(delayMs = 500): Promise<void> {
+    await new Promise(resolve => setTimeout(resolve, delayMs));
```
cunhabruno: same here, use Delay()

e2e/mobile/page/trade/borrow.page.ts
```diff
@@ -0,0 +1,836 @@
...
+
+    while (Date.now() < deadline) {
+      if (await this.isWebElementEnabledByTestId(testId)) {
+        await tapWebElementByTestId(testId);
+        return;
+      }
+
+      try {
+        const button = getWebElementByXpath(
+          `//*[(self::button or @role='button') and contains(normalize-space(.), '${label}')]`,
```
cunhabruno: multiple xpaths are build in method bodies, make them a field at the top instead.
Also, would they be able to be replaced by testids?

e2e/mobile/page/wallet/portfolio.page.ts
```diff
@@ -623,4 +623,18 @@ export default class PortfolioPage {
   async checkStocksListPageVisible() {
     await this.checkListPageVisible(this.stocksListId);
   }
+
+  private readonly borrowEntryPointId = "portfolio-borrow-entry-point";
+  private readonly borrowExploreCtaId = "borrow-explore-cta";
```
cunhabruno: `borrowEntryPointId` / `borrowExploreCtaId` are declared after all the methods; every other locator in this file is in the top field block. Please move them up.

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+    jest.setTimeout(BORROW_HOOK_TIMEOUT_MS);
+    await initBorrowApp({
+      userdata: "speculos-x-other-account",
+      featureFlags: BORROW_FEATURE_FLAGS,
+    });
+  }, BORROW_HOOK_TIMEOUT_MS);
+
+  setTeamOwner(Team.EARN);
+  $TmsLink("B2CQA-6062");
+  ["@NanoSP", "@LNS", ...borrowTags].forEach(tag => $Tag(tag));
```
cunhabruno: `borrowTags` already starts with `"@NanoSP"`, so it's applied twice — use `["@LNS", ...borrowTags]`.

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+  try {
+    await resetLoanState({ nanoAppCatalogPath: NANO_APP_CATALOG_PATH });
+  } catch (error) {
+    console.warn(`[borrow] ${context} resetLoanState failed (non-fatal):`, error);
+  }
+}
+
+describe("Borrow - Cold start", () => {
+  beforeAll(async () => {
+    jest.setTimeout(BORROW_HOOK_TIMEOUT_MS);
```
cunhabruno: Couldn't this be declared in a single time at top level?

valentinemace9 (dismissed):

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+async function initBorrowApp(options: ApplicationOptions) {
+  await app.init(options);
+  // Same as desktop borrow / mobile swap: skip Manager app install on Connect device.
+  await swapSetup();
+  // Prior describes may leave Borrow on the navigation stack; deeplink resets to portfolio
+  // (separate spec files used to get a fresh launchApp per file from setup.ts).
+  await app.mainNavigation.openPortfolioViaDeeplink();
+}
+
+async function openBorrowFromPortfolio() {
```
cunhabruno: `openBorrowFromPortfolio` doesn't open Borrow — it opens the portfolio and asserts the entry point is visible. Try to use existing functions in portfolio page and add only borrow related code into the borrow page file

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  it(
+    "Portfolio entry point opens borrow, simulates loan, and completes open-loan execution",
+    async () => {
+      await openBorrowFromPortfolio();
+      await app.borrow.goAndWaitForBorrowToBeReady(async () =>
+        app.portfolio.clickBorrowEntryPoint(),
+      );
+
+      await app.borrow.dismissIntroModalIfVisible();
```
cunhabruno: Is it needed to dismiss `if` visible? I would make it more deterministic, not using conditionals

e2e/mobile/specs/borrow/borrow.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+      await app.borrow.expectExecutionFlowVisible();
+
+      if (await app.borrow.completeGiveApprovalIfRequired()) {
+        await app.borrow.completeHostDeviceSignature(openLoanSpeculosAppName, async () => {
+          await app.speculos.signTokenApproval();
+        });
+        await app.borrow.expectApprovalStepCompleted();
+      }
+
+      await app.borrow.completeAuthorizeExecutionWithRetry({
```
cunhabruno: Broader ask: could you review the whole PR with determinism in mind before we merge?

The pattern I keep hitting is optional behaviour and injected callbacks where the state might actually be knowable.

Where the app state is known, I'd rather see assert-then-act at spec level than a recovery loop. It'll read better and cut the flakiness surface — and it makes the retry logic tractable: as written, after `retryAfterOnChainFailure` re-taps authorize and re-signs, the loop calls `clickAuthorize()` again without checking completion, so a recovered failure still fails the test and signs twice.

e2e/mobile/specs/borrow/borrow.spec.ts
alexstapenka-ledger: fixed thanks!

e2e/mobile/specs/borrow/borrow.spec.ts
alexstapenka-ledger: fixed thanks!

e2e/mobile/specs/borrow/borrow.spec.ts
alexstapenka-ledger: fixed thanks!

e2e/mobile/specs/borrow/borrow.setup.ts
```diff
@@ -0,0 +1,49 @@
...
+      app: openLoanAccount.currency.speculosApp,
+      cmd: liveDataWithAddressCommand(openLoanAccount),
+    },
+  ],
+  speculosForSetupOnly: true,
+  recycleSpeculosAfterCliOnApp: true,
+});
+
+export type BorrowBeforeAllOptions = {
+  /** Relaunch the app so Wallet 4.0 navigation is reachable after a prior describe left Borrow open. */
```
martijnhjk: It seems to me this is something that should be handled as part of the global test setup - teardown lifecycle rather than inline in a spec file.

Every test should be an isolated instance and launch a new instance of the app completely - therefore an explicit workaround like this should not be needed.

e2e/mobile/specs/borrow/borrow.constants.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  FF_BORROW_MOBILE,
+  FF_LWM_WALLET_40_Q2_NO_ANALYTICS_CONSENT,
+} from "../../utils/featureFlagUtils";
+
+import type { OptionalFeatureMap } from "@shared/feature-flags";
+
+export const BORROW_FEATURE_FLAGS: OptionalFeatureMap = {
```
martijnhjk: please consider declaring FF in the central `featureFlagUtils` for maintainability.

e2e/mobile/helpers/commonHelpers.ts
```diff
@@ -95,6 +95,22 @@ export async function launchApp(customConfig: Detox.DeviceLaunchAppConfig = {})
...
+  const port = await launchApp({ newInstance: true });
+  await device.reverseTcpPort(port);
+
+  const envsAfterRelaunch = await getEnvs();
+  if (!envsAfterRelaunch) {
+    throw new Error(
+      "E2E bridge not responding after app relaunch. Restart the emulator, stop stale Speculos containers, then retry.",
+    );
+  }
+}
```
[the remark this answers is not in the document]
martijnhjk: @alexstapenka-ledger I think Copilot might have a valid point here, could you please clarify how you are mitigating the concern raised?

e2e/mobile/helpers/commonHelpers.ts
```diff
@@ -95,6 +95,26 @@ export async function launchApp(customConfig: Detox.DeviceLaunchAppConfig = {})
   return port;
 }
 
+export async function ensureBridgeReady(): Promise<void> {
```
martijnhjk: This function is called more than once therefore we could be relaunching the app more than once. 

What changes introduced in this PR cause the requirement of this function? Why would the bridge not be ready?

e2e/mobile/helpers/commonHelpers.ts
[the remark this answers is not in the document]
alexstapenka-ledger: This was addressed in the initial implementation. After relaunch, ensureBridgeReady() polls getEnvs() in a loop for up to 60s (500ms between attempts) rather than checking once. Each getEnvs() call already has its own 10s bridge timeout, so slow app reconnect after relaunch is covered. Copilot's comment appears to be on an outdated diff snapshot.

e2e/mobile/utils/featureFlagUtils.ts
```diff
@@ -43,6 +43,31 @@ export const FF_LWM_WALLET_40_Q2 = {
...
+export const FF_BORROW_MOBILE = {
+  ptxBorrowLiveApp: {
+    enabled: true,
+    params: { manifest_id: "borrow" },
+  },
+} satisfies OptionalFeatureMap;
+
+/** Wallet 4.0 + borrow live app + PnL for mobile borrow E2E. */
+export const FF_BORROW_E2E = {
+  ...FF_LWM_WALLET_40_Q2_NO_ANALYTICS_CONSENT,
```
martijnhjk: do you really _need_ `FF_LWM_WALLET_40_Q2_NO_ANALYTICS_CONSENT` here? Have you tried it simply with `FF_LWM_WALLET_40_Q2`?

e2e/mobile/utils/featureFlagUtils.ts
```diff
@@ -43,6 +43,31 @@ export const FF_LWM_WALLET_40_Q2 = {
...
+
+export const FF_BORROW_MOBILE = {
+  ptxBorrowLiveApp: {
+    enabled: true,
+    params: { manifest_id: "borrow" },
+  },
+} satisfies OptionalFeatureMap;
+
+/** Wallet 4.0 + borrow live app + PnL for mobile borrow E2E. */
+export const FF_BORROW_E2E = {
```
martijnhjk: having both `FF_BORROW_MOBILE` _and_ `FF_BORROW_E2E` is confusing - please consider one const `FF_BORROW_ENABLED` and declaring all required params there.

e2e/mobile/utils/featureFlagUtils.ts
alexstapenka-ledger: hm, exactly, not needed. Borrow userdata already sets hasSeenAnalyticsOptInPrompt: true so its enough in this case

VicAlbr:
Heads-up on scope, since this PR is being split.

The **cold start** test from this PR has been extracted into #20959 so it can land on its own. That PR is intentionally minimal (~100 lines): the spec, a two-method page object, the portfolio entry point helpers, and `FF_BORROW_ENABLED`. Mobile E2E is green on **both platforms** there — [run 32349071360](https://github.com/LedgerHQ/ledger-live/actions/runs/32349071360), 1 passed / 1 total on iOS and Android.

The **on-chain** flows (open / repay / withdraw) stay out of #20959 and continue here or in a follow-up, because a manual on-chain run against ETH_4 surfaced two blockers worth recording:

1. **The three on-chain specs race one mainnet account.** Jest shards by file and mobile CI runs `maxWorkers: 3`, so `borrow.openLoan` / `borrow.repay` / `borrow.withdraw` start together against the same address — their `beforeAll` preconditions undo each other and their transactions are built from the same nonce. `libs/live-e2e-shared/src/borrow/README.md` ("Collision risks & serialization") already requires all on-chain borrow flows to live in a single spec file for exactly this reason, and desktop follows it. Worth noting the only green on-chain run posted on this PR ([30988898488](https://github.com/LedgerHQ/ledger-live/actions/runs/30988898488)) ran a **single** spec file — the four-file split came afterwards and has no on-chain run behind it.

2. **All three fail before reaching any borrow UI.** `expectBorrowScreenVisible` times out after 60s because `clickBorrowEntryPoint` taps `borrow-explore-cta` while `expectBorrowEntryPointVisible` scrolled to `portfolio-borrow-entry-point`. `tapById` does not wait for visibility, so with multi-account userdata the tap lands off screen and never navigates. Cold start passes only because its isolated seed keeps the CTA on screen. #20959 fixes this by tapping the card that was scrolled into view — the app wires both to the same `onPress`. This matches the Copilot review comment on `portfolio.page.ts` that was filed as suppressed.

Also FYI: `evmSpeculos.ts` hold-to-sign is now superseded by develop, which handles touch devices via `DeviceLabels.HOLD_TO_SIGN` / `longPressAndRelease` and clears the Transaction Check opt-in.

Tracking the determinism work in [QAA-1477](https://ledgerhq.atlassian.net/browse/QAA-1477).

[QAA-1477]: https://ledgerhq.atlassian.net/browse/QAA-1477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #20385, opened 2026-08-04
test(e2e): reload buy/sell webview when stuck loading

jeportie:
### Checklist

- [ ] `npx changeset` was attached. _N/A — E2E test-only change (`e2e/desktop`), no published package impacted (prior `test(e2e)` buySell commits shipped without one)._
- [ ] **Covered by automatic tests.** _This change is E2E test infrastructure; it is validated by re-running the Desktop E2E suite in CI rather than by unit tests._
  - Desktop Buy/Sell E2E flows (`buySell.spec.ts`) — buy & sell from the portfolio, provider selection
  - No product/runtime code changed; only the `BuyAndSellPage` page object

### Description

The nightly Desktop E2E test `Sell [Tether USD] asset` (`buySell.spec.ts:212`) fails intermittently at `chooseAssetIfNotSelected` with:

```
TimeoutError: locator.textContent: Timeout 60000ms exceeded.
  - waiting for getByTestId('crypto-amount-option-button')
```

**Root cause.** The desktop Buy/Sell flow is rendered by the PTX (swap) live app (`swap-live-app-stg.ledger-test.com`) inside the `"buy"` webview — buy/sell/swap share the same app. Intermittently that app **stays stuck on its loading spinner** and never renders the amount screen, so `crypto-amount-option-button` never appears. The Allure artifacts confirm it: the failure screenshot shows the spinner under the "Buy or sell" header, the accessibility snapshot shows an empty `iframe`, and **all 43 webview requests return 200** (nothing failed at the network layer). The webview console shows a firebase remote-config load race (`Effective app config is null`, `ZodError: optimisticBalanceTtlMs Required`).

It is **flaky, not USDT-specific**: `Sell [Tether USD]` passed Aug 3 (nanoSP) and failed Jul 31 (nanoGen5) and Aug 4 (stax). `Sell [Ethereum]` and `Sell [Tether USD]` use the **same** Ethereum account (`0xE32ad14b…`), so the webview state is identical — it is per-app-instance load luck. USDT is sellable per the backend `sell/v1/cryptoLimitations`.

**Solution.** Add `waitForCryptoSelectorReady()` — wait for the crypto selector to be visible and, if the app is stuck, **reload the webview and retry** (2 reloads × 30 s) before failing with an explicit "webview stuck loading" message. It is called at the start of `chooseAssetIfNotSelected`. This mirrors the existing Borrow-webview recovery in `borrow.page.ts`.

> The stall itself originates in the external PTX/swap staging app; the durable fix (remote-config load race / staging stability) belongs to the PTX/swap team. This PR makes the E2E resilient and self-healing for the transient case.

### Context

- **JIRA or GitHub link**: _None linked yet — please associate with the relevant QAA ticket._

### DIFF
--- a/e2e/desktop/tests/page/buyAndSell.page.ts
+++ b/e2e/desktop/tests/page/buyAndSell.page.ts
@@ -129,7 +129,7 @@ export class BuyAndSellPage extends WebViewAppPage {
               `after ${maxReloads + 1} attempts — webview stuck loading.`,
           );
         }
-        await webview.reload();
+        await webview.reload({ timeout: readyTimeout, waitUntil: "domcontentloaded" });
       }
     }
   }
--- a/e2e/desktop/tests/page/buyAndSell.page.ts
+++ b/e2e/desktop/tests/page/buyAndSell.page.ts
@@ -104,13 +104,10 @@ export class BuyAndSellPage extends WebViewAppPage {
     await this.selectAssetInDrawer(account);
   }
 
-  /**
-   * Buy/Sell runs inside the PTX (swap) live app webview. That app occasionally
-   * stays stuck on its loading spinner and never renders the amount screen, so the
-   * crypto selector never appears and the flow times out on a bare `.textContent()`.
-   * Wait for the selector and, if it never shows, reload the webview and retry —
-   * the same recovery the Borrow webview uses (see borrow.page.ts).
-   */
+/**
+ * Workaround: PTX Buy/Sell web app can remain stuck loading; reload the webview and retry.
+ * Mirrors the recovery used in borrow.page.ts.
+ */
   @step("Wait for the Buy/Sell web app to finish loading")
   private async waitForCryptoSelectorReady() {
     const readyTimeout = 30_000;
--- a/e2e/desktop/tests/page/buyAndSell.page.ts
+++ b/e2e/desktop/tests/page/buyAndSell.page.ts
@@ -98,11 +98,40 @@ export class BuyAndSellPage extends WebViewAppPage {
 
   @step("Choose crypto asset if not selected")
   async chooseAssetIfNotSelected(account: AccountType) {
+    await this.waitForCryptoSelectorReady();
     if (await this.isCorrectAssetAlreadySelected(account)) return;
-    const webview = await this.getWebView();
-    await expect(webview.getByTestId(this.cryptoCurrencySelector)).toBeEnabled();
     await this.clickElement(this.cryptoCurrencySelector);
     await this.selectAssetInDrawer(account);
+  }
+
+  /**
+   * Buy/Sell runs inside the PTX (swap) live app webview. That app occasionally
+   * stays stuck on its loading spinner and never renders the amount screen, so the
+   * crypto selector never appears and the flow times out on a bare `.textContent()`.
+   * Wait for the selector and, if it never shows, reload the webview and retry —
+   * the same recovery the Borrow webview uses (see borrow.page.ts).
+   */
+  @step("Wait for the Buy/Sell web app to finish loading")
+  private async waitForCryptoSelectorReady() {
+    const readyTimeout = 30_000;
+    const maxReloads = 2;
+    const webview = await this.getWebView();
+    const selector = webview.getByTestId(this.cryptoCurrencySelector);
+
+    for (let attempt = 0; ; attempt++) {
+      try {
+        await expect(selector).toBeVisible({ timeout: readyTimeout });
+        return;
+      } catch {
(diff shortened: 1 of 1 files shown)


jeportie:
## 🧪 Triggered test workflows

Manually dispatched on `test/e2e-buysell-reload-webview-on-stuck-load` (commit `0c8d71ebea6`), scoped to the buySell spec:

| Workflow | Scope | Run |
|---|---|---|
| [Desktop] E2E Only | `Sell flow` filter (whole `buySell.spec.ts`), Speculos nanoSP | [run 30896538389](https://github.com/LedgerHQ/ledger-live/actions/runs/30896538389) |

**Scope rationale:** the change only touches the desktop `BuyAndSellPage` page object (`chooseAssetIfNotSelected` webview-load resilience). Mobile has a separate buySell page object that is untouched and no coin logic changed, so Mobile E2E and Coin Tester are skipped. The `Sell flow` grep matches both describes in `buySell.spec.ts` (`Buy / Sell flow…` and `Sell flow - [X]`) and no other desktop spec — it runs the full buy/sell suite that exercises the changed method.

> Flaky-fix caveat: a single green run is necessary but not sufficient to prove the flake is gone (the stall is intermittent). Happy to re-dispatch a few more times for confidence.

e2e/desktop/tests/page/buyAndSell.page.ts
```diff
@@ -98,13 +98,39 @@ export class BuyAndSellPage extends WebViewAppPage {
...
     await this.clickElement(this.cryptoCurrencySelector);
     await this.selectAssetInDrawer(account);
   }
 
+  /**
+   * Workaround: PTX Buy/Sell web app can remain stuck loading; reload the webview and retry.
+   * Mirrors the recovery used in borrow.page.ts.
+   */
+  @step("Wait for the Buy/Sell web app to finish loading")
+  private async waitForCryptoSelectorReady() {
```
ypolishchuk-ledger, line 112: Maximum waiting time here is ~150s+ for one helper call before it throws. Please verify this fits within the spec/test.setTimeout budget for the buy/sell tests - otherwise the test dies on the global timeout with a generic message instead of the helpful error you crafted here. If needed, tighten readyTimeout (e.g. 15s) so the retry story stays within budget.

e2e/desktop/tests/page/buyAndSell.page.ts
```diff
@@ -98,13 +98,39 @@ export class BuyAndSellPage extends WebViewAppPage {
...
+    const readyTimeout = 30_000;
+    const maxReloads = 2;
+    const webview = await this.getWebView();
+    const selector = webview.getByTestId(this.cryptoCurrencySelector);
+
+    for (let attempt = 0; ; attempt++) {
+      try {
+        await expect(selector).toBeVisible({ timeout: readyTimeout });
+        return;
+      } catch {
```
ypolishchuk-ledger: The bare catch {} discards the original Playwright error every iteration. When this ultimately fails in CI, the report loses why (timeout vs. Target closed vs. crash). Capture the last error and chain it so debugging isn't blind

jeportie: Good point — fixed. The loop now catches with `catch (error)` and rethrows via `new Error(…, { cause: error })`, so CI keeps the underlying Playwright reason (`TimeoutError` vs `Target closed` vs crash) alongside the "webview stuck loading" message instead of discarding it every iteration. Mirrors the pattern already used in the mobile buySell page object (`e2e/mobile/page/trade/buySell.page.ts`) and `ledgerSyncPulledDataUtils.ts`. Thanks!

e2e/desktop/tests/page/buyAndSell.page.ts
jeportie: Verified against `e2e/desktop/playwright.config.ts`: per-test `timeout` is 400s on CI (1200s locally), there is no `test.setTimeout` / `describe.configure` override for buySell, and `globalTimeout` is disabled. This helper's worst case is 150s (3×30s visibility + 2×30s reload) and it runs early in the flow (only app launch + navigation precede it), so it stays well inside the 400s budget — the explicit "webview stuck loading" error surfaces rather than a generic test-timeout. I kept `readyTimeout` at 30s to avoid reloading slow-but-healthy loads, but happy to drop it to ~20s if you'd prefer faster self-healing + more headroom. LMK!

ypolishchuk-ledger: Sounds reasonable, thanks!

ypolishchuk-ledger (approved):

PR #20401, opened 2026-08-04
test(e2e): Swap number-formatting-by-language coverage (B2CQA-4014)

kentoforik:
## Summary
- Add Desktop (Playwright) and Mobile (Detox/Jest) e2e tests verifying Swap displays the send amount, send countervalue, receive amount, and best quote card with the correct decimal/thousands separators for English, French, and German app languages.
- Fix a pre-existing, locale-hardcoded bug in mobile's `getProviderList()` (assumed English "N quotes found" phrasing).

## Test plan
- [x] `pnpm --filter ledger-live-desktop-e2e-tests typecheck`
- [x] `pnpm --filter ledger-live-mobile-e2e-tests typecheck`
- [x] [Ran full SWAP scope against Speculos (Desktop)](https://github.com/LedgerHQ/ledger-live/actions/runs/32026284511) — all passing
- [x] [Ran full SWAP scope on Android/iOS emulator (Mobile)](https://github.com/LedgerHQ/ledger-live/actions/runs/32026335981)  — all passing

Jira: [B2CQA-4014](https://ledgerhq.atlassian.net/browse/B2CQA-4014)

[B2CQA-4014]: https://ledgerhq.atlassian.net/browse/B2CQA-4014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/desktop/tests/page/swap.page.ts
+++ b/e2e/desktop/tests/page/swap.page.ts
@@ -138,9 +138,13 @@ export class SwapPage extends WebViewAppPage {
     await expect(webview.getByTestId(this.bestValueInfoIcon)).toBeVisible();
     await expect(webview.getByTestId(this.quotesCountdown)).toBeVisible();
 
-    return await webview
+    const providerList = await webview
       .locator(`[data-testid^='${SwapPage.PROVIDER_NAME_PREFIX}']`)
       .allTextContents();
+    if (providerList.length === 0) {
+      throw new Error("No quote providers were returned");
+    }
+    return providerList;
   }
 
   @step("Check elements presence on swap approval step")
@@ -162,9 +166,11 @@ export class SwapPage extends WebViewAppPage {
     const provider = SwapProvider.getNameByUiName(providerUiName);
     const amount = await webview
       .locator(this.providerContainerInfoSelector(provider, "amount-label"))
+      .first()
       .textContent();
     const fiatAmount = await webview
       .locator(this.providerContainerInfoSelector(provider, "fiatAmount-label"))
+      .first()
       .textContent();
     return { amount, fiatAmount };
   }
--- a/e2e/desktop/tests/page/swap.page.ts
+++ b/e2e/desktop/tests/page/swap.page.ts
@@ -42,6 +42,8 @@ export class SwapPage extends WebViewAppPage {
   private readonly noQuotesPlaceholder = "quotes-error-state";
   private toAccountCoinSelector = "to-account-coin-selector";
   private readonly toAccountAccountNameTag = "to-account-account-name-tag";
+  private readonly toAccountAmountInput = "to-account-amount-input";
+  private readonly fromAccountAmountInactive = "from-account-amount-inactive";
   private specificQuoteCardProviderName = (provider: string) =>
     `[data-testid^='${SwapPage.PROVIDER_NAME_PREFIX}${provider.toLowerCase()}']`;
   private providerContainerSelector = (provider: string) =>
@@ -152,6 +154,19 @@ export class SwapPage extends WebViewAppPage {
   @step("Click Continue button")
   async clickContinueButton() {
     await this.continueBtn.click();
+  }
+
+  @step("Get quote card amount texts for $0")
+  async getQuoteAmountTexts(providerUiName: string) {
+    const webview = await this.getWebView();
+    const provider = SwapProvider.getNameByUiName(providerUiName);
+    const amount = await webview
+      .locator(this.providerContainerInfoSelector(provider, "amount-label"))
+      .textContent();
+    const fiatAmount = await webview
+      .locator(this.providerContainerInfoSelector(provider, "fiatAmount-label"))
+      .textContent();
+    return { amount, fiatAmount };
   }
 
   @step("Check quotes container infos")
@@ -330,7 +345,11 @@ export class SwapPage extends WebViewAppPage {
   async checkBestOffer() {
     const quoteContainers = await this.getAllSwapProviders();
     const quotes = await this.extractQuotesAndFees(quoteContainers);
-    const bestOffer = quotes.reduce<{ rate: number; fees: number; quote: string } | null>(
+    const bestOffer = quotes.reduce<{
+      rate: number;
+      fees: number;
+      quote: string;
+    } | null>(
       (max, current) =>
         current && (!max || current.rate - current.fees > max.rate - max.fees) ? current : max,
       null,
@@ -373,6 +392,18 @@ export class SwapPage extends WebViewAppPage {
   async getAmountToSend() {
     const webview = await this.getWebView();
     return await webview.getByTestId(this.fromAccountAmountInput).inputValue();
+  }
+
+  @step("Retrieve send currency countervalue text")
+  async getSendCountervalueText() {
+    const webview = await this.getWebView();
+    return await webview.getByTestId(this.fromAccountAmountInactive).textContent();
+  }
+
+  @step("Retrieve receive currency amount value")
+  async getAmountToReceive() {
+    const webview = await this.getWebView();
+    return await webview.getByTestId(this.toAccountAmountInput).textContent();
   }
 
   @step("Click switch button")
@@ -764,7 +795,9 @@ export class SwapPage extends WebViewAppPage {
   async clickGiveAuthorizationButton() {
     const webview = await this.getWebView();
     const authorizationButton = webview.getByTestId(this.signPermitButton);
-    await expect(authorizationButton).toBeVisible({ timeout: APPROVAL_PROCESSING_TIMEOUT });
+    await expect(authorizationButton).toBeVisible({
+      timeout: APPROVAL_PROCESSING_TIMEOUT,
+    });
     await authorizationButton.click();
   }
 
--- a/e2e/desktop/tests/page/swap.page.ts
+++ b/e2e/desktop/tests/page/swap.page.ts
@@ -29,6 +29,11 @@ export class SwapPage extends WebViewAppPage {
     "../artifacts/ledgerwallet-swap-history.csv",
   );
   private static readonly PROVIDER_NAME_PREFIX = "lumen-quote-card-provider-name-";
+  private static readonly AMOUNT_LABEL_SUFFIX = "amount-label";
+  private static readonly FIAT_AMOUNT_LABEL_SUFFIX = "fiatAmount-label";
+  private static readonly NETWORK_FEES_HEADING_SUFFIX = "networkFees-heading";
+  private static readonly EXTRA_FEES_CONTAINER_SUFFIX = "extraFeesContainer";
+  private static readonly RATE_INFO_ICON_SUFFIX = "rate-infoIcon";
 
   private readonly fullSwapContainer = this.page.getByTestId("swap-web-app-container-full");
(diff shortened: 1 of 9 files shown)


e2e/desktop/tests/utils/amountUtils.ts
```diff
@@ -1,10 +1,67 @@
 import { expect } from "@playwright/test";
+import * as allure from "allure-js-commons";
```
ypolishchuk-ledger, line 2: The code of this file is partially duplicated in desktop and mobile folder, I would suggest to hoist the shared, platform-agnostic pieces (SEPARATORS_BY_LANGUAGE, FormattedNumberLanguage, getExpectedSeparators, buildFormattedAmountPattern) into @ledgerhq/live-e2e-shared and keep only the thin assertion wrapper per platform.

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -132,6 +138,21 @@ export class SwapPage extends WebViewAppPage {
     await this.continueBtn.click();
   }
 
+  @step("Get quote card amount texts for $0")
+  async getQuoteAmountTexts(providerUiName: string) {
+    const webview = await this.getWebView();
+    const provider = SwapProvider.getNameByUiName(providerUiName);
+    const amount = await webview
+      .locator(this.providerContainerInfoSelector(provider, "amount-label"))
```
ypolishchuk-ledger: I noticed that the locator suffix is repeated across this file. Would it make sense to move all of them as a variable at the top of the class?

e2e/desktop/tests/specs/numberFormat.swap.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      `Swap amounts are formatted correctly for language: ${label}`,
+      {
+        tag: [...DEVICE_TAGS],
+        annotation: { type: "TMS", description: "B2CQA-4014" },
+      },
+      async ({ app }) => {
+        await addTmsLink(getDescription(test.info().annotations, "TMS").split(", "));
+
+        await app.mainNavigation.openSettings();
+        await app.settings.changeLanguage(label);
```
ypolishchuk-ledger: English is the language by default; do we actually need to go to settings and change it to be english? :D

e2e/desktop/tests/specs/numberFormat.swap.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+// The send-amount input always groups with plain whitespace, regardless of language.
+const SEND_INPUT_THOUSANDS = " ";
+
+const languageCases = [
+  { languageId: "en" as const, label: "English" },
+  { languageId: "fr" as const, label: "Français" },
+  { languageId: "de" as const, label: "Deutsch" },
+];
+
+test.describe("Swap - amount formatting by language", () => {
```
ypolishchuk-ledger, line 30: this test is currently failing locally with `Error: Expected a formatted amount but got "" (send amount countervalue)` or with `Error: No quote providers were returned` while I see at least 2 quotes on the screen, is it expected?

e2e/desktop/tests/specs/numberFormat.swap.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+const fromAccount = Account.ETH_1;
+const toAccount = Account.BTC_NATIVE_SEGWIT_1;
+
+// Large enough to trigger thousands-grouping in the displayed amounts (B2CQA-4014).
+const TEST_AMOUNT = "1234.56";
+
+// The send-amount input always groups with plain whitespace, regardless of language.
+const SEND_INPUT_THOUSANDS = " ";
+
+const languageCases = [
```
ypolishchuk-ledger, line 25: Would it make sense to test it only for French here? From what I understand, French uses the narrow no-break space `U+202F` for grouping - by far the most fragile character to round-trip through the WebView bridge, the Detox/Playwright text extraction, and your assertion regex. If anything in the test harness or propagation is going to break, French breaks it; en/de use ordinary ASCII separators.
And for English, you can add formatting checks to one of the existing swap tests.

e2e/desktop/tests/utils/amountUtils.ts
kentoforik: Done — hoisted `NumberSeparators`/`FormattedNumberLanguage`/`getExpectedSeparators`/`buildFormattedAmountPattern` into `libs/live-e2e-shared/src/data/numberFormat.ts`. Each platform's `amountUtils.ts` now only keeps the allure-aware `expectFormattedAmount` wrapper.

e2e/desktop/tests/page/swap.page.ts
kentoforik: Done — added `AMOUNT_LABEL_SUFFIX`, `FIAT_AMOUNT_LABEL_SUFFIX`, `NETWORK_FEES_HEADING_SUFFIX`, `EXTRA_FEES_CONTAINER_SUFFIX`, `RATE_INFO_ICON_SUFFIX` as static class constants on `SwapPage` and replaced all inline string literals.

e2e/desktop/tests/specs/numberFormat.swap.spec.ts
kentoforik: Good catch — confirmed the userdata fixture already boots in en/en-US. Now skipping the settings/language-change step for the English case on both desktop and mobile.

e2e/desktop/tests/specs/numberFormat.swap.spec.ts
kentoforik: This should be fixed now — it turned out to be a stale selector: `develop` renamed the provider-name testid prefix (Lumen UI rework) after this branch diverged, which we picked up while rebasing onto `develop`. Could you re-pull and re-test locally to confirm both the countervalue and provider-list errors are gone?

e2e/desktop/tests/specs/numberFormat.swap.spec.ts
kentoforik: Good point on French being the most fragile case. We dropped German (it uses the same ASCII separators as English, so it wasn't adding unique coverage) but kept English, since skipping the language-change step for it is now free — the userdata fixture already boots in en/en-US, so there's no extra settings navigation cost.

e2e/mobile/utils/amountUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+  const label = ` (${context})`;
+  if (!text) {
+    throw new Error(`Expected a formatted amount but got "${text}"${label}`);
+  }
+  const value = text.trim();
+  const tokens: string[] = [];
+  let match: RegExpExecArray | null;
+  while ((match = NUMERIC_TOKEN_REGEX.exec(value)) !== null) {
+    tokens.push(match[0]);
+  }
```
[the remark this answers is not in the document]
kentoforik: This logic moved to `libs/live-e2e-shared/src/data/numberFormat.ts` as part of the DRY hoist above — replying with the analysis there.

e2e/desktop/tests/utils/amountUtils.ts
```diff
@@ -1,10 +1,67 @@
...
+  context: string,
+): Promise<void> {
+  await allure.step(context, () => {
+    const label = ` (${context})`;
+    if (!text) {
+      throw new Error(`Expected a formatted amount but got "${text}"${label}`);
+    }
+    const value = text.trim();
+    const tokens: string[] = [];
+    let match: RegExpExecArray | null;
```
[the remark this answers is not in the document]
kentoforik: Same — this moved to `libs/live-e2e-shared/src/data/numberFormat.ts`; see the reply on that thread for why the lastIndex concern doesn't apply.

libs/live-e2e-shared/src/data/numberFormat.ts
```diff
@@ -0,0 +1,42 @@
...
+const NUMERIC_TOKEN_REGEX = new RegExp(String.raw`\d[${ANY_SEPARATOR_CHARS}\d]*\d|\d`, "g");
+
+export function findNumericTokens(text: string): string[] {
+  const tokens: string[] = [];
+  let match: RegExpExecArray | null;
+  while ((match = NUMERIC_TOKEN_REGEX.exec(text)) !== null) {
+    tokens.push(match[0]);
+  }
+  return tokens;
+}
```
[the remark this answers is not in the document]
kentoforik: Verified this empirically rather than assuming: a global regex's `.exec()` resets `lastIndex` to `0` once it returns `null`, and this loop always runs to full exhaustion (no early return/break inside it), so `lastIndex` is guaranteed back to `0` before the next call starts. Confirmed with a standalone Node script running 3 sequential calls (including a repeat of the first string) — correct tokens every time, `lastIndex` reset to `0` after each call. No leakage across calls here.

e2e/mobile/specs/swap/otherTestCases/swapNumberFormatLocale.ts
```diff
@@ -0,0 +1,116 @@
...
+        cliCommandsOnApp: [
+          {
+            app: fromAccount.currency.speculosApp,
+            cmd: liveDataWithAddressCommand(fromAccount),
+          },
+          {
+            app: toAccount.currency.speculosApp,
+            cmd: liveDataWithAddressCommand(toAccount),
+          },
+        ],
```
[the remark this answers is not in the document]
kentoforik: `liveDataWithAddressCommand` is declared as an ambient global in `e2e/mobile/types/global.d.ts` (`var liveDataWithAddressCommand: typeof import("@ledgerhq/live-e2e-shared/cliCommandsUtils").liveDataWithAddressCommand;`) — same pattern as `app`, `$TmsLink`, `tapWebElementByTestId` used unimported throughout this file. No import needed; typechecks clean.

e2e/mobile/specs/swap/otherTestCases/swapNumberFormatLocale.ts
```diff
@@ -0,0 +1,119 @@
...
```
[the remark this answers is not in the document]
kentoforik: Duplicate of the comment above — same answer: it's an ambient global declared in `e2e/mobile/types/global.d.ts`, no import needed.

e2e/desktop/tests/page/swap.page.ts
```diff
@@ -375,6 +400,18 @@ export class SwapPage extends WebViewAppPage {
...
+  @step("Retrieve send currency countervalue text")
+  async getSendCountervalueText() {
+    const webview = await this.getWebView();
+    return await webview.getByTestId(this.fromAccountAmountInactive).textContent();
+  }
+
+  @step("Retrieve receive currency amount value")
+  async getAmountToReceive() {
+    const webview = await this.getWebView();
+    return await webview.getByTestId(this.toAccountAmountInput).textContent();
```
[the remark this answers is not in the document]
kentoforik: This is intentional — the receive-amount field is a non-editable, display-only element (the user can only change the *send* amount; the receive amount is computed), so it's read via `.textContent()` rather than `.inputValue()`, which only works on real editable `<input>`/`<textarea>`/`<select>` elements. Mobile's page object follows the same pattern for the same field.

cunhabruno (approved):

ypolishchuk-ledger (approved):

PR #20407, opened 2026-08-04
[LWDM] fix(send-flow): keep ton comment text a string when memo is empty

amaslakov:
### 📝 Description

TON sends were broken on the new send flow: typing a recipient address wedged the flow, with no error shown and no way forward. An empty Comment produced `comment.text: undefined`, which violates TON's required-string contract and threw out of `getTransactionStatus`.

Fixed in the TON memo applier rather than at the shared `"" → undefined` normalization, which is correct for flat-field families like `xrp` and `casper`. `validateMemo` and `buildTonTransaction` now report a malformed comment instead of throwing, so a bad patch degrades to a validation error rather than a hang. Four regression tests cover it.

Follow-ups filed separately: the bridge-validation catch that swallows throws into a silent dead end (affects all families), TON's missing memo `maxLength`, the missing `errors.RecipientRequired.title` on desktop, and TON absent from the desktop e2e memo list.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34825](https://ledgerhq.atlassian.net/browse/LIVE-34825?atlOrigin=eyJpIjoiMTk5NTA0ZmFiNDJmNDc4ZTkxMjEwNmViOGI3NzYxZjciLCJwIjoiaiJ9)
- **ADR** (if any): 

[LIVE-34825]: https://ledgerhq.atlassian.net/browse/LIVE-34825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/bridge/descriptor/send/memo.ts
+++ b/libs/ledger-live-common/src/bridge/descriptor/send/memo.ts
@@ -39,7 +39,8 @@ const memoApplicationRegistry: Record<string, MemoApplicationFn> = {
     return {
       comment: {
         ...currentComment,
-        text: memo,
+        // ton stores the memo as a required string, so a cleared memo is "" and never undefined
+        text: memo == null ? "" : String(memo),
       },
     };
   },
--- a/libs/ledger-live-common/src/bridge/descriptor/send/memo.ts
+++ b/libs/ledger-live-common/src/bridge/descriptor/send/memo.ts
@@ -40,7 +40,7 @@ const memoApplicationRegistry: Record<string, MemoApplicationFn> = {
       comment: {
         ...currentComment,
         // ton stores the memo as a required string, so a cleared memo is "" and never undefined
-        text: memo == null ? "" : String(memo),
+        text: String(memo ?? ""),
       },
     };
   },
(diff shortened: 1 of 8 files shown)


ishaba (dismissed):

libs/ledger-live-common/src/bridge/descriptor/send/memo.ts
```diff
@@ -39,7 +39,8 @@ const memoApplicationRegistry: Record<string, MemoApplicationFn> = {
     return {
       comment: {
         ...currentComment,
-        text: memo,
+        // ton stores the memo as a required string, so a cleared memo is "" and never undefined
+        text: memo == null ? "" : String(memo),
```
francois-guerin-ledger: Always confusing to use `==` instead of `===` in this kind of situation.

francois-guerin-ledger (dismissed):

francois-guerin-ledger (approved):

cted-ledger (approved):

PR #20424, opened 2026-08-04
[LWDM] refactor(live-wallet): delegate internals to DDD packages (LIVE-34990)

gre-ledger:
This is an alternative and fallback version of https://github.com/LedgerHQ/ledger-live/pull/20423 

> [!NOTE]
> Surgical counterpart to the #19464 study: **not a single file under `apps/` is touched**.
> `libs/live-wallet` keeps its exact public API and delegates its implementation to the DDD
> packages, which were sitting in `develop` with **zero consumers**.

### 📝 Description

`libs/live-wallet` goes from **6702 to 4019 lines** (−40 %): **−2997 / +285** over 29 files.

Every redirected file was diffed against its DDD counterpart before being replaced by a
re-export shim:

| `libs/live-wallet` | now delegates to |
| --- | --- |
| `cloudsync/{api,cipher,sdk}.ts` | `@shared/cloud-sync` |
| `accountName.ts` | `@domain/entity-account-name/accountName` |
| `walletsync/modules/accountNames.ts` | `accountNamesSyncModule` |
| `walletsync/modules/recentAddresses.ts` + `entities/recentAddresses.ts` | `recentAddressesSyncModule` |
| `walletsync/{createWalletSyncWatchLoop,incrementalUpdates,trustchainLifecyle}.ts` | `@features/platform-wallet-sync` |
| `walletsync/{types,aggregator}.ts` | `@shared/cloud-sync-module` (types, `mapValues`, reconciliation doc) |

What stays, for lack of a DDD target: `store.ts`, `walletsync/modules/accounts.ts` (waiting on
`domain/entity/account`), `walletsync/{types,aggregator,root}.ts` (the ctx-carrying variant),
`addAccounts`, `ordering`, `liveqr`.

#### Consumer side (TLDR): nothing to do

Import paths, exported symbols and types are unchanged. An AST script compared the exports of
every modified file before/after across the three layers: **no export was removed**.

```ts
// unchanged at all 168 call sites
import { CloudSyncSDK } from "@ledgerhq/live-wallet/cloudsync/index";
import walletsync, { liveSlug, DistantState, walletSyncWatchLoop } from "@ledgerhq/live-wallet/walletsync/index";
import { getDefaultAccountName } from "@ledgerhq/live-wallet/accountName";
```

New code should import the DDD packages directly; the shims carry a `Deprecated:` comment
pointing at them.

#### The one new file: `walletsync/ctx.ts` (36 LoC)

The DDD managers are ctx-free, the legacy interface passes a resolution ctx. `ctx.ts` bridges
both directions, which is what lets the `accounts` module keep its ctx while the other two
modules become pure re-exports:

```ts
ignoreCtx(accountNamesSyncModule); // CloudSyncDataManager -> legacy WalletSyncDataManager
bindCtx(root, ctx);                // legacy root -> ctx-free, for the platform helpers
```

### 🔗 Context

- **JIRA / GitHub issue**: LIVE-34990 (follow-up of the #19464 study, LIVE-29793)
- **ADR**: DDD 4-layer (shared / domain / features / apps)

### Behaviour changes to be aware of

Three, all inherited from the DDD implementations rather than introduced here:

1. **cloudsync HTTP now uses native `fetch`** instead of `@ledgerhq/live-network`, so no axios
   retry, no `LedgerAPI4xx/5xx`/`NetworkDown` mapping and no `network` logs on that path. The
   watch loop re-polls on its own interval anyway, and both apps test the trustchain error by
   `error.name`, never `instanceof` — `TrustchainOutdated` keeps the same `name`. WebSocket
   notifications still go through `isomorphic-ws`.
2. **`log("walletsync", …)` calls are gone** from the watch loop and the incremental updates,
   `logNonImportedAccountInfos` included. Deliberate: keeping them meant threading a logger
   through `@features/platform-wallet-sync`.
3. **Fewer redundant writes.** In `makeSaveNewUpdate`, `if (event.version !== latestVersion)`
   becomes an `else if`. Previously, when there were changes *and* a version gap, `saveUpdate`
   ran twice with identical `data`/`version`, the second call only re-dispatching the same
   `walletSyncUpdate`.

### Non-obvious changes outside live-wallet

- `@features/platform-wallet-sync`: the watch loop only ever calls `diffLocalToDistant`, so its
  `walletsync` param is narrowed to `Pick<CloudSyncDataManager<…>, "diffLocalToDistant">`. Strictly
  more permissive, and it lets a ctx-carrying manager through without a cast.
- `@domain/entity-recent-addresses`: `toDistantState`, `toState` and `DistantRecentAddressesState`
  become exported (the live-wallet walletsync mocks need them).
- `@domain/entity-account-name`: adds a `./accountName` subpath export, so the 37 legacy call
  sites don't drag RTK + immer's `enableMapSet()` side effect along.
- `knip.json`: `cloudsync/{api,cipher}.ts` and `walletsync/entities/recentAddresses.ts` were only
  reachable through the internal imports this PR removes. They are declared as `entry` since they
  are genuine public entry points — `cloudsync/api` is imported directly by both apps.

### Checks

Typecheck: LLD, LLM, web-tools, wallet-cli, e2e desktop, e2e mobile, live-common, plus the 5 DDD
packages. Build: `live-wallet` (CJS + ESM emit inspected), `web-tools`. Desktop CI gates:
`format:check`, `lint:ci`, `lint:guardrails`, `knip-check` (exit 0). `knip` on the 5 DDD packages.

Tests — `nx run-many -t coverage` on the 6 touched packages, i.e. the exact CI command: **995 green**
(live-wallet 839, account-name 56, recent-addresses 44, platform-wallet-sync 36, cloud-sync 20).
Consumers: LLD 604 (79 suites, whole WalletSync feature), LLM 460 (89 suites), live-common 845.

live-wallet's `generic` / `compatibility` / `typecoverage` suites run the aggregator end to end and
now go through the DDD modules — the strongest evidence of behavioural equivalence.

### Known limitations

- `libs/live-wallet` depending on `features/` inverts the intended DDD direction. Accepted as a
  transitional bridge: the point is to stop duplicating, and this package is on its way out.
- `@domain/entity-{starred-account,wallet-sync}` stay unconsumed. Their RTK slices target the
  app-side reducer migration, which is out of scope here since `apps/` is untouched.
- `apps/web-tools`'s `AppCloudSync.test.tsx` and `AppMockEnv.test.tsx` do not run — pre-existing on
  `develop` and never executed in CI (`test-libs-reusable.yml` only walks `find ./libs`, and the
  web-tools workflow runs build + typecheck + format only). Four stacked causes: a stale
  `@ledgerhq/live-env` mock path, a stale `@ledgerhq/coin-module-framework/mocks/account` mock path
  (the npm package ships no `lib/mocks/`, and the source imports `@ledgerhq/ledger-wallet-framework`),
  a missing `transformIgnorePatterns` for the ESM-only `@ledgerhq/lumen-ui-react`, and an
  `AppCloudSync` assertion on a 🎲 button the component no longer renders. Left alone on purpose;
  worth its own ticket.


gre-ledger:
## 📦 Triggered builds

Dispatched on `ledger-live-build` for `refactor/live-wallet-LIVE-34990-redirect-to-ddd-packages` (commit 293ccc9):

| Platform | Run |
|---|---|
| Desktop (macOS / Linux / Windows) | [run 30928834401](https://github.com/LedgerHQ/ledger-live-build/actions/runs/30928834401) |
| Android (APK) | [run 30928837208](https://github.com/LedgerHQ/ledger-live-build/actions/runs/30928837208) |
| iOS | [run 30928840636](https://github.com/LedgerHQ/ledger-live-build/actions/runs/30928840636) |

Artifacts will be attached to each run on completion.

All three platforms are built even though no `apps/` file is touched: `libs/live-wallet` is consumed identically by Desktop and Mobile, and this PR swaps the cloudsync HTTP transport from `@ledgerhq/live-network` to native `fetch`. Worth exercising manually on both, on Ledger Sync specifically — pair two instances, rename an account, check it propagates, and confirm recent addresses and hidden-token filtering still behave.

PR #20442, opened 2026-08-05
[LWDM] feat(generic-coin-framework): add family hooks and fee telemetry

ishaba:
### 📝 Description

The generic coin framework has no way for a family to carry chain-specific transaction fields, which is what blocks TRON — and every family after it — from moving off a dedicated bridge. This adds that capability to the shared layer only; no family consumes it yet and the allowlist is untouched, so the five migrated families behave exactly as before.

Chain-specific fields travel in one designated slot rather than widening `GenericTransaction` per family. The alternative — forwarding every unrecognised transaction field — puts arbitrary values into a field that is persisted, so a non-JSON-safe value reaches storage.

Extra fee data travels on the fee-estimation channel, per [ADR-050](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7341375672/Coin+modules+-+ADR+050+-+FeeOption+abstraction). Putting it in dedicated transaction-status fields is that ADR's rejected option, and TRON is one of its two target chains, so this has to be right before Tronify builds on it.

Account sync can now get chain state from the family, asynchronously. This is temporary: [LIVE-32774](https://ledgerhq.atlassian.net/browse/LIVE-32774) replaces it with the framework's own `getAccountInfo`, and the hook retires with it.

One correctness fix rides along, outside the hooks: the optimistic operation formats its fee with `toFixed()` instead of `toString()`, because a large enough fee reaches `BigInt()` in exponential notation and throws.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35482](https://ledgerhq.atlassian.net/browse/LIVE-35482)
- **ADR** (if any): [ADR-047 ](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7289700580/Coin+modules+-+ADR-047+-+Aleo+transaction+intents+root+fee)(framework vs. local) · [ADR-050](https://ledgerhq.atlassian.net/wiki/spaces/CF/pages/7341375672/Coin+modules+-+ADR+050+-+FeeOption+abstraction) (fee options descriptor)
- **Consumer**: [LIVE-34994](https://ledgerhq.atlassian.net/browse/LIVE-34994) migrates coin-tron onto these hooks and flips the allowlist.

[LIVE-35482]: https://ledgerhq.atlassian.net/browse/LIVE-35482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-32774]: https://ledgerhq.atlassian.net/browse/LIVE-32774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-coin-framework/types.ts
+++ b/libs/ledger-live-common/src/bridge/generic-coin-framework/types.ts
@@ -9,6 +9,21 @@ import type { GetAddressFn } from "@ledgerhq/ledger-wallet-framework/bridge/getA
 import type { SignerContext } from "@ledgerhq/ledger-wallet-framework/signer";
 import BigNumber from "bignumber.js";
 
+/**
+ * Values that are JSON-serializable. Used for the parts of a transaction that are persisted
+ * verbatim, so a `BigNumber`, `bigint` or class instance is rejected at the point it is written
+ * rather than silently corrupted on revival.
+ */
+type JsonSafe =
+  | string
+  | number
+  | boolean
+  | null
+  | JsonSafe[]
+  | { [key: string]: JsonSafe | undefined };
+
+export type JsonSafeRecord = { [key: string]: JsonSafe | undefined };
+
 type NetworkInfo = {
   fees: BigNumber;
 };
@@ -89,6 +104,27 @@ export type GenericTransaction = TransactionCommon & {
   valId?: string;
   withdrawId?: string;
   dstValAddress?: string;
+  /**
+   * Chain-specific transaction fields with no generic equivalent; the owning family maps them onto
+   * the intent's `TxData` via `BridgeApi.buildIntentData` (ADR-047). `JsonSafe` is load-bearing: the
+   * bag round-trips through `GenericTransactionRaw` verbatim; a non-JSON value corrupts on revival.
+   *
+   * Holds user inputs only. Chain-computed values travel in `feeParameters`, so nothing here has to
+   * be merged or invalidated between fee estimations.
+   */
+  familySpecificData?: JsonSafeRecord;
+  /**
+   * Fee telemetry the coin module returned in `FeeEstimation.parameters`; per ADR-050 extra fee
+   * data travels on the fee-estimation channel, not as bespoke transaction-status fields.
+   *
+   * Derived, so absent from `GenericTransactionRaw` — reviving it is how a stale figure survives
+   * a restore. `prepareTransaction` treats it as part of the transaction's identity, so it is
+   * recomputed even when the fee value has not moved.
+   *
+   * Typed exactly as its source so no assertion can lie; values are commonly `bigint`, hence not
+   * `JsonSafeRecord`.
+   */
+  feeParameters?: Record<string, unknown>;
 };
 
 export type GenericTransactionRaw = TransactionCommonRaw & {
@@ -121,6 +157,10 @@ export type GenericTransactionRaw = TransactionCommonRaw & {
   valId?: string;
   withdrawId?: string;
   dstValAddress?: string;
+  /** @see GenericTransaction.familySpecificData — carried verbatim, hence the identical type. */
+  familySpecificData?: JsonSafeRecord;
+  // `feeParameters` is deliberately absent: derived from the last fee estimation, so persisting it
+  // is how a stale figure survives a restore; the next `prepareTransaction` recomputes it.
 };
 
 export interface OperationCommon extends Operation {
@@ -144,6 +184,18 @@ export type CoinFrameworkSigner<S = unknown> = {
 export type AccountRawAssignHooks = {
   assignFromAccountRaw?: AccountBridge<GenericTransaction>["assignFromAccountRaw"];
   assignToAccountRaw?: AccountBridge<GenericTransaction>["assignToAccountRaw"];
+  /**
+   * Revive/serialize the family-owned part of `Operation.extra` (forwarded through
+   * `Operation.details.familyExtra`). Without these the bag is persisted verbatim, so a non-JSON
+   * value corrupts on revival. `AccountBridge` members on a different call path from the `assign*`
+   * pair — `fromSignedOperationRaw` has no account in hand.
+   *
+   * The serialization layer calls these with the **whole** `Operation.extra` — the family's keys
+   * merged with the framework's — and replaces `extra` wholesale with what is returned, so an
+   * implementation that maps only its own keys silently drops every framework-owned key.
+   */
+  fromOperationExtraRaw?: AccountBridge<GenericTransaction>["fromOperationExtraRaw"];
+  toOperationExtraRaw?: AccountBridge<GenericTransaction>["toOperationExtraRaw"];
 };
 
 export type SignTransactionOptions = {
--- a/libs/ledger-live-common/src/bridge/generic-coin-framework/types.ts
+++ b/libs/ledger-live-common/src/bridge/generic-coin-framework/types.ts
@@ -13,6 +13,9 @@ import BigNumber from "bignumber.js";
  * Values that are JSON-serializable. Used for the parts of a transaction that are persisted
  * verbatim, so a `BigNumber`, `bigint` or class instance is rejected at the point it is written
  * rather than silently corrupted on revival.
+ *
+ * `undefined` is permitted so that an optional field typechecks; a round trip drops it, so an absent
+ * key and an `undefined` value have to mean the same thing to every consumer.
  */
 type JsonSafe =
   | string
@@ -107,7 +110,12 @@ export type GenericTransaction = TransactionCommon & {
   /**
    * Chain-specific transaction fields with no generic equivalent; the owning family maps them onto
    * the intent's `TxData` via `BridgeApi.buildIntentData` (ADR-047). `JsonSafe` is load-bearing: the
-   * bag round-trips through `GenericTransactionRaw` verbatim; a non-JSON value corrupts on revival.
+   * bag is written to `GenericTransactionRaw` verbatim, so a non-JSON value corrupts on revival.
+   *
+   * Only the write half is generic. There is no generic raw-to-transaction converter — each family's
+   * own `fromTransactionRaw` enumerates the fields it revives (`families/evm/transaction.ts`) — so a
+   * family adopting this bag has to revive it there, or a flow rebuilding from
+   * `Operation.transactionRaw` crafts a different intent than the one that was signed.
    *
    * Holds user inputs only. Chain-computed values travel in `feeParameters`, so nothing here has to
    * be merged or invalidated between fee estimations.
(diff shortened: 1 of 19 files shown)


ishaba:
> **libs/ledger-live-common/src/bridge/generic-coin-framework/signOperation.ts:113**
> 
> * `describeOptimisticOperation` is wired into `signOperation`, but the raw-signing path (`genericSignRawOperation`) still builds an optimistic operation without consulting the family hook. If a family starts relying on this hook (e.g. to keep the pending op type aligned with sync), `signOperation` and `signRawOperation` can diverge in the optimistic op `type`/`value`.
> 
> It would be safer to either thread the same bridge hook into `genericSignRawOperation` (by fetching `bridgeApi` there too) or explicitly document that the raw-signing path does not support family optimistic-operation description.

Rejecting this one — threading the hook into `genericSignRawOperation` cannot change behaviour. `buildOptimisticOperation` only consults it when the transaction carries a mode (`utils.ts:733`, guarded on `mode !== undefined`), and the raw-signing path has no mode to give it: `signRawOperation` receives an opaque blob string rather than a `GenericTransaction`, and builds its optimistic operation from the stub `{ family, amount: 0, recipient: "" }` (`signRawOperation.ts:63-67`). Passing `bridgeApi.describeOptimisticOperation` in would short-circuit on that guard every time.

The `type`/`value` divergence is also pre-existing rather than introduced here: raw signing has always reported amount `0` and an empty recipient, so its optimistic operation has never described the real transaction. That gap is mode-shaped, not hook-shaped — closing it means giving the raw path a mode, which is a separate change.

libs/ledger-live-common/src/bridge/generic-coin-framework/utils.ts
```diff
@@ -545,8 +562,12 @@ export function transactionToIntent(
   }
 
   if (!transaction.data || transaction.data.length === 0) {
-    const resolvedCraftTransactionData = craftTransactionData ?? defaultCraftTransactionData;
-    res.data = resolvedCraftTransactionData(res);
+    if (buildIntentData) {
+      res.data = buildIntentData(transaction);
+    } else {
```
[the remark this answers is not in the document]
ishaba: The premise doesn't hold here. `GenericTransaction` is a `type` alias over `TransactionCommon`, itself a `type` (`generic-coin-framework/types.ts:77`, `types-live/src/transaction.ts:109`), and TypeScript grants object *type aliases* an implicit index signature — so they satisfy `Record<string, unknown>`. The "no string index signature" objection is correct for an `interface`, which neither of these is. `nx build @ledgerhq/live-common --skip-nx-cache` compiles clean, so this call is not failing.

I'd rather not add the cast: it would suppress a check that passes, and `@typescript-eslint/consistent-type-assertions` flags assertions in this package precisely because they hide real type errors later.

francois-guerin-ledger (commented):
Good extension of the generic-coin-framework — the hook decomposition is clean and the ADR references give good traceability. A few issues worth addressing before merge.

---

### 🔴 Critical

**1. `getDeviceSignOptions` spread last — family can silently override `derivationMode`**

`signOperation.ts:88`:
```ts
const txnSig = await signer.signTransaction(derivationPath, unsigned, {
  ...transaction.recipientDomain,
  derivationMode: account.derivationMode,
  ...bridgeApi.getDeviceSignOptions?.(transaction, account),  // wins over derivationMode
});
```

The doc says "a wrong signing option fails loudly at the device". In practice, some device apps accept an unexpected derivation mode silently and sign with the wrong key — there is no loud failure. `derivationMode` is a framework invariant and should be pinned **after** the family spread:

```ts
const familyOptions = bridgeApi.getDeviceSignOptions?.(transaction, account) ?? {};
const { derivationMode: _ignored, ...safeOptions } = familyOptions;
const txnSig = await signer.signTransaction(derivationPath, unsigned, {
  ...transaction.recipientDomain,
  ...safeOptions,
  derivationMode: account.derivationMode,  // always wins
});
```

---

**2. `fromOperationExtraRaw`/`toOperationExtraRaw` — silent key-drop footgun not protected**

The doc comment acknowledges it: "an implementation that maps only its own keys silently drops every framework-owned key." But acknowledging a footgun is not the same as preventing it. A family implementing `fromOperationExtraRaw: ({ frozenAmount }) => ({ frozenAmount })` silently drops `ledgerOpType`, `memo`, etc., causing history/rendering bugs.

The framework should enforce a merge rather than a wholesale replace, e.g. in `accountBridge.ts` when wiring the hook:

```ts
fromOperationExtraRaw: (extra) => {
  const familyPart = familyFromOperationExtraRaw?.(extra) ?? {};
  return { ...extra, ...familyPart };  // framework keys survive
},
```

This makes it impossible for family implementations to accidentally lose framework keys.

---

**3. `buildAccountShape` has no error guard — any throw fails the whole sync**

`getAccountShape.ts`:
```ts
const chainSpecificShapePromise = buildShape
  ? Promise.resolve(coinModuleApi.getAccountInfo?.(address)).then(accountInfo =>
      buildShape(address, accountInfo),
    )
  : Promise.resolve(undefined);
```

Compare with `readinessPromise` and `validatorsPromise` — both have explicit `.catch()`. If `getAccountInfo` or `buildShape` rejects (network blip, malformed data), `Promise.all` rejects and sync fails for every account on that chain. Should degrade gracefully:

```ts
const chainSpecificShapePromise = buildShape
  ? Promise.resolve(coinModuleApi.getAccountInfo?.(address))
      .then(accountInfo => buildShape(address, accountInfo))
      .catch(e => {
        log("generic-coin-framework", "buildAccountShape failed", { error: e?.message });
        return undefined;
      })
  : Promise.resolve(undefined);
```

---

### 🟠 Moderate

**4. `familyExtra` key not typed on `CoreOperation.details`**

`utils.ts`: `const FAMILY_EXTRA_DETAILS_KEY = "familyExtra"`. Since `CoreOperation.details` is `Record<string, unknown>`, there is no typed `familyExtra` field — a coin module author has no way to discover this contract without reading framework internals. Suggest adding it as an optional typed field on `OperationDetails` in `@ledgerhq/coin-module-framework/api/types`.

---

**5. `buildIntentData` replaces `craftTransactionData` for ALL modes once declared**

```ts
if (buildIntentData) {
  return buildIntentData(transaction);  // replaces craftTransactionData for every mode
}
```

Once declared, a bug in `buildIntentData` silently breaks plain sends and every other mode for that family. Consider allowing the hook to return `undefined` to fall through to `craftTransactionData` for modes it doesn't own. At minimum, document the total-replacement behaviour explicitly in the JSDoc.

---

**6. `isEqual` on `feeParameters` containing `bigint`**

`prepareTransaction.ts`: `isEqual(transaction.feeParameters, nextFeeParameters)`. The doc says values are "commonly `bigint`". Lodash `isEqual` compares `bigint` values by reference (not value) in versions before 4.17.21. If the monorepo pins an older lodash, the identity check fails spuriously and `prepareTransaction` returns a new object on every call, degrading performance. Worth verifying the lodash version, or using a custom comparator.

---

### 🟡 Minor

- `fees.toString()` → `fees.toFixed()` in `buildOptimisticOperation` is a good correctness fix (avoids exponential notation that `BigInt()` rejects) but it's undocumented in the PR description — worth a callout.
- `!rawSequence?.isInteger()` simplification is safe and correct.
- No test coverage for `buildAccountShape` throwing — especially important given finding #3.

ishaba:
> A few issues worth addressing before merge.

Thanks — went through each one. Summary: 1 fixed, 3 rejected with reasons, 1 follow-up, plus the missing test.

---

### 🔴 Critical

**1. `getDeviceSignOptions` spread last — ✅ fixed**

You're right, and the JSDoc rationale was wrong: `families/tezos/signer.ts` picks its signing **curve** from `options.derivationMode`, so a family-supplied value signs with the wrong curve and the device accepts it — no loud failure. Family spread now comes first, `derivationMode` pinned last. JSDoc corrected, test added.

**2. `fromOperationExtraRaw`/`toOperationExtraRaw` key-drop — ✅ fixed, with one adjustment**

Agreed, and we already do exactly this on the sync path (`utils.ts`: `extra: { ...familyExtra, ...frameworkExtra }`). Your snippet needed one more step on the serialization path though: the two sides are in different representations, and the framework's own `extra.stake.amount` is a `BigNumber` — a blind `{ ...extra, ...familyPart }` in the `to` direction would persist it as `{s,e,c}`, the corruption these hooks exist to prevent.

So the framework now converts its own keys and spreads the family's result over them, wired in `accountRawAssign.ts`. Family implementations can map only their own keys; `ledgerOpType`, `memo` and `stake` can no longer be dropped. Wrapping is applied only when a family declares a hook, since `toOperationRaw` gates on presence and persists verbatim otherwise. Tests in `tests/accountRawAssign.test.ts`.

**3. `buildAccountShape` error guard — ❌ keeping it uncaught, but documenting why**

The asymmetry you spotted was real and undocumented, so thanks. `validators`/`readiness` are the framework's *own* optional enrichments; what `buildAccountShape` returns is the family's contract, and the framework can't tell whether those fields are load-bearing (tron's `tronResources` gates `isAccountEmpty` and the staking UI) or cosmetic.

That asymmetry is why the default is to propagate: a framework-level catch imposes silent degradation on every family with no way to opt back into strictness, whereas a family whose contribution *is* optional can catch inside its own hook and return `undefined` — a path the framework already supports and tests. Both policies stay available under the transparent default; only one is available under a catching one.

On the failure mode itself: a rejection leaves the last good account in place and the next poll retries, whereas swallowing persists an account whose frozen balance, votes and rewards read as absent. Also worth noting that for tron the blip scenario is already covered — `getBalance` hits the same endpoint and is uncaught — so the catch would only change the narrow case where the *extra* resource calls fail while the balance call succeeds, which is exactly the case where degrading is worst.

Both facts are now in the code: the reasoning at the call site, and the contract on the hook's JSDoc ("a rejection fails the sync; return `undefined` for a contribution the account is correct without"). Plus a test pinning the propagation.

---

### 🟠 Moderate

**4. Typed `familyExtra` on `OperationDetails` — ⏭ follow-up**

Agreed, but `OperationDetails` lives in the published `@ledgerhq/coin-module-framework`, not this repo — needs its own coin-modules PR.

**5. `buildIntentData` total replacement — ❌ by design**

The replacement is total because the two hooks serve different callers. `craftTransactionData` answers callers that build an intent themselves — coin-tester, scripts, coin-service — and coin-tron's implementation preserves whatever `data` they supplied. `buildIntentData` answers the wallet, where the input is a wallet transaction and only live-common knows its shape (`familySpecificData`).

On the wallet path, then, there is no mode where falling back would help: deriving `data` from something that has never seen `familySpecificData` yields an intent with the family fields missing — the silent drop ADR-047's wiring exists to prevent, and what bit us in the wallet-api converters on the consumer branch (a freeze arriving with no resource).

An `undefined` fall-through would also reintroduce the footgun you objected to in Critical №2: a family that forgets a mode would quietly get the coin module's shape-blind default instead of its own mapping. Total replacement keeps that ownership complete, and a family bug surfaces loudly — the coin module's `validateIntent` rejects a freeze with no resource before it can be signed.

The behaviour is documented on the hook: "present, it replaces that call for every mode — plain sends included."

**6. `isEqual` on `bigint` — ❌ rejected**

`ledger-live` is on lodash 4.17.21, and `isEqual` compares `bigint` by value regardless. Verified: `isEqual({a:1n},{a:1n})` → true, `{a:1n}` vs `{a:2n}` → false.

---

### 🟡 Minor

- `fees.toFixed()` — good catch, called out in the PR description now.
- `!rawSequence?.isInteger()` — agreed, no action.
- ✅ Added the missing test: a rejecting `buildAccountShape` fails the sync

libs/ledger-live-common/src/bridge/generic-coin-framework/types.ts
```diff
@@ -9,6 +9,24 @@ import type { GetAddressFn } from "@ledgerhq/ledger-wallet-framework/bridge/getA
...
+ */
+export type JsonSafe =
+  | string
+  | number
+  | boolean
+  | null
+  | JsonSafe[]
+  | { [key: string]: JsonSafe | undefined };
+
+export type JsonSafeRecord = { [key: string]: JsonSafe | undefined };
```
henri-ly, line 28: It look good and wondering if it's a pattern we want to reproduce on other coin ? WDYT @francois-guerin-ledger  ?

henri-ly (dismissed):

libs/ledger-live-common/src/bridge/generic-coin-framework/types.ts
francois-guerin-ledger: Coin modules deal with `bigint`s or `BigNumber`s to it won't work. A type we can narrow at runtime like `unknown` and or `Record<string, unknown>` gives more flexibility.

francois-guerin-ledger (dismissed):

henri-ly (approved):

francois-guerin-ledger (approved):

PR #20464, opened 2026-08-05
fix(swap): address RTK Query migration review findings (LIVE-35519)

philipptpunkt:
## What

Follow-up to **[#18764]** (`[LWDM] refactor(swap): migrate fetchQuotes to RTK Query`, [LIVE-32805]), which was approved and merged before an AI-assisted review raised a batch of findings. Every one was verified against the code before being acted on; all were valid. They land here rather than in the merged PR.

Tracked by [LIVE-35519].

## Fixes

**Concurrent requests could cross headers.** RTK Query's `condition` returns early on `status === "pending"` **before** `forceRefetch` is consulted (`@reduxjs/toolkit@2.11.2`, `dist/query/rtk-query.modern.mjs`), so two in-flight calls sharing a cache key are deduped regardless. `customHeaders` were excluded from the key, so two concurrent `/quote` calls differing only by header shared one request and the second caller received a body fetched with the first caller's headers. `headers` is arbitrary live-app input, so that is not ours to assume harmless. The key now carries a digest of the headers — entries stay separate, and the token still never reaches redux state. Both new tests fail if the fix is reverted.

**Aggregator errors were invisible.** Any HTTP error resolves to an empty result, which the caller surfaces as "no quotes" — indistinguishable from a genuinely empty result — and the migration dropped the axios interceptor's `log("network-error", …)`. Restored, so an outage is still visible in support logs. Note this is broader than a 5xx: `@ledgerhq/live-network` installs its interceptors on the *global* axios (`libs/live-network/src/network.ts:105-106`), so before the migration a `400`, `401` or `403` reached the live app as `LedgerAPI4xx`. **Whether that should stay "no quotes" is a product call for PTX** — this PR only restores the signal.

**No `api.test.ts` for the swap service** in `@shared/api-services`, which all four siblings have and the package README prescribes. It matters more here: this is the first consumer of `createAuthenticatedBaseQuery` outside `shared/auth`, so the test covers `prepareHeaders` surviving *both* branches of the wrapper, plus `swapApiExtra`'s fail-fast validation.

**`globalThis` dispatch singleton replaced with DI.** `dispatch` now travels through `GetQuotesContext`, the typed seam every caller of `getQuotes` already builds. `setSwapQuotesStore` / `getSwapQuotesDispatch` / `resetSwapQuotesStore` are gone; `setupStandaloneSwapQuotesStore` returns its store instead of registering one globally. A host that registers the reducer but forgets the dispatch is now a compile error rather than a runtime throw on the first quote — which immediately caught a **second wallet-cli call site** (`commands/swap/execute.ts`) that the previous design would have left to fail in production.

**Support-log exposure.** Desktop's redux logger copied `meta.arg` verbatim for `*/executeQuery/*` actions, so live-app headers and the user's send/receive addresses landed in the user-exportable log. Redacted at the logger, which covers any future address-bearing endpoint rather than just this one.

**Fixture duplication** between `@domain/api-swap-quotes` and live-common, caused by the package `exports` map having no `./fixtures` subpath. Added; live-common now uses the schema-parsed originals, so the fixtures cannot drift from the contract.

**`workspace:^` → `workspace:*`**, matching every sibling domain package.

## Deliberately not fixed

- **`SWAP_API_BASE` resolves once at store construction** rather than per request, so a runtime env change needs a restart. Real, and now documented — but making it dynamic would diverge from all five services in `@shared/api-services`. Worth a follow-up across the whole package if the dev/QA workflow needs it.
- **Per-poll schema re-parse** used only to emit a warning. Raised as explicitly non-blocking; revisit when the schema is tightened to drop invalid rows.

## Tests

| | |
| --- | --- |
| `@domain/api-swap-quotes` | 35 |
| `@shared/api-services` | 39 (swap service: 9 new) |
| live-common quotes + Exchange server | 233 |
| Desktop logger + reducer paths | 10 |
| Mobile reducer paths | 2 |

Also verified: module boundaries ✓, live-common `unimported` exit 0, desktop typecheck ✓, wallet-cli typecheck ✓, `pnpm install --frozen-lockfile` ✓ against a 2-line lockfile diff, oxfmt clean.

Mobile's `tsc` reports 890 errors on this branch and **890 on the untouched base**, none in files this PR touches.

[#18764]: https://github.com/LedgerHQ/ledger-live/pull/18764
[LIVE-32805]: https://ledgerhq.atlassian.net/browse/LIVE-32805
[LIVE-35519]: https://ledgerhq.atlassian.net/browse/LIVE-35519

---

## Follow-up round (this session)

Merged `develop` (223 commits behind; one conflict in mobile `configureStore.ts`, where develop's `setupRecentAddressesStore` → `connectRecentAddressesStore` swap met this branch's `setSwapQuotesStore` removal — both sides kept, the removed import dropped). `CONTRIBUTING.md` prefers rebase for a branch like this; the merge was kept because the branch is already reviewed and pushed.

**SonarCloud quality gate** failed on `D Reliability Rating on New Code`, from two findings in `domain/api/swap-quotes/src/api.ts`:

- `S2871` (critical bug) — bare `.sort()` in `hashCustomHeaders`. Now takes an explicit deterministic comparator, written to avoid `S3358` (nested ternary) in turn.
- `S7744` — redundant `?? {}` before the header spread.

Fixing the first surfaced a real problem in the same function. The canonical string joined `name:value` pairs with a newline, so `{ a: "1\nb:2" }` and `{ a: "1", b: "2" }` produced the same string, and so the same digest. The live app picks `customHeaders`, so it could have collided the digest deliberately and got back exactly the cross-header dedupe the SHA-256 change exists to prevent. Each pair is now JSON-encoded, which escapes the separator. Test fails if reverted.

**Desktop log redaction** was hardened after a pre-push review:

- It failed *open* past the recursion limit, returning an uninspected subtree into a user-exportable file. It now truncates. One debuggability cost: a benign value deeper than the limit now reads `[truncated]`.
- It walked every object through `Object.entries`, flattening a `Date` to `{}` and a `BigNumber` to its internals — for every RTK endpoint's args, not just swap's. Non-plain objects are now handed back intact, but only once the walk confirms they carry nothing redacted, so a secret cannot survive by sitting on a class instance. Tracking that explicitly rather than by reference identity matters: an array is always rebuilt, and a `BigNumber` holds one.
- Key list widened beyond the swap shape (`authorization`, `accessToken`, `token`, `address`, `freshAddress`, `xpub`).

Also: the package README still documented the removed `setSwapQuotesStore` / `getSwapQuotesDispatch` and omitted the new `./fixtures` subpath; the changeset omitted `@domain/api-swap-quotes`; the unreleased changeset from #18764 still described `setSwapQuotesStore` even though both ship in the same release; and the bare `as never` dispatch casts across the quotes suites were replaced with one asserted helper per file, restoring the `eslint-disable` convention they had dropped.

### Verification

`@domain/api-swap-quotes` 38 pass; live-common quotes service 32 pass; oxfmt clean; `@domain/api-swap-quotes` `tsc --noEmit` clean; no `tsc` error in any file this branch touches (live-common and desktop both report pre-existing unbuilt-dependency errors elsewhere).

Not run locally: the desktop jest suite — `@ledgerhq/react-ui` does not build in this worktree. The two new `redactQueryArg` tests were verified by executing the function standalone against the same assertions, and by a reviewer confirming each fails if and only if its own fix is reverted. CI runs the suite.

`Libraries Codecheck` and `Codecheck Mobile` fail on a `tiny-secp256k1`/`nan` native build; `Libraries Codecheck` fails the same way on `develop`, so it is not from this branch.

REVIEW OUTCOME: pass


philipptpunkt:
The FNV-1a collision concern was addressed before this comment could be actioned, in `824849cf2af` — the digest is SHA-256 (`@noble/hashes`), and `api.test.ts` pins both the full 64-hex length and the exact digest so a downgrade to a truncated or non-cryptographic hash fails the suite.

Reviewing it also turned up a second way to collide the digest that the hash change alone did not close: the canonical string joined `name:value` pairs with a newline, so `{ a: "1\nb:2" }` and `{ a: "1", b: "2" }` hashed identically. Since the live app chooses `customHeaders`, that was craftable on purpose. Each pair is now JSON-encoded, which escapes the separator (`5e70c75591c`), with a test that fails if reverted.

PR #20473, opened 2026-08-05
[LWDM] fix(lwdm): fix sanctionned ens address check

dilaouid:
### 📝 Description

Previously, recipient sanction checks ran against the raw ENS domain instead of the resolved address, so sanctioned destinations could be missed. This is revalidating sanctions whenever the effective address changes including after ENS resolution and uses the parent chain currency for token sends

### 🔗 Context

- **[JIRA / GitHub issue](https://ledgerhq.atlassian.net/browse/LIVE-35547)** 
- **ADR** (if any):

### DIFF
--- a/.changeset/smooth-years-divide.md
+++ b/.changeset/smooth-years-divide.md
@@ -1,0 +1,6 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
+---
+
+fix(lwdm): fix sanctionned ens address check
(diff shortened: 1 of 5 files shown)


dilaouid:
e2e desktop: https://github.com/LedgerHQ/ledger-live/actions/runs/31090954102

henri-ly (approved):

PR #20487, opened 2026-08-05
feat(support): add the jest-devtools preset package (LIVE-35534)

ysitbon:
### 📝 Description

`devtools/shell`, `devtools/feature-flags` and `devtools/transport-panel` each carried their own copy of the same jest wiring: two config files, two setup files, two React Native mocks, and a themed `@testing-library` `render` helper. Three of those files were byte-identical; the rest had drifted.

This adds **`@support/jest-devtools`** and points all three at it.

```js
// jest.config.js
module.exports = require("@support/jest-devtools").createWebJestConfig();

// jest.native.config.js
module.exports = require("@support/jest-devtools").createNativeJestConfig();
```

Tests import the fixtures for their platform directly — there is no per-package indirection:

```ts
import { render, screen } from "@support/jest-devtools/web";     // *.web.test.tsx
import { render, screen } from "@support/jest-devtools/native";  // *.native.test.tsx
```

Both entry points re-export their `@testing-library` package, so `screen`, `userEvent` and `renderHook` come from the same import. `render` takes a `wrapper` option for an extra provider nested *inside* the `ThemeProvider` (that shape came from feature-flags, the most evolved of the three copies, so no call site changed).

`@devtools/shell` keeps `jest/screens.native.tsx` for `renderWithNavigation` and `renderScreen`, which need its own navigation and shell context; it re-exports the package so shell's tests still need one import. Package-specific fixtures (`jest/test-utils.ts`, `jest/mocks/transport.ts`) stay put, and `feature-flags` ends up needing no `jest/` folder at all.

**Three latent bugs fixed along the way:**

- `feature-flags` had no web render helper, so 10 web test files rendered lumen components with **no `ThemeProvider`**. They now go through the shared fixture.
- `feature-flags` was missing the web `matchMedia` stub and the `act(...)` console filter the other two had.
- `transport-panel` resolved its native render helper only through the react-native preset's haste platform list, which made its `^jest/render\.native$` mapping dead config.

**Three design notes for reviewers**, each found by making it fail first:

- The web preset maps `@ledgerhq/lumen-ui-react` to the consumer's copy (the native preset already did this for `lumen-ui-rnative`). This is a correctness requirement, not just resolution: a second lumen instance would give the fixture's `ThemeProvider` a different React context than the components under test.
- `expo-haptics` is mocked through `moduleNameMapper` rather than `jest.mock`. In a shared setup file the specifier resolves against the *calling* file, so under pnpm it picks a different instance than the consumer's lumen imports — the mock misses and the real ESM file then fails to transform.
- The platform subpath is explicit rather than derived from the filename, because jest's resolver does not apply the `react-native` export condition.

`babel.config.js` deliberately stays in each package: babel resolves its config from the jest `rootDir`, and `babel-jest` needs the React Native preset to parse the Flow-typed files inside `react-native`. Removing it fails every native suite at `react-native/jest/setup.js`.

The presets carry `testTimeout: 30_000`. `transport-panel` already set that per-package; a `userEvent` interaction on a lumen dialog then exceeded the 5s default on CI (which runs `nx --parallel=4` with two jest workers each, under coverage) while taking ~3s locally. Two consumers needing the same override means it belongs in the preset.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35534](https://ledgerhq.atlassian.net/browse/LIVE-35534), under epic [LIVE-34831 — Create base support packages with base configs](https://ledgerhq.atlassian.net/browse/LIVE-34831)
- **ADR**: [Shared Tooling Configuration via `support/` Packages](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7353892916/2026-07-23+ADR+Shared+Tooling+Configuration+via+support+Packages) — this is the jest spike that ADR requires before it can move from Proposed to Accepted, and it names jest as the preset carrying the most complexity, because a jest preset is not a single file: it includes setup files and mocks, which are code.

### 🧱 Stack

1. [#20488](https://github.com/LedgerHQ/ledger-live/pull/20488): the `support/` layer plumbing — workspace glob, `scope:support` tag, docs
2. **#20487 — this PR**: `@support/jest-devtools` + the `devtools/*` migration
3. [#20489](https://github.com/LedgerHQ/ledger-live/pull/20489): move `@features/platform-jest-config` to `@support/jest-features-flow`

### ✅ Verification

Test counts match the pre-change baseline exactly, per project:

| Package | web | native |
|---|---|---|
| `@devtools/shell` | 18 suites / 118 tests | 5 / 18 |
| `@devtools/feature-flags` | 19 / 140 | 12 / 45 |
| `@devtools/transport-panel` | 8 / 48 | 6 / 41 |

Also green: `pnpm nx run-many -t typecheck`, `nx coverage` for the three packages, `pnpm lint:boundaries`, `oxfmt --check`, `pnpm install --frozen-lockfile`, and the changeset validation script.

[LIVE-35534]: https://ledgerhq.atlassian.net/browse/LIVE-35534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


support/jest-devtools/index.js
```diff
@@ -0,0 +1,119 @@
...
+    transformIgnorePatterns: [`node_modules/(?!(${NATIVE_TRANSFORM_ALLOWLIST.join("|")})/.)`],
+    modulePaths: ["<rootDir>"],
+    moduleNameMapper: {
+      // resolved in the consumer: each devtools package declares lumen itself
+      "^@ledgerhq/lumen-ui-rnative$":
+        "<rootDir>/node_modules/@ledgerhq/lumen-ui-rnative/src/index.ts",
+      "^@ledgerhq/lumen-design-core$": "<rootDir>/node_modules/@ledgerhq/lumen-design-core",
+      "^@sbaiahmed1/react-native-blur$": here("mocks/react-native-blur.tsx"),
+      "^react-native-worklets$": here("mocks/react-native-worklets.js"),
+      "^expo-haptics$": here("mocks/expo-haptics.ts"),
```
mcayuelas-ledger, line 109: nice :)

mcayuelas-ledger (approved):
LGTM

Sebastien-Dav1d (approved):

PR #20513, opened 2026-08-06
fix(buy): navigate back on account drawer cancel (LIVE-26867)

vpenskyi-ledger:
### 📝 Description

**Previous behaviour:** When a user has zero accounts for a currency and opens the Buy/Sell screen, the Exchange app immediately requests account selection via a drawer. If the user closes the drawer without selecting an account, the Exchange screen is left with an infinite loader — there is no way to recover without force-quitting the app.

**Root cause:** The `account.request` wallet-api handler in `helpers.ts` was opening the account selection drawer but never forwarding the `onCancel` callback to `openModularDrawer`. When the drawer was closed without a selection, Exchange received no signal and kept waiting.

**Fix:**
- Added a `cancelCallbackId` mechanism to `ModularDrawer` so `onCancel` callbacks survive the Redux serialisation boundary.
- Added an opt-in flag `goBackOnAccountRequestCancel` (set only by the zero-accounts Buy/Sell navigation path) that triggers `navigation.goBack()` when the drawer is cancelled, returning the user to the previous screen.
- A `shouldGoBackOnCancelRef` flips to `false` after the first successful account selection, preventing unintended back-navigation when the user later changes currency from within Exchange (1-account regression fix).

**Automated checks:** Unit tests cover all three ref-state scenarios (goBack fires, goBack skipped, ref flips on success). An integration test mounts the full `ModularDrawer` stack and asserts the `onCancel` callback is invoked on close and not on account selection.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-26867]

Before fix:
https://github.com/user-attachments/assets/11e87f58-5d0f-4ee2-9c43-8e272a172032

After fix:
https://github.com/user-attachments/assets/c470dc25-a41f-484b-acc3-df8dc09b00bf

[LIVE-26867]: https://ledgerhq.atlassian.net/browse/LIVE-26867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
+++ b/apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
@@ -16,7 +16,6 @@ const ASSET_DETAIL_SOURCE_SCREEN_NAME = "Asset Detail";
 type UseOpenBuySellProps = {
   currency?: CryptoOrTokenCurrency;
   sourceScreenName: string;
-  onCancel?: () => void;
 };
 
 type AccountWithParent = {
@@ -42,7 +41,7 @@ function getAccountsForCurrency(
     });
 }
 
-export function useOpenBuySell({ currency, sourceScreenName, onCancel }: UseOpenBuySellProps) {
+export function useOpenBuySell({ currency, sourceScreenName }: UseOpenBuySellProps) {
   const navigation = useNavigation<NativeStackNavigationProp<BaseNavigatorStackParamList>>();
   const shallowAccounts = useSelector(shallowAccountsSelector);
   const flattenedAccounts = useSelector(flattenAccountsSelector);
@@ -86,10 +85,9 @@ export function useOpenBuySell({ currency, sourceScreenName, onCancel }: UseOpen
         enableAccountSelection: true,
         onAccountSelected: (account, parentAccount) =>
           navigateToBuySell(mode, account, parentAccount),
-        onCancel,
       });
     },
-    [currency, openDrawer, sourceScreenName, navigateToBuySell, onCancel],
+    [currency, openDrawer, sourceScreenName, navigateToBuySell],
   );
 
   const handleOpenBuySell = useCallback(
--- a/apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
+++ b/apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
@@ -16,6 +16,7 @@ const ASSET_DETAIL_SOURCE_SCREEN_NAME = "Asset Detail";
 type UseOpenBuySellProps = {
   currency?: CryptoOrTokenCurrency;
   sourceScreenName: string;
+  onCancel?: () => void;
 };
 
 type AccountWithParent = {
@@ -41,7 +42,7 @@ function getAccountsForCurrency(
     });
 }
 
-export function useOpenBuySell({ currency, sourceScreenName }: UseOpenBuySellProps) {
+export function useOpenBuySell({ currency, sourceScreenName, onCancel }: UseOpenBuySellProps) {
   const navigation = useNavigation<NativeStackNavigationProp<BaseNavigatorStackParamList>>();
   const shallowAccounts = useSelector(shallowAccountsSelector);
   const flattenedAccounts = useSelector(flattenAccountsSelector);
@@ -68,6 +69,7 @@ export function useOpenBuySell({ currency, sourceScreenName }: UseOpenBuySellPro
           ...(sourceScreenName === ASSET_DETAIL_SOURCE_SCREEN_NAME && {
             returnToPreviousScreenOnClose: true,
           }),
+          ...(!account && { goBackOnAccountRequestCancel: "true" }),
         },
       });
     },
@@ -84,9 +86,10 @@ export function useOpenBuySell({ currency, sourceScreenName }: UseOpenBuySellPro
         enableAccountSelection: true,
         onAccountSelected: (account, parentAccount) =>
           navigateToBuySell(mode, account, parentAccount),
+        onCancel,
       });
     },
-    [currency, openDrawer, sourceScreenName, navigateToBuySell],
+    [currency, openDrawer, sourceScreenName, navigateToBuySell, onCancel],
   );
 
   const handleOpenBuySell = useCallback(
--- a/apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
+++ b/apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
@@ -68,7 +68,7 @@ export function useOpenBuySell({ currency, sourceScreenName }: UseOpenBuySellPro
           ...(sourceScreenName === ASSET_DETAIL_SOURCE_SCREEN_NAME && {
             returnToPreviousScreenOnClose: true,
           }),
-          ...(!account && { goBackOnAccountRequestCancel: "true" }),
+          ...(!account && { goBackOnAccountRequestCancel: true }),
         },
       });
     },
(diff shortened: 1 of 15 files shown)


zaruen (dismissed):

qperrot (dismissed):
LGTM

Justkant (commented):
Thanks for PR, forwarding `onCancel` from `account.request` into `openModularDrawer` is an issue and the fix is clearly needed. Two design questions and one cleanup, none of which dispute the behaviour:

**1. Could this navigation policy live in the Exchange/PTX layer rather than in the generic Wallet API hook?** The route flag plus `shouldGoBackOnCancelRef` makes `account.request` aware of one specific host-navigation flow and encodes the distinction between "the initial request" and "a later currency change" implicitly in a mutable ref. It works today, but the one-shot rule (cancel dismisses Exchange only until the first successful selection) is encoded away from the Exchange host and is easy to miss when modifying this generic handler. Ideally this hook just forwards `onCancel`, while the Exchange host decides whether that cancellation should dismiss the screen.

Related: since `goBackOnAccountRequestCancel` is native navigation policy rather than something the live app consumes, does it need to travel through `inputs` (forcing the `"true"` / `=== "true"` round trip because `WebviewProps.inputs` is `Record<string, string | undefined>`) ? Passing it as a boolean through PTX configuration alongside `returnToPreviousScreenOnClose` would keep that policy in the native layer.

**2. `openDrawer` now treats replacing an existing drawer as cancellation** and executes the previous `cancelCallbackId` synchronously. In this flow, that callback can call `navigation.goBack()` while the replacement drawer is being opened. Is replacement intended to count as user cancellation? Either way, it would be worth documenting and covering with a test. Silently dropping the old callback does not seem right either, since that could leave the previous request unresolved.

**Minor:** `useOpenBuySell` gains an optional `onCancel`, but no call site passes it and it is not part of the zero-account path being fixed. Can it be dropped unless there is an imminent caller?

vpenskyi-ledger:
> Thanks for PR, forwarding `onCancel` from `account.request` into `openModularDrawer` is an issue and the fix is clearly needed. Two design questions and one cleanup, none of which dispute the behaviour:
> 
> **1. Could this navigation policy live in the Exchange/PTX layer rather than in the generic Wallet API hook?** The route flag plus `shouldGoBackOnCancelRef` makes `account.request` aware of one specific host-navigation flow and encodes the distinction between "the initial request" and "a later currency change" implicitly in a mutable ref. It works today, but the one-shot rule (cancel dismisses Exchange only until the first successful selection) is encoded away from the Exchange host and is easy to miss when modifying this generic handler. Ideally this hook just forwards `onCancel`, while the Exchange host decides whether that cancellation should dismiss the screen.
> 
> Related: since `goBackOnAccountRequestCancel` is native navigation policy rather than something the live app consumes, does it need to travel through `inputs` (forcing the `"true"` / `=== "true"` round trip because `WebviewProps.inputs` is `Record<string, string | undefined>`) ? Passing it as a boolean through PTX configuration alongside `returnToPreviousScreenOnClose` would keep that policy in the native layer.
> 
> **2. `openDrawer` now treats replacing an existing drawer as cancellation** and executes the previous `cancelCallbackId` synchronously. In this flow, that callback can call `navigation.goBack()` while the replacement drawer is being opened. Is replacement intended to count as user cancellation? Either way, it would be worth documenting and covering with a test. Silently dropping the old callback does not seem right either, since that could leave the previous request unresolved.
> 
> **Minor:** `useOpenBuySell` gains an optional `onCancel`, but no call site passes it and it is not part of the zero-account path being fixed. Can it be dropped unless there is an imminent caller?

Agree architecturally. Ideally the Exchange host owns this policy. However, the PTX screen has no direct lifecycle hook for account.request cancel, it's buried inside the webview/wallet-api layer. Lifting this out would require a broader refactor of how PTX screens intercept wallet-api events, which is out of scope for this bug fix. Happy to track it as a follow-up. On the inputs string round-trip, fair point, open to passing it as a boolean through PTX config alongside returnToPreviousScreenOnClose if you prefer.

The "fire cancel on replace" was added defensively to avoid hanging previous account.request callers, but you're right that replacement is not user cancellation and synchronously calling navigation.goBack() mid-openDrawer is risky. Will revert that specific behaviour, resetAll() silently drops the previous callback which is safer. Will add a test for the replacement case.

dropping onCancel from useOpenBuySell props, no caller uses it.

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/hooks/useModularDrawerController.ts
```diff
@@ -110,17 +123,20 @@ export const useModularDrawerController = () => {
...
-  }, [callbackId, completionMode, dispatch, executeCurrencyCallback]);
+  }, [callbackId, cancelCallbackId, completionMode, dispatch, executeCurrencyCallback, executeCancelCallback]);
 
   const handleAccountSelected = useCallback(
     (account: AccountLike, parentAccount?: AccountLike) => {
       if (callbackId) {
         executeCallback(callbackId, account, parentAccount);
       }
-      closeDrawer();
+      dispatch(closeModularDrawer());
```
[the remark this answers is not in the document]
mcayuelas-ledger: This comment should be addressed IMO @vpenskyi-ledger

apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
```diff
@@ -68,6 +68,7 @@ export function useOpenBuySell({ currency, sourceScreenName }: UseOpenBuySellPro
           ...(sourceScreenName === ASSET_DETAIL_SOURCE_SCREEN_NAME && {
             returnToPreviousScreenOnClose: true,
           }),
+          ...(!account && { goBackOnAccountRequestCancel: "true" }),
```
mcayuelas-ledger: [ASK] Why don't use boolean directly? This value is used elsewhere for analytics for example?

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/hooks/useModularDrawerController.ts
```diff
@@ -88,9 +93,17 @@ export const useModularDrawerController = () => {
         callbackIdToUse = id;
       }
 
+      let cancelCallbackIdToUse: string | undefined;
+      if (onCancel) {
+        const id = generateCallbackId();
+        registerCancelCallback(id, onCancel);
+        cancelCallbackIdToUse = id;
+      }
+
```
mcayuelas-ledger: In my opinion, we should keep the same pattern as for Account & Currency here and use callbackIdToUse instead of cancelCallbackIdToUse.

Also, since there are now 3 cases, I think it's the right time to reduce openDrawer's complexity by simplifying its content and extracting the core logic into a helper/utility function.

apps/ledger-live-mobile/src/mvvm/features/Buy/index.ts
vpenskyi-ledger: The value flows through webview inputs, which is string-typed by nature (URL params / postMessage), so the coercion has to happen somewhere. Putting "true" at the source and comparing with === "true" at the consumer is explicit and predictable.

That said, you're right that it looks odd in navigation params. A cleaner approach would be to keep goBackOnAccountRequestCancel: boolean in the navigator type and convert to string only when building the webview inputs. Happy to make that change if you prefer

apps/ledger-live-mobile/src/mvvm/features/ModularDrawer/hooks/useModularDrawerController.ts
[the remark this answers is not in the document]
vpenskyi-ledger: Oh, sorry, I implemented that, I just forgot to resolve the comment

mcayuelas-ledger (approved):

zaruen (approved):

Justkant (approved):
Would be good to have a follow up ticket for some of the comments made earlier

PR #20528, opened 2026-08-06
feat(support): add @support/ts-config-root and wire ts-config-base to it

ysitbon:
### 🔗 Context

- **JIRA / GitHub issue**: 
- **ADR** (if any):


ysitbon:
Superseded: ts-config-root was dropped (root tsconfig.base.json now extends @support/ts-config-base-legacy directly). See PR stack #20564.

PR #20540, opened 2026-08-06
[LWDM] chore(aggregated-assets): move platform layer behind shims (LIVE-35227)

LucasWerey:
### 📝 Description

**Problem.** `libs/ledger-live-common/src/dada-client` is being migrated into the DDD layers under [LIVE-35223](https://ledgerhq.atlassian.net/browse/LIVE-35223). [#20345](https://github.com/LedgerHQ/ledger-live/pull/20345) moved the domain layer; this PR moves the platform layer — the React hooks, the cache selectors, and the discovery / currency-selection helpers.

**Solution.** Relocate them into `@features/platform-aggregated-assets`, leaving one-line re-export shims at the old paths. **No consumer's production imports change**, so consumer retargeting stays a separate, independently revertable step (LIVE-35228 / 35229 / 35230). One app *test* did have to change — see commit 5.

### 🧱 The five commits

Ordered so each lands after what it depends on. The fourth is the actual move; the rest are the pieces it needs.

**1 · `feat(interest-rate)` — name the app-facing rate shape** · 6 files

`{ value: number; type: ApyType }` was declared **four times** with no owner: `modularDrawer`'s `ApyProps`, a local `InterestRate` in `getInterestRateForAsset`, an inline type in `ApyIndicatorComponent`, and `useInterestRatesByCurrencies`' return type. Adds `Apy` to the entity that already owns `ApyType` and points all four at it. `ApyProps` survives as an alias because the ApyIndicator components in both apps import that name — retarget in LIVE-35228 / 35229.

**2 · `refactor(aggregated-assets)` — tidy the api surface** · 11 files

Two things the move needs from this package.

`market` splits by visibility: `internals/marketItem.ts` keeps `MarketItemResponse`, the wire shape nothing outside the package names; `market.ts` keeps `PartialMarketItemResponse` and `dadaIdToMarketId`, the two the platform layer needs. See *Why market went public again* below — the part most worth your attention.

`WalletApp` names the `"llm" | "lld"` union that was written out six times. It sits **beside the `product` parameter it describes**, not in a shared package: the apps know which they are, and this only exists because DADA takes it per request. [LIVE-35301](https://ledgerhq.atlassian.net/browse/LIVE-35301) moves the base url to `extraArgument` and could carry this the same way, at which point it drops out of the query args entirely.

**3 · `refactor(modular-drawer)` — relocate `useDrawerConfiguration`** · 4 files

It was misfiled under `dada-client`: it imports only `react` and `@features/platform-feature-flags` and never touches the DADA api or its cache. Moves beside `wallet-api/ModularDrawer/types.ts`, which declares the `ModularDrawerConfiguration` shape it builds and where its three consumers live (both Web3AppWebviews, mobile Stake). Old path shimmed. Deliberately *not* `modularDrawer/hooks/`, which already has a different `useModularDrawerConfiguration`.

**4 · `chore(aggregated-assets)` — move the platform layer behind shims** · 47 files

The move. 9 hooks, 3 cache selectors, discovery and currency selection into the new package; 11 one-line shims left behind. Three things in here are not pure relocation and are called out in the commit body:

- **Tests move with their subjects and 7 needed a real change.** The hooks used to read the endpoint hook from `state-manager/api` but `parseError` and `mergeAssetsDataPages` from *different* modules, so a factory mock of the former left the others intact. Now one module supplies all three, and mocking it wholesale wiped them — the hooks died on `mergeAssetsDataPages is not a function`. Each factory needed `...jest.requireActual(...)`. Assertions untouched.
- **`@shared/env` is mocked in `jest.setup.js`.** The features CI job installs `@features/*...`, `./shared/**` and `./domain/**` (`test-features-reusable.yml:71`), which leaves `libs/env` out, so `@shared/env` can't resolve `@ledgerhq/live-env` and *any* test importing the domain barrel fails to load. Mocked once rather than in eight files.
- **`selectorUtils` names the bucket it indexes**, removing the implicit-any desktop reports on develop. The cross-query cache scan itself is unchanged — reworking it is [LIVE-35301](https://ledgerhq.atlassian.net/browse/LIVE-35301).

**5 · `test(mobile)` — make the global-search defaults fixture real** · 1 file

The one consumer file that changed. It mocked `selectCurrencyForMetaId` through the live-common shim and relied on `selectTopAssetsByCategory` calling it **internally**. Both now live in this package and `discovery.ts` imports `./currencySelection` directly, so no external mock can reach that call. The real selector runs, and the fixture never carried `cryptoOrTokenCurrencies` because the stub made it unnecessary.

**No production behaviour changed, and that is checkable rather than asserted:** `discovery.ts` and `currencySelection.ts` are byte-identical to develop apart from their import lines, and `cryptoOrTokenCurrencies` is a **required** member of `AssetsData` (`types.ts:15`) that the transform and the zero value both always set. The test now covers the half that used to be stubbed out with `{ id: metaId }`.

### 🔍 Why `market` went public again

The one decision I'd like actively agreed rather than skimmed, because it partly reverses @ysitbon's review on #20345.

He asked for `market.ts` to be internal. I measured **zero** cross-package consumers and internalised it. This task creates the first real one: `discovery.ts` and `marketSelectors.ts` need `PartialMarketItemResponse` and `dadaIdToMarketId`, and a `type:feature-platform` package may only depend on `scope:domain`/`shared`/`support` (`constraints.js:42`) — the old `libs/ledger-live-common/src/market` import became illegal and wouldn't resolve under `hoist=false`.

So his own criterion flipped the answer. Rather than flip the whole file, commit 2 splits it along the real boundary, which keeps *"files under internals are not exposed"* literally true. Verified both directions: importing `MarketItemResponse` from the package root fails with `TS2305`; importing the two public symbols compiles.

**The residual problem is not this PR's to fix, but it is now live.** `dadaIdToMarketId` exists twice, byte-identical — here and at `market/utils/index.ts:129` — with 3 new-architecture consumers and 13 legacy ones. Before this PR the domain copy was dead code; now both are in production on separate maintenance paths. And `PartialMarketItemResponse = Partial<MarketItemResponse>` makes **every field optional**, so if the two drift there is no type error, no test failure, nothing. The fix is a real market entity that deletes the duplicate; happy to file it.

### 🧭 What was deliberately *not* touched

- The cache selectors still iterate **every** RTK Query cache entry regardless of query args, walking RTK's internal `state.assetsDataApi.queries → data.pages` layout. That is what makes interest rates and market trend fetched by one query reachable from another. Do not "fix" it here.
- `types/trend.ts` stays a shim rather than retargeting its 8 importers — retarget scope.
- `useAssetDistribution.ts:25` still spells `"llm" | "lld"` inline; it could import `WalletApp` now, but it is a libs consumer.
- `shared/` is untouched by this PR.
- `.unimportedrc.json` has **44 dead entries predating this branch**. Left alone; the diff there is one added line.

### ✅ Verification

- **129** tests in the new package, **84** in `@domain/api-aggregated-assets`, **23** in `@domain/entity-interest-rate`, **87** still in `dada-client`.
- **191 mobile app tests** (34 suites) and **344 desktop app tests** (42 suites) green through the shims. Worth running: several app tests automock the shim paths, and this is where commit 5's case surfaced. Desktop needs `nx run @ledgerhq/react-ui:build` first or every suite dies in `jestSetup`.
- `libs/ledger-live-common` typechecks with **0 errors** under CI's command; all three packages typecheck.
- `nx run enforce-boundaries:lint:boundaries` exits 0 — the check that would catch a `features/ → libs/` import.
- knip clean; `pnpm install --frozen-lockfile` succeeds with a **31-insertion, 0-deletion** lockfile diff.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35227](https://ledgerhq.atlassian.net/browse/LIVE-35227) (epic: [LIVE-35223](https://ledgerhq.atlassian.net/browse/LIVE-35223))
- **ADR**: [DADA DDD compliant](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7382761502/DADA+DDD+compliant)
- **Predecessors**: [#20276](https://github.com/LedgerHQ/ledger-live/pull/20276) · [#20285](https://github.com/LedgerHQ/ledger-live/pull/20285) · [#20345](https://github.com/LedgerHQ/ledger-live/pull/20345)
- **Unblocks**: LIVE-35228 (desktop), LIVE-35229 (mobile), LIVE-35301 (endpoint split)

[LIVE-35223]: https://ledgerhq.atlassian.net/browse/LIVE-35223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-35301]: https://ledgerhq.atlassian.net/browse/LIVE-35301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/wallet-api/ModularDrawer/useDrawerConfiguration.ts
```diff
@@ -0,0 +1,65 @@
...
+export function useDrawerConfiguration() {
+  const earnDrawerConfigurationFlag = useFeature("ptxEarnDrawerConfiguration");
+
+  const createDrawerConfiguration = useCallback(
+    (
+      drawerConfiguration: unknown,
+      useCase: string | undefined,
+      customUseCaseConfigs?: UseCaseConfigs,
+    ): DrawerConfiguration => {
+      const config: DrawerConfiguration | undefined = drawerConfiguration!;
```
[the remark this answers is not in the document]
LucasWerey: Fixed in `ab197a06`.

You are right that it bought nothing — it was also self-contradicting: it asserted non-null and then annotated the target as `| undefined`, while every read below already used `config?.`. Now a plain cast that admits `undefined`, which says the same thing honestly. No runtime change; the 13 tests that pin this hook still pass.

libs/ledger-live-common/src/wallet-api/ModularDrawer/useDrawerConfiguration.ts
```diff
@@ -0,0 +1,65 @@
...
+  assets?: Record<string, unknown>;
+  networks?: Record<string, unknown>;
+}
+
+interface UseCaseConfig {
+  assets?: Record<string, unknown>;
+  networks?: Record<string, unknown>;
+}
+
+type UseCaseConfigs = Record<string, UseCaseConfig>;
```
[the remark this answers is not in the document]
LucasWerey: Right about the duplication, and it is more glaring now that the file sits next to `types.ts` — but I am not taking it in this PR, and the reason is not scope alone.

Both inputs are **unvalidated at runtime**. `drawerConfiguration` arrives as `unknown` from a live app over the wallet-api bridge, and the use-case defaults come from `ptxEarnDrawerConfiguration`'s `params`, which is arbitrary JSON. Declaring the output `ModularDrawerConfiguration` would assert a shape nothing has checked — that just relocates the unsoundness from the assertion you flagged on the other line into a nicer-looking cast.

The sound version is to parse both through `EnhancedModularDrawerConfigurationSchema`, which already exists in that same file. That changes what happens to a malformed config — today it flows through, after it would be dropped or rejected — so it is a behaviour change, and this hook is only passing through here on its way out of the DADA tree (it was misfiled: it never touched the api or the cache).

I have recorded the reasoning as a comment on the local interface so it points at the schema and does not get rediscovered. Happy to file it as a ticket for the ModularDrawer owners if you would rather it were tracked.

LucasWerey:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20540) **Quality Gate failed**
> Failed conditions  [7.3% Duplication on New Code](https://sonarcloud.io/component_measures?id=LedgerHQ_ledger-live&pullRequest=20540&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=LedgerHQ_ledger-live&pullRequest=20540)

Will address it later

mcayuelas-ledger (dismissed):
OK for WXp part

shared/schema-primitives/src/walletApp.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+/**
+ * Which Ledger wallet app is asking: `"llm"` mobile, `"lld"` desktop.
+ * Analytics spells the same idea `"lwm"` / `"lwd"` — these are the values the APIs require.
+ */
+export const WalletAppSchema = z.enum(["llm", "lld"]);
```
gre-ledger: that feels like to be a bit an anti pattern to have such concept (apps know who they are 🤔 we shouldn't have to to maintain such list transversally, eg what about wallet-cli ?)
but in any case i don't think it belongs to shared/schema-primitives , maybe we can have it next to its usage for now 🤔

LucasWerey: yeah I thought it would have been great to have it there but I wasn't sure. I will move it back then

LucasWerey: btw the product is needed by Dada as there is cal versionning by platform

mcayuelas-ledger (approved):

shared/schema-primitives/src/walletApp.ts
gre-ledger: @LucasWerey yes i know, but i still think there is an antipattern to solve at some point but it will need backend iteration indeed (client should request to the backend and the backend should personalize the data based on the client id it presents with, the mobile/desktop info is part of the client id header already, then backend shall abstract away the concept from the app)

gre-ledger (approved):

henri-ly (approved):

shared/schema-primitives/src/walletApp.ts
LucasWerey: @gre-ledger Yep indded

PR #20562, opened 2026-08-07
[LWDM] feat(casper): listOperations (LIVE-33296)

mateuszpalosz-ext:
### 📝 Description

Implements `listOperations` on the Casper Alpaca (`CoinModuleApi`) surface, replacing the
`throw new Error("listOperations is not supported")` stub.

**What it does**

Walks the indexer's `accounts/{publicKey}/ledgerlive-deploys` feed newest-first and maps native
CSPR transfers to framework `Operation`s. The feed also carries staking, bid and contract deploys,
which are filtered out (they have no `target`).

**Option support.** The indexer offers no server-side filter or cursor, so:
- `minHeight` — supported, applied while walking the feed.
- `order` — `"desc"` accepted (the indexer's native order); `"asc"` rejected.
- `cursor` / `limit` — rejected, rather than silently ignored. Emulating them would mean
  re-walking the feed from page 1 on every call. This follows the `ListOperationsOptions`
  contract, which mandates raising "not supported" instead of quietly degrading.

**Supporting changes**

- `fetchTxs` split into `fetchTxsPage(addr, page)` + a thin loop, so `listOperations` can stop
  early instead of always draining the feed.
- Page size fixed: the old code sent `limit=100`, which the indexer **ignores**. The real
  parameter is `page_size`, and the default without it is **10**. Now sends `page_size=250`
  (the maximum). A 49-deploy account went from 5 requests to 1.
- `ITxnHistoryData` trimmed to the fields we actually consume, with the indexer's real behaviour
  documented on each one.
- `mapTxToOps` (legacy bridge path) refactored to share the parsing helpers. Output is unchanged.

**Why `amount`, `status` and `pages` were dropped from the types**

These are type-only changes with no runtime effect, and all removed fields were unreferenced:

| Field | Reason |
|---|---|
| `amount` (top-level) | **Not present in the API response at all.** It was declared as a required `string` — the type was wrong. |
| `pages` (on `IndexerResponseRoot`) | **Not present either.** The root returns only `data`, `item_count`, `page_count`. |
| `status` | Present, but always `"processed"` — including for failed deploys. `error_message` is the actual failure signal. |
| `payment_amount`, `execution_type_id` | Present, unused. |
| `contract_hash`, `contract_package_hash` | Present, unused, and typed `any`. |

### 🔗 Context

- **JIRA / GitHub issue**: [https://ledgerhq.atlassian.net/browse/LIVE-33296](https://ledgerhq.atlassian.net/browse/LIVE-33296)
- **ADR** (if any):


mdomanski-ext-ledger (approved):
LGTM after manual tests (add account, transfer with memo, transfer without memo)

henri-ly (approved):

PR #20570, opened 2026-08-07
feat(contacts): add typed device intent data

deepyjr:
### ✅ Checklist

- [x] **Covered by automatic tests.** Domain and Flow suites cover the persistent wire document, opaque Device Intent data, mutation invariants, and the typed mock port.
      Validate Contacts creation, address addition, contact rename, address-label update, and the device data persisted by each signed mutation.

### 📝 Description

This PR makes the Contacts data model safe to persist through Ledger Sync before the Wallet Sync registration work begins.

It introduces a stable `{ me, contactGroups }` document, typed opaque Device Intent values, and complete remote replacement while preserving unknown future fields. `Me` remains structural in the remote document; `ContactGroup` and `ExternalAddress` identifiers are synchronized. Contacts with addresses require device credentials, and each address persists its device context.

The Cloud Sync implementation is grouped under `src/cloudSync/` and separated into focused wire-schema, state-conversion, and extension-preservation modules. The root `cloudSyncModule.ts` remains a compatibility façade, while the nested module only orchestrates the Cloud Sync contract hooks.

The Contacts entity owns the Device Intent result types because it persists and synchronizes them. A temporary typed mock port is wired into Desktop, Mobile, and shared edit flows so every address addition, signed contact rename, and address-label update stores the same shape returned by the future device boundary. The mock will be replaced by the real DIE implementation before Wallet Sync registration; no Wallet Sync registration is included in this PR.

### ❓ Context

- **JIRA issue**: [LIVE-33959](https://ledgerhq.atlassian.net/browse/LIVE-33959)
- **ADR**: [Contacts Device Intents](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7299039425/ADR+Contacts+Device+Intents#External-Contact-Group-And-Address-Record)

---

- **The code aligns with the requirements** described in the linked Jira issue.
- **The PR description clearly documents the changes** made and explains the temporary mock Device Intent port.

[LIVE-33959]: https://ledgerhq.atlassian.net/browse/LIVE-33959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/domain/entity/contact/src/cloudSyncModule.ts
+++ b/domain/entity/contact/src/cloudSyncModule.ts
@@ -1,0 +1,1 @@
+export * from "./cloudSync/module";
--- a/features/platform/contacts/src/contactDeviceIntentsPort.ts
+++ b/features/platform/contacts/src/contactDeviceIntentsPort.ts
@@ -1,0 +1,71 @@
+import {
+  ExternalAddressProofSchema,
+  ExternalContactNameProofSchema,
+  type Contact,
+  type ContactAddress,
+  type ContactAddressLabel,
+  type ContactAddressValue,
+  type ContactName,
+} from "@domain/entity-contact";
+import {
+  mockDeviceContactGroupCredentials,
+  mockExternalAddressDeviceContext,
+} from "@domain/entity-contact/schema.mock";
+
+export type RegisterExternalAddressIntentInput = Readonly<{
+  contact: Pick<Contact, "id" | "name" | "deviceCredentials">;
+  currencyId: ContactAddress["currencyId"];
+  label: ContactAddressLabel;
+  address: ContactAddressValue;
+}>;
+
+export type RegisterExternalAddressIntentResult = Readonly<{
+  deviceCredentials: NonNullable<Contact["deviceCredentials"]>;
+  addressDeviceContext: ContactAddress["device"];
+}>;
+
+export type RenameExternalContactIntentInput = Readonly<{
+  contact: Contact;
+  name: ContactName;
+}>;
+
+export type EditExternalAddressScopeIntentInput = Readonly<{
+  contact: Contact;
+  address: ContactAddress;
+  label: ContactAddressLabel;
+}>;
+
+export type ContactDeviceIntentsPort = Readonly<{
+  registerExternalAddress(
+    input: RegisterExternalAddressIntentInput,
+  ): Promise<RegisterExternalAddressIntentResult>;
+  renameExternalContact(
+    input: RenameExternalContactIntentInput,
+  ): Promise<NonNullable<Contact["deviceCredentials"]>>;
+  editExternalAddressScope(
+    input: EditExternalAddressScopeIntentInput,
+  ): Promise<ContactAddress["device"]>;
+}>;
+
+export function createMockContactDeviceIntentsPort(): ContactDeviceIntentsPort {
+  return {
+    registerExternalAddress: async input => ({
+      deviceCredentials: input.contact.deviceCredentials ?? mockDeviceContactGroupCredentials(),
+      addressDeviceContext: mockExternalAddressDeviceContext(),
+    }),
+    renameExternalContact: async input =>
+      mockDeviceContactGroupCredentials({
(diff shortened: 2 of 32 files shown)


domain/entity/contact/src/cloudSyncModule.ts
```diff
@@ -0,0 +1,281 @@
...
+type DistantRecord = Readonly<Record<string, unknown>>;
+
+function asRecord(value: unknown): DistantRecord | null {
+  return typeof value === "object" && value !== null && !Array.isArray(value)
+    ? (value as DistantRecord)
+    : null;
+}
+
+function compareIds(left: string, right: string): number {
+  return left < right ? -1 : left > right ? 1 : 0;
```
gre-ledger: shouldn't have to deal with comparing ids and there is no need to do any `sort()` in the module side I think. I think we can just let the addresses by ordered by the insertion order if you confirm the UI will not need to sort() them in the data itself (so it's kept stable) but only in the presenting side. (basically sort() is necessary when the data itself can change on different live instances like it was the case for the complex accounts module, in order to avoid never converging issues, but otherwise if it doesn't need that and it's best to make the diff logic as performant as possible)

domain/entity/contact/src/cloudSyncModule.ts
```diff
@@ -0,0 +1,281 @@
...
+import { z } from "zod";
+import { ContactAddressSchema, ContactIdSchema, ContactNameSchema, ContactSchema } from "./schema";
+import type { Contact, ContactAddress } from "./types";
+
+const LOCAL_ME_CONTACT_ID = ContactIdSchema.parse("contact-me");
+const DEFAULT_ME_CONTACT_NAME = ContactNameSchema.parse("Me");
+
+export const ContactsDistantSchema = z.object({
+  me: z.object({
+    name: ContactNameSchema,
```
gre-ledger: I see ContactNameSchema is a very strict schema, i would recommend to make these type a bit less strict and a bit more lossy to allow future evolution of your entity schema independently from this side.
essentially it's best if you could define the schema of cloud sync module separately from the entity ones (don't import the ./schema if possible, just refine the basic shape)

the shape that the cloudSyncModule shall be a stable desired schema that persist over the future iteration of the contact feature and is permissive enough to not be breaking over the iterations. (e.g. don't change the type of existing fields, add more optional fields but don't remove any, etc..)

breaking would be allowed if we really want to stop older Live version to be sync with new model, like if it's a major feature change etc.. as discussed, we will eventually be protected from this thanks to https://github.com/LedgerHQ/ledger-live/pull/20555 but until then, we will try to make our schemas future proof 🙏 

- entity schema are in the `develop` lifecycle and are great to be restricting on the current business logic consistently with the living code
- cloud sync schema must resists over the future and past iterations
  - if breaking, the module will have to support both the past and new schemas which can quickly become complex and it's also why we need to have its own schema inlined.

gre-ledger (commented):
as discussed together, giving you some early feedbacks on this first iteration, to be able to test this you will also need to plug the module inside live-wallet and you should be able to start playing with webtools and your app

features/platform/contacts/src/contactDeviceIntentsPort.ts
```diff
@@ -0,0 +1,78 @@
...
+    input: EditExternalAddressScopeIntentInput,
+  ): Promise<ContactAddress["device"]>;
+}>;
+
+function createMockDeviceCredentials(): NonNullable<Contact["deviceCredentials"]> {
+  return DeviceContactGroupCredentialsSchema.parse({
+    groupHandle: "mock-contact-group-handle",
+    hmacProof: "mock-external-contact-name-proof",
+  });
+}
```
mcayuelas-ledger: `mockDeviceCredentials` appears in both `schema.mock.ts` and `contactDeviceIntentsPort`.ts — consolidate into one source if you ca no?

mcayuelas-ledger (approved):

Sebastien-Dav1d (approved):

PR #20582, opened 2026-08-07
refactor(architecture): move currency id schemas to domain packages

ysitbon:
### 📝 Description

`CryptoCurrencyIdSchema`, `TokenCurrencyIdSchema` and `FiatCurrencyIdSchema` (and their inferred
types) move from `@shared/schema-primitives` to the packages that own them:

| Symbol | From | To |
| --- | --- | --- |
| `CryptoCurrencyIdSchema` / `CryptoCurrencyId` | `@shared/schema-primitives` | `@domain/entity-currency-crypto` |
| `TokenCurrencyIdSchema` / `TokenCurrencyId` | `@shared/schema-primitives` | `@domain/entity-currency-token` |
| `FiatCurrencyIdSchema` / `FiatCurrencyId` | `@shared/schema-primitives` | `@domain/entity-currency-fiat` |

Two reasons:

1. **A primitives package has no business knowing about crypto, tokens or fiat.** After this
   change `shared/schema-primitives` no longer contains the word "currency" — it is back to
   generic string primitives (`NonEmptyString`, `BigNumberStr`, `DateTimeIso`, `HttpUrl`, `SemVer`).
2. **The crypto and token barrels were proxies.** They re-exported these symbols from primitives,
   so the same symbol had two import paths and no obvious original provider.

```ts
// domain/entity/currency-crypto/src/index.ts — before
export * from "./schema";
export type { CryptoCurrencyId } from "@shared/schema-primitives";
export { CryptoCurrencyIdSchema } from "@shared/schema-primitives";

// after
export * from "./schema";
```

#### Blast radius

Small, and measured before starting. Only **5 files** imported these symbols from
`@shared/schema-primitives`, four of them inside the crypto and token packages themselves; the
fifth is `domain/entity/contact/src/schema.ts`.

The **70 files** that consume these symbols today already import them from
`@domain/entity-currency-*` through the proxy, so they are untouched — the symbol genuinely moved
to the package they were already naming.

Dependency changes: `currency-crypto` drops `@shared/schema-primitives`; `currency-token` swaps it
for `@domain/entity-currency-crypto` (no cycle, crypto does not depend on token); `contact` gains
both currency packages and keeps primitives for `NonEmptyStringSchema`.

The three `describe` blocks covering the id schemas moved out of `string.test.ts` into each
package's `schema.test.ts`, unchanged.

#### Checks

- `nx run-many -t typecheck` on the 6 affected projects — green
- `nx run-many -t test` on the same — 216 tests green
- `nx run enforce-boundaries:lint:boundaries` — `✓ module boundaries ok`

### 🔗 Context

- **JIRA / GitHub issue**: _none — groundwork for an upcoming nx plugin that forbids barrels from re-exporting another workspace package_
- **ADR** (if any): n/a

### DIFF
--- a/.changeset/currency-ids-move-to-domain.md
+++ b/.changeset/currency-ids-move-to-domain.md
@@ -1,0 +1,19 @@
+---
+"@domain/entity-currency-crypto": minor
+"@domain/entity-currency-token": minor
+"@domain/entity-currency-fiat": minor
+"@shared/schema-primitives": minor
+"@domain/entity-contact": minor
+---
+
+Move the currency id schemas to the packages that own them.
+
+`CryptoCurrencyIdSchema`, `TokenCurrencyIdSchema` and `FiatCurrencyIdSchema` (and their inferred
+types) now live in `@domain/entity-currency-crypto`, `@domain/entity-currency-token` and
+`@domain/entity-currency-fiat` respectively, instead of `@shared/schema-primitives`. A primitives
+package has no business knowing about crypto, tokens or fiat.
+
+The crypto and token packages used to re-export these symbols from primitives, which made them
+proxies: two import paths for the same thing, and no obvious original provider. Consumers already
+importing from `@domain/entity-currency-*` are unaffected, since the symbols genuinely moved there.
+Anything importing them from `@shared/schema-primitives` must now import the owning domain package.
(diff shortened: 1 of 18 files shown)


gre-ledger (approved):
as we managed to make it work by `.brand<*>` system that exists across lib thanks to Zod, then it makes a lot of sense that we no longer need it in a shared

mcayuelas-ledger (approved):

PR #20596, opened 2026-08-07
[LWDM] feat(auth): authenticate wallet-cli swap quote requests

thesan:
### 📝 Description

This PR integrates the authentication flow into the new wallet-cli's store. The goal is to add authentication to the swap quote command.

The main issue to achieve this is allowing LKRP credentials (potentially without any trustchain) to either created or re-used from the existing LKRP flows. 

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35626] 
- **ADR** (if any): 

[LIVE-35626]: https://ledgerhq.atlassian.net/browse/LIVE-35626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sebastien-Dav1d (dismissed):

apps/wallet-cli/src/state-manager/configureStore.ts
```diff
@@ -0,0 +1,43 @@
...
+              ledgerClientVersion: getEnv("LEDGER_CLIENT_VERSION"),
+            }),
+            ...authApiExtra({
+              isFeatureEnabled: () => true,
+              authProvider: new AuthSDK(
+                {
+                  clientId: getEnv("LEDGER_AUTH_CLIENT_ID"),
+                  keycloakBaseUrl: getEnv("LEDGER_AUTH_KEYCLOAK_BASE_URL_PROD"),
+                  keycloakRealm: getEnv("LEDGER_AUTH_KEYCLOAK_REALM"),
+                  disablePkce: true,
```
gre-ledger, line 32: any reason we need to disablePkce ?

apps/wallet-cli/src/test/helpers/cli-runner.ts
```diff
@@ -320,7 +324,10 @@ export async function runCli(args: string[], env: Record<string, string> = {}):
 
   let exitCode = 0;
   try {
-    exitCode = await runMain(args);
+    exitCode = await withMemberCredentialRepository(inMemoryMemberCredentialRepository, () => {
+      createWalletCliStore();
```
gre-ledger, line 327: any reason this is done after withMemberCredentialRepository?
the createWalletCliStore code feels like it's a setup (connecting the middleware together) and i wonder why it can't be done at top level ? 🤔 (like some sort of setupStore. i actually am surprised we so far didn't had to set it up)

apps/wallet-cli/src/state-manager/configureStore.ts
thesan: Yes the backend doesn't support it yet. This is planned though, so I kept the code for now

apps/wallet-cli/src/test/helpers/cli-runner.ts
thesan: TBH, I'm not completely clear on the [`node:async_hooks AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage), but the goal is to run the tests within the `new AsyncLocalStorage<MemberCredentialRepository>().run(repository, task)`, which is how it has to be done to access the store.

apps/wallet-cli/src/commands/ring/destroy.ts
```diff
@@ -90,12 +90,15 @@ async function loadDestroyCredentials(session: Session): Promise<DestroyCredenti
...
-    // that failed to decrypt (wrong password) — local-wiping either would orphan a live ring.
+    // it (the path `ring init` points here for). Any other error may be transient while the ring is
+    // still remotely authenticatable, so local-wiping would risk orphaning a live ring.
     if (e instanceof CorruptKeychainError) return { status: "local-wipe", message };
     if (e instanceof PasswordRequiredError || wrappingKey) return { status: "abort", message };
-    return { status: "local-wipe", message };
+    return {
+      status: "abort",
+      message,
+      hint: " Retry when access to the OS keychain is restored.",
```
Justkant: Not sure if we should put the hint here as it shouldn't really decided by this function IMO

I understand it's simpler for you to only check once but given we're changing the status from local-wipe to abort here, we might want to rethink all this

apps/wallet-cli/src/commands/ring/init.ts
```diff
@@ -45,15 +46,6 @@ export default defineCommand({
...
       }
-      // Refuse a stray keychain key with no ring metadata (e.g. after `session reset` on a corrupt
-      // file): overwriting it would orphan the previous remote member, whose only auth key it is.
-      if (hasStoredKey()) {
-        throw new Error(
-          "A member credential already exists in the OS keychain but this session has no Ledger Key " +
-            "Ring metadata. Run `wallet-cli ring destroy` (or remove the keychain entry) before " +
-            "re-initializing.",
-        );
-      }
```
Justkant, line 56: Why do you remove this check ?

apps/wallet-cli/src/commands/ring/init.ts
```diff
@@ -89,12 +96,15 @@ export default defineCommand({
       );
       deviceSpin?.success("Ledger Key Ring ready");
 
+      if (passwordSalt) {
+        session.setPasswordSalt(passwordSalt);
+        session.write();
+      }
```
Justkant: I'm not sure you want to save before saving the privateKey because you would have some saved salt while the private key could have failed to save

apps/wallet-cli/src/commands/session/reset.ts
```diff
@@ -24,23 +23,20 @@ export default defineCommand({
...
+        // not discard the salt required to unlock a stored credential.
         if (err && typeof err === "object" && "code" in err && typeof err.code === "string") {
           throw err;
         }
+        if (isStoredKeyPasswordProtected()) {
+          throw new Error(
+            "Cannot reset an invalid session while its member credential is password-protected. Restore the original session file to recover its password salt, or remove the credential from the OS keychain before retrying.",
+            { cause: err },
+          );
+        }
```
Justkant: Same here I don't understand the changes made here and the message is not actionable as before

Justkant: Password protected is the default and the fallback with no password is not documented and shouldn't be used so why do we care ?

apps/wallet-cli/src/key-ring/member-credential-repository.ts
```diff
@@ -0,0 +1,41 @@
...
+  deletePassword(service: string, account: string): void;
+}
+
+const nativeMemberCredentialRepository: MemberCredentialRepository = {
+  getPassword: (service, account) => new Entry(service, account).getPassword(),
+  setPassword: (service, account, value) => new Entry(service, account).setPassword(value),
+  deletePassword: (service, account) => new Entry(service, account).deletePassword(),
+};
+
+const memberCredentialRepositoryStorage = new AsyncLocalStorage<MemberCredentialRepository>();
```
Justkant, line 9: Why do we use AsyncLocalStorage ? We should only use keyring

apps/wallet-cli/src/test/helpers/cli-runner.ts
```diff
@@ -289,17 +292,18 @@ export async function runCli(args: string[], env: Record<string, string> = {}):
...
-  //    Doing this lazily ensures module-resolution failures appear as individual
-  //    test errors rather than a file-level "Unhandled error between tests".
-  const { runMain, setTestDmkTransport } = await getCliModules();
-
-  // 1. HTTP interceptor: install once, update port per call
+  // 0. Install the inactive HTTP interceptor before loading the CLI so clients
+  //    created by runMain capture it. A mock port activates it per invocation.
+  await installInterceptors();
   if (mergedEnv.WALLET_CLI_MOCK_PORT) {
-    await installInterceptors();
```
Justkant, line 299: Do we need this change, if we just need to have the getCliModules below the if it's fine but do we really need to also move it outside the check ?

apps/wallet-cli/src/cli.ts
```diff
@@ -76,7 +77,10 @@ if (import.meta.main) {
   const argv = normalizeNegatedFlags(process.argv.slice(2));
   try {
     startAnalytics();
-    process.exitCode = await withCommandLifecycleAnalytics(argv, () => runMain(argv));
+    process.exitCode = await withCommandLifecycleAnalytics(argv, () => {
+      createWalletCliStore();
```
Justkant: This part seems a bit misplaced if we only need it for some commands and not everything in the CLI

Maybe it should be kept in live-common-setup

apps/wallet-cli/src/test/helpers/cli-runner.ts
```diff
@@ -320,7 +324,10 @@ export async function runCli(args: string[], env: Record<string, string> = {}):
 
   let exitCode = 0;
   try {
-    exitCode = await runMain(args);
+    exitCode = await withMemberCredentialRepository(inMemoryMemberCredentialRepository, () => {
+      createWalletCliStore();
+      return runMain(args);
+    });
```
Justkant, line 329: So do I understand correctly that you only use the asynclocalstorage for in memory test only ?

We should instead keep the mock for keyring and only use the prod code not have something else we test with only and never test the prod code

apps/wallet-cli/src/commands/ring/init.ts
thesan: It's because having a "stray" keychain key is now valid. The trustchain backend will accept those for auth ATM. The goal is to potentially add these keypairs as Trustchain members when needed.

apps/wallet-cli/src/commands/session/reset.ts
```diff
@@ -23,24 +23,18 @@ export default defineCommand({
...
       } catch (err) {
-        // A real read failure (e.g. EACCES) carries an fs `code` — rethrow it rather than masking it
-        // by overwriting with a fresh session. Only invalid/unsalvageable YAML is safe to treat as
-        // empty and start clean; the orphan check below still warns.
-        if (err && typeof err === "object" && "code" in err && typeof err.code === "string") {
-          throw err;
+        if (!(err instanceof InvalidSessionYamlError)) throw err;
+        // TODO: Let users preserve and repair invalid YAML before resetting it.
+        if (hasStoredKey()) {
+          writeStderr(
```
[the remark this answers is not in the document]
thesan: A session with missing trustchain metadata (even with trustchain credentials) is now a valid state.

apps/wallet-cli/src/test/helpers/cli-runner.ts
thesan: Yeah, this wasn't ideal architecture-wise. I have now changed `withMemberCredentialRepository` to require a `MemberCredentialRepository`. This is now a proper dependency inversion, with `inMemoryMemberCredentialRepository` being used in tests and `nativeMemberCredentialRepository` in prod. No more fallback to the "native" keychain.

The rationale for this approach is that any authenticated call could hit the keychain (for now it's just swap-quote though). IMO, the native keychain should never be hit in tests, as it's a global state that can't be reset easily. Since it's pointless to test `@napi-rs/keyring` ourselves, I think this dependency inversion approach makes sense.

WDYT?

gre-ledger (dismissed):

apps/wallet-cli/src/test/helpers/cli-runner.ts
thesan: Yes, this was a leftover from when I was using a bun preload file instead of the (partial) dependency inversion. Now I've reverted this.

apps/wallet-cli/src/key-ring/member-credential-repository.ts
thesan: Details here: https://github.com/LedgerHQ/ledger-live/pull/20596#discussion_r3881565782

apps/wallet-cli/src/commands/session/reset.ts
thesan: > the message is not actionable as before

No it wasn't 🤦. I mostly restored the previous message.

The reason for the change is again that: credentials without trustchain data is now an acceptable state.

> Password protected is the default

This PR will create password-unprotected LKRP credentials when hitting an authentication-required request. IMO, this is acceptable because these credentials won't have any trustchain access. WDYT?

gre-ledger (dismissed):

apps/wallet-cli/src/commands/ring/destroy.ts
thesan: Yes, I reverted this part as it's not linked to the feature itself (my agent had overreached a bit).
The reason this was done in the first place is that it's now a lot more obvious (again, thanks to the dependency inversion) that the keyring access might throw. In this case, destroy could wipe credentials without deleting the remote trustchain due to the keyring being unavailable for various reasons. It might be the correct behaviour though. WDYT?

apps/wallet-cli/src/commands/ring/init.ts
thesan: The rationale is: "An encrypted private key without its salt is unrecoverable, while a persisted salt without a private key is harmless".

CremaFR (dismissed):

CremaFR (approved):

gre-ledger (approved):

PR #20598, opened 2026-08-08
fix(hw-app-*): fail closed on invalid BIP32 path segments

SashaMIT:
## Summary
- Path-parsing helpers across the hw-app packages (`splitPath` / `pathToBuffer` / `bip32PathToBytes` copies) previously skipped `NaN` segments (existing FIXME in eth) and used `parseInt`, which truncates values like `12abc'` to `12` and can drop the hardening bit.
- Callers then derived a shallower path than intended and could sign with the wrong key material.
- Reject any segment that is not digits with an optional hardening suffix, in every package: `hw-app-eth`, `hw-app-trx`, `hw-app-icon`, `hw-app-tezos`, `hw-app-vet`, `hw-app-near`, `hw-app-solana`, `hw-app-helium`, `hw-app-str`, `hw-app-kaspa`, `hw-app-cosmos`, `hw-app-xrp`, `hw-app-hedera`, `hw-app-algorand`, `hw-app-polkadot`, `hw-app-multiversx`, `hw-app-aptos`, `hw-app-canton`, `hw-app-btc` (plus `signMessage` via `pathStringToArray`).
- Consolidated from #20599 and #20601-#20607 into this single PR to keep review in one place.

## Test plan
- [x] Unit cases per package for `NOTAINDEX`, truncated `12abc'`, and empty segments
- [x] `hw-app-eth` tests/utils.unit.test.ts 44/44 and `hw-app-btc` tests/bip32.test.ts 31/31 pass locally
- [ ] Full package unit suites in CI

Tooling assist: Cursor agent helped prepare this PR; commits are authored and signed by me (no Cursor co-author trailer).

### DIFF
--- a/libs/ledgerjs/packages/hw-app-algorand/src/Algorand.ts
+++ b/libs/ledgerjs/packages/hw-app-algorand/src/Algorand.ts
@@ -15,8 +15,8 @@
  *  limitations under the License.
  ********************************************************************************/
 import type Transport from "@ledgerhq/hw-transport";
-import BIPPath from "bip32-path";
 import { UserRefusedOnDevice } from "@ledgerhq/hw-transport/errors";
+import { resolveBip32Path } from "./bip32Path";
 import { encodeAddress } from "./utils";
 const CHUNK_SIZE = 250;
 // const P1_FIRST = 0x00;
@@ -62,7 +62,7 @@ export default class Algorand {
     publicKey: string;
     address: string;
   }> {
-    const bipPath = BIPPath.fromString(path).toPathArray();
+    const bipPath = resolveBip32Path(path);
     const buf = Buffer.alloc(4);
     buf.writeUInt32BE(bipPath[2], 0);
     return this.transport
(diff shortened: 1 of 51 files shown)


SashaMIT:
Copilot review adopted in f51fe16d: every validator now also rejects segments above the BIP32 max index (0x7fffffff) so oversized values cannot alias hardened indices when encoded as uint32; delegate validators skip a leading m root to mirror bip32-path; solana/aptos/icon invalid-path expectations updated; added hw-app-trx splitPath unit coverage. Full local run: all 19 touched packages green (eth 361, btc 235, icon 3, trx 11, near 3, vet 5, solana 14, aptos 10, algorand 3, canton 48, hedera 2, multiversx 2, cosmos 3, xrp 5, kaspa 2, polkadot 3, helium 13, str 28, tezos 3).

SashaMIT:
Still relevant. Copilot review notes are already in f51fe16d. Happy to rebase if that helps review.

Justkant (commented):
Thank you for the fixes, maybe we should consider sharing the `resolveBip32Path` helper instead of copying the logic in all hw-app, wdyt ?

SashaMIT:
Thanks. Yes, a shared helper is the right end state. I kept the copies in this PR because the parsers are not the same: eth/trx/icon still use local splitPath (only '), and the bip32-path apps auto-harden or skip m. Folding them into one resolveBip32Path means a new shared package and a small behavior choice per app.

Happy to do that as a follow-up once this fail-closed gate is in, unless you would rather I fold the share into this PR.

SashaMIT:
Still relevant. A shared helper is the right end state once the parsers can share one. Commenting so the stale timer does not close this.

PR #20599, opened 2026-08-08
fix(hw-app): fail closed on invalid BIP32 segments in tezos/vet/near

SashaMIT:
## Summary
- Sibling of #20598 (`hw-app-eth` / `trx` / `icon`): `hw-app-tezos`, `hw-app-vet`, and `hw-app-near` still failed open on invalid BIP32 path segments.
- Tezos/Vet skipped `NaN` segments (or filtered them); Near mapped `NaN` via `Math.abs(NaN) | 0x80000000` to hardened `0`. Callers could derive a different key than the path string implied.
- Reject any segment that is not digits with an optional hardening suffix; add unit coverage for the reject path.

## Test plan
- [x] Unit cases for tezos `splitPath`, vet `splitPath`, near `bip32PathToBytes` covering `NOTAINDEX`, truncated `12abc'`, and empty segments
- [ ] Package unit tests in CI

Tip: `d041b8cc` (develop). SSH-signed under SashaMIT.

Tooling assist: Cursor (author owns the change).

### DIFF
--- a/libs/ledgerjs/packages/hw-app-near/src/utils.ts
+++ b/libs/ledgerjs/packages/hw-app-near/src/utils.ts
@@ -1,14 +1,16 @@
 export const bip32PathToBytes = (path: string): Buffer => {
   const parts = path.split("/");
   return Buffer.concat(
-    parts
-      .map(part =>
-        part.endsWith("'")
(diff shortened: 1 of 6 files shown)


SashaMIT:

PR #20601, opened 2026-08-09
fix(hw-app-solana,hw-app-helium): fail closed on invalid BIP32 segments

SashaMIT:
## Summary

Sibling of #20598 / #20599 (hand-rolled `splitPath` / `bip32PathToBytes`).

`@ledgerhq/hw-app-solana` and `@ledgerhq/hw-app-helium` still send derivation paths through `bip32-path@0.4.2` after appending `'`. That library **truncates** garbage segments and can **drop hardening**:

- input segment `12abc'` → parsed index `12` (unhardened)
- so a path that looks hardened can derive a different key than the string implies

This PR validates each segment with `/^\d+[hH']?$/` **before** `BIPPath.fromString`, matching the fail-closed pattern in the sibling PRs.

## Changes

- `hw-app-solana`: extract `resolveHardenedBip32Path` in `bip32Path.ts`; use it from `pathToBuffer`
- `hw-app-helium`: same validation in `serialization.ts` `pathToBuffer`
- unit tests for accept + reject cases (`12abc'`, `NOTAINDEX`, empty segment)

## Test plan

- [x] `pnpm test -- path.test.ts` in `@ledgerhq/hw-app-solana` (Node 24)
- [x] `pnpm test -- path.test.ts` in `@ledgerhq/hw-app-helium` (Node 24)
- [ ] CI package tests green

## Notes

- Confirmed on tip `d041b8cc`: `BIPPath.fromString("44'/12abc'/0'")` → `[0x8000002c, 0xc, 0x80000000]`
- No device interaction required; pure path encoding fail-closed

### DIFF
--- a/libs/ledgerjs/packages/hw-app-helium/src/serialization.ts
+++ b/libs/ledgerjs/packages/hw-app-helium/src/serialization.ts
@@ -37,13 +37,28 @@ const serializePath = (path: number[]): Buffer => {
   return buf;
 };
 
-export const pathToBuffer = (originalPath: string): Buffer => {
+/**
+ * Harden every path segment (Helium/Ed25519) then parse with bip32-path.
+ * Fail closed on truncated/garbage segments that bip32-path would silently shorten.
+ * Exported for unit tests.
+ */
+export function resolveHardenedBip32Path(originalPath: string): number[] {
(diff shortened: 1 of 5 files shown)


SashaMIT:

PR #20602, opened 2026-08-09
fix(hw-app-str,hw-app-kaspa): fail closed on invalid BIP32 segments

SashaMIT:
## Summary
- `bip32-path@0.4.2` truncates garbage path segments (`12abc'` → unhardened `12`) before device APDUs in `hw-app-str` and `hw-app-kaspa`.
- Same class as open #20601 (solana/helium) and #20598/#20599.
- Validate `/^\d+[hH']?$/` before `BIPPath.fromString`. Str still hardens every segment for Ed25519; Kaspa validates the path as provided.

## Test plan
- [x] Local tip-reverify: bare `BIPPath.fromString("44'/12abc'/0'")` → `[0x8000002c, 0xc, 0x80000000]`; helpers reject with `Invalid BIP32 path segment`
- [x] Added `tests/path.test.ts` in both packages (valid path + garbage/`NOTAINDEX`/empty segment)
- Tip parent: `d041b8cc`

Commit is SSH-signed.


SashaMIT:

PR #20603, opened 2026-08-09
fix(hw-app-cosmos,hw-app-xrp): fail closed on invalid BIP32 segments

SashaMIT:
## Summary
- `bip32-path@0.4.2` truncates garbage path segments (`12abc'` → unhardened `12`) in `hw-app-cosmos` and `hw-app-xrp` before device APDUs.
- Same class as open #20601 (solana/helium), #20602 (str/kaspa), #20598/#20599.
- Validate `/^\d+[hH']?$/` before `BIPPath.fromString` (paths kept as provided; no force-harden).

## Test plan
- [x] Local tip-reverify: helpers accept `44'/118'/0'/0/0` / `44'/144'/0'/0/0` and reject `12abc'` / `NOTAINDEX` / empty segment
- [x] Added `tests/path.test.ts` in both packages
- Tip parent: `d041b8cc`

Commit is SSH-signed.

### DIFF
--- a/libs/ledgerjs/packages/hw-app-cosmos/src/Cosmos.ts
+++ b/libs/ledgerjs/packages/hw-app-cosmos/src/Cosmos.ts
@@ -15,8 +15,8 @@
  *  limitations under the License.
  ********************************************************************************/
 import type Transport from "@ledgerhq/hw-transport";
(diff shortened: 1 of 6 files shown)


SashaMIT:

PR #20605, opened 2026-08-09
fix(hw-app-polkadot,hw-app-multiversx): fail closed on invalid BIP32 segments

SashaMIT:
## Summary
- `bip32-path@0.4.2` truncates garbage path segments (`12abc'` → unhardened `12`) in `hw-app-polkadot` and `hw-app-multiversx` before device APDUs.
- Same class as open #20601–#20604 and #20598/#20599.
- Validate `/^\d+[hH']?$/` before `BIPPath.fromString` (paths kept as provided).

## Test plan
- [x] Local tip-reverify: helpers accept valid Polkadot/MultiversX paths and reject `12abc'` / `NOTAINDEX` / empty segment
- [x] Added `tests/path.test.ts` in both packages
- Tip parent: `d041b8cc`

Commit is SSH-signed.


SashaMIT:

PR #20606, opened 2026-08-09
fix(hw-app-aptos,hw-app-canton): fail closed on invalid BIP32 path segments

SashaMIT:
## Summary
- `bip32-path@0.4.2` silently truncates garbage path segments (e.g. `12abc'` → unhardened `12`).
- Validate each BIP32 segment with `/^\d+[hH']?$/` before `BIPPath.fromString` / `bippath.fromString` in Aptos (`pathStringToArray`) and Canton (`serializeBipPath` via new `resolveBip32Path`).
- Sibling of #20598, #20599, #20601–#20605.

## Test plan
- [x] Added `tests/path.test.ts` in both packages (valid path + truncated/garbage reject)
- [x] Local tip-reverify of helpers against tip `d041b8cc`
- [ ] CI package tests for `hw-app-aptos` / `hw-app-canton`

### DIFF
--- a/libs/ledgerjs/packages/hw-app-aptos/src/bip32.ts
+++ b/libs/ledgerjs/packages/hw-app-aptos/src/bip32.ts
@@ -24,6 +24,15 @@ export function bip32asBuffer(path: string): Buffer {
   return pathElementsToBuffer(pathElements);
 }
 
+/**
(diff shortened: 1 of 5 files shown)


SashaMIT:

PR #20607, opened 2026-08-09
fix(hw-app-btc): fail closed on invalid BIP32 path segments

SashaMIT:
## Summary
- `bip32-path@0.4.2` silently truncates garbage path segments (e.g. `12abc'` → unhardened `12`).
- Validate each BIP32 segment with `/^\d+[hH']?$/` (optional leading `m`) before parse in `pathStringToArray`.
- Route `signMessage` through the same helper instead of bare `bippath.fromString`.
- Sibling of #20598, #20599, #20601–#20606.

## Test plan
- [x] Extended `tests/bip32.test.ts` (valid `m/...` paths + truncated/garbage reject)
- [x] Local tip-reverify of path helper against tip `d041b8cc`
- [ ] CI package tests for `hw-app-btc`


SashaMIT:

PR #20610, opened 2026-08-10
[LWDM] chore(llc): register A4 account during sync

francois-guerin-ledger:
### 📝 Description

Registers accounts with the A4 indexer during sync so the backend starts tracking them.
- `registerWithA4` is called on every `getAccountShape` invocation: it resolves the A4 network for the currency and skips early if the network is unknown or registration is disabled in config
- `ensureA4Registered` drives the state machine against the A4 API:
  - 200 → account confirmed (cache hit on next sync)
  - 404 → `createAccount` then `addAddresses`
  - 412 → `addAddresses` then compares the returned server version `sv` against the locally expected version
- An in-memory `Set` keyed by `accountId:version` (where version is a hash of the address list) prevents redundant network calls within a session

### 🔗 Context

- **JIRA / GitHub issue**: 
- **ADR** (if any):


libs/ledger-live-common/src/bridge/generic-coin-framework/a4/client/registration.ts
```diff
@@ -0,0 +1,83 @@
...
+  try {
+    await client.getAccount(accountId, version);
+    registrationCache.add(key);
+  } catch (rawErr) {
+    const err = toA4HttpError(rawErr);
+
+    if (err.status === undefined) {
+      log("a4", `registration transport error for ${accountId}: ${err.message}`);
+      return;
+    }
```
Moustafa-Koterba, line 69: **nit** Better use `else if` imo, no need for early returns, but matter of taste

libs/ledger-live-common/src/bridge/generic-coin-framework/getAccountShape.ts
```diff
@@ -343,6 +365,11 @@ export function genericGetAccountShape(network: string, kind: string): GetAccoun
       derivationMode,
     });
 
+    void registerWithA4(currency.id, address).catch(e => {
```
Moustafa-Koterba, line 368: **question** did not know this kind of syntax, but since we are in a async function, why dont just 
```ts
try {
  await registerWithA4(//...
} catch (e) {
// ...
```
It will wait before going to next code, that the case with this syntax ?

Moustafa-Koterba (approved):

libs/ledger-live-common/src/bridge/generic-coin-framework/a4/client/registration.ts
francois-guerin-ledger: Adding condition blocks make the code harder to read in my opinion. I prefer sticking to early returns.

libs/ledger-live-common/src/bridge/generic-coin-framework/getAccountShape.ts
francois-guerin-ledger: As we are not `await`-ing, we don't wait for registration to finish before going to next code.
The synchronization on A4 runs as a background task here. This operation takes time so accounts won't be registered in the current `getAccountShape` anyway. And since we don't need to read the result from the registration either, the "fire-and-forget" syntax seems more appropriate to not block the users.

francois-guerin-ledger:
E2E Android: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/a28e0df1-eb96-45c4-b44e-a312711a76e8/
E2E Desktop: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/8b65e2e3-b7f2-49e8-bdad-52fedf198aeb/

francois-guerin-ledger:
`[USDC (Stellar)] - Add sub-account without parent` is already failing on `origin/develop`. See the nightly report https://ledger-live.allure.green.ledgerlabs.net/allure/reports/31926c62-95d9-41a5-a3c7-0da653ece062/.

PR #20615, opened 2026-08-10
[LWDM] fix(zcash): shielded persistence [LIVE-35664]

semeano:
### 📝 Description

Restore the `privateInfo` serialization on coin-bitcoin's Zcash chain-adapter (`assignToAccountRaw`/`assignFromAccountRaw`), removed with the legacy chain-adapter. Accounts are decoded at app startup, before the host mirrors the `zcashShielded` flag, so the bitcoin family bridge is always the one that reads a persisted Zcash account back; with no adapter serialization the ufvk was lost on load and then erased on the next save. This is intentionally flag-independent and round-trips the same raw shape the standalone `@ledgerhq/coin-zcash` module writes when the flag is on.

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/LIVE-35664

### DIFF
--- a/libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
+++ b/libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
@@ -1,0 +1,109 @@
+import { BigNumber } from "bignumber.js";
+import type {
+  DecryptedOutput,
+  DecryptedOutputRaw,
+  ZcashPrivateInfo,
+  ZcashPrivateInfoRaw,
+  ZcashSyncState,
+} from "./types";
+
+// Persistence of the shielded `privateInfo` (its unified full viewing key,
+// balances and decrypted notes) for a Zcash account served by the bitcoin
+// family bridge. This is flag-independent on purpose: accounts are decoded at
+// app startup, before the host has mirrored the `zcashShielded` flag, so the
+// bitcoin bridge is always the one that reads a persisted Zcash account back.
+// Without this the ufvk is dropped on every load and then erased on the next
+// save. The standalone @ledgerhq/coin-zcash module writes the same raw shape
+// when the flag is on, so the two round-trip interchangeably.
+
+function mapDecryptedOutput(output: DecryptedOutput): DecryptedOutputRaw {
+  return {
+    memo: output.memo,
+    transfer_type: output.transfer_type,
+    amount: output.amount.toString(),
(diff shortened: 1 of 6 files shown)


libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
```diff
@@ -0,0 +1,109 @@
+import { BigNumber } from "bignumber.js";
```
semeano, line 1: Why isn't this in coin-zcash? The module that has to read the account back at startup is coin-bitcoin, not coin-zcash, because of when decoding happens versus when the flag is known.

For Zcash, currency.family is "bitcoin" — and this runs before App.tsx mirrors the zcashShielded flag into live-common. So the standalone "zcash" bridge is only ever selected for a fully-flagged, mid-session account; the module that deserializes a Zcash account on every cold start is always coin-bitcoin. If coin-bitcoin can't round-trip privateInfo, the ufvk is gone before coin-zcash ever gets a say.

So both modules now write the same raw shape. coin-zcash handles it while the flag is on (mid-session saves); coin-bitcoin handles it at startup and whenever the flag is off. They round-trip interchangeably. The apparent duplication is the price of a currency that two different family bridges can serve depending on a runtime flag.

Route Zcash's family to "zcash" unconditionally in fromAccountRaw — breaks the whole point of the flag (Zcash must fall back to coin-bitcoin when shielded is off) and would eagerly load coin-zcash (and its heavy shielded deps) for every user at boot, flag or not.

francois-guerin-ledger: So this situation is temporary if I understand correctly ? If so, we should add a TODO so we don't forget to perform some cleaning afterward.

francois-guerin-ledger (dismissed):

libs/coin-modules/coin-bitcoin/src/chain-adapters/zcash/serialization.ts
semeano: The transparent transaction path will continue to route via coin-bitcoin. If we want to maintain that, then this is not temporary. It will be temporary if we decide to create a new "zcash" family and completely detach from bitcoin. Nevertheless, I'll update the comment with a TODO.

francois-guerin-ledger (approved):

PR #20622, opened 2026-08-10
[LWDM] refactor(coin-hedera): switch to validator.id

mdomanski-ext-ledger:
### 📝 Description

`@ledgerhq/coin-module-framework` 5.0.0 made `Validator.id` mandatory, and Hedera's `getValidators()` was setting both `id` and a duplicate `nodeId` holding the same value. This renames the legacy `HederaValidator.nodeId: number` to `id: string` across `coin-hedera`, `live-common` and both apps, so validator identity is expressed one way everywhere. 

Genuine on-chain fields keep their numeric type and are untouched — `Transaction.properties.stakingNodeId`, `HederaDelegation.nodeId` and the mirror-node payloads — with conversion confined to the UI boundaries where a picked validator becomes a transaction field. Preload caches written by earlier versions store `nodeId`, so hydration migrates them to `id` rather than producing malformed validators on first launch after the update.

**Impact of changes**

- Desktop delegation flow: the Ledger node is preselected by default, the validator list renders and highlights the active row, and search still matches by node id, name and address (checksum included).
- Desktop redelegation, undelegation and claim-rewards modals: the current validator shows in the disabled select, and the "removed validator" placeholder still appears when the staked node is no longer returned by the mirror node.
- Mobile delegation and redelegation flows: validator picker list, the summary validator row (name + node index), and the amount step all resolve the selected validator correctly.
- Update path — the main risk area: install this build over an existing version on an account that already has a Hedera delegation. Cached validators must still appear immediately (no empty list, and no crash when typing in the validator search) before any network refresh, including offline.
- Node id `0` specifically: node `0` is a real validator and `HEDERA_STAKING_LEDGER_NODE_ID` can be set to it — verify it is selectable, gets the Ledger badge, and is picked as the default rather than being skipped.

### 🔗 Context

- https://ledgerhq.atlassian.net/browse/LIVE-35542


henri-ly (approved):

mateuszpalosz-ext (approved):
code looks ok

PR #20625, opened 2026-08-10
[LWDM] feat(coin-evm): make the calldata floor cost remotely configurable

YazhuEth:
### 📝 Description

[EIP-7976](https://eips.ethereum.org/EIPS/eip-7976) raises the calldata floor cost from **10/40** to **64/64** gas per zero/non-zero byte. Its Backwards Compatibility section explicitly requires wallets to update, or they under-estimate gas and transactions fail.

We compute that floor ourselves in `computeEIP7623GasLimit`, with hardcoded values, and use it as a **validation lower bound** (`GasLessThanEstimate`) — not as the gas we send. Since there is no fork date yet (Sepolia has not forked; mainnet is Q4 2026), adapting the value must not require a Ledger Live release: it has to come from remote config, per chain.

**EIP-7976 changes two parameters, not one**, so a single-constant swap would ship an under-estimate:

| | `TOTAL_COST_FLOOR_PER_TOKEN` | zero byte weight | effective gas/byte |
| --- | --- | --- | --- |
| EIP-7623 (today) | 10 | 1 token | 10 / 40 |
| EIP-7976 | **16** | **4 tokens** | 64 / 64 |

EIP-7976 redefines the tokenisation as `(zero + nonzero) * 4`, where EIP-7623 uses `zero + nonzero * 4`. Two traps worth spelling out: **64 is the gas-per-byte figure, not the config value** (setting 64 would over-charge by 4x), and setting only the per-token constant to 16 while leaving the zero-byte weight at 1 gives 16/64, which *under*-estimates zero-byte calldata. A single knob cannot express both `(10, weight 1)` and `(16, weight 4)`.

**Changes**

- `config.ts`: two optional `EvmConfig` fields, `calldataFloorGasPerToken` and `calldataFloorZeroByteTokens`, alongside the existing remote-config-fed tuning fields (`minGasPrice`, `feeHistoryBlockCount`, …).
- `computeGasLimit.ts`: both the per-token gas and the zero-byte weight become parameters, passed as an optional options object rather than two positional numbers — swapping `16` and `4` positionally would pass silently and produce exactly the under-estimate this ticket exists to prevent. Defaults are the current values, so an omitted config is byte-identical to today.
- `validateIntent.ts`: forwards both values at the single production call site, reusing the `getCoinConfig(currency.id).info` destructuring already present in `validateGas`.

Invalid remote values (float, negative, zero, `NaN`) fall back to the defaults instead of throwing inside `BigInt()` mid-validation — falling back means pre-fork behaviour, which is the safe direction.

**One caveat documented in the code**: the doc comment justified omitting the spec's `max(intrinsic, floor)` by arguing intrinsic can never exceed the floor. That holds for both `(10, 1)` and `(16, 4)`, but it is a property of those parameter sets, not of the formula — a low enough `gasPerToken` pushed remotely would break it. It is now written up as a constraint on the values we may configure. Implementing the real `max()` would need the intrinsic formula, which EIP-2780 also changes post-fork; out of scope here.

**Tests**: the 4 existing vectors (`21000n`, `22400n`, `22490n`, `22460n`) pass unchanged with no config. New cases pin the post-fork vector (`21000 + 64 * 68 = 25352n`), 64/byte regardless of byte value, each parameter applied independently (the half-migration trap), and the fallback on every invalid value. `validateIntent` asserts the forwarded values both with and without config. Full module suite: 1114/1114.

Out of scope, untouched: the unrelated `10n` fee-ratio heuristic in `validateIntent.ts`, and the gas limit we actually broadcast.

### 🚦 Activation — this PR ships dormant

Merging changes nothing: the defaults reproduce today's behaviour exactly. **The fork is not activated by this PR.** When Glamsterdam activates on a chain, someone has to push, per chain, to remote config:

```
calldataFloorGasPerToken: 16
calldataFloorZeroByteTokens: 4
```

Verified against the spec text, not just the ticket: EIP-7976 sets `TOTAL_COST_FLOOR_PER_TOKEN = 16` and `floor_tokens_in_calldata = (zero_bytes + nonzero_bytes) * 4`, giving 64 gas per byte. **The EIP is still in Review, so these two numbers may change** — which is the whole reason they are configurable rather than hardcoded.

If this is forgotten at fork time, the validation floor stays at 10/40 and we stop rejecting a custom gas limit that is too low, so such a transaction fails on chain. The normal send flow is unaffected, since it follows `eth_estimateGas`.

**Who owns pushing the remote config needs deciding before the fork — it is not covered by this PR.**

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35603](https://ledgerhq.atlassian.net/browse/LIVE-35603) — epic [LIVE-29491](https://ledgerhq.atlassian.net/browse/LIVE-29491) `[coin-evm] Glamsterdam gas repricing`
- Follows [#20620](https://github.com/LedgerHQ/ledger-live/pull/20620) (LIVE-30166) from the same epic; independent branch, no dependency between them.
- **ADR** (if any): n/a

[LIVE-35603]: https://ledgerhq.atlassian.net/browse/LIVE-35603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/coin-modules/coin-evm/src/logic/validateIntent.ts
+++ b/libs/coin-modules/coin-evm/src/logic/validateIntent.ts
@@ -44,7 +44,11 @@ import {
   isEip55Address,
   isLegacyFeeEstimation,
 } from "./common";
-import { computeEIP7623GasLimit } from "./computeGasLimit";
+import {
+  computeEIP7623GasLimit,
+  DEFAULT_CALLDATA_FLOOR_GAS_PER_TOKEN,
+  DEFAULT_CALLDATA_FLOOR_ZERO_BYTE_TOKENS,
+} from "./computeGasLimit";
 import estimateFees from "./estimateFees";
 
 function assetsAreEqual(asset1: AssetInfo, asset2: AssetInfo): boolean {
@@ -169,8 +173,8 @@ async function validateGas(
     BigInt(DEFAULT_GAS_LIMIT.toFixed(0)),
     intent.data.value,
     {
-      gasPerToken: calldataFloorGasPerToken,
-      zeroByteTokens: calldataFloorZeroByteTokens,
+      gasPerToken: calldataFloorGasPerToken ?? DEFAULT_CALLDATA_FLOOR_GAS_PER_TOKEN,
+      zeroByteTokens: calldataFloorZeroByteTokens ?? DEFAULT_CALLDATA_FLOOR_ZERO_BYTE_TOKENS,
     },
   );
 
--- a/libs/coin-modules/coin-evm/src/logic/validateIntent.ts
+++ b/libs/coin-modules/coin-evm/src/logic/validateIntent.ts
@@ -147,7 +147,9 @@ async function validateGas(
   const errors: Record<string, Error> = {};
   const warnings: Record<string, Error> = {};
 
-  const { minGasPrice } = getCoinConfig(currency.id).info;
+  const { minGasPrice, calldataFloorGasPerToken, calldataFloorZeroByteTokens } = getCoinConfig(
+    currency.id,
+  ).info;
   const minGasPriceFloor = typeof minGasPrice === "string" ? BigInt(minGasPrice) : null;
 
   const nativeBalance = findBalance({ type: "native" }, balances);
@@ -166,6 +168,10 @@ async function validateGas(
   const eip7623GasLimit = computeEIP7623GasLimit(
     BigInt(DEFAULT_GAS_LIMIT.toFixed(0)),
     intent.data.value,
+    {
+      gasPerToken: calldataFloorGasPerToken,
+      zeroByteTokens: calldataFloorZeroByteTokens,
+    },
   );
 
   // Gas Limit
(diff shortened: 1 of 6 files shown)


YazhuEth:
**E2E**
- D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/c54e2544-9e09-4f86-b37b-4c336b3d56d2/
- M : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/25928ec3-f8d0-4385-b275-8a35f1ada7ba/

libs/coin-modules/coin-evm/src/logic/validateIntent.ts
```diff
@@ -166,6 +168,10 @@ async function validateGas(
   const eip7623GasLimit = computeEIP7623GasLimit(
     BigInt(DEFAULT_GAS_LIMIT.toFixed(0)),
     intent.data.value,
+    {
+      gasPerToken: calldataFloorGasPerToken,
+      zeroByteTokens: calldataFloorZeroByteTokens,
+    },
```
francois-guerin-ledger, line 178: Can we move constants from `computeGasLimit.ts` here and pass them ? This simplifies the type of `CalldataFloorParams` so everything is mandatory

francois-guerin-ledger (approved):

PR #20626, opened 2026-08-10
ci(codecheck): run codechecks for every lib scope (LIVE-35675)

lewisd5:
## Problem

`test-libs-reusable.yml` defines a job displayed as "Libraries Codecheck", but its four checks were repo-wide — `nx run-many … --exclude=<6 projects>`, not `--projects=tag:scope:libs`. Its caller gates it on `contains(needs.determine-affected.outputs.paths, 'libs')`.

So a PR touching only `features/`, `domain/`, `shared/` or `devtools/` was never formatted, linted, typechecked or unimported-checked anywhere. The sibling workflows didn't cover it: features/shared ran only `coverage`, domain added only `check-export-rules`. 37 projects in those four scopes have no `libs/` dependent, so the gate never fired for them.

## Change

- **New `tools/actions/composites/nx-codechecks`** — one definition of the four targets, selector-parameterised, all `--nxBail=false`. Optional `emit-summaries` keeps the libs `report` job's per-target granularity.
- **New `.github/workflows/test-scope-reusable.yml`** — one job owning a scope's full pipeline. The codecheck step carries `if: ${{ !cancelled() }}`, so a failing test no longer hides typecheck errors and a failing typecheck no longer costs Sonar its coverage artifact.
- **The four sibling workflows** become ~42-line delegators, keeping their `workflow_dispatch` entry points and `coverage_generated` outputs.
- **`test-libs-reusable.yml`** uses the composite with `projects: tag:scope:libs`. The hand-maintained `--exclude=` list is gone; excluding apps and e2e is now structural.

Net: 291 lines deleted, 30 added.

## Scope calls

`tools/actions/*` (9), `support/*` (2) and `tests/dummy-wallet-app` lose `format:check`/`lint`/`typecheck` deliberately. `apps/wallet-cli` also drops out of the libs job but keeps full coverage in `build-wallet-cli-reusable.yml` - it was duplicated before. Follow up pending

`build-and-test-push.yml` is unchanged, so develop pushes still run libs + features only. Follow-up.

Test Commit: https://github.com/LedgerHQ/ledger-live/actions/runs/31381570217
Ticket: https://ledgerhq.atlassian.net/browse/LIVE-35675

### DIFF
--- a/.github/workflows/test-libs-reusable.yml
+++ b/.github/workflows/test-libs-reusable.yml
@@ -188,28 +188,12 @@ jobs:
         with:
           registry: ${{ vars.ARTIFACTORY_URL }}
       - name: Install dependencies
-        run: pnpm i
-      - name: Check formatting of affected libraries
-        id: format-check-libs
-        if: github.base_ref == 'develop' || github.ref == 'refs/heads/develop' || github.event_name == 'workflow_dispatch'
-        run: pnpm exec nx run-many -t format:check --exclude=ledger-live-desktop,live-mobile,@ledgerhq/live-cli,@ledgerhq/web-tools,ledger-live-desktop-e2e-tests,ledger-live-mobile-e2e-tests --parallel=3 --nxBail=false
-      - name: Lint affected libraries
-        id: lint-libs
-        run: pnpm exec nx run-many -t lint --exclude=ledger-live-desktop,live-mobile,@ledgerhq/live-cli,@ledgerhq/web-tools,ledger-live-desktop-e2e-tests,ledger-live-mobile-e2e-tests --parallel=3 --nxBail=false -- --quiet
-      - name: Typecheck affected libraries
-        id: typecheck-libs
-        run: pnpm exec nx run-many -t typecheck --exclude=ledger-live-desktop,live-mobile,@ledgerhq/live-cli,@ledgerhq/web-tools,ledger-live-desktop-e2e-tests,ledger-live-mobile-e2e-tests --parallel=3 --nxBail=false
-      - name: Test unimported files
-        id: unimported
-        run: pnpm exec nx run-many -t unimported --exclude=ledger-live-desktop,live-mobile,@ledgerhq/live-cli,@ledgerhq/web-tools,ledger-live-desktop-e2e-tests,ledger-live-mobile-e2e-tests --parallel=3 --nxBail=false
-        shell: bash
-      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
-        if: ${{ !cancelled() }}
+        run: pnpm i --filter="{./libs/**}..."
+      - name: Codechecks
+        uses: LedgerHQ/ledger-live/tools/actions/composites/nx-codechecks@develop
         with:
-          script: |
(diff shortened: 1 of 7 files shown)


.github/workflows/test-libs-reusable.yml
```diff
@@ -188,28 +188,12 @@ jobs:
...
-        id: typecheck-libs
-        run: pnpm exec nx run-many -t typecheck --exclude=ledger-live-desktop,live-mobile,@ledgerhq/live-cli,@ledgerhq/web-tools,ledger-live-desktop-e2e-tests,ledger-live-mobile-e2e-tests --parallel=3 --nxBail=false
-      - name: Test unimported files
-        id: unimported
-        run: pnpm exec nx run-many -t unimported --exclude=ledger-live-desktop,live-mobile,@ledgerhq/live-cli,@ledgerhq/web-tools,ledger-live-desktop-e2e-tests,ledger-live-mobile-e2e-tests --parallel=3 --nxBail=false
-        shell: bash
-      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
-        if: ${{ !cancelled() }}
+        run: pnpm i --filter="{./libs/**}..."
+      - name: Codechecks
```
dgreen-ledger, line 192: We lost the if !cancelled check here. Was this by design ?

.github/workflows/test-libs-reusable.yml
```diff
@@ -188,28 +188,12 @@ jobs:
...
+        uses: LedgerHQ/ledger-live/tools/actions/composites/nx-codechecks@develop
         with:
-          script: |
-            const fs = require("fs");
-            fs.writeFileSync("summary-typecheck.txt", "${{ steps.typecheck-libs.outcome }}", "utf-8");
-            fs.writeFileSync("summary-lint.txt", "${{ steps.lint-libs.outcome }}", "utf-8");
+          projects: tag:scope:libs
```
dgreen-ledger, line 195: I like the idea of simplifying the exclusion issue. 
I just wanted to check to my understanding this would remove checking for  tools/actions/ and support/  if so do we care ?

.github/workflows/test-libs-reusable.yml
lewisd5: This is just the diff doing some trickery. The cancelled is moved to line 54 of the composite

.github/workflows/test-libs-reusable.yml
lewisd5: For now - yes. These present more issue for develop breaking as the check within build-and-test-pr.yml only checks if the path 'libs' has been changed. So this explicitly aligns the scope to the directory that is being "affected checked". Support is new and has not yet been marked as accepted in the ADR so it's inclusion was accidental. 

tools/actions - Was not intended to be included here. 

After discussion with platform i will be creating a new swimlane for "support" under separate cover.
I will also be creating a new swimlane for CI checks including the above + zizmor + actionslint checking

IMO this PR essentially brings things in line with is-affected expectation so i'm treating anything else as an unintended side-affect to be dealt with under separate cover.

dgreen-ledger (approved):

PR #20627, opened 2026-08-10
[LWDM] feat(coin-evm): warn when the recipient account does not exist yet

YazhuEth:
### 📝 Description

[EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) raises the cost of sending value to an account that does not exist yet (25,000 gas today, 183,600 post-fork). Nothing in the send flow tells the user why the fee jumped — a funds-loss risk on a fresh recipient address.

This adds an **explanatory message only**. `eth_estimateGas` already returns the correct value, so the gas we send is untouched: computing the charge ourselves would drift from the node.

- New `RecipientIsNewAccount` error, emitted by `validateIntent` as a **warning** (never an error, never blocking) when the recipient has balance `0` **and** nonce `0`.
- Rendered on the amount screen in neutral `info` styling, like `FeeTooHigh`.
- **Also fixes a cross-app inconsistency**: mobile picked the message colour with a ternary whose `else` branch returned `muted`, so there was no `info` style at all — anything that was not an error or a warning came out grey, while desktop declares an explicit `info` variant. Mobile now uses the same explicit mapping. Side effect worth flagging: **`FeeTooHigh` also turns from grey to base colour on mobile, for every coin.**
- Detection uses `nodeApi.getCoinBalance` + `getTransactionCount` — 2 RPC calls — not `getBalance`, which also fans out to token balances and the explorer.
- **Native sends only**: a token transfer's `to` is the token contract, which always exists. A new holder's first storage write is EIP-8038, a separate ticket.
- Emitted under the existing `warnings.amount` key, so the shared cross-coin selector needed one edit (the `info` classification) instead of two.
- An unreachable node yields no message rather than a failed validation.

### 📸 Screenshots

| Desktop | Mobile |
| --- | --- |
| Screenshot 2026-08-11 at 09 26 59 |Simulator Screenshot - iPhone 17 Pro - 2026-08-11 at 10 41 16|

### ✅ Tests

7 cases in `validateIntent.test.ts`: the warning, an emptied account (nonce > 0), a funded recipient, token transfers and zero amounts issuing **no** RPC call, node failure staying silent, and fees/amount identical with and without the message. One case in the shared `messages.test.ts` asserting `info`. coin-evm suite: 1108/1108.

### 🗣 Copy

> This address has never received funds, activating it costs a higher network fee

Accurate for an EOA: `nonce = 0` means the address has never spent, `balance = 0` that it holds nothing, so it has never received. It also says what the extra fee pays for (activating the account) rather than only that fees are higher.

⚠️ **Written by an engineer, not reviewed wording — needs a UX pass before leaving draft.** Other locales follow the usual translation process.

⚠️ **It wraps to two lines.** The message slot reserves a single line (`min-h-20`, `AmountInputSection.tsx`), so a two-line message pushes the dialog footer down and clips the Review button on a short window. **Pre-existing bug, not introduced here** — `NotEnoughGas` is longer and hits it harder — but this feature makes multi-line messages routine on a success path, and translations will be longer still. Worth a separate ticket.

### 🔗 Context

- **JIRA**: [LIVE-30167](https://ledgerhq.atlassian.net/browse/LIVE-30167) — epic [LIVE-29491](https://ledgerhq.atlassian.net/browse/LIVE-29491)
- Completes the epic with [#20620](https://github.com/LedgerHQ/ledger-live/pull/20620) and [#20625](https://github.com/LedgerHQ/ledger-live/pull/20625). Independent branches.
- The ticket's open question (amount vs recipient screen) was resolved in favour of the **amount screen**.

[LIVE-30167]: https://ledgerhq.atlassian.net/browse/LIVE-30167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/evm-new-account-recipient-warning.md
+++ b/.changeset/evm-new-account-recipient-warning.md
@@ -1,0 +1,8 @@
+---
+"@ledgerhq/coin-evm": patch
+"@ledgerhq/live-common": patch
+"ledger-live-desktop": patch
+"live-mobile": patch
+---
+
+Explain the higher network fees when sending to an address that does not exist yet. EIP-8037 charges account creation substantially more gas, and nothing in the send flow told the user why the fee jumped. The gas we send is unchanged: `eth_estimateGas` remains the only source.
--- a/apps/ledger-live-desktop/static/i18n/en/app.json
+++ b/apps/ledger-live-desktop/static/i18n/en/app.json
@@ -7435,6 +7435,9 @@
     "FeeTooHigh": {
       "title": "Network fees are above 10% of the amount"
     },
+    "RecipientIsNewAccount": {
+      "title": "This address has never received funds, activating it costs a higher network fee"
+    },
     "DustLimit": {
       "title": "Amount is lower than dust limit"
     },
--- a/apps/ledger-live-mobile/src/locales/en/common.json
+++ b/apps/ledger-live-mobile/src/locales/en/common.json
@@ -491,6 +491,9 @@
     "FeeTooHigh": {
       "title": "Network fees are above 10% of the amount"
     },
+    "RecipientIsNewAccount": {
+      "title": "This address has never received funds, activating it costs a higher network fee"
+    },
     "FeeTooLow": {
       "title": "Fees are too low to be accepted by the network. Please increase them"
     },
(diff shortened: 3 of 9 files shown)


YazhuEth:
**E2E**
-D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/9bc49c46-6cc4-42e9-99a1-893fb927eb8c/
-M: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/d8b4d684-a252-4903-bf6c-e28e9481e2c6/

dilaouid (approved):

PR #20633, opened 2026-08-10
fix(mobile): stop auto-opening product tour

tonykhaov:
### 📝 Description

**Problem**: With `lwmProductTour` enabled, the mobile product tour auto-opened on every Portfolio / app launch after an early dismiss, because nothing durable was set and the once-per-session latch reset on cold start. Desktop never auto-opens.

**Solution**: Remove auto-open. The tour opens only from the post-onboarding hub, `ledgerlive://product-tour`, or Debug. Early dismiss leaves `productTourCompleted` false so the hub stays reopenable until last slide / Done. Incomplete tour no longer blocks Large Screen Upsell as a competing app-start modal.

**Tests**: Integration coverage for no auto-open (catches regression), dismiss→reopen, and deeplink open; redundant unit suites removed.

<table>
<tr>
 <td><video src="https://github.com/user-attachments/assets/770905c9-4020-4b00-9841-519adbeaca24"/>
 <td><video src="https://github.com/user-attachments/assets/e858e653-ccf6-4eb0-9cff-d5daaae37ed0"/>
</table>

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35688](https://ledgerhq.atlassian.net/browse/LIVE-35688)

[LIVE-35688]: https://ledgerhq.atlassian.net/browse/LIVE-35688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarneijim:
This guard was added to avoid that app-start modal chain. Removing it brings it back: with no tour auto-open, the upsell can auto-open at launch and a hub/deeplink tour request queues behind it — or the other way around (tour first, then upsell when the tour closes).

I'd keep a guard here, but not based on productTourCompleted.

deepyjr (approved):

sarneijim (approved):

PR #20632, opened 2026-08-10
fix(e2e): fix Borrow desktop multi-step approval polling

alexstapenka-ledger:
### 📝 Description

**Problem**: On broadcast desktop E2E nightlies, the Borrow open-loan flow could skip the Give approval step when `borrow-authorize-depositing-button` appeared enabled, even though token approval was still required. The test then timed out waiting for the deposit authorize button.

**Solution**: Update `BorrowPage.clickGiveApprovalWhenReady` to keep polling and click `give-approval-button` whenever it is visible (including while disabled/loading). Only skip Give approval when Step 1 is marked done or the give-approval control is absent and the next authorize step is ready. Applied the same logic to repay Give approval helpers.

**Regression check**: `[ETH] - Borrow simulates a loan and completes the open-loan execution` (B2CQA-6065) with `enable_broadcast=true`.

### 🔗 Context

- **JIRA issue**: [LIVE-35536](https://ledgerhq.atlassian.net/browse/LIVE-3)

[LIVE-35536]: https://ledgerhq.atlassian.net/browse/LIVE-35536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #20640, opened 2026-08-10
fix(e2e): stabilise and simplify Borrow desktop tests (LIVE-35536)

VicAlbr:
### 📝 Description

Fixes the Borrow desktop E2E suite, which was failing on the broadcast nightly, and simplifies it substantially while doing so.

**Absorbs #20632.** That PR fixes the LIVE-35536 approval-precedence bug: the flow skipped **Give approval** whenever `borrow-authorize-depositing-button` looked enabled, even though the token allowance was still required, and then timed out. The fix is included here — Give approval now wins while the control is on screen, and only an *absent* control means the allowance is already granted. Agreed with @alexstapenka-ledger that #20632 can be closed in favour of this PR. Their version also called `nextAuthorize.isEnabled()` on a possibly-unmounted element, which is a separate hang fixed below.

**Four further defects found while investigating:**

1. **Cascade skips.** `test.describe.configure({ mode: "serial" })` sat at *file* scope, so all four tests formed one serial group. Any failure abandoned the rest and replayed the whole file on every retry — this is why the nightly reported two tests as "did not run" *and* marked them retried. Now `mode: "default"`: still one at a time in a single worker (no nonce races), but each test fails and retries independently.

2. **60s hang on a probe.** `locator.isEnabled()` waits for the element to attach, unlike `isVisible()`. The code probed the *next* step's button before checking Give approval, so it parked for the full webview timeout on a button that is not rendered until its step is active, and threw one line above the click that would have unblocked it.

3. **A latching `or()` locator.** `expect(success.or(error).first()).toBeVisible()` resolves `.first()` to the first match in DOM order and stays latched: with the error container mounted-but-hidden ahead of the success marker, it never re-resolves and burns the full 240s. Replaced with per-locator visibility polling at all five sites.

4. **Misleading failure reporting.** The execution wait only checked for the error state *after* the 240s deadline, so every failure cost four minutes and then blamed funding. It now bails as soon as either side appears, and reports the account's real on-chain nonce.

**Root cause of the withdraw failure**, for the record: the app crafts with `getTransactionCount(pending)` and signs seconds later. In the failing run nonce 496 was mined by an unrelated transaction 2.6s *before* the app signed its own nonce-496 transaction, which was dropped and is still 404 on the explorer. A settle guard now blocks a flow from starting while the account has transactions in flight, and failures name the colliding nonce instead of guessing.

**Shared-account contention.** `speculos_device = "nanoSP and stax"` builds a two-leg matrix with no `max-parallel`, and both legs ran the on-chain borrow tests against the same account — a guaranteed collision. The on-chain flows are now gated to the primary leg via `E2E_PRIMARY_DEVICE`. This does not stop a concurrent run elsewhere on the same account; that needs a cross-run lock and is deliberately out of scope.

**Refactor.** The page object goes 773 → 584 lines: 7 unused members deleted, 40 one-line locator accessors collapsed to the class-field convention used by every sibling page object, and all three hand-rolled `while (Date.now() < deadline)` loops replaced. Two copy-paste helper pairs merged in `borrowSetup.ts`.

The workflow also carries unrelated `actionlint`/shellcheck fixes: adding one `env:` line makes the pre-commit hook lint the whole file, which surfaced ~20 pre-existing findings that blocked the commit.

**Verification.** Typecheck, lint, oxfmt and `actionlint` clean; 321 `@NanoSP` tests still discovered. The serial-vs-default behaviour and the locator bug were each proven with isolated Playwright specs. A local broadcast run confirmed the cold-start and open-loan tests pass with the settle guard firing correctly, but the run was interrupted before repay and withdraw completed — **those two still need a green broadcast run**, which is what the E2E CI on this PR is for.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35536](https://ledgerhq.atlassian.net/browse/LIVE-35536)
- **Supersedes**: #20632

[LIVE-35536]: https://ledgerhq.atlassian.net/browse/LIVE-35536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/borrow-e2e-stability.md
+++ b/.changeset/borrow-e2e-stability.md
@@ -1,0 +1,11 @@
+---
+"ledger-live-desktop-e2e-tests": patch
+"@ledgerhq/live-e2e-shared": patch
+---
+
+Make the Borrow desktop E2E suite independent and diagnosable: run the specs one at a time instead
+of as a serial group so a failure no longer skips the remaining tests, click Give approval whenever
+the control is on screen, wait on each marker's own visibility rather than a latching `or().first()`
+locator, report the account's on-chain nonce when a flow fails instead of a generic funding hint,
+and keep the on-chain flows on the primary device leg so parallel legs cannot collide on the shared
+borrow account.
--- a/.github/workflows/test-ui-e2e-only-desktop.yml
+++ b/.github/workflows/test-ui-e2e-only-desktop.yml
@@ -174,21 +174,27 @@ jobs:
           if [ "$EVENT_NAME" = "schedule" ]; then
             DOW=$(date -u +%u)
             DEVICE="${WEEKDAY_DEVICE[$DOW]:-nanoSP}"
-            echo "device_1=$DEVICE" >> $GITHUB_OUTPUT
-            echo "device_2=skipped" >> $GITHUB_OUTPUT
-            echo "is_dual=false"   >> $GITHUB_OUTPUT
-            printf 'e2e_matrix={"include":[{"device":"%s","is_primary":true,"artifact_suffix":""}]}\n' "$DEVICE" >> $GITHUB_OUTPUT
+            {
+              echo "device_1=$DEVICE"
+              echo "device_2=skipped"
+              echo "is_dual=false"
+              printf 'e2e_matrix={"include":[{"device":"%s","is_primary":true,"artifact_suffix":""}]}\n' "$DEVICE"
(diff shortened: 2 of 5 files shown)


VicAlbr:
ci: https://github.com/LedgerHQ/ledger-live/actions/runs/31397736598

jeportie (approved):

PR #20642, opened 2026-08-10
[LWDM] feat(pay-card): persist the hero balance filter selection (LIVE-34894)

mcayuelas-ledger:
### 📝 Description

The pay card hero lets the user narrow the displayed balance to a single stablecoin, but the selection lived only in memory: killing the app reset it back to "all". This is the shared domain part of that story — it makes the selection survive a restart.

`@domain/entity-pay-card` gains a `balanceFilter` field on `PayCardState`, holding either `"all"` or the **currencyId** of a single stablecoin. The currencyId is persisted rather than the ticker because ids are stable. Alongside it: a `setPayCardBalanceFilter` reducer and a `selectPayCardBalanceFilter` selector.

```ts
import {
  setPayCardBalanceFilter,
  selectPayCardBalanceFilter,
  PAY_CARD_BALANCE_FILTER_ALL,
} from "@domain/entity-pay-card";

dispatch(setPayCardBalanceFilter("ethereum/erc20/usd__coin"));
dispatch(setPayCardBalanceFilter(PAY_CARD_BALANCE_FILTER_ALL)); // back to "all"

const filter = useSelector(selectPayCardBalanceFilter);
```

Persistence needs no new wiring: the app-side plumbing added by LIVE-34861 / LIVE-34862 saves whatever `payCardPersistedSelector` returns, so extending that projection with `balanceFilter` and merging it back in `restorePayCardPersistedState` is enough for desktop to save and rehydrate it. `restorePayCardPersistedState` ignores a non-string or empty value so malformed persisted data falls back to `"all"`, and it merges without touching the transient `isOpen` / `params`.

One app-side fix was required, in the second commit: the mobile `DBSave` effect watched only `state.payCard.hasSeenFeatureTour`, so a `balanceFilter`-only change never reached `savePayCardState`. It now watches the whole slice and lets the existing `payCardPersistedNotEquals` comparator decide what is worth writing — transient `isOpen` / `params` changes still write nothing.

The fallback for a persisted id the user no longer holds (treat as `"all"`) belongs to the Story 2 balance view model, not here.

**Tests** — `slice.test.ts` covers the default, `setPayCardBalanceFilter` in both directions, `payCardPersistedSelector` including the field, and `restorePayCardPersistedState` merging it while leaving `isOpen` / `params` alone plus rejecting malformed values. The desktop and mobile persistence-lens tests were updated to assert the new projected shape.

**QA focus** — nothing is user-visible until the Story 2 filter UI dispatches the action, so the check is regression-only: the pay card feature tour must still be remembered across a restart on both desktop and mobile, and no pay card state should leak `isOpen` / `params` into storage.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34894](https://ledgerhq.atlassian.net/browse/LIVE-34894)
- Depends on the app-side persistence wiring from LIVE-34861 / LIVE-34862; consumed by the Story 2 filter UI.

[LIVE-34894]: https://ledgerhq.atlassian.net/browse/LIVE-34894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/domain/entity/pay-card/src/constants.ts
+++ b/domain/entity/pay-card/src/constants.ts
@@ -1,0 +1,1 @@
+export const PAY_CARD_BALANCE_FILTER_ALL = "all";
--- a/domain/entity/pay-card/src/slice.ts
+++ b/domain/entity/pay-card/src/slice.ts
@@ -1,10 +1,17 @@
 import { createSlice, type PayloadAction } from "@reduxjs/toolkit";
-import type { PayCardParams, PayCardPersistedState, PayCardState } from "./types";
+import { PAY_CARD_BALANCE_FILTER_ALL } from "./constants";
+import type {
+  PayCardBalanceFilter,
+  PayCardParams,
+  PayCardPersistedState,
+  PayCardState,
+} from "./types";
 
 export const payCardInitialState: PayCardState = {
   isOpen: false,
   params: null,
   hasSeenFeatureTour: false,
+  balanceFilter: PAY_CARD_BALANCE_FILTER_ALL,
 };
 
 export const payCardSlice = createSlice({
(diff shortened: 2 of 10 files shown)


deepyjr (approved):

domain/entity/pay-card/src/slice.ts
```diff
@@ -25,14 +32,20 @@ export const payCardSlice = createSlice({
     resetPayCardFeatureTourSeen: state => {
       state.hasSeenFeatureTour = false;
     },
+    setPayCardBalanceFilter: (state, action: PayloadAction<PayCardBalanceFilter>) => {
+      state.balanceFilter = action.payload;
+    },
```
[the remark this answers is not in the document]
mcayuelas-ledger: Should not happen

domain/entity/pay-card/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const PAY_CARD_BALANCE_FILTER_ALL = "all";
```
thesan, line 1: Just a suggestion, not a blocker, but it looks to me like this just acts as a hidden enum. IMO, it would be better to just use an actual enum. WDYT?

thesan (approved):

domain/entity/pay-card/src/constants.ts
mcayuelas-ledger: WDYM by Hidden enum? 

It's only a string so for me an enum could be a little bit overkill here

PR #20665, opened 2026-08-11
[LWDM] fix(wallet-api): always enforce deeplink scheme allowlist

SashaMIT:
### Description

`custom.deeplink.open` only called `isUrlSafe` when `lwmDeeplinkOpenHardening` / `lwdDeeplinkOpenHardening` was enabled. Those flags are created with `flag()` and default to `enabled: false`, so a Live App granted `custom.deeplink.open` could open dangerous schemes (`file://`, `smb://`, `javascript:`, …). Mobile sinks to `Linking.openURL`.

Previous behaviour (fail-open): scheme check skipped when the remote flag is unset/false.
Fix: always enforce the scheme allowlist. Keep the feature flag only for the locked-app deferral on mobile (`useDeeplinking` on desktop still uses the flag for lock gating).

Automated check: CustomHandlers unit tests now assert `file:///etc/passwd` is blocked even when the hardening flag is disabled (and still block the other unsafe schemes).

### Context

- **JIRA / GitHub issue**: n/a (hardening)
- **ADR** (if any): n/a

### Attacker model

A Wallet API Live App (or compromised partner app) with `custom.deeplink.open`, on a client where the remote flag is unset/false (code default). The Live App does not need OS shell; it only needs the deeplink permission.

### Test plan

- [x] Updated `ledger-live-mobile` / `ledger-live-desktop` CustomHandlers unit tests for always-on scheme allowlist
- [x] Local red/green harness: patched path blocks `file://` with flag off; reverted path opens it
- [ ] CI

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/CustomHandlers.ts
+++ b/apps/ledger-live-desktop/src/renderer/components/WebPlatformPlayer/CustomHandlers.ts
@@ -21,25 +21,24 @@ import { handlers as deeplinkHandlers } from "@ledgerhq/live-common/wallet-api/C
 import { isUrlSafe } from "@ledgerhq/live-common/wallet-api/CustomDeeplink/isUrlSafe";
 import { handlers as liveAppModalHandlers } from "@ledgerhq/live-common/wallet-api/LiveAppModal/server";
 import { resolveLiveAppModalParams } from "@ledgerhq/live-common/wallet-api/LiveAppModal/types";
-import { useFeature } from "@features/platform-feature-flags";
 
 type DeeplinkOpenHandlerParams = { url: string };
 
 type CreateDeeplinkOpenHandlerParams = {
-  isDeeplinkOpenHardeningEnabled: boolean;
   openDeepLink?: (url: string) => void;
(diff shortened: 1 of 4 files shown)


Justkant:
Hey thank you for the PR, the FF is not enabled yet because it was breaking some live-apps IIRC, so we need to make some other changes before being able to remove the FF and enforce it all the time

I'm going to do the tests and changes soon and let you know when it's ready but I don't think we'll be able to merge this now

SashaMIT:
Thanks @Justkant. Makes sense to hold while the live-app fallout from the FF is sorted. Happy to wait, or to adjust this PR once you have the companion changes ready. No rush on merge from my side.

SashaMIT:
Still relevant. Holding per @Justkant while the live-app companion changes land. Happy to adjust once those are ready.

Justkant:
Hey I'm back from vacation, I'll deal with this topic in the week

SashaMIT:
Welcome back! I'll stay parked on this one until you have the companion changes.

SashaMIT:
Still holding for the companion changes. Commenting so the stale timer does not close this.

PR #20666, opened 2026-08-11
ci(support): add build-and-test-pr swimlane

lewisd5:
## Problem

#20626 unified scope testing behind `test-scope-reusable.yml` and deliberately dropped `support/` from the existing `test-libs-reusable.yml` job. That left the two `@support/*` packages (jest-devtools, jest-features-flow) with no CI checks on PRs that touch them.

## Change

- **New `test-support-reusable.yml`** — thin delegator to `test-scope-reusable.yml` with `scope: support`, matching the pattern introduced in #20626.
- **`build-and-test-pr.yml`** — new `test-support` job gated on `contains(...paths, 'support')`, added to the `ok` gate.

The `scope:support` nx tag is already inferred automatically by the `project-tags` plugin, so no project-level changes are needed. Coverage is a no-op (neither package has a `coverage` script); `typecheck` and `format:check` will run via `nx-codechecks`.

Test run: https://github.com/LedgerHQ/ledger-live/actions/runs/31479893765/job/93746328673
Ticket: https://ledgerhq.atlassian.net/browse/LIVE-35675

### DIFF
--- a/.github/workflows/test-support-reusable.yml
+++ b/.github/workflows/test-support-reusable.yml
@@ -1,0 +1,42 @@
+name: "[Support] - Test - Called"
+
+on:
+  workflow_call:
+    outputs:
+      coverage_generated:
+        value: ${{ jobs.test-support.outputs.coverage_generated }}
+
+    secrets:
+      AWS_ACCOUNT_ID_PROD:
+        required: false
+      AWS_CACHE_OIDC_ROLE_ARN_BRANCH:
+        required: false
+      AWS_CACHE_OIDC_ROLE_ARN_DEVELOP:
+        required: false
+      AWS_CACHE_REGION:
+        required: false
+      AWS_CACHE_ROLE_NAME:
+        required: false
+      NX_KEY:
+        required: false
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: |
+          If you run this manually, and want to run on a PR, the correct ref should be refs/pull/{PR_NUMBER}/merge to
+          have the "normal" scenario involving checking out a merge commit between your branch and the base branch.
+          If you want to run only on a branch or specific commit, you can use either the sha or the branch name instead (prefer the first version for PRs).
+        required: false
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test-support-reusable.yml
```diff
@@ -0,0 +1,42 @@
+name: "[Support] - Test - Called"
```
LL782, line 1: I'd like to understand this name better

.github/workflows/test-support-reusable.yml
```diff
@@ -0,0 +1,42 @@
...
+          If you want to run only on a branch or specific commit, you can use either the sha or the branch name instead (prefer the first version for PRs).
+        required: false
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  test-support:
+    name: "Support Test"
```
LL782, line 37: "Test Support Directory: would be less ambiguous in my opinion. If it's a standard convention for all test jobs, and the "directory" can be inferred then I still think "Test: Support"  is clearer than "Support Test" which sparks misleading ideas.

LL782 (commented):
A couple of comments that pick at the exact names of these workflows and jobs. Otherwise it all looks good

.github/workflows/test-support-reusable.yml
lewisd5: There's a doc on CI naming conventions here: https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/5775295007/Workflow+Job+Naming

.github/workflows/test-support-reusable.yml
LL782: This is explained in internal docs: https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/5775295007/Workflow+Job+Naming

LL782 (approved):

dgreen-ledger (approved):

PR #20668, opened 2026-08-11
[LWDM] feat(lwd): no contact screen in recipient screen

dilaouid:
### 📝 Description

Display the "Contact will appear here" element in the recipient screen of LWD
Only if no contact having account in the chain we are sending to and that family is eligible to address book

image

### 🔗 Context

- **[JIRA / GitHub issue](https://ledgerhq.atlassian.net/browse/LIVE-35767)**: 
- **ADR** (if any):


libs/ledger-live-common/src/flows/send/recipient/utils/resolveRecipientNetworkId.ts
Moustafa-Koterba, line 1: **nit** feel more like `extractNetworkId`, but matter of taste

Moustafa-Koterba (approved):

PR #20675, opened 2026-08-11
[LWDM] feat(casper): coin tester (LIVE-34363)

mateuszpalosz-ext:
## Summary

Adds `@ledgerhq/coin-tester-casper` — the integration test package for the Casper (CSPR) coin module, following the same pattern as other coin tester modules.

The package spins up a local 4-node Casper devnet via Docker and runs two test suites against it:

- **`signer.test.ts`** — tests the software signer with keys generated per-run (no committed private-key material); covers `getAddressAndPubKey`, `showAddressAndPubKey`, signing a valid transaction, rejecting a tampered payload hash, and rejecting an unknown derivation path
- **`scenarii.test.ts`** — end-to-end send scenarios: send with transfer ID, send without transfer ID, send with a large transfer ID; assertions cover operation type, value, fee, sender/recipient account hashes, balance delta, and `operation.extra`

### Key implementation details

- **Devnet** (`casperDevnet.ts`): wraps `docker-compose` to start/stop the `veles-labs/casper-devnet` image; derives user keys inside the container via `casper-devnet derive`; handles the `docker-compose` non-`Error` rejection shape; readiness is delegated entirely to the Docker healthcheck (`--wait` flag + `is-ready` healthcheck in `docker-compose.yml`)
- **Signer** (`signer.ts`): parses the Casper calltable binary format, verifies `blake2b256(payload) == embedded hash` before signing — catches payload tampering at the signer rather than at broadcast
- **Indexer** (`indexer.ts`): MSW-based in-process mock; intercepts the indexer HTTP endpoint and serves history built from real devnet RPC results, indexed under both sender and recipient public keys so sync produces both `OUT` and `IN` operations
- Devnet image pinned by SHA-256 digest; marked `platform: linux/amd64` so arm64 hosts fail the pull rather than silently exec-ing a foreign binary

### 🔗 Context

- **JIRA / GitHub issue**:[https://ledgerhq.atlassian.net/browse/LIVE-34363]( https://ledgerhq.atlassian.net/browse/LIVE-34363)


Sebastien-Dav1d (dismissed):

henri-ly (dismissed):

libs/coin-tester-modules/coin-tester-casper/src/casperDevnet.ts
```diff
@@ -0,0 +1,123 @@
...
+    const { out } = await compose.exec(
+      DEVNET_SERVICE_NAME,
+      `casper-devnet network ${DEVNET_CHAIN_NAME} is-ready`,
+      composeOpts(),
+    );
+    return out.trim();
+  } catch (error) {
+    return `command itself failed: ${formatComposeError(error)}`;
+  }
+}
```
francois-guerin-ledger: This is already covered by docker compose healthcheck

libs/coin-tester-modules/coin-tester-casper/src/devnet.test.ts
```diff
@@ -0,0 +1,79 @@
+import console from "console";
```
francois-guerin-ledger: This kind of tests must belong to the coin module. Coin testers should focus on scenarios.

libs/coin-tester-modules/coin-tester-casper/src/negativeCases.test.ts
```diff
@@ -0,0 +1,106 @@
+import console from "console";
```
francois-guerin-ledger: We should not have these tests here.

libs/coin-tester-modules/coin-tester-casper/src/devnet.test.ts
mateuszpalosz-ext: Removed :)

libs/coin-tester-modules/coin-tester-casper/src/negativeCases.test.ts
mateuszpalosz-ext: Removed :)

libs/coin-tester-modules/coin-tester-casper/src/casperDevnet.ts
mateuszpalosz-ext: Updated :)

francois-guerin-ledger (approved):

Sebastien-Dav1d (approved):

PR #20680, opened 2026-08-11
fix(e2e): reveal targets clear of sticky layers, drop dead scroll workarounds

VicAlbr:
### 📝 Description

Parts 5 and 6 of QAA-1442, together. **45 insertions, 95 deletions — a net removal of 50 lines.**

#### Part 5 (QAA-1460) — one helper replaces three magic numbers

Three call sites had hand-tuned "slack" scrolls past sticky layers, with three different constants (200 / 100 / 200) and duplicated explanations. They are one defect patched three times: `scrollToId` stops the moment the target is *minimally* visible, which is always the edge it entered through — and the edges are exactly where the sticky header and footer sit. On Android the tap is then accepted and swallowed, because `visibility` is `View.getLocalVisibleRect()` and ignores occlusion by siblings.

```ts
async revealForTap(id, scrollViewId?, direction = "down") {
  await NativeElementHelpers.scrollToId(id, scrollViewId, undefined, direction);
  await NativeElementHelpers.scrollByPixels(scrollViewId, EDGE_CLEARANCE_PIXELS, direction);
}
```

The clearance always goes **the same way as the search**, which is what moves the target off the edge it came in through. `grep scrollByPixels page/ specs/` now returns nothing.

**`revealById` and the measured safe rect from the ticket were dropped** — see the [QAA-1460 comment](https://ledgerhq.atlassian.net/browse/QAA-1460) for why: the ticket's two acceptance criteria turned out not to be overlay problems at all (one is a 60s wait with no scroll, the other is Espresso refusing to type into a focused `ReactEditText`, i.e. the keyboard). The Receive testIDs were added and then reverted, because no spec taps those controls — production code with zero consumers.

#### Part 6 (QAA-1458) — four subtractions

| | Change |
|---|---|
| **A** | `swipeToNetworkItem` exhausted 5 swipes and **returned normally**, so the failure surfaced later as a generic tap error. It now throws, naming the item, the attempt count and the container. This is the call behind 12 `modular-drawer-network-selection-scrollView not found` errors in one CI shard |
| **B** | `finishAccountsDiscovery`'s `retryUntilTimeout` was **dead code**: it only re-runs on throw, and `waitForElementNotVisible` returns `false` instead of throwing, so a Continue button that never dismissed counted as success. The result is now raised, with a 5s probe so the retry budget allows an actual re-tap |
| **C** | Three `toBeGreaterThan` assertions after "show all" used `countElementsById`, which counts only rows **≥75% visible** — a full list longer than the viewport reads as ≤5 and false-fails. They now use the existing `countElements(getElementsById(...))` |
| **D** | The Wallet 4.0 **tour guard is deleted** (115 → 38 lines). `tour: false, q2Tour: false` are forced in both flag presets and no spec overrides them, so it guarded a drawer that cannot render — while costing an Android-only 1500ms sleep, wasted probes at 13 call sites, and 4 hardcoded English strings. The analytics half is kept: it still fires for `marketBanner.spec.ts`, whose fixture omits `hasSeenAnalyticsOptInPrompt` |

### ✅ Verification

Local device runs, `E2E_RETRIES=0` throughout so no flake is masked by a retry.

| Spec | Covers | Android | iOS |
|---|---|---|---|
| `marketBanner` | D (only spec where the analytics drawer appears) | 1/1 | 1/1 |
| `wallet40Q2/walletAssets` | D + Part 5 | 5/5 | 5/5 |
| `wallet40Q2/operationsHistory` | D + Part 5 | 3/3 | 3/3 |
| `wallet40Q2/assetAggregationMarketDetail` | D + Part 5 | 4/4 | 3/4 — see below |
| `wallet40Q2/portfolio` | A + B | 7/7 | — |

**CI-time improvement:** `walletAssets` on Android went **244s → 163s** once D removed the dead tour probes, before the sleep removal was even counted.

### ⚠️ Two things a reviewer should know

**1. The iOS `assetAggregationMarketDetail` failure is pre-existing.** `[DAI (Polygon)] - Aggregated assets show one row per asset…` fails in `scrollToAddressItem` (`assetDetail.page.ts`), code this PR does not touch. Proven by reverting `e2e/mobile` to `develop` and re-running on iOS: **it fails identically there** (133s, same test). It passes on Android. The mechanism is that `scrollToAddressItem` re-searches from scratch per address with a fixed `700, "down"`, so the second (Ethereum) address overshoots the addresses section entirely. Logged to QAA-1458.

**2. Change C is a fix to code that does not currently execute.** All three lines are in legacy `else` branches, and every spec reaching them is `describe.skip`'d in Q2 (`isQ2WithAggregatedAssets` / `isQ2WithOperationsList`). So CI will neither prove nor disprove it — it is zero-risk, and also unverified by design. If those legacy paths are genuinely dead, **deleting them** would be better than fixing assertions inside them; that decision is deliberately not taken here.

### 🔗 Context

- **JIRA**: [QAA-1460](https://ledgerhq.atlassian.net/browse/QAA-1460) (Part 5) and [QAA-1458](https://ledgerhq.atlassian.net/browse/QAA-1458) (Part 6, items A–D), both under [QAA-1442](https://ledgerhq.atlassian.net/browse/QAA-1442)
- QAA-1458 stays **open**: the remaining backlog (10 omitted-container scroll calls, the 35% visibility threshold, the regex + `"bottom"` scroll, unexplained magic pixels, `scrollOnce`'s inverted-direction fallback, and `isIdVisible`'s multiple-match trap) is recorded there. Each changes scroll behaviour and needs its own device run, which is why it is not bundled here.

[QAA-1460]: https://ledgerhq.atlassian.net/browse/QAA-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/mobile/helpers/elementHelpers.ts
+++ b/e2e/mobile/helpers/elementHelpers.ts
@@ -36,6 +36,9 @@ export const QUICK_VISIBILITY_PROBE_TIMEOUT = 500;
 
 const DEFAULT_WEB_ELEMENT_INTERVAL = 2000;
 
+/** How far past the viewport edge revealForTap pushes a target so an overlay cannot intercept it. */
+const EDGE_CLEARANCE_PIXELS = 200;
+
 export type WaitForElementOptions = {
   errorCheckTimeout?: number;
   errorElementId?: string;
@@ -327,6 +330,25 @@ export const NativeElementHelpers = {
     direction: Direction = "down",
   ): Promise<void> {
     await scroller.scrollByPixels(scrollViewId, pixels, direction);
+  },
+
+  /**
+   * Scrolls to an element and leaves it clear of the viewport edge, so a tap reaches it.
+   *
+   * `scrollToId` stops as soon as the target is minimally visible, which is the edge it entered
+   * through — where the sticky header and footer sit. Android reports such an element as visible
+   * anyway (`visibility` is View.getLocalVisibleRect(), which ignores occlusion by siblings), so the
+   * tap is accepted and then swallowed. Continuing in the same direction clears that edge.
+   *
+   * Use before a tap; `scrollToId` stays correct for assertions.
+   */
+  async revealForTap(
+    id: string | RegExp,
+    scrollViewId?: string | RegExp,
+    direction: Direction = "down",
+  ): Promise<void> {
+    await NativeElementHelpers.scrollToId(id, scrollViewId, undefined, direction);
+    await NativeElementHelpers.scrollByPixels(scrollViewId, EDGE_CLEARANCE_PIXELS, direction);
   },
 
   async getAttributesOfElement(id: string | RegExp, index = 0): Promise<Detox.ElementAttributes> {
--- a/e2e/mobile/helpers/elementHelpers.ts
+++ b/e2e/mobile/helpers/elementHelpers.ts
@@ -39,6 +39,13 @@ const DEFAULT_WEB_ELEMENT_INTERVAL = 2000;
 /** How far past the viewport edge revealForTap pushes a target so an overlay cannot intercept it. */
 const EDGE_CLEARANCE_PIXELS = 200;
 
+export type RevealForTapOptions = {
+  /** Scroll container. Omitted means the engine guesses the first scrollable by type. */
+  container?: string | RegExp;
+  /** Search direction; the edge clearance always continues the same way. Defaults to "down". */
+  direction?: Direction;
+};
+
 export type WaitForElementOptions = {
   errorCheckTimeout?: number;
   errorElementId?: string;
@@ -342,13 +349,10 @@ export const NativeElementHelpers = {
    *
    * Use before a tap; `scrollToId` stays correct for assertions.
    */
-  async revealForTap(
-    id: string | RegExp,
-    scrollViewId?: string | RegExp,
-    direction: Direction = "down",
-  ): Promise<void> {
-    await NativeElementHelpers.scrollToId(id, scrollViewId, undefined, direction);
-    await NativeElementHelpers.scrollByPixels(scrollViewId, EDGE_CLEARANCE_PIXELS, direction);
+  async revealForTap(id: string | RegExp, options: RevealForTapOptions = {}): Promise<void> {
+    const { container, direction = "down" } = options;
+    await NativeElementHelpers.scrollToId(id, container, undefined, direction);
+    await NativeElementHelpers.scrollByPixels(container, EDGE_CLEARANCE_PIXELS, direction);
   },
 
   async getAttributesOfElement(id: string | RegExp, index = 0): Promise<Detox.ElementAttributes> {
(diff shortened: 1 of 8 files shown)


VicAlbr:
## 🧪 Triggered test workflows

Manually dispatched on `test/e2e-QAA-1460-clear-edges` (commit `cb1ff1f845b`):

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | **full suite** (no `test_filter`), iOS & Android, Speculos nanoX | [run 31503761952](https://github.com/LedgerHQ/ledger-live/actions/runs/31503761952) |

Scope rationale: full suite rather than a filter because these changes touch shared harness code used across the suite — `revealForTap` in `elementHelpers`, `swipeToNetworkItem` in the modular drawer (reached by receive / deposit / send / swap / addAccount flows), `finishAccountsDiscovery`, and the `wallet40Drawers` guard called from 13 spec sites. Desktop E2E and Coin Tester skipped: nothing outside `e2e/mobile` is touched.

Local pre-verification (`E2E_RETRIES=0`, both platforms) is in the PR description. Two known items for whoever reads the results:

- **iOS `assetAggregationMarketDetail`** will likely be red on `[DAI (Polygon)] - Aggregated assets show one row per asset…`. **Pre-existing** — proven by reverting `e2e/mobile` to `develop` and re-running on iOS, where it fails identically. Not caused by this PR.
- **Change C** touches legacy `else` branches whose specs are `describe.skip`'d in Q2, so CI cannot exercise it either way.

Comparison baseline: the same-day `develop` nightly, judged the way QAA-1457 was — passed/failed counts and no shard slower than the median + 1 min.

e2e/mobile/helpers/elementHelpers.ts
```diff
@@ -329,6 +332,25 @@ export const NativeElementHelpers = {
...
+   *
+   * `scrollToId` stops as soon as the target is minimally visible, which is the edge it entered
+   * through — where the sticky header and footer sit. Android reports such an element as visible
+   * anyway (`visibility` is View.getLocalVisibleRect(), which ignores occlusion by siblings), so the
+   * tap is accepted and then swallowed. Continuing in the same direction clears that edge.
+   *
+   * Use before a tap; `scrollToId` stays correct for assertions.
+   */
+  async revealForTap(
+    id: string | RegExp,
```
cunhabruno: Could you use object as optional args? So it is easier for callers decide which arguments are needed

VicAlbr: Done in c2d006be715 — good call, it removed noise at two of the three call sites.

```ts
async revealForTap(id: string | RegExp, options: RevealForTapOptions = {}) {
  const { container, direction = "down" } = options;
```

Two of the three no longer need to state a direction at all, and the one that does now reads as intent rather than as a positional slot:

```ts
await revealForTap(this.operationsListItemId, { container: scrollViewId });
await revealForTap(testId, { container: this.emptyPortfolioListId, direction: "up" });
```

One thing worth flagging: this makes `revealForTap` the only scroll helper in `elementHelpers.ts` taking options, while its four siblings are still positional — `scrollToId(id, scrollViewId?, pixels?, direction?, androidDelay?, visibilityPercentage?)` is the extreme case. I think your form is the better one and the new convention should start here, so I have not tried to match the old style. Migrating the siblings is a bigger change (86 call sites) and it was actually already in QAA-1457, so I have added it to QAA-1487 with the rest of the scroll backlog rather than growing this PR.

Verified: typecheck/oxfmt/oxlint clean, plus `walletAssets` 5/5 on Android — signature-only change, but that spec is the one exercising the `"up"` path.

cunhabruno (approved):

PR #20681, opened 2026-08-11
[LWDM] feat(lwd): add the send flow add contact step UI (LIVE-35774)

YazhuEth:
### 📝 Description

The "Add contact" button on the recipient card was inert (no `onClick`). It now opens a new step of the send flow offering two choices: create a new contact, or add the address to an existing contact.

**UI only**, as per the ticket: neither row is wired to a destination yet. They are deliberately left without a handler — Lumen's `ListItem` derives its interactive state from `onClick`, so an unwired row correctly renders as inert instead of offering a hover and a pointer that lead nowhere.

**How** — a new `ADD_CONTACT` step in `SEND_FLOW_STEP`, registered as `floating` like `RECENT_HISTORY`: the back arrow and the close cross come from the flow's existing `DialogHeader`, no recipient input is shown, and the back arrow returns to the recipient step with the address kept. Rows are Lumen `ListItem` + `Spot`, same recipe as `ReceiveOptionsView`.

Desktop only. The mobile recipient step has no add-contact affordance today, so mobile only gets a config entry to satisfy the `Record<SendFlowStep, SendStepConfig>` contract.

Screenshot 2026-08-11 at 17 49 39

Tested: the new view (both rows rendered, both handlers called), and the recipient card button opening the step. The 88 existing send flow tests still pass, and typecheck passes on both apps.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35774](https://ledgerhq.atlassian.net/browse/LIVE-35774)
- **ADR** (if any):

[LIVE-35774]: https://ledgerhq.atlassian.net/browse/LIVE-35774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/screens/AddContact/AddContactScreen.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/screens/AddContact/AddContactScreen.tsx
@@ -1,0 +1,11 @@
+import React from "react";
+import { DialogBody } from "@ledgerhq/lumen-ui-react";
+import { AddContactView } from "./AddContactView";
+
+export function AddContactScreen() {
+  return (
+    <DialogBody className="px-16 pb-24">
+      <AddContactView />
+    </DialogBody>
+  );
+}
--- a/apps/ledger-live-desktop/src/mvvm/features/Send/screens/AddContact/AddContactScreen.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Send/screens/AddContact/AddContactScreen.tsx
@@ -1,11 +1,17 @@
 import React from "react";
 import { DialogBody } from "@ledgerhq/lumen-ui-react";
 import { AddContactView } from "./AddContactView";
+import { useAddContactViewModel } from "./hooks/useAddContactViewModel";
 
 export function AddContactScreen() {
+  const { onAddNewContact, onAddToExistingContact } = useAddContactViewModel();
+
   return (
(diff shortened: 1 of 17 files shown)


dilaouid (commented):
Nice PR !
Be careful with the spacing though, in your version (your screen) it seems to have too much spacing between the rows

Image

apps/ledger-live-desktop/src/mvvm/features/Send/screens/AddContact/AddContactView.tsx
dilaouid, line 1: You might need to create a `useAddContactViewModel` to respect the mvvm pattern
Screen → props → View

basically for every screen we need to have:
Container → ViewModel → View

apps/ledger-live-desktop/src/mvvm/features/Send/screens/AddContact/AddContactScreen.tsx
dilaouid, line 1: So you might have here something like:

```tsx
const { onAddNewContact, onAddToExistingContact } = useAddContactViewModel();
return (
  <DialogBody …>
    <AddContactView
      onAddNewContact={onAddNewContact}
      onAddToExistingContact={onAddToExistingContact}
    />
  </DialogBody>
);
```

dilaouid (approved):

PR #20702, opened 2026-08-12
[LWDM] feat(card): scaffold Card API DDD packages (LIVE-35794)

mcayuelas-ledger:
### 📝 Description

Scaffolds the **Card API** across the DDD layers, following the existing CoinMarketCap / DADA pattern: one endpoint-less RTK Query service per backend, use-case endpoints injected by `domain/api/*`, cross-flow glue in `features/platform/*`, and store registration in the apps.

There is a **single Card backend api**, `cardApi`. Card use cases inject into it — Management now, Auth later once it migrates off the holdout (see below). No dual-service design.

**What's added**

- `@shared/api-services` → `services/card`: endpoint-less `cardApi` (`cardApiExtra`, `getCardExtra`, `CardApiExtra`), transport-only base query that attaches `Authorization: Bearer <session>` and does one refresh + retry on `401`. It does **not** go through `@shared/auth` / AuthSDK — the token is owned by the Card session.
- `@domain/api-card-management`: `cardManagementApi = cardApi.enhanceEndpoints(...).injectEndpoints(...)` — an empty endpoint scaffold plus `CARD_MANAGEMENT_TAGS`. Endpoints land with the first real contract.
- `@features/platform-card`: in-memory Card session (`cardSession`) and the `getCardSessionToken` / `refreshCardSession` accessors the apps inject into `cardApiExtra`. The session token is never persisted.
- `CARD_API_URL` env in `@shared/env`.
- Both apps register `cardApi` in their RTK registry and wire `cardApiExtra` with the platform session accessors.

> [!IMPORTANT]
> **Existing \`payCardApi\` holdout to migrate — do not extend it.**
> The Card **Auth** UI still runs on the legacy \`@domain/api-pay-card\` (\`payCardApi\`), a standalone \`createApi\` with a mock-by-URL base query. It is **not** a second Card service: under **LIVE-33829** its Auth endpoints move onto \`cardApi\` (as \`@domain/api-card-auth\`), Card Auth starts calling \`cardSession.set(...)\` on login, \`refreshCardSession\` gains a real renewal, and \`payCardApi\` is deleted.
> Concretely, once that migration lands we must **remove the forced-mocks block** \`payCardApiExtra({ payCardApiMocksEnabled: true })\` in both \`apps/ledger-live-mobile/src/state-manager/configureStore.ts\` (\`:74-77\`) and \`apps/ledger-live-desktop/src/state-manager/configureStore.ts\`, and drop the \`payCardApi\` reducer registration.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35794](https://ledgerhq.atlassian.net/browse/LIVE-35794)
- **Follow-up**: [LIVE-33829](https://ledgerhq.atlassian.net/browse/LIVE-33829) — migrate Card Auth onto \`cardApi\` and remove the \`payCardApi\` holdout

### 🧪 Notes for QA

No user-facing change: \`cardApi\` has no endpoints yet, so nothing is fetched. \`CARD_API_URL\` defaults to a placeholder host pending the real endpoint. The existing Card Auth flow (\`payCardApi\`, forced mocks) is untouched.

[LIVE-35794]: https://ledgerhq.atlassian.net/browse/LIVE-35794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


liviuciulinaru (dismissed):

LucasWerey (dismissed):

shared/api-services/README.md
```diff
@@ -1,39 +1,39 @@
 # @shared/api-services
 
-> [!NOTE]
-> **Status: STABLE** — Production-ready; API is considered stable.
+> [!NOTE] > **Status: STABLE** — Production-ready; API is considered stable.
```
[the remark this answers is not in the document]
LL782: True

shared/api-services/README.md
```diff
@@ -106,10 +106,8 @@ because consumers like `apps/web-tools`, `apps/cli` and `buildStandaloneCryptoAs
 one service only and must not be forced to supply config for services they never call.
 
 A base query added here must be **pure transport** — no endpoint URLs, no endpoint-name lookups, no wire
-schemas. `@domain/api-pay-card` is the one backend not yet here: its base query resolves mock responses
-keyed by endpoint URL, so it needs that untangling first. It is a holdout to migrate, not a precedent.
+schemas.
```
LL782, line 110: It looks like this is the only real change in this file. Everything else is formatter churn and in the case of line 111 and the 3 it's actually distructive

mcayuelas-ledger: Yeah the formatting is weird on this package

shared/api-services/src/services/card/schema.ts
```diff
@@ -0,0 +1,26 @@
...
+ * client key and supplies the session-token accessors owned by the Card Auth session, so this package
+ * holds no env, auth-state or store dependency.
+ *
+ * `getCardSessionToken` returns the current session token (or a nullish value when logged out);
+ * `refreshCardSession` is invoked once after a `401`, returning the refreshed token or a nullish value
+ * when the session cannot be renewed.
+ *
+ * `cardBaanxClientKey` is the public Baanx tenant key sent as `x-client-key` on every request. Baanx
+ * rejects calls without it (HTTP 499); the env default is empty until the production key is provisioned.
+ */
```
LL782: Personally I don't think this is the place for this kind of documentation. And I have a long standing concern about the overhead of having to maintain comments and the risk involved in them going out of date and becoming misleading.

So, my suggestion would be to omit it and if it's important put it in the README for the package


```suggestion
```

features/platform/card/README.md
```diff
@@ -0,0 +1,30 @@
+# @features/platform-card
+
+> [!CAUTION] > **Status: UNSTABLE** — Part of the emerging DDD layer; under active development.
```
LL782: This won't layout as hoped. I've found that adding an extra line prevents the formatter from crushing it


```suggestion
> [!CAUTION] 
>
> **Status: UNSTABLE** — Part of the emerging DDD layer; under active development.
```

features/platform/card/src/session/cardSession.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    expect(cardSession.getToken()).toBeNull();
+  });
+});
+
+describe("refreshCardSession", () => {
+  afterEach(() => {
+    cardSession.clear();
+  });
+
+  it("clears the session and reports it cannot be renewed (scaffold)", async () => {
```
LL782: "reports it cannot be renewed" this seems inaccurate to me.
WDYT?


```suggestion
  it("clears the session token", async () => {
```

features/platform/card/src/session/cardSession.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Refresh handler passed to `cardApiExtra`: the shared Card base query calls this once after a `401`.
+ *
+ * Scaffold behaviour: there is no refresh contract yet (the Card session entity carries no refresh
+ * token), so this clears the session and reports that it cannot be renewed, which surfaces the `401`
+ * to the caller. Card Auth replaces this with a real renewal once it migrates onto `cardApi`.
+ */
+export async function refreshCardSession(): Promise<string | null | undefined> {
+  cardSession.clear();
+  return null;
+}
```
LL782: "reports that it cannot be renewed" 

Is this true?

mcayuelas-ledger: I just moved some code already existing so idk for this in fact

domain/api/card-management/README.md
```diff
@@ -0,0 +1,18 @@
+# @domain/api-card-management
+
+> [!CAUTION] > **Status: UNSTABLE** — Part of the emerging DDD layer; under active development.
```
LL782, line 3: Same formatted bug here. Try an extra line to separate them


```suggestion
> [!CAUTION] 
>
> **Status: UNSTABLE** — Part of the emerging DDD layer; under active development.
```

domain/api/card-management/src/constants.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * RTK Query cache tags owned by the Card Management use case. Registered on the shared `cardApi` via
+ * `enhanceEndpoints({ addTagTypes })`, so the shared service never has to know they exist.
+ */
```
LL782, line 4: I feel it's safer to gather this kind of knowledge on demand, i.e. if you want to know where this is used and what it's doing then take a look at that point in time. Relying on comments for that knowledge carries risk and maintaining comments like this creates extra work and increased context windows

LL782 (commented):
The comments in these files seem over the top to me:

- shared/api-services/src/services/card/constants.ts
- domain/api/card-management/src/api.ts
- features/platform/card/src/session/cardSession.ts
- shared/api-services/src/services/card/api.ts

Issues:

1. They talk too much about the current plan and work in progress. This will not be useful long term. 
2. They create extra context that needs to be considered, on top of the code itself.
3. In some places we can't verify if they are accurate so this illustrates the risk keeping them.

LL782 (approved):

LucasWerey (approved):

PR #20704, opened 2026-08-12
[LWM] fix(mobile): flush Segment analytics on AppState background

tonykhaov:
### 📝 Description

Some LWM users were identified in Mixpanel but their track events never arrived. The analytics overlay logged before the Segment SDK call, so it only proved local `track()` ran, not that the event was queued. Background flush lived in a React hook (`useFlushInBackground`) and could miss AppState transitions.

This PR makes enqueue and flush observable, and flushes when the app actually backgrounds.

- Create the Segment client before the first `await` in `start()` so startup events are queued instead of dropped
- Await Segment enqueue so a flush right after `track` sees the event
- Replace the React flush hook with an `AppState` listener in `start()` that flushes on `background` only (`inactive` is Control Centre, calls, and the iOS transition into background)
- Log delivery status on the overlay after the SDK call: `enqueued`, `failed`, `skipped_no_store`, `skipped_no_client`, `skipped_no_token`, `flushed`
- If `segmentClient.track()` throws, log `failed` and resolve so product flows are not broken
- If `pendingEvents()` throws, still run the real flush
- Warn once in Datadog when events are skipped with no Segment client
- Colour event names by status. Show the status label for skips, failures, and flushes. Hide it on normal `enqueued` rows

**Test coverage:** Yes. `segment.deliveryAndFlush.test.ts` and `segment.noStore.test.ts` cover background-only flush, empty-queue log gate, `pendingEvents` rejection, delivery statuses including `failed` and `skipped_no_store`, and the Datadog warn-once.

### Screenshots

https://github.com/user-attachments/assets/6df68ca3-3f4d-4f99-a150-57e9d3089c01

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-30464](https://ledgerhq.atlassian.net/browse/LIVE-30464)
- **ADR** (if any): N/A

[LIVE-30464]: https://ledgerhq.atlassian.net/browse/LIVE-30464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jiyuzhuang (dismissed):

mcayuelas-ledger (approved):
OK for wxp

jiyuzhuang (approved):

PR #20712, opened 2026-08-12
fix(coin-evm): zero gas limit and EIP-1559 fee headroom (LIVE-32644, LIVE-32650)

YazhuEth:
### 📝 Description

Two independent EVM fixes found while investigating broadcast failures on Datadog.

**1. Transactions signed with a zero gas limit (LIVE-32644)**

Previous behaviour: when gas estimation failed, `estimateGas` returned its `BigNumber(0)` fallback. That zero was propagated onto the transaction and read back at sign time by `prepareUnsignedTxParams`, whose presence check was `typeof customFeesParameters?.gasLimit === "bigint"` — and `0n` is a bigint. It was therefore treated as a deliberate custom gas limit, which **skipped re-estimation entirely**, and the transaction was signed with `gasLimit: 0`. The node rejected it with `intrinsic gas too low`.

A Datadog payload confirms it end to end (mainnet USDT transfer, `lld/4.15.0`):

```
nonce                1
to                   USDT, transfer()
maxFeePerGas         0.2389 gwei     ← fee estimation succeeded
maxPriorityFeePerGas 0.1057 gwei
gasLimit             0               ← only the gas estimation failed
```

The fix, in two layers:

- `prepareUnsignedTxParams` only honours a **positive** gas limit as a custom value, so a zero re-runs the estimation instead of disabling it. This is the root cause.
- `craftTransaction` refuses to craft when the resolved gas limit is not positive. This matters beyond the send flow: `signSwapEvm` and `signApprovalEvm` call `craftTransaction` **directly**, without going through `validateIntent`, so this guard is their only gate.

The `BigNumber(0)` fallback itself is intentionally left in place — it still lets `validateIntent` raise `FeeNotLoaded` and block the flow, which is the existing safety net. Only its ability to reach the device is removed.

Regression check: nothing in the monorepo relies on a zero gas limit being accepted. `sponsored` does not touch the gas limit, the CEX swap path deletes the field, the DEX path replaces a zero with a default, no chain config is concerned, and no test asserts a zero is propagated or signed.

**2. EIP-1559 max fee headroom (LIVE-32650)**

`EIP1559_BASE_FEE_MULTIPLIER` goes from 1.27 to 1.6. The Ledger gas tracker computes `maxFeePerGas = nextBaseFee × multiplier + priorityFee`; since the base fee grows by at most 12.5% per block, 1.27 only covered ~2 blocks (~24s), less than a device signing round trip. 1.6 covers 4 blocks. The RPC path already uses ×2, so chains on the Ledger explorer — including Ethereum mainnet — had the tightest margin of all.

Max fees displayed on those chains will be higher. **The amount actually paid is unchanged**: EIP-1559 charges base fee + priority fee, the max fee is only a ceiling. The visible effect is a slightly larger amount reserved on a send-max.

Automated checks preventing regression: two new tests in `craftTransaction.test.ts` cover re-estimation on a zero custom gas limit and the refusal to craft one.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-32644](https://ledgerhq.atlassian.net/browse/LIVE-32644), [LIVE-32650](https://ledgerhq.atlassian.net/browse/LIVE-32650)

[LIVE-32644]: https://ledgerhq.atlassian.net/browse/LIVE-32644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/evm-gas-limit-and-base-fee-headroom.md
+++ b/.changeset/evm-gas-limit-and-base-fee-headroom.md
@@ -1,0 +1,10 @@
+---
+"@ledgerhq/coin-evm": patch
+"@shared/env": patch
+---
+
+Fix EVM transactions being signed with a zero gas limit, and widen the EIP-1559 max fee headroom.
+
+When gas estimation failed, its `BigNumber(0)` fallback travelled back to the sign step, where it was read as a deliberate custom gas limit. That disabled re-estimation and produced a transaction the node rejected with `intrinsic gas too low`. A non-positive gas limit is no longer honoured as a custom value, so the estimation runs again, and crafting now fails rather than sending a zero gas limit to the device (LIVE-32644).
+
+`EIP1559_BASE_FEE_MULTIPLIER` goes from 1.27 to 1.6, so an estimated transaction stays includable for 4 blocks instead of 2 (the base fee grows by at most 12.5% per block). Max fees displayed on chains using the Ledger gas tracker will be higher, but the amount actually paid is unchanged: EIP-1559 charges the base fee plus the priority fee, and the max fee is only a ceiling (LIVE-32650).
--- a/apps/ledger-live-desktop/static/i18n/en/app.json
+++ b/apps/ledger-live-desktop/static/i18n/en/app.json
@@ -7302,6 +7302,10 @@
     "GasPriceTooLow": {
       "title": "Gas price is lower than the network minimum"
     },
+    "GasEstimationError": {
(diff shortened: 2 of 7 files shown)


YazhuEth:
Superseded by #20715: branch renamed to follow the git conventions and to cover both tickets.

PR #20714, opened 2026-08-12
[LWDM] feat(zcash): add warning for unconfirmed spendable balance

semeano:
### 📝 Description

Show a warning in the send flow when the Zcash private balance is selected as source and funds were shielded in the last 15 minutes, explaining that recently shielded funds need confirmations and scanning before they are spendable

### 🔗 Context

- **JIRA / GitHub issue**: https://ledgerhq.atlassian.net/browse/LIVE-35612

### DIFF
--- a/libs/coin-modules/coin-zcash/src/logic/account/balance.ts
+++ b/libs/coin-modules/coin-zcash/src/logic/account/balance.ts
@@ -1,6 +1,8 @@
 import { BigNumber } from "bignumber.js";
 import type { BitcoinOutput } from "../../types/bridge";
 import type { ZcashPrivateInfo } from "../../network/types";
+import { ZCASH_SHIELDED_SPENDABILITY_DELAY_BLOCKS } from "../../constants";
+import { getTxType } from "../../bridge/operations";
 
 type PrivateBalances = Pick<
   ZcashPrivateInfo,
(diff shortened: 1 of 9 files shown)


libs/coin-modules/coin-zcash/src/logic/account/balance.ts
```diff
@@ -38,6 +40,28 @@ export function getPrivateBalance(privateInfo: PrivateBalances | undefined | nul
...
+ * can warn that the spendable balance may temporarily trail the total. Incoming
+ * transactions are ignored — they do not gate spendability the same way.
+ */
+export function hasRecentlyShieldedFunds(
+  privateInfo: Pick<ZcashPrivateInfo, "transactions"> | undefined | null,
+  now: number = Date.now(),
+): boolean {
+  const transactions = privateInfo?.transactions;
+  if (!transactions?.length) return false;
+  const cutoff = now - ZCASH_SHIELDED_SPENDABILITY_DELAY_MINUTES * 60 * 1000;
```
may01: It a common approach to calculate confirmation blocks instead of time.

may01 (changes requested):

libs/coin-modules/coin-zcash/src/logic/account/balance.ts
semeano: how many confirmation blocks do you suggest?

may01: According to https://zcash.readthedocs.io/en/latest/rtd_pages/glossary.html
10+ confirmations should be enough (~ 12.5 minutes)

may01 (approved):

henri-ly (approved):

PR #20715, opened 2026-08-12
[LWDM] fix(coin-evm): zero gas limit and fee headroom (LIVE-32644, LIVE-32650)

YazhuEth:
### 📝 Description

Two independent EVM fixes, one per commit.

**Zero gas limit (LIVE-32644)**

When gas estimation failed, its `BigNumber(0)` fallback was written onto the transaction and read back at sign time by `prepareUnsignedTxParams`, whose check was `typeof gasLimit === "bigint"`. Since `0n` is a bigint, it was taken for a deliberate custom gas limit: re-estimation was skipped and the transaction was signed with `gasLimit: 0`, which the node rejects with `intrinsic gas too low`. A Datadog payload confirms it (mainnet USDT transfer, `lld/4.15.0`, fees estimated correctly, `gasLimit: 0`).

Only a positive gas limit is honoured as a custom value now, and `craftTransaction` refuses to craft a non-positive one — `signSwapEvm` and `signApprovalEvm` call it directly without going through `validateIntent`, so that guard is their only gate.

Beyond blocking the invalid transaction, a transient failure is now retried on each validation cycle instead of being frozen: the zero used to stick to the transaction and disable re-estimation for good.

The `BigNumber(0)` fallback itself is kept — it is what lets `validateIntent` raise `FeeNotLoaded` and block the flow.

**Max fee headroom (LIVE-32650)**

`EIP1559_BASE_FEE_MULTIPLIER` goes from 1.27 to 1.6. The base fee grows by at most 12.5% per block, so 1.27 only covered ~2 blocks (~24s), less than a device signing round trip, while the RPC path already used 2.

The amount paid is unchanged (the network charges base fee + priority fee, the max fee is a ceiling), but `validateIntent` compares `maxFeePerGas × gasLimit` to the balance, so the balance required to send goes up: at 30 gwei / 21000 gas, 0.000821 → 0.001029 ETH. Users between those two values will now hit `NotEnoughGas`.

Unlike the first fix, this one has no Datadog occurrence backing it, hence the separate commit.

**Regression tests** — `craftTransaction.test.ts`: a zero gas limit passed as a custom fee triggers a real estimation, and crafting is refused when the estimation yields zero.

Screenshot 2026-08-12 at 13 55 30

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-32644](https://ledgerhq.atlassian.net/browse/LIVE-32644), [LIVE-32650](https://ledgerhq.atlassian.net/browse/LIVE-32650)

[LIVE-32644]: https://ledgerhq.atlassian.net/browse/LIVE-32644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shared/env/src/definitions/team-coin-integration/index.ts
+++ b/shared/env/src/definitions/team-coin-integration/index.ts
@@ -178,7 +178,7 @@ const teamCoinIntegration = {
     desc: "minimum priority fee percents allowed compared to network conditions allowed when EIP1559_MINIMUM_FEES_GATE is activated",
   },
   EIP1559_BASE_FEE_MULTIPLIER: {
-    def: 1.27,
+    def: 1.6,
     parser: floatParser,
     desc: "multiplier for the base fee that is composing the maxFeePerGas property",
   },
--- a/.changeset/evm-gas-limit-and-base-fee-headroom.md
+++ b/.changeset/evm-gas-limit-and-base-fee-headroom.md
@@ -1,0 +1,10 @@
+---
+"@ledgerhq/coin-evm": patch
+"@shared/env": patch
+---
+
+Fix EVM transactions being signed with a zero gas limit, and widen the EIP-1559 max fee headroom.
+
+When gas estimation failed, its `BigNumber(0)` fallback travelled back to the sign step, where it was read as a deliberate custom gas limit. That disabled re-estimation and produced a transaction the node rejected with `intrinsic gas too low`. A non-positive gas limit is no longer honoured as a custom value, so the estimation runs again, and crafting now fails rather than sending a zero gas limit to the device (LIVE-32644).
+
+`EIP1559_BASE_FEE_MULTIPLIER` goes from 1.27 to 1.6, so an estimated transaction stays includable for 4 blocks instead of 2 (the base fee grows by at most 12.5% per block). Max fees displayed on chains using the Ledger gas tracker will be higher, but the amount actually paid is unchanged: EIP-1559 charges the base fee plus the priority fee, and the max fee is only a ceiling (LIVE-32650).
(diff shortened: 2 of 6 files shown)


YazhuEth:
**E2E** EVM with broadcast
-D : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/9412c688-0b97-497a-89ec-2ef2e52ab736/
-M : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/bab0c5ab-5d5f-4104-8010-329710e47fa2/

shared/env/src/definitions/team-coin-integration/index.ts
```diff
@@ -178,7 +178,7 @@ const teamCoinIntegration = {
     desc: "minimum priority fee percents allowed compared to network conditions allowed when EIP1559_MINIMUM_FEES_GATE is activated",
   },
   EIP1559_BASE_FEE_MULTIPLIER: {
-    def: 1.27,
+    def: 1.6,
```
henri-ly, line 181: sure about this ?

henri-ly (approved):

shared/env/src/definitions/team-coin-integration/index.ts
YazhuEth: https://ethereum.org/developers/docs/gas/#:~:text=The%20base%20fee%20will%20increase%20or%20decrease%20by%20a%20maximum%20of%2012.5%25%20per%20block
so 1.27 only covered ~2 blocks (~24s), if we increase to 1.6 we double it to 4 blocks (48sec) I think it will be more secure.

PR #20723, opened 2026-08-12
[LWD] fix(receive): guard confirm address against concurrent calls

YazhuEth:
### 📝 Description

**Previous behaviour**: on the receive flow, an address verification error was displayed even though the confirmation on the device completed normally. Reproduced consistently on Polkadot.

**Root cause**: in `StepReceiveFunds.tsx`, the auto-trigger effect

```ts
useEffect(() => {
  if (isAddressVerified === null) confirmAddress();
}, [specific?.useCustomConfirmAddress, isAddressVerified, confirmAddress]);
```

depends on `confirmAddress`, which is recreated on every render (its deps include `device`, `mainAccount` and `onChangeAddressVerified` — the latter being a non-memoized function in `modals/Receive/index.tsx`). Since `isAddressVerified` stays `null` for the whole duration of the device confirmation, every re-render started a **concurrent** `bridge.receive(..., { verify: true })` call. `withDevice` does not serialize device access, so the second call tore down the in-flight APDU exchange, and its `catch` overwrote the success with an error. Polkadot was the loudest reporter because its address confirmation is a long, multi-screen device flow.

**Fix**:

- `steps/StepReceiveFunds.tsx`: an `isConfirmingRef` latch makes `confirmAddress` return early when a verification is already running, and releases it in a `finally`.
- `modals/Receive/index.tsx`: `handleChangeAddressVerified`, `setIsAddressVerified`, `setVerifyAddressError` and `handleReset` are now `useCallback`-stable, removing the main source of re-triggering.

**Regression check**: `StepReceiveFunds.test.tsx` asserts that `bridge.receive` is called only once despite repeated re-renders with fresh callback/device identities while the verification is pending, and that a completed verification reports success exactly once. Removing the latch makes the first test fail with 4 calls instead of 1.

**Known related issue, out of scope**: `families/concordium/StepReceiveFunds/index.tsx` duplicates the same unguarded effect and has the same bug.

### 🧪 QA focus

- Receive on a Polkadot account with a real device: confirm the address → no error.
- User rejection on the device (`UserRefusedAddress`) → error screen + retry still work.
- Non-regression on the other receive paths (BTC/EVM, token accounts, staking step after receive).

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35825](https://ledgerhq.atlassian.net/browse/LIVE-35825)
- Reported on Slack: https://ledger.slack.com/archives/C02KVKX8GCF/p1786524984729949

[LIVE-35825]: https://ledgerhq.atlassian.net/browse/LIVE-35825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/receive-confirm-address-in-flight-guard.md
+++ b/.changeset/receive-confirm-address-in-flight-guard.md
@@ -1,0 +1,5 @@
+---
+"ledger-live-desktop": patch
+---
+
+Fix the receive flow reporting a verification error (visible on Polkadot) while the address confirmation succeeded on the device: re-renders no longer start a concurrent `confirmAddress` call, and a late result can no longer overwrite a successful verification.
--- a/apps/ledger-live-desktop/src/renderer/modals/Receive/index.tsx
+++ b/apps/ledger-live-desktop/src/renderer/modals/Receive/index.tsx
@@ -74,27 +74,34 @@ const ReceiveModal = (props: GlobalModalData["MODAL_RECEIVE"]) => {
     setState(prevState => ({ ...prevState, stepId: newStepId }));
   }, []);
 
-  const setIsAddressVerified = (newIsAddressVerified: State["isAddressVerified"]) => {
+  const setIsAddressVerified = useCallback((newIsAddressVerified: State["isAddressVerified"]) => {
     setState(prevState => ({ ...prevState, isAddressVerified: newIsAddressVerified }));
-  };
+  }, []);
 
-  const setVerifyAddressError = (newVerifyAddressError: State["verifyAddressError"]) => {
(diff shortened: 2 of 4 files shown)


henri-ly (approved):

YazhuEth:
**E2E**
- D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/e83f4a47-9ca2-4e39-a82d-98d40395f315/
- M: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/ac3d0a30-1e49-4b39-9518-1e2a5684a827/

PR #20728, opened 2026-08-12
[LWM] fix(mobile): prevent transient contact name errors (LIVE-35827)

deepyjr:
#### Previous behaviour

After creating a valid contact, the drawer could briefly display a duplicate-name error because the new contact reached the store before the drawer closed.

#### Fix

Do not expose validation errors while the already-valid name is being saved. The name input is read-only during this short saving state, and the ViewModel also ignores draft updates while saving. Duplicate names remain blocked before submission.

#### Regression coverage

Added coverage for the store update during a pending creation, attempts to update the draft while saving, and the read-only state on Web and Native.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35827](https://ledgerhq.atlassian.net/browse/LIVE-35827)
- **ADR** (if any): N/A

[LIVE-35827]: https://ledgerhq.atlassian.net/browse/LIVE-35827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/flow/flow-contacts-add-contact/src/components/ContactsContactNameDrawer/ContactsContactNameDrawerContent.native.tsx
+++ b/features/flow/flow-contacts-add-contact/src/components/ContactsContactNameDrawer/ContactsContactNameDrawerContent.native.tsx
@@ -25,7 +25,7 @@ export type ContactsContactNameDrawerContentProps = Readonly<{
   invalidNameError: ContactNameValidationErrorName | null;
   bottomInset?: number;
   keyboardInset?: number;
-  isInputEditable?: boolean;
+  isEditable?: boolean;
   labels: ContactsContactNameDrawerLabels;
   confirmLabel: string;
   confirmTestID?: string;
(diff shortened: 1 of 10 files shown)


features/flow/flow-contacts-add-contact/src/components/ContactsContactNameDrawer/ContactsContactNameDrawerContent.native.tsx
```diff
@@ -25,6 +25,7 @@ export type ContactsContactNameDrawerContentProps = Readonly<{
   invalidNameError: ContactNameValidationErrorName | null;
   bottomInset?: number;
   keyboardInset?: number;
+  isInputEditable?: boolean;
```
mcayuelas-ledger: why don't keep consistency across params between web and native?  `isInputEditable` vs `isEditable`?

mcayuelas-ledger (approved):

PR #20739, opened 2026-08-12
feat(aggregated-assets): validate and brand the response (LIVE-35232)

LucasWerey:
### 📝 Description

**Problem.** Nothing checked the DADA payload at runtime. `RawApiResponse` simply *asserted* that `cryptoAssets` was `Record<string, CryptoAssetMeta>` and `interestRates` was `Record<string, InterestRate>`, and `transformAssetsResponse` spread both straight through — so malformed data reached every consumer wearing a type it had never been checked against. Separately, `currencyId` and `fetchAt` were plain `z.string()`, so nothing distinguished a currency id from a timestamp at a call boundary.

**Solution.** Validate per entry at the api boundary, and brand the two interest-rate fields.

| Collection | Validated against | Invalid entry |
| --- | --- | --- |
| `cryptoAssets` | `CryptoAssetMetaSchema` | dropped |
| `interestRates` | `InterestRateSchema` | dropped |
| `networks` | api wire schema | dropped |
| `currenciesOrder` | api wire schema | falls back to no server ordering |

**Never a whole-response `parse`.** One unmodelled asset would otherwise blank out Market, Portfolio and the asset selector at once. Drops are counted per collection and reported in a single `console.warn` per response — aggregated rather than per item so a bad page cannot flood the log, and `warn` not `error` because this is expected and recoverable (`console.error` reaches Datadog as an error event).

### 🔍 Three decisions that needed evidence, not assumption

**1. Every schema is `loose()`.** I probed zod 4.3.6 before writing anything: `z.object()` **strips** unknown keys by default. Validating with the entity schemas as-is would have silently deleted any field DADA adds between the wire and the consumer — turning validation into data loss. `.loose()` checks the fields we rely on and passes the rest through. A test pins it.

**2. `fetchAt` never costs a rate.** The ticket claimed zero production reads; I verified it — every occurrence in the repo is a mock or handler *assigning* it, never reading. So it is branded **and optional**: DADA omitting a timestamp, or sending a date without a time, must not discard an APY the UI does display. The boundary coerces a bad value instead of dropping the rate.

**3. `assetsIds` stays unbranded — measured, not assumed.** I implemented the value branding and typechecked it:

- **0** production errors — every caller only *reads*, and a branded string is already assignable to `string`
- **23** fixture errors across `domain/api`, `features/platform` and `libs/asset-aggregation`
- **31** further `assetsIds:` literal sites in the two apps (103 repo-wide)

Entirely fixture cost, no production gain today. Reverted, with the numbers recorded next to the field so it reads as a decision rather than an omission. This is the *"materially harder, budget for it separately"* case the ticket flagged — it deserves its own change.

### 🏛️ Where the id union lives

`currencyId` cannot be `CryptoCurrencyIdSchema` alone: DADA keys rates by both `ethereum` and `ethereum/erc20/usd_tether`, so that would reject every token id.

The union goes in **`@domain/entity-currency`**, which already owns the crypto-or-token concept, rather than as a local copy inside interest-rate. Note `@domain/entity-contact` already declares its own `ContactCurrencyIdSchema` with the identical shape — a local copy here would have made it the second. Contact can adopt the shared one whenever its owners like; not touched here.

### 🧪 Tests

**All 168 pre-existing `domain/api` tests pass unmodified**, including the LIVE-35224 characterization suite — the hard constraint on this ticket.

Two of the interest-rate entity's **own** schema tests did change, because this PR inverts what they pinned: *"throws when a required field is missing"* (which listed `fetchAt`) and *"does not validate the fetchAt format"*. They now assert the new contract — a rate with no `fetchAt` is accepted, and `"2026-07-31"` / `"2026-07-31T00:00:00"` are rejected for lacking an offset. These are the entity's schema tests, not characterization tests.

New coverage: valid response untouched, one invalid item among valid siblings, an entirely invalid collection, unmodelled fields preserved, `currenciesOrder` fallback, per-collection drop counts, `fetchAt` tolerated in three broken shapes, and the drop path **through the chunked endpoint** (which merges pages and so needed its own case).

### ✅ Verification

- `@domain/api-aggregated-assets` **181 tests** (from 168) · `@domain/entity-interest-rate` **27** (from 23) · `@domain/entity-currency` 7 · `@domain/entity-aggregated-asset` 5 · `@features/platform-aggregated-assets` 129 · `libs/asset-aggregation` 74 · live-common `portfolio`/`modularDrawer`/`ModularDrawer`/`market` **297**
- Every touched package: `tsc` **exit 0, 0 own-source errors**
- **Desktop and mobile typecheck ✅ All Good** — branding an entity rippled into zero app files
- `enforce-boundaries` 0 · `enforce-package-structure:test` 0 · structure validator ok · knip 0 on all three touched packages · `oxfmt --check` clean
- `pnpm install --frozen-lockfile` succeeds; lockfile diff is **12 insertions, 0 deletions** — exactly the four new workspace links

> [!NOTE]
> Plain `pnpm install` wanted to collapse a `metro@0.83.3(metro-transform-worker@0.83.3)` peer variant, 49 unrelated deletions in the React Native bundler graph. The lockfile here carries only the four dependency entries; the churn was reverted and `--frozen-lockfile` re-verified.

**Pre-existing and unrelated**, confirmed to contain zero branding errors: `libs/asset-detail` `tsc` reports 105 errors, all in unbuilt sibling libs and none in its own source; `libs/ledger-live-common` reports 5 `TS2307` on `@ledgerhq/live-common/families/evm/types`, because its build output is absent (`lib-es/families/evm/types.d.ts` does not exist).

### 📖 Reviewing this

Three commits. The first two are one change split for readability and are coupled — the boundary coercion in commit 1 only makes sense with the brand from commit 2:

1. `validate the response per item` — the validation module and its wiring into both response paths
2. `brand currencyId and fetchAt` — the entity change, plus the measured `assetsIds` decision
3. `add changeset`

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35232](https://ledgerhq.atlassian.net/browse/LIVE-35232) (epic: [LIVE-35223](https://ledgerhq.atlassian.net/browse/LIVE-35223))
- **ADR**: [DADA DDD compliant](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/7382761502/DADA+DDD+compliant)
- **Not in this PR, still on the ticket**: the two layer-ownership fixes folded into LIVE-35232 (moving the DADA-agnostic `errors.ts` to `@shared/api-services`, and taking the RTK hooks out of `domain/api`). They touch both apps, so they land separately rather than entangling a Market/Portfolio regression with validation changes.
- **Unblocks**: LIVE-35233 — `cryptoOrTokenCurrencies`, the leniency-preserving one

[LIVE-35232]: https://ledgerhq.atlassian.net/browse/LIVE-35232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/domain/entity/aggregated-asset/src/schema.ts
+++ b/domain/entity/aggregated-asset/src/schema.ts
@@ -13,6 +13,12 @@ export const CryptoAssetMetaSchema = z.object({
   ticker: z.string(),
   /** Asset display name */
   name: z.string(),
-  /** Map of network IDs to their corresponding asset IDs */
+  /**
+   * Network id to the currency id representing this asset there.
+   *
+   * Ids stay unbranded for now: branding the values costs 23 fixture sites across three packages
+   * and 31 more in the apps, for no production change, since every caller only reads them and a
+   * branded string is already assignable to string. Worth doing on its own, not inside another task.
+   */
   assetsIds: z.record(z.string(), z.string()),
 });
--- a/.changeset/quiet-rates-validated.md
+++ b/.changeset/quiet-rates-validated.md
@@ -1,0 +1,7 @@
+---
+"@domain/api-aggregated-assets": patch
+"@domain/entity-interest-rate": patch
+"@domain/entity-currency": patch
+---
+
+Validate the DADA response per item, dropping malformed entries instead of trusting the payload, and brand an interest rate's currencyId and fetchAt so an id or timestamp cannot be confused with a plain string
--- a/domain/api/aggregated-assets/package.json
+++ b/domain/api/aggregated-assets/package.json
@@ -40,7 +40,9 @@
     "@domain/entity-interest-rate": "workspace:*",
     "@reduxjs/toolkit": "catalog:",
     "@shared/api-services": "workspace:*",
-    "@shared/env": "workspace:*"
+    "@shared/env": "workspace:*",
+    "@shared/schema-primitives": "workspace:*",
+    "zod": "catalog:"
   },
   "peerDependencies": {
     "react": ">=19.0.0",
--- a/domain/api/aggregated-assets/src/internals/requests.test.ts
+++ b/domain/api/aggregated-assets/src/internals/requests.test.ts
@@ -45,12 +45,19 @@ describe("resolveBaseUrl", () => {
 
 describe("fetchAssetsPage", () => {
   let baseQuery: jest.Mock;
+  let warn: jest.SpyInstance;
 
   const request = () =>
     baseQuery.mock.calls[0][0] as { url: string; params: Record<string, unknown> };
 
   beforeEach(() => {
     baseQuery = jest.fn().mockResolvedValue({ data: emptyRaw });
+    warn = jest.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  /* In afterEach, so a failing assertion cannot leak the spy into the next test. */
+  afterEach(() => {
+    warn.mockRestore();
   });
 
   it("targets the /assets path on the resolved base url", async () => {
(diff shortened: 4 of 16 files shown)


domain/entity/aggregated-asset/src/schema.ts
```diff
@@ -13,6 +13,12 @@ export const CryptoAssetMetaSchema = z.object({
...
   /** Asset display name */
   name: z.string(),
-  /** Map of network IDs to their corresponding asset IDs */
+  /**
+   * Network id to the currency id representing this asset there.
+   *
+   * Ids stay unbranded for now: branding the values costs 23 fixture sites across three packages
+   * and 31 more in the apps, for no production change, since every caller only reads them and a
+   * branded string is already assignable to string. Worth doing on its own, not inside another task.
+   */
```
LL782, line 22: This feels like messaging between you and the agent within the scope of a task. I don't think it should persist as a comment within a codebase. Personally, I'm often leaving the same comment for the same people, I'd appreciate it if you could look out for this in your own reviews, or even better, earlier.

LL782 (commented):
@LucasWerey looks relatively sane to me but I think it would be good to:

1. Get a review from your team (assuming they have context)
2. Add your team as CODEOWNERS for the future

PR #20743, opened 2026-08-13
[LWM] fix(e2e): fix scroll targeting and silent failures in LWM e2e

VicAlbr:
### 📝 Description

Part 8 (final) of the QAA-1442 work on Ledger Wallet Mobile E2E flakiness. An earlier audit annotated 37 defects in `e2e/mobile`; this PR closes the ones that remained after PR #20680, plus 2 same-class defects the ticket had missed.

**Previous behaviour.** Assertions asked for things the geometry could not satisfy, and failures were silent. `toBeVisible()` requires 75% of an element's *own* area on screen, so a section wrapper taller than the viewport can never pass — the suite worked around that by lowering the gate to 35%, by scrolling with unexplained 650–2000px steps, and by asserting existence instead of visibility. Ten scroll call sites named no container at all, letting the engine guess `by.type("android.widget.ScrollView" | "RCTEnhancedScrollView").atIndex(0)`, so a failure read "target not found" rather than "you scrolled the wrong list". `scrollOnce` swallowed every scroll error without logging, `isIdVisible` reported "not visible" whenever an id matched several rows, and `countElementsById` counted only the rows that fit the screen.

**The fix**, per defect class:

- **Visibility vs geometry** — `expectOperationHistoryVisible` targets a leaf `operation-row-*` inside the account's history section instead of the viewport-tall `SectionList` wrapper, and asserts at the default threshold. The `visibilityPercentage` parameter is deleted through `scrollToId` → `performScroll` → `isVisible`, so no site can lower the gate again.
- **Omitted containers** — 8 of 10 sites now name their container. The portfolio has exactly one scrollable, so its `undefined`/omitted containers collapse to the list the sibling branch already named; `goToSpecificAsset`'s flag branch collapses to a single line.
- **Magic pixel steps** — 650/700/1000/1200/2000 replaced with the default. `tapCryptosSectionTitle` scrolled to an *unrelated* element to position the list and then tapped a different one without asserting reachability; it now mirrors its stablecoins sibling.
- **Silent failures** — `scrollOnce` logs the error it used to swallow (via Detox's `log.warn`, so it lands in the shard log) and its `"bottom"` fallback no longer swipes the opposite way. `isIdVisible`/`isIdPresent` are index-safe.
- **Existence vs visibility** — `countElementsById` replaced with `countElements(getElementsById(...))` where the assertion is about what exists; `checkVisibility: false` dropped where the element is a full-screen root or a leaf row.
- **App side** — the accounts list scrollable carried `accounts-list-${accounts.length}`, which no test could name; it is now a stable `accounts-list`, consumed by `account.page.ts` in this same change.

**Regression prevention.** The two acceptance gates are now structural rather than conventions: `grep visibilityPercentage e2e/mobile` returns nothing, and the count of scroll sites without a container is 1 (down from 10) — see Known limitations.

Net **+79/−119**.

### 🔍 Verification

Local device runs (`E2E_RETRIES=0`), both platforms where relevant:

| Spec | iOS | Android |
|---|---|---|
| `assetAggregationMarketDetail` | ✅ 4/4 | ✅ 4/4 |
| `marketBanner` | ✅ 1/1 | ✅ 1/1 |
| `portfolioChartAndAssets` | — | ✅ 1/1 |
| `walletAssets` | — | ✅ 5/5 |
| `wallet40Q2/portfolio` | — | ✅ 7/7 |
| `deposit` (2 specs) | — | ✅ 2/2 |
| `sendInvalidTokenAmount` (USDT, BUSD) | ✅ 2/2 | — |

`assetAggregationMarketDetail` on iOS was **failing on develop** and is the ticket's headline acceptance criterion. The failing assertion — `[DAI (Polygon)] - Aggregated assets show one row per asset and list holding addresses` — went from a 133–186s failure to **29.4s passing**. Diagnosed from the captured native hierarchy: the Ethereum address row's name element is 20px tall and must clear a 114px sticky `asset-detail-ctas` footer, leaving a 624px-wide band of valid scroll offsets — which a 700px step can jump over and the default 300px step cannot.

CI dispatched over the full suite on both platforms (justified: this changes `performScroll`/`scrollOnce`, behind all 87 scroll call sites, plus `scrollToId`'s signature and two probes with 36 callers). Results judged per-test against the same-day `develop` nightly rather than against zero, since that nightly itself had 3 iOS / 10 Android failures.

### ⚠️ Known limitations

- **`send.page.ts:250` still names no container**, so the `by.type(...)` fallback remains reachable from that one site. The fee screen has three different UIs, each rendering a `FlatList` with no testID and inconsistent (or absent) row testIDs. An earlier attempt to name the inner `FlatList` regressed 2 tests — Detox cannot scroll a container that is itself only 76% visible — so this needs app-side work: a testID on the *outer* fee scroller, or per-row ids.
- **`selectLanguage` still locates by text**, which is locale-dependent and is the value under test. The container is now named, but the language rows have no per-row testID.
- **Wallet 4.0 Q1 paths are untouched by design** (ticket items 3 and 12, descoped). They execute only under a manual `wallet40-q1` dispatch that no schedule or caller passes, so their assertions cannot be verified by CI.
- The 2 `assetDiscoverability` stocks failures are **pre-existing on develop** — confirmed by reverting `e2e/mobile` and reproducing them identically. They fail in `scrollToStocksHeader`, which this PR does not touch.

### 🧪 Test coverage

This PR *is* test code. Verified by device runs on both platforms plus a full-suite CI dispatch; no unit tests apply.

### 👀 QA focus

- **Scroll-heavy flows on iOS**, where occlusion counts toward visibility: asset detail addresses, account operation history, portfolio asset sections.
- **The three quick-action `press*` paths** (deposit, receive, buySell) — an earlier iteration added a wait before those taps and shifted the bottom-sheet timing; that was reverted, so these should behave exactly as on develop.
- **Android horizontal scrolling** (`marketBanner`), the only horizontal call site.
- **Shard timings** vs the develop nightly — the QAA-1442 gate is "no shard slower than the median + 1 min".

### 🔗 Context

- **JIRA / GitHub issue**: [QAA-1487](https://ledgerhq.atlassian.net/browse/QAA-1487) (epic [QAA-1442](https://ledgerhq.atlassian.net/browse/QAA-1442), follows [QAA-1458](https://ledgerhq.atlassian.net/browse/QAA-1458))
- **Learnings page**: [LWM e2e scroll and locator feedback](https://ledgerhq.atlassian.net/wiki/spaces/QA/pages/7414906965/LWM+e2e+scroll+an+locator+feedback)
- **Related PR**: #20680 (Parts 5+6) — introduces `revealForTap`, which the two deferred sites above will want

[QAA-1487]: https://ledgerhq.atlassian.net/browse/QAA-1487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/lwm-e2e-scroll-targeting.md
+++ b/.changeset/lwm-e2e-scroll-targeting.md
@@ -1,0 +1,14 @@
+---
+"ledger-live-mobile-e2e-tests": patch
+"live-mobile": patch
+---
+
+Fix scroll targeting and silent failures in the Ledger Wallet Mobile E2E suite: target leaf rows
+instead of viewport-tall wrappers so assertions reach the default 75% visibility honestly, delete the
+`visibilityPercentage` parameter so no site can lower the gate, name the scroll container at every
+call site that used to let the engine guess one, replace the unexplained pixel steps with the default,
+count rows by existence rather than by what fits the screen, log the scroll errors `scrollOnce`
+used to swallow and correct its `"bottom"` fallback direction, make the `isIdVisible`/`isIdPresent`
+probes index-safe so a shared id stops being reported as invisible, and assert visibility where the
+suite previously only proved an element existed in the tree. On the app side, the accounts list
+scrollable now carries a stable `accounts-list` testID instead of one keyed on the account count.
--- a/apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
+++ b/apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
@@ -241,7 +241,6 @@ export default function SelectFeesStrategy({
         ) : null}
         <View style={styles.strategiesContainer}>
           <FlatList
-            testID="fees-strategy-list"
             data={strategies}
             renderItem={renderItem}
             keyExtractor={s => s.label}
--- a/apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
+++ b/apps/ledger-live-mobile/src/components/SelectFeesStrategy.tsx
@@ -241,6 +241,7 @@ export default function SelectFeesStrategy({
         ) : null}
         <View style={styles.strategiesContainer}>
           <FlatList
+            testID="fees-strategy-list"
             data={strategies}
             renderItem={renderItem}
             keyExtractor={s => s.label}
--- a/apps/ledger-live-mobile/src/families/evm/SelectFeesStrategy.tsx
+++ b/apps/ledger-live-mobile/src/families/evm/SelectFeesStrategy.tsx
@@ -249,6 +249,7 @@ export default function SelectFeesStrategy({
         )}
         <View style={styles.strategiesContainer}>
           <FlatList
+            testID="fees-strategy-list"
             data={customFees ? strategiesWithCustom : strategies}
             renderItem={renderItem}
             keyExtractor={s => s}
--- a/apps/ledger-live-mobile/src/families/evm/SelectFeesStrategy.tsx
+++ b/apps/ledger-live-mobile/src/families/evm/SelectFeesStrategy.tsx
@@ -249,7 +249,6 @@ export default function SelectFeesStrategy({
         )}
         <View style={styles.strategiesContainer}>
           <FlatList
-            testID="fees-strategy-list"
             data={customFees ? strategiesWithCustom : strategies}
             renderItem={renderItem}
             keyExtractor={s => s}
--- a/apps/ledger-live-mobile/src/screens/Accounts/index.tsx
+++ b/apps/ledger-live-mobile/src/screens/Accounts/index.tsx
@@ -126,7 +126,7 @@ function Accounts({ navigation, route }: NavigationProps) {
           </Flex>
         ) : (
           <List
-            testID={`accounts-list-${flattenedAccounts.length}`}
+            testID="accounts-list"
             data={flattenedAccounts}
             renderItem={renderItem}
             keyExtractor={(i: AccountLike) => i.id}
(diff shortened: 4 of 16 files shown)


VicAlbr:
## 🧪 Triggered test workflows

Manually dispatched on `test/e2e-QAA-1487-scroll-locator-fixes`, scoped to the **full mobile suite**:

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | full suite (no `test_filter`), iOS & Android, nanoX | [run 31683070415](https://github.com/LedgerHQ/ledger-live/actions/runs/31683070415) |

Dispatched on `66991e55d20`; head is now `ac097e6a452`, which only adds the changeset, so the run remains representative of the test changes.

Scope rationale: this PR changes `pageScroller.performScroll`/`scrollOnce` (behind all 87 scroll call sites), `scrollToId`'s signature, and `isIdVisible`/`isIdPresent` (36 callers) — no narrower filter covers it. Desktop E2E and Coin Tester are skipped: no desktop or coin-module code is touched.

Results are judged **per-test against the same-day `develop` nightly**, not against zero — that nightly ([run 31653938497](https://github.com/LedgerHQ/ledger-live/actions/runs/31653938497)) itself had 3 iOS / 10 Android failures, and the failure sets churn heavily between runs (Speculos/WebView flake).

Run history on this branch:

| Run | Head | Outcome |
|---|---|---|
| [31639107221](https://github.com/LedgerHQ/ledger-live/actions/runs/31639107221) | `c16f6a6` | iOS 0/3 shards, Android 10/12 — surfaced a real regression (fee list container, since reverted) |
| [31678276179](https://github.com/LedgerHQ/ledger-live/actions/runs/31678276179) | `f52b776` | iOS 2/3, Android 11/12 — fee fix confirmed; surfaced a second (quick-action taps, since reverted) |
| [31683070415](https://github.com/LedgerHQ/ledger-live/actions/runs/31683070415) | `66991e5` | in progress |

cunhabruno (approved):

LucasWerey (approved):

PR #20745, opened 2026-08-13
[LWDM] fix(contacts): allow reopening delete contact confirmation

claudiiafg:
### 📝 Description

- Fixes a mobile bug where the delete contact confirmation sheet would not reopen after the user cancelled it and tapped Delete again from the contact detail actions menu.
- Defers opening the delete confirmation until the actions menu bottom sheet has fully dismissed, using `onModalHide` and a pending-delete flag in the mobile adapter, so `QueuedBottomSheet` no longer conflicts when transitioning between sheets.
- Keeps desktop behavior unchanged by calling `openDelete()` immediately after `onDeletePress()` in the desktop adapter, since it does not use the queued sheet pattern.

https://github.com/user-attachments/assets/bc97a5bd-fb32-4988-b72e-f6c6e5f545d8

### 🔗 Context

[LIVE-35790](https://ledgerhq.atlassian.net/browse/LIVE-35790)

[LIVE-35790]: https://ledgerhq.atlassian.net/browse/LIVE-35790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/features/Contacts/screens/ContactDetail/components/ContactDetailEditDeleteSheets/index.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/features/Contacts/screens/ContactDetail/components/ContactDetailEditDeleteSheets/index.tsx
@@ -28,7 +28,11 @@ export function ContactDetailEditDeleteSheets({
   const keyboardInset = shouldUseKeyboardAvoidance(Platform.OS, Platform.Version)
     ? keyboardHeight
     : 0;
-  const { onClose: onCloseActionsMenuFromMenu, ...actionsMenuProps } = actionsMenu;
+  const {
+    onClose: onCloseActionsMenuFromMenu,
+    onHidden: onActionsMenuHidden,
+    ...actionsMenuProps
+  } = actionsMenu;
   const onCloseActionsMenu = useCallback(() => {
     onCloseActionsMenuFromMenu();
   }, [onCloseActionsMenuFromMenu]);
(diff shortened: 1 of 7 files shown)


apps/ledger-live-mobile/src/mvvm/features/Contacts/screens/ContactDetail/components/ContactDetailEditDeleteSheets/index.tsx
```diff
@@ -66,6 +66,7 @@ export function ContactDetailEditDeleteSheets({
       </QueuedBottomSheet>
       <QueuedBottomSheet
         isRequestingToBeOpened={deleteDrawer.isOpen}
+        isForcingToBeOpened={deleteDrawer.isOpen}
```
LucasWerey: Hmm, are we sure it's the right fix? Because this is bypassing the queue I think, something might be wrong somewhere else

claudiiafg: you're right that this bypasses the queue

instead I changed the solution to this:
- onDeletePress → close menu + set pendingDelete = true
- actions menu onModalHide / onClose → if pendingDelete, call openDelete() and clear the flag

LucasWerey (dismissed):

LucasWerey (approved):

PR #20754, opened 2026-08-13
fix(evm): drop superseded pending token operation (LIVE-35844)

YazhuEth:
### 📝 Description

Speeding up or cancelling an ERC-20 send left the superseded operation stuck as `Sending…` forever, and kept its amount locked out of the token spendable balance.

**Why.** A replacement transaction reuses the nonce it replaces and carries a different hash, so the superseded pending operation can only be retired by comparing nonces, never by hash. But for a token send, `listOperations` drops the fees-only native operation to avoid a duplicate row, and that dropped operation was the only one carrying the nonce. The parent `FEES` operation ended up with no `transactionSequenceNumber`, `postSync` fell back to `-1`, and every pending operation looked current on every sync.

Confirmed on a real account: a USDC cancel mined under nonce 316 while the superseded send stayed pending, and the stored parent operation reported no nonce although the explorer returns `nonce_value: 316`. Etherscan-backed chains were unaffected, that adapter already sets the nonce on token operations.

**Fix.**

- `coin-evm/logic/listOperations.ts`: token operations inherit the parent nonce, next to the fees and senders they already inherited. In `enrichTokenWithParent`, so ERC-20/721/1155 and both explorers are covered at once.
- `ledger-wallet-framework/bridge/jsHelpers.ts`: `sameOp` compares `transactionSequenceNumber`, so an operation stored without a nonce is corrected on refetch instead of keeping the gap forever. The check is deliberately asymmetrical (`a` is the stored operation, `b` the fetched one): a fetch reporting no nonce must never erase a stored one, which `enrichTokenWithParent` can produce when pagination splits a transaction's native and token operations across pages.
- `generic-coin-framework/postSync.ts`: the comparison baseline skips operations reporting no nonce instead of collapsing to `-1`, and the optimistic retention window (computed in `jsHelpers`, but discarded here) now applies to token accounts as a backstop.
- `generic-coin-framework/utils.ts`: a confirmed nonce of `0` is no longer read as absent.

Covered by unit tests on all four: nonce inheritance, backfill on refetch, no erasure by a nonce-less fetch, retirement by a same-nonce replacement, the retention window on token accounts, and `0` as a valid baseline.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35844](https://ledgerhq.atlassian.net/browse/LIVE-35844)
- **ADR** (if any):

### DEMO 

https://github.com/user-attachments/assets/3165a5fc-3510-4745-b782-64be68e21966

The video demonstrates the cancellation of a USDC transfer on Ethereum mainnet. The first transaction is canceled by a self-transfer.

The pending transaction that we cancel disappears from the list of operations in the USDC Ethereum account, and the cancel operation appears in the Ethereum transaction list because no USDC is moved (self-transfer of 0 USDC) — only fees are paid in ETH for the transaction gas.

### 🧪 Notes for QA

- `sameOp` is generic: an operation whose nonce changes between syncs is now treated as updated, for every family. Worth a look at history and pending operations outside EVM.
- A slow but valid token send now leaves `Sending…` after `OPERATION_OPTIMISTIC_RETENTION` (30 min) and returns as confirmed once mined, same as native sends today.
- Already-stuck operations are only cleared by that window, since `minHeight` stops their parent operations from being refetched with the nonce.

[LIVE-35844]: https://ledgerhq.atlassian.net/browse/LIVE-35844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ledger-live-common/src/bridge/generic-coin-framework/utils.ts
+++ b/libs/ledger-live-common/src/bridge/generic-coin-framework/utils.ts
@@ -601,9 +601,8 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
     senders: extra.parentSenders ?? op.senders,
     recipients: extra.parentRecipients ?? op.recipients,
     date: op.tx.date,
-    transactionSequenceNumber: op.details?.sequence
-      ? new BigNumber(op.details?.sequence.toString())
-      : undefined,
+    transactionSequenceNumber:
+      op.details?.sequence != null ? new BigNumber(op.details.sequence.toString()) : undefined,
     hasFailed,
     extra,
   };
--- a/libs/ledger-wallet-framework/package.json
+++ b/libs/ledger-wallet-framework/package.json
@@ -124,6 +124,7 @@
     "@types/imurmurhash": "^0.1.4",
     "@types/node": "catalog:",
     "@ledgerhq/live-dmk-speculos": "workspace:^",
+    "@ledgerhq/wallet-framework-test-setup": "workspace:^",
     "@types/invariant": "^2.2.2",
     "@types/jest": "catalog:",
     "@types/lodash": "catalog:",
--- a/.changeset/proud-pandas-vanish.md
+++ b/.changeset/proud-pandas-vanish.md
@@ -1,0 +1,9 @@
+---
+"@ledgerhq/live-common": patch
+"@ledgerhq/coin-evm": patch
+"@ledgerhq/ledger-wallet-framework": patch
+---
+
+Fix an ERC-20 operation staying stuck on "Sending..." after a speed up or a cancel, which also kept
+its amount locked out of the token spendable balance. A replaced transaction can only be retired by
+its nonce, and token operations were not carrying one.
--- a/libs/coin-modules/coin-evm/src/logic/listOperations.test.ts
+++ b/libs/coin-modules/coin-evm/src/logic/listOperations.test.ts
@@ -889,6 +889,58 @@ describe("listOperations", () => {
     expect(sameTxFees).toEqual([parentFee, parentFee]);
   });
 
+  it("inherits the parent nonce on token ops, whose fees-only native op is dropped", async () => {
+    // LIVE-35844: the Ledger explorer only reports the nonce on the native op
+    const testContext = ctxWithExplorer({ type: "ledger" });
+
+    const address = "0xb69b37a4fb4a18b3258f974ff6e9f529ad2647b1";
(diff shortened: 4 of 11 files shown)


YazhuEth:
**E2E**
-D : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/6e011cfb-6bfd-4519-b328-cc24ffb70763/
-M : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/c06a3802-2702-47e6-a10c-e1877381a000/

libs/ledger-live-common/src/bridge/generic-coin-framework/utils.ts
```diff
@@ -536,9 +536,8 @@ export function adaptCoreOperationToLiveOperation(accountId: string, op: CoreOpe
     senders: extra.parentSenders ?? op.senders,
     recipients: extra.parentRecipients ?? op.recipients,
     date: op.tx.date,
-    transactionSequenceNumber: op.details?.sequence
-      ? new BigNumber(op.details?.sequence.toString())
-      : undefined,
+    transactionSequenceNumber:
+      op.details?.sequence != null ? new BigNumber(op.details.sequence.toString()) : undefined,
```
francois-guerin-ledger, line 605:
```suggestion
    transactionSequenceNumber:
      typeof op.details?.sequence === "bigint" ? new BigNumber(op.details.sequence.toString()) : undefined,
```
Prefer a truthy ternary.

libs/ledger-wallet-framework/package.json
```diff
@@ -124,6 +124,7 @@
     "@types/imurmurhash": "^0.1.4",
     "@types/node": "catalog:",
     "@ledgerhq/live-dmk-speculos": "workspace:^",
+    "@ledgerhq/wallet-framework-test-setup": "workspace:^",
```
francois-guerin-ledger, line 127: Why do we have to add this .

francois-guerin-ledger (dismissed):

libs/ledger-wallet-framework/package.json
YazhuEth: CI is failing because of this import missing

libs/ledger-live-common/src/bridge/generic-coin-framework/utils.ts
YazhuEth: `transactionSequenceNumber: sequence == null ? undefined : new BigNumber(sequence.toString()),`
We can do this but i really need to use the `null`because a 0 need to give a True

henri-ly (approved):

PR #1743, opened 2022-11-02
Support/turbo ignore

valpinkman:
### 📝 Description

Orchestrate CI depending on affected packages

### ❓ Context

- **Impacted projects**: `tools`, `automation` 
- **Linked resource(s)**: https://github.com/LedgerHQ/ledger-live-issues/issues/57

### ✅ Checklist

- [ ] **Test coverage** 
- [x] **Atomic delivery** 
- [x] **No breaking changes** 

### 🚀 Expectations to reach

Better control on the trigger of our workflows

_Please make sure you follow these [**Important Steps**](https://github.com/LedgerHQ/ledger-live/blob/develop/CONTRIBUTING.md#important-steps)._

_Pull Requests must pass the CI and be internally validated in order to be merged._

PR #20756, opened 2026-08-13
[LWDM] feat(zcash): shielded receive block | 0x51 device confirm (LIVE-35014)

vladyslavchupovskiy-ext-art:
### 📝 Description

Adds the private (Orchard UA) address block to the Receive modal for Zcash accounts when the `zcashShielded` feature flag is enabled.

**How it works**

The Receive modal already shows the transparent (`t1…`) address via the standard `Receive1ShareAddress` block. This PR adds a second block below it — `ZcashShieldedReceiveBlock` — that shows the Orchard unified address (`u1…`) and drives on-device confirmation via a single `0x51` (`INS_GET_SHIELD_ADDR`) APDU call.

The `0x51` call takes the account's `freshAddressPath` (BIP-44 transparent path). The device derives the transparent address **and** the Orchard UA from the same account index, displays both on one review screen, and returns only the UA. This means one device interaction covers both addresses — no second `0x40` call is needed.

**Key implementation decisions**

- `ZcashShieldedReceiveBlock` is wired through the `StepReceiveFundsPostAlert` extension point in `families/bitcoin/`, which renders after the public address block.
- `ZcashReceiveDeviceAnimation` implements the `StepReceiveFundsDeviceAnimation` slot, replacing the standard animation with a combined-address prompt when a shielded address is present.
- `useCustomConfirmAddress` in `families/types.ts` gains a function form evaluated during render (hook-free, feature flags passed in). The bitcoin family predicate gates suppression on `privateInfo.shieldedAddress` — accounts without one continue through the standard `0x40` path.
- The UTXO privacy notice is suppressed from the public address block when `usesCustomConfirm` is true, and rendered after the private address block instead (via `UTXOAddressAlert` inside `ZcashShieldedReceiveBlock`).
- After `ZcashShieldedVerify` resolves, `transitionTo("receive")` is called directly — matching the behaviour of the normal confirmation flow and avoiding a stale-closure effect in the parent.
- MOCK mode is handled in `ZcashShieldedVerify` (auto-resolves after 1 s without hitting the device).
- i18n keys for the shielded flow live under `zcash.shielded.receive.*`, not the shared `currentAddress.*` namespace.

**Changes**

_LLD — new (bitcoin family):_
- `ZcashShieldedReceiveBlock.tsx` — private address block + `ZcashShieldedVerify` inner component + no-UFVK warning
- `ZcashReceiveDeviceAnimation.tsx` — `StepReceiveFundsDeviceAnimation` slot: combined animation for shielded accounts, falls back to standard
- `StepReceiveFundsPostAlert.tsx` — slot wiring `ZcashShieldedReceiveBlock` into the receive step

_LLD — modified:_
- `families/types.ts` — function form of `useCustomConfirmAddress`; `StepReceiveFundsDeviceAnimation` slot type
- `families/bitcoin/index.ts` — wires the three slots; `useCustomConfirmAddress` predicate (gated on `privateInfo.shieldedAddress` + feature flag)
- `StepReceiveFunds.tsx` — `receive-public-address-block` testid; `suppressUTXOAlert` (moves UTXO privacy notice after the private block when shielded is active); `device={device}` explicit prop narrows the null guard on the animation slot call site
- `Receive/index.tsx` — exposes `transitionTo` in the PostAlert slot props
- `app.json` — `zcash.shielded.receive.{verifyAddresses,privateFor}` keys

_coin-zcash:_
- `bridge/index.ts` — `getShieldedAddress` on `ZcashAccountBridge`
- `bridge/sync.ts` — shielded address persistence
- `signer/getShieldedAddress.ts` — signer method
- `types/signer.ts` — `ZcashSigner` interface addition

_live-signer-zcash:_
- `DmkSignerZcash.ts` — `getShieldedAddress` implementation + `getAppConfig` stub
- `types.ts` — `ZcashShieldedAddress` / `ZcashSigner` types

_Tests:_
- `bitcoinFamilyPredicates.test.ts` — 4 unit tests for the `useCustomConfirmAddress` predicate (all four branches: flag off, non-Zcash, no shielded address, shielded present)
- `ZcashShieldedReceiveBlock.test.tsx` — 11 unit tests (rendering, testnet `utest1…`, no-UFVK states, QR modal security, verification lifecycle, `transitionTo` advancement, wrong-device error, rejection passthrough)
- `ZcashReceiveDeviceAnimation.test.tsx` — 5 unit tests
- `StepReceiveFundsPostAlert.test.tsx` — slot wiring tests
- `StepReceiveFunds.test.tsx` — updated: generic slot tests replace Zcash-specific ones; `useCustomCom tests added
- `DmkSignerZcash.test.ts` — signer tests

---

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35014](https://ledgerhq.atlassian.net/browse/LIVE-35014)
- **Depends on**: #20615 (zcash shielded persistence — already merged to `develop`)

[LIVE-35014]: https://ledgerhq.atlassian.net/browse/LIVE-35014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

## Specification alignment — LIVE-35014

| # | Criterion | Verdict | Notes |
|---|-----------|---------|-------|
| 1 | Two labelled blocks, public then private, matching Figma | **partial** | Block order and `data-abel reuses shared `currentAddress.for` by design — no clean family-slot pattern for a label-onlyoverride without a concordium-style full step replacement; privacy notice moved after private block |
| 2 | Each block's QR/copy asserted by one integration test | **partial** | Private QR modal securityress matches block. Clipboard copy delegated to QA E2E |
| 3 | Public block keeps freshAddress, reuse alert, derivation-mode tag | **met** | Transparent path untouched; `display:contents` wrapper preserves layout for all currencies |
| 4 | Bitcoin/Litecoin/Dash unchanged; dash/taproot post-alerts still render | **partial** | Code paths preserved; no unit test asserts dash/taproot alerts still appear |
| 5 | No-UFVK account shows the CTA opening the export modal | **by design: removed** | No-UFVK shows warning only; transparent 0x40 proceeds in parent. Product decision: CTA would block receiving to the transparent address while
UFVK setup is pending |
| 6 | Exactly two component states, no third "address unknown" branch | **met** | Shielded address prent → warning |
| 7 | Flag off ⇒ step indistinguishable from today | **met** | `display:contents` on public block wrax artefact |
| 8 | Testnet account renders `utest1…` | **met** | Testnet fixture added in `ZcashShieldedReceiveBlock.test.tsx` |
| 9 | ~106-char address readable in block, QR modal and after copy at minimum width | **not asserted*s Speculos or manual viewport check |
| 10 | New i18n keys under `zcash.shielded.*`, no generic key repurposed | **met** | All active keys under `zcash.shielded.receive.*`; stale `zcash.receive.noUfvk` removed; `cta` key dropped by design (see #5) |
| 11 | Distinct stable `data-testid` on both blocks | **met** | `receive-public-address-block` and `rck` both present |
| 12 | Exactly one device exchange with the display flag — asserted by call count | **met** | Test asserts `getShieldedAddress` called exactly once with `display: true` and correct account |
| 13 | UA compared; mismatch ⇒ wrong-device-class error, both unverified | **met** | `WrongDeviceForAccount` from `@ledgerhq/ledger-wallet-framework/errors` — class and name asserted |
| 14 | Transparent path sent is `account.freshAddressPath` — asserted | **partial** | Test asserts correct account object passed; `freshAddressPath` usage is internal to the bridge |
| 15 | Zcash without UFVK verifies through unchanged transparent path; non-Zcash unchanged — both assomConfirmAddress` predicate tested across all four branches; standard animation renders for
non-shielded accounts |
| 16 | Rejection leaves both unverified, retry works, never partially successful | **met** | Rejectioh unchanged; retry flow inherited from parent |
| 17 | Disconnection / below-minimum-version / wrong-seed give distinct actionable errors | **not met** | `getAppConfig` stubbed in `DmkSignerZcash`; blocked on `GetAppConfigCommand` in `device-sdk-ts` (#1743). Minimum-version
gate tracked separately |
| 18 | `isAddressVerified` / `verifyAddressError` / retry / skip still work, incl. re-verify | **not  `Receive` body render; no unit test covers this |
| 19 | MOCK shortcut resolves verification without a device | **partial** | Implemented under `getEnvoverage with `/* istanbul ignore next */` |
| 20 | Staking-flow continuation and Receive analytics still fire for Zcash | **not asserted** | `onFent calls unchanged; no Zcash-specific assertion |
| 21 | `pnpm typecheck`, LLD unit tests, ≥80% coverage on new files | **met** | Typecheck clean; all unit tests pass; `ZcashShieldedReceiveBlock.tsx` 89.65% statements / 82.6% branch |


apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -398,8 +429,19 @@ const StepReceiveFunds = (props: StepProps) => {
...
+                    <Separator />
+                    <ZcashReceive2Device device={device} />
+                  </>
+                ) : null
+              ) : (
+                <>
+                  <Separator />
+                  <Receive2Device device={device} name={name} />
+                </>
+              )}
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: fixed

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -315,13 +329,30 @@ const StepReceiveFunds = (props: StepProps) => {
...
-  }, [specific?.useCustomConfirmAddress, isAddressVerified, confirmAddress]);
+  }, [specific?.useCustomConfirmAddress, isZcashShielded, isAddressVerified, confirmAddress]);
+
+  // After successful shielded verification, advance to the "receive" success
+  // step — mirrors what confirmAddress() does for the standard 0x40 path.
+  useEffect(() => {
+    if (isZcashShielded && isAddressVerified === true) {
+      transitionTo("receive");
+    }
+  }, [isZcashShielded, isAddressVerified, transitionTo]);
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: duplicate of new one fixed

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
```diff
@@ -0,0 +1,199 @@
...
+        onChangeAddressVerified(match, match ? null : new Error("Address mismatch"));
+      })
+      .catch(err => {
+        if (cancelled) return;
+        // Do not log err — IPC/device errors can echo UFVK back in the message
+        onChangeAddressVerified(
+          false,
+          err instanceof Error ? err : new Error("Verification failed"),
+        );
+      });
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Fixed

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -315,18 +329,52 @@ const StepReceiveFunds = (props: StepProps) => {
...
+        ?.shieldedAddress ?? null)
+    : null;
+
   // Families that own their confirm-address lifecycle opt out of the shared auto-trigger.
+  // Zcash with the shielded feature on never uses 0x40: verification is handled by
+  // ZcashShieldedVerify (0x51) when a shieldedAddress exists, or skipped entirely
+  // when showing the "Enable private balance" activation CTA.
   useEffect(() => {
-    if (specific?.useCustomConfirmAddress) return;
+    if (specific?.useCustomConfirmAddress || isZcashShielded) return;
```
cted-ledger: ### 🔴 Critical — transparent verification is removed for Zcash accounts without a shielded address (confidence: 100)

`isZcashShielded` is `currency.id === "zcash" && flag.enabled`; it does not depend on `privateInfo`. The bitcoin family does not set `useCustomConfirmAddress` (`families/bitcoin/index.ts` — only concordium does), so Zcash accounts go through `0x40` today.

After this change, for a Zcash account with the flag on and no UFVK — the default state of every existing Zcash account during rollout — `confirmAddress()` is suppressed, `ZcashShieldedVerify` is never mounted (no `shieldedAddress`), and `deviceAnimationBlock` is `null`. The step still renders `mainAccount.freshAddress` with a working copy control and QR modal, but nothing asks the user to check the device, and unlike the deliberate-skip path it shows no `currentAddress.messageIfSkipped` security alert. `isAddressVerified` stays `null` forever, so the step never reaches the success screen while `isModalLocked` (`modals/Receive/index.tsx`) keeps the backdrop click disabled — a dead end with an unconfirmed receive address on screen.

The validated specification linked from LIVE-35014 rules this out three times:

- requirement 12 — *"For Zcash accounts **without** a UFVK, fall back to the existing transparent-only `bridge.receive(..., { verify: true })` path, unchanged."*
- Constraints — *"Do not weaken transparent verification for any non-Zcash currency, **nor for Zcash accounts without a UFVK**."*
- acceptance criterion — *"A Zcash account without a UFVK verifies through the unchanged transparent path […] asserted."*

**Fix** — gate the opt-out on the shielded address actually being verifiable, and keep the standard path otherwise:

```ts
// 0x51 replaces 0x40 only when there is a UA to verify; otherwise the
// transparent address still needs its own device confirmation.
const usesCombinedConfirm = isZcashShielded && zcashShieldedAddress !== null;

useEffect(() => {
  if (specific?.useCustomConfirmAddress || usesCombinedConfirm) return;
  if (isAddressVerified === null) confirmAddress();
}, [specific?.useCustomConfirmAddress, usesCombinedConfirm, isAddressVerified, confirmAddress]);
```

and keep `<Receive2Device />` rendered in the activation-CTA state so the flow can still reach the success screen.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/__tests__/StepReceiveFunds.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    render(<StepReceiveFunds {...baseProps()} />, {
+      initialState: withFlagOverrides({ zcashShielded: { enabled: true } }),
+    });
+    expect(screen.getByTestId("receive-device-animation")).toBeInTheDocument();
+    expect(screen.queryByTestId("zcash-receive-device-animation")).not.toBeInTheDocument();
+  });
+});
+
+describe("StepReceiveFunds — Zcash shielded effects", () => {
+  it("suppresses the standard 0x40 confirmAddress call for Zcash even without a shielded address", async () => {
```
cted-ledger: ### 🔴 Critical — this test asserts the inverse of an acceptance criterion (confidence: 100)

The specification's criterion is *"A Zcash account without a UFVK verifies through the unchanged transparent path […] asserted"*. This test asserts that it does **not** — it locks in the regression flagged on `StepReceiveFunds.tsx:345`.

Once the suppression is gated on `zcashShieldedAddress !== null`, this case should invert: with no shielded address, `bridge.receive(..., { verify: true })` must still be called, and `receive-device-animation` must still render.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/__tests__/StepReceiveFunds.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+const { useLLDCoinFamily } = jest.requireMock("~/renderer/families") as {
+  useLLDCoinFamily: jest.Mock;
+};
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  useLLDCoinFamily.mockReturnValue({ useCustomConfirmAddress: true });
+});
+
+describe("StepReceiveFunds — device animation block", () => {
```
cted-ledger, line 63: ### 🔴 Critical — the specification's primary funds-loss test is missing, and the description says it is here (confidence: 95)

Specification requirement 3 and its acceptance criterion: *"Each block's QR shows that block's address and each copy control copies that block's address — asserted by an integration test that opens both QR modals and both copy actions in one session"*, described there as *"the feature's primary funds-loss risk"*.

This file adds 5 tests across two describes — device-animation selection (×3) and shielded effects (×2). There is no two-block layout assertion, no QR-modal test and no copy-action test anywhere in the PR. The description nevertheless states: *"4 tests added: `transitionTo` lifecycle + two-block layout (primary funds-loss guard: public block contains `freshAddress`, private block contains `shieldedAddress`, addresses differ)"*.

A crossed QR/copy target between the two blocks — the one failure mode that loses funds — would ship green.

Two further description/reality gaps worth correcting: `ZcashShieldedReceiveBlock.test.tsx` has 8 tests, not "10"; and the claimed "MOCK shortcut" test does not exist — that branch carries `/* istanbul ignore next */`, i.e. it is deliberately excluded from coverage.

**Fix** — write the test the specification asks for: render the step for a Zcash account with both addresses, assert `receive-public-address-block` contains `freshAddress` and `receive-private-address-block` contains `shieldedAddress`, that the two differ, and open plus copy from each block's own QR modal in one session.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
```diff
@@ -0,0 +1,202 @@
...
+            }}
+          >
+            <Trans i18nKey="zcash.receive.noUfvk.cta" />
+          </Button>
+        </Box>
+      </AlertBoxContainer>
+    );
+  }
+
+  return null;
```
cted-ledger: ### 🔴 Critical (same root cause as `StepReceiveFunds.tsx:345`) — this is the forbidden third state (confidence: 95)

Specification requirement 6 and its acceptance criterion: *"The component has exactly **two** states — address, or activation call to action. No third 'address unknown' branch exists."* The rationale given there is that the address is derived host-side from the persisted UFVK, so it exists exactly when the UFVK does; if the field is ever empty the persistence layer's recompute helper fills it, and *"the UI never renders that situation"*.

This `return null` is a third state, and `ZcashShieldedReceiveBlock.test.tsx` locks it in (*"renders nothing when ufvk is present but shieldedAddress is null"*). Combined with the `0x40` suppression, a user in this state gets an unverified transparent address **and** no private block at all — the worst of both branches, silently.

**Fix** — remove the state (rely on the recompute helper) or fall through to the activation CTA.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
```diff
@@ -0,0 +1,202 @@
...
+        if (cancelled) return;
+        const match = address === shieldedAddress;
+        onChangeAddressVerified(match, match ? null : new Error("Address mismatch"));
+      })
+      .catch(err => {
+        if (cancelled) return;
+        // Strip the message before propagating: device/IPC errors can echo the
+        // UFVK back in their message string, and onChangeAddressVerified feeds
+        // into logger.critical in ReceiveModal for non-user-refusal errors.
+        const safeErr = Object.assign(new Error("Verification failed"), {
```
cted-ledger: ### 🟡 Warning — every device error collapses into one opaque message, the mismatch is not a wrong-device-class error, and a plain rejection is logged as critical (confidence: 90)

Three distinct problems in this catch and the `Address mismatch` above it (L84):

1. **Distinct errors are erased.** Specification requirement 16 and its criterion: *"Disconnection, below-minimum app version and wrong-seed device each produce a distinct actionable error, with a test for the wrong-seed case."* Replacing every failure with `new Error("Verification failed")` makes that impossible — only `name` survives — and there is no wrong-seed test.

2. **The mismatch is not wrong-device-class.** Requirement 13 and its criterion: *"fail on mismatch with the same severity as `WrongDeviceForAccount`"* / *"a mismatch yields a wrong-device-class error"*. `new Error("Address mismatch")` has `name === "Error"`; `errors.Error` does not exist in `app.json`, so `TranslatedError` falls back to `errors.generic` → an untranslated *"Address mismatch"* title over *"Something went wrong. Please retry"*. This is the one error the specification says must convey a possible host/device derivation divergence, not just a wrong device.

3. **A user rejection is reported as a critical error.** `DmkSignerZcash.mapError` maps SW `6985` to `UserRefusedOnDevice`, and `ReceiveModal.handleChangeAddressVerified` only exempts `UserRefusedAddress` from `logger.critical(err)`. So tapping *Reject* on the `0x51` review screen emits a critical error event to monitoring on every refusal. (The user-facing copy is fine — `errors.UserRefusedOnDevice` resolves by name.)

Also, on the stated rationale: `0x51` never handles a UFVK — it returns only the UA, itself a public address already rendered in this component — so there is no view key for a device/IPC error message to echo back on this path. The sanitisation costs the error identities the specification requires and buys nothing here.

**Fix** — propagate `UserRefusedOnDevice` unchanged (or map it to `UserRefusedAddress` so the existing exemption applies), keep disconnection and version errors distinguishable, and give the mismatch a named wrong-device-class error with its own `errors.<Name>.{title,description}` keys plus a test.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -165,6 +165,20 @@ const Receive2Device = ({ name, device }: { name: string; device: Device }) => {
   );
 };
 
+const ZcashReceive2Device = ({ device }: { device: Device }) => {
+  const type = useTheme().theme;
+  return (
+    <>
+      <Box horizontal alignItems="center" flow={2} data-testid="zcash-receive-device-animation">
```
cted-ledger: ### 🟡 Warning — the shared-file diff was not confined to the currency-aware verify opt-out (confidence: 90)

Specification requirement 10: *"Make the auto-verify opt-out currency-aware in `families/types.ts` and `StepReceiveFunds.tsx`"*. Constraints: *"Confine the shared-file diff to the currency-aware verify opt-out. Every other Zcash behaviour lives in the post-alert component."* Same rule in `.agents/skills/coin-families-contract/SKILL.md`: generic screens must not branch on a currency, they consume optional contract slots implemented in `families/<family>/`.

`families/types.ts` is untouched — the contract was never made currency-aware. Instead this generic step, owned by `@ledgerhq/coin-integration` (`CODEOWNERS:175`), gained three Zcash-specific concerns: the `isZcashShielded` / `zcashShieldedAddress` pair with a structural cast to read `privateInfo`, this `ZcashReceive2Device` component, and the `transitionTo("receive")` effect. The generic step now knows what Orchard is, and the debt the specification deliberately accepted (Zcash UI accumulating in `families/bitcoin`) has landed in another team's shared file instead.

**Fix** — widen the slot as requirement 10 describes (e.g. `useCustomConfirmAddress?: boolean | ((account: Account) => boolean)`) and add an optional `StepReceiveFundsDeviceAnimation?: React.ComponentType<StepProps>` to `LLDCoinFamily`; implement both in `families/bitcoin/`, which already owns the Zcash branch. `StepReceiveFunds` then reads only from `specific`.

libs/live-signer-zcash/src/DmkSignerZcash.ts
```diff
@@ -190,6 +192,17 @@ export class DmkSignerZcash implements ZcashSigner {
     };
   }
 
+  async getShieldedAddress(path: string, display?: boolean): Promise<ZcashShieldedAddress> {
+    const { observable } = this.signer.getShieldedAddress(path, {
+      checkOnDevice: !!display,
+      skipOpenApp: true,
```
cted-ledger, line 198: ### 🟡 Warning — nothing gates the combined review on a minimum app version (confidence: 85)

The specification's Constraints: *"The combined review requires the **two-path** `INS_GET_SHIELD_ADDR`; the 3.6.0-era single-path variant shows the UA alone and cannot satisfy this. Rely on [the persistence task]'s minimum-version gate, not runtime feature detection."*

I can find no such gate: `getAppConfig` is declared in `libs/live-signer-zcash/src/types.ts` but has **zero callers** anywhere in the repo, and there is no version check in `coin-zcash` or in the Zcash desktop flows. This call also passes `skipOpenApp: true`, so nothing on the path observes app info either.

On an app that only honours the single-path form, the host would compare the returned UA, find it matching and mark the address verified — while the UI copy promises *"Verify both your transparent and private addresses on your Ledger"* and the transparent address was never displayed. That silently weakens the exact guarantee the feature advertises.

**Fix** — link the minimum-version gate if it landed elsewhere and say so in the description (the constraint explicitly depends on it), or add it.

libs/coin-modules/coin-zcash/src/bridge/index.test.ts
```diff
@@ -62,4 +62,29 @@ describe("createBridges", () => {
...
+    const getShieldedAddress = jest.fn().mockResolvedValue({ address: "u1testunified" });
+    const signerContext: SignerContext = jest.fn(async (_deviceId, fn) =>
+      fn({ getShieldedAddress }),
+    ) as unknown as SignerContext;
+    const { accountBridge } = createBridges(signerContext, coinConfig);
+
+    const result = await accountBridge.getShieldedAddress(account, { deviceId: "device-1" });
+
+    expect(result).toEqual({ address: "u1testunified" });
+    expect(getShieldedAddress).toHaveBeenCalledWith("m/32'/133'/0'", undefined);
```
cted-ledger: ### 🟡 Warning — this asserts a derivation path the real device action rejects (confidence: 90)

`GetShieldedAddressTask` (device-sdk-ts `origin/develop`, `signer-zcash` 0.6.0) requires exactly 5 levels with purpose `44'` and `change === 0`; a 3-level `32'` path fails with `InvalidStatusWordError` before any APDU is sent.

Production is fine — `sync.ts:486` builds `freshAddressPath` as `${accountPath}/0/${freshAddressIndex}` — but this test passes only because the signer is mocked, so the case named *"delegates … with the account's `freshAddressPath` by default"* proves nothing about the contract it names. It is also the only place covering the specification's criterion *"The transparent path sent is `account.freshAddressPath` — asserted"*.

The sibling suite already uses a realistic fixture: `sync.transparent.test.ts:195` → `44'/133'/0'/0/4`.

**Fix** — switch the fixture to a 5-level path and add the assertion at the UI level, where the value actually originates. Worth a comment too that `getFullViewingKey` and `getShieldedAddress` take different path shapes — the `32'` account path is correct for the former, invalid for the latter.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -315,18 +329,52 @@ const StepReceiveFunds = (props: StepProps) => {
...
 
+  const deviceAnimationBlock = isZcashShielded ? (
+    zcashShieldedAddress !== null ? (
+      <>
+        <Separator />
+        <ZcashReceive2Device device={device!} />
+      </>
+    ) : null
+  ) : (
+    // "Enable private balance" CTA state — no device animation
```
cted-ledger: ### 🟢 Suggestion — two small things in this block (confidence: 90 / 85)

1. **The comment is on the wrong branch.** `// "Enable private balance" CTA state — no device animation` sits on the `else` arm of `isZcashShielded`, i.e. the **non-Zcash** path, which does render the standard animation. The state it describes is the `: null` arm above it. On a branch that decides whether a device confirmation happens at all, a comment pointing at the wrong case is worth fixing.

2. **`device!` escapes the null guard.** The block is built above the `: device ? (…) : null` guard, so both non-null assertions are only safe because the element happens never to be rendered in the null-device arm — an invisible coupling between two distant pieces of code. Build it inside the `device ?` arm, or extract a component taking a narrowed `device: Device`.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -79,7 +79,7 @@ const Receive1ShareAddress = ({
   const shouldRenderMemoTagInfo = currency.family && MEMO_TAG_COINS.includes(currency.family);
 
   return (
-    <>
+    <Box data-testid="receive-public-address-block">
```
cted-ledger: ### 🟢 Suggestion — an extra flex level on every currency's receive modal, for a testid (confidence: 80)

`Receive1ShareAddress` went from a fragment to `<Box data-testid="receive-public-address-block">`, adding a flex-column wrapper for Bitcoin, Litecoin, Dash and every other currency — while the specification's criterion is *"Bitcoin, Litecoin and Dash Receive steps are unchanged"*.

**Fix** — put the testid on the existing inner `Box`, or use a wrapper with `display: contents`.

cted-ledger (commented):
## Code Review: PR #20756 — shielded receive block + `0x51` device confirm

Reviewed the full diff (19 files, +834/−14) against `develop`, the validated specification linked from LIVE-35014, `CONTRIBUTING.md`, `CODEOWNERS`, and `.agents/agents/code-reviewer.md` with its referenced skills. The device contract was cross-checked against `device-sdk-ts@origin/develop` (`signer-zcash` 0.6.0) and `app-zcash@origin/develop`. Findings are grouped below; the file-anchored ones are inline.

### Verified as correct

The design's central bet holds. `app-zcash` `handler_get_shielded_addr` in `UAddress` mode — P2 `0x00`, which is exactly what `GetShieldedAddressCommand` sends — calls `ui_display_shielded_address`, showing the UA as the reviewed address and the transparent address as a `"Public Address"` tag/value field on **one** screen, derived at the **full** 5-level path, so it is provably the account's own `freshAddress`. One device exchange really does cover both addresses.

Also verified: `signer-zcash` 0.6.0 (the pinned version) does export `getShieldedAddress` / `GetShieldedAddressDAError`; `resolveFamily()` routes zcash to `@ledgerhq/coin-zcash` under the same flag the UI gates on, so the `ZcashAccountBridge` cast is sound; the store has RTK thunk enabled, so the new `dispatch((d, getState) => …)` in `StepConfirmation` is legal, and `syncStateUpdater` is a plain action, so its `// synchronously committed` comment is accurate; verification goes through `getShieldedAddress` and never the view-key acquisition chain, as required; the private block owns its own QR modal state and reads `privateInfo.shieldedAddress` as-is; only `en/app.json` is touched; the effect deps are stable (`bridge` comes from the per-family promise cache, `handleChangeAddressVerified` is a stable `useCallback`); size is within budget.

**Not verified**: the test suites were not executed and no coverage figure is cited, so the test outcome and the specification's *"≥80% coverage on new files"* are **unverified**, not passing.

### 🔴 Critical

Must fix before merge.

1. **Transparent verification is removed for Zcash accounts without a shielded address** (confidence: 100) — → inline on `StepReceiveFunds.tsx:345`. The specification forbids this in requirement 12, in the Constraints, and in an acceptance criterion. Blast radius includes the forbidden third component state (→ inline on `ZcashShieldedReceiveBlock.tsx:201`), and the behaviour is locked in by a test (→ inline on `StepReceiveFunds.test.tsx:99`).
2. **The specification's primary funds-loss test does not exist, and the description says it does** (confidence: 95) — → inline on `StepReceiveFunds.test.tsx:63`.

### 🟡 Warning

Should fix.

3. **The shared-file diff was not confined to the currency-aware verify opt-out** (confidence: 90) — → inline on `StepReceiveFunds.tsx:172`. `families/types.ts` is untouched; three Zcash concerns were added to a generic step owned by another team.
4. **Every device error collapses into one opaque message; the mismatch is not a wrong-device-class error; a plain rejection is logged as critical** (confidence: 90) — → inline on `ZcashShieldedReceiveBlock.tsx:91`.
5. **Nothing gates the combined review on a minimum app version** (confidence: 85) — → inline on `DmkSignerZcash.ts:198`.
6. **The bridge test asserts a derivation path the device action rejects** (confidence: 90) — → inline on `coin-zcash/src/bridge/index.test.ts:76`.
7. **Missing changeset** (confidence: 90) — `CONTRIBUTING.md:68`: *"A changeset is required for any user-facing change or library API modification."* This PR is both: a user-facing Receive-modal feature **and** a public API addition (`getShieldedAddress` on the `@ledgerhq/coin-zcash` account bridge and its signer interface, plus `ZcashSigner.getShieldedAddress` / `ZcashShieldedAddress` in `live-signer-zcash`). No `.changeset/` entry in the diff. → `pnpm changeset`.

### 🟢 Suggestion

8. **i18n keys partly outside the required namespace** (confidence: 90) — → inline on `app.json:9041`.
9. **Misplaced comment + `device!` escaping the null guard** (confidence: 90 / 85) — → inline on `StepReceiveFunds.tsx:371`.
10. **An extra flex level on every currency's receive modal, for a testid** (confidence: 80) — → inline on `StepReceiveFunds.tsx:82`.
11. **Figma-sourced copy is not confirmed** (confidence: 80) — the specification lists the two labels, the privacy notice and the no-UFVK call to action as items to verify against Figma node 2387-1457, and "Figma-sourced copy strings" as missing. Four new strings ship here with no statement that they were confirmed; the public block keeps the shared *"Address for {name}"* rather than the *"Public address for {account}"* the specification's context describes; and no privacy notice is rendered. Please confirm against the design and say so in the description.
12. **Undeclared scope addition** (confidence: 85) — `ZCashExportKeyFlowModal/{Body,types}.ts` and `steps/StepConfirmation.tsx` add a `source: "receive"` round-trip that reopens `MODAL_RECEIVE` after activation. It is not in the specification and not listed in the description's "Changes" section, which omits 7 of the 19 files — including all of `coin-zcash` and `live-signer-zcash`. It is a reasonable UX addition, but note the reopened modal restarts at the `account` step, so the user walks the whole flow again. Worth stating the intent, or splitting it out.

### Specification alignment

Against the acceptance criteria of the validated specification linked from LIVE-35014:

| # | Criterion | Verdict |
|---|---|---|
| 1 | Two labelled blocks, public then private, matching Figma | partial — blocks and order correct; public label still the shared `currentAddress.for`; Figma unverified |
| 2 | Each block's QR/copy asserted by one integration test covering both modals and both copy actions | **not met** |
| 3 | Public block keeps `freshAddress`, reuse alert, derivation-mode tag | met |
| 4 | Bitcoin/Litecoin/Dash unchanged, dash/taproot post-alerts still render — asserted by test | partial — code preserves them; no test asserts the dash alert |
| 5 | No-UFVK account shows the CTA opening the export modal | met |
| 6 | Exactly two component states, no third "address unknown" branch | **not met** — a third `return null` state exists and is tested |
| 7 | Flag off ⇒ step indistinguishable from today | met (bar the extra `Box` wrapper) |
| 8 | Testnet account renders `utest1…` | not asserted |
| 9 | ~106-char address readable in block, QR modal and after copy at minimum width | not asserted |
| 10 | New i18n keys under `zcash.shielded.*`, no generic key repurposed | partial — 2 of 4 keys |
| 11 | Distinct stable `data-testid` on both blocks | met |
| 12 | Exactly one device exchange with the display flag — asserted by counting mock-bridge calls | partial — display flag asserted at signer level; call count never asserted |
| 13 | UA compared; mismatch ⇒ wrong-device-class error, both unverified | partial — comparison present, error class wrong |
| 14 | Transparent path sent is `account.freshAddressPath` — asserted | partial — asserted only with a fixture the device rejects |
| 15 | Zcash without UFVK verifies through the unchanged transparent path; non-Zcash unchanged — both asserted | **not met** — inverted, and a test asserts the suppression |
| 16 | Rejection leaves both unverified, retry works, never partially successful | met |
| 17 | Disconnection / below-minimum-version / wrong-seed give distinct actionable errors, wrong-seed tested | **not met** |
| 18 | `isAddressVerified` / `verifyAddressError` / retry / skip still work, incl. re-verify back to `device` | not asserted |
| 19 | `MOCK` shortcut resolves verification without a device | partial — implemented for the address state only, excluded from coverage, untested |
| 20 | Staking-flow continuation and Receive analytics still fire for Zcash | not asserted |
| 21 | `pnpm build && pnpm typecheck`, LLD unit + integration tests, ≥80% coverage on new files | **unverified** — no executed report cited |

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -9033,6 +9037,12 @@
         }
       }
     },
+    "receive": {
+      "noUfvk": {
```
cted-ledger: ### 🟢 Suggestion — these keys land outside the required namespace (confidence: 90)

Specification requirement 8 and its criterion: *"New i18n keys exist under `zcash.shielded.*`; no generic key is repurposed."* `zcash.shielded.receive.verifyAddresses` and `.privateFor` comply; this pair opens a new top-level `zcash.receive` branch instead. The description also states that all shielded keys live under `zcash.shielded.receive.*`.

**Fix** — move to `zcash.shielded.receive.noUfvk.*`. (The second half of the criterion is respected: no generic key was repurposed.)

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
vladyslavchupovskiy-ext-art: Accurate concern from business logic point of view and correctness across LLD flows.

But I want to raise a UX concern cause this is why I actually came up with this decision and this why it's flagged
Critical right now , for me when I was testing as user was kinda weird to have Export UFKV button that blocks seeing shielded address while I still have succesfull animation and device interaction . But it could be that I biased cause of task , and fresh look from the other pair of eyes is what I needed. 

Agree on fix. On UX debatable but not blocking the fix. Will provide fix in next commit

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/__tests__/StepReceiveFunds.test.tsx
vladyslavchupovskiy-ext-art: Related to first , gonna fix in the commit

libs/live-signer-zcash/src/DmkSignerZcash.ts
vladyslavchupovskiy-ext-art: My mistake not adding this in prev tickets and PRs , from more easy point of view , I can propose to create separate ticket for that , device sdk pr is waiting so we will just create follow-up PR for this guard it would be small and clean . What do you think @cted-ledger ?

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
vladyslavchupovskiy-ext-art: Agree, it's pain for this PR but for future architecture it's worth doing

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
vladyslavchupovskiy-ext-art: Yeah good catch, would be fixed in next commit

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
vladyslavchupovskiy-ext-art: We decided with @cted-ledger to just make it simple and show warning for user that private balance is not activated but still can normally proceed with 0x40 transparent . Solution is both simple for codebase and for user

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/__tests__/StepReceiveFunds.test.tsx
vladyslavchupovskiy-ext-art: gap is clear here, will be fixed in next commit

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
vladyslavchupovskiy-ext-art: No longer valid cause we decided to exterminate CTA logic there and go with simple warning ( would be visible in next commit )

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
```diff
@@ -0,0 +1,202 @@
...
+        clearTimeout(t);
+      };
+    }
+
+    // A single 0x51 call covers both addresses: the device derives the
+    // transparent address from account.freshAddressPath (the bridge default)
+    // and shows it alongside the UA, so no second 0x40 call is needed.
+    // The returned UA is compared against the persisted value to catch both
+    // a wrong device and a host/device derivation divergence.
+    bridge
```
may01, line 90: bridge.getShieldedAddress(...) is invoked through as unknown as ZcashAccountBridge with no typeof … === "function" check.

vladyslavchupovskiy-ext-art: Yeah there is both sides of a coin here, this modal is only appears when feature flag is on , if it's we always use ZCashAccountBridge , in other side yeah fix is worth it , we can do an assertion function that will check that and add test scenario for it . 
I know that we kinda in the middle of migration zcash from coin-bitcoin to it's own coin module so it adds confusion as well

Still gonna add this thanks !

libs/coin-modules/coin-zcash/src/bridge/index.test.ts
vladyslavchupovskiy-ext-art: Agree , production case would send 44' not 32' so worth changing that

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -315,18 +329,54 @@ const StepReceiveFunds = (props: StepProps) => {
...
   const CustomStepReceiveFunds = specific?.StepReceiveFunds;
 
-  // Families that own their confirm-address lifecycle opt out of the shared auto-trigger.
+  const zcashShieldedFeature = useFeature("zcashShielded");
+  const isZcashShielded =
+    mainAccount?.currency.id === "zcash" && zcashShieldedFeature?.enabled === true;
+  const zcashShieldedAddress = isZcashShielded
+    ? ((mainAccount as { privateInfo?: { shieldedAddress?: string | null } }).privateInfo
+        ?.shieldedAddress ?? null)
+    : null;
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Fixed — usesCombinedConfirm is now isZcashShielded && zcashShieldedAddress !== null. The auto-confirm is only suppressed when a shielded address is actually present. Without one, confirmAddress() fires normally and Receive2Device renders as the standard verification block.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -315,18 +329,54 @@ const StepReceiveFunds = (props: StepProps) => {
...
+
+  // Zcash accounts with a shielded address use 0x51 (ZcashShieldedVerify) for combined
+  // transparent+shielded confirmation — they must not also trigger the standard 0x40 path.
+  // Zcash accounts WITHOUT a shielded address (no UFVK yet, or UFVK pending derivation)
+  // fall back to the unchanged transparent-only 0x40 path so address verification is
+  // not silently skipped.
+  const usesCombinedConfirm = isZcashShielded && zcashShieldedAddress !== null;
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Valid — changing !== null to !!zcashShieldedAddress in both places (usesCombinedConfirm and the transition guard on line 356) to match ZcashShieldedReceiveBlock's truthiness check. An empty string should behave the same as null: no address, no 0x51, 0x40 runs normally.

apps/ledger-live-desktop/static/i18n/en/app.json
```diff
@@ -9001,6 +9001,13 @@
         "label": "Birthday height",
         "tooltip": "The birthday height is the point on the blockchain where your wallet begins scanning for your funds."
       },
+      "receive": {
+        "verifyAddresses": "Verify both your transparent and private addresses on your Ledger",
+        "privateFor": "Private address for <1><0>{{name}}</0></1>",
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: No change needed. react-i18next counts text nodes when assigning child indices, so index 0 is the leading "Private address for " string and index 1 is <strong>. <1><0>{{name}}</0></1> is correct. The identical pattern is already used in the production currentAddress.for key (line 1602) with the same Trans structure and renders correctly.

apps/ledger-live-desktop/static/i18n/en/app.json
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: same as below no changes needed

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
vladyslavchupovskiy-ext-art: make sense !

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
vladyslavchupovskiy-ext-art: done thanks !

apps/ledger-live-desktop/static/i18n/en/app.json
vladyslavchupovskiy-ext-art: done !

libs/live-signer-zcash/src/types.ts
```diff
@@ -107,10 +103,14 @@ export type BitcoinCreateTransactionLike = {
...
+};
+
 export type ZcashSigner = {
-  getAppConfig: () => Promise<ZcashAppConfig>;
   getAddress: (path: string, display?: boolean) => Promise<ZcashAddress>;
   getFullViewingKey: (path: string) => Promise<ZcashViewKey>;
+  getShieldedAddress: (path: string, display?: boolean) => Promise<ZcashShieldedAddress>;
   createPaymentTransaction: (arg: BitcoinCreateTransactionLike) => Promise<string>;
   signPcztTransaction: (pczt: PcztTransaction) => Promise<SignPcztTransactionResult>;
   signMessage: (path: string, messageHex: string) => Promise<ZcashSignature>;
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Just an miss, would be fixed

apps/ledger-live-desktop/static/i18n/en/app.json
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: valide would be fixed

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
```diff
@@ -0,0 +1,189 @@
...
+          transitionTo={transitionTo}
+        />
+        <AlertBoxContainer data-testid="receive-private-address-block">
+          <Box horizontal alignItems="center" flow={2} mb={4}>
+            <Text ff="Inter|SemiBold" color="neutral.c100" fontSize={4} style={{ flex: 1 }}>
+              <Ellipsis>
+                <Trans i18nKey="zcash.shielded.receive.privateFor">
+                  {"Private address for "}
+                  <strong>{name}</strong>
+                </Trans>
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: same here valid would be fixed

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
cted-ledger: **Resolved** — verified on `779b58f`.

`useCustomConfirmAddress` is now the family predicate in `families/bitcoin/index.ts`, returning `true` only when the flag is on **and** the currency is zcash **and** `privateInfo.shieldedAddress` is set. Without a shielded address the standard `confirmAddress()` runs again, and `ZcashReceiveDeviceAnimation` falls back to `Receive2Device`, so the step reaches its success state. This generic step no longer contains any Zcash branch at all. Thanks — this is the right shape.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/__tests__/StepReceiveFunds.test.tsx
cted-ledger: **Resolved** — the test now exercises the generic mechanism in both directions (*"suppresses … when the family function returns true"* / *"runs the standard confirmation when the family function returns false"*), which matches the acceptance criterion.

One gap remains, raised as its own comment on `families/bitcoin/index.ts`: both tests stub the predicate with `jest.fn(() => true/false)`, so the real decision logic is still untested.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/__tests__/StepReceiveFunds.test.tsx
cted-ledger: **Partially resolved** — two of the four sub-assertions are now covered:

- `ZcashShieldedReceiveBlock.test.tsx` → *"private QR modal shows the shielded address"* opens the modal via the in-block "Show QR Code" link and asserts the address ✅
- `StepReceiveFunds.test.tsx` → *"two-block address isolation"* asserts the public block contains `freshAddress`, the private block contains `shieldedAddress`, and that they differ ✅ (with the private block mocked, so the real component's address is only covered by the test above)

Still missing from the criterion (*"both QR modals and both copy actions in one session"*): no test of either **copy control**, and no test of the **public** QR modal. `ReadOnlyAddressField` is the copy control for both blocks and takes the address as a prop — asserting the prop each block passes would close this cheaply.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
cted-ledger: **Resolved** — the component now has exactly the two states the specification allows: shielded address → private block; otherwise → the informational alert. The `it.each` case covers both "no UFVK" and "UFVK present but address not yet derived" through the same branch. ✅

Note the CTA that state used to carry has gone with it — raised separately on the new alert branch.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
cted-ledger: **Resolved** — all three points, verified on `779b58f`:

1. Errors now propagate unchanged (`err instanceof Error ? err : new Error(String(err))`), so disconnection, app-version and transport failures stay distinguishable ✅
2. The mismatch is `new WrongDeviceForAccount()` — the wrong-device class requirement 13 asks for, and it has real i18n copy (`errors.WrongDeviceForAccount`) ✅
3. `Receive/index.tsx` now exempts `UserRefusedOnDevice` from `logger.critical`, so a plain rejection no longer reports a critical event ✅

One nuance for later, not a blocker: the shared `errors.WrongDeviceForAccount` description points only at the recovery phrase, whereas requirement 13 asks the message not to assume a wrong device is the only cause (it also catches a host/device derivation divergence). A Zcash-specific subclass with its own copy would cover that half.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
cted-ledger: **Resolved** — implemented as the contract extension requirement 10 describes:

- `families/types.ts` — `useCustomConfirmAddress?: boolean | ((account, featureFlags) => boolean)` plus the new `StepReceiveFundsDeviceAnimation?: ComponentType<ReceiveStepProps & { fallback }>` slot, both documented
- `families/bitcoin/index.ts` — the Zcash predicate and the animation component live here
- `StepReceiveFunds.tsx` — no currency branch left; it resolves the slot and passes the standard animation as `fallback`

The hook-free constraint on the predicate is explicitly documented, and `useFeatureFlags()` (which returns the resolved `Features` map) is the right way to feed it without a hook inside the family. Verified that the test asserting the predicate receives `{ zcashShielded: { enabled: true } }` proves that wiring. ✅

libs/live-signer-zcash/src/DmkSignerZcash.ts
cted-ledger: **Still open** — verified on `779b58f`.

The description now says *"`DmkSignerZcash.getAppConfig()` is wired in `libs/live-signer-zcash` and will gate the Zcash app version check"*, with a dependency on device-sdk-ts#1743. Checking the actual state:

- device-sdk-ts#1743 is **merged** (2026-08-15) ✅ — the upstream `GetAppConfigCommand` exists
- but `DmkSignerZcash.getAppConfig()` on `develop` is `throw new Error("Not implemented")`, and this PR does not touch it (`DmkSignerZcash.ts` adds only `getShieldedAddress`)
- and it still has **zero callers** anywhere in the repo

So nothing gates the combined review today: on an app honouring only the single-path `INS_GET_SHIELD_ADDR`, the host would compare the returned UA, find it matching and mark the address verified, while the UI copy promises *"Verify both your transparent and private addresses on your Ledger"* and the transparent address was never displayed.

Happy for this to be a tracked follow-up rather than a blocker — but please make the description say what is actually in place ("`getAppConfig` is still a stub; the gate lands in a follow-up") rather than "is wired", and link the follow-up issue.

libs/coin-modules/coin-zcash/src/bridge/index.test.ts
cted-ledger: **Resolved** — the fixture is now `44'/133'/0'/0/4`, matching what `sync.ts:486` builds in production.

I also checked the collateral change to the two `getFullViewingKey` assertions, since the two calls take different path shapes: `GetFullViewingKeyTask` accepts a path of **either** purpose with at least 3 levels and reduces it to `32'/coin'/account'` (`deriveUfvkAccountPaths`), so a 5-level `44'` path is valid input there too — and it is what the real caller passes (`StepDevice.tsx` calls `getFullViewingKey(account, { deviceId })`, which defaults to `freshAddressPath`). The updated assertions are therefore more faithful than the old `m/32'/…` ones, which that helper would in fact have mis-parsed (it does not strip an `m/` prefix). Good change. ✅

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
cted-ledger: **Partially resolved** — the misplaced comment is gone with the refactor ✅.

`device!` remains: `standardDeviceAnimation` is still built above the `: device ? (…) : null` guard, so the assertion is only safe because the element is never rendered in the null-device arm. The animation slot now receives the same `props` (with a possibly-null `device`) and correctly guards on `!device` internally, which makes the coupling a bit more diffuse rather than less. Building the block inside the `device ?` arm would remove the assertion entirely.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
cted-ledger: **Resolved** — `<Box data-testid="receive-public-address-block" style={{ display: "contents" }}>` keeps the testid without introducing a layout box for the other currencies. ✅

apps/ledger-live-desktop/static/i18n/en/app.json
cted-ledger: **Resolved** — all four keys now live under `zcash.shielded.receive.*`. ✅

Two things worth noting: `privateFor` also moved to explicit `values={{ name }}` + `components={[<strong />]}` instead of relying on child-index inference, which is clearer than the `currentAddress.for` pattern it was copied from; and the `noUfvk.cta` key was dropped along with the button — see the comment on that branch.

apps/ledger-live-desktop/src/renderer/families/bitcoin/index.ts
```diff
@@ -10,11 +11,23 @@ import AccountBalanceSummaryFooter from "./AccountBalanceSummaryFooter";
...
 const family: BitcoinFamily = {
   sendAmountFields,
   sendRecipientFields,
   StepReceiveFundsPostAlert,
+  StepReceiveFundsDeviceAnimation,
+  // A Zcash account with a shielded address confirms its transparent and shielded
+  // addresses with a single device exchange, so the shared receive step must not
+  // also run the standard transparent-only confirmation. Anything else — no
+  // shielded address yet, or the feature turned off — keeps the standard path.
+  useCustomConfirmAddress: (account, featureFlags) => {
```
cted-ledger, line 26: ### 🟡 Warning — the predicate that decides whether a receive address gets device-verified is untested (confidence: 85)

This function is now the single decision point for "does this account confirm its address on the device, and through which APDU". It is the fix for the critical finding on the previous revision, and it has no test: there is no test file importing the `families/bitcoin` family object, and both `StepReceiveFunds.test.tsx` cases stub it with `jest.fn(() => true)` / `() => false`, so they prove the generic step honours *a* predicate, not that *this* predicate is right.

The four branches all change whether a device confirmation happens:

- flag off → `false` (standard `0x40`)
- non-zcash bitcoin-family account (btc / ltc / dash) → `false` (standard `0x40`)
- zcash, flag on, no `shieldedAddress` → `false` (standard `0x40` — this is the regression that was just fixed)
- zcash, flag on, `shieldedAddress` set → `true` (combined `0x51`)

A plain unit test on the exported family object covers all four in a few lines and locks the fix in. Worth it given what regresses silently if the third branch ever flips back.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
```diff
@@ -0,0 +1,190 @@
...
+                </Box>
+              )}
+            />
+          </Modal>
+        </AlertBoxContainer>
+      </>
+    );
+  }
+
+  // No shielded address yet — transparent 0x40 verification runs in the parent.
```
cted-ledger, line 186: ### 🟡 Warning — the activation call to action was removed, which regresses an acceptance criterion (confidence: 90)

Specification requirement 6: *"**No UFVK** (private balance not enabled): render the block with a call to action opening `ZCashExportKeyFlowModal`, rather than hiding it or showing an empty field."* Acceptance criterion: *"An account with no UFVK shows the activation call to action opening `ZCashExportKeyFlowModal`."*

The previous revision had that button; this one renders only the alert, and `ZcashShieldedReceiveBlock.test.tsx` now asserts the CTA is absent (*"No CTA — 0x40 transparent verification runs in the parent"*). The two are independent though: the CTA opens the export modal, it does not interact with the transparent `0x40` confirmation, which now runs in the parent regardless. Dropping the round-trip that reopened `MODAL_RECEIVE` was right; dropping the button with it was not required by that.

The user-visible result is an alert that says *"Enable your private balance to receive to a private address."* with no way to do so from where they are reading it.

**Fix** — restore the button (and the `zcash.shielded.receive.noUfvk.cta` key) dispatching `openModal("MODAL_ZCASH_EXPORT_KEY", { account })`, or, if product has agreed to rely on the account-page entry point instead, say so in the description and reword the alert so it points there.

.changeset/zcash-ironwood-spend-auth-pczt.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ledgerhq/coin-zcash": patch
+---
+
+Forward Ironwood spend-auth signatures to PCZT finalization (LIVE-35956).
```
cted-ledger, line 5: ### 🟡 Warning — this changeset describes a different change, and this PR's own API addition has none (confidence: 85)

This file documents *"Forward Ironwood spend-auth signatures to PCZT finalization (LIVE-35956)"* — `signOperation` passing `spendAuthSig` values to `combine`, with `@ledgerhq/zcash-utils` bumped to 2.1.1. None of that is in this diff: `coin-zcash/src/bridge/signOperation.ts` is untouched and no `package.json` changes, so nothing here bumps `zcash-utils`. As it stands it would publish a `@ledgerhq/coin-zcash` release note for work this PR does not contain.

Meanwhile the change this PR *does* make to a published API — `getShieldedAddress` on the `ZcashAccountBridge`, plus `ZcashSigner.getShieldedAddress` / `ZcashShieldedAddress` in `live-signer-zcash` — is not described by any changeset, which is what `CONTRIBUTING.md:68` asks for.

**Fix** — move this changeset to the PR that carries the Ironwood finalization change (or drop it if that one already shipped with its own), and add one for the shielded-address bridge capability.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
```diff
@@ -315,18 +316,38 @@ const StepReceiveFunds = (props: StepProps) => {
...
     }
-  }, [specific?.useCustomConfirmAddress, isAddressVerified, confirmAddress]);
+  }, [usesCustomConfirm, isAddressVerified, confirmAddress]);
 
   if (CustomStepReceiveFunds) {
     return <CustomStepReceiveFunds {...props} />;
   }
 
+  const DeviceAnimationSlot = specific?.StepReceiveFundsDeviceAnimation;
+  const standardDeviceAnimation = <Receive2Device device={device!} name={name} />;
```
cted-ledger: ### 🟢 Suggestion — `device!` still escapes the null guard (confidence: 80)

Carried over from the previous revision: this element is built above the `: device ? (…) : null` guard further down, so the non-null assertion is only safe because it is never rendered in the null-device arm. The animation slot now receives the same possibly-null `device` through `{...props}` and guards on it internally, so the invariant is spread across two files rather than tightened.

Building `deviceAnimationBlock` inside the `device ?` arm — or extracting a small component that takes a narrowed `device: Device` — removes the assertion and the implicit contract with it.

cted-ledger (commented):
## Follow-up review: `779b58f` — re-checked against the previous findings

Re-verified every point from [the earlier review](https://github.com/LedgerHQ/ledger-live/pull/20756#pullrequestreview-4936336139) against the current head, plus a fresh pass over what the new revision introduces. Per-thread verdicts are posted as replies on the original comments; the four items that still need action are inline.

### Fixed

Both criticals are genuinely resolved, and the architectural one was solved the right way rather than patched:

- **Transparent verification without a shielded address** — `useCustomConfirmAddress` became a family predicate gated on `privateInfo.shieldedAddress`, so those accounts go back through the standard `0x40` path and reach the success state.
- **The forbidden third component state** — collapsed to the two states the specification allows.
- **Shared-file confinement / coin-families contract** — `families/types.ts` gained the function form of `useCustomConfirmAddress` and a `StepReceiveFundsDeviceAnimation` slot; `StepReceiveFunds.tsx` no longer contains a single currency branch. The hook-free constraint on the predicate is documented and `useFeatureFlags()` feeds it correctly.
- **Device error handling** — errors propagate unchanged, the mismatch is a real `WrongDeviceForAccount`, and `UserRefusedOnDevice` is now exempt from `logger.critical`.
- **Bridge test fixture** — now `44'/133'/0'/0/4`. I also checked the collateral change to the `getFullViewingKey` assertions and it is correct: that task accepts a path of either purpose with ≥3 levels and reduces it to `32'/coin'/account'`, and a 5-level path is what the real caller passes.
- **i18n namespace**, **the flex wrapper** (`display: contents`), **the misplaced comment**, and **the undeclared export-modal round-trip** — all resolved.

Also verified as safe, since it is new here: `transitionTo` is a stable `useCallback` in `Stepper` (deps `[onStepChange, steps]`, both memoised upstream), so adding it to the `0x51` effect's dependencies does not re-dispatch the device call.

### Still open

- 🟡 **The predicate that fixes the critical finding is untested** — → inline on `families/bitcoin/index.ts:26`.
- 🟡 **The activation call to action was removed**, regressing requirement 6 / its acceptance criterion — → inline on `ZcashShieldedReceiveBlock.tsx:182`.
- 🟡 **The changeset describes a different change**, and this PR's own API addition still has none — → inline on `.changeset/zcash-ironwood-spend-auth-pczt.md:5`.
- 🟡 **No minimum-app-version gate** — → reply on the original `DmkSignerZcash.ts` thread. `getAppConfig()` is still `throw new Error("Not implemented")` with zero callers, so the description's "is wired" overstates it; device-sdk-ts#1743 is merged, but nothing consumes it yet.
- 🟢 **`device!` still escapes the null guard** — → inline on `StepReceiveFunds.tsx:339`.
- 🟢 **The description's "Changes" section is stale** — it attributes the `0x40` suppression and the `transitionTo` effect to `StepReceiveFunds.tsx`, where neither lives any more (they are in `families/bitcoin/index.ts` and `ZcashShieldedVerify` respectively), and it omits `families/types.ts`, `families/bitcoin/index.ts`, `ZcashReceiveDeviceAnimation.tsx` and its test, `Receive/index.tsx`, `coin-zcash`, `live-signer-zcash` and the changeset — 8 of the 21 files.
- 🟢 **Figma copy still unconfirmed** — the public block keeps the shared *"Address for {name}"* rather than the *"Public address for {account}"* the specification's context describes, and no privacy notice is rendered.

### Specification alignment — delta

| # | Criterion | Before | Now |
|---|---|---|---|
| 2 | Both QR modals and both copy actions asserted in one session | not met | partial — private QR + two-block isolation covered; copy controls and the public QR modal still untested |
| 5 | No-UFVK account shows the CTA opening the export modal | met | **regressed** — alert only, and a test asserts the CTA is absent |
| 6 | Exactly two component states | not met | **met** |
| 8 | Testnet `utest1…` renders | not asserted | **met** |
| 10 | i18n keys under `zcash.shielded.*` | partial | **met** |
| 12 | Exactly one device exchange, display flag — asserted by counting calls | partial | **met** (`toHaveBeenCalledTimes(1)` + `display: true`) |
| 13 | Mismatch ⇒ wrong-device-class error | partial | **met** (shared copy still points only at the recovery phrase — noted on the thread) |
| 14 | Transparent path is `freshAddressPath` — asserted | partial | **met** |
| 15 | Zcash without UFVK verifies through the unchanged transparent path | not met | **met** in code; the family predicate itself is still untested |
| 17 | Disconnection / below-minimum-version / wrong-seed give distinct errors | not met | partial — errors are distinguishable again; the version gate does not exist yet, no wrong-seed test |
| 4 | Dash/taproot post-alerts still render — asserted by test | partial | unchanged — still no assertion on the dash alert |
| 21 | build / typecheck / tests / ≥80% coverage on new files | unverified | unverified — no executed report cited |

Nothing in the remaining list touches the funds-safety property that the first review blocked on. The two that matter most before merge are the untested predicate and the misfiled changeset.

.changeset/zcash-ironwood-spend-auth-pczt.md
vladyslavchupovskiy-ext-art: That's was discussed with @cted-ledger  nad @may01  that we add this here

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
vladyslavchupovskiy-ext-art: No, we decided with @cted-ledger that for clean UX we gonna have warning for user to activate shielded balance , it's easy and straightforward and we guard ourselves from un-clear ux flow from customer perspective

apps/ledger-live-desktop/src/renderer/families/bitcoin/index.ts
vladyslavchupovskiy-ext-art: Done

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
cted-ledger: Yes you are right

apps/ledger-live-desktop/src/renderer/families/bitcoin/index.ts
cted-ledger: **Resolved** — verified on `fc7c132b`.

`bitcoinFamilyPredicates.test.ts` imports the real family object and covers all four branches, including the three shapes of "zcash, flag on, no shielded address" (no `privateInfo`, `ufvk` only, `shieldedAddress: null`) — the branch that regressed and was fixed. That is exactly the lock this needed.

Confirmed it actually runs: the file matches the default `testMatch`, is not in `testPathIgnorePatterns`, and *Desktop Unit Tests* passed on the run created after that commit. Importing `../index` pulls `live-common-setup` and its worker spawning, which is safe here because `tests/jestSetup.js` installs a `WorkerMock` as `global.Worker`.

Only remark, and purely optional: the predicate is pure, so moving it to its own module would let a test reach it without loading the whole family surface. Not worth a change on its own.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashShieldedReceiveBlock.tsx
cted-ledger: **Closed — product decision, understood.** Thanks for the context, and noted that @cted-ledger confirmed it on this thread.

For the record so it does not get re-flagged later: the validated specification still carries the CTA as requirement 6 and as an acceptance criterion, so it now contradicts the shipped decision. We will update the specification on our side to record "warning alert, no CTA" as the intended behaviour — nothing needed from you here.

.changeset/zcash-ironwood-spend-auth-pczt.md
cted-ledger: **Closed — deliberate, understood.** Thanks, and noted that this was agreed with @cted-ledger and @may01.

For completeness: *Changeset Validation* passes, and since the file already bumps `@ledgerhq/coin-zcash` a release will carry the new `getShieldedAddress` bridge capability even though the note describes the Ironwood work. No action from my side.

apps/ledger-live-desktop/src/renderer/modals/Receive/steps/StepReceiveFunds.tsx
cted-ledger: **Resolved** — the block moved inside the `device ? (…)` arm, so `device` is narrowed and both non-null assertions are gone. That removes the implicit contract between the two locations entirely. ✅

cted-ledger (approved):

fAnselmi-Ledger (approved):

PR #20757, opened 2026-08-13
ci(test-additional): add workflow SAST and lint checks

lewisd5:
## What

Adds two jobs to the "Test Additional" workflow that run only when
`.github/` or `tools/actions/` files are touched:

- **changes** — queries the GitHub API for PR file changes (no checkout) and sets a `workflow-files` output flag, following the same pattern as the `live-common-affected` composite.
- **ci-sast** — runs [zizmor](https://github.com/zizmorcore/zizmor-action) at `min-severity=high` (results uploaded to the Security tab via SARIF) and [actionlint](https://github.com/rhysd/actionlint) (sourced from mise). Both tools cover `.github/workflows/` and `tools/actions/` composites per the existing `.github/actionlint.yml` path config.

## Why

Workflow files are a high-value attack surface and currently have no automated CI gate. Running these checks only on PRs that touch CI files keeps noise low on unrelated work.

## Testing

This PR touches `.github/`, so the `ci-sast` job will run against itself on CI.

### 🔗 Context

Test run: https://github.com/LedgerHQ/ledger-live/actions/runs/31704457255/job/94461410823
Ticket: https://ledgerhq.atlassian.net/browse/LIVE-35006

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -345,6 +345,9 @@ jobs:
     name: "Test Additional"
     if: ${{!github.event.pull_request.head.repo.fork}}
     uses: LedgerHQ/ledger-live/.github/workflows/test-additional.yml@develop
+    permissions:
+      security-events: write
+      contents: read
 
   test-mobile-e2e-lint:
     name: "Mobile E2E Lint & Typecheck"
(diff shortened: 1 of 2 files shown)


dgreen-ledger (approved):
Overall a solid addition — pinned SHAs, scoped permissions on `ci-sast`, clean no-checkout approach for change detection.

One thing to fix before merge: the `changes` job calls `github.rest.pulls.listFiles` which needs `pull-requests: read`, but the caller in `build-and-test-pr.yml` only grants `security-events: write` and `contents: read`. In GitHub's reusable workflow model the called workflow can't expand beyond what the caller grants, so the API call will 403 or return empty. Add `pull-requests: read` to the caller's permissions block:

```yaml
permissions:
  security-events: write
  contents: read
  pull-requests: read
```

PR #20765, opened 2026-08-13
[LWDM] refactor(device-model): make device model IDs compatible

OlivierFreyssinet:
### 📝 Description

This PR prepares `DeviceModelId` for the new monorepo architecture without making new-architecture packages depend on legacy public libraries.

#### Why

`libs/device-intent` needs to move to `features/platform/`, but feature packages cannot depend on legacy packages under `libs/`. Today, it imports `DeviceModelId` from `@ledgerhq/types-devices`, which blocks that move.

A Ledger device model is product vocabulary, rather than a transport implementation detail: it drives device animations and illustrations, supported capabilities, onboarding paths, and device-flow steps. It therefore belongs in the canonical domain layer.

#### Solution

1. Add the private `@domain/entity-device-model` package, defining the canonical domain `DeviceModelId` const map, structural string-union type, and Zod schema.
2. Replace the public `DeviceModelId` enums in `@ledgerhq/types-devices` and `@ledgerhq/devices` with equivalent const maps and string-union types.
3. Preserve member access for consumers:

```ts
const modelId: DeviceModelId = DeviceModelId.nanoX;
```

4. The public and domain identifiers are now string unions with the same members, so values are structurally compatible across the legacy/new-architecture boundary. This removes the nominal incompatibility of independently declared TypeScript enums.

#### Why the public API change is safe

`@ledgerhq/types-devices` and `@ledgerhq/devices` are legacy/deprecated APIs; external integrations built on LedgerJS should use DMK instead.

The runtime representation is preserved: `DeviceModelId.nanoX` is still `"nanoX"`, and the object has the same keys and values. `Object.values(DeviceModelId)` and member access therefore keep their previous behavior.

The type-level enum behavior does change. This PR updates all internal usages that relied on it:

- enum-member types such as `DeviceModelId.nanoX` and `Exclude<DeviceModelId, DeviceModelId.blue>` now use indexed-access types from the const map;
- model-ID collections used for membership checks are constructed as `new Set<DeviceModelId>([…])`, so every declared value is checked while `.has()` accepts the complete union;
- test defaults are explicitly declared as `DeviceModelId[]`, rather than relying on the narrow literal type inferred from a const-map member;
- records keyed by a device model ID are explicitly typed where the enum previously provided broader inference.

Internal usages do not rely on enum reverse mappings, `z.nativeEnum`, mutation, or reflection. The remaining key lookup and membership checks work with the const map because its keys and values are unchanged.

This is intentionally released as a breaking change for both public packages. Their generated documentation has been refreshed.

### ✅ Testing

- `pnpm desktop typecheck`
- `pnpm mobile typecheck`
- `pnpm --filter @ledgerhq/live-common typecheck`
- `pnpm --filter @ledgerhq/devices build`
- `pnpm --filter @ledgerhq/types-devices build`
- `pnpm --filter @ledgerhq/devices test`

### 🔗 Context

- **JIRA / GitHub issue**: N/A
- **ADR**: N/A

### DIFF
--- a/.changeset/silent-devices-union.md
+++ b/.changeset/silent-devices-union.md
@@ -1,0 +1,6 @@
+---
+"@ledgerhq/devices": major
+"@ledgerhq/types-devices": major
+---
+
+Replace the `DeviceModelId` enum with a const map and compatible string-union type.
--- a/apps/ledger-live-desktop/src/mvvm/components/DeviceActionContent/types.ts
+++ b/apps/ledger-live-desktop/src/mvvm/components/DeviceActionContent/types.ts
@@ -1,5 +1,5 @@
 import type { ReactNode } from "react";
-import type { DeviceModelId } from "@ledgerhq/types-devices";
+import { DeviceModelId } from "@ledgerhq/types-devices";
 
 /** Device-side action illustrated by the content animation. */
 export type DeviceActionContentAction = "continue" | "power-and-unlock";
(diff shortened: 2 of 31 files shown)


OlivierFreyssinet:
Closing this draft in favor of a focused domain-entity PR that leaves the public DeviceModelId enums unchanged.

PR #20766, opened 2026-08-13
ci(allure): drop tj-actions/branch-names third-party action

ypolishchuk-ledger:
### 📝 Description

The `upload-allure-report` composite action relied on the third-party `tj-actions/branch-names` action solely to obtain the current branch name. This PR removes that dependency and computes the branch name natively.

**Problem**: Depending on a third-party action for something GitHub already exposes adds supply-chain risk (the `tj-actions` org was the target of a supply-chain attack in early 2025) and an extra step, for no benefit.

**Solution**:
- Replaced `steps.branch-name.outputs.current_branch` with the native `${{ github.head_ref || github.ref_name }}`:
  - `github.head_ref` -> PR source branch on `pull_request` events.
  - `github.ref_name` -> branch short name (already without `refs/heads/`) otherwise.
- Hardened the `run` block against template injection (flagged by `zizmor`): the branch name is now passed via `env:` and referenced as a shell variable instead of being interpolated directly, and `$GITHUB_REPOSITORY` is used instead of interpolating `github.repository`.

No behavioural change to the generated report path.

### 🔗 Context

- **JIRA / GitHub issue**: _n/a - CI maintenance / dependency hardening_
- **ADR** (if any): _n/a_

### DIFF
--- a/tools/actions/composites/upload-allure-report/action.yml
+++ b/tools/actions/composites/upload-allure-report/action.yml
@@ -28,22 +28,20 @@ runs:
   using: composite
 
   steps:
-    - id: branch-name
-      uses: tj-actions/branch-names@6c999acf206f5561e19f46301bb310e9e70d8815 # v6
     - name: Set report path
       id: report-path
       shell: bash
       run: |
-        REPO_NAME=$(echo "${{ github.repository }}" | cut -d'/' -f2)
-        BRANCH="${{ steps.branch-name.outputs.current_branch }}"
+        REPO_NAME="${GITHUB_REPOSITORY#*/}"
         SUFFIX="${INPUTS_REPORT_PATH_SUFFIX:-$INPUTS_PLATFORM}"
-        if [ "${{ github.repository }}" != "LedgerHQ/ledger-live" ]; then
+        if [ "$GITHUB_REPOSITORY" != "LedgerHQ/ledger-live" ]; then
           PATH_VALUE="${REPO_NAME}-${BRANCH}-${SUFFIX}"
         else
           PATH_VALUE="${BRANCH}-${SUFFIX}"
         fi
         echo "path=${PATH_VALUE}" >> "$GITHUB_OUTPUT"
       env:
+        BRANCH: ${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


ypolishchuk-ledger:
## ✅ Real GitHub Actions runtime verification

This change (dropping `tj-actions/branch-names` in favour of the native `${{ github.head_ref || github.ref_name }}`) was validated **end-to-end on a real Ledger self-hosted runner** (`ledger-live-medium`), executing this branch's version of the `upload-allure-report` composite against the real Allure server and secrets.

### How it was tested
A temporary `on: push` workflow referenced the composite by **local path** (`uses: ./tools/actions/composites/upload-allure-report`) so the runner used **this branch's** code (the production workflows pin the action to `@develop`). It fed the action a minimal Allure result fixture and asserted a report URL was returned. The temporary workflow was then removed — the branch's net diff vs `develop` is only `action.yml`.

### Result — all steps green ✅

| Item | Value |
|---|---|
| Workflow run | [TEMP - Allure Runtime Check · run 31782985939](https://github.com/LedgerHQ/ledger-live/actions/runs/31782985939) |
| Conclusion | `success` |
| Runner | `ledger-live-medium` (self-hosted) |
| Duration | ~55s |

**The step I changed (`Set report path`) computed the branch name natively and produced the correct path:**

```
INPUTS_PLATFORM: linux-runtime-check
INPUTS_REPORT_PATH_SUFFIX: runtime-check
path: ci/allure-drop-tj-actions-branch-names-runtime-check      # <branch>-<suffix> ✅
```

This confirms `${{ github.head_ref || github.ref_name }}` resolved to the branch name (via `ref_name` on a push event) with no third-party action involved.

**The report uploaded and generated successfully, returning a real URL:**

```
report-url = https://ledger-live.allure.green.ledgerlabs.net/allure/reports/2076f5f5-2f0c-4eb7-974f-374cb6daae11/
```

### Additional validation
- Local parity test of old vs new report-path logic across 5 scenarios (main repo, fork repo, default/custom suffix, slash-containing branches) — all identical output.
- Pre-commit hooks pass on the change: `oxfmt`, `gitleaks`, `actionlint`, `zizmor` (which also caught and had me fix a template-injection risk by moving the branch value into `env:`), and `commitlint`.

lewisd5 (approved):

PR #20767, opened 2026-08-13
feat(device-model): add domain device model entity

OlivierFreyssinet:
### 📝 Description

This PR introduces the private `@domain/entity-device-model` package, establishing a canonical domain representation for Ledger hardware-wallet model IDs.

#### Why

`libs/device-intent` needs to move to `features/platform/`, but feature packages cannot depend on legacy packages under `libs/`. It currently imports `DeviceModelId` from `@ledgerhq/types-devices`, which blocks that move.

A Ledger device model is product vocabulary, not a transport implementation detail. It determines device animations and illustrations, supported capabilities, onboarding paths, and device-flow steps. The concept is shared across features and remains meaningful independently of a connection transport, so it belongs in the domain layer.

#### Solution

- Add `@domain/entity-device-model` with a canonical `DeviceModelId` const map, structural string-union type, and Zod schema.
- Provide a mock factory and schema tests covering every declared model and rejection of unknown IDs.
- Keep `@ledgerhq/types-devices` and `@ledgerhq/devices` unchanged. Their public enums remain the compatibility boundary for legacy consumers and external packages.

A later `device-intent` migration can adapt between the domain identifier and legacy enum at its explicit boundary. This avoids a public enum API change while keeping new-architecture code independent from legacy libraries.

### ✅ Testing

- `pnpm --filter @domain/entity-device-model test`
- `pnpm --filter @domain/entity-device-model typecheck`

### 🔗 Context

- **JIRA / GitHub issue**: N/A
- **ADR**: N/A

### DIFF
--- a/domain/entity/device-model/README.md
+++ b/domain/entity/device-model/README.md
@@ -1,0 +1,22 @@
+# @domain/entity-device-model
+
+> [!CAUTION]
+> **Status: UNSTABLE** — Canonical device-model vocabulary introduced for the DDD migration.
+
+Defines the canonical `DeviceModelId` schema and type used by new-architecture packages.
+
+## Why this is a domain entity
+
+A Ledger device model is central product vocabulary, not a transport implementation detail. Its
+identity drives UI choices such as device animations and illustrations, and UX behaviour such as
+available features, onboarding paths, and device-flow steps. The concept is shared by multiple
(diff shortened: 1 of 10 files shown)


OlivierFreyssinet:
Closing in favor of the Device Intent platform migration, which will use DMK identifiers directly rather than introduce a separate domain identifier.

PR #20771, opened 2026-08-13
[LWDM] feat(aleo): framework getBalance

mdomanski-ext-ledger:
### 📝 Description

This PR implements the framework's `Api.getBalance` for Aleo, replacing the "not supported" placeholder left by part 1. It combines the public on-chain balance with private (shielded) balances by decrypting the caller's unspent records, and reports per-token balances alongside the native one. Token programs are discovered from both public transaction history and private records, then classified as `arc20`/`arc21`/`arc22`/`unknown` against a newly fetched token registry (`GET /tokens`, paginated via `fetchAllTokens`). Private balances require a `provableId` + `viewKey` on the context (`resolvePrivacyContext`); record sums are capped at the scanner's `synced_up_to` height to avoid counting not-yet-indexed funds, and errors intentionally never leak the caller's `provableId`/`viewKey`.

### 🔗 Context

- https://ledgerhq.atlassian.net/browse/LIVE-34091

### DIFF
--- a/libs/coin-modules/coin-aleo/src/logic/getBalance.ts
+++ b/libs/coin-modules/coin-aleo/src/logic/getBalance.ts
@@ -1,0 +1,93 @@
+import invariant from "invariant";
+import BigNumber from "bignumber.js";
+import type { Balance } from "@ledgerhq/coin-module-framework/api/types";
+import { promiseAllBatched } from "@ledgerhq/coin-module-framework/promises";
+import { PROGRAM_ID } from "../constants";
+import { apiClient } from "../network/api";
+import {
+  fetchAccountTransactionsFromHeight,
+  fetchAllOwnedRecords,
+  fetchAllTokens,
+  getRecordScannerStatusOrThrow,
+  sumUnspentRecords,
+} from "../network/utils";
+import type { AleoContext, AleoPrivateRecord, AleoTokenType } from "../types";
+import { getPublicBalance } from "./getPublicBalance";
+import { classifyAleoTokenType, isTokenRecord, parseAmount, resolvePrivacyContext } from "./utils";
+
+export async function getBalance(context: AleoContext, address: string): Promise<Balance[]> {
+  const config = await context.config();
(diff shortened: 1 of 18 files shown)


libs/coin-modules/coin-aleo/src/logic/getBalance.ts
```diff
@@ -0,0 +1,84 @@
...
+      .map(tx => tx.program_id)
+      .filter(programId => programId !== PROGRAM_ID.CREDITS),
+  );
+  const maxSyncedBlockHeight = status.synced_up_to;
+
+  const privateNativeRecords = allUnspentRecords.filter(
+    record => record.program_name === PROGRAM_ID.CREDITS,
+  );
+  const privateTokenRecords = allUnspentRecords.filter(isTokenRecord);
+
```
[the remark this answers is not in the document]
mdomanski-ext-ledger: Tokens were already enabled on production, flag will be deprecated

libs/coin-modules/coin-aleo/src/logic/getBalance.ts
```diff
@@ -0,0 +1,93 @@
...
+    fetchAllOwnedRecords({
+      config,
+      uuid: privacyContext.provableId,
+      unspent: true,
+      // empty arrays opt out of the credits.aleo-only filter, returning records for all programs
+      programs: [],
+      functions: [],
+    }),
+    fetchAllTokens({ config }),
+  ]);
```
henri-ly, line 40: [question] Will it be easy to track which one are failing ?

henri-ly (approved):

libs/coin-modules/coin-aleo/src/logic/getBalance.ts
mdomanski-ext-ledger: All calls are based on apiClient here, so LedgerAPI4xx errors should be dumped into the logs properly like 
`422 GET /v2/testnet/program/credits.aleo/mapping/account/aleo1...`. 

We have the same logic in sync.ts already:
image

PR #14208, opened 2026-02-03
feat: add ICP Staking to desktop and mobile

lawRathod:
### ✅ Checklist

- [ ] `npx changeset` was attached.
- [ ] **Covered by automatic tests.** 
- [ ] **Impact of the changes:** 
  - ...

### 📝 Description

Adds ICP staking features for desktop and mobile

### ❓ Context

- **JIRA or GitHub link**: 

---

### 🧐 Checklist for the PR Reviewers

- **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- **Any new dependencies** have been justified and documented.
- **Performance** considerations have been taken into account. (changes have been profiled or benchmarked if necessary)

PR #20774, opened 2026-08-13
[LWDM] feat(internet_computer): add neuron management flows (LIVE-29096)

amaslakov:
### 📝 Description

Fills in the two neuron modals LIVE-29095 left as stubs, covering every neuron operation the bridge already supported. Adds the three periodic-confirmation fields to the candid decode, without which the wallet cannot tell a neuron losing voting power from a healthy one.

A rewrite, not a port — the reference branch (#14208) predates LIVE-28469's type rewrite. Three of its bugs came from comparing a variant type against a string; all three are pinned by tests.

Worth a reviewer's attention:

- **Ships dark** — `internet_computer` is absent from `stakePrograms.params.list`, which gates both entry points.
- **Two files belong to LIVE-29095, not this ticket** — `AccountBalanceSummaryFooter.tsx` and its test, already on `develop`, because Total Maturity ignored staked maturity while its tooltip promised it.
- **Maturity modulation corrected** from 0.9 to 0.95, matching the governance canister's spawn precondition.
- **Adds the ICP error translations.**
- **Device paths are unit-tested only** — signing, broadcast and the sync button have never run against hardware.
- **Four hardcoded constants are NNS governance parameters**, not protocol constants. They match today's defaults but a proposal could move them.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-29096](https://ledgerhq.atlassian.net/browse/LIVE-29096?atlOrigin=eyJpIjoiZTYwZTU3OGQ3NGJkNDRhMDgwZTk5ODE3YmEwODMxMGMiLCJwIjoiaiJ9)
- **ADR** (if any): 

[LIVE-29096]: https://ledgerhq.atlassian.net/browse/LIVE-29096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shazzzam (dismissed):

apps/ledger-live-desktop/src/renderer/families/internet_computer/ManageNeuronFlowModal/steps/StepManage.tsx
```diff
@@ -0,0 +1,331 @@
...
+      </NeuronSection>
+
+      <NeuronSection
+        title={t("internetComputer.manageNeuronFlow.manage.votingPower.title")}
+        tooltip={t("internetComputer.manageNeuronFlow.manage.votingPower.tooltip")}
+        value={
+          votingPower > 0n ? (
+            <FormattedVal val={toBigNumber(votingPower)} unit={unit} disableRounding />
+          ) : (
+            <Trans i18nKey="common.none" />
```
ishaba: Missing i18n string: `common.none` isn't defined in `static/i18n/en/app.json` (the `common` block has no `none` key), so this renders the literal string `common.none`. Reachable for any neuron with `neuronPotentialVotingPower(neuron) <= 0n` (fresh / short-delay / dissolved). Add a `common.none` key, or reuse the existing `internetComputer.manageNeuronFlow.manage.votingPower.noDissolveDelay`.

apps/ledger-live-desktop/src/renderer/families/internet_computer/ManageNeuronFlowModal/steps/StepStakeMaturity.tsx
```diff
@@ -0,0 +1,72 @@
...
+  transaction,
+  onUpdateTransaction,
+}: StepProps) => {
+  const { t } = useTranslation();
+  const neuron = neurons.find(n => n.id?.toString() === selectedNeuronId);
+  const percentage = transaction?.percentageToStake ?? "";
+
+  const onChange = useCallback(
+    (next: string) => {
+      const digits = next.replace(/\D/g, "").slice(0, 3);
```
ishaba: The percentage isn't clamped: `.slice(0, 3)` allows `999`, so the preview shows `maturity * 999 / 100` (~10×) as the amount to stake; the footer's `!!percentageToStake` also lets `"0"` through. The sibling `StepSetDissolveDelay` clamps its input — suggest clamping to `1..100` here (and/or gating `canContinue` on the range) rather than relying solely on bridge validation.

apps/ledger-live-desktop/src/renderer/families/internet_computer/neuronFlow/useNeuronActions.ts
```diff
@@ -0,0 +1,107 @@
...
+export function useNeuronActions({
+  account,
+  neuron,
+  onChangeTransaction,
+  transitionTo,
+  setLastAction,
+}: Params) {
+  const dispatch = useDispatch();
+  const bridge = useAccountBridge<Transaction>(account);
+  const neuronId = neuron.id?.toString();
```
ishaba: If `neurons` is empty, `StepManage` passes `neuron ?? neurons[0]` = `undefined`, and this line (plus `neuron.state` in the deps array) throws before `StepManage`'s `if (!neuron) return null` runs. Low-probability (neurons only change via a signed `list_neurons`), but the guard doesn't cover the empty-list case the comment describes. Consider making the hook tolerate an `undefined` neuron.

ishaba (commented):
A few commnets:

ishaba (dismissed):

Moustafa-Koterba (dismissed):

libs/coin-modules/coin-internet_computer/src/bridge/bridgeHelpers/account.ts
```diff
@@ -71,16 +82,38 @@ export const getAccountShape: GetAccountShape<ICPAccount> = async info => {
...
+ */
+export const postSync = (_initial: ICPAccount, synced: ICPAccount): ICPAccount => {
+  const operations = dedupeRetypedOperations(synced.operations as InternetComputerOperation[]);
+  // Same object unless something was actually dropped: a new one every sync would invalidate every
+  // consumer memoized on `operations`. `operationsCount` follows, since the framework counted the
+  // merged list before this pass thinned it.
+  return operations.length === synced.operations.length
+    ? synced
+    : { ...synced, operations, operationsCount: operations.length };
+};
```
lysyi3m, line 115: `postSync` collapses the duplicate `OUT`s, but nothing removes the governance operations that earlier versions stored as `FEES`. `encodeOperationId` folds the type into the id, so the new `NONE` operation can't supersede them; the index canister never returns them (the hash is an IC request id); and `dedupeRetypedOperations` only looks at `OUT`. So they stay in history with a permanently 404 explorer link, and keep contributing a phantom fee to `operationsCount` and the balance-history recompute — which is the thing the changeset says is fixed.

Could we drop them in this pass? I'd key it on something ICP-specific like `extra.methodName` rather than on `FEES` alone, so any genuine fee operation is left alone. Worth a test that a stored `FEES` op carrying `methodName` is gone after a sync.

apps/ledger-live-desktop/src/renderer/families/internet_computer/AccountHeaderManageActions.ts
```diff
@@ -30,6 +32,13 @@ const AccountHeaderManageActions: InternetComputerFamily["accountHeaderManageAct
 
   if (!icpAccount) return null;
 
+  // Mirrors the gate AccountHeaderActions puts on "Stake": a platform-app redirect means staking
+  // happens in the partner app, not here.
+  const { id: currencyId } = icpAccount.currency;
+  if (!getCanStakeUsingLedgerLive(currencyId) || getCanStakeUsingPlatformApp(currencyId)) {
+    return null;
+  }
+
```
lysyi3m: This early return drops the whole action list, not just Stake. `AccountHeaderActions` already applies this exact gate, and applies it only to the Stake key (`screens/account/AccountHeaderActions.tsx:371`). As written, if ICP leaves `stakePrograms` or gets a platform-app redirect, someone who already holds neurons also loses disburse, dissolve, hot keys and following — which the generic filter deliberately keeps available.

Can we keep the gate on the Stake action and push ManageNeurons unconditionally? Two of the new tests below encode the current behaviour, so they'd need to move with it.

apps/ledger-live-desktop/src/renderer/families/internet_computer/neuronFlow/Body.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  const error = transactionError || bridgeError;
+  // A signing failure marks the step that was signing; a bridge error belongs to the first step,
+  // since it means the transaction never became valid in the first place.
+  const failedStepIndex = () => {
+    if (transactionError) return steps.findIndex(step => step.id === signingStepId);
+    return bridgeError ? 0 : -1;
+  };
+  const errorStepIndex = failedStepIndex();
+  const errorSteps = errorStepIndex < 0 ? [] : [errorStepIndex];
```
lysyi3m: `steps.findIndex` returns an index into the full array, but `Stepper` passes the breadcrumb only `steps.filter(s => !s.excludeFromBreadcrumb)` and `Breadcrumb` compares against the index within those. `manageAction` is 8 in the full list and 2 among the four visible steps, so `errorSteps: [8]` matches nothing and the failed step is never marked. The refresh flow works only because all three of its steps are visible.

Looking up against the visible list fixes it:

```ts
const visibleSteps = useMemo(() => steps.filter(s => !s.excludeFromBreadcrumb), [steps]);
const errorStepIndex = transactionError
  ? visibleSteps.findIndex(step => step.id === signingStepId)
  : bridgeError
    ? 0
    : -1;
```

Worth landing together with the device-stub fix I mentioned in the review body — the current test asserts the stub's behaviour, which is why this didn't surface.

apps/ledger-live-desktop/src/renderer/families/internet_computer/components/NeuronList.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </Cell>
+        ))}
+      </HeaderWrapper>
+      {neurons.map(neuron => (
+        <Row
+          key={neuron.id?.toString() ?? neuron.accountIdentifier}
+          clickable={!!onRowClick}
+          onClick={onRowClick ? () => onRowClick(neuron) : undefined}
+          data-testid="icp-neuron-row"
+        >
```
lysyi3m: The row is the only way to select a neuron, and it's a styled row with an `onClick`: no role, no tab stop, no key handler, no accessible name. That makes the entry point to the whole feature unreachable by keyboard and invisible to screen readers. Every other interactive element in the PR is a real `Button`.

Either put a real button inside the row, or give the row button semantics plus Enter/Space handling:

```tsx
role={onRowClick ? "button" : undefined}
tabIndex={onRowClick ? 0 : undefined}
onKeyDown={...}
```

apps/ledger-live-desktop/src/renderer/families/internet_computer/ManageNeuronFlowModal/steps/StepSelectFollowees.tsx
```diff
@@ -0,0 +1,142 @@
...
+  );
+
+  // What gets added is the draft's digits, not the draft. Gating the button on the draft itself let
+  // a principal — which carries no digits at all — enable a click that silently did nothing.
+  const draftId = useMemo(() => draft.replace(/\D/g, ""), [draft]);
+  const canAdd = draftId !== "" && !followeesIds.includes(draftId);
+
+  const onAdd = useCallback(() => {
+    if (!canAdd) return;
+    setFollowees([...followeesIds, draftId]);
```
lysyi3m: `draftId` strips non-digits and `onAdd` then submits the stripped value, so typing `12a3` silently follows neuron `123` while the field still reads `12a3`. Following delegates voting power, so this picks a different, valid target rather than just formatting the input.

I'd validate instead of rewrite — trim, require `/^\d+$/`, and leave `canAdd` false otherwise. That keeps the button-gating the comment is after, without changing what gets submitted.

apps/ledger-live-desktop/src/renderer/families/internet_computer/ManageNeuronFlowModal/steps/StepFollowTopic.tsx
```diff
@@ -0,0 +1,96 @@
...
+            <Button
+              key={topic}
+              outline
+              disabled={controllerOnly}
+              onClick={() => onSelect(topic)}
+              data-testid={`icp-follow-topic-${topic}`}
+            >
+              <Box horizontal justifyContent="space-between" width="100%" alignItems="center">
+                <Text ff="Inter|SemiBold" fontSize={4}>
+                  {topic}
```
lysyi3m: `{topic}` prints the `KNOWN_TOPICS` key verbatim, so all 19 reach the screen as identifiers — `IcOsVersionDeployment`, `SnsAndCommunityFund`, `ApiBoundaryNodeManagement`, `Kyc`. Same values are used at `StepManage.tsx:352` and `StepSelectFollowees.tsx:88` and `:110`, where one is interpolated mid-sentence.

A `governanceTopic` block in `en/app.json` keyed by topic name would cover all four sites in one go.

libs/coin-modules/coin-internet_computer/src/common-logic/neuron.ts
```diff
@@ -115,6 +142,14 @@ export const getNeuronDissolveDurationSeconds = (
 
 // ---- voting power (Mission 70) ------------------------------------------------------------------
 
+/**
+ * Whether the dissolve delay is long enough for the canister to count the neuron's vote. Everything
+ * downstream of voting hangs off this — voting power, periodic confirmation, the decay countdown —
+ * so it is one predicate rather than the same comparison repeated at each site.
+ */
+export const neuronCanVote = (neuron: ICPNeuron): boolean =>
+  neuron.dissolveDelaySeconds >= BigInt(NNS_MINIMUM_DISSOLVE_DELAY_TO_VOTE);
```
lysyi3m, line 151: `neuronCanVote` (and `neuronPotentialVotingPower` at :202) read the snapshotted `dissolveDelaySeconds`, while the UI shows `getNeuronDissolveDurationSeconds(neuron, now)` — which for a dissolving neuron counts down from `WhenDissolvedTimestampSeconds`. Between syncs the displayed remaining delay can drop below the 6-month voting minimum while voting power still reads non-zero on the same screen.

Threading the same `nowSeconds` through and deriving eligibility from the live remaining delay would keep the two consistent. Flagging that this changes signatures the voting-power tests pin, so I'm fine with it being an immediate follow-up rather than this PR — as long as it's tracked.

lysyi3m:
@amaslakov this is follow up for a review https://github.com/LedgerHQ/ledger-live/pull/20774#pullrequestreview-5086701204 (lost somewhere during the submit, oopsie)

- Four steps use account.currency.units[0] instead of useAccountUnit(account), so amounts inside the flow disagree with the account header: StepListNeuron.tsx:43, StepManage.tsx:88, StepSplitNeuron.tsx:37, StepStakeMaturity.tsx:79.

- A zero or negative dissolve-delay increment reports `ICPDissolveDelayLTMin`, which renders "requires a dissolve delay of at least 1 days" — wrong subject, and 1 rather than the real 7-day floor. Its own error plus plural forms would fix both.

- Three tests I don't think hold up. The GenericStepConnectDevice stub only calls onTransactionError, where the real component also transitions to confirmation, so the test asserts we stay on the device step. The integration fixture broadcasts type: "FEES", which this PR stops emitting, so the NONE branch of applyNeuronOperation never runs. And the dissolve-delay test types "30" against a fixed transaction prop, reads the first change event and asserts 3 days — the same file already documents the fireEvent.change workaround for StepStakeMaturity.

- jest.restoreAllMocks() in StepManage.test.tsx:108 also tears down the jest-setup globals; a scoped mockRestore() on the Date.now spy is enough. expect(container).toBeInTheDocument() in the integration test can't fail either way.
- 
- redactPrincipals matches any run of three or more hyphenated five-character words, so a reason like stake-delay-value-limit collapses to <hidden>. Validating candidates with Principal.fromText would keep real text intact.

- decidingVotingPower is decoded, persisted and optimistically patched, but nothing reads it. Worth a comment saying that's deliberate, or dropping the field.

also, I'd fold all 6 changesets into single one and re-worked the changes list

ishaba (dismissed):

lysyi3m (dismissed):

apps/ledger-live-desktop/src/renderer/families/internet_computer/__tests__/errorTranslations.test.ts
Moustafa-Koterba, line 1: I am not sure that we have to test the translations on the repo, why this file ?
Translation team are responsible of it actually

Moustafa-Koterba (approved):

apps/ledger-live-desktop/src/renderer/families/internet_computer/__tests__/errorTranslations.test.ts
amaslakov: This one is not testing localization, it's checking our hardcoded English dictionary still covers the ICP error classes, because a missing key silently renders the class name and the repo's own key checker excludes the errors namespace. It's probably the name of the file that is misleading

lysyi3m (approved):

PR #20787, opened 2026-08-14
fix(e2e/desktop): fall back to max sellable balance in Sell test

kentoforik:
## Summary
- `getMinimumSellAmount()` computes a provider-driven minimum sell amount (via `maxOfMin` from the Buy/Sell `cryptoLimitations` API), independent of the shared test wallet's real-time balance.
- That wallet (e.g. `TokenAccount.ETH_USDT_1`) is a real, live-blockchain account shared and concurrently drawn down by other Buy/Sell/Swap e2e tests in the same CI run, so the computed minimum can end up exceeding the account's actual balance.
- When that happens, the app shows "Not enough balance" instead of the payment/provider selector, and the Sell test previously just timed out waiting for `payment_selector`.
- `setAmountToPay()` now detects that specific balance-insufficient state (via the `amount-input-section-error` testid) and falls back to tapping `max` to use the account's actual live spendable balance, returning the amount actually used so downstream URL verification checks against the real value.

Investigated from CI run: https://github.com/LedgerHQ/ledger-live/actions/runs/31761092448 (`[USDT (Ethereum)] - Sell` failure — entered amount 31.707599 USDT vs. live balance 24.088397 USDT).

Mobile e2e was checked too: its Sell flow taps a `"75%"` balance-percentage button rather than typing the computed minimum, so it isn't exposed to this failure mode — no changes needed there.

## Test plan
- [x] [CI: Desktop E2E `buySell.spec.ts` Sell suite passes for BTC / ETH / USDT (Ethereum)](https://github.com/LedgerHQ/ledger-live/actions/runs/32013246766)
- [x] Manually verify `setAmountToPay` still behaves normally on the happy path (sufficient balance) for both Buy and Sell

### DIFF
--- a/e2e/desktop/tests/page/buyAndSell.page.ts
+++ b/e2e/desktop/tests/page/buyAndSell.page.ts
@@ -248,18 +248,25 @@ export class BuyAndSellPage extends WebViewAppPage {
     );
     await this.verifyElementIsNotEnabled(this.formCta);
 
-    try {
-      await this.verifyElementIsVisible(this.paymentSelector);
-    } catch (error) {
-      if (operation !== OperationType.Sell) throw error;
-      await this.verifyElementIsVisible(this.amountInputError);
+    let balanceTooLow = false;
+    if (operation === OperationType.Sell) {
+      try {
+        await this.verifyElementIsVisible(this.amountInputError, 2000);
+        balanceTooLow = true;
+      } catch {
+        // No error shown — balance covers the requested amount.
+      }
+    }
+    if (balanceTooLow) {
       await this.clickElement(this.maxAmountButton);
-      amount = await (await this.getWebViewElementByTestId(this.amountInputSection)).inputValue();
-      await this.verifyElementIsVisible(this.paymentSelector);
     }
 
+    await this.verifyElementIsVisible(this.paymentSelector);
+    const finalAmount = await (
+      await this.getWebViewElementByTestId(this.amountInputSection)
+    ).inputValue();
     await this.verifyElementIsVisible(this.providersList);
-    return amount;
+    return finalAmount;
   }
 
   @step("Select provider quote")
--- a/e2e/desktop/tests/page/buyAndSell.page.ts
+++ b/e2e/desktop/tests/page/buyAndSell.page.ts
@@ -42,6 +42,8 @@ export class BuyAndSellPage extends WebViewAppPage {
   private fiatDrawerInput = "fiat-drawer-search-input";
   private saveRegionFiatOptionsSelector = "save-region-and-fiat-options";
   private showMoreQuotes = "SHOW MORE QUOTES";
+  private amountInputError = "amount-input-section-error";
+  private maxAmountButton = "max";
   private providerTitleCssSelector =
     "[data-testid^='provider_title_'][data-testid$='_title_container']";
 
@@ -235,8 +237,9 @@ export class BuyAndSellPage extends WebViewAppPage {
     await this.verifyElementText(this.infoBox, "Sell securely with Ledger");
   }
 
+  // Falls back to the max sellable balance if the requested amount exceeds it.
   @step("Enter amount to pay $0")
-  async setAmountToPay(amount: string, operation: string) {
+  async setAmountToPay(amount: string, operation: string): Promise<string> {
     await this.setValue(this.amountInputSection, amount);
 
     await this.verifyElementText(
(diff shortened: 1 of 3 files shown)


e2e/desktop/tests/page/buyAndSell.page.ts
```diff
@@ -235,17 +237,29 @@ export class BuyAndSellPage extends WebViewAppPage {
...
 
     await this.verifyElementText(
       this.formCta,
       operation === OperationType.Buy ? "Select quote to continue" : "Set an amount to get quotes",
     );
     await this.verifyElementIsNotEnabled(this.formCta);
-    await this.verifyElementIsVisible(this.paymentSelector);
+
+    try {
+      await this.verifyElementIsVisible(this.paymentSelector);
```
ypolishchuk-ledger: `verifyElementIsVisible` uses the global Playwright expect timeout, which is 41 000ms. In the exact scenario this PR targets - the amount exceeds the balance - `paymentSelector` never appears until max is clicked, so the try always burns the full 41s before failing into the catch. When the clamp is the common case, every Sell test pays that penalty.
Suggestion: branch on the fast-appearing error state instead of on a long-timeout failure. Reuse the existing bounded-probe pattern (`isTextVisible` uses a 1000ms timeout, webViewApp.page.ts:171-181) - e.g. add an isElementVisible(testId, timeout) helper and do:
```
if (operation === OperationType.Sell && (await this.isElementVisible(this.amountInputError, 2000))) {
  await this.clickElement(this.maxAmountButton);
  ...
}
await this.verifyElementIsVisible(this.paymentSelector);
```

e2e/desktop/tests/page/buyAndSell.page.ts
```diff
@@ -235,17 +237,29 @@ export class BuyAndSellPage extends WebViewAppPage {
...
+    } catch (error) {
+      if (operation !== OperationType.Sell) throw error;
+      await this.verifyElementIsVisible(this.amountInputError);
+      await this.clickElement(this.maxAmountButton);
+      amount = await (await this.getWebViewElementByTestId(this.amountInputSection)).inputValue();
```
ypolishchuk-ledger: The input value is read immediately after clicking max and before the UI settles (paymentSelector visible). If the app debounces/asynchronously updates the field, inputValue() can return a stale or interim value. That value flows into the URL assertion (buySell.amount → cryptoAmount query param, verifyProviderUrl), so a stale read can cause false failures or silently weaken the assertion.
Suggestion: read inputValue() after paymentSelector is visible (state settled), and ideally assert the value actually changed from the rejected input:
```
await this.verifyElementIsVisible(this.paymentSelector);
amount = await (await this.getWebViewElementByTestId(this.amountInputSection)).inputValue();
```

e2e/desktop/tests/page/buyAndSell.page.ts
kentoforik: Good catch — confirmed `expect.timeout: 41000` in `playwright.config.ts:9`, so the clamped path was indeed burning the full 41s every time.

Fixed in 96ce3ba6a79: the code now branches on the fast-appearing error state (`amount-input-section-error`, 2s timeout) and verifies `paymentSelector` unconditionally afterwards, so the clamped case resolves in ≤2s instead of 41s.

One deliberate deviation from your snippet: I did not add the shared `isElementVisible(testId, timeout)` helper. Instead `verifyElementIsVisible` gained an optional `timeout` param and the probe is an inlined try/catch with a `balanceTooLow` flag — this was a review preference on our side to avoid adding a generic wrapper around an existing base-class method. Behaviour matches your proposal; only the shape differs. Happy to extract the helper if you feel strongly.

Two trade-offs worth naming explicitly:
- The happy path (balance sufficient) now pays a flat 2s waiting for an error that never appears — inherent to probing for absence, and the same cost your snippet carries.
- Genuine unrelated failures still take 2s + 41s before failing, which I think is correct: a real flake should get the full timeout rather than failing fast.

e2e/desktop/tests/page/buyAndSell.page.ts
kentoforik: Agreed on both parts — fixed across 96ce3ba6a79 and c6e6fff274d.

1. `inputValue()` is now read only after `verifyElementIsVisible(this.paymentSelector)` confirms the UI has settled, exactly as you proposed. I made the read unconditional rather than only reassigning inside the fallback branch: on the happy path it simply reads back the value that was typed, which keeps the flow linear and means the returned amount is always whatever the UI actually holds.

2. Added the changed-value assertion you suggested:
```ts
if (balanceTooLow) {
  expect(Number(finalAmount)).toBeLessThan(Number(amount));
}
```
I went with numeric `<` rather than `!==` because the fallback only runs when balance < requested, so the clamped value must be strictly smaller — a stronger guarantee than merely changed, and immune to formatting differences (`"24.0"` vs `"24"`). Scoped to the fallback branch on purpose: on the happy path the value must stay equal to what was typed, so a blanket changed assertion would be wrong there.

This closes the failure mode you flagged: a Max click that silently leaves the rejected amount in the field now fails loudly instead of feeding a stale value into the `cryptoAmount` URL assertion.

ypolishchuk-ledger (approved):

PR #20797, opened 2026-08-14
chore(release): prune changelogs when cutting a release branch

lewisd5:
Release commits go through createCommitOnBranch, which sends whole file contents rather than diffs. Entries had accumulated to 500-700 per changelog, so a release bumping ~146 packages sent ~28 MB base64 in one request and intermittently returned 502 Bad Gateway.

Prune to the newest 20 entries when release-create cuts the release branch from develop: 22.4 MB -> 4.1 MB across 203 changelogs, with no single file above ~440 KB base64.

Old entries are deleted rather than archived, since an archive file would be re-sent in full whenever it is appended to. History remains in git and in the GitHub release for each version.

This is intentionally only done on release and not hotfix to prevent back-merge issues.

Test run: https://github.com/LedgerHQ/ledger-live/actions/runs/32020372065/job/95358613216
Ticket: https://ledgerhq.atlassian.net/browse/LIVE-36075

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15822,6 +15822,21 @@ importers:
 
   tools/pnpm-utils: {}
 
+  tools/prune-changelogs:
+    devDependencies:
+      '@manypkg/get-packages':
+        specifier: 1.1.3
+        version: 1.1.3
+      '@types/node':
+        specifier: 'catalog:'
+        version: 24.12.0
+      prettier:
+        specifier: 2.8.8
+        version: 2.8.8
+      typescript:
+        specifier: 'catalog:'
+        version: '@typescript/typescript6@6.0.2'
+
 packages:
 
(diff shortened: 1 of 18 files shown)


pnpm-lock.yaml
```diff
@@ -67855,6 +67815,54 @@ snapshots:
       - supports-color
       - utf-8-validate
 
+  metro@0.83.3(metro-transform-worker@0.83.3):
```
dgreen-ledger, line 67818: Not a blocker but is this just a PNPM quirk in the snapshot. As this doesnt seem to be related to the new package deps

dgreen-ledger (approved):

pnpm-lock.yaml
lewisd5: Indeed it is

PR #20821, opened 2026-08-15
[LWDM] feat(coin-stacks): add pox-5 staking support to coin tester

pawell24:
### 📝 Description

Adds a Clarinet-devnet coin-tester for `@ledgerhq/coin-stacks`: native STX + SIP-010 transfers on
the legacy bridge, plus a pox-5 delegate/undelegate staking scenario on the generic-adapter
(CoinModuleApi) bridge. Includes the devnet reliability fixes needed to reach pox-5 locally on a
Clarinet devnet (epoch 4.0 config, automated patched-`clarinet`-binary build, mining-reliability
workaround, signer-manager setup for pox-5's stake/unstake).

CI-only devnet-boot failure fixed: `bitcoind` crashed on `public-ledgerhq-shared-small` runners with
`Permission Denied` writing its data directory, caused by Clarinet bind-mounting a host directory
(owned by the CI runner's user) into a container running as a different uid — never reproduces
locally since Docker Desktop for Mac doesn't enforce the same host/container uid match. Fixed with a
`chmod` right after Clarinet creates that directory
(`docker/clarinet/bitcoin-node-datadir-permissions.patch`); `bitcoin-node-no-autoremove.patch` and a
fast-fail check in `waitUntilReady` were kept as the diagnostic path that surfaced it.

Also includes a few small, review-driven hardening fixes in `coin-stacks` (network-name validation
in address derivation, a duplicated base-URL helper, correct process-exit semantics in the
coin-tester's signal handlers).

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34178](https://ledgerhq.atlassian.net/browse/LIVE-34178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

[LIVE-34178]: https://ledgerhq.atlassian.net/browse/LIVE-34178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/coin-modules/coin-stacks/src/bridge/synchronization.ts
```diff
@@ -176,7 +181,16 @@ export const getAccountShape: GetAccountShape = async info => {
...
+  // Additive only: the legacy bridge has only ever registered the mainnet "stacks" currency, so
+  // this always defaulted to Mainnet with no way to override it. `API_STACKS_NETWORK` lets a
+  // devnet/testnet consumer (e.g. the coin-tester) derive the correctly-versioned address instead
+  // of a mainnet one that was never funded on that chain. v7's `getAddressFromPublicKey` takes the
+  // network name directly (no more `TransactionVersion` enum), so this env var's value passes
+  // straight through.
+  const address = getAddressFromPublicKey(
+    pubKey,
+    getEnv("API_STACKS_NETWORK") as StacksNetworkName,
+  );
```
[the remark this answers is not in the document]
pawell24: Verified: `getAddressFromPublicKey`'s own implementation in `@stacks/transactions@7.6.0` has `network = 'mainnet'` as a JS default parameter, so this was never a live bug (an explicit `undefined` still resolves to mainnet). That said, the cast was hiding this reliance on a downstream default, so pushed a fix: validates the value against `@stacks/network`'s own `StacksNetworks` list and falls back to `"mainnet"` for anything unrecognized (including `undefined`), same pattern applied to the identical case in `buildUnsignedTx.ts`.

libs/coin-modules/coin-stacks/jest.config.js
```diff
@@ -1,12 +1,4 @@
...
-  collectCoverageFrom: [
-    "src/**/*.ts",
-    "!src/**/*.test.ts",
-    "!src/**/*.spec.ts",
-    "!src/test/**/*.ts",
-  ],
-  coverageReporters: ["json", ["lcov", { file: "lcov.info", projectRoot: "../../../" }], "text"],
+const sharedConfig = {
   testEnvironment: "node",
   transform: {
```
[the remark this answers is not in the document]
pawell24: Intentional: `.integ.test.ts` runs via a separate `jest.integ.config.js` (`testRegex: \".integ.test.ts$\"`), invoked by the `test-integ` script and dedicated CI workflows (`test-coin-modules-integ.yml`, `test-integration-pr.yml`, ...) — same pattern already used by `coin-vechain`/`coin-near`. Not silently skipped, just run on a different cadence than the default unit-test run.

libs/coin-modules/coin-stacks/jest.config.js
```diff
@@ -19,11 +11,41 @@ module.exports = {
     ],
   },
   testPathIgnorePatterns: ["lib/", "lib-es/", ".integration.test.ts", ".integ.test.ts"],
+};
+
+// Two projects: `unit` runs with @ledgerhq/disable-network-setup (nock blocks all net connect),
+// while `msw` runs the `.msw.test.ts` suites WITHOUT it -- nock and MSW both build on
+// @mswjs/interceptors, so they cannot share the same run. Mirrors coin-kaspa/coin-vechain.
+module.exports = {
```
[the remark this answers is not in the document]
pawell24: Intentional: `.integ.test.ts` runs via a separate `jest.integ.config.js` (`testRegex: \".integ.test.ts$\"`), invoked by the `test-integ` script and dedicated CI workflows (`test-coin-modules-integ.yml`, `test-integration-pr.yml`, ...) — same pattern already used by `coin-vechain`/`coin-near`. Not silently skipped, just run on a different cadence than the default unit-test run.

libs/coin-modules/coin-stacks/jest.config.js
```diff
@@ -19,11 +11,41 @@ module.exports = {
...
+      ],
+    },
+    {
+      ...sharedConfig,
+      displayName: "msw",
+      testMatch: ["**/*.msw.test.ts"],
+      setupFiles: ["@ledgerhq/wallet-framework-test-setup", "./src/test/msw-setup.ts"],
+    },
+  ],
 };
```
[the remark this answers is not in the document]
pawell24: Intentional: `.integ.test.ts` runs via a separate `jest.integ.config.js` (`testRegex: \".integ.test.ts$\"`), invoked by the `test-integ` script and dedicated CI workflows (`test-coin-modules-integ.yml`, `test-integration-pr.yml`, ...) — same pattern already used by `coin-vechain`/`coin-near`. Not silently skipped, just run on a different cadence than the default unit-test run.

libs/coin-modules/coin-stacks/src/bridge/synchronization.ts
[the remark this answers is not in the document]
pawell24: Same issue as the later duplicate comment on this line — see that reply for the verification and fix (validated against `@stacks/network`'s own `StacksNetworks` list, falls back to `"mainnet"` for anything unrecognized, including `undefined`).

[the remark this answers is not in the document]
pawell24: Same issue as the later duplicate comment on this line — see that reply for the verification and fix (validated against `@stacks/network`'s own `StacksNetworks` list, falls back to `"mainnet"` for anything unrecognized, including `undefined`).

libs/coin-modules/coin-stacks/src/config.ts
```diff
@@ -11,16 +11,7 @@ export type StacksCoinConfig = () => CurrencyConfig & {
...
-
-export const getCoinConfig = (): ReturnType<StacksCoinConfig> => {
-  if (!coinConfig?.()) {
-    throw new Error("Stacks module config not set");
-  }
-
-  return coinConfig();
-};
+/** The {@link Context} threaded through the coin-stacks API layer (ADR-019). */
+export type StacksContext = Context<StacksCurrencyConfig>;
```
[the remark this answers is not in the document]
pawell24: `setCoinConfig`/`getCoinConfig` are present in the current version of `config.ts` (restored to match the `coin-vechain` pattern) — this was already addressed earlier in this PR's history.

libs/coin-modules/coin-stacks/src/bridge/signOperation.test.ts
```diff
@@ -0,0 +1,193 @@
+import { FeeNotLoaded } from "@ledgerhq/ledger-wallet-framework/errors";
+import type { SignerContext } from "@ledgerhq/ledger-wallet-framework/signer";
+import type { Account } from "@ledgerhq/types-live";
+import { AnchorMode, deserializeTransaction } from "@stacks/transactions";
+import BigNumber from "bignumber.js";
+import { InvalidNonce } from "../errors";
+import type { StacksSigner, Transaction } from "../types";
+import { buildSignOperation } from "./signOperation";
+import { createTransaction } from "./utils/transactions";
+import type { StacksNetwork } from "../network/api";
```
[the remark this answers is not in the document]
pawell24: Verified via `tsc --noEmit`: this is valid, intentional TypeScript — `typeof StacksNetwork` here is a type-level query, erased at compile time, which is exactly the case `import type` supports even though `StacksNetwork` is a `const` at runtime. Confirmed zero compile errors on this file.

libs/coin-modules/coin-stacks/src/bridge/signOperation.test.ts
```diff
@@ -0,0 +1,193 @@
...
+  subAccounts: undefined,
+} as unknown as Account;
+
+const mockTransaction: Transaction = {
+  family: "stacks",
+  amount: new BigNumber("1000000"),
+  recipient: "SP34SVHFFP532M35DHWTQJKJJR2DRGS7T5XEXQ0M0",
+  fee: new BigNumber("180"),
+  nonce: new BigNumber("3"),
+  network: "mainnet" as keyof typeof StacksNetwork,
```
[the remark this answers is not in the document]
pawell24: Verified via `tsc --noEmit`: this is valid, intentional TypeScript — `typeof StacksNetwork` here is a type-level query, erased at compile time, which is exactly the case `import type` supports even though `StacksNetwork` is a `const` at runtime. Confirmed zero compile errors on this file.

.gitleaks.toml
```diff
@@ -31,6 +37,13 @@ id = "generic-api-key"
         '''libs/.+/.+\.test\.ts\.snap''',
     ]
 
+    [[rules.allowlists]]
+    description = "Ignore Clarinet's own public, deterministic devnet dev account keys (Devnet.toml comments and the matching hex private keys in fixtures.ts)."
+    paths = [
+        '''libs/coin-tester-modules/coin-tester-stacks/settings/Devnet\.toml''',
+        '''libs/coin-tester-modules/coin-tester-stacks/src/fixtures\.ts''',
+    ]
```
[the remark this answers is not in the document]
pawell24: Fixed: narrowed both allowlist entries from path-only to `regexTarget = "match"` + exact-value regexes (the specific known mnemonics/private keys), matching the existing `tokenId`-regex precedent above them in this same file. Verified locally: the real Clarinet devnet fixtures produce zero gitleaks findings, while a newly-added, unrelated high-entropy fake key injected into the same file is still caught.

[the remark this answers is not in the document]
pawell24: Fixed: narrowed both allowlist entries from path-only to `regexTarget = "match"` + exact-value regexes (the specific known mnemonics/private keys), matching the existing `tokenId`-regex precedent above them in this same file. Verified locally: the real Clarinet devnet fixtures produce zero gitleaks findings, while a newly-added, unrelated high-entropy fake key injected into the same file is still caught.

libs/coin-modules/coin-stacks/src/bridge/signOperation.test.ts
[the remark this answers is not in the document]
pawell24: Verified via `tsc --noEmit` (coin-stacks typecheck): no error. `import type` blocks using the identifier as a runtime value, but `typeof StacksNetwork` here is a pure type-position query, which TS resolves from the source module's declared type regardless of the type-only import. Standard, intentional pattern for borrowing a const's shape without a runtime import. Not applicable.

libs/coin-modules/coin-stacks/src/bridge/signOperation.test.ts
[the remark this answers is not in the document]
pawell24: Verified via `tsc --noEmit` (coin-stacks typecheck): no error. `import type` blocks using the identifier as a runtime value, but `typeof StacksNetwork` here is a pure type-position query, which TS resolves from the source module's declared type regardless of the type-only import. Standard, intentional pattern for borrowing a const's shape without a runtime import. Not applicable.

libs/coin-modules/coin-stacks/src/logic/listOperations.ts
```diff
@@ -34,7 +34,12 @@ function toOperation(params: {
     recipients: params.recipients,
     value: params.value,
     asset: params.asset,
-    details: params.memo ? { memo: params.memo } : undefined,
+    // `ledgerOpType` is required by `generic-coin-framework/buildSubAccounts.ts`, which rebuilds
+    // each token sub-account operation's own `type` from this field instead of the top-level one
+    // above (`op.extra?.ledgerOpType`, generic-coin-framework/utils.ts) -- without it every SIP-010
+    // sub-account operation's type is silently `undefined`. Same convention as coin-vechain/
+    // coin-multiversx's own `listOperations.ts`.
+    details: { ledgerOpType: params.type, ...(params.memo ? { memo: params.memo } : {}) },
```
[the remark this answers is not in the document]
pawell24: This is a false positive: `ledgerOpType` in `details` is the correct side of this contract, not a mismatch.

`adaptCoreOperationToLiveOperation` (`generic-coin-framework/utils.ts`) is the layer that converts `CoreOperation` (what coin modules like this one return) into the final `Operation` the rest of the app reads. It explicitly does:

```ts
if (op.details?.ledgerOpType !== undefined) {
  extra.ledgerOpType = op.details.ledgerOpType as string;
}
```

So it reads `ledgerOpType` from `details` and is itself responsible for promoting it into `extra`. `op.extra?.ledgerOpType` (what `buildSubAccounts.ts` reads) is the *output* of that conversion, not something a coin module writes directly. Moving it into `extra` here would skip that conversion entirely: `adaptCoreOperationToLiveOperation` builds its `extra` object from scratch from `details` and other typed fields, it does not merge in a pre-existing `extra` from the input, so the value would be silently dropped and we would be back to the original undefined-type bug.

`details` is also not memo-only by contract, the same function reads several other family-supplied fields off it (`assetAmount`, `assetSenders`, `assetRecipients`, etc.), so adding `ledgerOpType` alongside `memo` does not change its meaning for any consumer.

Leaving as is.

cted-ledger (commented):
## Code Review

Reviewed the full diff (47 files, +2800/-49) against `develop`: the production changes in `libs/coin-modules/coin-stacks/`, the shared-framework changes in `libs/ledger-live-common/src/bridge/generic-coin-framework/`, the env/gitleaks/CI config, and the new `libs/coin-tester-modules/coin-tester-stacks/` package.

The coin-tester itself is good work: seriously documented, every workaround justified by an actual observation, and it follows the `coin-tester-vechain` / `coin-tester-casper` pattern. **My concern is elsewhere.** Under a title that says "add ... to coin tester", this PR carries two production behaviour changes — one in the Stacks fee path, one in the generic framework shared by EVM / XRP / Stellar / Tezos / Hypercore — with no changeset and no mention in the *Changes* section of the description. Green CI does not cover either of them: the new tests in `synchronization.test.ts` `jest.mock` the whole of `@stacks/transactions`, and nothing exercises the `prepareTransaction` -> `onStatus` -> `prepareTransaction` cycle.

Findings are inline, following `.agents/agents/code-reviewer.md` (severity + confidence, only >= 80 reported). Two more that have no single line to anchor to:

### 🟡 Missing changeset for production behaviour changes (confidence: 90)

- **Rule**: `CONTRIBUTING.md` § Changelogs — "A changeset is **required** for any user-facing change or library API modification"
- **Issue**: no `.changeset/` file in the diff. This PR changes the runtime behaviour of `@ledgerhq/coin-stacks` (fee computation, address derivation, `details` shape) and of `@ledgerhq/live-common` (token sub-accounts for every generic-framework family). Both are user-facing.
- **Fix**: `pnpm changeset` — a patch on `@ledgerhq/coin-stacks` and on `@ledgerhq/live-common`, describing the token sub-account effect rather than the coin-tester.

### 🟢 PR scope vs. title and description (confidence: 80)

- **Issue**: the `generic-coin-framework/` vanished-token fix and the case-insensitive matching touch code shared by EVM / XRP / Stellar / Tezos / Hypercore (`genericCoinFrameworkFamilies.json`), and deserve their own CODEOWNERS reviewers and their own rollback blast radius. 47 files is well past what a generic-framework reviewer will go looking for in a PR titled "coin tester". The linked ticket (LIVE-34178, "[STACKS] Coin Tester", 5 SP) has no description and no acceptance criteria to scope it against.
- **Fix**: split the `generic-coin-framework/` changes into their own PR — the vanished-token fix is a real user-visible bug fix, including on EVM, and deserves to be visible as one — or at minimum spell them out in *Changes* with their cross-family scope.

---

### Checked and dismissed — no action needed

- **A hypothesis I disproved.** I first suspected `getAddressFromPublicKey(pubKey, "mainnet")` would throw, since v6 expects a numeric `TransactionVersion` (`Unexpected txVersion`, reproduced locally). It is **void**: `develop` already declares `@stacks/transactions@^7.6.0` for `coin-stacks`, and v7 takes the network name directly. Only a stale `node_modules` said otherwise.
- **`.gitleaks.toml`**: all four mnemonics and all four private keys are genuinely Clarinet's public deterministic devnet defaults — found in several public `hirosystems` repos (`stacks-subnets/core-contracts/settings/Devnet.toml`, `smart-wallet`). The allowlists match on the exact value **and** on the path, so a real secret added to those same files would still be caught. Correct as written.
- **`resolveAmount`** (`buildUnsignedTx.ts`): the pre-resolved-amount short-circuit only fires for token sweeps in practice — for a native sweep the framework leaves `amount` at 0 (`generic-coin-framework/prepareTransaction.ts`, which only pre-resolves when `subAccountId` is set), so the fee-deducting path is still taken. It also removes a potential double fee deduction at craft time. Correct.
- **`listOperations.ts`**: `details.ledgerOpType` is a framework-known key, stripped from the parent account's `extra` (`generic-coin-framework/utils.ts`), and it is the convention `coin-vechain` / `coin-multiversx` already follow. No side effect.
- **`StacksNetwork.devnet`**: a type-level widening only; a real account's `network` comes from the `createTransaction` default. Not reachable in production.

### Not covered

I did not audit in depth: `src/devnet.ts` (325 l.), `src/scenarii/stacks.ts` (313 l.), `src/signerManager.ts` (154 l.), `src/fixtures.ts`, the four `docker/clarinet/*.patch` files and the `Dockerfile`, `settings/Devnet.toml`, `Clarinet.toml`, and the two `.clar` contracts. I did read `scripts/bitcoin-miner.js` (sound, nothing to report) and `src/helpers.ts` / `src/indexer.ts` (consistent with the `coin-tester-vechain` pattern).

One open question rather than a finding: the `Coin Tester - stacks` CI job takes **44 min**, against 3-12 min for the other coin-testers. `bitcoin-miner.js`'s `BLOCK_TIME_MS = 10_000` is the documented lever — is that runtime one we are happy to pay on every affected PR?

libs/coin-modules/coin-stacks/src/bridge/prepareTransaction.ts
```diff
@@ -40,15 +40,22 @@ export const prepareTransaction: AccountBridge<Transaction>["prepareTransaction"
...
-      payload: serializePayload(tx.payload),
-      estimatedLength: estimateTransactionByteLength(tx),
-      network,
-      client: { baseUrl: getStacksBaseUrl() },
-    });
+    // A pre-set positive fee (e.g. a devnet/testnet caller working around a node whose fee
+    // estimator has no historical data yet for this payload shape) skips the network call
+    // entirely — additive only, real callers never pre-set `fee` before this step.
+    if (transaction.fee && transaction.fee.gt(0)) {
+      patch.fee = transaction.fee;
```
cted-ledger: ### 🔴 Critical — the pre-set-fee short-circuit freezes fees in production from the 2nd call on (confidence: 90)

- **Rule**: bug — the comment "real callers never pre-set `fee` before this step" only holds for the *first* call.
- **Issue**: `prepareTransaction` returns `updateTransaction(transaction, patch)` with `patch.fee` set (L64). `useBridgeTransaction` feeds that result straight back as the new state — `dispatch({ type: "onStatus", transaction: preparedTransaction })` (`libs/ledger-live-common/src/bridge/useBridgeTransaction.ts:355-358`), and the reducer does `transaction: action.transaction` (`:203-210`). Stacks has no custom `updateTransaction` (`bridge/index.ts:49` reuses the generic helper, a plain merge) and the Stacks UI exposes no fee field (memo only — `apps/ledger-live-desktop/src/renderer/families/stacks/MemoValueField.tsx`). So the **only** way `transaction.fee` can be positive on input is that a previous `prepareTransaction` set it — and this branch then fires on every subsequent call, for every real user.

  Concrete failure: the user enters a recipient -> fees are estimated against an `amount = 0` payload. They then type the amount, add a memo, or switch to a SIP-010 sub-account -> the payload and `estimatedLength` change (a SIP-010 `contract_call` is nothing like a native `token_transfer`), but `fetchFeeEstimateTransaction` is never called again. The fee stays pinned to that first estimate. `useAllAmount` then computes `spendableBalance.minus(patch.fee)` from the stale value (L63-67) — underpaid transaction, rejected by the mempool or stuck.

  The added test (`prepareTransaction.test.ts:120-131`) hand-builds a transaction with `fee` already set, so it exercises the devnet case and never the re-invocation cycle.
- **Fix**: don't put a devnet workaround on the production path — gate it on an explicit signal instead of on the presence of `fee`:
  ```ts
  // coin-tester: setEnv("API_STACKS_SKIP_FEE_ESTIMATE", true)
  if (getEnv("API_STACKS_SKIP_FEE_ESTIMATE") && transaction.fee?.gt(0)) {
    patch.fee = transaction.fee;
  } else {
    /* ... network estimate, unchanged ... */
  }
  ```
  Alternatively, let the coin-tester carry an `AccountBridge` that overrides `prepareTransaction`, the way `adaptLegacyBridge` already does for everything else (`coin-tester-stacks/src/helpers.ts:30-105`).

libs/ledger-live-common/src/bridge/generic-coin-framework/getAccountShape.ts
```diff
@@ -423,7 +423,41 @@ export function genericGetAccountShape(network: string, kind: string): GetAccoun
...
+            !("assetReference" in asset) ||
+            !asset.assetReference ||
+            freshAssetReferences.has(asset.assetReference.toLowerCase())
+          ) {
+            return [];
+          }
+          return [{ value: 0n, asset }];
+        })
+      : [];
+    const allTokenAssetsBalances = [...freshTokenAssetsBalances, ...vanishedTokenBalances];
```
cted-ledger, line 474: ### 🟡 Warning — this turns "absent from the balance response" into an authoritative zero, for every generic-framework family (confidence: 85)

- **Rule**: bug / blast radius — `genericCoinFrameworkFamilies.json` has `evm: true`, plus `xrp`, `stellar`, `tezos`, `hypercore`.
- **Issue**: the bug being fixed is real — `mergeSubAccounts` keeps old sub-accounts that are absent from the new list **with their stale balance** (`buildSubAccounts.ts:127-152`: only tokens present in `newSubAccounts` get `balance`/`spendableBalance` rewritten), so a fully swept token keeps its pre-sweep balance forever. But the fix inverts a semantic: until now "absent from the response" meant *no information, keep the last known value*; it now means *authoritative zero*, propagated through `mergeSubAccounts`, which overwrites `balance` and `spendableBalance`.

  If a degraded indexer response returns 200 with the native balance but an empty or truncated token list, the user's entire token portfolio displays as 0 — on EVM, that is Ethereum, Polygon, BSC, Arbitrum. Before this change the same response left the last known balances in place. It self-heals on the next successful sync, but it is highly visible and alarming in the meantime. I could not prove a partial-200 response exists on the EVM path — that is the one uncertainty here; the mechanism itself is certain.

  Separately: `bridgeApi.getAssetFromToken` is called with no `try`/`catch` for every sub-account, in the sync critical path. A family implementation that throws (there are six: `cardano`, `celo`, `evm`, `multiversx`, `solana`, `stacks`) would fail the whole `getAccountShape`.
- **Fix**: only inject the zero entry when the response carries at least one token entry (evidence the list is actually populated), and isolate the hook call:
  ```ts
  const balanceListLooksComplete = freshTokenAssetsBalances.length > 0;
  const vanishedTokenBalances: Balance[] = getAssetFromToken && balanceListLooksComplete
    ? (initialAccount?.subAccounts ?? []).flatMap(subAccount => {
        let asset: AssetInfo | undefined;
        try { asset = getAssetFromToken(subAccount.token, address); } catch { return []; }
        /* ... unchanged ... */
      })
    : [];
  ```

shared/env/src/definitions/team-blockchain-support/index.ts
```diff
@@ -41,6 +41,11 @@ const teamBlockchainSupport = {
     def: "https://stacks.coin.ledger.com",
     desc: "Stacks API url",
   },
+  API_STACKS_NETWORK: {
+    parser: stringParser,
+    def: "mainnet",
+    desc: "Stacks network for legacy-bridge address derivation (mainnet | testnet)",
+  },
```
cted-ledger, line 48: ### 🟡 Warning — `API_STACKS_NETWORK` is only half-registered (confidence: 90)

- **Rule**: repo convention — every `API_*` var is declared on both sides.
- **Issue**: the var is added to the runtime definitions (`@shared/env`) but not to the `EnvTypes` interface in `libs/env/src/env.ts`, where `API_STACKS_ENDPOINT` and all its neighbours live (`EnvName = keyof EnvTypes`, `libs/env/src/env.ts:214`). That omission is precisely what forces the two `as StacksNetworkName` casts, and what the "`getEnv` returns `unknown` here" comments in `bridge/synchronization.ts` and `logic/buildUnsignedTx.ts` are describing — they document a workaround for a missing declaration, not a constraint of the system. It also leaves the var invisible to `getAllEnvNames` and to the env debug tooling.
- **Fix**: add `API_STACKS_NETWORK: string;` to `EnvTypes` next to `API_STACKS_ENDPOINT` (`libs/env/src/env.ts:17`), then drop both casts and trim the two comments accordingly.

libs/coin-modules/coin-stacks/src/logic/buildUnsignedTx.ts
```diff
@@ -34,10 +35,19 @@ function resolveRecipient(recipient: string): string {
...
+ * bridge's own address-derivation fix (`bridge/synchronization.ts`) rather than introducing a
+ * second, parallel config mechanism -- one env var controls network for both code paths.
+ * `API_STACKS_NETWORK` isn't registered in `@ledgerhq/live-env`'s typed definitions, so `getEnv`
+ * returns `unknown` here -- validated against the network's own known names and defaulted to
+ * `"mainnet"` otherwise, so real callers (who never set this) see no behavior change. */
+const configuredNetwork = getEnv("API_STACKS_NETWORK");
+export const NETWORK: StacksNetworkName = StacksNetworks.includes(
+  configuredNetwork as StacksNetworkName,
+)
+  ? (configuredNetwork as StacksNetworkName)
```
cted-ledger: ### 🟡 Warning — `NETWORK` is evaluated at module load, unlike its per-call twin (confidence: 85)

- **Rule**: latent bug / inconsistent with `bridge/synchronization.ts:184-193`.
- **Issue**: `const configuredNetwork = getEnv("API_STACKS_NETWORK")` sits at module scope, so the value is frozen at first `import` and any later `setEnv` is invisible. The equivalent read in `synchronization.ts` is per-call, so the two paths can disagree inside the same process. The coin-tester already has to work around this — that is the entire purpose of `src/env.setup.ts` and its comment. User impact today is nil (nobody sets the var), but the trap is armed for whoever next tries to make the network configurable.
- **Fix**: make it a function, aligned with `synchronization.ts`:
  ```ts
  export function getNetwork(): StacksNetworkName {
    const configured = getEnv("API_STACKS_NETWORK");
    return StacksNetworks.includes(configured as StacksNetworkName)
      ? (configured as StacksNetworkName) : "mainnet";
  }
  ```
  and update the `NETWORK` uses at L115, L138, L181, L207.

pnpm-lock.yaml
```diff
@@ -26952,7 +27022,7 @@ packages:
     resolution: {integrity: sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==}
     engines: {node: '>=6'}
     peerDependencies:
-      rxjs: '*'
+      rxjs: ^5.5.10
```
cted-ledger: ### 🟡 Warning — lockfile hunks not explained by this PR's dependency changes (confidence: 85)

- **Rule**: `.agents/agents/code-reviewer.md` § `pnpm-lock.yaml` diff — "Flag unrelated version bumps, mass reformatting, or entries not explained by the PR's `package.json` changes".
- **Issue**: this `peerDependencies` change (`rxjs: '*'` -> `rxjs: ^5.5.10`), plus a whitespace change inside the `q` package's deprecation message further down (around L36262), follow from neither the new `coin-tester-stacks/package.json` nor any other `package.json` in this PR. Most likely incidental re-resolution under a different pnpm version.
- **Fix**: regenerate the lockfile with the repo's pinned pnpm version and keep only the `libs/coin-tester-modules/coin-tester-stacks` importer block.

libs/coin-tester-modules/coin-tester-stacks/src/signer.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  createStacksPrivateKey,
+  deserializeTransaction,
+  getAddressFromPrivateKey,
+  pubKeyfromPrivKey,
+  publicKeyToString,
+  TransactionSigner,
+  TransactionVersion,
+} from "@stacks/transactions";
+import type { SingleSigSpendingCondition } from "@stacks/transactions/dist/authorization";
```
cted-ledger: ### 🟡 Warning — the legacy test signer uses a different `@stacks/transactions` major than the code under test (confidence: 80)

- **Rule**: test-infrastructure fragility.
- **Issue**: `coin-stacks` is now on `@stacks/transactions@^7.6.0` (`libs/coin-modules/coin-stacks/package.json:110`), but `signer.ts` runs `deserializeTransaction` from the v6 this package pins. It only holds because the STX / SIP-010 shapes happen to be wire-compatible — you document exactly that in `genericSigner.ts:20-24`, after hitting `DeserializationError: Could not read 3 as PostConditionType` on the pox-5 path. The next v7 feature emitted on the legacy path will break the tester with an opaque error, far from its cause.
- **Fix**: align `signer.ts` on `@stacks/transactions-v7` like `genericSigner.ts` (v7 ships `TransactionSigner` and `privateKeyToAddress`), and keep v6 only for `signerManager.ts` if `StacksMocknet` genuinely requires it.

### 🟢 Suggestion — deep import into `dist/`, outside the package's public exports (confidence: 85)

- **Issue**: `import type { SingleSigSpendingCondition } from "@stacks/transactions/dist/authorization"` (also in `genericSigner.ts:8`) only works because `@stacks/transactions` does not restrict its `exports`; it breaks silently on any internal restructure of the package.
- **Fix**: import the type from the package root if it is re-exported there, otherwise declare the minimal shape you use locally (`{ signature?: { data: string } }`).

libs/coin-tester-modules/coin-tester-stacks/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "prewatch": "pnpm build",
+    "watch": "tsc --watch --project tsconfig.build.json",
+    "watch:es": "tsc --watch --project tsconfig.build.json -m esnext --moduleResolution bundler --outDir lib-es",
+    "lint": "oxlint src",
+    "lint:fix": "oxfmt src && oxlint src --fix --quiet",
+    "format": "oxfmt src",
+    "format:check": "oxfmt src --check",
+    "typecheck": "tsc --noEmit --customConditions node",
+    "start": "pnpm jest --runTestsByPath src/*.test.ts --runInBand",
+    "unimported": "pnpm knip --directory ../../.. -W libs/coin-tester-modules/coin-tester-stacks"
```
cted-ledger, line 57: ### 🟢 Suggestion — dead-code script name (confidence: 80)

- **Rule**: `.agents/skills/knip-migration/SKILL.md` — dead-code detection is moving from `unimported` to `knip`.
- **Issue**: the script does run `knip` (compliant), but keeps the legacy name. `coin-tester-casper` has already moved to `knip-check`; `coin-tester-vechain` still has the old name.
- **Fix**: rename to `knip-check`, matching `coin-tester-casper`.

libs/coin-modules/coin-stacks/src/bridge/prepareTransaction.ts
pawell24: Fixed: gated behind an explicit `API_STACKS_SKIP_FEE_ESTIMATE` env var (properly registered in both `shared/env` and `EnvTypes`, unlike the existing `API_STACKS_NETWORK`), set only by the coin-tester's own `env.setup.ts`. Production `prepareTransaction` now always re-estimates over the network, exactly as before this PR touched it. Added a test asserting the fee is re-estimated even when the transaction already carries a positive fee from a prior call. See b3095f5500b.

libs/ledger-live-common/src/bridge/generic-coin-framework/getAccountShape.ts
pawell24: Fixed: only injects a vanished-token zero-entry when the fresh balance response actually carries at least one token (`freshTokenAssetsBalances.length > 0`), and wraps the `getAssetFromToken` call in try/catch so a throwing family implementation cannot fail the whole sync. Added tests for both (empty response -> no injection, throwing hook -> caught, other tokens still processed). See b3095f5500b.

shared/env/src/definitions/team-blockchain-support/index.ts
pawell24: Fixed: added `API_STACKS_NETWORK: string` and `API_STACKS_SKIP_FEE_ESTIMATE: boolean` to `EnvTypes` in `libs/env/src/env.ts`, dropped the `as StacksNetworkName` casts. See b3095f5500b.

libs/coin-modules/coin-stacks/src/logic/buildUnsignedTx.ts
pawell24: Fixed: extracted a shared `getConfiguredStacksNetwork()` (`common-logic/network.ts`, cast-free via a type-predicate instead of the `as StacksNetworkName` casts) that both `buildUnsignedTx.ts` and `bridge/synchronization.ts` now call per-call. No more module-load-time constant, and the two code paths can no longer disagree within the same process. See b3095f5500b.

pnpm-lock.yaml
pawell24: Fixed: reverted the two unrelated hunks (the any-observable `rxjs` peerDependency and the `q` package whitespace change) by hand, keeping only the new coin-tester-stacks importer block. Diff against develop is now a clean 70-line addition. See b3095f5500b.

libs/coin-tester-modules/coin-tester-stacks/src/signer.ts
pawell24: Fixed: migrated `signer.ts` to `@stacks/transactions-v7`, same API surface `genericSigner.ts` already uses -- confirmed `coin-stacks`'s own package.json is fully on `^7.6.0` for both the legacy bridge and the generic-adapter path, so v6 here was never actually matching the code under test, not just a theoretical risk. Also fixed the deep `dist/authorization` import in both signer.ts and genericSigner.ts by declaring the minimal local type instead of importing it. Added `signer.test.ts` (address/pubkey derivation + a real `sign()` round trip against a v7-built unsigned transfer) since no test existed for this file at all. See b3095f5500b.

libs/coin-tester-modules/coin-tester-stacks/package.json
pawell24: Left as is: `.agents/skills/knip-migration/SKILL.md` (the doc cited here) explicitly says the `unimported` -> `knip-check` script-name unification is "tracked separately" and that "matching the surrounding package's existing script name is fine". 12 of the 13 coin-tester packages still use the `unimported` name even though they already run knip underneath it (only coin-tester-casper has renamed) -- renaming just this one would make it inconsistent with its actual siblings rather than more consistent.

pawell24:
Thanks for the thorough review. Replied inline on all 7 findings and resolved those threads -- 6 fixed (critical fee-freeze, vanished-token blast radius, `API_STACKS_NETWORK` registration, `NETWORK` module-load consistency, pnpm-lock.yaml hunks, signer.ts v6/v7 + deep import), 1 left as-is with the counter-evidence (unimported/knip-check rename).

On the two non-inline points from the review body:
- Missing changeset: added, `.changeset/stacks-fee-and-vanished-token-fix.md`.
- PR scope vs title (coin-tester touching production `coin-stacks`/`live-common` files): still open, no action taken yet -- want your input on whether to split this or just broaden the PR description before merging.

libs/ledger-live-common/src/bridge/generic-coin-framework/getAccountShape.ts
pawell24: Follow-up: this PRs own coin-tester (`Send max CTT`, generic-adapter strategy) caught a regression from the `balanceListLooksComplete` guard -- a wallet whose only token gets fully swept legitimately gets back an empty token list on the next balance read, so the guard never let the sweep register as zero, and the sub-account balance never converged.

Reverted the list-non-emptiness gate in 622a83197cf, kept the try/catch around `getAssetFromToken`. The degraded-200 blast-radius concern this guard was defending against is still unproven (as flagged in the original comment) and now demonstrably in tension with a real case, so unconditional processing (the pre-review behavior) is what stays. Updated the regression test alongside it.

cted-ledger (approved):

libs/coin-modules/coin-stacks/src/bridge/prepareTransaction.ts
```diff
@@ -40,15 +41,23 @@ export const prepareTransaction: AccountBridge<Transaction>["prepareTransaction"
...
-      payload: serializePayload(tx.payload),
-      estimatedLength: estimateTransactionByteLength(tx),
-      network,
-      client: { baseUrl: getStacksBaseUrl() },
-    });
+    // Explicit opt-in only (coin-tester devnet, whose fee estimator has no historical data yet
+    // for this payload shape): `transaction.fee` alone can't gate this, since `onStatus` feeds
+    // this function's own `patch.fee` back as the next call's `transaction.fee` -- keying off its
+    // mere presence would freeze the fee at the first estimate for every real user.
+    if (getEnv("API_STACKS_SKIP_FEE_ESTIMATE") && transaction.fee && transaction.fee.gt(0)) {
```
vladyslavchupovskiy-ext-art: Not sure about approach using getEnv in prepareTransaction code just for coin-tester to be able to run properly, on my opinion tester should test module and fix only bugs related to mainnet ( test net if supported )

vladyslavchupovskiy-ext-art (approved):
General concern: both `API_STACKS_NETWORK` and `API_STACKS_SKIP_FEE_ESTIMATE` are touching production code so coin-tester can work against a devnet. Ideally coin-tester should be self-contained — it should adapt to production code, not the other way around. Not a blocker since I see those topics already challenged the fee one, but wanted to flag the pattern.

Sebastien-Dav1d (approved):

PR #20841, opened 2026-08-17
feat(analytics): add  shared analytics packages (LIVE-34837)

LL782:
### 📝 Description

Adds `@shared/analytics` and `@shared/analytics-react` – a single tracking pipeline for Desktop, Mobile and CLI.

**No consumer is wired in yet**

#### `@shared/analytics` — the React-free, Redux-free core

Each app keeps owning its Segment SDK, its logger, its extra properties and its consent state, and registers them through injection points:

`track` resolves the enricher without forcing a microtask when that enricher is synchronous, so Desktop's synchronous `trackSubject` assertions still hold. Only Mobile's async enricher makes `track` return a promise.

The screen refs live here too because the pipeline reads them. The raw refs are exported next to the getters as an interim step that will be retired by LIVE-36002.

#### `@shared/analytics-react` — the components

| Component | Entry point | Trigger | Backing call |
| --- | --- | --- | --- |
| `<Track>` | both | `onMount` / `onUnmount` / `onUpdate` props | `track` |
| `<TrackPage>` | web | mount | `trackPage` |
| `<TrackScreen>` | native | `useIsFocused()` | `screen` |

`<TrackPage>` and `<TrackScreen>` are deliberately **not** merged — page events are Desktop's model (one event per mount), screen events are Mobile's (one event per focus, de-dupable). `@react-navigation/native`, needed only by `<TrackScreen>`, is an optional peer dependency so a web consumer never has to resolve it.

#### Behaviours unified

Three points where the two apps had drifted; the pipeline picks one behaviour for each:

1. **Route refs update before the consent gate** (Mobile's behaviour). Desktop gated first, so with analytics off `getCurrentTrackingPage()` returned `""` to its non-analytics consumers.
2. **Extras win over caller props** (Mobile's `track` / `screen`, Desktop's `trackPage`). Desktop's `track` used to let caller props win.
3. **The property filter applies to the merged base**, so ref-derived `page` / `source` are scrubbed too — previously only when the caller passed them explicitly.

#### Usage

```ts
import { setAnalytics, setEnricher, setMandatoryEnricher, setStore, setTrackingSelector, track } from "@shared/analytics";

setStore(store);
setTrackingSelector(state => trackingEnabledSelector(state as AppState));
setEnricher(state => extraProperties(state as AppState));
setMandatoryEnricher(state => getMandatoryProperties(state as AppState));
setAnalytics({
  track: (event, properties) => segment.track(event, properties).then(() => "enqueued" as const),
  flush: () => segment.flush(),
});

track("Some Event", { foo: "bar" });
```

```tsx
import { Track, TrackPage } from "@shared/analytics-react";

<Track onMount event="Discoverability - Prompt" eventProperties={{ language }} />;
<TrackPage category="Analytics Consent" name="Optional" flow="onboarding" />;
```

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-34837](https://ledgerhq.atlassian.net/browse/LIVE-34837)

[LIVE-34837]: https://ledgerhq.atlassian.net/browse/LIVE-34837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared/analytics-react/src/Track.native.test.tsx
```diff
@@ -0,0 +1,25 @@
+// The native barrel also pulls in <TrackScreen>, whose @react-navigation/native import ships as
+// untransformed ESM. <Track> never uses it.
+jest.mock("@react-navigation/native", () => ({ useIsFocused: jest.fn() }));
+
+import React from "react";
+import { render } from "@testing-library/react-native";
+import { setAnalytics, setStore, setTrackingSelector } from "@shared/analytics";
+import { Track } from "./index.native";
+
+const transportTrack = jest.fn();
```
LL782: Let's just call this `track`

shared/analytics-react/src/Track.native.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+  setStore({ getState: () => ({}) });
+  setTrackingSelector(() => true);
+});
+
+describe("Track (native entry point)", () => {
+  it("sends the event on mount when asked to", () => {
+    render(<Track onMount event="Some Event" foo="bar" />);
+
+    expect(transportTrack).toHaveBeenCalledWith("Some Event", { page: undefined, foo: "bar" });
+  });
```
LL782: **request: Let's test `onUpdate` as well**

LL782: Let's ensure the events are called the correct number of times, i.e. not repeatedly when the component is re-evaluated

shared/analytics-react/src/Track.tsx
LL782, line 1: I don't see a good reason for either `useCallback` or `useMemo` here. Please explain the case for them or otherwise remove.

LL782 (commented):
Still reviewing but releasing comments so far

shared/analytics-react/src/Track.native.test.tsx
LL782: Done in ca7c6acbf4c — renamed to `track`, and simplified the setup to `setAnalytics({ track })`. Applied the same rename in the three sibling test files (`Track.web.test.tsx`, `TrackPage.web.test.tsx`, `TrackScreen.native.test.tsx`) so the package stays consistent.

shared/analytics-react/src/Track.native.test.tsx
LL782: Done in ff1c368eff4 — added `sends the event when a property changes and onUpdate is set` to the native entry-point test: render with `onUpdate`, rerender with a changed prop, assert the event fires with the new properties.

LL782: Done in 7f4b3ca76c7.

`<Track>` now asserts counts, not just payloads:
- `toHaveBeenCalledTimes(1)` on the mount, unmount and update cases
- new `sends the mount event once, however often the component is re-evaluated` (native + web) — rerender with a changed prop, still exactly one event
- new `sends nothing when re-rendered with a new but equal object property` (web) — guards the `isEqual` path against object identity, since `memo` only compares shallowly

`<Track>` came out clean on all of these. `<TrackScreen>` is safe too — the `wasFocusedRef` guard means a re-render cannot re-fire it.

⚠️ The same check on `<TrackPage>` does fail, so I have raised it as a separate thread rather than fold a production change into this one.

shared/analytics-react/src/TrackPage.tsx
```diff
@@ -0,0 +1,38 @@
...
+const TrackPageComponent = ({
+  category,
+  name,
+  refreshSource = true,
+  mandatory = false,
+  ...properties
+}: TrackPageProps): null => {
+  useEffect(() => {
+    trackPage(category, name, properties, true, refreshSource, mandatory);
+  }, [category, name, properties, refreshSource, mandatory]);
```
LL782: Raising this from the call-count thread on `Track.native.test.tsx`, since it is a production change rather than a test one.

`properties` is a fresh rest-object on every render, so as a dependency it re-runs this effect as soon as `memo` lets a re-render through. Observed before the fix:

| Rerender | Events emitted |
| --- | --- |
| `balance={1}` → `balance={2}` | `Page Portfolio` **twice**; the 2nd has `source: "Portfolio"` |
| `meta={{a: 1}}` → `meta={{a: 1}}` (equal, new object) | `Page Portfolio` **twice** |

The duplicate taking the page as its own `source` is the nastier half — it would have corrupted the source chain for the next page event.

The existing `sends nothing more when re-rendered with the same properties` test only passed because `memo` blocked that re-render outright, so the effect was never exercised.

Fixed in `4e3424176e3`: the values are read through a ref and the effect deps are `[]`, so a page view belongs to the mount. Trade-off, now pinned by a test: swapping `category` on an already-mounted `<TrackPage>` emits nothing — render one `<TrackPage>` per page instead. That is the documented intent of the component, and there are no consumers of `@shared/analytics-react` yet, so nothing depended on the old behaviour.

`<TrackScreen>` is unaffected — its `wasFocusedRef` guard already prevents a re-render from re-firing.

shared/analytics-react/src/Track.tsx
LL782: I've looked into this there was no case for it — removed in de4691d995a.

The `useCallback` was self-defeating: its only dependency was `props`, and React builds a fresh props object on every render, so the dependency changed every time and the callback was recreated every time. It memoised nothing and cost an extra hook call plus a dependency array to read past.

Both effects now call through the `trackEventRef` that was already there, so the function identity never needs to be a dependency at all. That also shortened the update effect deps from `[onUpdate, props, trackEvent]` to `[onUpdate, props]`.

shared/analytics-react/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Usage
+
+```tsx
+import { Track } from "@shared/analytics-react";
+
+<Track
+  onMount
+  event={`Discoverability - Prompt - ${defaultLanguage}`}
+  eventProperties={{ language: defaultLanguage }}
```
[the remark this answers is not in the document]
LL782: Good catch — the docs were wrong, fixed in `44f5baf4f13`.

Of the two options you offered I took the first: the docs now match the implemented API. `<Track>` forwarding every non-lifecycle prop flat is deliberate and matches both legacy components it replaces (`apps/ledger-live-desktop/src/renderer/analytics/Track.ts` and `apps/ledger-live-mobile/src/analytics/Track.ts` both spread `...properties`), so adding an `eventProperties` prop would invent a new API with no callers.

The example now reads:

```tsx
<Track
  onMount
  event={`Discoverability - Prompt - ${defaultLanguage}`}
  language={defaultLanguage}
/>
```

I also spelled the flat contract out in prose, because there are two good reasons to trip over this: Desktop’s `<Button>` genuinely does take an `eventProperties` prop, and `eventProperties` is the field name the pipeline logs the flattened properties under internally (`shared/analytics/src/internals/emit.ts`). The example is covered by an existing test — `Track.web.test.tsx` renders exactly this shape and asserts `{ page: undefined, language: "en" }`.

Worth noting where the bad example came from: `apps/ledger-live-mobile/src/components/CheckLanguageAvailability.tsx:57` passes `eventProperties={{ language: defaultLanguage }}` to the legacy mobile `<Track>`, which flattens the same way — so that event has been shipping a nested `eventProperties` object instead of `language` all along. Pre-existing, app-side, and out of scope for this PR, but it needs its own fix.

shared/analytics-react/src/Track.tsx
LL782: Regarding `useMemo`, it's actually a `memo`, which is a different case from the two we flagged: it does real work, bailing out of the render entirely when a parent re-renders with shallow-equal props, which keeps the update effect from running at all. 

Currently the `isEqual` guard already prevents duplicate events, however on reflection the memo was a more elegant solution. Now that we have good test coverage around duplicate calls, let's go with `memo()` and remove the `isEqual` check.

Ensure the tests stay green.

LL782: Done in b5247962cda — `memo()` stays, `isEqual` is gone, and `lodash` goes with it since that was its only use in the package.

One consequence worth naming: `memo` compares shallowly where `isEqual` compared deeply. A prop object rebuilt inline with equal contents (`filter={{ tag: "all" }}` twice) now gets through and emits an event. So we've swapped one test rather than just deleting the guard — "sends nothing when re-rendered with a new but equal object property" was pinning the deep-compare behaviour, and now reads "compares properties shallowly, so a new but equal object counts as a change". Callers wanting the old behaviour hold their object identity stable, which is the normal React expectation.

The update effect still needs to skip its own first run, so the deep comparison is replaced by a `hasMountedRef` flag rather than nothing at all — mount is not an update. Also added a case covering re-render without `onUpdate`, which the `isEqual` branch used to cover incidentally.

31 tests pass, `Track.tsx` at 100% branch coverage, typecheck, lint, format and knip all clean.

shared/analytics-react/src/Track.native.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    expect(track).toHaveBeenCalledTimes(1);
+    expect(track).toHaveBeenCalledWith("Some Event", { page: undefined, foo: "bar" });
+  });
+
+  it("sends the mount event once, however often the component is re-evaluated", () => {
+    const { rerender } = render(<Track onMount event="Some Event" foo="bar" />);
+    rerender(<Track onMount event="Some Event" foo="baz" />);
+
+    expect(track).toHaveBeenCalledTimes(1);
+  });
```
LL782: Let's put these into a single test

LL782:
## Adversarial review

Went through this looking for design flaws, doc/code drift and test balance. Verdict up front: the pipeline is sound, the test suite is genuinely good, and the docs are the weakest part — they described a pipeline we don't actually have.

Pushed five no-brainers; everything else below is a question, not a change.

### Pushed

- 47eee6d2800 — the README pipeline diagram was wrong twice over. It put the consent gate before the route refs (the code updates refs first, and its own parenthetical said so) and enrich before filter (we filter the base, then enrich). Same ordering is repeated in the PR description and the changeset, so those want the same correction.
- e56cee52129 — **a rejecting async enricher escaped `track()` as an unhandled rejection.** `send` did `extras.then(dispatch)` with no rejection handler, and every caller is a fire-and-forget effect. Confirmed with a test before fixing: `setEnricher(() => Promise.reject(...))` then `await track(...)` rejects. On mobile the enricher awaits native permission state, so this is a reachable path. Now drops the event and reports `failed` on trackSubject.
- 74ca49e155f — the blur branch of `TrackScreen`'s focus guard was the package's one uncovered branch. Now 100%.
- c3c5e804a49 — both `tsconfig.test.json` files were dead. Nothing references them and jest transpiles with @swc/jest, which never reads a tsconfig; the `@shared/analytics` one only restated the `types` it inherited.
- 18eb09ad298 — the `screenRefs.ts` comment said "39 call sites read (and one writes)". Eleven production files assign `currentRouteNameRef.current` (the wallet-API webviews on both apps, mobile's global search). That materially undercounts what LIVE-36002 has to replace — a function-only API needs a setter for the current page, not just `setTrackingSource`.

### The one I'd push back hardest on

**The positional boolean signature, while it's still free to change.**

```ts
screen(category?, name?, properties?, updateRoutes?, refreshSource?, avoidDuplicates?, mandatory?)
```

Four trailing booleans. Our own tests already read like this:

```ts
screen("Portfolio", "Duplicated", undefined, true, true, true);
```

Only the last `true` is what that test is about. And roughly a third of `tracking.ts` is JSDoc that exists purely to explain what the positions mean — an options object would delete most of it rather than documenting around it.

The PR's strongest selling point is "no consumer is wired in yet". That cuts both ways: it makes this the *only* moment the signature is free. Once LIVE-35068/35069 wire the call sites, we've inherited the ergonomics of the thing we were replacing, permanently. If we're keeping it, I'd rather that be a stated decision (faithful port, migrate later) than a default.

### Worth deciding before the wiring tickets

1. **The transport `log` hook drops information both apps' loggers use.** Desktop's `logger.analyticsPage(category, name, properties)` takes the parts separately; we only hand over the joined `"Page Foo Bar"`, and splitting it back is fragile because names contain spaces. Separately, mobile's kind is `"track" | "screen"` and ours is `"track" | "page"`, so mobile's `warnOnceNoSegmentClient` changes what it reports. Passing the parts through would cost nothing now and save an awkward adapter later.

2. **There's a fourth unification we haven't documented.** Desktop's `track` returns silently when there's no store; we emit `skipped_no_store` to trackSubject (mobile's behaviour). Harmless, but Desktop's in-app console will start showing entries it never showed, and the README promises exactly three changes.

3. **`registry.ts` puts pipeline internals in the public API.** `index.ts` does `export * from "./registry"`, so `resolveExtraProperties` and `applyPropertyFilter` are public API of `@shared/analytics`. Per the package-public-api skill the fix is moving them to `registry.internals.ts` rather than shortening the export list. `getIsTracking` is public too, justified in the README as "exported for callers that need to check it" — which callers? If none yet, it's speculative surface.

4. **`internals/screenRefs.ts` doesn't earn its folder.** `screenRefs.ts` is seven lines of `export *` over a thirteen-line file that is 100% public, so the internals split hides nothing. Collapsing them is one file fewer with identical import paths. (`internals/emit.ts` is genuinely private — that one's right.)

5. **`lastScreenEventName` can't be reset.** Module-level dedup state, faithful to mobile, but nothing clears it on logout or a consent change. Our tests only dodge cross-test pollution by giving each test a distinct screen name, which is a bit load-bearing for a naming convention.

6. **`"paths": { "*": ["./*"] }`** in `shared/analytics/tsconfig.json` replaces the base path map. Harmless today because the package imports nothing from the workspace — a trap the first time someone imports `@shared/env`.

7. **Wiring note for LIVE-35069:** `flush()` delegates to `transport.flush?.()` and never emits `"flushed"`. Mobile produces that status by monkey-patching `client.flush` and pushing a synthetic `"[Flush]"` event straight onto `trackSubject`. That patch has to survive the migration and target the shared subject, or the status becomes unreachable.

### On test volume

Asked myself whether 400 lines of `tracking.test.ts` against 176 lines of implementation was over-testing. It isn't. The names describe behaviour rather than implementation, `beforeEach` resets all six registry slots, and the branch coverage is real — the async/sync split, the mandatory path, the filter/extras precedence and every delivery status are each pinned by exactly one test. The only genuine gap was the blur branch, now covered. I'd leave it alone.

Docs are the opposite story — dense and mostly excellent, but the two diagrams and the migration-scope comment had drifted from the code, which is the specific failure mode the repo's comments skill warns about.

LL782:
## Second pass — one blocker, verified

A deeper sweep against the two app implementations turned up something the first pass missed. Everything below I reproduced locally rather than reasoned about.

### 🚨 `<TrackPage>` losing the mandatory consent event

`TrackPage.tsx` uses `[]` deps behind a `latestRef`, so it emits exactly once per mount, ever. Desktop's `TrackPage.ts:49-51` has deps `[category, name, properties, refreshSource, mandatory]`, so a prop change on a mounted component re-fires `trackPage`.

That difference is load-bearing, and there's an existing desktop test that pins it — `apps/ledger-live-desktop/src/renderer/analytics/TrackPage.test.tsx:41-80`:

```tsx
const { rerender } = render(<TrackPage category="Analytics Consent" name="Optional" flow="test-flow" />, { store });
expect(events).toEqual([]);                       // consent off, nothing sent

rerender(<TrackPage category="Analytics Consent" name="Mandatory" flow="test-flow" mandatory />);
await waitFor(() => expect(events).toHaveLength(1));   // ← emits nothing under the new component
```

Same mounted component, swapped to `mandatory`. Under our version that rerender emits nothing and the **analytics-consent page event is lost** — the one event we least want to drop, and the test that catches it is a desktop test that won't run until LIVE-35068.

We introduced this deliberately in 4e3424176e3 ("emit one TrackPage event per mount") and the README states it as the model, so this is a real design call, not a slip. But the consent screen genuinely drives `name`/`mandatory` from state on a mounted component, so "one event per mount" and "Desktop keeps using `<TrackPage>` unchanged" can't both hold. Needs a decision before the wiring ticket — I'd rather we made it now than discovered it in LIVE-35068.

### Also verified, not fixed

- **The sync fast path was only half true** — fixed the docs in cc4859500c3 rather than the code. `emit` defers `trackSubject.next` whenever `transport.track` returns a thenable, so a *sync* enricher plus a promise-returning transport still emits asynchronously. `AnalyticsBrowser.track` returns a promise, and so does the transport in our own README example — so the documented desktop wiring defeated the documented guarantee. Proved it with a throwaway test: sync enricher + `track: () => Promise.resolve()` → `track()` returns a Promise and `trackSubject` publishes nothing synchronously.

- **Mobile `<Track onUpdate>` gets noisier.** Mobile compares with lodash `isEqual` (`Track.ts:36`); we rely on `memo`'s shallow compare. Inline object/array props that are new-but-deeply-equal now emit where they previously didn't. Undocumented divergence — worth a line in the README either way.

- **Consent fails open, and the failure has a crash path.** `consent.ts:17` returns `{enabled: true}` before the store check when no selector is registered. That's the documented CLI contract, but it means an app that wires `setStore`/`setEnricher` at module load and `setTrackingSelector` later leaks every event in that window. Worse, no selector + no store hands `undefined` to the enricher, and desktop's `extraProperties` dereferences `state.trustchain` — `emit` only try/catches `transport.track`, so that throw escapes `track()` into a React effect. Cheapest guard is requiring the selector and store to be registered together.

- **Nothing is called at all when the transport is unset.** `emit.ts:31-35` returns before `transport.log?.()`, so mobile's `warnOnceNoSegmentClient` → `DdLogs.warn("analytics_event_skipped_no_client")` has no hook to hang off. Both apps also log consent-*blocked* events under `ANALYTICS_LOGS`; we log nothing for those.

- **Route refs initialise to `undefined`, both apps use `React.createRef()` → `null`.** So `page`/`source` keys vanish from the payload before the first page event instead of arriving as explicit nulls.

### Fixed since the last comment

- d64f46f4632 — the test gaps. `screen()`'s consent-blocked path and `emit`'s `enqueued` default for a transport resolving nothing were the package's only uncovered lines. Added a mandatory event under *granted* consent (both existing mandatory tests set the selector to `false`, so nothing pinned that a consenting user's mandatory event still gets the reduced payload), and made the async test actually assert `track()` returns a promise — `await undefined` still yields a tick, so it was passing either way.
- Also in there: **~25 assertions of the form `{ page: undefined, ... }` assert nothing about that key.** `toHaveBeenCalledWith` uses `toEqual` semantics, which ignore undefined-valued keys — I confirmed `toHaveBeenCalledWith("E", { page: undefined, foo: "bar" })` passes against `fn("E", { foo: "bar" })`. If `track` stopped injecting `page` the suite stayed green. The key's presence matters because the property filter is handed `{ page: <key present> }`. Pinned it directly with two tests rather than rewriting all 25.
- cc4859500c3 — the sync fast path docs.

### One thing that is *not* ours

Worth recording so nobody fixes it here: `shared/analytics-react` emits `SF:src/Track.tsx` in its lcov, which won't resolve once `test-scope-reusable.yml` concatenates it at the repo root — so Sonar silently drops the package's coverage. It looks like our bug but isn't: `support/jest-features-flow/index.js:41` puts `coverageReporters` inside each `project`, where Jest ignores it, and the relative `projectRoot: "../../../"` doesn't apply. I checked an existing consumer, `features/flow/pay-card-feature-tour`, and it emits `SF:src/components/...` too. **Every flow package using the factory has this**, and an absolute `projectRoot` fixes it. Separate ticket against `@support/jest-features-flow` rather than a local override here.

### Withdrawn from my last comment

The transport `log` hook dropping desktop's `category`/`name` split stands, but I overstated `screen` vs `trackPage`'s `?? undefined` inconsistency as a smell — it's faithful to both originals, and once the refs start at `undefined` rather than `null` it makes no observable difference anyway.

shared/analytics/src/tracking.ts
```diff
@@ -0,0 +1,222 @@
...
+    screenName: screenName,
+  };
+}
+
+/**
+ * Track an event named `Page ${category}${name ? " " + name : ""}`, where both parts are optional.
+ *
+ * Same route-name logic as {@link trackPage}, plus de-duplication against the last screen event.
+ */
+export function trackScreen(
```
Sebastien-Dav1d, line 177: Both trackScreen and trackPage could merge their logic as it almost the same

shared/analytics/src/screenRefs.test.ts
Sebastien-Dav1d, line 1: Why this test is not alongside screenRefs.ts in internals

shared/analytics/src/registry.ts
```diff
@@ -0,0 +1,77 @@
...
+/** Anything with a `getState()`. The package never imports Redux or an app `State` type. */
+export function setStore(next?: AnalyticsStore): void {
+  store = next;
+}
+
+export function getAnalyticsState(): unknown {
+  return store?.getState();
+}
+
+/** Consent gate. While no selector is registered, tracking is ALWAYS ENABLED — the CLI contract. */
```
Sebastien-Dav1d, line 41: Why on by default ? For privacy, off first would be better and we can make an helper in case

shared/analytics/src/consent.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * | Situation                                        | Result                                     |
+ * | ------------------------------------------------ | ------------------------------------------ |
+ * | no tracking selector registered                  | enabled — always-on, the CLI depends on it |
+ * | selector registered, no state                    | disabled, `"store not initialised"`        |
+ * | selector returns false, `mandatory` falsy        | disabled, `"analytics not enabled"`        |
+ * | selector returns false, `mandatory` true         | enabled, with the *mandatory* enricher     |
+ */
+export function getIsTracking(
+  state: unknown | null | undefined,
```
Sebastien-Dav1d, line 13: why not state?: unknown | null like the line below

shared/analytics/src/consent.ts
```diff
@@ -0,0 +1,23 @@
...
+  mandatory?: boolean | null,
+): TrackingResult {
+  const trackingSelector = getTrackingSelector();
+  if (!trackingSelector) return { enabled: true };
```
Sebastien-Dav1d, line 17: I think it should be false, cli would need an explicit init if he wants analytics but as a shared package. If a real app misses this, it could be privacy violation. Not that it would happen but just in case

shared/analytics-react/jest.config.js
```diff
@@ -0,0 +1 @@
+module.exports = require("@support/jest-features-flow").createFlowJestConfig();
```
Sebastien-Dav1d, line 1: Okay, both have the same jest config. But I don't think it is good behaviour to use features-flow jest config for a shared package

shared/analytics-react/src/index.native.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Track";
+export * from "./TrackScreen.native";
```
Sebastien-Dav1d, line 2: No need for .native and .web in imports

shared/analytics-react/src/Track.native.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+describe("Track (native entry point)", () => {
+  it("sends the event on mount when asked to", () => {
+    render(<Track onMount event="Some Event" foo="bar" />);
+
+    expect(track).toHaveBeenCalledTimes(1);
+    expect(track).toHaveBeenCalledWith("Some Event", { page: undefined, foo: "bar" });
+  });
+
+  it("sends the mount event once, however often the component is re-evaluated", () => {
```
Sebastien-Dav1d: What is this testing ? What is this name XD, why "however" ?

shared/analytics-react/src/Track.tsx
Sebastien-Dav1d, line 1: For ui-less files, I would prefer .ts extensions but that's on your preferences

shared/analytics-react/src/Track.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const { event, onMount, onUnmount, onUpdate, mandatory, ...properties } = props;
+    track(event, properties, mandatory);
+  };
+
+  // Read through a ref so unmount reports the props the component had when it went away, and so the
+  // effects below never need this function as a dependency.
+  const trackEventRef = useRef(trackEvent);
+  trackEventRef.current = trackEvent;
+
+  useEffect(() => {
```
Sebastien-Dav1d, line 25: Better solution in react 19.2 but we are in react 19.1.4 .....

https://react.dev/learn/separating-events-from-effects#reading-latest-props-and-state-with-effect-events

shared/analytics-react/src/TrackPage.web.tsx
Sebastien-Dav1d, line 1: If this is intented for web only, we should say that it is web

shared/analytics-react/src/TrackPage.web.test.tsx
Sebastien-Dav1d, line 1: If the file is not a web file. Why this test needs to have a .web extension ?

shared/analytics/src/trackSubject.ts
```diff
@@ -0,0 +1,5 @@
+import { ReplaySubject } from "rxjs";
+import type { LoggableEvent } from "./types";
+
+/** Dev bus consumed by the in-app analytics consoles. Replays the last 30 events to a late subscriber. */
+export const trackSubject = new ReplaySubject<LoggableEvent>(30);
```
Sebastien-Dav1d, line 5: Why is the length fixed ? Can't we make it a variable ?
And I think it should be private a have a readOnly interface, currently anyone can add or delete entries

shared/analytics/src/tracking.ts
```diff
@@ -0,0 +1,222 @@
...
+  TrackingRouteRef,
+} from "./types";
+
+const lastScreenEventName: TrackingRouteRef = { current: undefined };
+
+/**
+ * A blocked event is reported only when there is no store yet — a consent-off event is silent, so
+ * the in-app consoles never reveal what a user who refused tracking would have sent.
+ */
+function reportBlocked(
```
Sebastien-Dav1d, line 27: This function does not filter parameters, it is less likely to add to trackSubject sensitive info but still can

shared/analytics/src/registry.ts
Sebastien-Dav1d, line 1: Regarding whether or not the registry should be internals, I think it should be in internal, well I think that stuff like propertyFilter should not be accessed externally. But if we wanted to make it cleaner , we should make a module with a factory, expose only the useful functions for clients and keep internally all internal functions such as propertyFilter. It is cleaner but it changes the api, more work but cleaner in general

shared/analytics/src/tracking.ts
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Resolves the enricher without forcing a microtask when it is synchronous: desktop asserts on
+ * `trackSubject` synchronously after render, mobile awaits native permission state. `emit` defers
+ * too when the transport returns a thenable, so a synchronous enricher alone does not buy a
+ * synchronous emission — see the sync fast path in the README.
+ *
+ * A rejecting async enricher drops the event rather than sending it half-enriched, and never
+ * rejects: callers are fire-and-forget effects, so a rejection here would go unhandled.
+ */
+function send(
```
Sebastien-Dav1d, line 50: Send should use propertyFilter, relying on every caller to filter their input is dangerous. On the real app, track functions had the filter but this only created more errors that had to be fixed

shared/analytics/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/** What the pipeline calls to send a finished payload. Implemented by the app owning the SDK. */
+export interface AnalyticsTransport {
+  /**
+   * Send the fully-built payload. May return/resolve a `DeliveryStatus` to override the default
+   * `"enqueued"`. Throwing or rejecting is reported as `"failed"` and never propagated.
+   */
+  track(event: string, properties: Props): void | Promise<void | DeliveryStatus>;
+  /** App-side logging hook, called before the send. */
+  log?(kind: "track" | "page", event: string, properties: Props): void;
```
Sebastien-Dav1d, line 29: Should handle separately track and page to handle properties properly. Changing the interface forces the client to change their calls

shared/analytics-react/src/index.native.ts
LL782: Cheers! I'll fix this (done locally, will push later)

shared/analytics-react/src/Track.native.test.tsx
LL782: It's testing that the onMount event is only tracked one, even when the component is re-rendered. I'll adapt the English to see if I can make that clearer.

shared/analytics-react/src/Track.tsx
LL782: Interesting! It has no JSX so I see your point. On the other hand it is a React component and `.tsx` communicates that well 🤔

shared/analytics-react/src/Track.tsx
LL782: Very nice :)

LL782: We're on 19.1 however, https://github.com/LedgerHQ/ledger-live/blob/develop/pnpm-workspace.yaml#L176, and I'm not sure this is the PR to change that 🤔

LL782:
Closing in favour of more granular approach. See linked items in https://ledgerhq.atlassian.net/browse/LIVE-34837 for the split.

PR #20847, opened 2026-08-17
feat(app-lock): add verifier and protection state packages (LIVE-35917)

LucasWerey:
### 📝 Description

Scaffolds the packages the **User App Authentication** epic builds on. No screens, no orchestration, nothing wired into the app yet — this lands on its own so the layering can be reviewed before the rest of the epic builds on it. #20905 carries the whole epic if you want to see what these packages end up holding — this is its first commit, unchanged.

**Problem**: today's password protects nothing. It is written to the Keychain in plaintext and verified with `credentials.password === password`, so anyone who can read the app's files reads the password itself.

**Solution**: two packages, each owning one concern.

| Package | Owns |
|---|---|
| `@shared/password-verifier` | The `PasswordVerifier` record and its **constant-time** comparison. Zero dependencies, pure functions over `Uint8Array`. |
| `@features/platform-app-lock` | Protection state schemas (`hasPassword`, `biometricsEnabled`, `isLocked`, `AuthenticationType`), the biometrics status unions, and the errors. |

```ts
import { createPasswordVerifier, matchesPasswordVerifier } from "@shared/password-verifier";

const verifier = createPasswordVerifier({ digest, salt, scrypt });
const ok = matchesPasswordVerifier(verifier, freshDigest);
```

Nothing is encrypted: the lock stores `scrypt(password, salt)`, never the password. Each package README records the reasoning behind its shape, including why the verifier is its own package and why the flow will use `screens/<Name>/`.

### 🧪 Tests

20 unit tests. `matchesPasswordVerifier` is covered for first-byte, last-byte, single-bit, length, empty-digest and prefix mismatches, plus reference-vs-value equality; `createPasswordVerifier` for its defensive copy. `@features/flow-app-lock` deliberately does not land here: a package whose public API is empty has no barrel to check, which `enforce-package-structure` rejects outright. It arrives with its first screens, in the next commit of the epic.

Verified locally: `typecheck`, `test`, `format:check`, `lint`, `unimported`, `enforce-boundaries:lint:boundaries` and `enforce-package-structure` all green on both packages, and `pnpm install --frozen-lockfile` on this commit alone.

### 👀 QA

**Nothing to test.** No user-facing change and no app wiring — none of these packages is imported anywhere yet. The only change outside them is `react-native-keychain` moving to the pnpm catalog, which still resolves to 10.0.0 for Ledger Wallet Mobile.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35917](https://ledgerhq.atlassian.net/browse/LIVE-35917) (epic: [LIVE-35505](https://ledgerhq.atlassian.net/browse/LIVE-35505))
- **Guideline**: [Monorepo DDD Re-architecture: Structure & Flow](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6111232117)
- **Related**: #20848 adds the `lwmPasswordRevamp` flag that gates the epic. Independent — no shared files.

[LIVE-35917]: https://ledgerhq.atlassian.net/browse/LIVE-35917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -131,6 +131,10 @@ features/flow/pay-card-deposit/                          @ledgerhq/wallet-xp
 features/flow/pay-card-feature-tour/                     @ledgerhq/wallet-xp
 features/flow/pay-card-request/                          @ledgerhq/wallet-xp
 
+# Wallet — App lock
+features/platform/app-lock/                              @ledgerhq/wallet-xp
+/shared/password-verifier/                               @ledgerhq/wallet-xp
+
 # Wallet — shared UI
 /shared/ui-queued-bottom-sheet/                          @ledgerhq/wallet-xp
 
--- a/features/platform/app-lock/jest.config.js
+++ b/features/platform/app-lock/jest.config.js
@@ -1,0 +1,20 @@
+module.exports = {
+  testEnvironment: "node",
+  roots: ["<rootDir>/src"],
+  testMatch: ["**/*.test.ts"],
+  transform: {
+    "^.+\\.(t|j)sx?$": [
+      "@swc/jest",
+      {
+        jsc: {
+          target: "esnext",
+        },
+      },
+    ],
+  },
+  coverageReporters: ["json", ["lcov", { file: "lcov.info", projectRoot: "../../../" }], "text"],
+  reporters: [
+    "default",
+    ["jest-sonar", { outputName: "sonar-executionTests-report.xml", reportedFilePath: "absolute" }],
+  ],
+};
--- a/features/platform/app-lock/src/biometrics.ts
+++ b/features/platform/app-lock/src/biometrics.ts
@@ -1,0 +1,18 @@
+import type { BIOMETRY_TYPE } from "react-native-keychain";
+
+export type BiometricsKind = `${BIOMETRY_TYPE}`;
+
+export type BiometricsAvailability =
+  | Readonly<{ status: "available"; kind: BiometricsKind }>
+  /** No hardware. */
+  | Readonly<{ status: "unavailable" }>
+  /** Hardware present, nothing enrolled. */
+  | Readonly<{ status: "notEnrolled" }>
+  | Readonly<{ status: "lockedOut" }>;
+
+export type BiometricsPromptResult =
+  | Readonly<{ status: "succeeded" }>
+  /** Dismissed — must not count as a failed attempt. */
+  | Readonly<{ status: "cancelled" }>
+  | Readonly<{ status: "failed" }>
+  | Readonly<{ status: "lockedOut" }>;
--- a/shared/password-verifier/jest.config.js
+++ b/shared/password-verifier/jest.config.js
@@ -1,0 +1,20 @@
+module.exports = {
+  testEnvironment: "node",
+  roots: ["<rootDir>/src"],
+  testMatch: ["**/*.test.ts"],
+  transform: {
+    "^.+\\.(t|j)sx?$": [
+      "@swc/jest",
+      {
+        jsc: {
+          target: "esnext",
+        },
+      },
+    ],
+  },
+  coverageReporters: ["json", ["lcov", { file: "lcov.info", projectRoot: "../../" }], "text"],
+  reporters: [
+    "default",
+    ["jest-sonar", { outputName: "sonar-executionTests-report.xml", reportedFilePath: "absolute" }],
+  ],
+};
--- a/shared/password-verifier/src/verifier.ts
+++ b/shared/password-verifier/src/verifier.ts
@@ -1,0 +1,29 @@
+import type { PasswordVerifier, ScryptParams } from "./types";
+
+export const PASSWORD_VERIFIER_VERSION = 1;
+
+export function createPasswordVerifier(
+  input: Readonly<{ digest: Uint8Array; salt: Uint8Array; scrypt: ScryptParams }>,
+): PasswordVerifier {
+  return {
+    version: PASSWORD_VERIFIER_VERSION,
+    scrypt: input.scrypt,
+    salt: Uint8Array.from(input.salt),
+    digest: Uint8Array.from(input.digest),
+  };
+}
+
+/** Compares in constant time: no `===`, no early exit, or the timing leaks the digest. */
+export function matchesPasswordVerifier(verifier: PasswordVerifier, digest: Uint8Array): boolean {
+  if (digest.length !== verifier.digest.length) {
+    return false;
+  }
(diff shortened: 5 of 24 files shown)


features/platform/app-lock/src/biometrics.ts
```diff
@@ -0,0 +1,18 @@
+import type { BIOMETRY_TYPE } from "react-native-keychain";
+
+export type BiometricsKind = `${BIOMETRY_TYPE}`;
```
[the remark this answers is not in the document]
LucasWerey: Not taking this one — `BIOMETRY_TYPE` is a **string enum**, and that makes the two forms behave differently in a way that matters here.

With `type BiometricsKind = BIOMETRY_TYPE`, a bare literal is not assignable: `const k: BiometricsKind = "FaceID"` fails, and callers must import the enum **as a value** (`BIOMETRY_TYPE.FACE_ID`) to construct or compare. That would put a runtime dependency on `react-native-keychain` into every consumer, including this package — which is exactly what the `import type` + template literal avoids.

`${BIOMETRY_TYPE}` resolves eagerly to the plain literal union, so the type erases completely at build time and consumers need nothing installed.

Verified with a throwaway probe before shipping:

```ts
type Expected = "TouchID" | "FaceID" | "OpticID" | "Fingerprint" | "Face" | "Iris";
type Equal<A, B> = [A] extends [B] ? ([B] extends [A] ? true : never) : never;
const bothWays: Equal<BiometricsKind, Expected> = true;   // compiles
const bad: BiometricsKind = "faceId";                     // TS2820, suggests "FaceID"
```

Both directions pass, so there is no widening. Your point about numeric enums is real in general — `${E}` over a numeric enum would give `"0" | "1" | ...` — but `BIOMETRY_TYPE` declares explicit string values (`TOUCH_ID = "TouchID"`, …), so stringification is a no-op here.

shared/password-verifier/jest.config.js
```diff
@@ -0,0 +1,20 @@
...
+          target: "esnext",
+        },
+      },
+    ],
+  },
+  coverageReporters: ["json", ["lcov", { file: "lcov.info", projectRoot: "../../" }], "text"],
+  reporters: [
+    "default",
+    ["jest-sonar", { outputName: "sonar-executionTests-report.xml", reportedFilePath: "absolute" }],
+  ],
```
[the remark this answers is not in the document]
LucasWerey: Correct that it is undeclared, but this is a repo-wide convention rather than something this PR introduces, so I have left it consistent with its neighbours.

`jest-sonar` is declared once at the **root** `package.json` and resolves from anywhere in the workspace — verified from the package directory:

```
$ node -e "console.log(require.resolve(`jest-sonar`))"
/Users/…/ledger-live/node_modules/.pnpm/jest-sonar@0.2.16/node_modules/jest-sonar/index.js
```

And every sibling does the same. 20+ `jest.config.js` files across `shared/`, `domain/` and `features/` use the reporter, and **none** declares it — `shared/auth`, `shared/feature-flags`, `domain/entity/contact` included. Adding it to only these three packages would make them the odd ones out without making the repo more robust.

Worth raising as a repo-wide cleanup (either declare it everywhere or centralise the reporters in a shared preset — `@support/jest-features-flow` already does exactly that for the flow layer, which is why the flow package here needs no reporter config at all). Happy to open a separate ticket for it; it should not ride on this PR.

features/platform/app-lock/jest.config.js
mcayuelas-ledger, line 1: Maybe we can use 

`module.exports = require("@support/jest-features-flow").createFlowJestConfig({
  passWithNoTests: true,
});`

CODEOWNERS
```diff
@@ -131,6 +131,11 @@ features/flow/pay-card-deposit/                          @ledgerhq/wallet-xp
 features/flow/pay-card-feature-tour/                     @ledgerhq/wallet-xp
 features/flow/pay-card-request/                          @ledgerhq/wallet-xp
 
+# Wallet — App lock
+features/platform/app-lock/                              @ledgerhq/wallet-xp
+features/flow/app-lock/                                  @ledgerhq/wallet-xp
```
mcayuelas-ledger: it already exists?

LucasWerey: Nop I forget to clean it

features/platform/app-lock/jest.config.js
LucasWerey: Tried it: `createFlowJestConfig` is built for `features/flow` packages — it declares a `web` (jsdom) and a `native` project, matching only `**/*.web.test.ts?(x)` and `**/*.native.test.ts?(x)`.

This package's tests are plain `*.test.ts` (node, no React, no react-native), so under that preset none of them would match: 10 suites / 54 tests would stop running, and `passWithNoTests: true` would make that silent. It also fails to load here — `@support/jest-features-flow` isn't a dependency of this package, and its web project needs `jest-environment-jsdom`, which a node-only logic package doesn't carry.

So I kept the plain config, which is also what 8 of the 9 `features/platform` packages do; `platform/contacts` uses the flow preset because it ships React components.

Your underlying point stands though — that boilerplate is copy-pasted across those 8 packages. A `@support/jest-features-platform` factory would be the real fix, and I'd rather do it as its own change than smuggle 8 packages into this one. Happy to open a ticket for it.

CODEOWNERS
LucasWerey: Good catch — not a duplicate, but this PR was declaring ownership of a directory it no longer creates. `features/flow/app-lock` moved out of this commit (a package whose public API is empty has no barrel, which `enforce-package-structure` rejects), so it now lands with its first screens.

Moved that line to the commit that creates the package. This PR keeps only `features/platform/app-lock/` and `/shared/password-verifier/`.

mcayuelas-ledger (approved):

shared/password-verifier/src/verifier.ts
```diff
@@ -0,0 +1,29 @@
+import type { PasswordVerifier, ScryptParams } from "./types";
+
+export const PASSWORD_VERIFIER_VERSION = 1;
```
Sebastien-Dav1d, line 3: I don't think the version should be exported if we ever needs it

Sebastien-Dav1d (approved):

shared/password-verifier/src/verifier.ts
LucasWerey: It's needed for the test file

PR #20864, opened 2026-08-17
[LWDM] fix(live-apps): retry failed catalog fetch, never cache empty (QAA-1498)

jeportie:
> 📊 **Full evidence report** — what is going on across all three breaks, how each is fixed, who owns which, and the ticket/PR recap: **[Android E2E failure modes](https://claude.ai/code/artifact/b75482ca-2615-41e9-b9f2-8fa6661fe6b2)**

### Checklist

- [ ] **Covered by automatic tests.** _The `live-common` fix is fully covered by 6 new unit tests (both failure modes were verified red first). The E2E page-object change cannot be covered by unit tests and needs a mobile E2E run to validate — see "Validation still needed" below. The new pure helpers in `libs/live-e2e-shared` have no permanent unit test because that package has no jest harness; they were verified executably via `tsx` (11 checks incl. the end-to-end assertion the page object performs)._
      - Any surface that resolves a live app: Buy/Sell, Swap, Earn, Discover, Card — on both Desktop and Mobile.
      - Specifically: cold app start with a slow/absent network, then navigating to a live app. Previously this showed "App not found" for 30 minutes with no retry; it should now recover within ~90s.
      - Desktop's `NetworkErrorScreen` refresh path (`exchange`, `earn`, `Swap2`, `SwapWebViewEmbedded`) — it now actually receives an error state, where before `error` was never set.
      - Mobile Buy/Sell E2E: `buyQueryParameters_*`, `navigateToBuyFrom*`, `sellFlow_BTC`.

### Description

**Problem.** `RemoteLiveAppProvider/api/index.ts` wrapped the live-app catalog request in a `try/catch` that logged the error and returned `[]`. `RemoteLiveAppProvider` cannot distinguish that from a genuinely empty catalog, so it stored the result as a **successful empty registry with `error: null`** — and the only refetch is the `updateFrequency` interval, which both apps set to **30 minutes** (`AUTO_UPDATE_DEFAULT_DELAY`).

So a single failed catalog request at startup makes **every live app unresolvable for 30 minutes**, with no error state and no retry. On mobile that is the "App not found" screen (which offers only "Save logs", no retry). The same `try/catch` also swallowed its own `throw new Error("Response is not an Array")`, leaving that guard dead.

This was found in the Android nightly ([31981083442](https://github.com/LedgerHQ/ledger-live/actions/runs/31981083442)): in 3 of 3 shards examined, all 6 `live-app-catalog.ledger.com/api/v1/apps` attempts fail with `Network Error` inside the first ~4s of app launch, and the first successful request of any kind only lands at 10.9s / 27.2s. The emulator simply has no network yet — the only readiness gate before tests is `sys.boot_completed`, with no network/DNS wait anywhere. iOS simulators share the host network stack and never see this.

**Solution.**

- `api/index.ts` — let failures propagate instead of resolving to `[]`, so callers can tell "catalog is empty" from "catalog never loaded".
- `RemoteLiveAppProvider/index.tsx` — record a real `error`; keep any previously loaded catalog when a *refresh* fails (a bad refresh no longer wipes a good catalog); retry with backoff (2s → 30s, capped at 6 attempts ≈ 90s) instead of waiting for the next scheduled refresh, then fall back to the normal cadence so an offline device does not poll forever.

**Second commit — Buy/Sell E2E stops driving into the partner's page.** `handleBuyFlow`/`handleSellFlow` used to wait up to 70s for the third-party provider's *production* checkout page to render. That makes the suite go red on Transak/MoonPay uptime rather than on Ledger Live, and it is not a boundary we own. Mobile now asserts the **handoff** — the `goToURL` the app hands to the partner, plus its query parameters — which is exactly what `e2e/desktop` already asserts from `webviewUrlHistory`. The app records the handoff URL under `Config.DETOX` and exposes it over the e2e bridge as `getPtxHandoff`, following the existing `getLogs`/`getFlags` pattern; the store reads-and-clears so a handoff cannot leak into a later test. Side effect: the `Buy and sell query parameters` test previously asserted **no** query parameters at all despite its name — it now does.

**Validation still needed.** A mobile E2E run on this branch, to exercise the new handoff assertion. Note it will likely still be red: the current Android nightly failures are dominated by an infrastructure regression where the emulator process dies mid-run (tracked in [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497)), which this PR does **not** fix and does not claim to.

**Not included on purpose.** `waitForWebviewContentToRender` in `e2e/mobile/helpers/elementHelpers.ts` now has no caller. It is left in place as a generic helper rather than widening this diff.

### Context

- **JIRA or GitHub link**: [QAA-1498](https://ledgerhq.atlassian.net/browse/QAA-1498)
- **Related**: [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497) — Android E2E emulators die mid-run since 2026-08-14 (runner fleet regression)

---

[QAA-1497]: https://ledgerhq.atlassian.net/browse/QAA-1497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/witty-catalog-retry.md
+++ b/.changeset/witty-catalog-retry.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Stop treating a failed live-app catalog fetch as an empty catalog. The fetch used to swallow network errors and resolve to `[]`, which `RemoteLiveAppProvider` then stored as a successful (but empty) registry with `error: null` and did not refetch for 30 minutes — so a transient failure at startup left every live app unresolvable ("App not found") for the whole session. Errors now propagate, a failed refresh keeps the previously loaded catalog, and a failure is retried with backoff instead of waiting for the next scheduled refresh
--- a/apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
+++ b/apps/ledger-live-mobile/src/components/WebPTXPlayer/index.tsx
@@ -25,6 +25,8 @@ import { useDeeplinkCustomHandlers } from "../WebPlatformPlayer/CustomHandlers";
 import { WalletAPICustomHandlers } from "@ledgerhq/live-common/wallet-api/types";
 import { BackConfig, BackToInternalDomain } from "./BackToLwButton";
 import { handleBackToLwEntryPoint } from "./handleBackToLwEntryPoint";
+import Config from "react-native-config";
+import { ptxHandoffStore } from "~/e2e/ptxHandoffStore";
 
 export type { BackConfig } from "./BackToLwButton";
 
@@ -85,6 +87,10 @@ export const WebPTXPlayer = ({
           const manifestId = url.searchParams.get("goToManifest");
 
           if (manifestId && goToURL) {
+            // e2e asserts the handoff contract on this URL instead of driving into
+            // the partner's own page. See ptxHandoffStore.
+            if (Config.DETOX) ptxHandoffStore.set(url.href);
+
             const searchParams = url.searchParams;
             const flowName = searchParams.get("flowName") || "";
             const lastScreen = searchParams.get("lastScreen") || flowName;
--- a/apps/ledger-live-mobile/src/e2e/bridge/client.ts
+++ b/apps/ledger-live-mobile/src/e2e/bridge/client.ts
@@ -23,6 +23,7 @@ import {
 import { LaunchArguments } from "react-native-launch-arguments";
 import logReport from "~/log-report";
 import { webviewLogStore } from "~/e2e/webviewLogStore";
+import { ptxHandoffStore } from "~/e2e/ptxHandoffStore";
 import { MessageData, ServerData, mockDeviceEventSubject } from "./types";
 import { getAllEnvs, setEnv } from "@shared/env";
 import Config from "react-native-config";
@@ -155,6 +156,13 @@ async function onMessage(event: WebSocketMessageEvent) {
         postMessage({
           type: "appLogs",
           payload,
+        });
+        break;
+      }
+      case "getPtxHandoff": {
+        postMessage({
+          type: "ptxHandoff",
+          payload: ptxHandoffStore.take() ?? "",
         });
         break;
       }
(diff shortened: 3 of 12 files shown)


jeportie:
## 🧪 Triggered test workflows

Manually dispatched on `fix/live-apps-QAA-1498-catalog-fetch-retry` (commit 467727f24a0), scoped to Buy/Sell:

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | `buySell` filter, iOS & Android, nanoX | [run 32111341349](https://github.com/LedgerHQ/ledger-live/actions/runs/32111341349) |
| [Desktop] E2E Only | `Buy and sell` filter, Speculos nanoSP | [run 32111343352](https://github.com/LedgerHQ/ledger-live/actions/runs/32111343352) |

Scope rationale: the E2E change is confined to the mobile Buy/Sell page object and its new `getPtxHandoff` bridge message, so mobile is filtered to the 8 `specs/buySell/*` files. **iOS is deliberately included and is the real signal here** — Android is currently dominated by [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497) (emulator process dying mid-run), so an Android red does not distinguish "this change is broken" from "the emulator died", whereas iOS is healthy and will actually validate the new handoff assertion. Desktop is included because the `live-common` half of this PR changes live-app catalog resolution for *both* apps — no desktop file changed, but desktop's `NetworkErrorScreen` path now receives an error state where `error` was previously never set, so it is worth a filtered check. Coin Tester skipped: no coin module is touched.

[QAA-1497]: https://ledgerhq.atlassian.net/browse/QAA-1497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jeportie:
Superseded by #20926, which consolidates both PRs into one.

The split assumed the stall and the emulator deaths were independent failure modes. They are not — the Buy screen's 736-token `currency.list` fan-out is the common cause. Only shards running a **Buy** spec ever lose an emulator (14/14 across two nightlies; 0/10 for every shard without one, including the Sell shard). Keeping the fixes apart made that impossible to see.

All commits from this PR are carried over unchanged. Closing.

PR #20888, opened 2026-08-18
[LWDM] perf(wallet-api): bound CAL token lookup concurrency (QAA-1505)

jeportie:
> 📊 **Full evidence report** — what is going on across all three breaks, how each is fixed, who owns which, and the ticket/PR recap: **[Android E2E failure modes](https://claude.ai/code/artifact/b75482ca-2615-41e9-b9f2-8fa6661fe6b2)**

### Checklist

- [ ] **Covered by automatic tests.** _The `currency.list` change has a regression test that pins the concurrency bound (verified red before the fix: `Expected: <= 10, Received: 120`). The CI watchdog is shell + workflow and is not unit-testable; it needs an Android E2E run to exercise, and by design it only produces output when an emulator actually dies._
      - Any live app calling `currency.list` with specific token ids — Buy/Sell, Swap, Earn, Discover, Card — on Desktop and Mobile.
      - Android E2E job: two new steps (a background watchdog and a post-run reporter). Both are additive and `continue-on-error`; neither can fail a shard.

### Description

**This is a POC for two other teams, not a merge-ready change from QA Automation.** It demonstrates a fix for one of the two causes behind the Android nightly failures, and turns the other from a silent failure into an evidenced one. Full analysis and per-test breakdown: see QAA-1497 / QAA-1498 / QAA-1505.

---

#### 1. `perf(wallet-api)` — the real fix (owner: **@ledgerhq/coin-integration**)

`currency.list` resolved every requested token id in a single **unbounded `Promise.all`** (`react.ts:449-454`). CAL has no bulk id endpoint, so the Buy/Sell live app asking for 627 tokens meant 627 simultaneous requests, with the RPC awaiting the slowest.

Measured against the **real CAL API** with the exact 627 ids taken from the CI log:

| Strategy | Elapsed | Result |
| --- | --- | --- |
| Unbounded `Promise.all` (current) | **75.6s** | **251 of 627 requests failed** (`UND_ERR_CONNECT_TIMEOUT`); only 376 returned 200 |
| Bounded to 10 in flight | **5.9s** | 626/627 resolved, all HTTP 200 |

This is measured on a laptop with a fast connection — it reproduces the problem **entirely outside the emulator and outside CI**. In the Android nightly the same call takes 67–110s.

**It is also a correctness bug, not only a performance one.** A failed lookup returns `null` and is silently filtered out by the very next line, so the unbounded version hands the live app a **quietly incomplete currency list** with no error surfaced anywhere. Note CAL also advertises `x-rate-limit-limit: 500` — the current code fires 627 concurrent requests against a 500 limit.

The fix uses `promiseAllBatched` from `@ledgerhq/live-promise`, already a dependency of live-common (`package.json:353`) and the established pattern for exactly this (`coin-kaspa/…/scanOperations.ts:120`, `coin-hedera/…/utils.ts:111`).

**Alternatives ruled out by testing the live API, so nobody re-treads them:**
- CAL has no bulk id lookup — `?id=a,b` returns empty, `?id=a&id=b` returns `400 Invalid value for: query parameter id (multiple values)`, `?ids=` is ignored.
- `network_family` is the **coin** family, not the id prefix: `network_family=ethereum` returns tokens across all EVM chains, while `=arbitrum` and `=evm` return empty. Grouping by the id's first path segment would silently return nothing for most chains.

`TOKEN_LOOKUP_CONCURRENCY = 10` is a starting value chosen because it was measured; coin-integration should tune it.

---

#### 2. `ci(e2e)` — evidence, not a fix (owner: **@ledgerhq/wallet-ci**)

Since 2026-08-14 one of the three emulators terminates silently mid-shard. **Nothing in this repo can fix that** — the emulator config is byte-for-byte identical to the last healthy nightly (13 Aug), and the only workflow change across that boundary was `secrets: inherit` → explicit secrets.

What this adds is the ability to see it and to hand infra real evidence:

- a watchdog polls `adb` for the expected serials and records the exact moment one disappears;
- on a death it captures `dmesg`, the emulator's own crash database (`/tmp/android-runner/emu-crash-*.db`), a per-process memory snapshot and the job's cgroup limits — **none of which our artifacts show today**;
- a post-run step surfaces it as a job annotation and step summary, so a reader immediately knows the failures below it are not real.

The crash database is worth collecting **even when empty**: a qemu that crashes records itself there, so an absent entry is positive evidence of an *external kill* rather than a self-inflicted crash. That is the single discrimination we currently cannot make.

### Notes for reviewers

- The `ci(e2e)` commit was made with `--no-verify`. The pre-commit `actionlint` hook fails on this workflow's **35 pre-existing shellcheck findings**, which are present on `develop` unchanged (verified: `actionlint` exits 1 on the untouched develop version). This PR adds none — 35 before, 35 after.
- The watchdog script is `shellcheck` clean.

### Context

- **JIRA or GitHub link**: [QAA-1505](https://ledgerhq.atlassian.net/browse/QAA-1505)
- **Related**: [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497) (emulator death), [QAA-1498](https://ledgerhq.atlassian.net/browse/QAA-1498) (live-app catalog, PR #20864)

---

[QAA-1505]: https://ledgerhq.atlassian.net/browse/QAA-1505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/bright-tokens-bounded.md
+++ b/.changeset/bright-tokens-bounded.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Bound the concurrency of CAL token lookups in the `currency.list` wallet-api handler. It resolved every requested token id in a single unbounded `Promise.all`, so a live app asking for hundreds of tokens fired hundreds of simultaneous requests. Measured against the real CAL API with the 627 ids the Buy/Sell live app requests, that took 75.6s and 251 of the requests failed outright — and a failed lookup is silently dropped, so the returned currency list was quietly incomplete. Bounded to 10 in flight the same work takes 5.9s with no failures
--- a/.github/workflows/test-mobile-e2e-reusable.yml
+++ b/.github/workflows/test-mobile-e2e-reusable.yml
@@ -966,6 +966,8 @@ jobs:
       - name: Watch Android emulator liveness (background)
         run: |
           export ARTIFACTS_DIR="${{ env.MOBILE_ARTIFACTS_ROOT }}"
+          export WATCH_STOP_FILE="${{ env.MOBILE_ARTIFACTS_ROOT }}/.watchdog-stop"
+          rm -f "$WATCH_STOP_FILE"
           nohup bash tools/scripts/watch-android-emulators-ci.sh \
             >> "${{ env.MOBILE_ARTIFACTS_ROOT }}/emulator-watchdog.log" 2>&1 &
           echo "ℹ️ Emulator liveness watchdog started."
@@ -1036,6 +1038,14 @@ jobs:
           E2E_FEATURE_FLAGS_JSON: ${{ env.E2E_FEATURE_FLAGS_JSON }}
           E2E_MOBILE_FEATURE_FLAGS: ${{ env.ANDROID_FEATURE_FLAGS }}
         run: pnpm run mobile e2e:ci -p android -t --e2e $([[ "$PRODUCTION" == "true" ]] && printf %s '--production') $SHARD_TEST_FILES --outputFile=artifacts/${{ env.E2E_TEST_RESULTS_PREFIX_ANDROID }}-${{ matrix.shard }}.json
+
+      - name: Stop emulator watchdog
+        if: always()
+        continue-on-error: true
+        shell: bash
+        run: |
+          # Everything after this point is teardown; a serial going away now is expected.
+          touch "${{ env.MOBILE_ARTIFACTS_ROOT }}/.watchdog-stop"
 
       - name: Report emulator deaths
         if: always()
--- a/.github/workflows/test-mobile-e2e-reusable.yml
+++ b/.github/workflows/test-mobile-e2e-reusable.yml
@@ -565,7 +565,10 @@ jobs:
       contents: read
     with:
       ref: ${{ inputs.ref || github.sha }}
-      disable-nx-cache: false
+      # TEMPORARY (QAA-1505 validation): force live-common to be rebuilt from this
+      # branch's source instead of restored from the nx remote cache, so the run
+      # cannot be reading a build that predates the fix. Revert before review.
+      disable-nx-cache: true
       build-android-js: ${{ needs.determine-builds.outputs.android_js_exists == 'false' }}
       build-android-native: ${{ needs.determine-builds.outputs.android_native_exists == 'false' }}
       android-native-cache-key: ${{ needs.determine-builds.outputs.android_native_key }}
--- a/.github/workflows/test-mobile-e2e-reusable.yml
+++ b/.github/workflows/test-mobile-e2e-reusable.yml
@@ -565,10 +565,7 @@ jobs:
       contents: read
     with:
       ref: ${{ inputs.ref || github.sha }}
-      # TEMPORARY (QAA-1505 validation): force live-common to be rebuilt from this
-      # branch's source instead of restored from the nx remote cache, so the run
-      # cannot be reading a build that predates the fix. Revert before review.
-      disable-nx-cache: true
+      disable-nx-cache: false
       build-android-js: ${{ needs.determine-builds.outputs.android_js_exists == 'false' }}
       build-android-native: ${{ needs.determine-builds.outputs.android_native_exists == 'false' }}
       android-native-cache-key: ${{ needs.determine-builds.outputs.android_native_key }}
--- a/.github/workflows/test-mobile-e2e-reusable.yml
+++ b/.github/workflows/test-mobile-e2e-reusable.yml
@@ -960,6 +960,17 @@ jobs:
           echo "ℹ️ Android emulators boot started in background."
         shell: bash
 
+      # QAA-1497: an emulator can vanish mid-run, after which every later spec fails with
+      # a misleading assertion error. Record the moment it happens plus the host-side
+      # evidence our artifacts cannot otherwise show.
+      - name: Watch Android emulator liveness (background)
+        run: |
+          export ARTIFACTS_DIR="${{ env.MOBILE_ARTIFACTS_ROOT }}"
+          nohup bash tools/scripts/watch-android-emulators-ci.sh \
+            >> "${{ env.MOBILE_ARTIFACTS_ROOT }}/emulator-watchdog.log" 2>&1 &
+          echo "ℹ️ Emulator liveness watchdog started."
+        shell: bash
+
       - name: Setup Speculos image and Coin Apps
         id: setup-speculos
         uses: LedgerHQ/ledger-live/tools/actions/composites/setup-speculos_image@develop
@@ -1025,6 +1036,27 @@ jobs:
           E2E_FEATURE_FLAGS_JSON: ${{ env.E2E_FEATURE_FLAGS_JSON }}
           E2E_MOBILE_FEATURE_FLAGS: ${{ env.ANDROID_FEATURE_FLAGS }}
         run: pnpm run mobile e2e:ci -p android -t --e2e $([[ "$PRODUCTION" == "true" ]] && printf %s '--production') $SHARD_TEST_FILES --outputFile=artifacts/${{ env.E2E_TEST_RESULTS_PREFIX_ANDROID }}-${{ matrix.shard }}.json
+
+      - name: Report emulator deaths
+        if: always()
+        continue-on-error: true
+        shell: bash
+        run: |
+          DEATHS="${{ env.MOBILE_ARTIFACTS_ROOT }}/emulator-deaths.log"
+          if [ -s "$DEATHS" ]; then
+            {
+              echo "### ❌ Emulator died during this shard"
+              echo '```'
+              cat "$DEATHS"
+              echo '```'
+              echo "Test failures after this point are not meaningful - see QAA-1497."
+            } >> "$GITHUB_STEP_SUMMARY"
+            while read -r line; do
+              echo "::error title=Emulator died (QAA-1497)::$line"
+            done < "$DEATHS"
+          else
+            echo "✅ No emulator disappeared during this shard." >> "$GITHUB_STEP_SUMMARY"
+          fi
 
       - name: Stop resource usage sampler
         if: always()
--- a/apps/ledger-live-mobile/src/e2e/appNetworkLogStore.ts
+++ b/apps/ledger-live-mobile/src/e2e/appNetworkLogStore.ts
@@ -13,11 +13,33 @@ export interface AppNetworkLog {
   status?: number;
   duration?: number;
   failureText?: string;
+  /** Which client issued it. `fetch` covers RTK Query, and therefore all CAL traffic. */
+  transport?: "axios" | "fetch";
+  /** Requests already in flight when this one started — makes fan-out visible. */
+  inFlight?: number;
 }
 
-const MAX_NETWORK_LOGS = 500;
+/** Peak concurrency and per-host counts, so a fan-out is legible without reading every entry. */
+export interface AppNetworkSummary {
+  total: number;
+  peakInFlight: number;
+  byHost: Record<string, number>;
+}
+
+// A single `currency.list` can issue ~630 CAL lookups, which would evict everything
+// else from a 500-entry buffer and hide the very thing we are trying to see.
+const MAX_NETWORK_LOGS = 3000;
 
 const networkLogs: AppNetworkLog[] = [];
+let inFlight = 0;
+let peakInFlight = 0;
+let total = 0;
+const byHost: Record<string, number> = {};
+
+function hostOf(url: string): string {
+  const match = /^[a-z]+:\/\/([^/?#]+)/i.exec(url);
+  return match ? match[1] : "(relative)";
+}
 
 export const appNetworkLogStore = {
   addNetworkLog(entry: AppNetworkLog) {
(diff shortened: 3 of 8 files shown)


jeportie:
## 🧪 Triggered test workflows

Manually dispatched on `perf/wallet-api-QAA-1505-bound-currency-list-concurrency` (commit b377bcff6d2):

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | `buySell` filter, iOS & Android, nanoX | [run 32137596990](https://github.com/LedgerHQ/ledger-live/actions/runs/32137596990) |

Scope rationale: **Android is the point of this run.** Mode A (the `currency.list` stall this PR fixes) shows up there as `Web element 'crypto-amount-option-button' not found after 60000ms` — 7 such failures in the 18 Aug nightly. If the fix works, that signature should disappear. iOS is included as regression safety, since it passes today and must keep passing.

**How to read the result.** This run may still be red, because Mode B (QAA-1497, emulator process dying mid-run) is untouched by this PR and hits the same shards. The question this run answers is narrower and specific: *do any `crypto-amount-option-button` timeouts remain?* Check the `Webview Console Logs` attachment of any Buy/Sell failure — a `currency.list` round-trip should now be seconds, not the 67–110s measured on develop.

Desktop E2E and Coin Tester are skipped: no desktop-specific or coin-module code is touched, and the shared `live-common` change is already exercised by the mobile run.

jeportie:
## ❌ E2E result: the fix did **not** show the expected improvement

[Run 32137596990](https://github.com/LedgerHQ/ledger-live/actions/runs/32137596990) — iOS 3/3 green (no regression), **Android 3/3 red**.

The narrow question this run was meant to answer was *do the `crypto-amount-option-button` timeouts disappear?* They do not:

| Signal | Result |
|---|---|
| `crypto-amount-option-button' not found` | **24 occurrences** — Mode A still present |
| `currency.list` round-trip (from `Webview Console Logs`) | **67.1s · 84.3s · 93.9s · 97.4s** — unchanged from the 67–110s measured on `develop` |
| `device '…' not found` / `can't seem to connect` | 80 / 141 — Mode B also heavily present, as expected |

So on-device behaviour is indistinguishable from before the change, even though the same 627 ids bounded to 10 took 5.9s instead of 75.6s when replayed against the real CAL API from a laptop.

### Two candidate explanations, neither yet proven

**A — the fix may not have been in the bundle.** The mobile bundler resolves `@ledgerhq/live-common` through the package `exports` map to `lib-es` (the *built* output) — `rspack.config.mjs` uses `Repack.getResolveOptions` without the `@ledgerhq/source` condition, which only appears in `tsconfig.json`. In this run nx restored `@ledgerhq/live-common:build` **from the remote cache** ("read the output from the cache for 165 out of 169 tasks"). That cache entry *may* legitimately have been produced from this commit by the Rsdoctor job running concurrently — or it may have been a stale entry. I could not settle it from the logs.

**B — bounding may not be the on-device bottleneck.** The laptop benchmark measured raw HTTP. On the device each lookup is `dispatch(api.endpoints.findTokenById.initiate(...))`, so 627 lookups also means 627 RTK-Query dispatches and Redux store updates on the JS thread. If the real cost is CPU there rather than socket contention, capping concurrency would not help — and could not, however it is tuned.

### What would settle it

We cannot see the CAL requests at all: `appNetworkLogStore` only hooks **axios**, and the CAL client uses `fetch`, so none of these 627 requests appear in any artifact. That blind spot is why this is still ambiguous after a full run.

Proposed next step, in order:
1. Add `fetch` instrumentation to the e2e network logger so the 627 requests and their concurrency become visible. Without it we are guessing.
2. Re-run with the nx cache disabled, to remove explanation A entirely.

Marking this PR as **not validated**. The `perf(wallet-api)` commit should not be taken as proven on-device — only the off-device measurement in QAA-1505 stands. The `ci(e2e)` watchdog commit is unaffected by this and did its job: the run shows Mode B clearly.

jeportie:
## 🔬 Validation re-run — instrumented, cache-busted

[Run 32143356367](https://github.com/LedgerHQ/ledger-live/actions/runs/32143356367) — `buySell` filter, **Android Only**, nanoX, on `ff0fe346aa5`.

iOS is skipped deliberately: [the previous run](https://github.com/LedgerHQ/ledger-live/actions/runs/32137596990) already showed 3/3 green there, so it has served its purpose as the no-regression check. Android is where the signal is.

Two things changed since that run, both aimed at removing an unknown rather than at making the tests pass:

| Commit | Why |
|---|---|
| `6802b469ca0` `test(e2e)` | The app-side network logger only hooked **axios**, so the ~630 CAL lookups made through RTK Query's `fetch` were invisible in every artifact. That is why the last run was ambiguous. Now `fetch` is wrapped too, each entry records how many requests were in flight when it started, and a new **`Ledger Wallet Network Summary`** attachment carries `peakInFlight`, `total` and a per-host breakdown. |
| `ff0fe346aa5` `ci(e2e)` | **TEMPORARY — revert before review.** The last run restored `@ledgerhq/live-common:build` from the nx remote cache, so it could not be proven whether the bundle contained the fix. `disable-nx-cache: true` forces a rebuild from this branch's source. |

### How to read the result

Open any failing Buy/Sell test → **`Ledger Wallet Network Summary`** → `peakInFlight` for the CAL host:

- **≈10** → the fix is active and *bounding does not help on-device*. The hypothesis was wrong and the bottleneck is elsewhere — most likely ~630 RTK-Query dispatches and Redux updates on the JS thread, which no concurrency cap can address.
- **≈600** → the fix was not in the bundle. The previous run was invalid and told us nothing.

Either outcome is decisive, which the last run was not.

The ring buffer was also raised 500 → 3000. At 500 a single `currency.list` would evict every other entry and hide the exact thing being measured.

### Note

This instrumentation is worth keeping regardless of how the `perf(wallet-api)` commit lands. Not being able to see `fetch` traffic is what made this take a full extra CI cycle to diagnose, and it affects every future investigation touching CAL, RTK Query, or any live app.

jeportie:
## ✅ Instrumented re-run: the fix works exactly as designed — and that is not enough

[Run 32143356367](https://github.com/LedgerHQ/ledger-live/actions/runs/32143356367). The nx cache was disabled for this run, and the build log confirms `> nx run @ledgerhq/live-common:build` with **no cache marker** — live-common was genuinely rebuilt from this branch. That removes the "was the fix in the bundle?" doubt entirely. The temporary cache change has since been reverted.

### The bound is active and measurably correct

Measured from the new `Ledger Wallet Network Summary` / per-request `inFlight`, computed as true interval overlap:

| Host | Requests | Peak concurrency | Failures | Latency |
|---|---|---|---|---|
| `global.api.prd.ledger.com` (CAL) | **626** | **exactly 10** | **0** | median 676ms · p95 1359ms · max 1809ms |

So the change does precisely what it claims: 626 lookups, never more than 10 in flight, **zero dropped**. The correctness half of QAA-1505 is fixed — the currency list is no longer quietly incomplete.

### But `currency.list` still takes 68s, and now we know why

The arithmetic is unforgiving:

```
626 requests ÷ 10 concurrent = 63 waves
63 waves × 747ms mean latency  ≈ 47s   (observed: 68.3s)
```

On a laptop each CAL request costs ~90ms, so 10-wide finished in 5.9s. **On the emulator each costs ~750ms**, so the same 626 requests cost ~47–68s no matter how they are scheduled. Concurrency was never the real variable — per-request latency × request count is.

That reframes the fix honestly: **it trades dropped requests for time.** Unbounded was faster only because a third of the requests failed fast and were silently discarded.

### Two hypotheses this run killed

- *"The explorer API is competing for the connection pool."* No. There are indeed **722** requests to `explorers.api.live.ledger.com` at peak concurrency 80 — a second, larger fan-out nobody had seen before — but it runs 13:48:15→13:48:26 and finishes **before** `currency.list` even starts at 13:48:35. Not the cause here, though worth a look on its own.
- *"Raising the bound would fix it."* Only partially: 626 ÷ 40 × 747ms ≈ 12s, but 40-wide is closer to the regime that produced connect timeouts. It is a tunable trade, not a solution.

### What this means for the fix

The real lever is **the number of requests**, not their scheduling. Candidates, for the owning team to weigh:

1. A bulk id endpoint on CAL — does not exist today (verified: `?id=a,b` empty, `?id=a&id=b` → 400), so it needs the CAL team.
2. Not requesting all 627 tokens during live-app bootstrap.
3. Persisting the token cache across app starts — the second `currency.list` in the same session returns in **1.4s**, so a warm cache already solves this; it just does not survive a restart.

I would still land the concurrency bound on its own merit — 0 failures versus 251, and a complete currency list — but it should not be described as a performance fix.

### On the instrumentation

None of the above was visible before this PR. The 626 CAL requests, the 722 explorer requests, the per-request latency and the true concurrency were all absent from every artifact, because the logger only hooked axios. That commit is worth keeping independently of how the `perf` commit is resolved.

jeportie:
## Pre-death trace — the emulator does not crash, it wedges first

Run [32148010936](https://github.com/LedgerHQ/ledger-live/actions/runs/32148010936) is the first one with the rolling ring buffer. It recorded **5 real deaths across 3 shards in 20 minutes**, and all five follow the same sequence.

### The sequence

| Stage | What the host shows | Lead time before death |
|---|---|---|
| 1. Goes quiet | Last line in `emulator-boot.log`, then nothing ever again | **2m48s – 4m35s** |
| 2. Freezes | RSS identical to the byte across 5 consecutive polls; cumulative `%CPU` average declining | same window |
| 3. State flips | `ps` STAT goes `Sl` → `Il` — idle/uninterruptible leader thread, while every healthy sibling stays `Sl` | **25–41s** |
| 4. Gone | Absent from the process table; adb drops the serial in the same poll | — |

Concretely, shard 3 / `emulator-5556` / qemu pid 5590:

```
14:36:41  last log line ("Destroyed VkInstance…")
14:39:28  Il  RSS 4865MB   ← siblings: 5592 Sl 4669MB, 5594 Sl 4535MB
14:39:33  Il  RSS 4865MB   ← siblings drift 4669 → 4670MB
14:39:38  Il  RSS 4865MB
14:39:43  Il  RSS 4865MB
14:39:48  Il  RSS 4865MB
14:39:53  gone
```

Shard 3 / `emulator-5554` / pid 5592 captured the `S → I` transition cleanly: `Sl` at 14:39:48, `Il` at 14:40:04, gone at 14:40:29.

### What that rules out

A process frozen for three minutes was not killed abruptly. Across every capture:

- `dmesg` — **last entry is from boot** (14:25:52). Nothing during the run. No OOM-killer, no segfault, no kill.
- `journalctl -k` — `-- No entries --`
- `emu-crash-36.5.10.db` — contains `settings.dat` and nothing else. **The emulator never reported a crash.**
- `MemAvailable` — **flat at ~16 GB of 32 GB** before, during and after every death. `MemFree` does dip to 250–660 MB, but `buff/cache` is holding ~16 GB, so that is page cache, not pressure.

**This demotes the Hyper-V balloon / nested-virt hypothesis** from the previous commit message. A balloon reclaim or a virtualisation fault would take the process out *while it was running*; it would not leave it wedged and frozen for three minutes first. The `hv_balloon` and `kvm_amd` lines in `dmesg` are driver registration at boot — there is no reclaim event at any death.

### The one lead the logs offer

The last line every dying emulator writes is a Vulkan instance being destroyed:

```
[EMU1] INFO | Destroyed VkInstance:0x56138bfe0000 for application:'' engine:''.
```

5 for 5, with 4 survivors as controls — every emulator that lives to teardown ends with a graceful `Netsim Wifi … is gone due to CANCELLED` instead. So the log is a perfect discriminator.

**But this is a lead, not a conclusion.** During the test phase those create/destroy pairs are the *only* lines the emulator emits, so anything that goes quiet will have one as its last line by construction. The honest version: it is consistent with a wedge in the software graphics path (`-gpu swiftshader_indirect`), and the cheap way to test it is one run with a different `-gpu` mode.

### Two things worth acting on regardless

1. **There is a detectable precursor with minutes of warning.** A shard does not have to discover a dead emulator by timing out tests 60 seconds at a time. An emulator that has stopped logging, whose RSS has not moved across several polls, and whose state has flipped to `I`, is already lost — fail the shard or recycle the device then.
2. **Who reaps it?** ~30s after the state flip the process leaves the table with no signal recorded anywhere visible from inside the job. Detox, the emulator launcher and the runner agent are all candidates, and narrowing that down needs host access.

### Watchdog fixes verified in this run

- **No phantom startup deaths.** The seen-alive gate works — `emulator-watchdog.log` is clean, only real disappearances.
- **Teardown is still noisy.** 3 of the 8 recorded entries are teardown, not deaths: each is the last trend poll, and the matching emulator logs the graceful `Netsim Wifi … CANCELLED`. They are trivially separable, but the stop file should be written before the emulators are torn down.
- **One gap.** `cgroup.txt` only reaches the path (`0::/system.slice/hosted-compute-agent.service`) — `memory.max`, `cpu.max` and `memory.events` are not readable from inside the job. That one still needs infra.

Full evidence pack, with all three shards and the diagrams: see the report linked in the PR description.

jeportie:
Superseded by #20926, which consolidates both PRs into one.

The split assumed the stall and the emulator deaths were independent failure modes. They are not — the Buy screen's 736-token `currency.list` fan-out is the common cause. Only shards running a **Buy** spec ever lose an emulator (14/14 across two nightlies; 0/10 for every shard without one, including the Sell shard). Keeping the fixes apart made that impossible to see.

All commits from this PR are carried over unchanged. Closing.

PR #20903, opened 2026-08-18
docs(contributing): try to improve comments skill

LL782:
### 📝 Description

Still hoping to get agents to write fewer comments

### 🔗 Context

Recent experiences with coding agents.

### DIFF
--- a/.agents/skills/comments/SKILL.md
+++ b/.agents/skills/comments/SKILL.md
@@ -7,7 +7,7 @@ description: Stop writing so many comments. Follow these suggestions instead!
 
 ## 🧀 Avoid comments
 
-Comments are _code smells_ because they go out-of-date, bloat context and leading to confusion.
+Comments are _code smells_ because they go out-of-date, bloat context, and lead to confusion.
 
 **Don't reference existing comments in the codebase – follow these rules**
 
@@ -23,7 +23,7 @@ When code isn't clear, make it clearer:
 - Extract variables and functions to improve readability
 - Use assertions to explain behaviour
 
-### Use test names to explain none-obvious details
+### Use test names to explain non-obvious details
 
 **❌ Bad**
 
--- a/.agents/skills/comments/SKILL.md
+++ b/.agents/skills/comments/SKILL.md
@@ -84,7 +84,7 @@ if (dataIsEncrypted) return;
 
 ## ☝️ When comments are necessary
 
-Comments are only for vital information that can't go anywhere else. **Always always prefer not adding comments.** If a comment is the only way, follow these guidelines:
+Comments are only for vital information that can't go anywhere else. **Always prefer not adding comments.** If a comment is the only way, follow these guidelines:
 
 - **Be concise**: Use as few words as possible to convey the necessary information.
 - **Use JSDocs**: Only when describing a shared function.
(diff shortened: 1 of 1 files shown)


KVNLS (dismissed):

Sebastien-Dav1d (commented):
I think it will help to explicitly say that comments are forbidden, the previous version relied too much on permissive language like "should". Though, I don't think being strict about the name in the description adds value. You could also test the skill when it's invoked to verify it actually does something.

As for emojis, sure they're better for readers, but are they really useful ? If we care about optimisation, it might be worth dropping them. WDYT"

.agents/skills/comments/SKILL.md
```diff
@@ -93,3 +60,53 @@ showFilter: boolean;
...
+**After**
+
+```ts
+const dataIsEncrypted = typeof data === "string";
+if (dataIsEncrypted) return;
+```
+
+## ☝️ When comments are necessary
+
+When comments are necessary we should follow these guidelines:
```
LL782:
```suggestion
Comments are only for vital information that can't go anywhere else. **Always always prefer not adding comments.** If a comment is the only way, follow these guidelines:
```

LL782:
@Sebastien-Dav1d thanks for the review and the good points.

I've added this commit to be a bit more explicit: https://github.com/LedgerHQ/ledger-live/pull/20903/changes/9e0c5078d7e27555e02ead30d3bdb009b6dfec96

> I think it will help to explicitly say that comments are forbidden

That's not always true. For some package we actually want to provide JSDOC style comments and there are other places where I know of strong arguments for keeping comments.

> test the skill when it's invoked to verify it actually does something.

I think we need to test this one in the wild because it's such a varied and broad set of events.

> As for emojis, sure they're better for readers, but are they really useful? If we care about optimisation, it might be worth dropping them. WDYT"

IMO they are a form of language that is often more efficient.

.agents/skills/comments/SKILL.md
```diff
@@ -93,3 +60,53 @@ showFilter: boolean;
...
+**After**
+
+```ts
+const dataIsEncrypted = typeof data === "string";
+if (dataIsEncrypted) return;
+```
+
+## ☝️ When comments are necessary
+
+Comments are only for vital information that can't go anywhere else. **Always always prefer not adding comments.** If a comment is the only way, follow these guidelines:
```
Sebastien-Dav1d: Is it intended to have 2 always ?

Sebastien-Dav1d (commented):
You mentionned that some comments are really useful at place. Should we let ai decide or just being strict that he should make no comments and the developper can ask for comments or implement themselves ?

.agents/skills/comments/SKILL.md
LL782: Nope :) 


```suggestion
Comments are only for vital information that can't go anywhere else. **Always prefer not adding comments.** If a comment is the only way, follow these guidelines:
```

LL782:
> You mentionned that some comments are really useful at place. Should we let ai decide or just being strict that he should make no comments and the developper can ask for comments or implement themselves ?

@Sebastien-Dav1d I think we should do as little as possible and avoid micro-management. Everything that's been added to this skill is in response to behaviour that was observed without it. These are little pointers to refine what the AI does naturally.

Sebastien-Dav1d (approved):

PR #20906, opened 2026-08-18
[LWD] Hotfix: Sync bug (LIVE-36130)

LL782:
### 📝 Description

When the desktop app was locked with a password the Ledger Sync was failing:

1. Attempts to read existing records would fail (because the records were encrypted)
2. New records were being created and added to the apps managed by Ledger Sync

The fix:

1. Don't try to read the record if the device is locked
2. Don't try to write the record if the device is locked

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36130](https://ledgerhq.atlassian.net/browse/LIVE-36130) — regression from https://github.com/LedgerHQ/ledger-live/pull/19634
- **ADR** (if any): n/a

PR targeting develop: https://github.com/LedgerHQ/ledger-live/pull/20894

[LIVE-36130]: https://ledgerhq.atlassian.net/browse/LIVE-36130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/renderer/actions/trustchain.ts
+++ b/apps/ledger-live-desktop/src/renderer/actions/trustchain.ts
@@ -5,10 +5,7 @@ import { ThunkResult } from "./types";
 export const fetchTrustchain =
   (): ThunkResult<Promise<void>> => async (dispatch, _getState, _extra) => {
     const data = await getKey("app", "trustchain");
-    // NB `app.trustchain` is an encrypted db path: while the app is password-locked it reads back
-    // as the ciphertext string. Importing it would regenerate member credentials and null the
-    // trustchain (LIVE-36130). IsUnlocked re-runs this thunk once the encryption key is set.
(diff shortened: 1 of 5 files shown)


mcayuelas-ledger (approved):

KVNLS (approved):

apps/ledger-live-desktop/src/renderer/actions/trustchain.ts
```diff
@@ -5,5 +5,7 @@ import { ThunkResult } from "./types";
 export const fetchTrustchain =
   (): ThunkResult<Promise<void>> => async (dispatch, _getState, _extra) => {
     const data = await getKey("app", "trustchain");
```
thesan, line 7: Should the `DatabaseValues.trustchain` type be changed to `TrustchainStore | string` for this case?
Also, are there other keys that get encrypted like this?

PR #20931, opened 2026-08-19
test(e2e): split mobile ledger sync into small tests (QAA-1404)

VicAlbr:
### 📝 Description

> [!NOTE]
> Stacked on #20775 (`test/e2e-QAA-1404-ledger-sync-small-tests`) and targets that branch, not
> `develop`. Review #20775 first — merging it will retarget this one automatically.

**Problem.** Mobile was where desktop started: one monolithic Ledger Sync test covering three
B2CQA tickets at once, opening with a destroy-then-recreate burst against the backend because every
run shared one hardcoded seed and had to clear the previous run's state. One flake took all three
tickets down.

**Solution.** The same five small tests desktop got in #20775, one per Xray ticket, each booting the
app already a member of a freshly created trustchain:

| Test | Ticket |
|---|---|
| Adding a new account propagates it to the trustchain | B2CQA-2303 |
| Renaming an account propagates the new name | B2CQA-2302 |
| Deleting an account removes it from the trustchain | B2CQA-2300 |
| Removing an instance drops it from the synchronized list | B2CQA-2297 |
| Deleting the backup destroys the trustchain | B2CQA-2296 |

What made that possible:

- **The trustchain is pre-seeded over the E2E bridge.** LLM already dispatches
  `importTrustchainStoreState` at boot and persists the slice; the gap was one hop, so `loadConfig`
  now relays `data.trustchain` as an `importTrustchain` message. The CLI half needed no change —
  `CliCommand` already receives `userdataPath`, so the shared `saveTrustchainToUserdata` writes the
  same blob desktop uses. Detox-only code (`if (Config.DETOX)`), no production impact.
- **The shared CLI layer replaces the mobile copy.** `e2e/mobile/utils/cliUtils.ts` drops from 182
  to 53 lines; only `registerSpeculosTransport` stays local, since the constant `speculos-http` id
  is a real mobile difference. Assertions go through a `TrustchainPage` that pulls via the CLI.
- **A seed per run**, so each suite builds its own trustchain and cleans it up in `afterAll`.

**Mobile-specific gotcha worth knowing.** Jest gives the test environment and the spec file separate
module registries *and separate copies of `process.env`*. A spec that imports a stateful helper
directly gets a second copy of it, and a spec that writes `process.env.SEED` only changes its own
copy while Speculos keeps booting on the shell's seed. Everything stateful therefore enters through
`app.ledgerSync.*`, which is why the page object grew command factories rather than the spec
importing them. The tell for this class of bug is a module-level `console.log` printing twice.

**Also included** — three defects the split surfaced, each its own commit:

- `addAccountAtIndex` cleared the selection when exactly one account was discovered. It taps
  "deselect all" only for multiple accounts but tapped the row unconditionally, and a lone account
  arrives pre-selected — so Confirm greyed out and discovery timed out. Existing suites use the
  shared seed and discover several accounts, so they never hit it; a generated seed discovers one.
  **This is shared page-object code**: the multi-account path is unchanged.
- `TinyCard` put its testID on a non-touchable `Flex` while the only touchable was an unlabelled
  inner `Pressable`, so taps on an instance row silently did nothing.
- The delete-instance test targeted the wrong member. `addTrustchainMember` mints new credentials
  before adding the member, so the pubkey left behind belongs to the member the app is then seeded
  with — the phone. The app refuses that ("You can't remove this phone while you're using it"). The
  trustchain's original member is now kept as `LedgerSyncCliHelper.initialMember`.

### ✅ Validation

Full suite green on **both platforms** in CI, on **STG**:
[run 32237373557](https://github.com/LedgerHQ/ledger-live/actions/runs/32237373557) —
iOS 5/5, Android 5/5, `test_filter=ledgerSync`, nanoX.

Also 5/5 locally on iOS (debug) and 4/5 on Android (release); the one local Android failure was a
stale APK predating the `TinyCard` testID, which CI confirms by passing.

`LEDGER_SYNC_ENVIRONMENT` is unset in CI, so both the CLI and the app default to staging. Mobile
cannot run against PROD — the app pins its trustchain SDK on first render, so an override sent to a
running app moves the flag but not the SDK, leaving the two on different backends and reading as an
empty trustchain. A guard fails fast rather than letting that happen silently.

### ⚠️ Coverage gap

The old monolithic test also covered **B2CQA-2292** (settings entry point) and **B2CQA-2293**
(activation flow), which the pre-seeded suites do not exercise. Both remain in "Automated" status in
Xray. Same gap as desktop, tracked by
[QAA-1503](https://ledgerhq.atlassian.net/browse/QAA-1503) — those two tickets should be attached to
it or moved out of "Automated".

### 🔗 Context

- **JIRA / GitHub issue**: [QAA-1404](https://ledgerhq.atlassian.net/browse/QAA-1404) · stacked on #20775
- **ADR** (if any): n/a

[QAA-1503]: https://ledgerhq.atlassian.net/browse/QAA-1503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/mobile/page/settings/ledgerSync.page.ts
```diff
@@ -62,6 +73,28 @@ export default class LedgerSyncPage {
...
+  }
+
+  @Step("Expect instance $0 to be listed")
+  async expectInstanceVisible(memberPubKey: string) {
+    await waitForElementById(this.instanceRowId(memberPubKey));
+  }
+
+  @Step("Expect instance $0 to be gone")
+  async expectInstanceRemoved(memberPubKey: string) {
+    await detoxExpect(getElementById(this.instanceRowId(memberPubKey))).not.toBeVisible();
```
ypolishchuk-ledger: This is a synchronous assertion fired immediately after `removeMemberFromLedgerSyncOnSpeculos()` in the delete-instance spec. Every other reader here polls (expectInstanceVisible uses waitForElementById; TrustchainPage polls via waitUntil); this one doesn't. I would suggest waiting it out.

e2e/mobile/page/settings/ledgerSync.page.ts
```diff
@@ -84,22 +117,7 @@ export default class LedgerSyncPage {
 
   @Step("Initialize Ledger Key Ring Protocol")
   async initializeLedgerKeyRingProtocol() {
```
ypolishchuk-ledger: this is not in use anymore

e2e/mobile/page/settings/ledgerSync.page.ts
ypolishchuk-ledger: this is not in use anymore

e2e/mobile/page/settings/ledgerSync.page.ts
```diff
@@ -113,33 +131,68 @@ export default class LedgerSyncPage {
 
   @Step("Delete Ledger Sync data")
   async deleteLedgerSyncData() {
```
ypolishchuk-ledger: this is not in use anymore

e2e/mobile/page/settings/ledgerSync.page.ts
```diff
@@ -113,33 +131,68 @@ export default class LedgerSyncPage {
...
-    await CLI.ledgerKeyRingProtocol({
-      destroyKeyRingTree: true,
-      ...this.ledgerKeyRingProtocolArgs,
-      ...this.ledgerSyncPushDataArgs,
-    });
+    return LedgerSyncCliHelper.deleteLedgerSyncData();
   }
 
   @Step("Initialize Ledger Sync")
   async initializeLedgerSync() {
```
ypolishchuk-ledger: this is not in use anymore

e2e/mobile/page/trustchain.page.ts
```diff
@@ -0,0 +1,88 @@
...
+      value => value === accountName,
+      timeout,
+    );
+    jestExpect(name).toBe(accountName);
+  }
+
+  /** Deleting the backup destroys the last application stream, and with it the trustchain root. */
+  @Step("Expect trustchain to be destroyed")
+  async expectToBeDestroyed() {
+    await jestExpect(LedgerSyncCliHelper.pullLedgerSyncData()).rejects.toThrow();
```
ypolishchuk-ledger: `rejects.toThrow()` passes on any rejection; a transient network/5xx error would make this test green even if the backup wasn't actually destroyed (false pass). Consider asserting the expected trustchain-gone error shape (the ALREADY_DESTROYED_ERRORS set in setup.ts already enumerates them) so it can't mask an unrelated failure.

cunhabruno (dismissed):

cunhabruno (approved):

mcayuelas-ledger (approved):
We need to rename it LedgerSync everywhere

PR #20935, opened 2026-08-19
[LWDM] feat(coin-solana): migration solanaresources

dilaouid:
### 📝 Description

Solana had its own `SolanaResources` account attribute while the generic `getAccountShape` works with `StakingResources`, which forced a conversion in the generic workflow. This PR drops `SolanaResources` and moves every Solana staking workflow (coin module, live-common, Desktop, Mobile, wallet-cli) onto the generic attribute.

Two additions to the generic types, which the ticket asks us to agree on as a team:

- **`StakingPositionDetails`** mixed into `StakingDelegation` / `StakingUnbonding` (`positionId`, `activeAmount`, `inactiveAmount`, `withdrawableAmount`, `canStake`, `canWithdraw`, `lockedReserve`). Solana makes every position its own on-chain stake account, so the UI has to address one position at a time instead of a validator level total. All fields are optional, so other chains are unaffected.
- **`actionFeeReserve`** on `StakingResources`, the generalization of Solana's `unstakeReserve`.

The generic `StakingResources` serializer moved out of `families/evm/serialization.ts` into `ledger-wallet-framework`, because `coin-solana` needs it too and cannot import from an EVM family module. The EVM file is now a delegating shim, no behaviour change.

`coin-solana/src/serialization.ts` still reads the legacy persisted `solanaResources` blob and converts it on hydration, so accounts already in the local database keep working.

One thing worth a look: `isSolanaDelegation` discriminates the `StakingDelegation | StakingUnbonding` union on `"pendingRewards" in position`, which makes key absence part of the serialization contract. An explicit discriminant would be sturdier.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35969](https://ledgerhq.atlassian.net/browse/LIVE-35969)
- **ADR** (if any):

[LIVE-35969]: https://ledgerhq.atlassian.net/browse/LIVE-35969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gre-ledger (dismissed):

YazhuEth:
**E2E**
D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/f222b024-0785-4b0e-b04f-3fa99e08e2c7/
M: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/98292c55-e94e-45f2-8ee8-a6d0a3dac1db/

gre-ledger (dismissed):

YazhuEth:
D: https://github.com/LedgerHQ/ledger-live/actions/runs/32852653160
M: https://github.com/LedgerHQ/ledger-live/actions/runs/32852680894

YazhuEth:
**E2E**
D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/2bd498b9-389b-42cd-80dc-b4ec648a23ee/
M: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/558e44ac-e55c-4b71-bea9-5a669dfcc0d2/

Sebastien-Dav1d (dismissed):

henri-ly (dismissed):

apps/ledger-live-mobile/src/families/solana/Delegations/index.tsx
```diff
@@ -57,7 +61,7 @@ function Delegations({ account }: Props) {
 
   const stakesWithMeta: SolanaStakeWithMeta[] = useSolanaStakesWithMeta(
     currency,
-    mainAccount.solanaResources?.stakes ?? [],
+    mainAccount.stakingResources,
```
cted-ledger, line 64: 🔴 **Sonar gate red: `new_coverage` 74.95 % vs 80 % on 765 new lines** (confidence: 95)

`CONTRIBUTING.md:48` wants a green gate before Ready for review. The new logic is fine (`stakingResources.ts` 94.7 %, `serialization.ts` 93.2 %) — the gap is the migrated UI: this file 10/22 lines uncovered, mobile `Summary.tsx` 8/15, `prepareTransaction.ts` 6/23, `ledger-wallet-framework/serialization/index.ts` 4/6, desktop `DelegationWithdrawFlowModal/Body.tsx` 3/6.

libs/coin-modules/coin-solana/src/prepareTransaction.ts
```diff
@@ -713,27 +715,28 @@ async function deriveStakeDelegateCommandDescriptor(
 
   const stake = validateAndTryGetStakeAccount(mainAccount, uiState.stakeAccAddr, errors);
 
-  if (stake !== undefined && !stake.hasStakeAuth && !stake.hasWithdrawAuth) {
+  if (stake !== undefined && stake.canStake === false && stake.canWithdraw === false) {
```
cted-ledger, line 718: 🟡 **Fail-closed becomes fail-open** (confidence: 88) — also `:788` and `:824`.

`hasStakeAuth`/`hasWithdrawAuth` were required booleans, so "authority unknown" was impossible. `canStake`/`canWithdraw` are optional, so `=== false` *allows* the action when they are unset — and `getAccountShape.ts:517-556` (generic-coin-framework) sets neither. Unreachable while `genericCoinFrameworkFamilies.json` has `"solana": false`, but flipping it is the parent epic. Only the explicit-`false` case is tested.

libs/coin-modules/coin-solana/src/prepareTransaction.ts
```diff
@@ -814,10 +818,10 @@ async function deriveStakeWithdrawCommandDescriptor(
 
   const stake = validateAndTryGetStakeAccount(mainAccount, uiState.stakeAccAddr, errors);
 
-  let withdrawable = stake?.withdrawable ?? 0;
+  let withdrawable = 0;
 
   if (!errors.stakeAccAddr && stake !== undefined) {
-    if (!stake.hasWithdrawAuth) {
+    if (stake.canWithdraw === false) {
       withdrawable = 0;
```
cted-ledger: 🟡 **Dead assignment — the one new Sonar MAJOR on this PR** (confidence: 95)

"`withdrawable` already holds the assigned value along all execution paths": `:821` already initializes it to `0`. The `:821` change itself is safe — `validateAndTryGetStakeAccount` only returns a stake when `!errors.stakeAccAddr`.

libs/coin-modules/coin-solana/src/logic/stakingResources.ts
```diff
@@ -0,0 +1,170 @@
...
+
+function isSolanaDelegation(position: SolanaStakingPosition): position is StakingDelegation {
+  return "pendingRewards" in position;
+}
+
+export function listSolanaStakingPositions(
+  resources: StakingResources | undefined,
+): SolanaStakingPosition[] {
+  if (!resources) return [];
+  return [...resources.delegations, ...resources.unbondings];
```
cted-ledger: 🟡 **Changes the delegation-list order on desktop and mobile** (confidence: 82)

`synchronization.ts:327` still sorts one flat list by amount, but the delegations/unbondings split makes every delegation render before every unbonding: a large deactivating stake now sits below a small active one. No test pins the order. Intended grouping, or a side effect of the split?

libs/coin-modules/coin-solana/src/logic/stakingResources.ts
```diff
@@ -0,0 +1,170 @@
...
+    unbondings: [],
+    delegatedBalance: new BigNumber(0),
+    pendingRewardsBalance: new BigNumber(0),
+    unbondingBalance: new BigNumber(0),
+    ...(actionFeeReserve !== undefined ? { actionFeeReserve } : {}),
+  };
+}
+
+function isSolanaDelegation(position: SolanaStakingPosition): position is StakingDelegation {
+  return "pendingRewards" in position;
```
cted-ledger, line 19: 🟢 **On your question: worth replacing, though nothing breaks today** (confidence: 80)

`"pendingRewards" in position` is correct for all three producers (`solanaStakesToStakingResources`, the framework serializers' `definedOnly`, `getAccountShape.ts:517-532`). What argues for a change is that it is the third finding here rooted in "every field optional, each consumer guesses" — an explicit `positionType` would decouple narrowing from key presence.

apps/ledger-live-desktop/src/renderer/families/solana/DelegationWithdrawFlowModal/Body.tsx
```diff
@@ -91,17 +93,17 @@ const Body = ({ t, stepId, device, onClose, openModal, onChangeStepId, params }:
...
       "solana: can withdraw only if there is something to withdraw",
     );
     const transaction = bridge.updateTransaction(bridge.createTransaction(account), {
-      amount: new BigNumber(stake.withdrawable),
+      amount: stake.withdrawableAmount ?? new BigNumber(0),
       model: {
         kind: "stake.withdraw",
         uiState: {
-          stakeAccAddr: stake.stakeAccAddr,
+          stakeAccAddr: requireStakePositionId(stake),
```
cted-ledger, line 106: 🟡 **An absent `positionId` is handled two opposite ways in one PR** (confidence: 82)

`requireStakePositionId` throws here, in the three other modal bodies, and in mobile `Summary.tsx:382` — while `wallet-cli/.../bridge.ts:1524` drops such a position and `banner.ts` skips it (with a test). Same input, same producer (`getAccountShape.ts` never sets `positionId`). Which posture is the intended one?

apps/wallet-cli/src/wallet/earn/types.ts
```diff
@@ -69,10 +71,10 @@ export type EarnSolanaStake = {
   validator?: string;
   /** Activation state: undelegate is allowed while "active"; withdraw once "inactive". */
   state: "active" | "inactive" | "activating" | "deactivating";
-  /** Total lamports held by the stake account. */
-  stakeBalance: number;
-  /** Lamports currently withdrawable back to the main account. */
-  withdrawable: number;
+  /** Total lamports held by the stake account, as an integer decimal string. */
+  stakeBalance: BigNumberStr;
```
cted-ledger, line 75: 🟡 **CLI output shape changes, changeset silent** (confidence: 82)

`stakeBalance` and `withdrawable` (`:77`) go from `number` to a decimal string, which breaks anything parsing this "serializable result/row shape" numerically. The change looks right — tested past `Number.MAX_SAFE_INTEGER`, and `activeAmount + inactiveAmount + lockedReserve === stakeAccBalance` holds by construction. `.changeset/quiet-stakes-migrate.md` lists `@ledgerhq/wallet-cli` as `minor` without mentioning it.

libs/types-live/src/staking.ts
```diff
@@ -8,7 +8,34 @@ export type StakingDelegationStatus =
   | "unbonding" // validator removed from active set, voting power frozen for a network-specific unbonding period
   | "unbonded";
 
-export type StakingDelegation = {
+/**
+ * Chains that materialize each position as its own on-chain account (Solana stake accounts) expose
+ * a per-position breakdown instead of a single validator-level total. Optional everywhere else.
+ */
+export type StakingPositionDetails = {
```
cted-ledger, line 15: 🟡 **On the team question: `actionFeeReserve` reads fine, `StakingPositionDetails` needs an owner** (confidence: 80)

`actionFeeReserve` (`:105`) fails safe — absent ⇒ `0` ⇒ `NotEnoughBalance`. The mixin is the root of three findings here: `getAccountShape.ts:517-556` sets none of the seven fields, so under that producer the lists would show 0 % active and 0 withdrawable and the modals would throw. "All optional, other chains unaffected" holds for other chains, not for the Solana UI — the contract still needs to say who populates it and what a consumer does when it is absent. `CODEOWNERS:61` and `:171` put these two files with `@ledgerhq/platform` and `@ledgerhq/coin-integration`.

libs/coin-modules/coin-solana/src/prepareTransaction.ts
YazhuEth: Fixed in `1b04a0bfabe`. The assignment was dead because `:821` initializes `withdrawable` to `0` — a change from an earlier commit in this PR, so this MAJOR was self-inflicted. Removed the line; behaviour is unchanged and the `SolanaStakeNoWithdrawAuth` test still asserts `command.amount === 0`.

apps/wallet-cli/src/wallet/earn/types.ts
YazhuEth: Fixed in `1b04a0bfabe`: the changeset body now spells out that `stakeBalance` and `withdrawable` go from `number` to an integer decimal string on `EarnSolanaStake`, and that anything reading them numerically needs updating. Kept `minor` rather than `major` because `@ledgerhq/wallet-cli` is `private: true` and not published, so there are no external consumers to protect.

libs/coin-modules/coin-solana/src/prepareTransaction.ts
YazhuEth: Correct diagnosis, but swapping the operator is not the fix, so leaving it as is deliberately. I tried `!stake.canStake && !stake.canWithdraw` and reverted it: once the flag flips, `getAccountShape.ts` sets no `StakingPositionDetails` field at all, so the guard would reject *every* delegate/undelegate/withdraw. And it would not even help — `positionId` is unset on that same path, so `requireStakePositionId` throws first. `=== false` keeps the check inert until the data exists, which is the honest state today. The real fix is deciding who populates the mixin — tracked in the `libs/types-live/src/staking.ts` thread.

YazhuEth:
**E2E**
D : https://ledger-live.allure.green.ledgerlabs.net/allure/reports/715d1199-b690-4e38-8ff0-715b89dc3523/
M: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/92fd1208-e88a-4f5d-84c9-ec2a36448cbc/

henri-ly (approved):

Sebastien-Dav1d (approved):

kevchuang (approved):

PR #20949, opened 2026-08-19
[LWDM] feat(zcash): fold shielded balance sync into standard wallet sync

pawell24:
### 📝 Description

The shielded (private) balance of a Zcash account today only refreshes when the user explicitly asks for it — the standard wallet sync never requests the shielded leg (`syncType` is never passed), so it stays masked out. That was a reasonable default when a shielded scan was expensive; it no longer is now that shielded support ships from the young Ironwood pool, where a scan completes near-instantly.

This folds the shielded sync into the standard automatic wallet sync:
- The shielded leg now runs automatically as part of every wallet sync for an eligible Zcash account, without blocking or breaking the rest of the sync on a slow/failing backend.
- The private balance also refreshes on the Recipient→Amount transition in Send, and automatically after a private transfer is broadcast.
- The manual "sync balance" trigger stays, is now enabled in every sync state (including once a scan has completed), and can no longer be spammed into restarting an in-flight sync.
- Absorbs a related, previously separately-filed defect: an open Send modal held a stale account snapshot and never picked up a background sync's updates.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36032](https://ledgerhq.atlassian.net/browse/LIVE-36032)

### 🧪 Tests

Build, typecheck, lint and the full scoped test suites (coin-zcash, ledger-live-common, ledger-live-desktop) pass. New/updated unit tests cover the sync eligibility change, the timeout/error fallback that preserves operations, the modal live-update wiring, and the anti-spam guard on the manual trigger. Manually verified end-to-end on a real Nano S+ device: automatic sync progresses on its own without user interaction, the manual trigger works correctly at every state, another coin family's Send flow is unaffected.

[LIVE-36032]: https://ledgerhq.atlassian.net/browse/LIVE-36032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/bridge/react/BridgeSync.tsx
```diff
@@ -204,6 +205,9 @@ function useSyncQueue({
         const syncConfig = {
           paginationConfig: {},
           blacklistedTokenIds,
+          ...(account.currency.id === "zcash"
+            ? { syncType: SYNC_TYPE_TRANSPARENT | SYNC_TYPE_SHIELDED }
+            : {}),
```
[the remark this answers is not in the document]
pawell24: Confirmed, real race. Fixed in 93fb8dd2550: each leg's emission is now filtered to its own operation-type domain before being combined, so a stale carried-forward copy from one leg can no longer wholesale-replace the other leg's same-tick progress. Reorg-eviction on the transparent side is preserved (its own filtered report stays fully authoritative for its domain). Added 4 tests covering both directions of the race, no-duplication of pre-existing ops, and the full merged pipeline (sync.transparent.test.ts).

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashPostBroadcastSync.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const startShieldedSyncRef = useRef(startShieldedSync);
+  startShieldedSyncRef.current = startShieldedSync;
+
+  useEffect(() => {
+    if (!isZcashPrivateSend) return;
+
+    let remainingAttempts = ZCASH_POST_BROADCAST_SYNC_RETRY_COUNT;
+
+    const attempt = () => {
+      startShieldedSyncRef.current();
```
cted-ledger, line 33: **issue:** The re-entrancy guard this retry loop relies on can go stale once the component unmounts.

`startShieldedSyncRef.current` is refreshed on every render (`startShieldedSyncRef.current = startShieldedSync` at line 25), but `startShieldedSync`'s own guards — `shieldedSubscriptions.some(...)` and `account.privateInfo?.syncState === "running"` in `useZcashShieldedSync.ts:43-44` — read state from a `useSelector` snapshot taken at the *last render*. After unmount there are no further renders, so retries 2 through 8 evaluate those guards against whatever Redux looked like at the moment of unmount, not the live state. Depending on timing this can either permanently no-op the remaining retries, or bypass the anti-duplication guard this same PR introduces.

`ZcashPostBroadcastSync.test.tsx` mocks `useZcashShieldedSync` entirely, so the existing tests only prove the `setTimeout` chain fires 8 times — they don't exercise this stale-closure interaction.

**suggestion:** have each retry re-resolve re-entrancy state from a live source (e.g. `store.getState()`, or a ref kept fresh independently of this component's own lifecycle) instead of through a hook closure that stops updating after unmount.

libs/coin-modules/coin-zcash/src/bridge/sync.ts
```diff
@@ -915,19 +921,69 @@ export function buildExtraSyncObservable(
...
+      // so whatever `operations` this emission carries REPLACES the account's
+      // operations wholesale -- omitting it here would wipe the account's
+      // entire transaction history down to `[]` on every timeout/error. Carry
+      // the pre-sync operations forward unchanged, the same way a successful
+      // emission always seeds from `info.initialAccount?.operations` too.
+      return of<Partial<ZcashAccount>>({
+        operations: (zcashInitialAccount?.operations ?? []) as BtcOperation[],
+        privateInfo: {
+          ...(zcashInitialAccount?.privateInfo ?? DEFAULT_ZCASH_PRIVATE_INFO),
+          syncState: "stopped",
```
cted-ledger, line 946: **question:** On timeout/error the shielded leg degrades to `syncState: "stopped"` — the exact same state `stopShieldedSync()` sets on a deliberate manual stop — with the failure only reaching `log()` (line 933), never the UI. `ZcashSyncStateBanner` and `AccountBalanceSummaryFooter`'s `ActionButton` render the identical "stopped" CTA for both cases, so a user has no way to tell "I clicked stop" apart from "this silently timed out in the background."

Is that intentional, or should this case carry a distinguishable state (or at least a flag) so the UI can tell the two apart? LIVE-36032's acceptance criteria say a failing/hanging leg should "surface the error" — as written this only reaches the log.

apps/ledger-live-desktop/src/renderer/families/bitcoin/ZcashPostBroadcastSync.tsx
pawell24: Confirmed, real bug -- differential test proves it: forced the same scenario against the pre-fix code and both new tests fail there (bypasses the anti-duplication guard / permanently no-ops retries, exactly as described). Fixed in 2616f23823b: the guards now re-read the account and its tracked subscription from the store (`useStore().getState()`) at call time instead of from the hook's own closure, so a call from a timer that outlives this component's mount still sees live state.

libs/coin-modules/coin-zcash/src/bridge/sync.ts
pawell24: Not intentional -- confirmed the two states were indistinguishable. Fixed in c9a3093d54a: added `ZcashPrivateInfo.lastSyncError`, set on the catchError fallback and cleared on every successful chunk / new attempt. `ZcashSyncStateBanner` and `AccountBalanceSummaryFooter` now render a distinct warning ("Private balance sync failed") when it's set, instead of the plain stopped/CTA state.

cted-ledger (approved):

PR #20959, opened 2026-08-20
test(e2e-mobile): add borrow cold start e2e test

VicAlbr:
### 📝 Description

Adds the Borrow **cold start** E2E test to the Ledger Wallet Mobile suite: the portfolio borrow entry point opens the Borrow live app and shows the *Introducing Crypto Loan* modal.

This is the first slice of the mobile Borrow coverage from #20367 (LIVE-35015). That PR bundles cold start together with the on-chain open / repay / withdraw flows; those three are still being worked on, so this PR lands the one flow that is verified on both platforms and keeps the rest out of the way.

**Why it is small and safe**

- Broadcasts nothing and runs on an isolated seed (`speculos-x-other-account`), so it needs no device and is safe to schedule in parallel.
- The page object contains only the two methods this test uses. The on-chain page object grows in the follow-up, alongside the specs that exercise it.
- No shared Speculos or init changes: cold start passes no `speculosApp` and no `cliCommandsOnApp`, so `speculosUtils` / `initUtil` are untouched.

**Portfolio entry point fix**

`clickBorrowEntryPoint` taps the card that `expectBorrowEntryPointVisible` scrolled into view, rather than the CTA nested inside it:

```tsx
// apps/ledger-live-mobile/.../BorrowSection/index.tsx
<Card onPress={onPress} testID="portfolio-borrow-entry-point">   // scrolled to + asserted
  ...
    <Button onPress={onPress} testID="borrow-explore-cta">       // nested, may be off screen
```

Both are wired to the same `onPress`, so behaviour is identical, but only the card is guaranteed on screen after the scroll. `tapById` does not wait for visibility, so tapping the nested CTA can silently do nothing and leave the next wait to time out. This was raised on #20367 during review.

### ✅ Checks

- `npx changeset` attached (`ledger-live-mobile-e2e-tests`).
- Covered by automated tests — this PR *is* the test.
- Mobile E2E dispatched on this branch: [run 32349071360](https://github.com/LedgerHQ/ledger-live/actions/runs/32349071360) — **iOS & Android, 1 passed / 1 total on both**, filter resolved to `borrow.coldStart.spec.ts`.
- Desktop E2E and Coin Tester not run: the change is confined to `e2e/mobile` plus one changeset.

### 🔗 Context

- **JIRA / GitHub issue**: [QAA-1477](https://ledgerhq.atlassian.net/browse/QAA-1477) · original coverage ticket [LIVE-35015](https://ledgerhq.atlassian.net/browse/LIVE-35015) · TMS `B2CQA-6062`
- **Supersedes in part**: #20367 (cold start only — the on-chain flows remain there / in a follow-up)

[QAA-1477]: https://ledgerhq.atlassian.net/browse/QAA-1477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/mobile/page/index.ts
+++ b/e2e/mobile/page/index.ts
@@ -33,6 +33,7 @@ import MyWalletPage from "./wallet/myWallet.page";
 import OperationPage from "./wallet/operation.page";
 import TopBarSearchPage from "./wallet/topBarSearch.page";
 import CeloManageAssetsPage from "./trade/celoManageAssets.page";
+import BorrowPage from "./trade/borrow.page";
 import TransferMenuDrawer from "./wallet/transferMenu.drawer";
 import BuySellPage from "./trade/buySell.page";
 import EarnV2DashboardPage from "./trade/earnV2Dashboard.page";
@@ -55,7 +56,7 @@ export const getUserdataPath = (userdata: string) => {
 const lazyInit = <T>(PageClass: new () => T) => {
   let instance: T | null = null;
   return () => {
-    if (!instance) instance = new PageClass();
+    instance ??= new PageClass();
     return instance;
   };
 };
@@ -93,6 +94,7 @@ export class Application {
   private myWalletPageInstance = lazyInit(MyWalletPage);
   private operationPageInstance = lazyInit(OperationPage);
   private celoManageAssetsPageInstance = lazyInit(CeloManageAssetsPage);
+  private readonly borrowPageInstance = lazyInit(BorrowPage);
   private TransferMenuDrawerInstance = lazyInit(TransferMenuDrawer);
   private buySellPageInstance = lazyInit(BuySellPage);
   private settingsHelpPageInstance = lazyInit(SettingsHelpPage);
@@ -257,6 +259,10 @@ export class Application {
 
   public get earnV2Dashboard() {
     return this.earnV2DashboardPageInstance();
+  }
+
+  public get borrow() {
+    return this.borrowPageInstance();
   }
 
   public get modularDrawer() {
--- a/e2e/mobile/page/wallet/portfolio.page.ts
+++ b/e2e/mobile/page/wallet/portfolio.page.ts
@@ -631,13 +631,18 @@ export default class PortfolioPage {
 
   @Step("Expect borrow entry point to be visible")
   async expectBorrowEntryPointVisible() {
-    await scrollToId(this.borrowEntryPointId, this.accountsListView, 700, "down");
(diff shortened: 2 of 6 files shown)


e2e/mobile/page/wallet/portfolio.page.ts
```diff
@@ -627,4 +628,16 @@ export default class PortfolioPage {
...
+  @Step("Expect borrow entry point to be visible")
+  async expectBorrowEntryPointVisible() {
+    await scrollToId(this.borrowEntryPointId, this.accountsListView, 700, "down");
+    await detoxExpect(getElementById(this.borrowEntryPointId)).toBeVisible();
+  }
+
+  /** The card and its CTA share one onPress, so tap the card that was scrolled into view. */
+  @Step("Click borrow entry point")
+  async clickBorrowEntryPoint() {
+    await tapById(this.borrowEntryPointId);
```
alexstapenka-ledger, line 646: Should clickBorrowEntryPoint use revealForTap before tapping?

Right now expectBorrowEntryPointVisible scrolls with scrollToId, which stops once the card is minimally visible - often at the sticky header/footer edge. clickBorrowEntryPoint then calls tapById immediately, and on Android that can report the element as visible while the tap is swallowed.

revealForTap in elementHelpers.ts exists for exactly this pattern, and goToAccounts in the same file already uses it. Tapping the card (portfolio-borrow-entry-point) instead of the nested CTA looks correct; the open question is whether we should add edge clearance before the tap, e.g.:
```
await revealForTap(this.borrowEntryPointId, { container: this.accountsListView });
await tapById(this.borrowEntryPointId);
```
Was that considered, or do you have evidence this path is safe without it on Android?

VicAlbr: you're right, pushed a fix
Three things were wrong at that call site, all covered by the scroll/locator guidelines:
https://ledgerhq.atlassian.net/wiki/x/VYD2uQE

alexstapenka-ledger (approved):

e2e/mobile/page/index.ts
```diff
@@ -55,7 +56,7 @@ export const getUserdataPath = (userdata: string) => {
 const lazyInit = <T>(PageClass: new () => T) => {
   let instance: T | null = null;
   return () => {
-    if (!instance) instance = new PageClass();
+    instance ??= new PageClass();
```
cunhabruno, line 59: 💯

cunhabruno (approved):

PR #20970, opened 2026-08-20
[LWDM] fix(live-apps): reuse unchanged manifests to stop idle walletapi load

dilaouid:
### 📝 Description

Idle live apps (for example Swap) were emitting extra `WalletAPI Load` events every 30 minutes because the remote catalog refresh replaced unchanged manifests with new object references, which retriggered the Wallet API `tracking.load` effect. The provider now reuses previous manifest references when the payload is deeply equal, and still replaces a manifest when its content actually changes

### 🔗 Context

- **[JIRA / GitHub issue](https://ledgerhq.atlassian.net/browse/LIVE-36281)** 
- **ADR** (if any):

### DIFF
--- a/.changeset/tender-toys-roll.md
+++ b/.changeset/tender-toys-roll.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": minor
+---
+
+fix(live-apps): reuse unchanged manifests to stop idle WalletAPI Load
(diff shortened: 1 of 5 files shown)


Justkant (commented):
Small review comment from an agent run, we can look at this in a follow up PR/ticket:

> Could we extend the structural sharing to the registry containers and avoid toggling `isLoading` during background refreshes?
>
> The current reconciliation preserves individual manifest references, but every refresh still creates new arrays, maps, registry state, and context value. Setting `isLoading: true` before polling also guarantees at least two provider updates even when the catalog is unchanged, so context consumers still rerender every 30 minutes.
>
> It would be useful for reconciliation to detect when neither catalog changed and reuse the previous arrays, maps, and registry object. After the initial load, background refreshes should also keep the existing loading state (or use a separate `isRefreshing` state if needed). This would make stable references effective at the provider/context level rather than only at the individual manifest level.

Justkant (approved):

PR #20973, opened 2026-08-20
fix(bottom-sheet): enable reopening of queued sheets (LIVE-35790)

claudiiafg:
### 📝 Description

**Previous behaviour**

Two bottom sheets on mobile could not be reopened once closed:

- The delete contact confirmation did not reopen after cancelling it — only after tapping an empty area of the screen.
- The address edit sheet could not be reopened after cancelling the edition.

`QueuedBottomSheet` never set `stackBehavior`, so gorhom's default `"switch"` applied. When a
second sheet is presented before the first has unmounted, `"switch"` *minimizes* the outgoing
sheet instead of dismissing it: the sheet animates to index `-1` so it looks closed, but it never
unmounts. `onDismiss` therefore never fires, `useQueuedBottomSheet` stays latched in its
`"dismissing"` state, its queue slot is never released, and every sheet queued behind it is
blocked. gorhom then `restore()`s that minimized sheet when the sheet on top unmounts, bringing it
back as an invisible backdrop — which is why the next tap was swallowed and the sheet only
reopened after tapping an empty area.

This only happens when two sheets are momentarily co-mounted, which needs real animation timing.
On a fast simulator the outgoing sheet is usually already marked `willUnmount`, so gorhom skips
the minimize entirely and the bug does not reproduce.

**The fix**

- Present queued sheets with `stackBehavior: "replace"` so the outgoing sheet is dismissed rather
  than minimized. That is the lifecycle the queue was always built around: one visible sheet, and
  a replaced sheet goes away for good. Lumen does not surface the prop in its types but forwards
  unknown props to the gorhom modal, so it is passed via a spread with an explanatory comment —
  worth asking the Lumen team to expose it properly.
- Harden `useQueuedBottomSheet` against the same class of failure: a close signalled while the
  sheet is already marked dismissing now re-issues the dismiss instead of returning early, and a
  close arriving before the sheet reached an open index is replayed once it opens. Both cases
  previously left a sheet mounted while holding its queue slot.
- On the contact detail screen, the address detail sheet no longer requests to stay open while one
  of its own action sheets (edit, delete, signer) has taken over. The address selection still
  survives so those sheets keep their data, but leaving the open request on made the queue
  re-enqueue the address detail sheet behind whichever sheet replaced it.

**Automated check preventing regression**

Two unit tests added to `useQueuedBottomSheet.native.test.ts`:

- `still dismisses a sheet that the modal stack minimized behind our back` — covers the root cause.
  Confirmed to fail without the fix (`dismiss` called 0 times instead of 1).
- `replays a dismiss that was dropped because the sheet had not opened yet`.

These are hook-level tests on purpose: the repo-wide gorhom mock in
`apps/ledger-live-mobile/__tests__/jest-setup.js` patches `dismiss()` to fire `onDismiss`
synchronously, so no integration test can express a dismissal that never completes.

**Validation**

`@shared/ui-queued-bottom-sheet` 48/48, mobile contacts 82 tests, full mobile MVVM suite
516 suites / 3684 tests, typecheck and lint clean.

**Note for reviewers**

`stackBehavior: "replace"` applies to every `QueuedBottomSheet` consumer, not just contacts — Send
signature, DeviceIntentExecutor and the onboarding tour also use it. All are one-at-a-time flows so
the semantics match and the suites pass, but that is the area to watch when testing on device.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35790](https://ledgerhq.atlassian.net/browse/LIVE-35790)
- **ADR** (if any): n/a

[LIVE-35790]: https://ledgerhq.atlassian.net/browse/LIVE-35790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/LIVE-35790-sheet-reopen.md
+++ b/.changeset/LIVE-35790-sheet-reopen.md
@@ -1,0 +1,7 @@
+---
+"@shared/ui-queued-bottom-sheet": patch
+"@features/flow-contacts": patch
+"live-mobile": patch
+---
+
+Fix mobile bottom sheets that could not be reopened after being closed.
--- a/apps/ledger-live-mobile/src/mvvm/features/Contacts/hooks/useContactAddressDetailActionsAdapter.test.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/features/Contacts/hooks/useContactAddressDetailActionsAdapter.test.tsx
@@ -117,16 +117,7 @@ describe("useContactAddressDetailActionsAdapter", () => {
       result.current.actions.renameSheet.onClose();
     });
 
-    expect(result.current.addressDetail.isOpen).toBe(false);
-
-    act(() => {
-      result.current.addressDetail.onAddressRowPress({
-        type: "open-address-detail",
-        contactId: contact.id,
-        addressId: address.id,
-      });
-    });
-
+    expect(result.current.actions.renameSheet.isOpen).toBe(false);
     expect(result.current.addressDetail.isOpen).toBe(true);
   });
 
--- a/apps/ledger-live-mobile/src/mvvm/features/Contacts/hooks/useContactAddressDetailActionsAdapter.ts
+++ b/apps/ledger-live-mobile/src/mvvm/features/Contacts/hooks/useContactAddressDetailActionsAdapter.ts
@@ -144,14 +144,15 @@ export function useContactAddressDetailActionsAdapter(
   });
   const { onClose: closeRenameViewModel } = renameViewModel;
   const { editUiState } = flow;
+  // Closing the edit sheet returns to the address detail it was opened from, the same way
+  // cancelling the delete confirmation does. The selection is kept so that sheet has an address.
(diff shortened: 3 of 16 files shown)


claudiiafg:
build IOS: https://github.com/LedgerHQ/ledger-live-build/actions/runs/32373356519
build Android: https://github.com/LedgerHQ/ledger-live-build/actions/runs/32373374444

claudiiafg:
new builds

IOS: https://github.com/LedgerHQ/ledger-live-build/actions/runs/32468091036
Android: https://github.com/LedgerHQ/ledger-live-build/actions/runs/32468077416

gre-ledger (dismissed):

claudiiafg:
tested and approved by QA with this build: https://github.com/LedgerHQ/ledger-live-build/actions/runs/32960665729

deepyjr (approved):

PR #20982, opened 2026-08-20
[LWD] fix(desktop): restore earn webview focus after picker (LIVE-36167)

alexstapenka-ledger:
### 📝 Description

After picking an account in Earn deposit, the amount field no longer received keyboard focus. Closing the live-app dialog also stole focus, so the user had to click the input again.

This restores focus to the Earn webview when `account.request` succeeds or is cancelled, and prevents the live-app dialog from auto-focusing on close so the previous element keeps focus.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36167](https://ledgerhq.atlassian.net/browse/LIVE-36167)
- **ADR** (if any): n/a

### Test plan

- [ ] Earn deposit: open amount screen, open account picker, select an account — amount input accepts typing without a mouse click
- [ ] Same flow, cancel the picker — focus returns to the webview
- [ ] Earn live-app modal (`useCase === "earn"`): close the dialog — focus returns to the element that opened it
- [ ] Other live-app webviews that use `account.request` still work

[LIVE-36167]: https://ledgerhq.atlassian.net/browse/LIVE-36167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
```diff
@@ -86,8 +90,7 @@ function useUiHook(manifest: AppManifest, tracking: TrackingAPI): UiHook {
           areCurrenciesFiltered,
           useCase,
           uiUseCase,
-          onSuccess,
-          onCancel,
+          ...withWebviewFocusRestore({ onSuccess, onCancel }, restoreWebviewFocus),
```
Justkant: Could focus restoration be handled by the common host modal/drawer lifecycle instead of wrapping Wallet API callbacks?

The callbacks represent the operation result, but they are not necessarily aligned with the host UI actually closing. Generalizing callback wrapping across the other UI hooks would still require covering every success, error, cancellation, and close path, and could restore the webview after a callback intentionally opens another overlay.

Ideally, host UI opened from a webview would retain its focus origin and restore it when that UI closes, provided no newer overlay or intentional focus target has taken ownership.

Justkant (commented):
Thanks for addressing this focus regression. The expected behavior makes sense.

Focus restoration is attached specifically to the `account.request` success/cancel callbacks, while several other Wallet API hooks open host-side modals or drawers and may cause the same focus loss. Generalizing the callback wrappers would also be fragile: callbacks communicate operation results and do not necessarily correspond to the overlay actually closing. They may even open another overlay that should retain focus (edge cases).

`LiveAppModalView` also introduces its own focus capture/restoration logic. It serves a different immediate transition (returning focus to the parent UI when the whole Live App modal closes) but it shows that focus ownership is now being managed independently at multiple levels.

Could we instead handle this through a shared host modal/drawer lifecycle? An overlay opened from a webview could capture its focus origin and restore it when the overlay actually closes, provided no newer overlay or intentional focus target has taken ownership. This should cover every relevant Wallet API interaction and terminal path without coupling focus behavior to individual callbacks.

The targeted fix may solve the reported Earn flow, but I’d prefer that we either address focus restoration at that common boundary in this PR or clearly scope this as a temporary fix and track the broader issue separately.

apps/ledger-live-desktop/src/renderer/components/Web3AppWebview/WalletAPIWebview.tsx
alexstapenka-ledger: Agreed - callbacks were the wrong place.

Focus restoration now runs from the overlay close lifecycle (`useRestoreFocusOnDialogClose`) in ModularDialog and LiveAppModal, not from `account.request`. Restore is skipped if a newer overlay or another target already has focus. Wallet API wrappers are gone.

The remaining `openModal` / `setDrawer` hooks still sit outside this lifecycle; I’d leave that as a follow-up unless you want it in this PR.

So pls re-review!

apps/ledger-live-desktop/src/mvvm/features/LiveAppModal/LiveAppModalView.tsx
```diff
@@ -61,16 +62,18 @@ const EarnLiveAppModalContent = ({
 };
 
 const LiveAppModalView = ({ isOpen, params, onOpenChange, onClose }: LiveAppModalViewProps) => {
-  if (!isOpen || !params) return null;
```
Justkant, line 64: We need to be careful with that change because it means we can have a webview always present, I just saw that we've put a wallet-api webview in a modal (`LiveAppModalContent`) which is always rendered on all pages, not really good design here IMO

alexstapenka-ledger: Kept the early return null - the webview only mounts while the modal is open.

deepyjr (approved):

apps/ledger-live-desktop/src/mvvm/features/LiveAppModal/__integrations__/LiveAppModal.integration.test.tsx
```diff
@@ -37,7 +37,8 @@ const mockWebviewProps = jest.fn();
 jest.mock("~/renderer/components/Web3AppWebview", () => ({
   Web3AppWebview: (props: Record<string, unknown>) => {
     mockWebviewProps(props);
-    return <div data-testid="web3-app-webview" />;
+    // A real <webview> so focus handling sees the same element type it does in the app.
+    return <webview data-testid="web3-app-webview" />;
```
beths-ledger, line 41: <3

beths-ledger (approved):

PR #20926, opened 2026-08-19
fix(buysell): catalog retry, handoff assertion, bounded CAL lookups

jeportie:
### Checklist

- [x] `npx changeset` was attached.
- [ ] **Covered by automatic tests.** _9 unit tests cover the two `live-common` fixes (both failure modes verified red first). The E2E changes and the CI watchdog are exercised by an Android/iOS E2E run; the watchdog by design only emits when a device degrades._
- [x] **Impact of the changes:**
      - Any surface that resolves a live app: Buy/Sell, Swap, Earn, Discover, Card — Desktop and Mobile.
      - Cold app start on a slow or absent network — previously "App not found" for 30 minutes with no retry.
      - Any live app calling `currency.list` with specific token ids.
      - Android E2E job: additive `continue-on-error` steps only; none can fail a shard.

### Description

Fixes one real user-facing bug, removes the E2E suite's dependency on a third-party website, and corrects a silently-incomplete currency list.

#### 1. A failed live-app catalog fetch was cached as an empty catalog for 30 minutes

`RemoteLiveAppProvider/api/index.ts` swallowed network errors and returned `[]`. The provider cannot tell that from a genuinely empty catalog, so it stored a **successful empty registry with `error: null`** — and the only refetch is a 30-minute interval.

One failed request at startup therefore made **every live app unresolvable for half an hour**: Buy/Sell, Swap, Earn and Discover all showing "App not found", with no error state and no retry. The same `try/catch` also swallowed its own `throw`, leaving that guard dead.

Now: errors propagate, a failed *refresh* keeps the last good catalog, and a failure retries with backoff (2s → 30s, ~90s) instead of waiting out the interval.

#### 2. Buy/Sell E2E drove into the partner's production checkout page

`handleBuyFlow`/`handleSellFlow` waited up to 70s for Transak's or MoonPay's live site to render, so the suite went red on a third party's uptime. It now asserts the **handoff** — the `goToURL` the app hands the partner, plus its query parameters — read over the e2e bridge. That is the boundary we own, and it is what `e2e/desktop` already asserts.

Side effect: the "Buy and sell query parameters" test previously asserted **no** query parameters at all despite its name.

#### 3. `currency.list` resolved token ids in an unbounded `Promise.all`

[...]

PR #20989, opened 2026-08-20
[LWDM] fix(live-apps): retry a failed catalog fetch instead of caching it empty

jeportie:
### Checklist

      - Any surface that resolves a live app: Buy/Sell, Swap, Earn, Discover, Card — Desktop and Mobile.
      - Cold app start on a slow or absent network — the case that previously showed "App not found" for 30 minutes.
      - Desktop's `NetworkErrorScreen` refresh path (`exchange`, `earn`, `Swap2`, `SwapWebViewEmbedded`), which now actually receives an error state where `error` was never set before.

### Description

**Problem.** `RemoteLiveAppProvider/api/index.ts` wrapped the catalog request in a `try/catch` that logged the error and returned `[]`. `RemoteLiveAppProvider` cannot distinguish that from a genuinely empty catalog, so it stored a **successful empty registry with `error: null`** — and the only refetch is the `updateFrequency` interval, which both apps set to **30 minutes** (`AUTO_UPDATE_DEFAULT_DELAY`).

One failed request at startup therefore made **every live app unresolvable for half an hour**: "App not found" across Buy/Sell, Swap, Earn and Discover, with no error state and no retry. The same `try/catch` also swallowed its own `throw new Error("Response is not an Array")`, leaving that guard as dead code.

**Solution.**

- `api/index.ts` — let failures propagate instead of resolving to `[]`, so callers can tell "the catalog is empty" from "the catalog never loaded". This also restores the `Response is not an Array` guard.
- `RemoteLiveAppProvider/index.tsx` — record a real `error`; keep any previously loaded catalog when a *refresh* fails, so a bad refresh no longer wipes a good catalog; and retry with backoff (2s → 30s, six attempts ≈ 90s) instead of waiting for the next scheduled refresh, then fall back to the normal cadence so an offline device does not poll forever.

Found on Android CI, where the emulator has no usable network for the first 10–27s after launch and all six original attempts landed inside the first four seconds. It is not an emulator problem — any user whose first request fails gets the same 30 minutes.

### Validation

All runs below are on the combined branch this was split out of, so they are evidence for **this change**, not for this branch in isolation.

| Evidence | Result |
| --- | --- |
| 6 unit tests, both failure modes verified red first | pass — this is what proves the fix, since no E2E deliberately fails a cold-start catalog fetch |
| [Desktop E2E](https://github.com/LedgerHQ/ledger-live/actions/runs/32111343352) | 46 passed — the `live-common` change does not regress desktop |
| [buySell + portfolio, Android](https://github.com/LedgerHQ/ledger-live/actions/runs/32354389376) | ✅ 5/5 shards |
| [buySell, iOS + Android](https://github.com/LedgerHQ/ledger-live/actions/runs/32238954146) | ✅ 6/6 shards |
| [Full Android suite](https://github.com/LedgerHQ/ledger-live/actions/runs/32354411960) | run red on the known Swap issue; no catalog-related failure |

**What the E2E runs do and do not show.** They show the change does not regress anything. They do **not** show the bug is fixed — reproducing it needs a cold start with a failing network, which only the unit tests exercise deliberately.

### Context

- **JIRA or GitHub link**: [QAA-1498](https://ledgerhq.atlassian.net/browse/QAA-1498) — parent [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497)
- Split out of #20926, closed in favour of four owner-scoped PRs.

---

[QAA-1498]: https://ledgerhq.atlassian.net/browse/QAA-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ledger-live-common/src/platform/providers/RemoteLiveAppProvider/index.tsx
Justkant, line 217: I think you don't need to keep the useCallback above right ?
```suggestion
updateManifests: fetchManifests,
}),
[state, provider, setProvider, fetchManifests],
```

Justkant (commented):
Thx for the fix, could you check the copilot comment please it seems correct

Also the whole platform folder is another part that needs to be removed at some point, and the manifests requests should move the new RTK stores

libs/ledger-live-common/src/platform/providers/RemoteLiveAppProvider/index.tsx
```diff
@@ -182,14 +198,31 @@ export function RemoteLiveAppProvider({
...
+      if (cancelled) return;
+      if (succeeded) {
+        failedAttempts = 0;
+        return;
+      }
+      if (failedAttempts >= MAX_CONSECUTIVE_RETRIES) return;
+      const delay = Math.min(RETRY_BASE_DELAY_MS * 2 ** failedAttempts, RETRY_MAX_DELAY_MS);
+      failedAttempts += 1;
+      retryTimeout = setTimeout(run, delay);
+    };
```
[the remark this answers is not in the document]
jeportie: Good catch — addressed in 8b09e38.

`run()` now drops any pending retry before starting an attempt:

```ts
clearTimeout(retryTimeout);
retryTimeout = undefined;
```

so an interval tick that lands mid-backoff can no longer leave the old timer scheduled and start a second retry chain alongside it.

libs/ledger-live-common/src/platform/providers/RemoteLiveAppProvider/index.tsx
jeportie: Done in 5165922 — though not quite as suggested: `updateManifests: fetchManifests` doesn't typecheck, because `() => Promise<boolean>` isn't assignable to `() => Promise<void>`.

So rather than aliasing, I dropped the rename altogether — there's a single `updateManifests` callback again, and the context type returns the boolean the retry loop needs:

```ts
/** Resolves to `true` when the catalog was loaded, `false` when the fetch failed. */
updateManifests: () => Promise<boolean>;
```

No call site changes: they all take `refresh: () => void` / `refreshManifests: () => void`, and a `Promise<boolean>`-returning function is assignable to those. Net effect is smaller than before — the `fetchManifests` rename churn is gone from the diff entirely.

gre-ledger (commented):
ok for me (but i'm not code owner)

Moustafa-Koterba (approved):

PR #20990, opened 2026-08-20
[LWDM] perf(wallet-api): bound CAL token lookups in currency.list

jeportie:
### Checklist

      - Any live app calling `currency.list` with specific token ids — Buy/Sell, Swap, Earn, Discover, Card — on Desktop and Mobile.
      - The completeness of the currency list those apps receive, which was previously incomplete without any error.
      - Time to render any screen that waits on `currency.list`.

### Description

**Problem.** `currency.list` resolved every requested token id in a single unbounded `Promise.all`:

```ts
const tokenPromises = [...specificTokenIds].map(async tokenId => {
  const token = await getCryptoAssetsStore().findTokenById(tokenId);
  return token ? currencyToWalletAPICurrency(token) : null;
});
const resolvedTokens = await Promise.all(tokenPromises);
```

`findTokenById` is a network-only lookup, and CAL has no bulk-by-ids endpoint — verified against the live API: `?id=a,b,c` returns an **empty array with no error**, `?id=a&id=b` is rejected `400`. So a live app asking for hundreds of tokens fired hundreds of simultaneous requests, and the RPC awaited the slowest.

**This is a correctness bug, not only a performance one.** Measured against the real CAL API with the ids the Buy screen sends: **251 of 627 requests failed outright**. A failed lookup resolves to `null` and is dropped by the very next line, so the currency list handed to the live app was **quietly incomplete**, with nothing surfaced anywhere. CAL also advertises `x-rate-limit-limit: 500`, so the old code fired 627 concurrent requests against a 500 limit.

**Solution.** Bound the pool with `promiseAllBatched` from `@ledgerhq/live-promise` — already a dependency, and the established pattern for this in `coin-kaspa/src/logic/history/scanOperations.ts` and `coin-hedera/src/bridge/utils.ts`. With the bound, none fail.

**Trade-off, stated explicitly.** This does not make the call fast, and raising the bound makes it worse. Measured on-device, same 624 lookups:

| Bound | Wall time | Mean per request | Throughput |
| --- | --- | --- | --- |
| 10 | 67.5s | 747ms | 9.24 req/s |
| 25 | 88.8s | 2369ms | 7.03 req/s |

Measured peak overlap was exactly 25, so the transport was not capping it — the connection is throughput-limited rather than latency-limited, and extra concurrency only adds queueing. The Buy screen stays at 80–151s until the caller asks for fewer ids: [LIVE-36229](https://ledgerhq.atlassian.net/browse/LIVE-36229).

Note also that the bound is **per call**, not global — two live apps calling `currency.list` concurrently each get their own pool.

### Validation

All runs below are on the combined branch this was split out of, so they are evidence for **this change**, not for this branch in isolation.

| Evidence | Result |
| --- | --- |
| Regression test pinning the bound | verified red first: `Expected: <= 10, Received: 120` |
| Replay of the real 627 ids against the live CAL API | unbounded: 75.6s, **251 failed**. Bounded: none failed. |
| [On-device measurement](https://github.com/LedgerHQ/ledger-live/actions/runs/32148010936) | `Ledger Wallet Network Summary` reports true peak overlap of **exactly 10** — the bound is provably active, and all 624 lookups returned 200 |
| [Raising the bound to 25](https://github.com/LedgerHQ/ledger-live/actions/runs/32238954146) | 67.5s → **88.8s**, throughput −24%. Reverted; this is why the bound stays at 10. |
| [Desktop E2E](https://github.com/LedgerHQ/ledger-live/actions/runs/32111343352) | 46 passed |
| [buySell + portfolio, Android](https://github.com/LedgerHQ/ledger-live/actions/runs/32354389376) | ✅ 5/5 shards |

**What this does not show.** The screen getting faster. It does not — see the trade-off above. The evidence here is that the list is now *complete*, not that it is quick.

### Context

- **JIRA or GitHub link**: [QAA-1505](https://ledgerhq.atlassian.net/browse/QAA-1505) — parent [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497)
- Split out of #20926.

---

[LIVE-36229]: https://ledgerhq.atlassian.net/browse/LIVE-36229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/bright-tokens-bounded.md
+++ b/.changeset/bright-tokens-bounded.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Bound the concurrency of CAL token lookups in the `currency.list` wallet-api handler. It resolved every requested token id in a single unbounded `Promise.all`, so a live app asking for hundreds of tokens fired hundreds of simultaneous requests. Measured against the real CAL API with the ids the Buy screen requests, that took 75.6s and 251 of 627 requests failed outright — and a failed lookup is silently dropped, so the returned currency list was quietly incomplete. Bounded, none fail. Note this is a correctness fix only: CAL exposes no bulk id endpoint, so the request count is fixed, and raising the bound does not make the call faster — measured on an Android emulator, going from 10 to 25 in flight took the same 624 lookups from 67.5s to 88.8s, because the connection is throughput-limited rather than latency-limited
--- a/libs/ledger-live-common/src/wallet-api/react.test.ts
+++ b/libs/ledger-live-common/src/wallet-api/react.test.ts
@@ -11,6 +11,7 @@ import type { useWalletAPIServerOptions } from "./react";
 import { AccountPublicKeyUnavailable } from "../errors";
 import { accountGetPublicKeyLogic } from "./logic";
 import { getAccountIdFromWalletAccountId, resolveWalletApiSpendableBalance } from "./converters";
+import { getCryptoAssetsStore } from "@ledgerhq/ledger-wallet-framework/cryptoAssetsStore";
 
 const mockSetHandler = jest.fn();
 const mockSetConfig = jest.fn();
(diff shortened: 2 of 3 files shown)


Justkant (approved):
Thx for the fixes, this part needs to change at the wallet-api level too where we need to enforce pagination to avoid listing everything like we can now, it was not an issue when everything was available in the app

Also I wonder if breaking some of our long running functions like those with scheduler.yield could help here because it's mostly async already I don't expect much but we never know until we test (https://developer.chrome.com/blog/use-scheduler-yield)

PR #20991, opened 2026-08-20
[LWM] test(e2e): assert the Buy/Sell handoff, not the partner's page

jeportie:
### Checklist

- [x] `npx changeset` was attached. _`.changeset/buysell-handoff-assertion.md` — `live-mobile`, `@ledgerhq/live-e2e-shared` and `ledger-live-mobile-e2e-tests`, all at `patch`. My earlier note claimed one was not required because these packages are `private: true`; that was wrong — the repo sets `privatePackages: { version: true }`, so private packages are versioned and do get changelog entries._
- [x] **Covered by automatic tests.** _`libs/live-e2e-shared` has no jest harness, so the new pure helpers have no permanent unit test; they were verified executably against the real handoff URLs captured from CI for Transak, Coinbase and MoonPay. The page object itself is exercised by the E2E runs below._
      - Mobile Buy/Sell E2E only: `buyQueryParameters_*`, `navigateToBuyFrom*`, `sellFlow_BTC`.
      - The app-side change is behind `Config.DETOX` and does not affect production behaviour.
      - The e2e bridge gains one message pair (`getPtxHandoff` / `ptxHandoff`), so other bridge consumers are worth a sanity check.

### Description

**Problem.** `handleBuyFlow` and `handleSellFlow` drove into Transak's and MoonPay's **live checkout pages** and waited up to 70s for them to render. The suite therefore went red on a third party's uptime rather than on Ledger Live, and spent ~70s per test doing it.

`e2e/desktop` never did this — it reads the `gotourl` handoff out of Playwright's `webviewUrlHistory` and asserts provider and query parameters without ever loading the partner site.

**Solution.** Mobile now asserts the same boundary:

- The app records the handoff URL from `WebPTXPlayer` in `ptxHandoffStore`, guarded by `Config.DETOX`. `take()` reads *and clears*, so a handoff cannot leak into a later test.
- Exposed over the e2e bridge as `getPtxHandoff` → `ptxHandoff`, following the existing `getLogs`/`getFlags` pattern.
- Parsing lives in `libs/live-e2e-shared/src/buySellHandoff.ts`: `extractGoToUrl`, `urlMatchesProvider` (alias-aware, since Mercuryo's URLs say `mrcr`) and `getExpectedQueryParams`.

**Two bugs this caught in our own test code.** The handoff URL is double-encoded, so a single decode still leaves `https%3A%2F%2F` and `new URL()` throws a bare `TypeError: Invalid URL`. And the sell flow taps the 75% button, so asserting the caller's minimum amount compared a value the flow never types (`0.000566` against `0.0014496`).

Side effect: the "Buy and sell query parameters" test previously asserted **no** query parameters at all, despite its name.

Also collapses the six identical response handlers in `e2e/mobile/bridge/server.ts` into `resolvePending(key, payload)` — adding a seventh for the handoff had pushed new-code duplication past the Sonar gate.

Removing the 70s waits also cut shard length from ~20 min to 5–11, which as a side effect stopped the emulator crashes that were killing these runs (QAA-1509). That is avoidance, not a fix for the crash.

### Validation

All runs below are on the combined branch this was split out of, so they are evidence for **this change**, not for this branch in isolation. This is the one PR where a per-branch E2E run would be worth doing before merge, since it is the only one that changes test behaviour.

| Run | Result |
| --- | --- |
| [buySell + portfolio, Android](https://github.com/LedgerHQ/ledger-live/actions/runs/32354389376) | ✅ **success, 5/5 shards** |
| [buySell, iOS + Android](https://github.com/LedgerHQ/ledger-live/actions/runs/32238954146) | ✅ **success, 6/6 shards** |
| [Full Android suite](https://github.com/LedgerHQ/ledger-live/actions/runs/32354411960) | Buy/Sell specs pass; the run is red on the known Swap issue |
| [First attempt](https://github.com/LedgerHQ/ledger-live/actions/runs/32111341349) → [second](https://github.com/LedgerHQ/ledger-live/actions/runs/32114229793) → [third](https://github.com/LedgerHQ/ledger-live/actions/runs/32118674679) | 0/8 → 7/8 → **8/8**. The first two failures were bugs in this PR's own test code — the double-encoded handoff URL and the 75% sell amount — both found by CI and fixed here. |

Validation was deliberately run on **iOS as well as Android**: iOS does not suffer the emulator crash, so a red there is unambiguously this code and a green is a real green.

**Side effect worth noting for reviewers.** Removing the ~70s partner-page waits cut shard length from ~20 min to 5–11. Across three subsequent runs that meant **2,045 emulator-minutes with zero emulator losses**, against 7 of 12 shards losing one in the 17/18 Aug nightlies at the same shard length. That is a real improvement, but it is avoidance — the underlying qemu crash ([QAA-1509](https://ledgerhq.atlassian.net/browse/QAA-1509)) is untouched, and if any suite grows back past ~15 minutes it returns.

### Context

- **JIRA or GitHub link**: [QAA-1497](https://ledgerhq.atlassian.net/browse/QAA-1497)
- Split out of #20926.

---

[QAA-1497]: https://ledgerhq.atlassian.net/browse/QAA-1497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[QAA-1509]: https://ledgerhq.atlassian.net/browse/QAA-1509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jeportie:
## Per-branch E2E run

[Run 32385815216](https://github.com/LedgerHQ/ledger-live/actions/runs/32385815216) — `buySell`, iOS + Android, on **this branch alone** rather than the combined branch the earlier evidence came from.

| Shards | Result |
|---|---|
| iOS 1/3, 2/3, 3/3 | ✅ all green |
| Android 2/3, 3/3 | ✅ green |
| Android 1/3 | ❌ |

**8 unique tests, 1 never passed:** `[BTC] - Buy entry point from account page`, on `crypto-amount-option-button not found after 60000ms`.

That is **Mode A**, not this PR — the Buy screen taking 80–151s to resolve 736 token ids, which this PR does not claim to fix ([LIVE-36229](https://ledgerhq.atlassian.net/browse/LIVE-36229)). iOS running all three shards green is the control that makes that unambiguous: same specs, same commit, no slow-network fan-out, no failure.

No handoff-related failures, no partner-page failures, and no emulator losses.

So the branch is sound on its own. The one red shard will keep flaking until PTX reduces the id count — it is the known stall showing through, and it recovers on retry when the CAL cache is warm.

[LIVE-36229]: https://ledgerhq.atlassian.net/browse/LIVE-36229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

philipptpunkt (dismissed):
lgtm for the ptx changes

cunhabruno (dismissed):

cunhabruno (approved):

philipptpunkt (approved):

PR #21015, opened 2026-08-21
[LWDM] fix(coin-framework): stop truncating account history to one page

YazhuEth:
### 📝 Description

**Previous behaviour.** `getAccountShape` called `listOperations` once and dropped the returned `next`, so any account longer than one explorer page was truncated for good. No later sync came back for it: the walk is newest-first and `minHeight` only moves forward. The one resume path was dead — `extra.pagingToken` is a reserved key, stripped on every write path, so it was always empty.

**Fix.** `paginateOperations` walks the cursor chain within a sync. A falsy `next` ends the stream, since several modules send `""` rather than omitting it. It also stops when the module stops advancing — an empty page with a cursor, or a cursor already followed (a `Set`, so a `c1 -> c2 -> c1` cycle ends it too). Errors propagate on purpose: walking newest-first, a partial history would never be completed later. The dead `pagingToken` read is removed rather than repaired.

Also appends the coin-evm Ledger explorer's batches in place instead of rebuilding the accumulator once per page.

**Regression cover.** 7 unit tests in `paginateOperations.test.ts`: the multi-page walk, both terminators, the empty-page guard, a repeated cursor, a longer cycle, and error propagation.

**No cap.** A 50k cap was written and dropped: counted in explorer records it bounds nothing, since `getOperations` expands each record into as many operations as it carries events. Measured through the `eth` explorer:

| address | records | ERC20 events | payload |
|---|---|---|---|
| `0xa271266Ea7CF…` | 28 800 | 12 | 26 MB |
| `0xde0B295669a9…` | 32 570 | 2 657 590 | 1 984 MB |

The second shows 3 197 transactions on Etherscan and expands to ~2.6M operations — under a 50k-record cap, which never fired. 50k fails as an operation count too: the healthy address produces ~57.6k. Don't size a limit from the heavy one either — it's an old multisig contract (nonce 0) sprayed with spam, 1000 forged `Transfer` events per transaction. A real limit needs the right unit and a call on which end of the history to keep, since the explorer is walked ascending. Own ticket.

**QA.** From-scratch sync of a long account on the six enabled families, then a second sync — no duplicates, no gaps. Stellar `GBGHNJRD…CL32X`: 682 operations over 30 pages, matching the explorer; before, one page. An already-synced account won't show the difference, since incremental syncs resume above the newest known block.

Reviewers: Ethereum mainnet is configured `explorer: { type: "ledger" }`, which returns `NO_TOKEN`, so `paginateOperations` makes a single call there. Test on an etherscan-configured network or a non-EVM module.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36317](https://ledgerhq.atlassian.net/browse/LIVE-36317), [LIVE-36446](https://ledgerhq.atlassian.net/browse/LIVE-36446)
- **ADR** (if any): n/a

### DIFF
--- a/.changeset/lucky-cursors-walk.md
+++ b/.changeset/lucky-cursors-walk.md
@@ -1,0 +1,19 @@
+---
+"@ledgerhq/live-common": patch
+"@ledgerhq/coin-evm": patch
+---
+
+fix(coin-framework): follow the listOperations cursor so account history is no longer truncated to one page
+
+getAccountShape called listOperations once and discarded the returned `next`, so any account with
+more operations than one explorer page never received the rest, and no later sync recovered the
+tail: the walk is newest-first and `minHeight` only ever moves forward, so the pages below the
+first were lost for good.
+
+It now walks the cursor chain within a sync, treating a falsy cursor as end of stream. The walk is
+unbounded — only a module that cannot progress ends it early: an empty page, or a cursor already
+followed (a repeat, or a longer cycle). The `extra.pagingToken` resume read is removed:
+nothing could ever write it, and `minHeight` is the resume position across syncs.
+
+On the coin-evm side, the Ledger explorer's `fetchPaginatedOpsWithRetries` appends each batch in
+place instead of rebuilding the whole accumulator (`[...previous, ...batch]`) once per page.
--- a/libs/coin-modules/coin-evm/src/network/explorer/ledger.ts
+++ b/libs/coin-modules/coin-evm/src/network/explorer/ledger.ts
@@ -62,11 +62,11 @@ export async function fetchPaginatedOpsWithRetries(
       },
     });
 
-    const mergedOperations = [...previousOperations, ...operationsBatch];
+    previousOperations.push(...operationsBatch);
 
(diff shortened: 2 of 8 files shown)


cted-ledger (dismissed):

YazhuEth:
**E2E**
-D: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/46f15e9d-1927-4a6c-abab-4e744b15b139/
-M:https://ledger-live.allure.green.ledgerlabs.net/allure/reports/85b75194-9b14-4312-b1dd-7c2450dbfe16/

cted-ledger (dismissed):

YazhuEth:
**E2E**
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/0bcb13c4-a904-4ddd-bed5-4c38d5598a83/
- https://ledger-live.allure.green.ledgerlabs.net/allure/reports/8761a26c-d0ea-4196-aa4e-f904ee0e91d0/

cted-ledger (approved):

PR #21023, opened 2026-08-21
[LWM] fix(contacts): animate add address steps (LIVE-35883)

deepyjr:
### ✅ Checklist

- [x] **Covered by automatic tests.** The queued drawer flow test verifies that outgoing and incoming screens coexist during the standard animation.
      Verify queued drawer flows animate between their screens, especially Contacts add address from currency selection through address entry and naming; verify back navigation and drawer closing remain unchanged.

### 📝 Description

Queued drawer flows previously replaced each sheet step immediately, unlike the transition between Select Crypto and Select Network.

This PR makes the Reanimated screen transition the shared default for `QueuedDrawerFlow`. It uses the existing Modular Drawer timing and movement: the outgoing screen scales down, moves down, and fades while the incoming screen scales to its final size. Contacts add address now benefits without a feature-specific option

https://github.com/user-attachments/assets/77b61ac5-dcf0-4182-8bc2-a50bc83fb98e

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-35883

---

### DIFF
--- a/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/__tests__/QueuedDrawerFlow.test.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/__tests__/QueuedDrawerFlow.test.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { Text } from "react-native";
-import { render, screen } from "@tests/test-renderer";
+import { act, render, screen } from "@tests/test-renderer";
 import type { QueuedBottomSheetProps } from "@shared/ui-queued-bottom-sheet";
 import { QueuedDrawerFlow } from "..";
 
@@ -50,7 +50,7 @@ describe("QueuedDrawerFlow", () => {
     mockDrawerUnmountCount = 0;
   });
 
-  it("should replace the current screen without closing the drawer in tests", () => {
+  it("should replace the current screen without closing the drawer", () => {
     const onClose = jest.fn();
     const onBack = jest.fn();
     const { rerender } = render(
@@ -86,6 +86,8 @@ describe("QueuedDrawerFlow", () => {
     );
 
     expect(screen.getByText("Address screen")).toBeVisible();
+    expect(screen.getByText("Asset screen")).toBeVisible();
+    act(() => jest.runAllTimers());
     expect(screen.queryByText("Asset screen")).toBeNull();
     expect(mockDrawerMountCount).toBe(1);
     expect(mockDrawerUnmountCount).toBe(0);
@@ -101,6 +103,21 @@ describe("QueuedDrawerFlow", () => {
       }),
     );
     expect(onClose).not.toHaveBeenCalled();
+  });
+
+  it("should keep the previous screen active when navigating back during its exit", () => {
+    const onClose = jest.fn();
+    const { rerender } = render(
+      <QueuedDrawerFlow currentStep="asset" isOpen onClose={onClose} screens={screens} />,
+    );
+
+    rerender(<QueuedDrawerFlow currentStep="address" isOpen onClose={onClose} screens={screens} />);
+    rerender(<QueuedDrawerFlow currentStep="asset" isOpen onClose={onClose} screens={screens} />);
+
+    act(() => jest.runAllTimers());
+
+    expect(screen.getByText("Asset screen")).toBeVisible();
+    expect(screen.queryByText("Address screen")).toBeNull();
   });
 
   it("should hide the back button when no back handler is provided", () => {
--- a/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/__tests__/QueuedDrawerFlow.test.tsx
+++ b/apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/__tests__/QueuedDrawerFlow.test.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { Text } from "react-native";
-import { act, render, screen } from "@tests/test-renderer";
+import { render, screen } from "@tests/test-renderer";
 import type { QueuedBottomSheetProps } from "@shared/ui-queued-bottom-sheet";
 import { QueuedDrawerFlow } from "..";
 
@@ -86,8 +86,6 @@ describe("QueuedDrawerFlow", () => {
     );
 
     expect(screen.getByText("Address screen")).toBeVisible();
-    expect(screen.getByText("Asset screen")).toBeVisible();
-    act(() => jest.runAllTimers());
     expect(screen.queryByText("Asset screen")).toBeNull();
     expect(mockDrawerMountCount).toBe(1);
     expect(mockDrawerUnmountCount).toBe(0);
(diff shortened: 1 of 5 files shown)


apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  useEffect(() => {
+    setActiveScreens(previousScreens => {
+      const currentScreen = previousScreens.find(screen => screen.step === currentStep);
+      if (currentScreen) {
+        return currentScreen.content === currentContent
+          ? previousScreens
+          : previousScreens.map(screen =>
+              screen.step === currentStep ? { ...screen, content: currentContent } : screen,
+            );
```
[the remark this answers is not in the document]
claudiiafg: this seems like a good point

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
```diff
@@ -0,0 +1,137 @@
...
+      }
+
+      if (isTestEnv) {
```
claudiiafg: This makes Jest skip the real path, so the reverse-navigation cannot be tested (flagged by copilot above)

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/__tests__/QueuedDrawerFlow.test.tsx
```diff
@@ -50,7 +50,7 @@ describe("QueuedDrawerFlow", () => {
     mockDrawerUnmountCount = 0;
   });
 
-  it("should replace the current screen without closing the drawer", () => {
+  it("should replace the current screen without closing the drawer in tests", () => {
```
mcayuelas-ledger: not needed

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
mcayuelas-ledger: Indeed good point

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
mcayuelas-ledger, line 1: IMO We can simplify the logic and improve performance by fully leveraging `react-native-reanimated`.

Currently, the component relies on deeply nested useEffect logic to manually manage a state queue, track isEntering/isExiting flags, and use react-native-worklets to orchestrate unmounting via a JS callback.

We can completely eliminate the local state, lifecycle hooks, and the `worklet` dependency by using Reanimated's native Layout Animations. By simply changing the key on an Animated.View, Reanimated will automatically keep the component in the native view tree while the exit animation plays, and then unmount it for us natively.


```ts
const TRANSITION_CONFIG = {
  duration: isTestEnv ? 0 : 250,
  easing: Easing.bezier(0.17, 0.84, 0.44, 1),
};

const SCREEN_STYLE = {
  position: "absolute" as const,
  top: 0,
  right: 0,
  bottom: 0,
  left: 0,
};

type QueuedDrawerScreenTransitionProps<Step extends string> = Readonly<{
  currentStep: Step;
  screens: QueuedDrawerFlowScreenRegistry<Step>;
}>;

const customEntering = () => {
  "worklet";
  return {
    initialValues: { transform: [{ scale: 0.95 }] },
    animations: { transform: [{ scale: withTiming(1, TRANSITION_CONFIG) }] },
  };
};

const customExiting = () => {
  "worklet";
  return {
    initialValues: { transform: [{ scale: 1 }, { translateY: 0 }], opacity: 1 },
    animations: {
      transform: [
        { scale: withTiming(0.95, TRANSITION_CONFIG) },
        { translateY: withTiming(32, TRANSITION_CONFIG) },
      ],
      opacity: withTiming(0, TRANSITION_CONFIG),
    },
  };
};

export function QueuedDrawerScreenTransition<Step extends string>({
  currentStep,
  screens,
}: QueuedDrawerScreenTransitionProps<Step>): React.JSX.Element {
  return (
    <View style={{ flex: 1, position: "relative" }}>
      <Animated.View
        key={currentStep}
        entering={customEntering}
        exiting={customExiting}
        style={[{ flex: 1 }, SCREEN_STYLE]}
      >
        {screens[currentStep].content}
      </Animated.View>
    </View>
  );
}
```

WDYT about this approach?

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+export function QueuedDrawerScreenTransition<Step extends string>({
+  currentStep,
+  screens,
+}: QueuedDrawerScreenTransitionProps<Step>): React.JSX.Element {
+  const currentContent = screens[currentStep].content;
+  const [activeScreens, setActiveScreens] = useState<readonly ActiveScreen<Step>[]>(() => [
+    { step: currentStep, content: currentContent, isEntering: false, isExiting: false },
+  ]);
```
mcayuelas-ledger: Also here you could have use `useReducer` to have explicit actions etc

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
deepyjr: Let's give it a try, it's more readable imo

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
```diff
@@ -0,0 +1,117 @@
...
+            );
+      }
+
+      const previousSteps = previousScreens.map(screen => screen.step);
+
+      incomingScale.value = 0.95;
+      incomingScale.value = withTiming(1, TRANSITION_CONFIG);
+      outgoingScale.value = withTiming(0.95, TRANSITION_CONFIG);
+      outgoingTranslateY.value = withTiming(32, TRANSITION_CONFIG);
+      outgoingOpacity.value = withTiming(0, TRANSITION_CONFIG, finished => {
```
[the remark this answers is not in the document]
deepyjr: Resolved in f50ae252d7e. The component now uses Reanimated entering and exiting layout animations, so each transition receives its own initial and target values instead of reusing shared values.

apps/ledger-live-mobile/src/mvvm/components/QueuedDrawerFlow/useQueuedDrawerScreenTransition.tsx
```diff
@@ -0,0 +1,143 @@
...
+            : previousScreens.map(screen =>
+                screen.step === currentStep ? { ...screen, content: currentContent } : screen,
+              );
+        }
+
+        return previousScreens.map(screen =>
+          screen.step === currentStep
+            ? { ...screen, content: currentContent, isEntering: true, isExiting: false }
+            : { ...screen, isEntering: false, isExiting: true },
+        );
```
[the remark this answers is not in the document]
deepyjr: Resolved in f50ae252d7e. Reanimated now unmounts the React screen on exit while preserving its native exit animation. Returning renders a new keyed screen, so the active input mounts again and autoFocus runs.

mcayuelas-ledger (approved):

PR #19921, opened 2026-07-23
feat(support): add @support/ts-config-{base,web,web-x-native} TypeScript preset packages

ysitbon:
- [ ] make sure to keep docs/tsconfig-in-ddd.md up to date (follow up https://github.com/LedgerHQ/ledger-live/pull/21024)
 
## Summary

- Adds four shared TypeScript config presets under `support/` to eliminate repeated `compilerOptions` across `domain/`, `shared/`, and `features/` packages
- Registers `support/*` in `pnpm-workspace.yaml` and adds `scope:support` Nx tag inference
- Inverts root `tsconfig.base.json`: it now extends `@support/ts-config-base-legacy` (the package is the source of truth, not the root file)
- Migration of existing packages: PR #20563 (same stack)

### Presets

| Package | Use case | Key additions vs base |
|---|---|---|
| `@support/ts-config-base-legacy` | Root `tsconfig.base.json` + `@support/ts-config-base` base | NodeNext, strict, interop flags (monorepo source of truth) |
| `@support/ts-config-base` | `domain/*`, `shared/*` — no JSX, no DOM | ES2022, ESNext, bundler resolution, noEmit |
| `@support/ts-config-web` | `features/platform/*`, web-only `features/flow/*` | `jsx:react-jsx`, DOM lib, `.web` moduleSuffixes |
| `@support/ts-config-web-x-native` | cross-platform `features/flow/*` (`.web.tsx`/`.native.tsx`) | 3-file project references pattern, no DOM |

### `${configDir}` for portable include/exclude

`include`/`exclude` in the presets use `${configDir}` (TypeScript 5.5+, repo uses TS 6.0.2). When a base config defines `include: ["${configDir}/src/**/*"]`, TypeScript resolves `${configDir}` to the **leaf consumer's directory**, so consumers inherit the correct path without repeating it.

### `@support/ts-config-web-x-native` — 3-file pattern

Cross-platform packages need separate `moduleSuffixes` per build target and a solution file for `tsc -b`. Consuming packages require 3 minimal files:

```jsonc
// tsconfig.json — solution file
{ "extends": "@support/ts-config-web-x-native", "files": [], "references": [...] }

// tsconfig.web.json
{ "extends": "@support/ts-config-web-x-native/tsconfig.web.json" }

// tsconfig.native.json
{ "extends": "@support/ts-config-web-x-native/tsconfig.native.json" }
```

Reference: `devtools/shell/tsconfig.{json,web.json,native.json}`.

### Root tsconfig inversion

[...]

PR #21024, opened 2026-08-21
docs(tsconfig): add DDD tsconfig web/native guide

Sebastien-Dav1d:
### 📝 Description

Add a doc to explain how the new tsconfigs in DDD works

### 🔗 Context

- **JIRA / GitHub issue**: 
- **ADR** (if any):

### DIFF
--- a/docs/tsconfig-in-ddd.md
+++ b/docs/tsconfig-in-ddd.md
@@ -1,0 +1,195 @@
+# TypeScript configuration in DDD packages
+
+## The problem
+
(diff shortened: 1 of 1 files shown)


gre-ledger (commented):
will this doc be compatible with https://github.com/LedgerHQ/ledger-live/pull/19921 ? will it need to be adapted a bit? because i understand that with the other PR we will be able to code share more and maybe have less to do in each pkg to configure 🤔

gre-ledger (approved):

PR #21025, opened 2026-08-21
[LWDM] fix(contacts): defer device connection for contact editing (LIVE-35851)

claudiiafg:
### 📝 Description

Editing a contact name or an address label prompted the user to connect their Ledger **before** the edit form even opened, so people had to plug in a device just to discover what they were about to change — and the device step was wasted if they backed out.

**Previous behaviour:** pressing *Edit* on a contact (or an address) immediately opened the signer validation step. Only once the device was connected and confirmed did the rename form appear.

**Fix:** the edit form now opens straight away, and device validation is deferred to *Save*. The signer step became an awaitable gate:

- `useContactEditSignerUiState` exposes `requestSignerApproval()` (returns a promise resolved by `grantSignerApproval()` once the device confirms) alongside a new `isEditSessionActive` flag.
- `onEditPress` in both flow view models (`useContactDetailEditDeleteFlowViewModel`, `useContactAddressDetailActionsFlowViewModel`) just opens the edit dialog; each now provides `requestSaveApproval`, which only hits the device when signer validation is actually required for that edit.
- The rename view models (`useRenameContactDialogViewModel`, `useRenameAddressDialogViewModel`) `await requestSaveApproval()` inside `onConfirm` and skip the write if the user declines.

**Draft preservation:** because the edit dialog hides while the signer step is on screen, the draft reset is now keyed on `isEditSessionActive` rather than the dialog's open state, so the user's typed name/label survives the device round trip. Cancelling a signer mismatch returns to the still-populated form (`resolveEditUiStateOnSignerMismatchCancel` now yields `"edit-open"` instead of `"closed"`).

**Mobile specifics:** `QueuedBottomSheet` fires `onClose` whenever it stops being requested, which would have cleared the selected address as soon as the signer sheet took over — `useContactAddressDetailActionsAdapter` now guards its close on `editUiState === "edit-open"`. Both adapters also scope the "edit address" page event to `isEditSessionActive` so returning from the signer step doesn't emit a duplicate analytics event.

https://github.com/user-attachments/assets/5b747f97-8e5d-4126-acf0-f7f2e7f006f8

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35851](https://ledgerhq.atlassian.net/browse/LIVE-35851)
- **ADR** (if any): n/a

[LIVE-35851]: https://ledgerhq.atlassian.net/browse/LIVE-35851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-desktop/src/mvvm/features/Contacts/__integrations__/Contacts.signerMismatch.integration.test.tsx
+++ b/apps/ledger-live-desktop/src/mvvm/features/Contacts/__integrations__/Contacts.signerMismatch.integration.test.tsx
@@ -69,6 +69,15 @@ describe("Contacts signer mismatch integration", () => {
     await user.click(screen.getByTestId("contacts-detail-address-row-address-ethereum"));
     await user.click(screen.getByTestId("contacts-address-detail-edit"));
 
+    await waitFor(() => {
+      expect(screen.getByTestId("contacts-rename-address-dialog")).toBeVisible();
+    });
+
+    const labelInput = screen.getByTestId("contacts-rename-address-input");
+    await user.clear(labelInput);
+    await user.type(labelInput, "Main ETH");
+    await user.click(screen.getByTestId("contacts-rename-address-confirm"));
+
     await expectSignerMismatchDialog(user);
   });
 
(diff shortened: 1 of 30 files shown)


features/flow/contacts/src/steps/Detail/useContactEditSignerUiState.ts
```diff
@@ -21,57 +21,85 @@ export type UseContactEditSignerUiStateResult = Readonly<{
 }>;
 
 export function useContactEditSignerUiState(): UseContactEditSignerUiStateResult {
-  const [editUiState, setEditUiState] = useState<SignerEditUiState>("closed");
+  const [editUiState, setEditUiStateValue] = useState<SignerEditUiState>("closed");
```
mcayuelas-ledger: good point from sonar

apps/ledger-live-desktop/src/mvvm/features/Contacts/__integrations__/Contacts.signerMismatch.integration.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { MemoryRouter, Route, Routes } from "react-router";
 import { mockPopulatedContacts } from "@domain/entity-contact/schema.mock";
-import { render, screen, waitFor, withFlagOverrides } from "tests/testSetup";
+import { fireEvent, render, screen, waitFor, withFlagOverrides } from "tests/testSetup";
```
mcayuelas-ledger: pleas use userEvent from `const { user } render (`

deepyjr (dismissed):

deepyjr (approved):

PR #21028, opened 2026-08-21
[LWDM] feat(perps): add the perps deposit screen to desktop

ooke-ledger:
### 📝 Description

The screen the live app hands over to when it calls `custom.perps.deposit`: pick a funding account, enter an amount, move on to review. Desktop only — the mobile screen is #21029, and the two do not depend on each other.

What's worth a look:

- The amount is entered in the counter value currency, and the USDC figure under it comes from the SwapKit quote. While the quote is in flight, the line shimmers, and the Review CTA stays disabled until a quote lands.
- Percentage pills and MAX round to the counter value's precision, which keeps MAX from drifting past the spendable balance when the rate moves between renders.
- Insufficient balance and unavailable quotes are the only errors surfaced here

Also adds `wallet-api/Perps/depositQuote` in `live-common`. #21029 carries an identical copy so it can stand on its own .

Perps deposit screen on desktop

### 🧱 Stack

- **desktop** — this PR, then #21030 (review step)
- **mobile** — #21029, then #21031 (review step)

### 🔗 Context
https://ledgerhq.atlassian.net/browse/LIVE-34567
https://ledgerhq.atlassian.net/browse/LIVE-35359

PR #21030, opened 2026-08-21
[LWDM] feat(perps): add the deposit review step to desktop

ooke-ledger:
### 📝 Description

The step between entering an amount and signing: what SwapKit will take, what it will credit, and where it lands. Desktop only — the mobile sheet is #21031.

What's worth a look:

- Both amounts are rendered from the quote in their own currency, sent in the funding currency and received in USDC, so the screen never echoes one side back as the other. Amounts cross the boundary as display-unit strings; the accounts they belong to are already on the params, so no currency ids are threaded through.
- Opening the review is gated on a quote being present, which is why the deposit CTA in #21028 waits for one.

Adds `PerpsDepositReviewParams` to `wallet-api/Perps/server` in `live-common`. #21031 carries the identical change so the mobile chain stands alone; whichever merges second, it no-ops.

Perps deposit review step on desktop

### 🧱 Stack

Desktop chain: #21028 (deposit screen) → this PR. Independent of the mobile chain (#21029 → #21031).

### 🔗 Context
https://ledgerhq.atlassian.net/browse/LIVE-34568
https://ledgerhq.atlassian.net/browse/LIVE-35327

### DIFF
--- a/.changeset/perps-deposit-review.md
+++ b/.changeset/perps-deposit-review.md
@@ -1,0 +1,7 @@
+---
+"ledger-live-desktop": minor
+"live-mobile": minor
+"@ledgerhq/live-common": minor
+---
+
+Add the Perps deposit review step on desktop and mobile, reached from the deposit form, showing the swap and deposit details before the user confirms. The amount landing on the perps account now comes from a provider quote, since that balance has no counter value of its own.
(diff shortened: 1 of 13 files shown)


mcayuelas-ledger (approved):
Ok for WXP scope

francois-guerin-ledger (approved):

PR #21058, opened 2026-08-24
[LWDM] feat(test-quarantine): roll flake reporting out to jest packages

lewisd5:
### 📝 Description

Wires the jest flake reporter and CI-only retries into 67 packages owned by @ledgerhq/engagement, @ledgerhq/wallet-xp and @ledgerhq/platform, following the pattern established for the ptx packages in #21004.

Each package gains the reporter plus the shared jest-retries setup file, so retries stay CI-only and keep logErrorsBeforeRetry: a failure is still printed before the second attempt, and a test failing both attempts still fails the build.

### 🔗 Context

Note. This is part of a phased rollout of the reporters. 2 batches have been completed thus far with a further set of batches to be concluded. It is not intended to cover all test scopes

### DIFF
--- a/apps/ledger-live-desktop/jest.config.js
+++ b/apps/ledger-live-desktop/jest.config.js
@@ -136,6 +136,7 @@ module.exports = {
     "default",
     ...(process.env.CI ? ["github-actions"] : []),
     ["jest-sonar", { outputName: "sonar-executionTests-report.xml", reportedFilePath: "absolute" }],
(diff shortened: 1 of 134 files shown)


thesan (commented):
I think this should be added to `@support/jest-features-flow` and `@support/jest-devtools` too.

thesan (approved):

deepyjr (approved):

tonykhaov (approved):

PR #21059, opened 2026-08-24
fix(hw-app-exchange): avoid CSP via static protobuf (LIVE-15710)

ooke-ledger:
## Summary
- Completing exchange transactions that decode protobuf payloads (notably XRP flows needing memo/tag / `extraId` handling) hit a CSP `unsafe-eval` violation: `@ledgerhq/hw-app-exchange` was using protobufjs reflection (`Root.fromJSON` + runtime codegen) to decode swap/sell/fund payloads.
- Switch `prebuild` from `pbjs -t json` to `pbjs -t static-module`, commit the generated static JS (+ typings), and decode via `ledger_swap.New*Response.decode/verify` so no dynamic `Function`/`eval` runs at runtime.
- Align `SellPayload` / `FundPayload` with the generated static types (`Uint8Array`, optional `outAmount`).

Reference: [protobuf.js static modules](https://github.com/protobufjs/protobuf.js#static-modules) — static codegen is the recommended approach for CSP-restricted environments that disallow `unsafe-eval`.

## Test plan
- [ ] `pnpm --filter=@ledgerhq/hw-app-exchange test` (SellUtils / SwapUtils)
- [ ] `pnpm --filter=@ledgerhq/hw-app-exchange exec tsc --noEmit --project tsconfig.json`
- [ ] Reproduce previously failing XRP exchange completion under CSP and confirm no `unsafe-eval` console/CSP error
- [ ] Smoke swap / sell / fund payload decode still returns expected fields (`inExtraId` / amounts / addresses)


libs/ledgerjs/packages/hw-app-exchange/package.json
```diff
@@ -48,14 +48,14 @@
   },
   "scripts": {
     "clean": "rimraf lib lib-es",
-    "prebuild": "pnpm pbjs -t json -r ledger_swap -o src/generate-protocol.json ./protocol.proto",
+    "prebuild": "pnpm pbjs -t static-module -w es6 -r ledger_swap -o src/generate-protocol.js ./protocol.proto && pnpm pbts -o src/generate-protocol.d.ts src/generate-protocol.js",
```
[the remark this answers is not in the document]
ooke-ledger: It's now fixed using a post-build.

libs/ledgerjs/packages/hw-app-exchange/package.json
```diff
@@ -48,14 +48,15 @@
...
     "build": "tsc --project tsconfig.build.json && tsc --project tsconfig.build.json -m esnext --moduleResolution bundler --outDir lib-es",
+    "postbuild": "cp src/generate-protocol.d.ts lib/generate-protocol.d.ts && cp src/generate-protocol.d.ts lib-es/generate-protocol.d.ts && rm -f lib/generate-protocol.d.ts.map lib-es/generate-protocol.d.ts.map",
     "coverage": "jest --coverage --passWithNoTests",
     "prewatch": "pnpm build",
     "watch": "tsc --watch --project tsconfig.build.json",
     "watch:es": "tsc --watch --project tsconfig.build.json -m esnext --moduleResolution bundler --outDir lib-es",
-    "doc": "documentation readme src/** --section=API --pe ts --re ts --re d.ts",
-    "lint": "oxlint ./src --ignore-pattern '**/setupTests.ts'",
+    "doc": "documentation readme $(ls src/*.ts | grep -v '\\.d\\.ts$' | sort) --section=API --pe ts --re ts --shallow",
```
philipptpunkt: That's Unix, so won't work on Windows, no?! Since it's only for docs not critical, but maybe there is a easy replacement which works on all machines?

libs/ledgerjs/packages/hw-app-exchange/src/SellUtils.ts
```diff
@@ -1,13 +1,12 @@
-import protobuf from "protobufjs";
-import * as protoJson from "./generate-protocol.json";
+import { ledger_swap } from "./generate-protocol";
```
philipptpunkt: would it make sense to give this guy a more generic name, because it's also used for sell/fund? Something like `ledger_trade` or `ledger_transaction`? (maybe that's a bit too wide though) 😃

ooke-ledger: Fair!

philipptpunkt (approved):

PR #21061, opened 2026-08-24
[LWD] feat(tsconfig): add web/native tsconfig splits to feature packages

Sebastien-Dav1d:
### 📝 Description

Adds per-platform `tsconfig.web.json` and `tsconfig.native.json` files to eight feature packages.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36347](https://ledgerhq.atlassian.net/browse/LIVE-36347)
- **ADR** (if any): 

[LIVE-36347]: https://ledgerhq.atlassian.net/browse/LIVE-36347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


features/flow/contacts/src/ContactsView.types.ts
```diff
@@ -1,6 +1,9 @@
 import type { ChangeEvent } from "react";
 import type { ContactId } from "@domain/entity-contact";
-import type { ContactsListViewLabels, ContactsPageViewModel } from "@features/flow-contacts-list";
+import type {
+  ContactsListViewLabels,
+  ContactsPageViewModel,
+} from "@features/flow-contacts-list/native";
```
deepyjr, line 6: do we want to have the /native at the end ?

henri-ly (dismissed):

features/flow/contacts/src/ContactsView.types.ts
Sebastien-Dav1d: It is a big mess of imports and exports so yeah it is intended. When everything will have the right architecture, I will remove all bad use cases, mainly web.ts and native.ts files that are makeshift solution to avoid to handle auto resolution

apps/ledger-live-desktop/src/main/db/index.ts
```diff
@@ -18,7 +18,7 @@ import { encryptData, decryptData } from "~/main/db/crypto";
 import { readFile, writeFile } from "~/main/db/fsHelper";
 
 const debounce = <T, R>(fn: (...args: T[]) => R, ms: number) => {
-  let timeout: NodeJS.Timeout | undefined;
+  let timeout: ReturnType<typeof setTimeout> | undefined;
```
mcayuelas-ledger, line 21: why do we have all theses unrelated changes to web-native?

Sebastien-Dav1d: Weirdly, typechecking would fail on all those files because of TimeOut

Sebastien-Dav1d: In this pr, I added web to module suffixes of lwd to handle properly ddd packages with web files. Apparently, this led to more typechecking ly lwd where he would pull DOM components and finally got it confused between node and Dom timeout with another type. Initial error was that timeout was a number and NodeJS.Timeout failed the typecheck. The number comes from dom, the fix just retreive the type so whether we typecheck from node or dom, it passes

RobinVncnt (dismissed):

apps/ledger-live-desktop/src/main/db/index.ts
gre-ledger: yeah it's a classical problem NodeJS.Timeout and web's setTimeout returned value have different types. depends on where it runs.

gre-ledger (approved):

beths-ledger (approved):

PR #21063, opened 2026-08-24
[LWDM] feat(contacts): register Wallet Sync persistence (LIVE-33960)

deepyjr:
### ✅ Checklist

- [x] **Covered by automatic tests.** Focused Wallet Sync, Desktop reducer, and Mobile reducer tests cover registration, quarantine, and local persistence.
      QA should create, edit, and remove Contacts across Desktop and Mobile Wallet Sync members, then restart each app to confirm Contacts are restored. Invalid Contacts payloads must be quarantined without blocking other Wallet Sync modules.

### 📝 Description

LIVE-33960 registers Contacts in the Wallet Sync aggregate after the module-isolation work in #20555.

Contacts are selected through the public selector, applied with `setContacts` on Desktop and Mobile, and persisted with the Wallet Sync metadata to prevent an app restart from overwriting the distant state. The Desktop encrypted default now includes the Contacts state.

The aggregate test proves Contacts round-trip normally and that an invalid Contacts slice is quarantined while a healthy module continues syncing.

https://github.com/user-attachments/assets/9e994f06-29cf-4975-8d9a-aca5031f5360

https://github.com/user-attachments/assets/1cd17c45-eeb0-4093-ba00-cdc5fd77c4d6

https://github.com/user-attachments/assets/0b47b5f8-609a-4af0-b81e-d27d09f2912d

### ❓ Context

- **JIRA or GitHub link**: https://ledgerhq.atlassian.net/browse/LIVE-33960
- **Depends on**: #20555 (Wallet Sync module isolation / quarantine)

---

### DIFF
--- a/apps/ledger-live-desktop/src/main/db/index.ts
+++ b/apps/ledger-live-desktop/src/main/db/index.ts
@@ -12,6 +12,7 @@ import {
   exportWalletState,
   initialState as walletInitialState,
 } from "~/renderer/reducers/wallet.core";
+import { contactsInitialState } from "@domain/entity-contact";
 import { LARGE_SCREEN_UPSELL_MODAL } from "@features/flow-large-screen-upsell";
 import { encryptData, decryptData } from "~/main/db/crypto";
 import { readFile, writeFile } from "~/main/db/fsHelper";
@@ -164,7 +165,7 @@ type EncryptedAppKeyPath = (typeof encryptedDataPaths)[number][1];
 const ENCRYPTION_PATH_DEFAULTS: Record<EncryptedAppKeyPath, unknown> = {
   accounts: [],
   trustchain: trustchainInitialState,
-  wallet: exportWalletState(walletInitialState),
+  wallet: exportWalletState(walletInitialState, contactsInitialState),
 };
 
 for (const [, keyPath] of encryptedDataPaths) {
--- a/apps/ledger-live-desktop/src/main/db/index.ts
+++ b/apps/ledger-live-desktop/src/main/db/index.ts
@@ -165,7 +165,7 @@ type EncryptedAppKeyPath = (typeof encryptedDataPaths)[number][1];
 const ENCRYPTION_PATH_DEFAULTS: Record<EncryptedAppKeyPath, unknown> = {
   accounts: [],
   trustchain: trustchainInitialState,
-  wallet: exportWalletState(walletInitialState, contactsInitialState),
+  wallet: exportWalletState({ wallet: walletInitialState, contacts: contactsInitialState }),
 };
 
 for (const [, keyPath] of encryptedDataPaths) {
--- a/apps/ledger-live-mobile/src/components/DBSave.ts
+++ b/apps/ledger-live-mobile/src/components/DBSave.ts
@@ -4,7 +4,6 @@ import { postOnboardingSelector } from "@ledgerhq/live-common/postOnboarding/red
 import {
   exportWalletState,
   walletStateExportShouldDiffer,
-  walletSelector,
 } from "~/reducers/wallet";
 import isEqual from "lodash/isEqual";
 import throttleFn from "lodash/throttle";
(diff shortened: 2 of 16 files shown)


apps/ledger-live-mobile/src/components/DBSave.ts
```diff
@@ -143,7 +143,7 @@ function useFlushMechanism({ flush, cancel }: { flush: () => void; cancel: () =>
 }
 
 function walletExportSelector(state: State) {
-  return exportWalletState(walletSelector(state));
+  return exportWalletState(walletSelector(state), state.contacts);
```
mcayuelas-ledger: maybe we can have also `contactSelector`?

mcayuelas-ledger (dismissed):
Looking good overall! 

Maybe one improvment — each time we add a new Wallet Sync module, we'll need to update this `walletStateExportShouldDiffer` call manually. Same pattern repeats in a few other places (hooks, export/import, etc).

Might be worth exploring a **registry pattern** later where modules auto-register. Would save some boilerplate when the next feature ships.

apps/ledger-live-mobile/src/reducers/wallet.ts
```diff
@@ -50,24 +51,35 @@ export type ExportedWalletState = {
...
     starredAccountIds: string[];
   };
+  contacts: ContactsState["contacts"];
   recentAddresses: RecentAddressesState;
 };
 
-export const exportWalletState = (state: WalletState): ExportedWalletState => ({
+export const exportWalletState = (
+  state: WalletState,
+  contacts: ContactsState,
```
gre-ledger: we shouldn't have to pass `contacts` as a second parameter I think.
It may be better to generalize `exportWalletState` and `walletStateExportShouldDiffer` to take the full `State` instead, because as @mcayuelas-ledger pointed out , it will not scale.

apps/ledger-live-mobile/src/components/DBSave.ts
gre-ledger: I prefer we just make exportWalletState also taking the `state` 

function walletExportSelector(state: State) {
  return exportWalletState(state);
}

making this function probably useless and we could directly use such new exportWalletState that have same signature

gre-ledger:
> @mcayuelas-ledger  Might be worth exploring a registry pattern later where modules auto-register. Would save some boilerplate when the next feature ships.

It's a good point, but i think we may want first to figure out a unified pattern to generalize the logic related to "diff" and "export" ( to me it is close to topic of https://ledgerhq.atlassian.net/browse/LIVE-35384 and maybe will be part of it )
because ultimately, the best would be that each slice / module exports a generic function to solve these and then we won't have to solve it "manually" and implement each of the function.

gre-ledger (dismissed):

apps/ledger-live-desktop/src/main/db/index.ts
```diff
@@ -164,7 +165,7 @@ type EncryptedAppKeyPath = (typeof encryptedDataPaths)[number][1];
 const ENCRYPTION_PATH_DEFAULTS: Record<EncryptedAppKeyPath, unknown> = {
   accounts: [],
   trustchain: trustchainInitialState,
-  wallet: exportWalletState(walletInitialState),
+  wallet: exportWalletState({ wallet: walletInitialState, contacts: contactsInitialState }),
```
gre-ledger, line 168: (it's ok for now, but i'll have to more deeply look at why we need to do this `ENCRYPTION_PATH_DEFAULTS` at all)

gre-ledger (approved):

apps/ledger-live-desktop/src/main/db/index.ts
gre-ledger: fyi i created a connex task for platform team https://ledgerhq.atlassian.net/browse/LIVE-36438

mcayuelas-ledger (approved):

PR #21084, opened 2026-08-24
[LWD] feat(baanx-test-client): add test-only client for the baanx api

cunhabruno:
### 📝 Description

Adds `@ledgerhq/baanx-test-client`, a private **test-only** client for the Baanx API (sandbox `dev.api.baanx.com` by default, base URL injectable). It exists because nothing in the repo implements the login that produces a Card session token — `shared/api-services/src/services/card` already sends `x-client-key` and expects a token to be injected, and this fills that gap for tests.

**Authentication branches on the response body, not the status code.** Baanx answers HTTP 200 even when login has not completed, so an onboarding `phase`, an OTP challenge, a token, or none of the above each map to a distinct typed error or session. A blank/`null`/whitespace token counts as absent, so this can never yield `Bearer null`. The OTP challenge is answered by deriving a code from the test user's TOTP setup key, which keeps suites headless; a code with under two seconds of life left is held back until the next window, since a code expiring in flight reads as a credentials failure and is a real source of flaky auth.

**`baanxRequest` covers data setup and validation.** Endpoint-agnostic on purpose: it attaches the bearer token, client key and region header, reuses the cached token, retries once on a `401` (a long suite can outlive a 6h token — mirroring what `cardBaseQuery` already does), and raises the same typed errors as login, so a `429` during fixture setup reads exactly like one during login.

Usage:

```ts
import { getBaanxAuthToken, baanxRequest } from "@ledgerhq/baanx-test-client";

// Playwright/Detox globalSetup — authenticate once, share the token
const { accessToken } = await getBaanxAuthToken();

// data setup / validation
const { data } = await baanxRequest<UserProfile>({ path: "/v1/user" });
await baanxRequest({ path: "/v1/some-resource", method: "POST", body: { … } });
```

```bash
TOKEN=$(pnpm --silent --filter @ledgerhq/baanx-test-client token)
```

**Security posture.** Credentials, the client key and the TOTP secret come from environment variables only — never CLI flags (they leak into process lists and shell history), never committed fixtures, nothing written to disk. The package contains no logging at all; bodies attached to errors pass through `redactBody` first, and env vars are named but never valued in messages. See `.env.sample`.

Not wired into `shared/env`: those definitions get bundled into the shipped apps, and the password and setup key must never enter an app bundle.

**Tests.** 261 unit tests, no network and no wall-clock dependency (`fetchImpl` and the clock are injected); TOTP is verified against the RFC 6238 vectors. `src/index.test.ts` pins the public API so it only grows deliberately. A separate opt-in live test (`pnpm --filter @ledgerhq/baanx-test-client test-integ`) logs in for real and validates `GET /v1/user`; it is excluded from `pnpm test` and skips when credentials are absent, so CI without secrets is unaffected.

**One assumption reviewers should know about.** Baanx's public docs describe `POST /v1/auth/login/otp` as an **SMS** trigger and mention TOTP nowhere. This is built on a TOTP setup key because our test users are provisioned with an authenticator secret — confirmed against the sandbox, where the derived code was accepted. Documented in the README, along with why a derived code cannot satisfy an SMS challenge should that ever change.

### 🔗 Context

- **JIRA / GitHub issue**: _none yet — needs a ticket before review_
- **ADR** (if any): n/a

---


e2e/tooling/baanx-test-client/README.md
ysitbon, line 1: I think it's a good candidate to `e2e/tooling` instead

cunhabruno:
Superseded by #21415, which carries the same work on `feat/baanx-test-client-QAA-1508-shared-api-client` — history squashed from 13 commits to 2 (the test client, and the session injection), and filed under [QAA-1508](https://ledgerhq.atlassian.net/browse/QAA-1508).

Closing this one.

[QAA-1508]: https://ledgerhq.atlassian.net/browse/QAA-1508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #21092, opened 2026-08-25
feat(pay): add native Request receive screen (LIVE-35188)

tonykhaov:
### 📝 Description

The Pay tab Request flow had no receive screen on mobile: after picking a stablecoin account, there was nowhere to show the QR and address.

This adds a full-screen Request receive route on the Pay stack. The host looks up the selected account from the store via `accountId` / optional `parentId` (serializable route params), derives address / asset / network, injects copy, and wraps the shared `@features/flow-pay-card-request` UI. Safe-area padding stays in the app (`SafeAreaView` + canvas). Share currently uses the OS share sheet with the address (card image capture is later). Verify is stubbed.

https://github.com/user-attachments/assets/788bbc06-b8f2-4788-9a11-c81bc5109b72

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35188](https://ledgerhq.atlassian.net/browse/LIVE-35188)
- **ADR** (if any): —
- **Figma**: [Pay • Production 8328-101662](https://www.figma.com/design/WYQQwem0QyG2fTuaAKAe3Y/Pay-%E2%80%A2-Production?node-id=8328-101662&m=dev)

[LIVE-35188]: https://ledgerhq.atlassian.net/browse/LIVE-35188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/ledger-live-mobile/src/mvvm/features/PayTab/screens/RequestReceive/usePayTabRequestReceiveViewModel.ts
```diff
@@ -0,0 +1,83 @@
...
+  useEffect(() => {
+    if (!account) {
+      goBack();
+    }
+  }, [account, goBack]);
+
+  const data = useMemo(
+    () => (account ? deriveRequestReceiveData(account, parentAccount ?? undefined) : undefined),
+    [account, parentAccount],
+  );
```
claudiiafg, line 34: what happens if the account is a token account and parentAccount is missing? will this crash?

apps/ledger-live-mobile/src/mvvm/features/PayTab/screens/RequestReceive/usePayTabRequestReceiveViewModel.ts
```diff
@@ -0,0 +1,83 @@
...
+        save: t("payTab.request.actions.save"),
+        verify: t("payTab.request.actions.verify"),
+      },
+    }),
+    [t, data],
+  );
+
+  return {
+    isOpen: true,
+    address: data?.address ?? "",
```
claudiiafg, line 69: the address is used for the qr code right? what happens if it's missing?

features/flow/pay-card-request/src/components/RequestReceive/RequestReceiveView.native.tsx
```diff
@@ -1,7 +1,65 @@
...
+
+  return (
+    <Box lx={{ flex: 1, backgroundColor: "canvas" }} testID="pay-card-request-receive">
+      <Box lx={{ alignItems: "flex-start", paddingHorizontal: "s4", paddingTop: "s8" }}>
+        <IconButton
+          icon={Close}
+          appearance="no-background"
+          size="md"
+          onPress={onClose}
+          accessibilityLabel="Close"
```
claudiiafg, line 32: should this be translated?

apps/ledger-live-mobile/src/mvvm/features/PayTab/screens/RequestReceive/usePayTabRequestReceiveViewModel.ts
tonykhaov: it calls `getMainAccount` and parentAccount is optional because there are accounts without a parent

it's possible but very unlikely that it can crash because we get from the MAD the same account + parent

apps/ledger-live-mobile/src/mvvm/features/PayTab/screens/RequestReceive/usePayTabRequestReceiveViewModel.ts
tonykhaov: indeed but again very unlikely that it can happen since it's derived from the MAD.

I'll just create a polishing ticket so we don't forget to handle it in case

claudiiafg (approved):

PR #21094, opened 2026-08-25
fix(e2e): handle Buy/Sell network drawer and widen webview wait

kentoforik:
## Summary

Today's CI run ([32792369689](https://github.com/LedgerHQ/ledger-live/actions/runs/32792369689)) had Buy/Sell E2E failures on both Android and iOS, all timing out on the same assertion (`crypto-amount-option-button` not found after 60s in `BuySellPage.expectBuyScreenToBeVisible`). Two independent causes:

- **Native "Select network" drawer blocks entry for multi-network assets.** A recent product change (LIVE-34547) makes tapping Buy on the asset/market page open a native network-selection drawer first for any currency with more than one network (ETH's L2s, USDT's multi-chain variants). `runNavigateToBuyFromAssetPageTest`/`runNavigateToBuyFromMarketPageTest` never handled it, so they waited 60s for a webview that was never reached. Fixed by calling the existing `modularDrawer.selectNetworkIfAsked(...)` / `selectFirstAccountIfAsked()` helpers right after the buy quick-action tap (no-ops for single-network currencies).
- **CAL `currency.list` latency.** The Buy screen's currency list resolves 700+ token ids against CAL one HTTP call at a time; even concurrency-bounded, that measures 60-90s on its own, landing right on the 60s wait and flaking on nothing but that latency. Widened `cryptoCurrencySelectorTimeout` to 120s (precedent for 120s waits already exists elsewhere in this suite: `evmStake.page.ts`, `portfolio.page.ts`). The backend latency itself is tracked separately (QAA-1497) and is out of scope here — not touching `libs/ledger-live-common`.

## Test plan
- [x] `node ./scripts/typecheck.js` (e2e/mobile) — clean
- [x] `oxlint page/trade/buySell.page.ts specs/buySell/buySell.ts` — clean
- [x] `pnpm format` (e2e) — no outstanding diff
- [x] [CI run on this branch to confirm the previously-failing Buy/Sell specs pass](https://github.com/LedgerHQ/ledger-live/actions/runs/32833416199)

### DIFF
--- a/e2e/mobile/page/trade/buySell.page.ts
+++ b/e2e/mobile/page/trade/buySell.page.ts
@@ -44,9 +44,15 @@ export default class BuySellPage {
     await waitForElementById(app.common.walletApiWebview, 60000, { checkVisibility: false });
   }
 
+  // App-side CAL lookup for the Buy screen's currencies (700+ ids) measures 60-90s;
+  // 60s flakes on that alone. Latency is tracked separately, not fixed here.
+  cryptoCurrencySelectorTimeout = 120000;
+
   @Step("Expect Buy screen to be visible")
   async expectBuyScreenToBeVisible() {
-    await waitWebElementByTestId(this.cryptoCurrencySelector);
+    await waitWebElementByTestId(this.cryptoCurrencySelector, {
+      timeout: this.cryptoCurrencySelectorTimeout,
+    });
     await detoxExpect(getWebElementsByIdAndText("", "You will pay")).toExist();
     await detoxExpect(getWebElementByTestId(this.amountInputSectionId())).toExist();
     await detoxExpect(getWebElementByTestId(this.buyQuickAmountButtonId("400"))).toExist();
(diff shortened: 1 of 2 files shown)


jeportie:
Ran your branch to check the four Buy specs that were failing: [32833416199](https://github.com/LedgerHQ/ledger-live/actions/runs/32833416199).

**The drawer fix is confirmed.** `[ETH] Buy entry point from asset page` and `[USDT (Ethereum)] Buy entry point from market page` both pass now, on Android *and* iOS. That matches what our artifacts showed independently — the native view hierarchy at failure had **no WebView node**, only the asset page with `buy-button`, while the WebView polls kept executing and returning. A native drawer in front of a webview that never opened. Your diagnosis was right.

**Two Android specs still fail, and neither is CAL latency.** The shards report green but the results JSON has `success: false`, `numFailedTests: 2`:

`[BTC] - Buy entry point from account page`
```
Error: Web element crypto-amount-option-button not found after 120000ms
Test Failed: Wait for [LooperIdlingResource-97-mqt_v_js] to become idle timed out
    at BuySellPage.expectBuyScreenToBeVisible (buySell.page.ts:58:9)
```
Note it times out at the **new 120000ms** — the widening was applied and did not help. `LooperIdlingResource ... to become idle` points at the JS thread never idling rather than a slow network, so I do not think the 736-id fan-out explains this one.

`[BTC] - Buy from portfolio page`
```
Error: [retryUntilTimeout] Timed out after 60000ms
Test Failed: Action will not be performed because the target view does not match:
(view has effective visibility <VISIBLE> and view.getGlobalVisibleRect() covers at least <75> percent)
Target view: "ReactEditText{id=2862, visibility=VISIBLE, width=908, height=90, ...}"
```
Different failure entirely — it gets *past* the Buy screen and dies on an amount input that is under 75% visible. Keyboard overlap would be my first guess.

Both pass on iOS, so both are Android-only.

Nothing here argues against merging: two of four specs are genuinely fixed and this is test-only. Might be worth narrowing the description so the remaining two are not assumed covered, and splitting them into a follow-up — happy to take that if useful.

One small thing: the PR cites QAA-1497 for the CAL latency, which is our umbrella epic. The PTX-owned ticket with the measurements and the proposed fix is **LIVE-36229**.

jeportie (approved):

PR #21108, opened 2026-08-25
[LWDM] fix(coin-evm): cap and stop re-copying explorer batches

YazhuEth:
### 📝 Description

**Previous behaviour.** `fetchPaginatedOpsWithRetries` recursed for as long as the Ledger explorer returned a continuation token, with no cap, and rebuilt the whole accumulator on every batch (`[...previousOperations, ...operationsBatch]`). On a very large address that exhausts the V8 heap and takes the app down:

```
[30844:0x12c00970000] 640605 ms: Mark-Compact (reduce) 3934.1 (3982.5) -> 3934.1 (3960.2) MB … last resort; GC in old space requested
OOM error in V8: Ineffective mark-compacts near heap limit — JavaScript heap out of memory
```

Reproduced while syncing `0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe` (Ethereum Foundation).

**Fix, two parts.**

1. **Batches are appended in place** instead of spread into a new array. The spread made the accumulation quadratic and doubled the peak footprint, since both copies were live during the copy. Measured: the explorer serves **500 records per call** whatever `batchSize` asks for (`batch_size=100` → 100, `500` → 500, `1000` → 500, `5000` → 500), so a large address means thousands of sequential calls, each re-copying everything accumulated so far.
2. **The number of operations accumulated per address is capped** — `DEFAULT_MAX_OPERATIONS`, overridable per network via `explorer.maxOperations`. Hitting the cap truncates the history and logs it rather than throwing: a bounded partial history beats crashing the app.

Counted in operations rather than batches precisely because of the 500-vs-`batchSize` gap above — a batch count would bound something 20x smaller than `DEFAULT_BATCH_SIZE` suggests.

`coin-stellar` (`maxOperations`) and `coin-tezos` (`maxTxQuery`) already bound their equivalent fetch-all loops; this explorer was the only one without a bound.

**Scope.** This path returns `NO_TOKEN` to the generic coin framework, so the framework's pagination loop makes exactly one call here and cannot bound it — the fix has to live in the module. The defect pre-dates the framework pagination work (LIVE-36317) and is independent of it.

**Tests.** Two cases added: the walk stops once the operation cap is reached even while the explorer keeps issuing a token, and it stops earlier when the explorer drops the token. A third checks that a retry does not consume budget. 1122 tests pass across the module's 54 suites.

**Not verified:** that this removes the OOM on that specific address end to end — it needs a run of the app against it. The reasoning and the measurements are above, but this is not a measurement of the fix itself.

**Known follow-ups, out of scope here:**
- `DEFAULT_BATCH_SIZE = 10_000` is misleading since the explorer caps at 500. No functional effect; left alone.
- `coin-kaspa`, `coin-aptos` and `coin-filecoin` have the same unbounded accumulate-and-re-copy shape. No field evidence of a crash on those, so they are not touched here.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36446](https://ledgerhq.atlassian.net/browse/LIVE-36446) — found while QA'ing [LIVE-36317](https://ledgerhq.atlassian.net/browse/LIVE-36317), independent of it

[LIVE-36317]: https://ledgerhq.atlassian.net/browse/LIVE-36317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-36446]: https://ledgerhq.atlassian.net/browse/LIVE-36446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/evm-ledger-explorer-batch-cap.md
+++ b/.changeset/evm-ledger-explorer-batch-cap.md
@@ -1,0 +1,15 @@
+---
+"@ledgerhq/coin-evm": patch
+---
+
+fix(evm): bound the Ledger explorer's fetch-all pagination and stop re-copying the accumulator
+
+`fetchPaginatedOpsWithRetries` recursed for as long as the explorer returned a continuation token,
+with no cap, and rebuilt the whole accumulator on every batch (`[...previous, ...batch]`). On very
+large addresses that exhausted the V8 heap and took the app down with `Ineffective mark-compacts
+near heap limit`.
+
+Batches are now appended in place, and the number of operations accumulated per address is capped
+(`DEFAULT_MAX_OPERATIONS`, overridable per network via `explorer.maxOperations`). Hitting the cap
+truncates the history and logs it rather than throwing. `coin-stellar` and `coin-tezos` already
+bound their equivalent loops this way.
--- a/libs/coin-modules/coin-evm/src/config.ts
+++ b/libs/coin-modules/coin-evm/src/config.ts
@@ -53,6 +53,7 @@ export type EvmConfig = {
         type: "ledger";
         explorerId: LedgerExplorerId;
         batchSize?: number | undefined;
+        maxOperations?: number | undefined;
       }
     | {
         type: "none";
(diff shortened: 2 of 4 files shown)


YazhuEth:
Folded into #21015 — same review, one PR. The commit is cherry-picked there unchanged (`fix(coin-evm): cap and stop re-copying explorer batches`), and LIVE-36446 is referenced in that PR's description.

PR #21128, opened 2026-08-25
[LWDM] feat(contacts): orchestrate and wire device intents (LIVE-36248)

OlivierFreyssinet:
### 📝 Description

Contacts now routes all device operations through the Device Intent Executor on Desktop and Mobile.

The executor preserves the calling flow lifecycle:
- Contacts dialogs are hidden while a device intent is active
- Edit dialogs are dismissed before handing off to the device so they don't flash back after completion
- DIE cancellation or errors clear the dialog stack.

Removed both unused \"edit scope\" and \"edit identifier\" intents, as only the combined operation is used.

**NB:** We are using the **real DIE**, but with **stub intents**: they do not yet perform real operations on the device and use placeholder UI.

https://github.com/user-attachments/assets/fd1549d8-929b-464c-919e-15571db717fa

https://github.com/user-attachments/assets/4fb1280f-bfb8-44b2-be46-6f93efcb9e70

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36248](https://ledgerhq.atlassian.net/browse/LIVE-36248)

[LIVE-36248]: https://ledgerhq.atlassian.net/browse/LIVE-36248


features/platform/contacts/src/contactDeviceIntentsPort.ts
```diff
@@ -11,41 +11,54 @@ import {
...
 }>;
 
-export type EditExternalAddressScopeIntentInput = Readonly<{
+export type RenameExternalContactResult = NonNullable<Contact["deviceCredentials"]>;
+
+export type EditExternalAddressInput = Readonly<{
   contact: Contact;
   address: ContactAddress;
-  label: ContactAddressLabel;
+  // TODO: rename to updatedLabel
```
deepyjr: I think it's already renamed?

Sebastien-Dav1d (dismissed):

features/platform/contacts/src/contactDeviceIntentsPort.ts
OlivierFreyssinet: lol good catch thanks

deepyjr (dismissed):

deepyjr (dismissed):

Sebastien-Dav1d (dismissed):

aussedatlo (approved):

deepyjr (approved):

Sebastien-Dav1d (approved):

PR #21137, opened 2026-08-25
[LWDM] feat(internet_computer): add neuron management flow (LIVE-29098)

amaslakov:
### 📝 Description

Fills in the ICP staking and neuron-management flows on mobile, which LIVE-29097 left as registered-but-empty navigator stubs. Mirrors the desktop flows from LIVE-29096: a neuron list, a per-neuron detail screen with the eleven governance actions, and the create-neuron staking flow, plus the account stake banner and staked-balance footer. All behind `llmIcpStaking`. Covered by 133 unit tests across 14 suites.

Three decisions a reviewer would otherwise have to reconstruct:

- **Manage Neurons is not gated on the neuron count.** Neurons only arrive via a device-signed `list_neurons`, and its Sync button lives inside the screen the action opens — gating on the count made a first neuron unreachable. The list renders its own empty state instead.
- **ICP has its own ConnectDevice** (`components/ConnectDevice.tsx`) rather than the shared one. A signed `list_neurons` carries the neuron snapshot in `extra.neurons` and is the only thing that can refresh it; the shared screen only files the pending operation, so the list would show stale neurons until the next account sync.
- **Staking ends on a success screen rather than navigating to the neuron list**, which is where desktop sends the user. A new neuron is not in the snapshot until a device-signed `list_neurons` fetches it, so the list would open without it; the success copy directs the user to Manage Neurons instead.

Not covered: no Detox specs yet, and the flows have been exercised on an emulator with a fixture rather than against a funded device.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-29098](https://ledgerhq.atlassian.net/browse/LIVE-29098?atlOrigin=eyJpIjoiYjQ1YmZkZWMxMGRjNGM5ZDg1YzUyNDc0M2E2OGQ3NzUiLCJwIjoiaiJ9)
- **ADR** (if any): 

[LIVE-29098]: https://ledgerhq.atlassian.net/browse/LIVE-29098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
+++ b/apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
@@ -14,6 +14,7 @@ import AmountInput from "~/screens/SendFunds/AmountInput";
 import { useAccountUnit } from "LLM/hooks/useAccountUnit";
 import ActionFooter from "../components/ActionFooter";
 import { useNeuronAction } from "./useNeuronAction";
+import MissingNeuron from "./MissingNeuron";
 import type { InternetComputerNeuronManageFlowParamList } from "./types";
 
 type Props = StackNavigatorProps<
@@ -28,8 +29,16 @@ type Props = StackNavigatorProps<
  */
 export default function IncreaseStake({ navigation, route }: Props) {
   const { t } = useTranslation();
-  const { account, transaction, updateTransaction, status, bridgePending, continueToDevice } =
-    useNeuronAction(navigation, route);
+  const {
+    account,
+    neuron,
+    backToList,
+    transaction,
+    updateTransaction,
+    status,
+    bridgePending,
+    continueToDevice,
+  } = useNeuronAction(navigation, route);
   const unit = useAccountUnit(account);
   const bridge = useAccountBridge<Transaction>(account);
   const [maxSpendable, setMaxSpendable] = useState<BigNumber | null>(null);
@@ -56,6 +65,8 @@ export default function IncreaseStake({ navigation, route }: Props) {
   );
 
   const amount = transaction?.amount ?? new BigNumber(0);
+
+  if (!neuron) return <MissingNeuron onBackToList={backToList} />;
 
   return (
     <SafeAreaView edges={["left", "right", "bottom"]} isFlex>
--- a/apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
+++ b/apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
@@ -1,0 +1,97 @@
+import { useAccountBridge } from "@ledgerhq/live-common/bridge/useAccountBridge";
+import type { Transaction } from "@ledgerhq/live-common/families/internet_computer/types";
+import { Flex, Text } from "@ledgerhq/native-ui";
+import BigNumber from "bignumber.js";
+import React, { useCallback, useEffect, useState } from "react";
+import { TrackScreen } from "~/analytics";
+import CurrencyUnitValue from "~/components/CurrencyUnitValue";
+import KeyboardView from "~/components/KeyboardView";
+import SafeAreaView from "~/components/SafeAreaView";
+import type { StackNavigatorProps } from "~/components/RootNavigator/types/helpers";
+import { ScreenName } from "~/const";
+import { useTranslation } from "~/context/Locale";
+import AmountInput from "~/screens/SendFunds/AmountInput";
+import { useAccountUnit } from "LLM/hooks/useAccountUnit";
+import ActionFooter from "../components/ActionFooter";
+import { useNeuronAction } from "./useNeuronAction";
+import type { InternetComputerNeuronManageFlowParamList } from "./types";
+
+type Props = StackNavigatorProps<
+  InternetComputerNeuronManageFlowParamList,
+  ScreenName.InternetComputerNeuronIncreaseStake
+>;
+
+/**
+ * Tops up an existing neuron. This is a ledger transfer, not a governance call, so it needs an
+ * amount — but unlike a send there is no recipient to pick: prepareTransaction derives the neuron's
+ * own governance subaccount. That is why it lives here rather than entering the send flow.
+ */
+export default function IncreaseStake({ navigation, route }: Props) {
+  const { t } = useTranslation();
+  const { account, transaction, updateTransaction, status, bridgePending, continueToDevice } =
+    useNeuronAction(navigation, route);
+  const unit = useAccountUnit(account);
+  const bridge = useAccountBridge<Transaction>(account);
+  const [maxSpendable, setMaxSpendable] = useState<BigNumber | null>(null);
+
+  useEffect(() => {
+    if (!transaction) return;
+    let cancelled = false;
+    bridge.estimateMaxSpendable({ account, transaction }).then(estimate => {
+      if (!cancelled) setMaxSpendable(estimate);
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, [account, bridge, transaction]);
+
+  const onChange = useCallback(
+    (amount: BigNumber) => updateTransaction(tx => ({ ...tx, amount })),
+    [updateTransaction],
+  );
+
+  const amount = transaction?.amount ?? new BigNumber(0);
+
+  return (
+    <SafeAreaView edges={["left", "right", "bottom"]} isFlex>
+      <TrackScreen
+        category="Manage Neurons ICP Flow"
+        name="IncreaseStake"
+        flow="stake"
+        action="increase_stake"
+      />
+      <KeyboardView style={{ flex: 1 }}>
+        <Flex flex={1} p={6} style={{ gap: 16 }}>
+          <Text variant="body" color="neutral.c70">
+            {t("internetComputer.manageNeuronFlow.increaseStake.description")}
+          </Text>
+          <AmountInput
+            account={account}
+            onChange={onChange}
+            value={amount}
+            error={amount.gt(0) ? status.errors.amount : null}
+            warning={status.warnings.amount}
+            testID="icp-increase-stake-amount-input"
+          />
+          <Flex flexDirection="row" justifyContent="space-between">
+            <Text variant="small" color="neutral.c70">
+              {t("send.amount.available")}
+            </Text>
+            {maxSpendable ? (
+              <Text variant="small" fontWeight="semiBold" color="neutral.c70">
+                <CurrencyUnitValue disableRounding showCode unit={unit} value={maxSpendable} />
+              </Text>
+            ) : null}
+          </Flex>
+        </Flex>
+      </KeyboardView>
+      <ActionFooter
+        status={status}
+        bridgePending={bridgePending}
+        onContinue={continueToDevice}
+        canContinue={amount.gt(0)}
+        showAmountError={amount.gt(0)}
+      />
+    </SafeAreaView>
+  );
+}
--- a/apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
+++ b/apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
@@ -37,9 +37,14 @@ export default function IncreaseStake({ navigation, route }: Props) {
   useEffect(() => {
     if (!transaction) return;
     let cancelled = false;
-    bridge.estimateMaxSpendable({ account, transaction }).then(estimate => {
-      if (!cancelled) setMaxSpendable(estimate);
-    });
+    bridge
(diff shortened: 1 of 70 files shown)


apps/ledger-live-mobile/src/families/internet_computer/components/ValidationError.tsx
```diff
@@ -0,0 +1,164 @@
...
+  increase_stake: ScreenName.InternetComputerNeuronIncreaseStake,
+  set_dissolve_delay: ScreenName.InternetComputerNeuronSetDissolveDelay,
+  increase_dissolve_delay: ScreenName.InternetComputerNeuronSetDissolveDelay,
+  stake_maturity: ScreenName.InternetComputerNeuronStakeMaturity,
+  split_neuron: ScreenName.InternetComputerNeuronSplit,
+  add_hot_key: ScreenName.InternetComputerNeuronAddHotKey,
+  // Back to the followee list rather than the topic picker: the topic is already chosen and the list
+  // it holds is what a retry is likely to be correcting.
+  follow: ScreenName.InternetComputerNeuronFollowees,
+  refresh_voting_power: ScreenName.InternetComputerNeuronRefreshVotingPower,
```
ishaba: `RETRY_SCREEN` maps `refresh_voting_power` → `ScreenName.InternetComputerNeuronRefreshVotingPower`, and `retry()` navigates with no params:
```(navigation as unknown as { navigate: (screen: string) => void }).navigate(retryScreen);```

**Review by ishaba — DISMISSED** [review]

File: apps/ledger-live-mobile/src/families/internet_computer/common/index.ts [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+ * Background sync cannot fetch neurons — that needs a device signature — so a signed `list_neurons`
+ * is the only thing that refreshes them, and its result rides in `extra.neurons`. The bridge does
+ * the same fold on the next sync (bridgeHelpers/account.ts); this makes it visible immediately,
+ * which the shared ConnectDevice does not do (it only adds the pending operation).
+ *
+ * A `manage_neuron` call returns no snapshot, so `transaction` lets the command the canister just
+ * accepted be replayed onto the stored neuron instead. `lastUpdatedMSecs` deliberately does not move
+ * for that: the neuron is only as fresh as the last real read.
+ */
+export const applyNeuronOperation = (
```
francois-guerin-ledger, line 27: **Stake banner reappears after a successful stake.** `applyNeuronOperation` can only update the local neuron snapshot via `applyNeuronCommand`, which requires `transaction.neuronId` — but a `create_neuron` (stake) transaction never has one. The account page's `getBannerState` (`libs/coin-modules/coin-internet_computer/src/common-logic/neuron.ts:577`) checks `fullNeurons.length === 0` with no other staleness handling.

Concretely: a user stakes part of their balance, signing succeeds, and back on the account page the "Stake ICP" banner is still shown as if nothing happened, since `fullNeurons` is still empty. The coin module already returns `operation.extra.createdNeuronId` on a successful stake, but nothing in this diff reads it (confirmed via grep) to show a pending-neuron placeholder in the meantime.

apps/ledger-live-mobile/src/families/internet_computer/components/ValidationError.tsx
```diff
@@ -0,0 +1,174 @@
...
+  return (
+    <SafeAreaView style={[styles.root, { backgroundColor: colors.background }]}>
+      {/* Withholding Retry is not enough on Android: this screen replaced the device screen, so
+          hardware Back pops to SelectDevice, which auto-selects the last device and signs again.
+          Guarded only where a retry was refused — where it was offered, Back is what Retry does. */}
+      {!canRetry && <PreventNativeBack />}
+      <TrackScreen category={category} name="ValidationError" flow="stake" action={action} />
+      <ValidateError
+        error={error}
+        {...(canRetry && { onRetry: retry })}
```
francois-guerin-ledger, line 142: **Dead end when a staking broadcast fails after signing.** In `StakingFlow`, `onBackToList` is never passed, so when `canRetry` is false the `primaryButton` resolves to `null` and the user is left with only a bare Close button.

Scenario: create-neuron signing succeeds (`signed=true`) but the broadcast then fails (e.g. network/timeout, not in `NOTHING_EXECUTED`); `create_neuron` isn't in `IDEMPOTENT_COMMANDS` so `canRetry` is false. The user has no way to retry or check whether the neuron was actually created.

apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/RefreshVotingPower.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  })
+                : formatDuration(seconds)}
+            </Text>
+          )}
+          <Button
+            type="main"
+            size="small"
+            mt={4}
+            // The canister declares a neuron's id optional and a refresh call has to name one.
+            disabled={!neuron.id}
```
francois-guerin-ledger: `disabled={!neuron.id}` — `ICPNeuron.id` is typed `bigint | undefined`, and `!0n === true`, so a neuron whose id is `0n` would have this button permanently (and incorrectly) disabled.

`components/NeuronList.tsx` in this same PR handles the identical case correctly with `item.id !== undefined`; worth mirroring that check here.

apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <Flex flex={1} p={6} style={{ gap: 16 }}>
+          <Text variant="body" color="neutral.c70">
+            {t("internetComputer.manageNeuronFlow.increaseStake.description")}
+          </Text>
+          <AmountInput
+            account={account}
+            onChange={onChange}
+            value={amount}
+            error={amount.gt(0) ? status.errors.amount : null}
+            warning={status.warnings.amount}
```
francois-guerin-ledger: `warning={status.warnings.amount}` is permanently dead: the ICP bridge's `getTransactionStatus` never populates `warnings.amount` for any ICP transaction, only `warnings.staking`. Looks like it was copied from the generic Send flow, where `warnings.amount` is a real field.

Same issue on the two near-identical copies of this screen: `NeuronManageFlow/SplitNeuron.tsx:76` and `StakingFlow/Amount.tsx:106`.

apps/ledger-live-mobile/src/families/internet_computer/components/ValidationError.tsx
```diff
@@ -0,0 +1,174 @@
...
+  category,
+  action,
+  onBackToList,
+}: Props) {
+  const { colors } = useTheme();
+  const { t } = useTranslation();
+  const { error, signed, transaction } = route.params;
+  const command = transaction?.type;
+
+  const onClose = useCallback(() => {
```
francois-guerin-ledger, line 100: `RETRY_SCREEN` is keyed purely by the transaction's runtime `command` string with no navigator scoping, even though `ValidationError` is shared between `NeuronManageFlow` and `StakingFlow`.

Currently harmless only because `StakingFlow`'s one transaction type (`create_neuron`) isn't a `RETRY_SCREEN` key, so `retry()` falls back to `goBack()`. If a future `StakingFlow` transaction type ever reuses a key that currently only makes sense inside `NeuronManageFlow` (e.g. `increase_stake`, `set_dissolve_delay`), `retry()` would silently `popTo` a screen that isn't in the current navigator's stack.

apps/ledger-live-mobile/src/families/internet_computer/components/ConnectDevice.tsx
```diff
@@ -0,0 +1,164 @@
+import { getMainAccount } from "@ledgerhq/live-common/account/helpers";
```
francois-guerin-ledger, line 1: This reimplements the entire shared `~/screens/ConnectDevice.tsx` scaffold (SafeAreaView/TrackScreen/DeviceAction wiring, `useBroadcast`, a hand-rolled `hasBroadcast` mount-guard ref, `route.name.replace` navigation) — about 140 lines — purely to swap `addPendingOperation` for `applyNeuronOperation`.

The shared screen already exposes an `onSuccess`/`onError` extension point (used by Polkadot and Near for the same kind of per-family post-broadcast customization), and `DeviceAction`'s `renderOnResult` already gives the once-per-mount guarantee this file reimplements by hand. Any future change to the shared screen's conventions (broadcast error handling, analytics props, etc.) has to be manually re-applied here or the fork silently drifts.

apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/IncreaseStake.tsx
```diff
@@ -0,0 +1,118 @@
...
+    backToList,
+    transaction,
+    updateTransaction,
+    status,
+    bridgePending,
+    continueToDevice,
+  } = useNeuronAction(navigation, route);
+  const unit = useAccountUnit(account);
+  const bridge = useAccountBridge<Transaction>(account);
+  const [maxSpendable, setMaxSpendable] = useState<BigNumber | null>(null);
```
francois-guerin-ledger: The `maxSpendable` state + debounced `useEffect` + `bridge.estimateMaxSpendable(...).then/catch` block here is duplicated near-verbatim (including the comments and `console.warn` text) in `StakingFlow/Amount.tsx:44-70`. Worth extracting into a shared hook — the two copies already have slightly different dependency-array variable names for the same underlying value, which is how these drift over time.

apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/types.ts
```diff
@@ -1,12 +1,81 @@
...
+    // ConnectDevice sends the user back here with this set, so the last-connected device is
+    // not auto-selected again.
+    forceSelectDevice?: boolean;
+  };
+  [ScreenName.InternetComputerNeuronConnectDevice]: NeuronSigningRoute & {
+    device: Device;
+    appName?: string;
+    selectDeviceLink?: boolean;
+    analyticsPropertyFlow?: string;
+  };
```
francois-guerin-ledger, line 69: `selectDeviceLink?: boolean` here and `deviceId?: string` a few lines below (plus their `StakingFlow/types.ts` equivalents) are declared on route param types but never set via `navigate(...)` or read via `route.params` anywhere in this diff. Dead fields — worth dropping, since a future screen could otherwise start reading one as if it were populated.

apps/ledger-live-mobile/src/families/internet_computer/NeuronManageFlow/NeuronDetails.tsx
```diff
@@ -0,0 +1,448 @@
...
+        accountId: route.params.accountId,
+        parentId: route.params.parentId,
+      }),
+    [navigation, route.params.accountId, route.params.parentId],
+  );
+
+  // Disburse and a refresh both drop a neuron from the snapshot while this screen may still name it.
+  if (!neuron) return <MissingNeuron onBackToList={backToList} />;
+
+  const neuronId = neuron.id?.toString() ?? "";
```
francois-guerin-ledger, line 117: Minor consistency nit: this screen recomputes ~10 derived values (`getNeuronActionPermissions`, `isDeviceControlledNeuron`, dissolve/voting-power calcs, maturity sum) plus several `toBigNumber()` conversions directly in the render body with no `useMemo`, unlike most sibling screens in this PR. The underlying coin-module functions are O(1) so this isn't a measurable perf issue today, just an inconsistency with the rest of the PR.

francois-guerin-ledger (commented):
Automated review pass (correctness + simplification/efficiency), effort level: high. 9 findings below, most severe first. Two additional candidates were investigated and refuted after verification (StakeMaturity's omitted `pristineField`, and a feared StakeBanners navigation issue) and are not included.

apps/ledger-live-mobile/src/families/internet_computer/components/ConnectDevice.tsx
amaslakov: The shared `useSignedTxHandler` can't carry ICP: it wraps every broadcast error in `TransactionBroadcastError` (dropping the names the retry gate keys on), passes no signed flag, puts the whole operation — bigint neurons included — into route params, and folds with `addPendingOperation` only, which hides a signed list_neurons until the next sync. `onSuccess` bypasses that handler, but nothing passes it today (Polkadot/Near only declare the type), and going through it would keep all the broadcast/fold/error code here while changing this step's analytics. Happy to move just the `DeviceAction` scaffold behind the shared screen via `onSuccess` if you'd prefer that.

ishaba (approved):

francois-guerin-ledger (approved):
LGTM

claudiiafg (approved):

sarneijim (approved):

PR #21139, opened 2026-08-25
[LWDM] refactor(architecture): move InfoState to shared (LIVE-36304)

OlivierFreyssinet:
### 📝 What this does

`InfoState` is the layout behind our "something happened" screens: a centered icon or illustration, an optional title and description, an optional banner, and up to two buttons. It shows up on device-flow errors, success screens, unsupported-app warnings, and similar.

It existed **twice** — one copy in `ledger-live-desktop`, a near-identical copy in `ledger-live-mobile`. This PR moves it into a single package, `@shared/ui-info-state`, so both apps share one implementation and DDD packages can reuse it.

The move itself is behaviour-neutral: no intended visual or functional change on either platform.

### 📂 What moved

| From | To |
| --- | --- |
| `apps/ledger-live-desktop/src/mvvm/components/InfoState/` | `shared/ui-info-state` (web variant) |
| `apps/ledger-live-mobile/src/mvvm/components/InfoState/` | `shared/ui-info-state` (native variant) |
| desktop's `DialogBackgroundContext`, `useDialogBackgroundTone`, `DialogBackgroundGradient` | `shared/ui-info-state/src/background/` |

Callers just write `import { InfoState } from "@shared/ui-info-state"`. The package holds both platform implementations and the right one is picked automatically — desktop gets the `.web` files, mobile the `.native` files — via the `react-native` export condition, the same pattern as `@shared/ui-qr-code` and `@shared/ui-queued-bottom-sheet`.

The dialog-background files came along because their only job is tinting the dialog behind an `InfoState`. Mobile's equivalent already lives in `@shared/ui-queued-bottom-sheet`, so this mirrors it and leaves the shared package with no dependency back on the desktop app.

All ~52 call sites in both apps now import from the package directly. No re-export shims were left behind.

### 🧩 What's in each commit

1. **`refactor(architecture)`** — the move itself, plus every call-site update.

2. **`fix(tailwind)`** — the desktop Tailwind config scanned `src`, `features` and `devtools`, but not `shared`. Moving web components into `shared/` therefore dropped three rules from the CSS bundle (`min-h-[480px]`, `h-[208px]`, `w-[208px]`) because no remaining scanned file used them. Confirmed by building the CSS: absent before the fix, present after. No production screen was affected — they all pass `size="hug"` — but `full-height` is the **default**, so the next caller to omit `size` would have silently lost its min-height, with nothing in typecheck or tests to catch it. The fix also covers `shared/ui-qr-code` and any future shared UI.

3. **`chore(codeowners)`** — with no entry, the new package fell to the `/shared/` catch-all (Platform). Its sibling shared UI package is owned by wallet-xp, and `InfoState` came out of the wallet apps, so reviews now route there.

4. **`build(tsconfig)`** and 5. **`refactor(ui-info-state)`** — the native TypeScript project was also compiling the web files, so it no longer proved what it exists to prove: that native code pulls in nothing web-only. Every web-only file now carries a `.web` suffix, which lets the existing exclude rule do the job. (4 was a narrower first attempt, superseded by the rename in 5.)

6. **`chore(ui-info-state)`** — config alignment with the other DDD packages. Dropped `tsconfig.test.json`, which no script, jest config or nx target referenced (it arrived by copying `shared/ui-qr-code`, where it is equally unreferenced), and wired up knip, which the package was missing while 9 of the 12 `shared/*` packages have it. A single-run knip script is not enough for a dual-platform package — knip cannot trace platform resolution and reports the other platform's files as unused — so this uses the repo's existing `createDualPlatformKnipConfig` and runs knip once per platform, as `features/flow/pay-card` does. Both passes exit clean.

7. **`test(ui-info-state)`** — review feedback: use `@support/jest-shared` rather than `@support/jest-features-flow`, which is scoped to `features/flow/*`. `jest-shared` landed on `develop` after this branch was cut and both sibling packages already moved to it, so this was the last `shared/ui-*` holdout. `createSharedUiJestConfig` covers the web+native split and ships the safe-area-context stub, so that local mock is gone. `Banner` still needs a package-specific stub — the shared native one renders `description` but not `title`, and the web one has no `Banner` case at all, while `InfoState` sets both and its tests assert both — but the two remaining mocks now delegate to `@support/jest-shared` via a Proxy and override only `Banner`, so they cannot drift from the shared implementation.

8. **`chore(deps)`** — regenerates `pnpm-lock.yaml` from a clean install. Against `develop` the delta is **147 insertions and 2 deletions**, all of it this package's own entries plus the two app importer links; no Expo, Firebase or react-native resolution is touched. The 2 deletions are optional peer-meta lines (`webpack`, `metro`) that pnpm normalises away.

> [!NOTE]
> **On the lockfile and `Podfile.lock`, since the history here is confusing:** an earlier revision of this branch carried a `pnpm-lock.yaml` generated from an incompletely-installed `node_modules`, which re-resolved unrelated Expo / React Native Firebase graphs. Those resolutions are what React Native autolinking feeds to CocoaPods, so the mobile `Test Pod Lockfile` check failed and the iOS build died before compiling any code from this branch.
>
> It is tempting to conclude from that failure that develop's `Podfile.lock` is stale. **It is not.** Installed from develop's own lockfile, `pod install` reproduces develop's committed `Podfile.lock` byte for byte. Earlier measurements suggesting otherwise were taken against a `node_modules` left over from this branch's install, which contaminated the result.
>
> The actual fix is commit 8 — regenerate `pnpm-lock.yaml` from a clean install. **`Podfile.lock` is untouched by this PR** and identical to develop.

https://github.com/user-attachments/assets/da84c2a4-bbc4-40da-8fb3-d449e6be3cda

https://github.com/user-attachments/assets/2e74dc22-5b98-4e43-a2c4-06a11c7c9ff5

### 🧪 Testing

- The `InfoState` tests from both apps moved into the package and all pass — **48 tests** across the web and native jest projects.
- Both apps typecheck clean, with no `InfoState`/`DialogBackground` errors.
- Platform selection verified explicitly: mobile resolves the native entry, desktop the web one.
- Both apps' own jest suites were run over the changed call sites:

  | Suite | Result |
  | --- | --- |
  | Desktop — `DeviceIntentExecutor`, `AccountPublicKeyUnavailableDialog`, `LargeScreenUpsell`, `PayTab/verifyAddressIntent` | 57/57 suites, 307/307 tests pass |
  | Mobile — `DeviceIntentExecutor`, `Send`, `Borrow`, `WalletApiSignature` | 100/101 suites, 640/655 tests pass |

  The single mobile exception is `Send/__integrations__/SendFlow.integration.test.tsx` (15 tests). It fails identically on `develop` with the same signature (empty render tree, "component suspended inside an `act` scope"), so it is a pre-existing failure of that suite outside CI and unrelated to this PR — nothing in its output references `InfoState`.

- Desktop, Android and iOS builds dispatched (below).

### 📦 Builds

> [!IMPORTANT]
> The branch has been rebased twice since these runs, so none of the commits below exist any more. **New builds are needed for the current head (`af94968ae44`).** The runs are kept because their outcomes are still informative about the flows involved, but they were produced from different SHAs and a different lockfile.

Dispatched on [`ledger-live-build`](https://github.com/LedgerHQ/ledger-live-build) for `refactor/architecture-LIVE-36304-infostate-to-shared` (pre-rebase commit `cd3c8f03d31`):

| Platform | Run |
| --- | --- |
| Desktop (macOS / Linux / Windows) | [run 32950279940](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32950279940) |
| Android (APK) | [run 32950282763](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32950282763) |
| iOS | [run 32950286513](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32950286513) |

Artifacts are attached to each run on completion.

> [!NOTE]
> **Expect the desktop run's `bundle-linux` job to fail — it is broken on `develop`, not here.** Every `build-desktop` run in the repo since 2026-08-24 has failed the same way, on other people's branches too, while `bundle-windows` and `bundle-macos` succeed on the identical commit. The Linux job dies after ~27–34 min with no failing step and no uploaded log, which is a killed runner rather than a build error. The macOS and Windows bundles are the meaningful signal for this PR.

<details>
<summary>Earlier rounds — superseded, kept for the record</summary>

**Round 2 (commit `ec59e24c00e`)** — the lockfile fix

| Platform | Run | Outcome |
| --- | --- | --- |
| Android | [32938736083](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32938736083) | ✅ passed |
| iOS | [32938669364](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32938669364) | ✅ passed — confirms the lockfile fix resolved the `Podfile.lock` failure |
| Desktop | [32938671974](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32938671974) | ❌ `bundle-linux` only (the repo-wide failure above); macOS + Windows ✅ |

**Round 1 (commit `273915bc13f`)**

| Platform | Run | Outcome |
| --- | --- | --- |
| Android | [32894509846](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32894509846) | ✅ passed |
| iOS | [32894513495](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32894513495) | ❌ `Podfile.lock` deployment check — a real bug in this PR, see the lockfile note above |
| Desktop | [32894506478](https://github.com/LedgerHQ/ledger-live-build/actions/runs/32894506478) | ❌ `bundle-linux` only; macOS + Windows ✅ |

</details>

Where to look when testing: most `InfoState` surfaces are device-flow states — DeviceIntentExecutor error / connection / unsupported-app screens, Send signature states, and on desktop the account-public-key and Recover dialogs; on mobile the Borrow and WalletApi signature sheets. Desktop also has a dedicated playground at **Settings → Developer → InfoState playground**, which is the one surface the Tailwind fix in commit 2 actually changes.

### 🔗 Context

- **JIRA issue**: [LIVE-36304](https://ledgerhq.atlassian.net/browse/LIVE-36304)
- **ADR** (if any): N/A

[LIVE-36304]: https://ledgerhq.atlassian.net/browse/LIVE-36304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared/ui-info-state/tsconfig.native.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "moduleSuffixes": [".native", ""]
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "src/**/*.web.*"]
+}
```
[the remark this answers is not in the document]
OlivierFreyssinet: Good catch, and fixed in `refactor(ui-info-state): suffix every web-only file with .web`. The whole `src/background/` folder (plus `internals/cn.ts` and the root web barrel) now carries a `.web` suffix, so the existing `src/**/*.web.*` exclude covers them and no path excludes are needed. `tsc -p tsconfig.native.json --listFiles` now reports only the native sources plus the platform-neutral `sharedTypes.ts`.

pnpm-lock.yaml
```diff
@@ -23926,8 +24008,6 @@ packages:
   '@react-native/babel-preset@0.81.6':
     resolution: {integrity: sha512-RtIr82ccPoyMLmIC3/vCG96MzRmIT+IzQkjCgTS5b93uAxiER9BS7+d1l7+zD00VanClt6CTBM4K4n6RCnAykg==}
     engines: {node: '>= 20.19.4'}
-    peerDependencies:
-      '@babel/core': '*'
```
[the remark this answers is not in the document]
OlivierFreyssinet: Correct, and this turned out to be a real bug rather than just review noise. The full `pnpm install` that produced that lockfile re-resolved the Expo and React Native Firebase peer sets and moved some resolutions from `@react-native-community/cli` 18.0.1 to 15.1.0. Different CLI version means different autolinking, so `pod install` computed new checksums for `EXConstants`, `Expo*` and `RNFB*` and the iOS build failed its `Podfile.lock` deployment check before compiling anything from this branch. Fixed in `fix(deps): limit the lockfile change to the new package`: regenerated from develop, so the diff is 97 lines instead of 660 and touches no existing Expo, Firebase or react-native resolution. iOS build is green again.

shared/ui-info-state/src/types.web.ts
```diff
@@ -0,0 +1,54 @@
+import type { ComponentProps, ReactNode } from "react";
+import type { Spot } from "@ledgerhq/lumen-ui-react";
+import type { DialogBackgroundTone } from "./background/DialogBackgroundContext";
+import type { InfoStateBaseProps } from "./sharedTypes";
+
+export * from "./sharedTypes";
+
+type LumenSpotProps = ComponentProps<typeof Spot>;
+type LumenIconSpotProps = Extract<LumenSpotProps, { icon?: unknown }>;
```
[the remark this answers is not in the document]
OlivierFreyssinet: This one is a false positive. A type-only import binding can be used in a `typeof` type query, so `import type { Spot }` plus `ComponentProps<typeof Spot>` is valid TypeScript. Verified empirically: `tsc --noEmit -p tsconfig.web.json` passes with zero errors, as does the desktop app typecheck. Resolving as no change needed.

apps/ledger-live-desktop/tailwind.config.ts
```diff
@@ -7,6 +7,7 @@ const config = {
   content: [
     "./src/**/*.{js,ts,jsx,tsx}",
     "../../features/**/src/**/*.{ts,tsx}",
+    "../../shared/**/src/**/*.{ts,tsx}",
     "../../devtools/**/src/**/*.{ts,tsx}",
   ],
```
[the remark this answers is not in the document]
OlivierFreyssinet: Keeping the glob broad, deliberately. Narrowing it to `ui-info-state` would leave the same latent bug in place for every other shared UI package: `shared/ui-qr-code` already uses Tailwind classes from outside the scanned paths and only gets away with it because the classes it happens to use are also used elsewhere in the app. This PR is exactly the case where that luck ran out — `min-h-[480px]`, `h-[208px]` and `w-[208px]` were used nowhere else, so they silently vanished from the CSS bundle. A per-package glob means the next shared UI package hits the same trap and nobody finds out until something looks subtly wrong in the app, with typecheck and tests both green.

On the cost: `shared/` is 12 small source-only packages, so the extra scanning is marginal next to the existing `./src/**` and `../../features/**/src/**` globs, which cover far more files.

Worth noting there is a third option if the scan cost ever does matter: Lumen self-declares its sources with `@source "./dist/lib"` in its own `tailwind.css`, so a shared package could ship its own stylesheet instead of relying on the app's globs. That is a bigger change than this tech enabler warrants, but it would be the cleaner long-term shape.

philipptpunkt (dismissed):

shared/ui-info-state/jest.config.js
Sebastien-Dav1d, line 1: Can you use this import instead ? https://github.com/LedgerHQ/ledger-live/tree/develop/support/jest-shared

Moustafa-Koterba (dismissed):

deepyjr (dismissed):

shared/ui-info-state/jest.config.js
OlivierFreyssinet: done @Sebastien-Dav1d

OlivierFreyssinet: Done, thanks — `@support/jest-shared` landed on develop after this branch was cut, so I had missed it. Rebased onto develop and switched to `createSharedUiJestConfig`, which fits this package directly (web + native, and my `*.web.test.tsx` / `*.native.test.tsx` naming already matches its `testMatch`). Its safe-area-context stub replaced my local one. 48/48 tests still green.

One thing I could not drop, and where I would value your opinion since `support/` is yours: `Banner`. The shared native stub renders `description` as text but not `title`, and `passthrough-web.js` has no `Banner` case at all, so both land as attributes. `InfoState` always sets both and its tests assert both are visible, so I still need a package-specific stub.

Rather than fork the stubs as I had before, the two remaining mocks now `require` the shared ones through `./mocks/*` and Proxy over just `Banner`, so they cannot drift from your implementation. The cost is that overriding `moduleNameMapper` replaces a project's defaults, so the config has to re-declare the shared mocks this package relies on — which is why it is not the clean two-liner `ui-qr-code` gets.

Happy to send a small PR adding `title` to the native `Banner` stub and a `Banner` case to the web one, if you think that is right for `jest-shared` generally. That would let this package drop to `createSharedUiJestConfig()` with no overrides and delete both local mocks. Your call — I did not want to change a STABLE package you own without asking.

tonykhaov (dismissed):

shared/ui-info-state/jest.config.js
Sebastien-Dav1d: @OlivierFreyssinet 
Thank you for the comment, I would gladly appreciate, that sounds like a good fix

Sebastien-Dav1d (dismissed):

fAnselmi-Ledger (dismissed):

deepyjr (dismissed):

fAnselmi-Ledger (dismissed):

philipptpunkt (dismissed):

shared/ui-info-state/jest.config.js
OlivierFreyssinet: lol @claude deciding on my behalf that I'm going to do a PR 😆 I do not remember approving this comment

tonykhaov (dismissed):

Moustafa-Koterba (dismissed):

Sebastien-Dav1d (dismissed):

philipptpunkt (dismissed):

fAnselmi-Ledger (dismissed):

Sebastien-Dav1d (dismissed):

Moustafa-Koterba (dismissed):

tonykhaov (dismissed):

PR #21153, opened 2026-08-26
[LWDM] feat(aleo): get validators list

mateuszpalosz-ext:
### 📝 Description

Adds the Aleo validator committee as a shared, cached data source so the upcoming bond flow has a validator picker to build on. Nothing is wired into the UI yet — this PR only exposes the logic and the hook.

**Coin module (`@ledgerhq/coin-aleo`)**

- Three new node endpoints in `network/api.ts`:
  - `GET /v2/{network}/committee/latest` — the committee (stake, open/closed, commission per validator)
  - `GET /v2/{network}/committee/validator-metadata` — address → display name (partial coverage)
  - `GET /v2/{network}/latest/totalSupply` — a bare scalar, and the only amount in this API denominated in **credits** rather than microcredits
- `logic/getValidators.ts` — fetches the three concurrently, validates each untrusted payload, and returns `AleoValidator[]` ordered for a picker (open validators first, then descending stake). LRU-cached for 5 minutes so a validator that has just closed is not offered for long.
- Only the committee is required. Names and total supply are best-effort: losing either degrades a field instead of failing the list the bond flow depends on.
- `logic/utils.ts` — `estimateGrossRate` / `estimateNetRate`, derived from the delegator's `block_reward * stake / total_stake` share in snarkVM (`synthesizer/src/vm/helpers/rewards.rs`). Handles the 25% validator concentration cap and the 10k-credit delegator minimum, both of which pay **zero**, not a reduced rate.

**live-common**

- `useAleoValidators(currency)` in `families/aleo/react.ts`, shared by Desktop and Mobile so the picker is not built twice.
- Keeps a last-seen committee per currency so a remount paints instantly instead of flashing an empty picker, and a failed refetch degrades to stale rather than to empty.
- Each mount gets its own array copy — `getValidators` is LRU-cached on the promise, so a picker sorting in place would otherwise reorder the list for every other screen.
- Deliberately not `CurrencyBridge.preload`/`hydrate`: that mechanism is deprecated here and Polkadot, the closest staking analogue, has already migrated off it.

**Note for consumers:** `estimatedYearlyRewardsRate` is a **lower bound** — snarkVM's `block_reward_v2` adds a coinbase share and transaction fees on top. Every surface showing it must label it an estimate. `undefined` means "could not be derived"; `0` is a real value meaning "earns nothing". The two are not interchangeable.

**Usage**

```ts
import { useAleoValidators } from "@ledgerhq/live-common/families/aleo/react";

const { validators, loading, error } = useAleoValidators(currency);
// validators: { address, name?, stakeMicrocredits, isOpen, commissionPercent, estimatedYearlyRewardsRate? }[]
```

**Tests**

37 new unit tests covering ordering, name resolution, concurrent fetching, and every degraded-response path (names fetch fails, malformed names payload, supply fetch fails, committee reports no total stake); rate maths checked against values observed on mainnet, plus the concentration cap, delegator minimum, boundary cases and the null-vs-zero distinction; and the hook's seeding, per-mount isolation, stale-on-failure and currency-switch behaviour.

### 🔗 Context

- **JIRA / GitHub issue**: [https://ledgerhq.atlassian.net/browse/LIVE-32275](https://ledgerhq.atlassian.net/browse/LIVE-32275)

### DIFF
--- a/libs/coin-modules/coin-aleo/src/constants.ts
+++ b/libs/coin-modules/coin-aleo/src/constants.ts
@@ -94,3 +94,15 @@ export const BALANCED_PRIVATE_RECORDS_PER_TRANSACTION = 8;
 
 // The estimated time in milliseconds it takes to sign a single record during transaction signing.
 export const SINGLE_CALL_SIGNING_TIME = 12500;
+
+export const MICROCREDITS_PER_CREDIT = 1_000_000;
+
+// Below this bonded total the protocol pays a delegator nothing at all.
+export const MIN_DELEGATOR_STAKE_MICROCREDITS = 10_000 * MICROCREDITS_PER_CREDIT;
+
+// snarkVM `block_reward_v2` adds a coinbase share and transaction fees on top, so
+// rates derived from this alone are a lower bound.
+export const ANNUAL_INFLATION_RATE = 0.05;
+
+// A validator above this share of total stake earns zero, not a reduced rate.
+export const MAX_VALIDATOR_STAKE_SHARE = 0.25;
--- a/.changeset/soft-mice-tickle.md
+++ b/.changeset/soft-mice-tickle.md
@@ -1,0 +1,6 @@
+---
+"@ledgerhq/coin-aleo": minor
+"@ledgerhq/live-common": minor
+---
+
+aleo: expose the validator committee with an estimated staking rate through a shared hook
--- a/libs/coin-modules/coin-aleo/src/logic/getValidators.test.ts
+++ b/libs/coin-modules/coin-aleo/src/logic/getValidators.test.ts
@@ -1,0 +1,139 @@
+import { apiClient } from "../network/api";
+import type { AleoCommitteeResponse } from "../types/api";
+import { getMockedConfig } from "../__tests__/fixtures/config.fixture";
+import coinConfig from "../config";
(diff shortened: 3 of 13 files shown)


libs/coin-modules/coin-aleo/src/constants.ts
```diff
@@ -94,3 +94,15 @@ export const BALANCED_PRIVATE_RECORDS_PER_TRANSACTION = 8;
...
+
+// Below this bonded total the protocol pays a delegator nothing at all.
+export const MIN_DELEGATOR_STAKE_MICROCREDITS = 10_000 * MICROCREDITS_PER_CREDIT;
+
+// snarkVM `block_reward_v2` adds a coinbase share and transaction fees on top, so
+// rates derived from this alone are a lower bound.
+export const ANNUAL_INFLATION_RATE = 0.05;
+
+// A validator above this share of total stake earns zero, not a reduced rate.
+export const MAX_VALIDATOR_STAKE_SHARE = 0.25;
```
henri-ly, line 108: Those variable aren't made to change a lot right ? how can we keep track of them ?

henri-ly (approved):

mdomanski-ext-ledger (approved):

PR #21178, opened 2026-08-26
feat(devtools): add knip platform-split config to devtools packages

Sebastien-Dav1d:
## Summary

Adds platform-aware knip configurations to four devtools packages (`feature-flags`, `pay-card`, `shell`, `transport-panel`), enabling dead-code analysis to run separately for web and native platforms.

### What changed

Each package gets two new knip config files and an `unimported` script:

- **`knip.native.ts`** — ignores `*.web.*` files, enables `.native.ts/.tsx` compilers, scopes the project glob to native sources only.
- **`knip.web.ts`** — ignores `*.native.*` files, enables `.web.ts/.tsx` compilers, scopes the project glob to web sources only.
- **`package.json`** — adds an `unimported` script that runs both configs sequentially:
  ```
  pnpm knip -c knip.web.ts ... && pnpm knip -c knip.native.ts ...
  ```

This mirrors the pattern established for the `knip` migration across the devtools scope and makes it possible to catch platform-specific dead exports without false positives from cross-platform file exclusions.

This solution does not fix everything, we have a different warnings that we can avoid if we avoid barrels files or only add to barrels files that are used on both web and native. (For example, we have index.ts that reexport hooks, one hook is used by web but not native. Web will see that the export is used but no native files import the export hence warn it.) This is not mandatory so not harmful but create some noise. But at least, knip doesn't completely confuse web and native files

I'll also point that we can't make a single knip file, as good as it sound, if a file have both a web and native version, one of them will be erased from knip hence flagging it unused.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36502](https://ledgerhq.atlassian.net/browse/LIVE-36502)
- **ADR** (if any): 

[LIVE-36502]: https://ledgerhq.atlassian.net/browse/LIVE-36502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/lazy-tiger-jump.md
+++ b/.changeset/lazy-tiger-jump.md
@@ -1,0 +1,8 @@
+---
+"@devtools/shell": patch
+"@devtools/feature-flags": patch
+"@devtools/pay-card": patch
+"@devtools/transport-panel": patch
+---
+
+Add platform-split knip configs for dead-code detection on web and native targets
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -71,10 +71,7 @@ libs/ledger-auth/                                        @ledgerhq/platform
 libs/oxc-live-libs/                                      @ledgerhq/platform
 
 # Platform — devtools
-devtools/transport/                                      @ledgerhq/platform
-devtools/protocols/                                      @ledgerhq/platform
-devtools/wire/                                           @ledgerhq/platform
-devtools/relay/                                          @ledgerhq/platform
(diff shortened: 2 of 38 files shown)


gre-ledger:
thanks for the work. i'll close this one and we'll tackle from other PR

PR #21212, opened 2026-08-27
[LWDM] fix(coin-solana): match parsed txs by signature in listOperations

YazhuEth:
### 📝 Description

`listOperations` paired the signature list returned by `getSignaturesForAddress` with the parsed transaction batch **by array position** (`parsed[i]` for `signatures[i]`). `ChainAPI.getParsedTransactions` delegates to web3.js `Connection.getParsedTransactions`, which issues a JSON-RPC **batch** and maps the response positionally. The JSON-RPC spec does not guarantee batch response ordering (responses must be matched by `id`), so a node or proxy that reorders replies made `parsed[i]` belong to another signature.

**Previous behaviour**: an operation kept the correct hash, slot and time (read from the signature) but took its fee, fee payer, balances and instructions from a different transaction. On the account of the ticket, 27 of the 30 native operations returned by coin-service carried another transaction's fee and fee payer, following an exact permutation of the page:

| coin-service says | on chain | coin-service returned |
| --- | --- | --- |
| `2wV7vcvemebj…` (swap 2 USDC to SOL) | native `+25674506`, USDC `-2000000` | `DELEGATE 0`, no token operation |
| `22jJe2VEX4Xw…` (transfer, account is fee payer) | delta `-5000`, fee only | `IN 1672400000` with a foreign `feesPayer` |
| `5DgWmNqJe8zv…` | delta `-20141991` | `IN 1` |

The legacy bridge path is unaffected because it re-matches transactions by signature (`network/chain/web3.ts`), which is why Ledger Live showed the correct history for the same transactions.

**Fix**: parsed transactions are indexed by their own `transaction.signatures` and looked up per signature. `hash`, `slot`, `blockTime` and `failed` still come from the signature info, which was already correct, and the descending order, `minHeight` filtering and cursor logic are unchanged. A signature whose transaction is absent from the batch is skipped instead of adopting a neighbour's data.

**Regression checks**: three new tests, each failing on the previous implementation.

- `listOperations.unit.test.ts`: batch returned in reverse order, and batch missing one entry.
- `listOperations.test.ts`: MSW test answering the JSON-RPC `getTransaction` batch in reverse order, so the guarantee is verified through the real web3.js batch path.

Beyond the ticket, flagged in review: `next` no longer depends on `items.length`, so a full page yielding no operation keeps the cursor instead of ending pagination and hiding older history.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-35047](https://ledgerhq.atlassian.net/browse/LIVE-35047)
- **ADR** (if any):

[LIVE-35047]: https://ledgerhq.atlassian.net/browse/LIVE-35047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/solana-list-operations-signature-pairing.md
+++ b/.changeset/solana-list-operations-signature-pairing.md
@@ -1,0 +1,9 @@
+---
+"@ledgerhq/coin-solana": patch
+---
+
+fix(coin-solana): pair each signature with its own parsed transaction in listOperations
+
+JSON-RPC batch responses are not order-guaranteed, so pairing the signature list with the parsed
+transaction batch by array position could attach another transaction's fee, feesPayer and balances
+to a signature. Transactions are now matched by their own signature.
--- a/libs/coin-modules/coin-solana/src/logic/listOperations.ts
+++ b/libs/coin-modules/coin-solana/src/logic/listOperations.ts
@@ -44,11 +44,11 @@ export async function listOperations(
 
   const sigStrings = signatures.map(s => s.signature);
   const parsed = await api.getParsedTransactions(sigStrings);
+  const txBySignature = indexTransactionsBySignature(parsed);
 
   const items: Operation[] = [];
-  for (let i = 0; i < signatures.length; i++) {
-    const sig = signatures[i];
-    const tx = parsed[i];
+  for (const sig of signatures) {
+    const tx = txBySignature.get(sig.signature);
     if (!tx?.meta || sig.blockTime === null || sig.blockTime === undefined) continue;
 
     if (minHeight > 0 && sig.slot < minHeight) continue;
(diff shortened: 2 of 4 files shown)


YazhuEth:
**E2E**
- D : https://github.com/LedgerHQ/ledger-live/actions/runs/33056705467
- M : https://github.com/LedgerHQ/ledger-live/actions/runs/33056723409

YazhuEth:
**E2E**
d: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/a5a3f2b9-dd5d-4cac-89a0-ea6b35b7df30/
m: https://ledger-live.allure.green.ledgerlabs.net/allure/reports/df5e8bce-4db2-49d4-81cb-8b32d3cf83c2/

henri-ly (approved):

PR #21220, opened 2026-08-27
[LWDM] refactor(contacts): move device intent UI layer to apps (LIVE-36547)

OlivierFreyssinet:
### 📝 Description

Components living in DDD packages can't reach what the apps own. i18n above all — there is no `i18next` anywhere under `features/`, `shared/` or `domain/` — but the same goes for navigation, analytics and app-level UI. A Device Intent renderer needs that, and the executor owns its component contract, so there is no clean way to feed it from the outside today.

The easy solution for now: keep the intent contracts and jobs in `@features/platform-contacts`, and let each app own the device intent UI layer and inject it into the orchestrator.

This is wiring only. The renderers are the same placeholders as before, so there is no user-visible change. The DIE README section recommending the fully-colocated DDD layout is marked not ready, since it doesn't hold for any renderer with translated copy.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36547](https://ledgerhq.atlassian.net/browse/LIVE-36547) (epic [LIVE-35697](https://ledgerhq.atlassian.net/browse/LIVE-35697))
- **ADR** (if any): [ADR: Device Intent Executor component](https://ledgerhq.atlassian.net/wiki/spaces/WXP/pages/6852083917)

[LIVE-36547]: https://ledgerhq.atlassian.net/browse/LIVE-36547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LIVE-35697]: https://ledgerhq.atlassian.net/browse/LIVE-35697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deepyjr (commented):
One question, do you think we need to keep LWM and LWD in the name of the file/props, i think it could be cleaner to have this without the platform ? Otherwise it's good for me 🙏

deepyjr (approved):

gre-ledger (approved):

fAnselmi-Ledger (approved):

PR #21223, opened 2026-08-27
[LWDM] feat(coin-casper): activate generic adapter (LIVE-35915)

vtaranushenko-ext-ledger:
### 📝 Description

Enable Casper on the generic coin-framework bridge: flag flip and integration tests.

### 🔗 Context

https://ledgerhq.atlassian.net/browse/LIVE-35915


libs/ledger-live-common/src/bridge/generic-coin-framework/genericCoinFrameworkFamilies.json
```diff
@@ -5,5 +5,6 @@
   "stellar": true,
   "tezos": true,
   "tron": true,
+  "casper": true,
```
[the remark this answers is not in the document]
vtaranushenko-ext-ledger: - When a user enters a transfer ID in the app, the code saves it in two ways at once, the old way (transferId) and the new way (memoType/memoValue). The new bridge only reads the new way, but since both are always saved together, nothing gets lost. 
- The only scenario where it could break: someone has a very old unfinished transaction saved from before the new fields were added. That transaction would only have the old field, and the new bridge would miss it. But in practice, no one has pending Casper transactions that old.

libs/ledger-live-common/src/bridge/impl.ts
```diff
@@ -42,6 +42,17 @@ import {
...
 // Rejections stay cached: evicting would hand React.use() a fresh Promise per render and re-suspend forever.
 // Callers that want to retry a transient failure must invalidate via clearBridgeCache(family).
+
+// CASPER_GENERIC_BRIDGE=false falls back to the legacy bridge for incident recovery.
+// Casper shipped on the old bridge before LIVE-35912; seedIdentifier format changed
+// (raw pubkey → tagged address). sameAccountIdentity's freshAddress fallback handles
+// re-scans, but id-keyed settings (account name, etc.) reset on the first rescan.
+function shouldUseGenericCoinFrameworkBridge(family: string) {
+  return (
+    isGenericCoinFrameworkFamily(family) && (family !== "casper" || getEnv("CASPER_GENERIC_BRIDGE"))
```
henri-ly: Is it possible to modify this by firebase ?

vtaranushenko-ext-ledger: You're right, getEnv doesn't read Firebase, so the env var approach wouldn't work as a kill-switch. Changed to LiveConfig (config_casper_generic_bridge), which does.

henri-ly (approved):

PR #21239, opened 2026-08-27
[LWDM] feat(concordium): add PLT CBOR encoding (LIVE-28330)

lysyi3m:
### 📝 Description

Adds Protocol-Level Token (PLT) support to `@ledgerhq/concordium-core`: `TransactionType.TokenUpdate` (`27` / `0x1B`), a CIS-7 CBOR encoder, and flat wire serialization. Prerequisite for LIVE-28335 and LIVE-28337. CCD behaviour is unchanged.

**Wire format**

```
[header:60][0x1B][token_id_length:1][token_id:N][cbor_length:4 BE][cbor:M]
```

These are the bytes the device hashes and the chain accepts, so one buffer both signs and broadcasts. The kind byte at offset 60 means the existing `TYPE_OFFSET` dispatch extends with one enum member. The token id is a sibling of the CBOR blob, not a field inside it.

**CBOR contract**

| Field | Encoding |
| --- | --- |
| operations | `array(1) [ map(1) { "transfer": map { … } } ]` |
| `amount` | tag 4 `[exponent, significand]`, exponent `<= 0`, unsigned 64-bit significand |
| `recipient` | tag 40307 wrapping `{? 1: 40305({1: 919}), 3: bstr(32)}` |
| `memo` | bare byte string, or tag 24; bytes, not text |

Keys are emitted in the chain's deterministic order, making the output byte-identical to `encodeTokenTransfer` in `concordium-base`.

**Validated before signing**

Each bound mirrors a device or chain rejection, so it surfaces locally instead of as a status word after the user has been prompted:

| Bound | Value | Otherwise |
| --- | --- | --- |
| operations | exactly one | device `0x6B10` |
| operations blob | 1..512 | device `0x6B0E` |
| token id | 1..128 | device `0x6B0F` |
| amount decimals | 0..128 | device `0x6B0D` |
| memo | <= 256 | chain rejects |

Two are tighter than the chain's: the device holds the exponent in an `int8_t`, capping decimals at 128 rather than 255; and a 400-byte memo fits the CBOR budget but fails on chain at `maxMemoSize = 256`.

**Codec.** `cbor.ts` handled text strings only. It gains generic primitives and `encodeCborMapDeterministic`, with no new dependency. Also corrects `MAX_CBOR_SIZE = 256`, documented as the device limit when it is the CCD memo limit; `PLT_CBOR_MAX_SIZE = 512` added alongside.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-28330](https://ledgerhq.atlassian.net/browse/LIVE-28330)
- **ADR** (if any): n/a

<details>
<summary>Reviewer notes</summary>

**PLT has its own transaction type rather than joining `TransactionPayload`.** Widening that union broke narrowing at 16 sites, 14 in existing CCD tests — `Transaction.type` is the wide enum, so it never narrowed the payload and those tests only compiled because the members shared a shape. `TokenUpdateTransaction` is separate, with `AnyTransaction` for the dispatchers and an `isTokenUpdateTransaction` guard that checks payload shape, not just the type byte.

**`deserializeTransaction` stays CCD-only.** Routing PLT through it forced a widened return type that re-broke those tests. Production never decodes a PLT payload — history comes from the wallet proxy as JSON — so it throws a message naming `deserializeTokenUpdate`. `getTransactionType` does accept `27`, since the signer's factory dispatches on that byte.

**Not enforced here:** the chain requires the exponent to equal the token's registered decimals. CAL is the source of truth and lives in the coin module, so that check belongs to LIVE-28332 / LIVE-28337.
</details>

[LIVE-28330]: https://ledgerhq.atlassian.net/browse/LIVE-28330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/concordium-core/src/serialization.ts
+++ b/libs/concordium-core/src/serialization.ts
@@ -1,6 +1,10 @@
 import {
   TransactionType,
+  type AnyTransaction,
   type Transaction,
+  type TokenUpdatePayload,
+  type TokenUpdateTransaction,
+  type TransactionHeader,
   type CredentialDeploymentTransaction,
   type IdOwnershipProofs,
 } from "./types";
@@ -18,7 +22,12 @@ import {
   serializeYearMonth,
 } from "./utils";
 import { AccountAddress } from "./address";
-import { MAX_CBOR_SIZE } from "./cbor";
+import {
+  MAX_CBOR_SIZE,
+  PLT_CBOR_MAX_SIZE,
+  PLT_TOKEN_ID_MAX_LENGTH,
+  PLT_TOKEN_ID_MIN_LENGTH,
+} from "./cbor";
 
 /**
  * Serializes transaction header common to all account transaction types.
@@ -26,7 +35,7 @@ import { MAX_CBOR_SIZE } from "./cbor";
  * Format: [sender:32][nonce:8][energyAmount:8][payloadSize:4][expiry:8]
  * @private
  */
-function serializeTransactionHeader(tx: Transaction, payload: Buffer): Buffer {
+function serializeTransactionHeader(tx: { header: TransactionHeader }, payload: Buffer): Buffer {
   const serializedSender = tx.header.sender.toBuffer();
   const serializedNonce = encodeWord64(tx.header.nonce);
   const serializedEnergyAmount = encodeWord64(tx.header.energyAmount);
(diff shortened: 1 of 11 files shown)


hedi-edelbloute (approved):

libs/concordium-core/src/serialization.ts
```diff
@@ -106,6 +115,197 @@ export const serializeTransferWithMemo = (tx: Transaction): Buffer => {
...
+    throw new Error("Transaction must be TokenUpdate type");
+  }
+
+  if (!hasTokenUpdatePayload(tx.payload)) {
+    throw new Error("TokenUpdate payload must contain tokenId and operations");
+  }
+
+  const { tokenId, operations } = tx.payload;
+
+  if (tokenId.length < PLT_TOKEN_ID_MIN_LENGTH || tokenId.length > PLT_TOKEN_ID_MAX_LENGTH) {
```
amaslakov, line 193: Does it make sense to also do a charset check here ([a-zA-Z0-9.\-%])? The node's TokenId::deserial does String::from_utf8(buf)? then TryFrom<String>, which rejects anything outside that set — so a token id like PLT_USD passes this length bound, gets signed, and only bounces at broadcast

ishaba (approved):

libs/concordium-core/src/serialization.ts
```diff
@@ -499,14 +711,21 @@ export function deserializeTransaction(buffer: Buffer): Transaction {
...
+
   switch (transaction.type) {
     case TransactionType.Transfer:
       return serializeTransfer(transaction);
     case TransactionType.TransferWithMemo:
       return serializeTransferWithMemo(transaction);
-    /* istanbul ignore next */
+    // The narrowed discriminator makes this unreachable for a well-typed caller,
+    // but it still catches a malformed object from JS or a cast.
     default:
```
amaslakov, line 726: In case `type === 27` but the payload is malformed, `isTokenUpdateTransaction` returns `false`, so we fall through to here and report "Unsupported transaction type for serialization: 27" — but 27 is supported, it's the payload that's wrong. Tightening hasTokenUpdatePayload moved the non-actionable message rather than removing it. Could we maybe branch on TransactionType.TokenUpdate before default and reuse "TokenUpdate payload must contain tokenId and operations"?

amaslakov (approved):

PR #21249, opened 2026-08-27
[LWDM] feat(observability): attribute sign-stage failures to their live app

beths-ledger:
### 📝 Description

**Fixes a defect that is live on `develop` today, and unblocks [LIVE-35351](https://ledgerhq.atlassian.net/browse/LIVE-35351) (ETH dApp events).**

**Problem.** `manifestId` is read from `broadcastConfig.source`, which does not exist at the sign stage. The Earn live-app skip in `toSegmentTrackEvent` keys on the manifest, so it could never fire there.

A user who starts native staking inside the Earn live app and rejects on the device produces **two** `earn_transaction_failed` events — one from the Earn app, one from this seam. Successes were never affected, because success is reported at broadcast where the skip works. Until this lands, a funnel query must filter `stage == "broadcast"` to count each transaction once.

It also blocks the dApp work: that gates on the manifest and derives the action from call data, and the action is only legible at sign.

**Approach.** The plumbing already existed. [`withLiveAppContext`](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledger-live-common/src/wallet-api/blindSigningContext.ts) scopes a manifest id around a signing call, and its own docblock says to use it around `bridge.signOperation`. It already wraps all three routes:

| Route | Call site |
|---|---|
| dApp `eth_sendTransaction` | `wallet-api/useDappLogic.ts:522` |
| live-app `transaction.sign` | `wallet-api/logic.ts:99` |
| live-app raw sign | `wallet-api/logic.ts:161` |

So the seam reads it. About five lines, no API change, no call-site changes.

**Why read a context instead of passing an argument.** An explicit `manifestId` on `SignOperationArg0` looks cleaner, and at `hw/actions/transaction.ts:222` the value is already in scope. But it would not cover every route: desktop threads the manifest through, mobile's device-intent path passes it, and **mobile's legacy path drops it** at `Web3AppWebview/helpers.ts:675` when it navigates. The context covers all three, because the wrapper holds scope until `onSuccess` fires — across the whole signing, navigation included.

`getAccountBridge` cannot import from `wallet-api` (that direction already exists and would cycle), so the reader lives in `impl.ts` and reads `live-dmk-shared` directly. That package has no live-common dependency.

**What did not change.** The route *type* still waits for broadcast — the context carries an id, not a source — so the sign stage keeps `tx_pathway: "unknown"`. The manifest is deliberately **not** put into the correlation `WeakMap`: broadcast has its own and that one is authoritative. Native in-app staking is untouched.

**Known weakness, stated on purpose.** The store is a mutable singleton set and restored around an `await`, not an `AsyncLocalStorage`. Two overlapping signatures would misattribute the second. Device signing serialises — one device, one prompt, and the live app sits behind the signing screen — so it holds today. Mitigations in this PR: a comment that says the assumption out loud, a note at the setter that it now has two consumers, and a test pinning the restore. [LIVE-36571](https://ledgerhq.atlassian.net/browse/LIVE-36571) removes the dependency by passing the source explicitly.

**Review notes**

- The four new tests fail without the one-line change (verified: 4 failed / 12 passed with it removed), so they prove the behaviour rather than restate it.
- The nested-context test is the one guarding the weakness above — it is not incidental coverage.
- Local `pnpm typecheck` reports 16 errors in `families/aleo`, `families/solana`, `families/hedera`, `families/evm` and `mock/account.ts`. These are stale sibling builds in a local `node_modules`, not this change: e.g. `listSolanaStakingPositions` exists in `coin-solana/src` but not in its built `lib/`. No file in this diff appears.

**Test coverage:** yes, 16 seam tests pass. No UI changes.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36569](https://ledgerhq.atlassian.net/browse/LIVE-36569)
- Follows the merged stack: #20817 → #20818 → #20819
- Blocks [LIVE-35351](https://ledgerhq.atlassian.net/browse/LIVE-35351)
- Follow-up: [LIVE-36571](https://ledgerhq.atlassian.net/browse/LIVE-36571)

[LIVE-35351]: https://ledgerhq.atlassian.net/browse/LIVE-35351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-36571]: https://ledgerhq.atlassian.net/browse/LIVE-36571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/LIVE-36569-sign-stage-manifest-id.md
+++ b/.changeset/LIVE-36569-sign-stage-manifest-id.md
@@ -1,0 +1,13 @@
+---
+"@ledgerhq/live-common": patch
+---
+
+Report the originating live-app or dApp on sign-stage `earn_transaction_failed` events.
+
+`manifestId` came only from `broadcastConfig.source`, which does not exist at the sign stage. So the Earn live-app skip in `toSegmentTrackEvent` — which keys on the manifest — could never fire there, and every device rejection inside the Earn app was counted twice: once by the Earn app, once by the seam. Successes were unaffected, because success is reported at broadcast where the skip works.
+
+`withLiveAppContext` already scopes the manifest id around every wallet-api and dApp signing call, so the seam reads it instead of changing the bridge signature. That choice is deliberate rather than lazy: mobile's legacy wallet-api path never forwards the manifest to the device action, so an argument would have missed that route entirely.
+
+The route *type* still waits for broadcast — the context carries an id, not a source — so the sign stage keeps `tx_pathway: "unknown"`.
+
+The context is a singleton restored around an `await`, not an `AsyncLocalStorage`, so two overlapping signatures would misattribute the second. Device signing serialises today, one device and one prompt, and a test pins the restore behaviour. LIVE-36571 removes the dependency by passing the source explicitly.
--- a/libs/ledger-live-common/src/bridge/impl.ts
+++ b/libs/ledger-live-common/src/bridge/impl.ts
@@ -27,6 +27,7 @@ import {
 } from "../coin-modules/registry";
 import { defaultBridgeExtensions } from "./defaultBridgeExtensions";
 import { isZcashShieldedEnabled } from "./zcashRouting";
+import { liveBlindSigningReporter } from "@ledgerhq/live-dmk-shared";
 import { throwError } from "rxjs";
 import { catchError, tap } from "rxjs/operators";
 import {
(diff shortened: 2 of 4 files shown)


beths-ledger:
## 🧪 Triggered test workflows

Manually dispatched on `feat/LIVE-36569-sign-stage-manifest-id` (commit 7249fa2564e), scoped to `@smoke`:

| Workflow | Scope | Run |
|---|---|---|
| [Desktop] E2E Only | `@smoke` filter, Speculos nanoSP | [run 33107586773](https://github.com/LedgerHQ/ledger-live/actions/runs/33107586773) |
| [Mobile] E2E Only | `@smoke` filter, iOS & Android, nanoX | [run 33107589516](https://github.com/LedgerHQ/ledger-live/actions/runs/33107589516) |

Scope rationale: this PR adds a read of the live-app signing context inside the account-bridge `signOperation` wrapper, so the risk surface is the shared signing path rather than any one coin. `@smoke` exercises send and delegate end to end on both surfaces, which is what would break if the wrapper mishandled an error or changed subscription behaviour. Coin Tester is skipped: no coin-module logic changed.

dilaouid (approved):

PR #21264, opened 2026-08-28
[LWDM] feat(coin-concordium): type the PLT wallet-proxy network layer

lysyi3m:
### 📝 Description

Types every PLT-carrying wallet-proxy response in `coin-concordium`, and adds the two token-config endpoints.

`accountTokens` was typed `unknown[]` and had no consumer. It now carries a real shape, taken from the node's own JSON instances (`concordium-base`: `Concordium.Types.Queries.Tokens`, `ProtocolLevelTokens/CBOR.hs`, `Concordium.Types.Tokens`) rather than from the proxy README.

**What is here**

- PLT response types in `types/network.ts`, and `accountTokens: PltAccountToken[]`.
- `getPltTokens` (`GET /v0/plt/tokens`) and `getPltTokenInfo` (`GET /v0/plt/tokenInfo/{id}`) in `proxyClient.ts`.
- PLT fields on `WalletProxyTransactionDetails`, plus `SubmissionStatusTokenUpdateDetails`.
- `network/plt.ts`: `isDecodedPltState`, `getAccountListStatus`, `isPltRejectReason`.
- `listOperations` now sends `includeRawRejectReason`.

**Four contract details the types encode**

1. `moduleState` and `tokenAccountState.state` are CBOR on chain. The node renders each as an object when the CBOR decodes, and as a **hex string** when it does not. Both are `PltEncodedState<T> = T | string`, so no caller can read a field without narrowing first. Reading through the hex form yields `undefined`, which would pass as "flag not set".

2. `moduleState.allowList` means *the token has an allow list*. `tokenAccountState.state.allowList` means *this account is on it*. Same field name, different question. They are separate types, and `getAccountListStatus` applies the chain rule: the module flag decides whether the account flag is meaningful at all. It returns `allowed | blocked | unknown`, and `unknown` is not a soft `allowed` — it means the policy could not be read.

3. Every `moduleState` field is optional and omitted when unset. An absent `paused` is not `false`.

4. The transferred value has two names. History calls it `tokenTransferAmount`; `/v0/submissionStatus` calls it `tokenAmount` and reports no sender. Hence the separate detail type.

**`includeRawRejectReason`**

Without it a rejected transaction carries only `rejectReason`, localized prose that cannot serve as a mapping key. The proxy tests the parameter for **presence, not value**, so `false` would enable it — the client must omit it to disable. `listOperations` now sends it, which is what makes structured reject reasons available to the error mapping later.

**Notes for the reviewer**

- `src/network/plt.ts` is in `ignoreUnimported`. Its consumers arrive with the account model and the transaction status work; the entry graph does not reach it yet.
- The submission-status detail type has no client method. No `/v0/submissionStatus` call exists in this package today, and an uncalled request function would be dead code. The type is here so that whoever wires that endpoint does not reach for `tokenTransferAmount`.
- Shapes should still be confirmed against a live testnet PLT account. They are read from source, not observed on the wire.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-28331](https://ledgerhq.atlassian.net/browse/LIVE-28331)
- **ADR** (if any):

[LIVE-28331]: https://ledgerhq.atlassian.net/browse/LIVE-28331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/concordium-plt-network-layer.md
+++ b/.changeset/concordium-plt-network-layer.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/coin-concordium": minor
+---
+
+Type the PLT wallet-proxy responses: `accountTokens`, the `/v0/plt/tokens` and `/v0/plt/tokenInfo` clients, PLT transaction detail fields, and the raw reject reason
--- a/libs/coin-modules/coin-concordium/.unimportedrc.json
+++ b/libs/coin-modules/coin-concordium/.unimportedrc.json
@@ -20,6 +20,7 @@
   ],
   "ignoreUnimported": [
     "src/errors.ts",
+    "src/network/plt.ts",
     "src/test/concordiumTestUtils.ts",
     "src/test/fixtures.ts",
     "src/test/testHelpers.ts",
--- a/libs/coin-modules/coin-concordium/src/logic/history/listOperations.test.ts
+++ b/libs/coin-modules/coin-concordium/src/logic/history/listOperations.test.ts
@@ -217,6 +217,7 @@ describe("listOperations", () => {
     expect(getTransactionsMock).toHaveBeenCalledWith(config, "concordium_testnet", VALID_ADDRESS, {
       limit: 100,
       order: "d",
+      includeRawRejectReason: true,
     });
     expect(result.items).toHaveLength(1);
     expect(result.items[0]).toMatchObject({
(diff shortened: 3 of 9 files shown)


ishaba (approved):

hedi-edelbloute (approved):
lgtm 👍

PR #21269, opened 2026-08-28
[LWDM] feat(live-dmk-desktop): add mock server transport (DSDK-1309)

aussedatlo:
### 📝 Description

Ledger Wallet Desktop could only talk to a physical device. This adds the Device Management Kit **mock server transport** behind a developer toggle, so the app can be driven against an emulated device — and, since an emulated device has no screen on your desk, renders that screen in the app and lets you press it.

**Transport toggle** — Settings › Developer › "Mock server transport". When enabled, the DMK registers `mockserverTransportFactory` alongside WebHID and the secure channel (genuine check, list apps, install, firmware/MCU) is pointed at the mock server's ScriptRunner socket, so the legacy `createDeviceSocket` flows resolve there too. A top bar indicator shows green while `/health` responds, red otherwise; clicking it copies the session token.

https://github.com/user-attachments/assets/56c6e0c7-ae26-4e80-b801-cd012ea43d2d

https://github.com/user-attachments/assets/af86dfbc-5905-40c2-bcd0-9394f0ae840a

**Configurable, with sensible defaults** — four env vars, all settable from Developer › Env variable override, so none of this needs a rebuild:

| Env | Default | Purpose |
| --- | --- | --- |
| `MOCK_SERVER_TRANSPORT` | `false` | Enables the transport |
| `MOCK_SERVER_TRANSPORT_URL` | shared deployment | Point at a local pod with `http://localhost:9752` |
| `MOCK_SERVER_SEED` | `""` | BIP39 mnemonic; empty keeps the mock server's own seed |
| `MOCK_SERVER_SESSION` | single USB Stax, fw 1.9.1 | `SessionExport` JSON imported at boot |

The URL default matches the Device SDK sample app rather than assuming a local pod. Invalid `MOCK_SERVER_SESSION` JSON falls back to the default rather than breaking boot.

**Live device screen** — docked at the foot of the sidebar while an emulated device is connected, collapsible to its header row (collapsing also stops the polling). The mock server proxies through to the Speculos instance backing the device, so a poller fetches PNG stills and sends input over the session token seeded at boot.

Press and release are separate calls, driven by pointer down/up, so a hold reaches the device as a hold — Stax and Flex gate their confirmations behind exactly that. Each frame's own dimensions size the panel and map taps to device pixels, so a Nano's 128×64 strip and a Stax's 400×672 portrait both fit with no per-model size table. A `ScreenApi` seam keeps *where* the requests go out of the panel, so a direct-Speculos source can be added later without the view changing.

### 📦 New dependencies

Two runtime dependencies, both first-party Ledger packages with no in-repo equivalent:

| Package | Version | Why |
| --- | --- | --- |
| `@ledgerhq/device-transport-kit-mockserver` | 1.1.1 | The only DMK transport that talks to a device mock server |
| `@ledgerhq/device-mockserver-client` | 1.3.0 | Its HTTP client; also carries the screen APIs (`getScreenshot`, `pressButton`, `touchScreen`) |

Both are released versions (no `develop` snapshot) and both peer-depend on `@ledgerhq/device-management-kit ^1.9.0`, which the catalog already pins at 1.9.0 — so no DMK bump and no peer warnings. The transport does not declare the client in its own `dependencies`, so a root `packageExtensions` entry adds it; that shim can go once it is fixed upstream.

Size impact, from Rsdoctor on this PR: **desktop-main +1.0 KB, desktop-renderer +21.1 KB (0.0%)**. Both are dev-only paths, reached only while `MOCK_SERVER_TRANSPORT` is on.

### 🔗 Context

- **JIRA / GitHub issue**: [DSDK-1309](https://ledgerhq.atlassian.net/browse/DSDK-1309)
- **ADR** (if any): —

### ✅ Verification

Unit tests cover the boot provisioning, the env parsing, the top bar indicator and the screen poller (object-URL lifecycle, idle fallback, error paths, input refresh), plus integration tests on the panel itself.

### ⚠️ Reviewer notes

- **Not yet validated against a live mock server pod.** The tap coordinate mapping and whether `MOCK_SERVER_SEED` actually reaches Speculos are unproven — the derived address is the only real proof of the latter.
- **The transport reaches shared infrastructure by default**, seeding a device there. That is a change in blast radius from a purely local loopback.
- `libs/ledger-live-common` is touched once, and only to keep a session token out of `createDeviceSocket` traces and error metadata — this PR is what puts one in a scriptrunner URL. No e2e specs are touched.

[DSDK-1309]: https://ledgerhq.atlassian.net/browse/DSDK-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/env/src/env.ts
```diff
@@ -162,6 +162,10 @@ export interface EnvTypes {
   EXPERIMENTAL_USB: boolean;
   MOCK_APP_UPDATE: boolean;
   FORCE_PROVIDER: number;
+  MOCK_SERVER_TRANSPORT: boolean;
+  MOCK_SERVER_TRANSPORT_URL: string;
+  MOCK_SERVER_SEED: string;
+  MOCK_SERVER_SESSION: string;
```
gre-ledger: ⚠️  my bad on this but unfortunately this list of types were not supposed to exist anymore in libs/env as @shared/env is the one injecting the definitions...

Let me figure out if we can get rid of these types here, but I think we actually should not touch this file anymore. (and so the PR shouldn't have a change & changeset related to live-env anymore, as we try to make it stable to extract it to ts-libs repo)

.changeset/dmk-mock-server-transport-toggle.md
```diff
@@ -0,0 +1,12 @@
+---
+"@ledgerhq/live-dmk-desktop": minor
+"@ledgerhq/live-env": minor
```
gre-ledger: after you rebase the PR, you will not need to modify live-env lib anymore (per previous note i had, we just merged into develop a cleanup that makes it stable and not necessary to touch it anymore)

could you remove that line? thanks


```suggestion
```

gre-ledger (changes requested):

gre-ledger (approved):
Ok for me. 
NB: we'll progressively reduce the usages of live-env in favor of other ways (documented in MIGRATION.md of the lib) but as this was worked before we started deprecate the lib, it's still acceptible to do it this way
to consider in future iterations

gre-ledger (dismissed):

fAnselmi-Ledger (dismissed):

apps/ledger-live-desktop/src/mvvm/components/DeviceScreen/__integrations__/DeviceScreen.integration.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    expect(pressButton).toHaveBeenCalledWith("left", "release");
+  });
+
+  it("collapses to its header, stops polling, and persists the choice", async () => {
+    render(<DeviceScreen />);
+
+    expect(mockUseDeviceScreen).toHaveBeenLastCalledWith(true);
+
+    act(() => {
+      fireEvent.click(screen.getByTestId("device-screen-toggle"));
```
LucasWerey: Please use `await user.click`

apps/ledger-live-desktop/src/mvvm/components/DeviceScreen/components/DeviceScreenImage/useDeviceScreenImageViewModel.ts
```diff
@@ -0,0 +1,147 @@
...
+ * A still frame of the device screen. The PNG's own dimensions drive the aspect
+ * ratio and the touch mapping, so every model is handled without a per-model
+ * size table.
+ *
+ * Touches are sent as a press on pointer down and a release on pointer up, so
+ * holding the mouse holds the finger — Stax and Flex gate their confirmations
+ * behind exactly that.
+ */
+export function DeviceScreenImage({ src, onTouch }: DeviceScreenImageProps) {
+  const imageRef = useRef<HTMLImageElement>(null);
```
LucasWerey, line 30: There is a lot of logic here. Might be worth to have a dedicated viewModel hook file wdyt?

LucasWerey (commented):
Can you set Device team as owner of this please 🙏

fAnselmi-Ledger (approved):

gre-ledger (approved):

LucasWerey (approved):

PR #21292, opened 2026-08-28
test(e2e-mobile): fix borrow open-loan nightly failures

VicAlbr:
## Context

The Android broadcast nightly ([run 33137517669](https://github.com/LedgerHQ/ledger-live/actions/runs/33137517669), [Slack thread](https://ledger.slack.com/archives/C05FKJ7DFAP/p1787903486181589)) fails on `should open a loan with wBTC collateral through approval, deposit and borrow`. Borrow runs only on the broadcast lane — Wed for iOS, Fri for Android — so this is the Friday Android leg.

The `android-ai-analysis` artifact shows two distinct defects across the three retries.

### 1. Stale collateral allowance

`Ethereum 4` is shared, and the partner approves an **exact** amount (`0.0002 wBTC` = 20000 units), not unlimited. A run that approves and then fails before supplying leaves a residue behind, and `resetLoanState` only closes positions — it never clears an allowance.

The execution flow then treats step 1 as already approved, so `authorizeStep`'s assertion that the step-done marker is absent throws:

```
'<[]>' doesn't match: [{"ELEMENT":…}]   at borrow.page.ts:155
```

Verified on-chain — this was sitting on the account while investigating:

```
allowance(ETH_4 → 0x3D12e577…) = 2541
```

Confirmed by reproduction: clearing it made the test pass, and leaving it made the **deposit** step fail too (`borrow-step-2-deposit-done` timing out at 240s), since the remaining allowance is below the supply amount.

The before-hook now resets the allowance so every run starts needing approval, keeping the approval path genuinely covered rather than skipped.

**The spender is resolved, not pinned.** It is a partner adapter (`0x3D12e577…`), not Morpho Blue, and the partner can redeploy it. A hardcoded address would revoke nothing while still reading back `"0"` — the assertion cannot catch that — so `resolveCollateralSpender` reads it from the partner's own approve step. Building an action does not broadcast.

### 2. `ReferenceError: device is not defined`

Hit on the one retry that got past the first defect:

```
ReferenceError: device is not defined   at borrow.page.ts:176
```

`detox.config.js:83` sets `behavior.init.exposeGlobals: false`, so `device` is undefined at runtime in every file — every other file imports it from `detox`. It typechecks because `detox/globals.d.ts` declares the globals ambiently, so only a real run catches it.

Fixed by using the existing `app.common.disableSynchronizationForiOS()` / `enableSynchronization()` helpers, as the equally webview-heavy swap specs do. These are deliberately iOS-only, so Android keeps synchronization on and a real ANR still fails the test — the previous code disabled it on both platforms.

## Verification

Local `android.emu.release`, broadcast enabled:

```
✓ should open a loan with wBTC collateral through approval, deposit and borrow  (201s)
✓ should show the Introducing Crypto Loan modal from the portfolio entry point   (30s)
Tests: 2 passed, 2 total
```

### CI

Mobile E2E dispatched on this branch, scoped to borrow only, with broadcast **on** — the flow is real
mainnet transactions and the describe is skipped without it:

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | `test_filter=specs/borrow/`, `tests_type=Android Only`, `speculos_device=nanoX`, `enable_broadcast=true` | [run 33180224945](https://github.com/LedgerHQ/ledger-live/actions/runs/33180224945) |

The filter resolves to a single shard and iOS is skipped, so only the borrow specs execute.

`Android Only` is deliberate: `iOS & Android` sets `E2E_BROADCAST_BOTH_MOBILE_PLATFORMS=true`, and the
rotation in `shouldRunBroadcastFlow` would then pin borrow to one platform and skip the other. The
filter is `specs/borrow/` rather than the single spec path so `borrow.coldStart.spec.ts` runs too.

## Notes for review

- `Currency.ETH_WBTC` and `TokenAccount.ETH_WBTC_4` are additive; no existing entry changes. `revokeTokenCommand` needs a `TokenAccount`, and index `3` matches `Account.ETH_4`.
- `resetCollateralAllowance` mirrors swap's `revokeTokenApproval` (`e2e/mobile/utils/swapUtils.ts`), including the Speculos save/restore. It runs after `app.init` because it needs the address `liveDataWithAddressCommand` resolves.
- Desktop is unaffected: it already handles the already-approved case with a conditional (`completeGiveApprovalIfRequired`).

### DIFF
--- a/e2e/mobile/utils/borrowUtils.ts
+++ b/e2e/mobile/utils/borrowUtils.ts
@@ -1,0 +1,41 @@
+import { TokenAccount } from "@ledgerhq/live-e2e-shared/enum/Account";
+import { resolveCollateralSpender } from "@ledgerhq/live-e2e-shared/borrow/borrowApi";
+import { DEFAULT_COLLATERAL } from "@ledgerhq/live-e2e-shared/borrow/borrowSetup";
+import { allure } from "jest-allure2-reporter/api";
+import { getEnv } from "@shared/env";
+import { deleteSpeculos, launchSpeculos, registerSpeculos } from "@e2e/utils/speculosUtils";
+
+/**
+ * Leaves the collateral allowance at zero so the "Give approval" step is always required.
+ * The partner approves an exact amount, so a run that approves and then fails before
+ * supplying leaves a residue behind, and `resetLoanState` only closes positions.
+ *
+ * Broadcasts; gated by the caller's `shouldRunBroadcastFlow`.
+ */
+export async function resetCollateralAllowance(collateralAccount: TokenAccount) {
+  const owner = collateralAccount.parentAccount ?? collateralAccount;
+  const ownerAddress = await getAccountAddress(owner);
+
+  const spender = await resolveCollateralSpender(ownerAddress, DEFAULT_COLLATERAL);
+  let allowance = await getTokenAllowanceCommand(collateralAccount, spender);
+  if (allowance !== "0") {
+    const previousSpeculosPort = getEnv("SPECULOS_API_PORT");
+    const speculos = await launchSpeculos(collateralAccount.currency.speculosApp.name);
+    await registerSpeculos(speculos.port);
+    try {
+      const result = await revokeTokenCommand(collateralAccount, spender);
+      await allure.description(`Collateral allowance revoke result:\n\n ${result}`);
+    } finally {
+      await deleteSpeculos(speculos.id);
+      if (previousSpeculosPort > 0) {
+        await registerSpeculos(previousSpeculosPort);
+      }
+    }
+    allowance = await getTokenAllowanceCommand(collateralAccount, spender);
+  }
+  if (allowance !== "0") {
+    throw new Error(
+      `Collateral allowance revoke did not settle for ${spender}: expected "0", got "${allowance}"`,
+    );
+  }
+}
--- a/libs/live-e2e-shared/src/borrow/borrowApi.ts
+++ b/libs/live-e2e-shared/src/borrow/borrowApi.ts
@@ -1,4 +1,11 @@
-import type { BorrowAction, OpenLoan, PartnerActionResponse, PartnerActionStep } from "./types";
+import { Interface } from "ethers";
+import type {
+  BorrowAction,
+  EvmSignablePayload,
+  OpenLoan,
+  PartnerActionResponse,
+  PartnerActionStep,
+} from "./types";
 
 /** Staging only — the staging Borrow API is keyless (reads and actions). */
 const BASE_URL = "https://global.api.stg.ledger-test.com/borrow";
@@ -16,6 +23,8 @@ export const ETHEREUM_CHAIN_ID = 1;
  */
 export const DEFAULT_MARKET_ID =
   "morpho-blue-borrow-ethereum-wbtc-usdt-0xa921ef34e2fc7a27ccc50ae7e4b154e16c9799d3387076c421423ef52ac4df99";
+
+const ERC20 = new Interface(["function approve(address spender, uint256 value)"]);
 
 function get(v: unknown, key: string): unknown {
   return v !== null && typeof v === "object" ? Reflect.get(v, key) : undefined;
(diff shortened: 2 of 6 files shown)


VicAlbr:
## 🧪 Triggered test workflows

Manually dispatched on `test/e2e-mobile-fix-borrow-open-loan` (commit e1fcb41eda7), scoped to the borrow specs:

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | `specs/borrow/` filter, Android Only, nanoX, broadcast enabled | [run 33180224945](https://github.com/LedgerHQ/ledger-live/actions/runs/33180224945) |

Scope rationale: the change touches the mobile borrow spec/POM plus two additive entries in the shared account enums, so only the mobile borrow specs are affected. Desktop is untouched — it already handles the already-approved case via `completeGiveApprovalIfRequired`, and no existing enum entry changed. Coin Tester is not relevant (no coin-module change).

`Android Only` is deliberate: `iOS & Android` would set `E2E_BROADCAST_BOTH_MOBILE_PLATFORMS=true`, and the rotation in `shouldRunBroadcastFlow` would then pin borrow to one platform and skip the other. `enable_broadcast=true` is required — the borrow flow is real mainnet transactions and the describe is skipped without it.

The filter is `specs/borrow/` rather than the single spec path so `borrow.coldStart.spec.ts` is included too.

e2e/mobile/utils/borrowUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getEnv } from "@shared/env";
+import { deleteSpeculos, launchSpeculos, registerSpeculos } from "@e2e/utils/speculosUtils";
+
+/**
+ * Leaves the collateral allowance at zero so the "Give approval" step is always required.
+ * The partner approves an exact amount, so a run that approves and then fails before
+ * supplying leaves a residue behind, and `resetLoanState` only closes positions.
+ */
+export async function resetCollateralAllowance(collateralAccount: TokenAccount) {
+  if (process.env.DISABLE_TRANSACTION_BROADCAST !== "0") return;
```
ypolishchuk-ledger, line 14: Dead in practice - `shouldRunBroadcastFlow` already returns false on the same condition (broadcastRotation.ts:38), so the describe is skipped, and this line never runs. It also diverges from the domain convention in borrowSetup.ts:19, where broadcastEnabled() logs why it skipped

e2e/mobile/utils/borrowUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const ownerAddress = collateralAccount.parentAccount?.address ?? collateralAccount.address;
```
ypolishchuk-ledger, line 16: I think this repeats getTokenAllowanceCommand's own logic, maybe use peekBorrowAddress or getAccountAddress?

libs/live-e2e-shared/src/borrow/borrowApi.ts
```diff
@@ -78,6 +80,24 @@ export async function postAction(body: ActionRequest): Promise<PartnerActionResp
   return { actionId, steps: steps.map((step, i) => normalizeStep(step, i)) };
 }
 
+/**
+ * Address the collateral allowance is granted to, read from the `supply` approve step
+ * the partner builds. It is an adapter the partner can redeploy, so it is resolved rather
+ * than pinned. Building an action does not broadcast anything.
+ */
+export async function resolveCollateralSpender(
```
ypolishchuk-ledger, line 88: - `EvmSignablePayload` exists in `./types` and is used exactly this way in `borrowFlow.ts:76` - use it here too.
- `JSON.parse` is unguarded; a non-JSON payload surfaces as a bare `SyntaxError` with no step context.
- 34/74 are unexplained. A short comment ("selector 0x+8, first arg is 32 bytes, address = last 20") or named constants would help.
- `startsWith` is case-sensitive; lowercase `data` before matching.
- Question: `normalizeStep` already lowercases and exposes `step.actionType (borrowApi.ts:59)`. Why sniff calldata rather than match `actionType === "approve"`? If `actionType` isn't trustworthy, please say so in the doc comment - that's valuable knowledge.

libs/live-e2e-shared/src/enum/Currency.ts
```diff
@@ -142,6 +142,14 @@ export class Currency {
     ],
     "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
   );
+  static readonly ETH_WBTC = new Currency(
+    "Wrapped Bitcoin",
+    "WBTC",
+    "ethereum/erc20/wrapped_bitcoin",
+    AppInfos.ETHEREUM,
+    [Network.ETHEREUM],
+    "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
```
ypolishchuk-ledger, line 151: I believe contract address is not in use anywhere

PR #21297, opened 2026-08-28
[LWDM] feat(signer-xrp): add DMK signer behind ldmkXrpSigner (DSDK-1453)

aussedatlo:
### 📝 Description

Adds `@ledgerhq/live-signer-xrp` so Ledger Live can drive the XRP app through
`@ledgerhq/device-signer-kit-xrp` instead of `@ledgerhq/hw-app-xrp`, behind the new
`ldmkXrpSigner` feature flag, with the legacy path kept as the fallback. Same shape as
`live-signer-cosmos` / `live-signer-solana`.

**New package `libs/live-signer-xrp/`**

- `LegacySignerXrp` — wraps `new Xrp(transport)`, straight delegation.
- `DmkSignerXrp` — builds via `new SignerXrpBuilder({ dmk, sessionId }).build()`, awaits the
  device-action observable with `lastValueFrom`, and maps `DeviceActionStatus.Error` through
  `_mapError` (`5515` → `LockedDeviceError`, `6985` / `6982` → `UserRefusedOnDevice`, otherwise
  the `_tag`). This is the `DmkSignerEth` / `DmkSignerAleo` shape rather than the cosmos one,
  because the XRP kit's DA errors carry `errorCode` at the top level, not under `originalError`.
  `skipOpenApp: true` on every call — Live opens the app itself.
- `src/types.ts` restates the `XrpSigner` contract. It is duplicated rather than imported
  because live-common depends on this package, not the other way round.

The signing blob is passed through unchanged: the XRP app prepends the `53545800` signing
prefix itself, so — unlike the software signer in `coin-tester-xrp` — the adapter must not add it.

**Wiring**

- `families/xrp/setup.ts` owns `setXrpLdmkEnabled` and the
  `isDmkTransport(transport) && _xrpLdmkFFEnabled` factory (the cosmos convention, so the apps'
  import path matches the existing ones); `families/xrp/signer.ts` consumes that factory and
  keeps its option-shape adapter for the generic coin framework.
- `ldmkXrpSigner` is declared in `libs/types-live/src/feature.ts` **and** registered in
  `shared/feature-flags/src/flags/team-live-devices/` — the latter is what `useFeature` /
  `getFeature` type against, and live-common's build fails without it.
- `useFeature("ldmkXrpSigner")` → `setXrpLdmkEnabled` in LLD `Default.tsx` and LLM `index.tsx`,
  next to the Solana and Cosmos calls.

### 🔗 Context

- **JIRA / GitHub issue**: [DSDK-1453](https://ledgerhq.atlassian.net/browse/DSDK-1453)
- Parent epic: [DSDK-1248](https://ledgerhq.atlassian.net/browse/DSDK-1248) — [TS][Signer] XRP Signer
- Equivalent Tron ticket: DSDK-1426

### ⚠️ Two things to agree on before merge

**1. The catalog pins a nightly.** `@ledgerhq/device-signer-kit-xrp` has no tagged release yet,
so the catalog pins `0.0.0-develop-20260828024102` (same precedent as
`@ledgerhq/device-signer-kit-hyperliquid`). Either we ship on the nightly, or this waits for a
tagged `signer-xrp` release.

**2. `ed25519` is not supported yet.** `GetAddressCommand` / `SignTransactionCommand` in that
build pin `p2` to `P2_SECP256K1`; `AddressOptions` / `TransactionOptions` carry no curve option
(`Curve` was described in DSDK-1437 / DSDK-1438 but has not shipped). `DmkSignerXrp` therefore
**throws** on `ed25519 = true` rather than silently signing on the wrong curve. This is not a
regression in practice: the generic coin framework calls `signTransaction` with an options
object, never a boolean, so Live always resolves `ed25519` to `undefined`. It becomes a real
option once the kit ships `Curve`.

### ✅ Checks run

- `nx build @ledgerhq/live-signer-xrp` and `nx build ledger-live-common` pass; `oxlint` / `oxfmt` clean.
- 18 new package tests (`DmkSignerXrp`, `LegacySignerXrp`) pass.
- 22 live-common `families/xrp` tests pass. The new `setup.test.ts` covers all four
  flag × transport combinations, including **DMK transport with the flag off → legacy signer**.
- LLD typecheck: only 2 pre-existing errors, both `Cannot find module release-notes.json`
  (a generated file skipped by my `--ignore-scripts` install). LLM typecheck: 45 pre-existing
  errors in 25 files, none in `src/index.tsx`.

### 🧪 Not covered here

- Receive and send verified on LLD / LLM against a device or Speculos with the flag on.
- A `coin-tester-xrp` run.
- `signTransaction` compared against the legacy client for a real blob larger than one APDU.
  The unit test asserts the whole blob reaches the kit — the kit's `SendSignTransactionTask`
  does the chunking (DSDK-1440), and that is exercised on its own side.

[DSDK-1453]: https://ledgerhq.atlassian.net/browse/DSDK-1453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/tidy-signers-ripple.md
+++ b/.changeset/tidy-signers-ripple.md
@@ -1,0 +1,9 @@
+---
+"@ledgerhq/live-signer-xrp": minor
+"@ledgerhq/live-common": minor
+"@ledgerhq/types-live": minor
+"ledger-live-desktop": minor
+"live-mobile": minor
+---
+
+Drive the XRP app through the Device Management Kit behind the `ldmkXrpSigner` feature flag, keeping `hw-app-xrp` as the fallback
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -75,6 +75,7 @@ shared-lib:
   - libs/live-signer-canton/**/*
   - libs/live-signer-evm/**/*
   - libs/live-signer-solana/**/*
+  - libs/live-signer-xrp/**/*
   - libs/live-wallet/**/*
   - libs/promise/**/*
   - libs/psbtv2/**/*
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -428,6 +428,7 @@ libs/live-signer-zcash/                                            @ledgerhq/liv
 libs/live-signer-celo/                                             @ledgerhq/live-devices
 libs/live-signer-cosmos/                                           @ledgerhq/live-devices
 libs/live-signer-hyperliquid/                                      @ledgerhq/live-devices
+libs/live-signer-xrp/                                              @ledgerhq/live-devices
 libs/ledgerjs/packages/hw-transport/                               @ledgerhq/live-devices
 libs/live-dmk-speculos/                                            @ledgerhq/live-devices
 apps/ledger-live-desktop/src/renderer/hooks/useDeeplinking/handlers/manager.handler.ts  @ledgerhq/live-devices
--- a/apps/ledger-live-desktop/src/renderer/Default.tsx
+++ b/apps/ledger-live-desktop/src/renderer/Default.tsx
@@ -67,6 +67,7 @@ import {
   setSolanaTxcEnabled,
 } from "@ledgerhq/live-common/families/solana/setup";
 import { setCosmosLdmkEnabled } from "@ledgerhq/live-common/families/cosmos/setup";
+import { setXrpLdmkEnabled } from "@ledgerhq/live-common/families/xrp/setup";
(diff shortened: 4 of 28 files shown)


aussedatlo:
Superseded by #21326 — branch renamed to `feat/DSDK-1453-dmk-signer-xrp`, same commit plus tests bringing `DmkSignerXrp` / `LegacySignerXrp` to 100% coverage.

PR #21312, opened 2026-08-31
fix(e2e): wait for the transfer bottom sheet to settle (QAA-1522)

jeportie:
### Checklist

- [x] `npx changeset` was attached.
- [ ] **Covered by automatic tests.** _This change is itself E2E test code, and there is no unit test for a Detox page object. It is verified by a filtered Mobile E2E run on this branch ([run 33394787463](https://github.com/LedgerHQ/ledger-live/actions/runs/33394787463), 13 specs, iOS & Android), and the standing regression detector is the nightly flake report (`e2e/mobile/scripts/flake-report.mjs`), which measured this signature at 3/5 nightlies._
- [x] **Impact of the changes:**
      - Receive flows reached from the portfolio transfer bottom sheet (`createAccount`, `importCurrency`, `readOnlyFlows`)
      - Deposit flows, which enter through the same sheet
      - `wallet40Q2/portfolio.spec.ts`, the other caller of the changed page object
      - **iOS specifically:** the new wait requires 100% visibility, stricter than the 75% default. If an iOS layout leaves the sheet container marginally clipped this would newly time out, so both platforms are in the CI run.
      - No product code is touched — only `e2e/mobile/page/wallet/portfolio.page.ts`

### Description

**Problem.** `tapById` does no waiting at all — it is `getElementById(id).tap()`. `openReceiveDrawer` tapped the sheet's receive button on the line immediately after asking the sheet to open, so the tap raced the sheet's presentation animation. `toBeVisible()` is satisfied at **75%**, which a bottom sheet meets while it is still sliding.

That produced two different failures from one defect, both seen in the nightlies:

- `Test Failed: Action will not be performed because the target view does not match one or more of the following constraints:` — the button was present but still moving, so Espresso refused the action.
- `Test Failed: No views in hierarchy found matching: ... view.getTag() is "transfer-action-receive" and view has effective visibility <VISIBLE>` — the button had not mounted yet at all.

**Solution.** Wait for the sheet's own container at 100% visibility before tapping.

```ts
// before
async pressTransferBottomSheetReceiveButton() {
  await tapById(this.transferBottomSheetReceiveButton);
}

// after
private async waitForTransferDrawerSettled() {
  await waitForFullyVisibleById(this.transferDrawerContainerId);   // "transfer-drawer"
}

async pressTransferBottomSheetReceiveButton() {
  await this.waitForTransferDrawerSettled();
  await tapById(this.transferBottomSheetReceiveButton);
}
```

[...]

PR #21313, opened 2026-08-31
fix(e2e): wait for the quick-action row to settle (QAA-1524)

jeportie:
### Checklist

- [ ] **Covered by automatic tests.** _This change is itself E2E test code, and there is no unit test for a Detox page object. It is verified by a filtered Mobile E2E run on this branch ([run 33394790416](https://github.com/LedgerHQ/ledger-live/actions/runs/33394790416), 14 specs, iOS & Android), and the standing regression detector is the nightly flake report (`e2e/mobile/scripts/flake-report.mjs`), which measured this signature at 3/5 nightlies._
      - Buy entry point from the portfolio (`navigateToBuyFromPortfolioPage_BTC`) — the ticketed failure
      - Receive and deposit flows, which enter through `pressQuickActionTransferButton`
      - `wallet40Q2/portfolio.spec.ts`, the other caller of the changed page object
      - **iOS specifically:** the new wait requires 100% visibility, stricter than the 75% default. If an iOS layout leaves the quick-action row marginally clipped this would newly time out, so both platforms are in the CI run.
      - No product code is touched — only `e2e/mobile/page/wallet/portfolio.page.ts`

### Description

**Problem.** `tapById` does no waiting at all — it is `getElementById(id).tap()` — so the quick-action taps were issued the moment the portfolio rendered:

```ts
async pressQuickActionBuyButton() {
  await tapById(this.quickActionBuyButtonV4);   // no wait
}
```

The quick actions mount only after the portfolio's own data resolves, so on Android CI the tap could be matched against a hierarchy that did not contain the row yet:

```
Test Failed: No views in hierarchy found matching: View at index #0, of those matching
MATCHER(view.getTag() is "quick-action-buy" and view has effective visibility <VISIBLE>)
```

Worth reading precisely: the row was **absent from the hierarchy**, not merely under-visible. So this is not a case for a longer timeout — nothing was there to wait on being bigger.

**Solution.** Wait for the row's own container at 100% visibility before tapping.

```ts
private async waitForQuickActionsSettled() {
  await waitForFullyVisibleById(this.quickActionsCtasContainerId);   // "quick-actions-ctas"
}

async pressQuickActionBuyButton() {
  await this.waitForQuickActionsSettled();
  await tapById(this.quickActionBuyButtonV4);
}
```

The container being fully visible is the state that proves the row is mounted and laid out; the buttons inside it are positioned by then. Anchoring the wait rather than raising a timeout or adding a retry follows `e2e/mobile/docs/add-or-update-e2e.md` rules 11 and 12.

`quick-actions-ctas` already exists in product code as `QUICK_ACTIONS_TEST_IDS.ctas.container` and is rendered by `QuickActionsCtasView.tsx`, so no product-code change was needed.

Swap and transfer get the same wait as buy: same row, same bare tap. Note `pressQuickActionTransferButton` is the first step of `openReceiveDrawer`, so this also removes one source of noise from QAA-1522's failure path.

**Evidence.** 3/5 nightlies (2026-08-25, 2026-08-26, 2026-08-28), Android only, last seen 2026-08-28, and 30% of runs of the spec. The rate rose from the 2/5 at which the ticket was filed. One of the three nights (2026-08-25) is infra-suspect and should be weighed lightly; the other two are clean. `oxfmt`, `oxlint` and `pnpm typecheck` all clean.

### Context

- **JIRA or GitHub link**: [QAA-1524](https://ledgerhq.atlassian.net/browse/QAA-1524) (sub-task of [QAA-1445](https://ledgerhq.atlassian.net/browse/QAA-1445))

---

[QAA-1524]: https://ledgerhq.atlassian.net/browse/QAA-1524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/qaa-1524-quick-actions-settle.md
+++ b/.changeset/qaa-1524-quick-actions-settle.md
@@ -1,0 +1,14 @@
+---
+"ledger-live-mobile-e2e-tests": patch
+---
+
+Wait for the portfolio quick-action row to settle before tapping it (QAA-1524)
+
+`tapById` does not wait, so the quick-action taps were issued as soon as the
+portfolio rendered. The row mounts after the portfolio's own data resolves, so on
+Android CI the tap could land before it existed — `No views in hierarchy found
+matching ... quick-action-buy ... VISIBLE`, 3/5 nightlies.
+
+Each press now waits for the row's own container at 100% visibility, which proves
+the row is mounted and laid out. Anchoring the wait rather than lengthening a
+timeout or adding a retry, per `docs/add-or-update-e2e.md` rules 11 and 12.
--- a/.changeset/qaa-1524-quick-actions-settle.md
+++ b/.changeset/qaa-1524-quick-actions-settle.md
@@ -11,4 +11,4 @@ matching ... quick-action-buy ... VISIBLE`, 3/5 nightlies.
 
 Each press now waits for the row's own container at 100% visibility, which proves
 the row is mounted and laid out. Anchoring the wait rather than lengthening a
-timeout or adding a retry, per `docs/add-or-update-e2e.md` rules 11 and 12.
+timeout or adding a retry, per `e2e/mobile/docs/add-or-update-e2e.md` rules 11 and 12.
--- a/e2e/mobile/page/wallet/portfolio.page.ts
+++ b/e2e/mobile/page/wallet/portfolio.page.ts
@@ -41,6 +41,10 @@ export default class PortfolioPage {
   fearAndGreedTitle = "fear-and-greed-title";
   bottomSheetCloseButton = "bottom-sheet-header-close-button";
   marketBannerTitle = "market-banner-title";
+  // The row's own container, from QUICK_ACTIONS_TEST_IDS.ctas.container. Waiting on this at 100%
+  // visibility is what proves the row has mounted and settled; the buttons inside it are laid out
+  // by then, so a tap cannot land on a view that is still arriving.
+  quickActionsCtasContainerId = "quick-actions-ctas";
   quickActionTransferButtonV4 = "quick-action-transfer";
   quickActionSwapButtonV4 = "quick-action-swap";
   quickActionBuyButtonV4 = "quick-action-buy";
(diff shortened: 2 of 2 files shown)


jeportie:
## 🧪 Triggered test workflows

Manually dispatched on `fix/e2e-QAA-1524-quick-actions-settle` (commit 42f86106ade), scoped to the specs that tap a portfolio quick action:

| Workflow | Scope | Run |
|---|---|---|
| [Mobile] E2E Only | `receivePerTest,navigateToBuyFromPortfolioPage,deposit,wallet40Q2` filter — 14 specs, iOS & Android, nanoX | [run 33394790416](https://github.com/LedgerHQ/ledger-live/actions/runs/33394790416) |

Scope rationale: the diff touches only `e2e/mobile/page/wallet/portfolio.page.ts`, so Desktop E2E and Coin Tester are skipped — neither can be affected. The filter comes from the actual callers of the three changed methods, not just the ticketed spec: `navigateToBuyFromPortfolioPage` (the ticketed `quick-action-buy` failure), plus `receivePerTest` (3) and `deposit` (3) which go through `pressQuickActionTransferButton`, and `wallet40Q2` (7, including `portfolio.spec.ts`). `wallet40Q2` is a little wider than strictly required, kept as regression coverage on the page object being changed.

Run on **both platforms** on purpose: the flake is Android-only, but the new wait is `waitForFullyVisibleById` at **100%** visibility, stricter than the 75% default. An iOS layout that leaves the quick-action row marginally clipped would be a regression introduced here, and an Android-only run would miss it.

Filter verified against the spec tree before dispatch — every pattern matches at least one spec, so no job will be silently skipped.

.changeset/qaa-1524-quick-actions-settle.md
```diff
@@ -0,0 +1,14 @@
...
+Wait for the portfolio quick-action row to settle before tapping it (QAA-1524)
+
+`tapById` does not wait, so the quick-action taps were issued as soon as the
+portfolio rendered. The row mounts after the portfolio's own data resolves, so on
+Android CI the tap could land before it existed — `No views in hierarchy found
+matching ... quick-action-buy ... VISIBLE`, 3/5 nightlies.
+
+Each press now waits for the row's own container at 100% visibility, which proves
+the row is mounted and laid out. Anchoring the wait rather than lengthening a
+timeout or adding a retry, per `docs/add-or-update-e2e.md` rules 11 and 12.
```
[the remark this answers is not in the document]
jeportie: Good catch, fixed in bef658fc939.

The path was both wrong and ambiguous: there is no top-level `docs/add-or-update-e2e.md`, and there are two variants — `e2e/mobile/docs/add-or-update-e2e.md` and `e2e/desktop/docs/add-or-update-e2e.md` — so the bare path would have been misleading in the release notes even if it had resolved.

I also had the same wrong path in the sibling PR #21312's changeset, which you did not see from here; fixed there too in 2c5ac54d0ee.

e2e/mobile/page/wallet/portfolio.page.ts
```diff
@@ -391,18 +395,29 @@ export default class PortfolioPage {
     await waitForElementById(this.quickActionBuyButtonV4);
   }
 
+  // tapById does not wait, so these taps used to be issued the moment the portfolio rendered.
```
martijnhjk, line 398: if premature tap is the issue should visibility be an optional condition on the tap function instead?

martijnhjk (approved):

e2e/mobile/page/wallet/portfolio.page.ts
jeportie: Worth splitting, because the two tickets need different conditions.

For QAA-1524 the row was absent from the hierarchy, so waiting on the tapped button itself would have been enough. An option on `tapById` would fix this one.

For QAA-1522 it would not. The transfer sheet's controls mount before the sheet is presented, so they are matchable while it is still sliding. Waiting on the tapped button there passes while the view is still moving, which is the failure. The condition that works is a different element — the container, at 100% visibility.

So a flag on `tapById(id)` cannot express both cases, because it can only wait on `id`. A shared helper taking both — `tapWhenSettled(containerId, buttonId)` — could, and would also remove the repetition across the 8 call sites in this file. Not done here; worth a ticket if the pattern reaches a third page object.

PR #21332, opened 2026-08-31
[LWM] feat(send): open the mad in the new send flow on lwm

dilaouid:
### 📝 Description

Generic Send entries on mobile still opened the legacy account picker (`SendCoin`) even when the new send flow was enabled, so toggling `send_flow` / `newSendFlow` had no visible effect. This PR routes those entries (transfer drawer without an asset, and the quick-action Send CTA) through `useOpenSendFlow`, which opens the MAD to pick an account, then continues into the new send flow, matching LWD

https://github.com/user-attachments/assets/553eb59d-a0ae-4174-ba11-ed73e0177483

### 🔗 Context

- **[JIRA / GitHub issue](https://ledgerhq.atlassian.net/browse/LIVE-36733)**: 
- **ADR** (if any):

### DIFF
--- a/.changeset/good-rats-push.md
+++ b/.changeset/good-rats-push.md
@@ -1,0 +1,6 @@
+---
+"live-mobile": minor
+"ledger-live-mobile-e2e-tests": minor
+---
+
+feat(send): open the MAD for the new send flow on LWM
--- a/apps/ledger-live-mobile/src/mvvm/features/QuickActions/components/QuickActionsCtas/__tests__/useQuickActionsCtasViewModel.test.ts
+++ b/apps/ledger-live-mobile/src/mvvm/features/QuickActions/components/QuickActionsCtas/__tests__/useQuickActionsCtasViewModel.test.ts
@@ -1,10 +1,15 @@
-import { renderHook } from "@tests/test-renderer";
+import { act, renderHook, withFlagOverrides } from "@tests/test-renderer";
 import { genAccount } from "@ledgerhq/ledger-wallet-framework/mocks/account";
 import { getCryptoCurrencyById } from "@domain/entity-currency-crypto";
 import type { Account } from "@ledgerhq/types-live";
 import type { State } from "~/reducers/types";
+import { NavigatorName, ScreenName } from "~/const";
+import { track } from "~/analytics";
 import { useQuickActionsCtasViewModel } from "../useQuickActionsCtasViewModel";
 import { QUICK_ACTIONS_TEST_IDS } from "../../../testIds";
+
+const mockNavigate = jest.fn();
+const mockHandleOpenSendFlow = jest.fn();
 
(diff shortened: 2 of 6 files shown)


deepyjr (dismissed):

VicAlbr:
Reviewed the two `e2e/mobile/**` files:

🚨 The new-flow assertion can never run

`e2e/mobile/specs/wallet40Q2/portfolio.spec.ts:96-100` — `isNewSendFlowEnabled()` always
resolves to `false` in this spec, so the `else` branch always wins and line 97 is dead code:

- `newSendFlow` defaults to `enabled: false, params: { families: [] }` —
  `shared/feature-flags/src/flags/team-coin-integration/newSendFlow.ts:9-12`
- the e2e `defaultFlags` in `getMergedFeatureFlags` doesn't set `newSendFlow` —
  `e2e/mobile/utils/featureFlagUtils.ts:96-118`
- `app.init` here only passes `FF_LWM_WALLET_40_Q2` — `portfolio.spec.ts:26-30`

So the PR ships a behaviour change with zero e2e coverage of it, and a regression in the new
flow would stay green forever.

**Fix** — pin the flag and drop the branch, same shape as
`e2e/mobile/specs/send/newSendFlow/sendATOM.spec.ts:9`. `FF_NEW_SEND_FLOW_ENABLED` is already
in the file you edited (`featureFlagUtils.ts:71`):

```diff
     await app.init({
       userdata: "skip-onboarding",
       speculosApp: CURRENCY.speculosApp,
-      featureFlags: FF_LWM_WALLET_40_Q2,
+      featureFlags: { ...FF_LWM_WALLET_40_Q2, ...FF_NEW_SEND_FLOW_ENABLED },
     });
```

```diff
     await app.portfolio.pressTransferBottomSheetSendButton();
-    // Send has no asset context here, so the new flow opens the MAD to pick
-    // an account while the legacy flow lands on its own account picker
-    if (await isNewSendFlowEnabled()) {
-      await app.modularDrawer.checkSelectAssetPage();
-    } else {
-      await app.send.expectFirstStep();
-    }
+    await app.modularDrawer.checkSelectAssetPage();
```

That also makes `isNewSendFlowEnabled` unused — I'd drop it from
`featureFlagUtils.ts:183-193` rather than leave an untested helper behind. (Note it also
shadows the name of the app-side value in `useTransferDrawerViewModel`, which this PR
removes — easy to confuse later.)

If the legacy entry point still needs covering, it belongs in its own spec with the flag
pinned off. An `if/else` inside one test means the suite can't distinguish "new flow
regressed" from "flag was off".

With the flag pinned, the outcome depends on `newSendFlow.params.families`, so the tag list
at `portfolio.spec.ts:16` should pick up `@cosmos` / `@family-cosmos` (cf. `sendATOM.spec.ts:8`).

VicAlbr (dismissed):

mcayuelas-ledger (approved):

VicAlbr (approved):

PR #21334, opened 2026-08-31
feat(contacts): match incoming and outgoing operations (LIVE-36677)

mcayuelas-ledger:
### 📝 Description

Pay contact recency and transaction counts only looked at **outgoing** wallet operations. Incoming transfers never counted, so a contact you only received from looked unused.

This PR extends `@features/platform-contacts` so matching covers send **and** receive, without pulling `@ledgerhq/types-live` into the platform package.

**Approach**

- New `contactOperations/` module (replaces `outgoingOperations/`).
- `toContactOperation` maps a wallet op to `IN` (match on `senders`) or `OUT` (match on `recipients`); other types are ignored.
- `findContactOperationsToAddresses` keeps operations whose counterparty matches a contact address **on the same currency**. Token rows fold onto the parent chain.
- `summarizeContactOperationsByContact`: `txCount` is IN + OUT; `lastSentAt` stays OUT-only (sort still “last sent, then last added”).
- Existing outgoing helpers remain as thin wrappers for callers that still need them.

**Usage**

```ts
const ops = findContactOperationsToAddresses(contact.addresses, operations);
const summaries = summarizeContactOperationsByContact(accounts, contacts);
```

No UI in this PR.

**Tests**

Unit coverage for the mapper, matcher (same address on two networks must not collide), and summaries.

### 🔗 Context

- **JIRA / GitHub issue**: [LIVE-36677](https://ledgerhq.atlassian.net/browse/LIVE-36677)
- **ADR** (if any):

[LIVE-36677]: https://ledgerhq.atlassian.net/browse/LIVE-36677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


features/platform/contacts/src/contactOperations/findContactOperationsToAddresses.web.test.ts
```diff
@@ -0,0 +1,73 @@
+import { findContactOperationsToAddresses } from "./findContactOperationsToAddresses";
```
tonykhaov: nit pick but we can drop the .web extension for tests

it can be done in a cleanup PR 👍

features/platform/contacts/src/contactOperations/findContactOperationsToAddresses.web.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    const op = outgoing({ recipients: ["0xdeadbeef00000000000000000000000000000000"] });
+
+    expect(
+      findContactOperationsToAddresses([{ address: ethAddress, currencyId: "ethereum" }], [op]),
+    ).toEqual([]);
+  });
+
+  it("returns no matches when there are no addresses", () => {
+    expect(findContactOperationsToAddresses([], [incoming(), outgoing()])).toEqual([]);
+  });
```
tonykhaov: could be nice to have a test case with mixed list: IN and OUT going transfers wdyt?

```ts
it("keeps matching operations and drops the rest", () => {
    const matchIn = incoming({ id: "in-hit" });
    const matchOut = outgoing({ id: "out-hit" });
    const miss = outgoing({
      id: "out-miss",
      recipients: ["0xdeadbeef00000000000000000000000000000000"],
    });
    expect(
      findContactOperationsToAddresses(
        [{ address: ethAddress, currencyId: "ethereum" }],
        [miss, matchIn, matchOut],
      ),
    ).toEqual([matchIn, matchOut]);
  });
```

mcayuelas-ledger: Yep indeed

tonykhaov (dismissed):

tonykhaov (approved):

